PR #5, opened 2019-03-04
Add Javascript files to PHPCS scan

philipjohn:
### All Submissions:

To make full use of the VIP Go standard we should include JS, as there are some handy rules in the Go standard that check for potential XSS vulnerabilties.


claudiulodro:
I've pushed an update to exclude the `dist` folder from the rules. Everything else seems to work well. The JSHint error I mention above is unrelated to this. Between this and Prettier we should have more than enough JS linting.

PR #6, opened 2019-03-06
Plugin management utility class

### LINKED ISSUE #3: Plugin installer class
Create a class to manage installation and activation of plugins from wordpress.org and third-party sources.

claudiulodro:
### All Submissions:

Closes #3 .

I've developed a class to handle general plugin management tasks: install/uninstall, activate/deactivate, etc. The class is designed so that we can just use it without having to worry about whether the underlying host is VIP Go or a different host:
- It handles installation from both WordPress.org repos and from URLs.
- The class is designed to work in all environments, including VIP Go.
- The `activate` method will install a plugin first if needed (when possible), so the idea is to just use `activate` in the OBW and that will automatically handle installation and activation of plugins.
- The class checks to make sure plugins installation is allowed in order to prevent issues with VIP Go. If plugin installation is not allowed, it gracefully falls back to only managing activation of existing plugins.
- There are also some general purpose methods that will probably come in handy.
- I've introduced namespaces. Since we're supporting PHP 5.6+, we can take advantage of this and not have to prefix all our classes/function names.

Questions (these can be adressed in follow-up PRs or down the road if we encounter a situation):
- Do we need bulk plugin install functionality? Activation should be quick enough to activate a bunch of plugins in one session, but install won't be.
- Do we need background install functionality? This would prevent the need for "loading" screens while something is installing but introduce UI/development complexity as plugins may not be available to use right away.
- How (and whether) should we integrate VIP Go-specific functionality like [wpcom_vip_load_plugin](https://vip.wordpress.com/functions/wpcom_vip_load_plugin/)?

1. Run/examine unit tests to see usage.
2. Examine code.

### Other information:

### DIFF
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -82,13 +82,19 @@ class Plugin_Manager {
 	/**
 	 * Deactivate a plugin.
 	 *
-	 * @param string $plugin_file The path to the plugin file. e.g. 'newspack/newspack.php'.
+	 * @param string $plugin The plugin slug (e.g. 'newspack') or path to the plugin file. e.g. ('newspack/newspack.php').
 	 * @return bool True on success. False on failure.
 	 */
-	public static function deactivate( $plugin_file ) {
+	public static function deactivate( $plugin ) {
 		$installed_plugins = self::get_installed_plugins();
-		if ( ! in_array( $plugin_file, $installed_plugins ) ) {
+		if ( ! in_array( $plugin, $installed_plugins ) && ! isset( $installed_plugins[ $plugin ] ) ) {
 			return new WP_Error( 'newspack_plugin_not_installed', __( 'The plugin is not installed.', 'newspack' ) );
+		}
+
+		if ( isset( $installed_plugins[ $plugin ] ) ) {
+			$plugin_file = $installed_plugins[ $plugin ];
+		} else {
+			$plugin_file = $plugin;
 		}
 
 		if ( ! is_plugin_active( $plugin_file ) ) {
@@ -158,17 +164,23 @@ class Plugin_Manager {
 	/**
 	 * Uninstall a plugin.
 	 *
-	 * @param string $plugin_file The path to the plugin file. e.g. 'newspack/newspack.php'.
+	 * @param string $plugin The plugin slug (e.g. 'newspack') or path to the plugin file. e.g. ('newspack/newspack.php').
 	 * @return bool True on success. False on failure.
 	 */
-	public static function uninstall( $plugin_file ) {
+	public static function uninstall( $plugin ) {
 		if ( ! self::can_install_plugins() ) {
 			return new WP_Error( 'newspack_plugin_failed_uninstall', __( 'Plugins cannot be uninstalled.', 'newspack' ) );
 		}
 
 		$installed_plugins = self::get_installed_plugins();
-		if ( ! in_array( $plugin_file, $installed_plugins ) ) {
+		if ( ! in_array( $plugin, $installed_plugins ) && ! isset( $installed_plugins[ $plugin ] ) ) {
 			return new WP_Error( 'newspack_plugin_failed_uninstall', __( 'The plugin is not installed.', 'newspack' ) );
+		}
+
+		if ( isset( $installed_plugins[ $plugin ] ) ) {
+			$plugin_file = $installed_plugins[ $plugin ];
+		} else {
+			$plugin_file = $plugin;
 		}
 
 		// Deactivate plugin before uninstalling.
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -41,14 +41,9 @@ class Plugin_Manager {
 			require_once ABSPATH . 'wp-admin/includes/plugin.php';
 		}
 
-		$plugin_slug = $plugin;
-
-		// if it's a url, parse the slug from the URL.
-		if ( wp_http_validate_url( $plugin ) ) {
-			$plugin_slug = self::get_plugin_slug_from_url( $plugin );
-			if ( ! $plugin_slug ) {
-				return new WP_Error( 'newspack_invalid_plugin', __( 'Invalid plugin URL.', 'newspack' ) );
-			}
+		$plugin_slug = self::get_plugin_slug( $plugin );
+		if ( ! $plugin_slug ) {
+			return new WP_Error( 'newspack_invalid_plugin', __( 'Invalid plugin.', 'newspack' ) );
 		}
 
 		$installed_plugins = self::get_installed_plugins();
@@ -118,18 +113,24 @@ class Plugin_Manager {
 	}
 
 	/**
-	 * Parse a plugin slug from the URL to download a plugin.
+	 * Parse a plugin slug from the slug or URL to download a plugin.
 	 *
-	 * @param string $url The URL to a plugin zip file.
+	 * @param string $plugin A plugin slug or the URL to a plugin zip file.
 	 * @return string|bool Parsed slug on success. False on failure.
 	 */
-	public static function get_plugin_slug_from_url( $url ) {
-		if ( ! is_string( $url ) ) {
+	public static function get_plugin_slug( $plugin ) {
+		if ( ! is_string( $plugin ) || empty( $plugin ) ) {
 			return false;
 		}
 
-		$url = wp_http_validate_url( $url );
-		if ( ! $url || ! stripos( $url, '.zip' ) ) {
+		$url = wp_http_validate_url( $plugin );
+
+		// A plugin slug was passed in, so just return it.
+		if ( ! $url ) {
+			return $plugin;
+		}
+
+		if ( ! stripos( $url, '.zip' ) ) {
 			return false;
 		}
 
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -20,7 +20,8 @@ class Plugin_Manager {
 	 * @return bool
 	 */
 	public static function can_install_plugins() {
-		if ( ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT ) || ( defined( 'DISALLOW_FILE_MODS' ) && DISALLOW_FILE_MODS ) ) {
+		if ( ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT ) ||
+			( defined( 'DISALLOW_FILE_MODS' ) && DISALLOW_FILE_MODS ) ) {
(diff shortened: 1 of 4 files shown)


tests/unit-tests/plugin-manager.php
```diff
@@ -0,0 +1,130 @@
...
+	 */
+	protected $plugin_url = 'https://downloads.wordpress.org/plugin/hello-dolly.1.6.zip';
+
+	/**
+	 * Compatibility checks and clean up.
+	 */
+	public function setUp() {
+		// These tests can't run on environments where we can't install plugins e.g. VIP Go.
+		if ( ! Plugin_Manager::can_install_plugins() ) {
+			$this->markTestSkipped( 'Plugin installation is not allowed in the environment' );
```
philipjohn: Am I right in thinking this will skip _all_ tests if plugin (un-)install can't be performed? If so, we should probably make it conditional on the test, as not being able to (un-)install plugins doesn't preclude (de-)activation.

tests/unit-tests/plugin-manager.php
```diff
@@ -0,0 +1,130 @@
...
+	/**
+	 * Test the plugin-slug-from-URL parser.
+	 */
+	public function test_get_plugin_slug_from_url() {
+		$this->assertEquals( 'hello-dolly', Plugin_Manager::get_plugin_slug_from_url( 'https://downloads.wordpress.org/plugin/hello-dolly.1.6.zip' ) );
+		$this->assertEquals( 'hello-dolly', Plugin_Manager::get_plugin_slug_from_url( 'https://downloads.wordpress.org/plugin/hello-dolly.zip/' ) );
+		$this->assertEquals( 'hello-dolly', Plugin_Manager::get_plugin_slug_from_url( 'https://downloads.wordpress.org/plugin/hello-dolly.zip?foo=blah&1' ) );
+		$this->assertEquals( false, Plugin_Manager::get_plugin_slug_from_url( 'https://downloads.wordpress.org/plugin/hello-dolly' ) );
+		$this->assertEquals( false, Plugin_Manager::get_plugin_slug_from_url( 'https://downloads.wordpress.org/plugin/hello-dolly.tar' ) );
+		$this->assertEquals( false, Plugin_Manager::get_plugin_slug_from_url( 'hello-dolly' ) );
```
philipjohn: Should we add some more extreme inputs here? For example, what if we passed an;

* `''` (empty string),
* `bool`,
* `'WP_Error',
* `int`

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,270 @@
...
+ */
+class Plugin_Manager {
+
+	/**
+	 * Determine whether plugin installation is allowed in the current environment.
+	 *
+	 * @return bool
+	 */
+	public static function can_install_plugins() {
+		if ( ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT ) || ( defined( 'DISALLOW_FILE_MODS' ) && DISALLOW_FILE_MODS ) ) {
```
philipjohn: It might be more readable to split this conditional into several lines.

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,270 @@
...
+		return true;
+	}
+
+	/**
+	 * Get a simple list of all installed plugins.
+	 *
+	 * @return array of 'plugin_slug => plugin_file_path' entries for all installed plugins.
+	 */
+	public static function get_installed_plugins() {
+		$installed_plugins = array_reduce( array_keys( get_plugins() ), array( __CLASS__, 'reduce_plugin_info' ) );
```
philipjohn: [`get_plugins()`](https://developer.wordpress.org/reference/functions/get_plugins/) uses the object cache - could that cause an issue when we [try to refresh the list](https://github.com/Automattic/newspack-plugin/pull/6/files#diff-044b9390f497ea129e19496244893c6eR63)?

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,270 @@
...
+		if ( empty( $installed_plugins ) ) {
```
philipjohn: Could `$installed_plugins` be anything other than an array at this point? I'm wondering whether this check is necessary.

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,270 @@
...
+			$installed_plugins = [];
+		}
+		return $installed_plugins;
+	}
+
+	/**
+	 * Parse a plugin slug from the URL to download a plugin.
+	 *
+	 * @param string $url The URL to a plugin zip file.
```
philipjohn: Could we be clever about this and allow anything that we can determine is a WP.org plugin URL and work out the actual ZIP URL from that?

claudiulodro:
Thanks for the feedback @philipjohn!

includes/class-plugin-manager.php
claudiulodro: It should be OK. After installing [the plugin cache is flushed](https://github.com/Automattic/newspack-plugin/pull/6/files#diff-044b9390f497ea129e19496244893c6eR253).

tests/unit-tests/plugin-manager.php
claudiulodro: It will skip all those tests if plugin installation is not allowed, even activation tests. The reason is because the activation tests can't run without first installing a plugin to test activation with. I'll investigate whether there's a good way of handling that situation.

includes/class-plugin-manager.php
claudiulodro: We could; it's just going to be e.g. `woo-gutenberg-products-block` for `https://wordpress.org/plugins/woo-gutenberg-products-block/`. I don't know if that's going to be a feature we need, since we already know the plugin slug for all .org plugins (it's `woo-gutenberg-products-block` for the plugin I just mentioned).

I'll add it if needed in a future PR, though. Good thinking and very feasible. :)

claudiulodro:
I've made requested and other changes:
- Added more test cases and hardened `Plugin_Manager::get_plugin_slug_from_url`.
- Split conditional in `Plugin_Manager::can_install_plugins` to improve readability.
- Improved error handling by returning WP_Errors with descriptive error messages.
- Added the ability to deactivate and uninstall plugins by slug. Previously this wasn't available, which is why your uninstall testing with Hello Dolly didn't work as expected. This change should increase the flexibility of the methods and make your expected test case pass.
- When uninstalling plugins, deactivate them first. This will resolve the error you saw when uninstalling Hello Dolly.

I've added unit tests for these changes where applicable.

tests/unit-tests/plugin-manager.php
claudiulodro: We can resolve the issue with our CI architecture, if e.g. Travis sets up a site with VIP Go configurations and a plugin preinstalled, the tests could be tweaked or we could have separate tests for that environment. I'll hold off on modifying this for now, though, since it wouldn't be super helpful without that environment.

jeffersonrabb:
If a plugin has been previously installed through wp-admin dashboard, installing via `Plugin_Manager::activate();` leads to duplicate instances of the plugin installed. To replicate:

- Navigate to `Plugins->Add New` and install AMP plugin. 
- In `class-newspack.php` add an action for `admin_init` hook.
- In the new function: `Plugin_Manager::activate( 'amp-wp' );` 
- Refresh Plugins page and you should see two instances of the AMP plugin.

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,302 @@
...
+	 */
+	public static function activate( $plugin ) {
+		if ( ! function_exists( 'get_plugins' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+
+		$plugin_slug = $plugin;
+
+		// if it's a url, parse the slug from the URL.
+		if ( wp_http_validate_url( $plugin ) ) {
```
jeffersonrabb: I expected to be able to install a plugin from a .zip file in the filesystem by passing in a local path, e.g. `Plugin_Manager::activate( '/srv/www/wordpress-develop/public_html/build/wp-content/plugins/newspack-plugin/amp-wp-develop.zip' )`, but only an absolute URL (or slug) can be used. I'm not sure we need to support this third approach, but if we do end up with some kind of a Newspack distro this might be a way to improve installation speed by avoiding unnecessary network activity. What do you think?

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,302 @@
...
+
+		if ( ! function_exists( 'plugins_api' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+		}
+
+		$plugin_info = plugins_api(
+			'plugin_information',
+			[
+				'slug'   => $plugin_slug,
+				'fields' => [
```
jeffersonrabb: Are the fields necessary? I'm getting the same result with just:

```
$plugin_info = plugins_api(
	'plugin_information',
	[ 'slug' => $plugin_slug ]
);
```

includes/class-plugin-manager.php
```diff
@@ -0,0 +1,302 @@
...
+	 *
+	 * @param string $url The URL to a plugin zip file.
+	 * @return string|bool Parsed slug on success. False on failure.
+	 */
+	public static function get_plugin_slug_from_url( $url ) {
+		if ( ! is_string( $url ) ) {
+			return false;
+		}
+
+		$url = wp_http_validate_url( $url );
```
jeffersonrabb: Very, very minor nit, but it struck me as weird that `wp_http_validate_url()` is being used both inside this function and in the logic where this function is called. With a slight rewrite we could have a `get_plugin_slug()` utility function that returns a clean slug from input that could be a slug or URL, which might save us a bunch of unnecessary logic down the road.

includes/class-plugin-manager.php
claudiulodro: Interesting idea. I would prefer not to have that feature if we don't actively need it, as it introduces a new category of security concerns we need to watch out for. Let's wait and add this feature if/when we need it. :)

jeffersonrabb: Agreed—YAGNI

includes/class-plugin-manager.php
claudiulodro: Those fields will limit the amount of information that the API call to WP.org returns. With the fields, the API will not transmit a bunch of extra info back. The end result is the same, since all the extra plugin info we don't need gets discarded, but it's faster with the fields.

includes/class-plugin-manager.php
claudiulodro: Good thinking. Much cleaner. I've made this change in https://github.com/Automattic/newspack-plugin/pull/6/commits/f08916b39cc2c3d7160582f1667b6bbbbf7a0500

PR #7, opened 2019-03-12
Add readme

claudiulodro:
A fairly basic readme with some general info about Newspack. We can add more details, links to docs, etc. in the future. I'll open a PR for the readme at the theme repo once this gets merged, so I don't have to make requested changes in two places at once.

Anything missing from here that should be added?
Anything not decided yet here that should be removed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Welcome to the Newspack plugin repository on GitHub. Here you can browse the source, look at open issues and keep track of development. We also recommend everyone [follow the Newspack blog](https://newspack.blog/) to stay up to date about everything happening in the project.
 
+The Newspack plugin provides tools and guidance for setting up and managing the important features and plugins a modern newsroom needs.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,27 @@
-# newspack-plugin
-The Newspack plugin. https://newspack.blog/
+# Newspack
+
+Welcome to the Newspack plugin repository on GitHub. Here you can browse the source, look at open issues and keep track of development. We also recommend everyone [follow the Newspack blog](https://newspack.blog/) to stay up to date about everything happening in the project.
+
+Newspack is an open-source publishing platform built on WordPress for small to medium sized news organizations. It is an “opinionated” platform that stakes out clear, best-practice positions on technology, design, and business practice for news publishers.
```
philipjohn: Should we use "plugin" here rather than "platform" as the latter is a much broader thing than just what's in this repo?

claudiulodro: The plugin and theme are part of the Newspack "platform", and I'm trying to give a 2-sentence overview of the whole thing. I think a couple sentences about the plugin specifically is a good idea, and should resolve some confusion. I'll make that update.

PR #12, opened 2019-04-04
Add SQL test data

laurelfulford:
### All Submissions:

This PR adds a SQL dump of test data, to be used for unit and visual tests.

At the moment, the content includes the [Theme Unit Test](https://codex.wordpress.org/Theme_Unit_Test) posts, a couple longer posts, with Gutenberg, the Newspack plugin, and Newspack Blocks enabled, as well as the Newspack theme. The one user included is VVV's default user.

It will need to be iterated on, but should work as a starting point for now!


claudiulodro:
In the interest of getting stuff done, let's keep it here for now. If it becomes an issue down the line we can always move this data somewhere else. :)

PR #15, opened 2019-04-25
[Try & don't merge] Subscriptions Wizard

claudiulodro:
**_The architecture of our WP Admin screens has changed since I began work on this, so this PR is just for testing flows and identifying flow/UI issues. Please ignore the code; it's rough, and I won't clean it up since the implementation is going to be JS-based soon._**

### To test:
1. Activate WooCommerce, WooCommerce Subscriptions, WooCommerce Name Your Price, and WooCommerce One Page Checkout plugins.
2. Navigate to Newspack > Subscriptions in WP Admin:
Screen Shot 2019-04-25 at 8 20 33 AM
3. Create a subscription:
Screen Shot 2019-04-25 at 8 21 55 AM
Screen Shot 2019-04-25 at 8 22 09 AM
4. Click subscription name/image to view subscription:
Screen Shot 2019-04-25 at 8 23 33 AM
5. Hover over card and click 'Delete' to delete subscription:
Screen Shot 2019-04-25 at 8 24 55 AM

## Discussion questions:
- How do we want to approach the product category for these subscriptions? The default category is "Uncategorized", and with no further customization subscriptions will display "Category: Uncategorized" and "Home / Uncategorized / Valued Donor". The *easiest* solution would be to change the name of the Uncategorized category to something else. The *most flexible* solution would be to create a new category for these to live in or add category selection to the screen somehow.
- Do we want short description, regular description, both, or none? They can help explain benefits of membership, but people need to write content and if done poorly it can distract from purchasing. Here is a screenshot of the above subscription but with short description and description added:
Screen Shot 2019-04-25 at 8 35 19 AM
The short description is to the right of the image. The description is underneath. We'll need some sort of UI if we want either of those.


claudiulodro:
See #73

Issue #18, opened 2019-04-29
Components: Formatted Header


claudiulodro:
A heading with a description to be used at the top of all pages.

Example usage:
<img width="278" alt="add-subscription" src="https://user-images.githubusercontent.com/7317227/56912697-ad671a80-6a64-11e9-98e6-26841d414df8.png">

Design:
<img width="508" alt="header" src="https://user-images.githubusercontent.com/7317227/56912718-b7891900-6a64-11e9-8cc8-768fea0ab18d.png">

Existing implementations:
- https://github.com/Automattic/wp-calypso/tree/master/client/components/formatted-header

PR #28, opened 2019-05-01
Muriel-styled Card component.

### LINKED ISSUE #17: Components: Card
An empty container within which we render all of our wizard content, subscription previews, etc.

Usage:
add-subscription

Design:
card

Existing implementations:
- https://github.com/Automattic/wp-calypso/tree/master/client/components/card
- https://github.com/woocommerce/woocommerce-admin/tree/master/packages/components/src/card

jeffersonrabb:
### All Submissions:

Closes #17.

1. Build the components: `npm ci && npm run build:webpack`
2. In a test environment running the Newspack plugin, navigate to Newspack->Subscriptions.
3. Observe two Cards, each with a single Checkbox in it.

### Some questions for discussion

1. The max width of this component is set to 484px to match the widths in the Figma mockups. Is it appropriate for the Card to determine its own width? Or, should we be placing Cards in a container that defines the width? If the former, we might need to introduce a prop to control the width of the Card (e.g. `cardWidth=small|medium|large`).
2. The Header/Subheader described [here](https://github.com/Automattic/newspack-plugin/issues/18) could be worked into the Card component as parameters? 
3. In the Figma document, some of the Card examples have top and bottom padding of 24px, while others which contain only a single element have top/bottom padding of 17px. We may need to control this via a prop as well, unless we can come up with logic to set this programmatically.

### Other information:

### DIFF
--- a/assets/src/components/card/card.test.js
+++ b/assets/src/components/card/card.test.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Card from './';
+
+describe( 'Card', () => {
+	describe( 'basic rendering', () => {
+		it( 'should render a Card element with only one class', () => {
+			const card = shallow( <Card /> );
+			expect( card.hasClass( 'newspack-card' ) ).toBe( true );
+		} );
+	} );
+} );
--- a/assets/src/components/card/style.scss
+++ b/assets/src/components/card/style.scss
@@ -6,7 +6,7 @@
 	background: $muriel-white;
 	border-radius: 3px;
 	box-shadow: 0px 1px 1px rgba(0, 0, 0, 0.25);
-	margin-bottom: 10px;
+	margin: 0 auto 10px auto;
 	max-width: 440px;
 	padding: 24px;
 }
--- a/assets/src/components/card/style.scss
+++ b/assets/src/components/card/style.scss
@@ -1,0 +1,12 @@
+/**
+ * Card styles.
+ */
+
+.newspack-card {
+	background: $muriel-white;
+	border-radius: 3px;
+	box-shadow: 0px 1px 1px rgba(0, 0, 0, 0.25);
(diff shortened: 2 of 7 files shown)


jeffersonrabb:
Added a single, rudimentary test in https://github.com/Automattic/newspack-plugin/pull/28/commits/5599fea28f89a185970667c88907cdd3264d8da4. This component is extremely simple at the moment so more tests probably aren't necessary, but this establishes a testing approach for the React components and wizards we'll be developing. cc: @claudiulodro @iuravic @philipjohn

assets/src/components/card/card.test.js
```diff
@@ -0,0 +1,12 @@
...
+import Card from './';
+
+describe( 'Card', () => {
+	describe( 'basic rendering', () => {
+		it( 'should render a Card element with only one class', () => {
+			const card = shallow( <Card /> );
+			expect( card.hasClass( 'newspack-card' ) ).toBe( true );
+		} );
+	} );
+} );
```
claudiulodro: Looks good and test worked great. 👍 Thanks for setting it up.

assets/src/components/card/style.scss
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+.newspack-card {
+	background: $muriel-white;
+	border-radius: 3px;
+	box-shadow: 0px 1px 1px rgba(0, 0, 0, 0.25);
+	margin-bottom: 10px;
+	max-width: 440px;
+	padding: 24px;
+}
```
claudiulodro: I think it may be good to add `margin-left: auto; margin-right: auto`. Currently cards are left-aligned, and we want them centered.

jeffersonrabb: Done in https://github.com/Automattic/newspack-plugin/pull/28/commits/c6231e5b5c29fe206c18e1c1126a3e34eba3bdb8.

PR #29, opened 2019-05-02
Image upload/media modal component

### LINKED ISSUE #23: Components: Image uploader
A link that opens the media modal when clicked and allows uploading and setting of an image.

This component needs some more research around it. There should be existing work that we can piggyback somehow.

Example usage:
add-subscription

Design:
image-uploader

Existing implementations:
- ???

claudiulodro:
### All Submissions:

Closes #23 .

This PR introduces a media modal component. The component allows uploading images or selecting existing images using the media modal, and can be pre-populated with an image. I've also added a couple tests for it.

No good pre-existing component exists for us to use, so this is a React wrapper over the standard WP media modal functionality. This should work well for our use.

1. `npm ci` and `npm run build:webpack`
2. On the Subscriptions page you should see a card with prompt to "Add an image".
3. Test adding and removing images.

#### Discussion questions

- i18n. Should we do it right from the get-go? How should we do it? [Similar to how wc-admin does it](https://github.com/woocommerce/woocommerce-admin/blob/master/package.json#L34)?
- What should it look like once an image has been uploaded? We need to balance being able to see the image with not taking up a ton of space. This is the current design I did (max width of 240px for the image preview):
Screen Shot 2019-05-02 at 11 45 48 AM

### Other information:

### DIFF
--- a/assets/src/components/imageUpload/image_upload_24px.svg
+++ b/assets/src/components/imageUpload/image_upload_24px.svg
@@ -1,0 +1,14 @@
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs>
+        <path d="M21.3333333,0 L2.66666667,0 C1.2,0 0,1.2 0,2.66666667 L0,21.3333333 C0,22.8 1.2,24 2.66666667,24 L21.3333333,24 C22.8,24 24,22.8 24,21.3333333 L24,2.66666667 C24,1.2 22.8,0 21.3333333,0 Z M21.3333333,21.3333333 L2.66666667,21.3333333 L2.66666667,2.66666667 L21.3333333,2.66666667 L21.3333333,21.3333333 Z M10.9466667,17.1066667 L14.6133333,12.3866667 L19.3333333,18.6666667 L4.66666667,18.6666667 L8.33333333,13.96 L10.9466667,17.1066667 Z" id="path-1"></path>
+    </defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="image_upload_24px">
+            <mask id="mask-2" fill="white">
+                <use xlink:href="#path-1"></use>
+            </mask>
+            <g id="Shape"></g>
+            <rect id="Rectangle-path" fill="#A2AAB2" fill-rule="nonzero" mask="url(#mask-2)" x="-4" y="-4" width="32" height="32"></rect>
+        </g>
+    </g>
+</svg>
--- a/assets/src/components/imageUpload/index.js
+++ b/assets/src/components/imageUpload/index.js
@@ -1,6 +1,14 @@
+/**
+ * Image uploader component.
+ */
+
+/**
+ * WordPress dependencies
+ */
 import { Component } from '@wordpress/element';
 import { Button } from '@wordpress/components';
 import { data } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Image select/upload button and modal.
@@ -17,11 +25,6 @@ class ImageUpload extends Component {
 			image_id : props.image_id || 0,
 			image_url: props.image_url || '',
 		}
-
-		this.openModal          = this.openModal.bind( this );
-		this.handleImageSelect  = this.handleImageSelect.bind( this );
-		this.removeImage        = this.removeImage.bind( this );
-		this.renderImagePreview = this.renderImagePreview.bind( this );
 	}
 
 	/**
@@ -36,18 +39,16 @@ class ImageUpload extends Component {
 	/**
 	 * Open the WP media modal.
 	 */
-	openModal( evt ) {
-		evt.preventDefault();
-
+	openModal = () => {
 		if ( this.state.frame ) {
 			this.state.frame.open();
 			return;
 		}
 
 		this.state.frame = wp.media( {
-			title: 'Select or upload image',
+			title: __( 'Select or upload image' ),
 			button: {
-				text: 'Select'
+				text: __( 'Select' )
 			},
 			library: {
 				type: 'image'
@@ -62,7 +63,7 @@ class ImageUpload extends Component {
 	/**
 	 * Update the state when an image is selected from the media modal.
 	 */
-	handleImageSelect() {
+	handleImageSelect = () => {
 		let attachment = this.state.frame.state().get( 'selection' ).first().toJSON();
 		this.setState( {
 			image_id: attachment.id,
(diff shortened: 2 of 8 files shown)


assets/src/components/imageUpload/style.scss
```diff
@@ -0,0 +1,40 @@
+/**
+ * Image uploader styles.
+ */
+
+.newspack-image-upload {
+	button {
+		color: $muriel-hot-pink-500;
```
jeffersonrabb: Nice!

assets/src/components/imageUpload/style.scss
```diff
@@ -0,0 +1,40 @@
...
+
+		img {
+			max-width: 240px;
+			height: auto;
+		}
+	}
+
+	&.no-image {
+		.add-image {
+			&:before {
```
jeffersonrabb: This is nice. Makes me think we should revisit the [`infoButton` styling](https://github.com/Automattic/newspack-plugin/blob/master/assets/src/components/infoButton/style.scss#L5-L12) as a `:before` to unify our approach?

package.json
```diff
@@ -11,6 +11,7 @@
     "@babel/core": "^7.4.0",
     "@wordpress/blocks": "^6.2.0",
     "@wordpress/browserslist-config": "^2.2.2",
+    "@wordpress/components": "^7.3.1",
```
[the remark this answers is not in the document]
claudiulodro: The only reason it's explicitly installed is because the unit tests need it, otherwise something like this happens:
```
 FAIL  assets/src/components/imageUpload/imageUpload.test.js
  ● Test suite failed to run

    Cannot find module '@wordpress/components' from 'index.js'

       7 |  */
       8 | import { Component } from '@wordpress/element';
    >  9 | import { Button } from '@wordpress/components';
         | ^
      10 | import { data } from '@wordpress/data';
      11 | 
      12 | /**

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:229:17)
      at Object.<anonymous> (assets/src/components/imageUpload/index.js:9:1)
```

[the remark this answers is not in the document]
jeffersonrabb: Ah, yup, I get the same thing. This problem rings a bell and I'm going to try to get to the bottom of it, but this definitely isn't a blocker—later on we can do a separate PR that trims the dependencies.

assets/src/components/imageUpload/style.scss
claudiulodro: Yeah that sounds like a good idea. It keeps the markup and the style separate. I'll just add that small tweak to the revisions for this PR.

claudiulodro: I gave it a shot. Tooltip needs to wrap an element, (i.e. the thing you hover over over to display the tooltip), so we still need the `newspack-info-button` div. Since we still need that div, there isn't really any benefit to changing the style to use `:before`.

assets/src/components/imageUpload/index.js
```diff
@@ -0,0 +1,116 @@
...
+		if ( 'onChange' in this.props ) {
+			this.props.onChange( this.state );
+		}
+	}
+
+	/**
+	 * Open the WP media modal.
+	 */
+	openModal( evt ) {
+		evt.preventDefault();
```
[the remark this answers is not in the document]
claudiulodro: Force of habit. It's not necessary.

claudiulodro:
I've made the requested changes. Thanks for the feedback @jeffersonrabb. Great suggestions!

claudiulodro:
I've updated this with the latest jest fixes, so tests should be passing.

assets/src/components/imageUpload/image_upload_24px.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jeffersonrabb: `@wordpress/components` has a few SVG-related classes (`SVG`, `Path`, `G`) that offer improved accessibility. They're used pretty regularly for block icons and other assets. Do you think we should use these here? 

https://github.com/WordPress/gutenberg/tree/master/packages/components/src/primitives/svg

assets/src/components/imageUpload/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Image select/upload button and modal.
+ */
+class ImageUpload extends Component {
+
+	/**
+	 * Constructor.
+	 */
+	constructor( props ) {
```
jeffersonrabb: Tiny nit, but we could probably remove `props` from the function signature since it's not being used anymore.

assets/src/components/imageUpload/index.js
```diff
@@ -0,0 +1,99 @@
...
+		this.state.frame.on( 'select', this.handleImageSelect );
+		this.state.frame.open();
+	}
+
+	/**
+	 * Update the state when an image is selected from the media modal.
+	 */
+	handleImageSelect = () => {
+		const { onChange } = this.props;
+		const attachment = this.state.frame.state().get( 'selection' ).first().toJSON();
```
jeffersonrabb: This is just a question, but do you think there's any wisdom in avoiding jQuery in our plugins?

assets/src/components/imageUpload/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+		return (
+			<Fragment>
+				{ !! image && (
+				<div className="newspack-image-upload has-image">
+					<div className="image-preview">
+						<img src={ image.url } />
+					</div>
+					<Button className="remove-image" onClick={ this.removeImage }>
+						{ __( 'Remove image' ) }
```
jeffersonrabb: Regarding the styling of the `Remove image` button, I was wondering if the vertically centered placement came from Muriel, or if that was just how it happened to lay out when there's a placeholder image next to it?

claudiulodro: There was no design for that in Muriel or our designs, so I just "designed" that. Do you have a better-looking idea for that?

assets/src/components/imageUpload/index.js
claudiulodro: jQuery is definitely not going anywhere any time soon, but in general we don't need it. In this specific case, the media modal is built with jQuery, so there's no avoiding it.

assets/src/components/imageUpload/index.js
jeffersonrabb: No, not really—mostly just wanted to see if you had found it somewhere I wasn't looking, and if not to flag it for Sonja and Muriel folks to assess at a later date. Definitely no blocker here!

assets/src/components/imageUpload/image_upload_24px.svg
claudiulodro: Seems like it may be a good idea. I wasn't able to get it working properly in my experiment with it, though. If you want to make an issue, I can spend more time investigating this for our icons when I have a chance.

claudiulodro:
I think this should be good to go now.

assets/src/components/imageUpload/image_upload_24px.svg
jeffersonrabb: Done: https://github.com/Automattic/newspack-plugin/issues/49

PR #36, opened 2019-05-06
Muriel-styled Text/Number Input component.

### LINKED ISSUE #20: Components: Text/Number Input
Components for text and number inputs. We may or may not want separate components for text and number inputs, or one component that can handle both. Examining existing implementations should provide the answer.

Example usage:
add-subscription

Design:
text-input
text-input-2

Existing implementations:
- https://github.com/WordPress/gutenberg/tree/master/packages/components/src/text-control
- https://github.com/Automattic/wp-calypso/tree/master/client/components/forms/form-text-input

iuravic:
### All Submissions:

Closes #20 .

1. the Input component is demoed in the Subscription wizzard page, in its 3 states: containing a pre-entered value, containing no (empty) value, and disabled

### Other information:

### DIFF
--- a/assets/src/components/inputText/index.js
+++ b/assets/src/components/inputText/index.js
@@ -5,116 +5,81 @@
 /**
  * WordPress dependencies
  */
+import { TextControl, withFocusOutside } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 
-class InputText extends Component {
-	/**
-	 * Constructor.
-	 */
- 	constructor( props ) {
-		super( props );
-		const { label, value, disabled } = props;
-
-		this.state = {
-			label,
-			value,
-			disabled,
-			isFocused: false
+const InputText = withFocusOutside(
+	class extends Component {
+		/**
+		 * Constructor.
+		 */
+		constructor( props ) {
+			super( props  );
+			this.state = {
+				value: props.value,
+				isFocused: false
+			};
 		};
 
-		this.textInputRef = null;
-		this.wrapperRef = null;
-	};
-
-	/**
-	 * componentDidMount.
-	 */
-	componentDidMount() {
-		document.addEventListener('mousedown', this.handleClickOutside);
-	};
-
-	/**
-	 * componentWillUnmount.
-	 */
-	componentWillUnmount() {
-		document.removeEventListener('mousedown', this.handleClickOutside);
-	};
-
-	/**
-	 * Handle component onClick.
-	 */
-	handleOnClick = () => {
-		this.setState({ isFocused: true });
-		this.textInputRef.focus();
-	};
-
-	/**
-	 * Handle clickOutside component.
-	 */
-	handleClickOutside = event => {
-	  const { isFocused } = this.state;
-		if ( this.wrapperRef && ! this.wrapperRef.contains(event.target) && isFocused ) {
-			this.setState({ isFocused: false });
-		}
-	};
-
-	/**
-	 * Handle input value change.
-	 */
-	handleOnChange = event => {
-		this.setState({ value: event.target.value});
-	};
-
-	/**
-	 * Get input element's className depending on context.
-	 */
-	getTextInputClassName = ( disabled, isEmpty, isActive ) => {
-		let className = "with-value";
-		if ( disabled ) {
-			className = "disabled";
-		} else if ( isEmpty ) {
-			className = "empty";
-		} else if ( isActive ) {
-			className = "active";
+		/**
+		 * withFocusOutside: handles component onClick.
+		 */
+		handleFocusOutside() {
+			this.setState( { isFocused: false } );
 		}
 
-		return className;
-	};
+		/**
+		 * Handle component onClick.
+		 */
+		handleOnClick = () => {
+			this.setState( { isFocused: true } );
+		};
 
-	/**
-	 * Render.
-	 */
-	render() {
-		const { label, value, disabled, isFocused } = this.state;
-		const isEmpty = ! value;
-		const isActive = isFocused && ! disabled;
-		const displayLabelDiv = ! isActive && ! isEmpty;
-		const classNameInput = this.getTextInputClassName( disabled, isEmpty, isActive );
+		/**
+		 * Handle component onChange.
+		 */
+		handleOnChange = value => {
+			this.setState( { value } );
+		};
 
-		return (
-				<div
-					className="newspack-input-text"
+		/**
+		 * Get component's className depending on state and context.
+		 */
+		getTextInputClassName = ( disabled, isEmpty, isActive ) => {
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
Discussed this a bit with @iuravic and the plan is rather than creating the Component from scratch to follow the "[Specialization](https://reactjs.org/docs/composition-vs-inheritance.html#specialization)"  pattern using the [`TextControl` component](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/text-control) in `@wordpress/components`.

iuravic:
Refactored to use the `TextControl` component in `@wordpress/components`. Also now using `withFocusOutside` from `@wordpress/components`, rather than detecting click-outside with own events. Restructured .scss as needed.

assets/src/wizards/subscriptions/index.js
```diff
@@ -9,6 +9,7 @@ import { Component, Fragment, render } from '@wordpress/element';
 import CheckboxInput from '../../components/checkboxInput';
 import Card from '../../components/card';
 import InputText from '../../components/InputText';
+import './style.scss';
```
iuravic: Added by resolving a merge conflict from master

assets/src/components/inputText/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+const InputText = withFocusOutside(
+	class extends Component {
+		/**
+		 * Constructor.
+		 */
+		constructor( props ) {
+			super( props  );
+			this.state = {
+				value: props.value,
```
jeffersonrabb: Similar to [this comment](https://github.com/Automattic/newspack-plugin/pull/29#discussion_r281025232) from a different PR, we should generally avoid setting state from props. The `value` of the component should be a prop, not state, and should be controlled by the parent component. To achieve this, I'd suggest removing `value` and `onChange` from the `TextControl` component in `render()`. Then, in `assets/src/wizards/subscriptions/index.js` either remove the `value` attribute entirely (a hardcoded value will make it impossible to type in the component), or better yet set the wizard up as a PoC demo of how the parent's state interacts with the component. Here's the bare minimum version of how this would look:

```
constructor() {
	super( ...arguments );
	this.state = {
		myText: '',
	}
}
/**
 * Render the example stub.
 */
render() {
	const { myText } = this.state;
	return(
		<InputText
			label="Text Input with value"
			value={ myText }
			onChange={ value => this.setState( { myText: value } ) }
		/>
	);
}
```

assets/src/components/inputText/index.js
```diff
@@ -0,0 +1,85 @@
...
+			const classNameInput= this.getTextInputClassName( disabled, isEmpty, isActive );
+
+			return (
+				<TextControl
+					className={ "newspack-input-text " + classNameInput }
+					placeholder={ label }
+					{ ...this.props }
+					value={ value }
+					onClick={ () => this.handleOnClick() }
+					onChange={ this.handleOnChange }
```
jeffersonrabb: As noted in another comment, I think you can omit `onChange` and leave it to the parent to define the `onChange` handler.

assets/src/components/inputText/index.js
```diff
@@ -0,0 +1,85 @@
...
+			const isEmpty = ! value;
+			const isActive = isFocused && ! disabled;
+			const classNameInput= this.getTextInputClassName( disabled, isEmpty, isActive );
+
+			return (
+				<TextControl
+					className={ "newspack-input-text " + classNameInput }
+					placeholder={ label }
+					{ ...this.props }
+					value={ value }
```
jeffersonrabb: As noted above, remove value attribute and leave that to the parent.

assets/src/components/inputText/index.js
```diff
@@ -0,0 +1,85 @@
...
+			const { value, isFocused } = this.state;
+			const { label, disabled } = this.props;
+			const isEmpty = ! value;
+			const isActive = isFocused && ! disabled;
+			const classNameInput= this.getTextInputClassName( disabled, isEmpty, isActive );
+
+			return (
+				<TextControl
+					className={ "newspack-input-text " + classNameInput }
+					placeholder={ label }
```
jeffersonrabb: Wouldn't there be a good argument for distinct `label` and `placeholder` values? If so, we might want to remove this and leave it to the parent to define, if desired?

iuravic: If I'm not mistaken, the Muriel implementation uses only one of them (label or placeholder), not both:
Muriel Input Control

The way I saw it:
- in the 2nd example on the image, we use the <label> above the <input> (which are children of the TextComponent from @wordpress/components)
- and to implement the 1st and 3rd examples, I've hidden the <label> and set the placeholder on the <input>. But I might have taken a wrong approach there, and maybe should have enlarged the <label> and hidden the <input> (in which case the placeholder prop would not be used). If you believe this should be changed, please just comment and I will. At this point, it seems to me that the effect is exactly the same, but it might be semantically cleaner.

assets/src/components/inputText/index.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * Muriel-styled Text/Number Input.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { TextControl, withFocusOutside } from '@wordpress/components';
+import { Component } from '@wordpress/element';
+
```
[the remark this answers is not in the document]
iuravic: Thank you for the tip. Imported.

assets/src/components/inputText/index.js
iuravic: Excellent points, and it makes a lot of sense! Updated now.
Lesson learned: I'm keeping in mind the uni-directional control flow from Parent to the Child component, and keeping the Child controllable by the Parent.

assets/src/components/inputText/index.js
iuravic: Updated

assets/src/components/inputText/index.js
iuravic: Updated

iuravic:
> Looks good! The main thing I'm confused about is whether this actually is the right styling for the component. In the Figma designs @rickybanister and Sonja Leix prepared, the text inputs look like this `(...)` When Sonja returns next week let's have her determine the correct styling and if necessary make changes in a future PR.
Sound OK?

Yes, sounds good, let's review what we need exactly together with Sonja. We might want to update and change this component right here, or create a new one which displays an icon.
And, just to mention, the no. 5 looks like it might be a Select dropdown, so that's a different component.

assets/src/components/inputText/index.js
jeffersonrabb: I see what you're saying. Let's put a pin in this and revisit in a later PR when we have determined if the styling here is in line with the Muriel approach Sonja had in mind.

PR #25, opened 2019-04-30
Build pipeline, Murielized Checkbox, and Subscriptions Wizard stub

claudiulodro:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes #22.
Closes #16.

This PR sets up the build pipeline for the wizards. A "wizard" is a self-contained admin-screen feature, and each wizard is its own single-page app. The modularity and independence of the wizards should make it easier to develop wizards in parallel and will prevent overhead from trying to coordinate a bunch of wizards in one master single-page app.

#### Webpack

The webpack config is heavily based on the [newspack-blocks](https://github.com/Automattic/newspack-blocks/blob/master/webpack.config.js) webpack config, but with some custom stuff to make it work better for components and wizards.

Within the `src` directory there are two directories: `components` and `wizards`.

The webpack slurps up all of the component css files and builds them into one master `components.css` file that gets enqueued automatically for each wizard.

The webpack builds JS and CSS files for each wizard, named `[name].js` and [name]-style.css` respectively. These can be enqueued as needed in the wizard's PHP class.

All webpack output goes into the `dist` folder.

#### Creating components

Examine the `components/checkboxInput` directory to see a working example.
1. Create a new directory in the `src/components` directory.
2. Add `index.js` and `style.scss` files.
3. In `index.js` make a "wrapper" component for an existing Gutenberg component if possible, adding a class for styling and making any other tweaks we need to the behavior.
4. In `style.scss`, add styling to make the component match the Muriel design styles.

#### Creating wizards

Examine the `assets/src/wizards/subscriptions` directory and `includes/wizards/class-subscriptions-wizard.php` file to see a working example.
1. Create a new directory in the `src/wizards` directory.
2. Add `index.js` and `style.scss` files.
3. Build/style the wizard frontend.
[...]

PR #37, opened 2019-05-06
Minor tweaks

claudiulodro:
This PR implements the suggestions from https://github.com/Automattic/newspack-plugin/pull/25#pullrequestreview-232478568 . The PR was merged since those were non-blocking suggestions, and I'm opening this PR before I forget about them.

Specifically:
- Use sass color variables instead of literal colors.
- Move inclusion of Wizard class to classes that extend it instead of the main Newspack class.
- Remove unnecessary module from webpack config.

The `package.json` dependencies mentioned in the review, we will re-approach down the line as part of a bigger dependency clean-up effort. There are some other dependencies @jeffersonrabb has mentioned we may be able to remove, but including unnecessary dependencies doesn't affect our product at the moment, so let's investigate that later.

### DIFF
--- a/assets/src/components/checkboxInput/style.scss
+++ b/assets/src/components/checkboxInput/style.scss
@@ -7,18 +7,18 @@
 	input[type="checkbox"] {
 		width: 22px;
 		height: 22px;
-		border: 2px solid #636D75;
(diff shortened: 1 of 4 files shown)


jeffersonrabb:
Rebased and pushed.

PR #38, opened 2019-05-06
Build: Simplify style build stack

ockham:
Having extra entry points for style files is a bit weird TBH :grimacing: `calypso-build` supports `import`ing SASS files out of the box, so this PR suggests using that technique instead.

Can be a first step towards dropping the webpack config altogether (which ironically will make adding back a babel config less messy. And adding a babel config is one way to fix failing tests).

### DIFF
--- a/includes/wizards/class-subscriptions-wizard.php
+++ b/includes/wizards/class-subscriptions-wizard.php
@@ -56,7 +56,7 @@ class Subscriptions_Wizard extends Wizard {
 		wp_register_style(
 			'newspack-subscriptions-wizard',
 			Newspack::plugin_url() . '/assets/dist/subscriptions.css',
-			null,
+			[ 'wp-components' ],
 			filemtime( dirname( NEWSPACK_PLUGIN_FILE ) . '/assets/dist/subscriptions.css' )
 		);
 		wp_style_add_data( 'newspack-subscriptions-wizard', 'rtl', 'replace' );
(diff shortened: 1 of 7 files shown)


includes/wizards/class-subscriptions-wizard.php
```diff
@@ -45,19 +45,19 @@ public function enqueue_scripts_and_styles() {
...
+			true
 		);
 
 		wp_register_style(
 			'newspack-subscriptions-wizard',
-			Newspack::plugin_url() . '/assets/dist/subscriptions-style.css',
-			[ 'newspack-components' ],
-			filemtime( dirname( NEWSPACK_PLUGIN_FILE ) . '/assets/dist/subscriptions-style.css' )
+			Newspack::plugin_url() . '/assets/dist/subscriptions.css',
+			null,
```
claudiulodro:
```suggestion
			[ 'wp-components' ],
```

The stylesheets have a dependency on the `wp-components` stylesheet (see the code you removed in `class-wizard.php` ). That stylesheet provides the base style layer, and we just tweak some colors and stuff to make things look Muriel-designed. 

For example, we've lost a lot of tooltip styling now:
Screen Shot 2019-05-06 at 2 28 40 PM

jeffersonrabb: Oops sorry about that - I didn't study closely enough. Added the dependency back in https://github.com/Automattic/newspack-plugin/pull/38/commits/0dce134ad698022e46a0377a05f0c554aeffa723. Does that work, or do you think it should be in the `Wizard` class instead?

PR #39, opened 2019-05-07
Fix jest tests

claudiulodro:
### All Submissions:

I've added some rules to package.json and a babel file to get the jest tests working again. Everything should build and test the JS properly now. The babel file is based on the [calypso-build](https://github.com/Automattic/wp-calypso/blob/master/packages/calypso-build/babel/default.js) one.

1. `npm ci` and `npm run test`
2. Tests should run and pass.

### Other information:

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,34 +1,7 @@
 const config = {
 	presets: [
-		[
-			'@babel/env',
-			{
-				useBuiltIns: 'entry',
-				corejs: 2,
-				// Exclude transforms that make all code slower, see https://github.com/facebook/create-react-app/pull/5278
-				exclude: [ 'transform-typeof-symbol' ],
-			},
-		],
-		'@babel/react',
-		'@babel/preset-typescript',
-		require( '@automattic/calypso-build/babel/wordpress-element' )
-	],
-	plugins: [
-		'@babel/plugin-proposal-class-properties',
-		'@babel/plugin-proposal-export-default-from',
-		'@babel/plugin-proposal-export-namespace-from',
-		'@babel/plugin-syntax-dynamic-import',
-		[
-			'@babel/transform-runtime',
-			{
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
This is working well for me, thanks for running it down! I'm going to leave it for @ockham to look over before approving, to determine if the elements of `babel.config.js` should be moved into `calypso-build`.

ockham:
I'd suggest a bit of a different approach, for various reasons:

- `babel.config.js` will only be used by tests but not by the actual build, which is somewhat unintuitive, and confusing for contributors.
- `babel.config.js` is rather complex -- we should be able to use something simpler, based on the `default` preset that comes with `@automattic/calypso-build`

I'm hoping to use something like this:

```js
module.exports = {
	presets: [
		'@automattic/calypso-build/babel/wordpress-element',
		'@automattic/calypso-build/babel/default',
	],
	env: {
		test: {
			presets: [ [ '@babel/env', { targets: { node: 'current' } } ] ],
		},
	},
};
```

(The crux is that we need to transpile ES Modules `import` statements to CommonJS `require`s for node/Jest.)

This is still failing currently, apparently Jest isn't setting the `test` env like it's supposed to. I need to investigate a bit more :thinking:

jeffersonrabb:
After a bit of flailing in the dark here's a finding that may or may not be relevant:

On this branch, with this `babel.config.js` build and  tests execute successfully:

```const config = {
	presets: [
		'@automattic/calypso-build/babel/default',
		'@automattic/calypso-build/babel/wordpress-element',
	]
};
module.exports = config;
```

However with this one (order of presets is reversed), build works, tests fail as before:

```
const config = {
	presets: [
		'@automattic/calypso-build/babel/wordpress-element',
		'@automattic/calypso-build/babel/default',
	]
};
module.exports = config;
```

PR #40, opened 2019-05-07
Formatted header component

### LINKED ISSUE #18: Components: Formatted Header

claudiulodro:
### All Submissions:

Closes #18 .

Adds the FormattedHeader component. The code is strongly influenced by [the Calypso component of the same name.](https://github.com/Automattic/wp-calypso/blob/master/client/components/formatted-header/index.jsx) I've also made a couple tweaks to the Subscriptions demo page, for better demoing.

I'll remove the Do Not Merge label and add tests to this as soon as the tests are working again. :)

1. `npm ci` `npm run build:webpack`
2. Navigate to the Subscriptions screen.
Screen Shot 2019-05-07 at 11 34 17 AM

### Other information:

### DIFF
--- a/assets/src/components/formattedHeader/index.js
+++ b/assets/src/components/formattedHeader/index.js
@@ -1,0 +1,34 @@
+/**
+ * Formatted header for titling screens/sections.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { Component } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+class FormattedHeader extends Component {
+	/**
+	 * Render.
+	 */
+	render() {
+		const { headerText, subHeaderText } = this.props;
+		const classes = 'newspack-formatted-header' + ( !! subHeaderText ? ' has-subheader' : '' );
+
+		return (
+			<header className={ classes }>
+				<h1 className="newspack-formatted-header__title">{ headerText }</h1>
+				{ subHeaderText && (
+					<p className="newspack-formatted-header__subtitle">{ subHeaderText }</p>
+				) }
+			</header>
+		);
+	}
+}
+
+export default FormattedHeader;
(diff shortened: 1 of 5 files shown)


assets/src/components/formattedHeader/index.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+import './style.scss';
+
+class FormattedHeader extends Component {
+	/**
+	 * Render.
+	 */
+	render() {
+		const { headerText, subHeaderText } = this.props;
+		const classes = 'newspack-formatted-header' + ( !! subHeaderText ? ' has-subheader' : '' );
```
jeffersonrabb: You could use the [`classnames`](https://jedwatson.github.io/classnames/) utility for a (maybe?) more elegant approach here. In this case your approach is fewer characters and requires no dependency so it's superior, but just mentioning this for future cases where we get into complex className logic. It would look like this:

`import classnames from 'classnames';`

then...

```
const classes = classnames(
			'newspack-formatted-header',
			!! subHeaderText ? 'has-subheader' : null
		);
```

assets/src/components/formattedHeader/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+		return (
+			<header className={ classes }>
+				<h1 className="newspack-formatted-header__title">{ headerText }</h1>
+				{ subHeaderText && (
+					<p className="newspack-formatted-header__subtitle">{ subHeaderText }</p>
```
jeffersonrabb: Why not use `h2` for  more semantic tagging?

jeffersonrabb:
Why the DO NOT MERGE tag?

assets/src/components/formattedHeader/index.js
claudiulodro: Yeah, I'm sure that utility will come in handy a lot, so let's just start using it. 👍

claudiulodro:
I've made the requested revisions. The DO NOT MERGE tag was because I wanted to add unit tests before we merged it. I've added unit tests (and updated the branch so the tests work), so we should be good to merge.

PR #46, opened 2019-05-09
Add Button Component

### LINKED ISSUE #19: Components: Buttons
Primary (pink) and secondary (gray) buttons for performing actions.

Example usage:
manage-subscriptions

Design:
buttons

Existing implementations:
- https://github.com/WordPress/gutenberg/tree/master/packages/components/src/button
- https://github.com/Automattic/wp-calypso/tree/master/client/components/button

philipjohn:
### All Submissions:

Adds a component providing a `NewspackButton`, inheriting largely from the default `Button` element but overridden with Muriel styling. Includes styling for the various [props available](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/button#props).

Closes #19 .

1. Apply the patch
2. In WP Admin, navigate to Newspack > Subscriptions
3. You will see a pink "Continue" button
4. In development, change the `isPrimary` prop of `<NewspackButton>` within [wizards/subscriptions/index.js](https://github.com/Automattic/newspack-plugin/blob/3ff96ef964c5b07a43f2a20b5bdd5b898ee317a5/assets/src/wizards/subscriptions/index.js#L38) to one of the [other values](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/button#props) to see the different styles.

### Other information:


philipjohn:
Note that the test is currently failing with the following output:

```
 FAIL  assets/src/components/button/button.test.js
  ● Test suite failed to run

    Cannot find module '@wordpress/components' from 'index.js'

       7 |  */
       8 | import { Component } from '@wordpress/element';
    >  9 | import { Button } from '@wordpress/components';
         | ^
      10 | 
      11 | class NewspackButton extends Component {
      12 | 

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:229:17)
      at Object.<anonymous> (assets/src/components/button/index.js:9:1)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        2.024s
Ran all test suites matching /button/i.
```

assets/src/components/button/index.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Muriel-styled buttons.
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { Component } from '@wordpress/element';
+import { Button } from '@wordpress/components';
+
```
jeffersonrabb: The stylesheet needs to be imported here:

```suggestion
import './style.scss';
```

assets/src/components/button/index.js
```diff
@@ -0,0 +1,28 @@
...
+class NewspackButton extends Component {
+
+	/**
+	 * Render.
+	 */
+	render( props ) {
+		const { value } = this.props;
+		return (
+			<div className="newspack-button">
```
jeffersonrabb: Is there any reason we need a wrapper, as opposed to simply adding the class to the `Button` component, like this?

`<Button className="newspack-button" { ...this.props } />`

assets/src/components/button/index.js
```diff
@@ -0,0 +1,28 @@
...
+				<Button { ...this.props } >
```
jeffersonrabb: The `@wordpress/components` `Button` component uses its children as the content, not `value`, and I think we should mirror this here. `children` is a prop, so this would mean changing the JSX here to simply:

```suggestion
				<Button className="newspack-button" { ...this.props } />
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -77,6 +78,8 @@ class SubscriptionsWizard extends Component {
 						disabled
 					/>
 				</Card>
+
+				<NewspackButton isPrimary value="Continue" />
```
jeffersonrabb: Per comment above about using `children` rather than `value`, this would change to:

```suggestion
				<NewspackButton isPrimary>Continue</NewspackButton>
```

jeffersonrabb: Let's move the button into the `Card` so we see it side-by-side with other example components.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,70 @@
+/**
+ * Button styles,
+ */
+
+.newspack-button {
+	.components-button {
```
jeffersonrabb: If we get rid of the wrapper `div`, we can get rid of the `.components-button` selector and apply all styles to `.newspack-button`

claudiulodro:
In addition to @jeffersonrabb's points, I think we should have some standard way of naming our components (either all prefixed or none prefixed with `Newspack` or `Muriel`). It's going to be confusing if we have e.g. `Card` and `NewspackButton` components in our library. I'd prefer non-prefixed names or maybe names prefixed with `Muriel`, as that will make it easy for us to release our components as a stand-alone package if we want. Thoughts?

jeffersonrabb:
> Note that the test is currently failing with the following output:

@philipjohn you can solve this by adding `@wordpress/components` as an NPM dependency, as @claudiu did [here](https://github.com/Automattic/newspack-plugin/pull/29/commits/067983bd8d285140762e91463598baf3b53b83d9). I suspect we'll eventually be able to remove this, since `@wordpress/components` is already part of `calypso-build` but for I don't have a solution and this is a fine stopgap.

assets/src/components/button/index.js
philipjohn: Nope, I think I just started out with this for some reason and it got left over.

assets/src/components/button/style.scss
philipjohn: We're inheriting styles from `.components-button` rather than duplicating here, so removing the selector will break our styling a bit.

philipjohn:
Yep, I assumed we'd hit a conflict by using `Button` but that solution looks good :+1:

philipjohn:
Thanks for the hint @jeffersonrabb - I now have a working test!

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
+/**
+ * Button styles,
+ */
+
+
+.newspack-button {
```
jeffersonrabb: As per the changes made in https://github.com/Automattic/newspack-plugin/pull/47, let's change the className to `muriel-button`. We'll need to make this change in the component file and the tests as well as here.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+				color: $muriel-gray-0;
+			}
+
+			&:active:enabled {
+				background-color: $muriel-hot-pink-700;
+				color: $muriel-white;
+			}
+		}
+
+		&.is-tertiary {
```
jeffersonrabb: Seems to me there are two ways to go with these components: 1) support all the variants that the Gutenberg component offers, making this a truly robust and reusable implementation, or 2) start out by supporting only the use case we need and add the other variants later. We've been taking the pragmatic 2nd approach in order to get moving quickly, and as we develop confidence around the components we'll go back fill in the gaps. Robust support for `Button` would mean, at minimum, supporting `isPrimary | isLarge | isSmall | isTertiary | isToggled | isBusy | isDefault | isLink | isDestructive`, although I believe the only one the Newspack mockups call for is `isPrimary`. In the spirit of this, let's limit our styles to the required variant(s) to avoid confusion later. All of which is a long-winded way of saying, let's remove the `is-tertiary` block.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+
+		&.is-button {
+			background-color: $muriel-white;
+			border: none;
+			border-radius: 3px;
+			cursor: pointer;
+			height: 40px;
+			max-width: 360px;
+			max-height: 40px;
+			margin-top: 13px;
```
jeffersonrabb: I think we may have margin and padding mixed up here. Let's do the following (all within `&.is-button`);

```
padding: 13px 0 9px 0;
margin: 14px auto;
```

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+		font-weight: 500;
+		color: $muriel-gray-700;
+		text-shadow: none;
+
+		&.is-button {
+			background-color: $muriel-white;
+			border: none;
+			border-radius: 3px;
+			cursor: pointer;
+			height: 40px;
```
jeffersonrabb: We shouldn't set explicit height or max-height, instead this should be determined by the size of the button + the vertical padding. For the sizing, how about this, all within `&.is-button`:

```
display: block;
height: auto;
max-width: 360px;
width: 100%;
```

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+			}
+
+			&:active:enabled {
+				border-width: 2px 1px 1px;
+			}
+		}
+
+		&.is-primary {
+			background-color: $muriel-hot-pink-500;
+			border-color: #992053;
```
jeffersonrabb: We have no border, so `border-color` can be removed.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+			border: none;
+			border-radius: 3px;
+			cursor: pointer;
+			height: 40px;
+			max-width: 360px;
+			max-height: 40px;
+			margin-top: 13px;
+			margin-bottom: 9px;
+
+			&:hover {
```
jeffersonrabb: These hovers can be removed since we have no borders.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+
+			&:active:enabled {
+				border-width: 2px 1px 1px;
+			}
+		}
+
+		&.is-primary {
+			background-color: $muriel-hot-pink-500;
+			border-color: #992053;
+			box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.25);
```
jeffersonrabb: Any reason the `box-shadow` is in the `is-primary` block vs. `is-button`? I'm not certain, but would guess that the shadow should be the same for primary and other buttons?

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+	.components-button {
+	    padding: 0 20px 1px;
+		text-align: center;
+
+		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", sans-serif;
+		font-size: 14px;
+		line-height: 18px;
+		font-weight: 500;
+		color: $muriel-gray-700;
+		text-shadow: none;
```
jeffersonrabb: After removing the wrapper `div`, `text-shadow` will need to move into the `&.is-button` block.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+ * Button styles,
+ */
+
+
+.newspack-button {
+	.components-button {
+	    padding: 0 20px 1px;
+		text-align: center;
+
+		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", sans-serif;
```
jeffersonrabb: Let's keep all font definitions out of the components. We might revisit this later after talking to the Muriel team, but for now let's keep them out for consistency.

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,83 @@
...
+			&:active:focus:enabled {
+				border: 0;
+			}
+		}
+
+	}
+}
+
+
+.newspack-card {
```
jeffersonrabb: Let's remove all of the `newspack-card` special-case styles. I don't think we need it, and even if we did we're aiming for reusable components so this would be more appropriate elsewhere. Other CSS changes I've suggested elsewhere should cover all of these styles.

assets/src/components/button/style.scss
claudiulodro: Agreed, we only have 2 button styles in our designs from what I can tell: primary and secondary.

assets/src/components/button/index.js
philipjohn: Yep, this is necessary in order to align the button properly when within a card component, hence [this code](https://github.com/Automattic/newspack-plugin/pull/46/files#diff-050da08994b8f727f91b4b19128cd8bdR73)

assets/src/components/button/style.scss
philipjohn: See [this comment](https://github.com/Automattic/newspack-plugin/pull/46/files#r283315652)

assets/src/components/button/style.scss
philipjohn: Yep, these were mostly straight from the Figma dimensions to make sure everything was exactly as presented there. I've tweaked things around a little in [0bd35eb32973dddeab931f11ba4ba1fc61ff7bbc] including padding.

assets/src/components/button/style.scss
philipjohn: There's not really a shadow on the others - like `isDefault` - but there is a strong bottom border instead.

assets/src/components/button/style.scss
philipjohn: I think we need this to make sure that the button is aligned properly when within a card component, to match up with what's presented in Figma

assets/src/components/button/style.scss
philipjohn: That will make the shadow appear on the default button style, too, which I don't think we really want to do

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,53 @@
...
+ * Button styles,
+ */
+
+.muriel-button {
+	text-align: center;
+	font-size: 14px;
+	line-height: 18px;
+	font-weight: 500;
+	color: $muriel-gray-700;
+	text-shadow: none;
```
jeffersonrabb: Now that Text Shadow is in the `.is-primary` block can this be removed?

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,53 @@
...
+
+	&.is-button {
+		display: inline-block;
+		background-color: $muriel-white;
+		border-color: #CCCCCC;
+		border-radius: 3px;
+		cursor: pointer;
```
jeffersonrabb: Tiny nit: don't think you need `cursor: pointer` here since you get it for free from `.components-button`

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,53 @@
...
+	&.is-primary {
+		border: none;
+		background-color: $muriel-hot-pink-500;
+		box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.25);
+		color: $muriel-white;
+		text-shadow: none;
+
+		&:hover,
+		&:focus:enabled {
+			background-color: $muriel-hot-pink-400;
```
jeffersonrabb: To my eye, the black line under the Primary button in hover state looks funny, and since we don't have a model for hover states (that I'm aware of?) we can make styling decisions here. My comment is only about Primary—I like the hover state of the Default button with the same shadow... Adding `box-shadow: inherit;` to the focus block would resolve it.  

Screen Shot 2019-05-14 at 8 16 58 AM

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,53 @@
...
+	text-shadow: none;
+
+	&.is-button {
+		display: inline-block;
+		background-color: $muriel-white;
+		border-color: #CCCCCC;
+		border-radius: 3px;
+		cursor: pointer;
+		height: auto;
+		padding: 13px;
```
jeffersonrabb: The buttons outside of the card seem a little tight, maybe could use more horizontal padding? I bring this up mostly because I'm working on the Checklist component right now which requires non-full-width buttons, and in the mockups these have a bit more horizontal spacing. First screenshot is what I see on this branch and second is from Figma:

Screen Shot 2019-05-14 at 8 22 37 AM

Screen Shot 2019-05-14 at 8 25 02 AM

assets/src/components/button/style.scss
```diff
@@ -0,0 +1,53 @@
...
+
+		&:active:enabled {
+			background-color: $muriel-hot-pink-700;
+			color: $muriel-white;
+		}
+	}
+
+	&.is-centered {
+		display: block;
+		width: 100%;
```
jeffersonrabb: In the mockups the centered Primary button doesn't fill full width. Adding `max-width` would achieve this. In Figma the button is 360px wide so that minus horizontal padding should do the trick:

Screen Shot 2019-05-14 at 8 26 01 AM

assets/src/components/button/style.scss
philipjohn: Copy pasta!

assets/src/components/button/style.scss
philipjohn: Yep, that makes sense. In [88826d5c343b4e580c6b7473704697683db34532] I've given more horizontal padding to those buttons, and reduced the vertical padding 'cause they looked weirdly fat otherwise. The centered buttons are then given different padding.

assets/src/components/button/style.scss
philipjohn: Good spot, padding is set at 13px so in [cfaff9b06b64c3805e992f17728dba01b9b634a6] I've set a max width of 334px.

philipjohn:
I also dropped centered button padding in [b4c318a0cbad52d88a52f6c4c37db07c34f2102b] because I thought they looked a bit fat compared to Figma designs.

PR #47, opened 2019-05-09
Standardize naming conventions

claudiulodro:
This PR standardizes the naming conventions for the components following these rules:
- Use the same JS class name as the Gutenberg component for Murielized components when applicable.
- Use the `muriel-` prefix for HTML classes.

In addition to making it easier to offer these components as stand-alone components in the future, this should make naming things easier and won't require getting creative with component names.

To test:
1. `npm ci`, `npm run build:webpack`, `npm run test`.
2. Go to the Subscriptions page. Observe nothing is broken or unstyled.

### DIFF
--- a/assets/src/components/infoButton/index.js
+++ b/assets/src/components/infoButton/index.js
@@ -5,6 +5,7 @@
 /**
  * WordPress dependencies
  */
+import { Component } from '@wordpress/element';
 import { Tooltip } from '@wordpress/components';
 
 /**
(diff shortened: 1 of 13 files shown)


assets/src/components/infoButton/index.js
```diff
@@ -5,20 +5,21 @@
...
+import { Component } from '@wordpress/element';
 import { Tooltip } from '@wordpress/components';
 
 /**
  * Internal dependencies
  */
 import './style.scss';
 
-class InfoButton extends Tooltip {
+class InfoButton extends Component {
```
claudiulodro: Also made this small tweak, as per #43.

PR #48, opened 2019-05-09
Progress Bar component

### LINKED ISSUE #32: Components: Checklist progress bar
A progress bar to indicate how many tasks in the checklist have been completed. 
The progress bar consists of a stepper, and conditionally shows a 'hide completed' toggle with an arrow.  

checklist-progress-bar-1

Once at least one item in the checklist has been completed a 'hide completed' toggle appears
checklist-progress-bar-2

### LINKED ISSUE #35: Components: Stepper
An indicator to indicate progress through a linear flow with more than one step. I'm not quite sure what the final UI design of this stepper is. Here are a few I found: 

Example from unified onboarding
Screenshot 2019-05-03 18 17 10

Woo onbaording flows
Screenshot 2019-05-03 18 18 26

claudiulodro:
### All Submissions:

Closes #35  .

Adds a Muriel progress bar component. I've followed the Muriel style for this instead of the rounded style in the screenshot in #32.

Screen Shot 2019-05-09 at 12 02 22 PM

1. `npm ci` `npm run build:webpack` `npm run test`
2. Go to Subscriptions page and observe the varieties of progress bar available:
Screen Shot 2019-05-09 at 12 03 45 PM

### Other information:

### DIFF
--- a/assets/src/components/progressBar/index.js
+++ b/assets/src/components/progressBar/index.js
@@ -1,0 +1,70 @@
+/**
+ * Progress bar for displaying visual feedback about steps-completed.
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { Component } from '@wordpress/element';
+
+/**
+ * Internal dependencies.
+ */
+import './style.scss';
+
+/**
+ * Progress bar.
+ */
+class ProgressBar extends Component {
+
+	/**
+	 * Get completion as a percentage.
+	 *
+	 * @return int
+	 */
+	getCompletionPercentage() {
+		let { completed, total } = this.props;
+		completed = parseInt( completed );
+		total = parseInt( total );
+
+		if ( ! total ) {
+			return 100;
+		}
+
+		return Math.min( Math.round( ( completed / total ) * 100 ), 100 );
+	}
+
+	/**
+	 * Render.
+	 */
+	render() {
+		const { label, completed, total, displayFraction } = this.props;
+
+		const barStyle = {
+			width: this.getCompletionPercentage() + '%',
+		}
+
+		return (
+			<div className="muriel-progress-bar">
+				{ ( label || displayFraction ) && (
+					<div className="muriel-progress-bar__headings">
+						{ label && (
+							<div className="muriel-progress-bar__label">
+								{ label }
+							</div>
+						) }
+						{ displayFraction && (
+							<div className="muriel-progress-bar__fraction">
+								{ completed }/{ total}
+							</div>
+						) }
+					</div>
+				) }
+
+				<div className="muriel-progress-bar__bar" style={ barStyle } ></div>
+			</div>
+		);
+	}
+}
+
+export default ProgressBar;
--- a/assets/src/components/progressBar/index.js
+++ b/assets/src/components/progressBar/index.js
@@ -20,18 +20,16 @@ class ProgressBar extends Component {
 	/**
 	 * Get completion as a percentage.
 	 *
+	 * @param  int completed The number of steps completed.
+	 * @param  int total     The total number of steps.
 	 * @return int
 	 */
-	getCompletionPercentage() {
-		let { completed, total } = this.props;
-		completed = parseInt( completed );
-		total = parseInt( total );
-
+	getCompletionPercentage( completed, total ) {
 		if ( ! total ) {
 			return 100;
 		}
 
-		return Math.min( Math.round( ( completed / total ) * 100 ), 100 );
+		return Math.max( Math.min( Math.round( ( completed / total ) * 100 ), 100 ), 0 );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


jeffersonrabb:
Note: rebase: https://github.com/Automattic/newspack-plugin/compare/37910cb0f0102b2186277e58aba4f51f6c4b1194..db7c7c88e8c407abe8e60d48510bad3ae8ce232d

assets/src/components/progressBar/index.js
```diff
@@ -0,0 +1,70 @@
...
+ * Progress bar.
+ */
+class ProgressBar extends Component {
+
+	/**
+	 * Get completion as a percentage.
+	 *
+	 * @return int
+	 */
+	getCompletionPercentage() {
```
jeffersonrabb: Suggestion: instead of getting these values from props rewrite the function to take two parameters (completed, total) and return a value. It would be more easily testable and allows for sanitization to occur in only one place (more on this in a separate comment).

assets/src/components/progressBar/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+		return Math.min( Math.round( ( completed / total ) * 100 ), 100 );
+	}
+
+	/**
+	 * Render.
+	 */
+	render() {
+		const { label, completed, total, displayFraction } = this.props;
+
```
jeffersonrabb: Suggestion: sanitize the completed value to check that it is a number and is between 0-`total`, e.g.:

`const cleanCompleted = Math.max( 0, Math.min( ( parseInt( completed ) || 0 ), parseInt( total ) ) );`

In my testing it's possible to set the `completed` prop to a number greater than total, a number less than 0, or a string that yields `NaN` from `parseInt()`.

assets/src/components/progressBar/style.scss
```diff
@@ -0,0 +1,36 @@
+/**
+ * Progress bar styles.
+ */
+
+.muriel-progress-bar {
+	margin: 16px 0;
+	position: relative;
+	z-index: 1;
```
jeffersonrabb: This is kind of a fly-by, but given our aspirations to publish these components as a package for other consumers this might be a good moment to discuss `z-index` strategy. `z-index` normalization seems to be a major pain point for packaging Calypso components, since the mixin that handles  them works across the entirety of Calypso styles, which returns values which might be completely inappropriate in a different consumer App. Since this is (I believe) the first `z-index` in our component work, it would be great to strategize.

assets/src/components/progressBar/style.scss
```diff
@@ -0,0 +1,36 @@
...
+		font-weight: 500;
+	}
+
+	.muriel-progress-bar__fraction {
+		margin-left: auto;
+	}
+
+	.muriel-progress-bar__bar {
+		height: 4px;
+		background-color: $muriel-hot-pink-500;
```
jeffersonrabb: I wonder if a red-ish color like this tracks as indicating errors rather than progress? Let's revisit with Sonja, but I'm wondering if it should be something more  like `$muriel-blue-400`

assets/src/wizards/subscriptions/index.js
```diff
@@ -92,6 +93,15 @@ class SubscriptionsWizard extends Component {
 						disabled
 					/>
 				</Card>
+				<Card>
+					<FormattedHeader
+						headerText="Progress bar"
+					/>
+					<ProgressBar completed="2" total="3" />
+					<ProgressBar completed="2" total="5" label="Progress made" />
+					<ProgressBar completed="0" total="5" displayFraction />
```
jeffersonrabb: Not suggesting this be committed, but as a way to test this component I set the `completed` attribute to `{ inputTextValue2 }`, which lets me change progress in real time by changing the value of the second text input.

assets/src/components/progressBar/style.scss
```diff
@@ -0,0 +1,36 @@
...
+		justify-content: space-between;
+		font-weight: 500;
+	}
+
+	.muriel-progress-bar__fraction {
+		margin-left: auto;
+	}
+
+	.muriel-progress-bar__bar {
+		height: 4px;
```
jeffersonrabb: For bonus points we could animate the progress bar?

```
-webkit-transition: width 0.25s; /* Safari */
  transition: width 0.25s;
```

claudiulodro:
I believe the bar design with the steps in it is supposed to be something similar to this:

Screen Shot 2019-05-10 at 10 36 42 AM

I'm unclear whether it should fill up like a progress bar, or if only the current wizard screen should be the highlighted chunk of that bar.

If it's not a big deal I'd prefer to use the bar in this PR for everything, though. :) That would likely just mean adding support for multiple colors to this bar.

assets/src/components/progressBar/style.scss
claudiulodro: [From here:](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)
> The z-index values of its child stacking contexts only have meaning in [the] parent.

As far as I can tell, we should be OK in all situations here unless e.g. the wizard's stylesheet overrides the z-indexing of this component. In this specific situation, the z-index is required so that the gray part of the bar is placed under the pink part of the bar.

Can you provide an example of the issues you describe? I'm having a hard time coming up with a strategy without knowing what issues to avoid.

claudiulodro:
I've made the requested revisions:
- Isolated `getCompletionPercentage` and added some tests.
- Cleaned variables better in `render`.
- Animated progress bar (looks cool 👍)

I don't believe the z-index here requires a greater z-index strategy, and we can follow up with any bar color issues in a separate PR.

PR #52, opened 2019-05-13
Button Centering (Do Not Merge)

jeffersonrabb:
Proposed solution to button sizing and centering in https://github.com/Automattic/newspack-plugin/pull/46.

- Remove the container div
- Remove special styling for button when it appears in `Card` component
- Add a few styles (display, width, max-width, text-shadow) to the `.is-primary` block.


assets/src/components/button/index.js
```diff
@@ -17,7 +22,8 @@ class Button extends Component {
 	 */
 	render( props ) {
 		const { value } = this.props;
-		return <BaseComponent className="muriel-button" { ...this.props } />;
+		const { classNames } = this.props;
+		return <BaseComponent { ...this.props } className={ murielClassNames( 'muriel-button', this.props.className ) }  />;
```
philipjohn: Should `this.props.className` here actually be `this.props.classNames` (just the one letter difference)?

claudiulodro:
Is this still needed now that #46 is merged?

jeffersonrabb:
> Is this still needed now that #46 is merged?

Nope. Closing!

Issue #45, opened 2019-05-08
Component: Checklist Task


claudiulodro:
One checklist task from #31. See #31 for designs/info.

jeffersonrabb:
Resolved as of https://github.com/Automattic/newspack-plugin/pull/58

PR #58, opened 2019-05-14
Checklist Component

### LINKED ISSUE #31: Components: Checklist
A checklist consists of a number of checklist items for the user to work through. Each checklist item can have one of three states: 
1. Next task ––active (has an action button
1.1 Might have a secondary action (e.g. skip stop option)
2. Upcoming task ––inactive (has no action button)
3. Completed task ––checked (has a secondary action to adjust configuration

checklist-full

Example of 3. Completed task and 1.1. List item with secondary action 
example of 3 (completed task) and 1 1 (secondary action)

jeffersonrabb:
### All Submissions:

Introduces a checklist component, based somewhat on https://github.com/Automattic/wp-calypso/tree/master/client/components/checklist. The component displays a list of tasks, each with a `title`, `description`, `buttonText`, and `completedTitle` which is displayed when the task has been committed. Each `Task` has `onClick` and `onSkip` events which fire when the primary or "Skip" buttons are clicked. The `Checklist` takes a `currentTask` integer as a prop which determines how far along in the list the user has gotten. In addition to visual differences between upcoming/active/complete `Task` there is a `ProgressBar` element that visualizes the user's progress. 

This is the mockup the Component was built to match:

Screen Shot 2019-05-14 at 7 44 00 PM

Some questions for discussion:

1) In the example implementation, the `onClick` and `onSkip` events both will advance the `currentTask`. This generally works well unless the user clicks `Edit` to make a previously completed `Task` current. At this point things get a little weird because completing that task again takes you to the next one, which may be a step backwards overall. @claudiulodro I'm curious to hear if you think this is the right model for handling Click/Skip/Edit buttons?

2) The Primary buttons have less horizontal padding than the mockups show. Rather than tinker with their styling in this PR, curious if we should make changes to the `Button` component to more closely match the mocks in this context. cc: @philipjohn 

3) The mockups also introduce a light text-only button (Edit/Skip). For this I used the `isLink` style button, and added a new style definition to the `Button` component. @philipjohn curious what you think of this, if `isLink` should be held for a different use case, and if so how to handle this new button style.

4) The mockups show an "I'm done setting up" button at the bottom. Is this meant to be part of this component, or is it a free-floating button?

Closes #31, #45 

2. Navigate to `/admin.php?page=newspack-subscriptions-wizard`
3. Observe and interact with the component, trying Primary, Edit and Skip buttons. After some tasks have been completed there should be Hide/Show Completed UI in the upper right corner. Verify that this is working as expected.

### Other information:

### DIFF
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -10,6 +10,7 @@ import { Component, Children, cloneElement } from '@wordpress/element';
 /**
  * Internal dependencies.
  */
+import murielClassnames from '../../shared/js/muriel-classnames';
 import ProgressBar from '../progressBar';
 import './style.scss';
 
@@ -18,9 +19,9 @@ class Checklist extends Component {
 	 * Render.
 	 */
 	render() {
-		const { children, currentTask, progressBarText } = this.props;
+		const { className, children, currentTask, progressBarText, ...otherProps } = this.props;
 		return (
-			<div className="muriel-checklist">
+			<div className={ murielClassnames( 'muriel-checklist', className ) } { ...otherProps }>
 				<div className="muriel-checklist__progress-bar">
 					<ProgressBar completed={ currentTask } total={ Children.count( children ) } displayFraction label={ progressBarText } />
 				</div>
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -1,0 +1,37 @@
+/**
+ * Checklist for tracking multi-step tasks.
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { Component, Children, cloneElement } from '@wordpress/element';
+
+/**
+ * Internal dependencies.
+ */
+import './style.scss';
+
+class Checklist extends Component {
+	/**
+	 * Render.
+	 */
+	render() {
+		const { children, currentTask } = this.props;
+		return (
+			<div className="muriel-checklist">
+				<div className="muriel-checklist__progress-bar">
+					OK
+				</div>
+				<div className="muriel-checklist__tasks">
+					{ Children.map( children, ( child, index ) => {
+						const newChild = cloneElement( child, { isCurrent: index === currentTask, isComplete: index < currentTask } );
+						return newChild;
+					} ) }
+				</div>
+			</div>
+		);
+	}
+}
+
+export default Checklist;
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -60,8 +60,8 @@ class Checklist extends Component {
 						return cloneElement(
 							child,
 							{
-								isCurrent: index === currentTask,
-								isComplete: index < currentTask
+								current: index === currentTask,
+								complete: index < currentTask
 							} );
 					} ) }
 				</div>
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -48,7 +48,7 @@ class Checklist extends Component {
 					</div>
 					{ shouldShowHideCompletionUI && (
 						<div className="muriel-checklist__header-secondary">
-							<label for="muriel-checklist__header-action">{ completedLabel }</label>
+							<label htmlFor="muriel-checklist__header-action">{ completedLabel }</label>
 							<Button id="muriel-checklist__header-action" onClick={ () => this.setState( { hideCompleted: ! hideCompleted } ) }>
 								<Dashicon icon={ completedIcon } />
 							</Button>
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -29,7 +29,7 @@ class Checklist extends Component {
 	 * Render.
 	 */
 	render() {
-		const { className, children, currentTask, progressBarText, ...otherProps } = this.props;
+		const { className, children, progressBarText, ...otherProps } = this.props;
 		const { hideCompleted } = this.state;
 		const completedLabel = hideCompleted ? __( 'Show completed' ) : __( 'Hide completed' );
 		const completedIcon = hideCompleted ? 'arrow-down-alt2' : 'arrow-up-alt2';
@@ -38,38 +38,31 @@ class Checklist extends Component {
 			className,
 			hideCompleted && 'is-hide-completed'
 		);
-		const shouldShowHideCompletionUI = currentTask > 0;
+		const completedCount = Children.toArray( children ).reduce( ( completedCount, child ) => {
+			return completedCount + ( child.props.completed ? 1 : 0 );
+		}, 0 );
 		return (
 			<div className={ classes } { ...otherProps }>
 				<div className="muriel-checklist__header">
 					<div className="muriel-checklist__header-main">
 						<ProgressBar
-							completed={ currentTask }
+							completed={ completedCount }
 							total={ Children.count( children ) }
 							displayFraction
 							label={ progressBarText }
 						/>
 					</div>
-					{ shouldShowHideCompletionUI && (
-						<div className="muriel-checklist__header-secondary">
-							<label htmlFor="muriel-checklist__header-action">{ completedLabel }</label>
-							<Button
-								id="muriel-checklist__header-action"
-								onClick={ () => this.setState( { hideCompleted: ! hideCompleted } ) }
-							>
-								<Dashicon icon={ completedIcon } />
-							</Button>
-						</div>
-					) }
+					<div className="muriel-checklist__header-secondary">
+						<label htmlFor="muriel-checklist__header-action">{ completedLabel }</label>
+						<Button
+							id="muriel-checklist__header-action"
+							onClick={ () => this.setState( { hideCompleted: ! hideCompleted } ) }
+						>
+							<Dashicon icon={ completedIcon } />
+						</Button>
+					</div>
 				</div>
-				<div className="muriel-checklist__tasks">
-					{ Children.map( children, ( child, index ) => {
-						return cloneElement( child, {
-							current: index === currentTask,
-							complete: index < currentTask,
-						} );
-					} ) }
-				</div>
+				<div className="muriel-checklist__tasks">{ children }</div>
 			</div>
 		);
 	}
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -10,6 +10,7 @@ import { Component, Children, cloneElement } from '@wordpress/element';
 /**
  * Internal dependencies.
  */
+import ProgressBar from '../progressBar';
 import './style.scss';
 
 class Checklist extends Component {
@@ -17,11 +18,11 @@ class Checklist extends Component {
 	 * Render.
 	 */
 	render() {
-		const { children, currentTask } = this.props;
+		const { children, currentTask, progressBarText } = this.props;
 		return (
 			<div className="muriel-checklist">
 				<div className="muriel-checklist__progress-bar">
-					OK
+					<ProgressBar completed={ currentTask } total={ Children.count( children ) } displayFraction label={ progressBarText } />
 				</div>
 				<div className="muriel-checklist__tasks">
 					{ Children.map( children, ( child, index ) => {
--- a/assets/src/components/checklist/index.js
+++ b/assets/src/components/checklist/index.js
@@ -5,30 +5,64 @@
 /**
  * WordPress dependencies.
  */
+import { __ } from '@wordpress/i18n';
 import { Component, Children, cloneElement } from '@wordpress/element';
+import { Dashicon } from '@wordpress/components';
 
 /**
  * Internal dependencies.
  */
 import murielClassnames from '../../shared/js/muriel-classnames';
+import Button from '../button';
 import ProgressBar from '../progressBar';
 import './style.scss';
 
 class Checklist extends Component {
+
+	constructor() {
+		super( ...arguments );
+		this.state = {
+			hideCompleted: false,
+		}
+	}
+
 	/**
 	 * Render.
 	 */
 	render() {
 		const { className, children, currentTask, progressBarText, ...otherProps } = this.props;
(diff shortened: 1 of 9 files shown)


iuravic:
Wow, this looks great! Very clear and nicely crafted -- Muriel is awesome. And the component is very slick and interactive! Playing around with it made me notice a few details; please decide for yourself (or let's ask someone else, too) whether these suggestions should be applied, or it's already fine just the way it is.

assets/src/components/checklist/index.js
```diff
@@ -0,0 +1,73 @@
...
+		const shouldShowHideCompletionUI = currentTask > 0;
+		return (
+			<div className={ classes } { ...otherProps }>
+				<div className="muriel-checklist__header">
+					<div className="muriel-checklist__header-main">
+						<ProgressBar completed={ currentTask } total={ Children.count( children ) } displayFraction label={ progressBarText } />
+					</div>
+					{ shouldShowHideCompletionUI && (
+						<div className="muriel-checklist__header-secondary">
+							<label for="muriel-checklist__header-action">{ completedLabel }</label>
```
iuravic: There's a small warning when the "Do it" button is clicked

image

And if we just update the `from` to `htmlFrom` in this line here, it goes away:
```suggestion
							<label htmlFor="muriel-checklist__header-action">{ completedLabel }</label>							
```

claudiulodro:
> In the example implementation, the onClick and onSkip events both will advance the currentTask. This generally works well unless the user clicks Edit to make a previously completed Task current. At this point things get a little weird because completing that task again takes you to the next one, which may be a step backwards overall. @claudiulodro I'm curious to hear if you think this is the right model for handling Click/Skip/Edit buttons?

I think making the task un-completed when clicking the 'Edit' button is weird behavior. :) I expect clicking the 'Edit' link will take me to the wizard (same behavior as 'Do it') but will not modify the completion of any tasks. It's just a handy way to re-run a completed wizard. I don't think we need to modify the implementation of the component itself to behave this way; it's just the way it is in the component demo.

assets/src/components/task/style.scss
```diff
@@ -0,0 +1,88 @@
...
+		background: $muriel-white;
+		border: 2px solid $muriel-gray-200;
+    border-radius: 16px;
+		display: block;
+		height: 16px;
+		left: 24px;
+    position: absolute;
+    top: 24px;
+    width: 16px;
+	}
```
claudiulodro: I think you've mixed spaces and tabs; that's why the formatting is messed up in this file.

iuravic:
The Wizard now does a nice demo of the basic changes in the Tasks' statuses. But playing around with it a bit made me think if we might need to add the ability to also mark individual Tasks as already complete (for example, in case we're reloading/displaying a List that's already been partially complete).

Something like this, 3rd step being already complete:

image

If you think this is a functionality we might also wish to add now, I tinkered with it a little bit and here's how it might be extended. Please consider all this just as a suggestion and a sketch, which you might or might not want to use at all. I'm not even fully sure it's the best way to do.

Next I'll immediately give and publish the "code suggestions" that could do this. I hope they'll display nicely following this comment!

assets/src/components/checklist/index.js
```diff
@@ -0,0 +1,73 @@
...
+						</div>
+					) }
+				</div>
+				<div className="muriel-checklist__tasks">
+					{ Children.map( children, ( child, index ) => {
+						return cloneElement(
+							child,
+							{
+								current: index === currentTask,
+								complete: index < currentTask
```
iuravic: -  first enabled the `isComplete` prop on every child:

```suggestion
								complete: child.props.complete || ( index < currentTask )
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -30,6 +33,7 @@ class SubscriptionsWizard extends Component {
 	constructor() {
 		super( ...arguments );
 		this.state = {
+			currentTask: 0,
 			inputTextValue1: "Input value",
 			inputTextValue2: "",
 			image: null,
```
iuravic: - here in the Wizard, added a `completedTasks` Array to state, to keep track of the completed ones (it holds indexes of completed tasks -- starting from 0). In this example the **3rd** task is marked as complete, with index **2**

```suggestion
			image: null,
			completedTasks: [ 2 ],
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -40,31 +44,65 @@ class SubscriptionsWizard extends Component {
 	 * Render the example stub.
```
iuravic: - right when the Wizard is loaded, updating the currentTask (since, for example, tasks 0 and 1 might already be done):

```suggestion
	 * componentDidMount.
	 */
	componentDidMount() {
		this.setState( { currentTask: this.getNextTask( 0, this.state.completedTasks) } )
	}
```

- just to keep things clean, extracted this into a separate function:

```suggestion

	/**
	 * Starting from the startingIndex, determine the next task to go to.
	 *
	 * @param int startingIndex
	 * @param Array completedTasks
	 * @returns int {number}
	 */
	getNextTask = ( startingIndex, completedTasks ) => {
		let nextTaskIndex = 0;
		while ( completedTasks.indexOf( nextTaskIndex ) !== -1 ) {
			nextTaskIndex++;
		}

		return nextTaskIndex;
	}
```

- here added a handler when the task is complete (besides updating the `currentTask`, now also adding it into the `completedTasks` Array)

```suggestion

	/**
	 * Upon completing the task, add its index to this.state.completedTasks, and increase this.state.currentTask.
	 *
	 * @param int taskIndex
	 */
	completeTask = ( taskIndex ) => {
		let completedTasks = [ ...this.state.completedTasks ];
		if ( completedTasks.indexOf( taskIndex ) === -1 ) {
			completedTasks.push( taskIndex );
			completedTasks.sort();

			this.setState( { completedTasks } );
		}

		this.setState( { currentTask: this.getNextTask( 0, completedTasks ) } );
	}

	/**
	 * Render the example stub.
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -40,31 +44,65 @@ class SubscriptionsWizard extends Component {
...
 	 */
 	render() {
-		const { inputTextValue1, inputTextValue2 } = this.state;
+		const { inputTextValue1, inputTextValue2, currentTask } = this.state;
```
iuravic: - here now also using the `completedTasks`:

```suggestion
		const { inputTextValue1, inputTextValue2, currentTask, completedTasks } = this.state;
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -40,31 +44,65 @@ class SubscriptionsWizard extends Component {
...
+					headerText={ __( 'Newspack Components' ) }
+					subHeaderText={ __( 'Temporary demo of Newspack components' ) }
 				/>
+				<Checklist currentTask={ currentTask } progressBarText={ __( 'Your setup list' ) }>
+					<Task
+						title={ __( 'Set up membership' ) }
+						description={ __( 'Optimize your site for search engines and social media by taking advantage of our SEO tools. We\'ll walk you through important SEO strategies to get more exposure for your business.' ) }
+						buttonText={ __( 'Do it' ) }
+						completedTitle={ __( 'All set!' ) }
+						onClick={ () => this.setState( { currentTask: 1 } ) }
```
iuravic: - finally, modifying `onClick`, `onSkip`, `complete` and `current` **for all four of the `<Task/>` components**, like this:
```suggestion
						onClick={ () => this.completeTask( 0 ) }
						onSkip={ () => this.completeTask( 0 ) }
						complete={ _.includes( completedTasks, 0 ) }
						current={ 0 === currentTask }
```

The only problem now is that the use of the `<Task/>` component has all these repetitions of the keys, which actually looks worse now... Makes me think if the `<Task/>`s could maybe be created with the use of a map() function, maybe. This I'm not sure of.

assets/src/components/checklist/index.js
```diff
@@ -0,0 +1,73 @@
...
+		const { className, children, currentTask, progressBarText, ...otherProps } = this.props;
+		const { hideCompleted } = this.state;
+		const completedLabel = hideCompleted ? __( 'Show completed' ) : __( 'Hide completed' );
+		const completedIcon = hideCompleted ? 'arrow-down-alt2' : 'arrow-up-alt2';
+		const classes = murielClassnames(
+			'muriel-checklist',
+			className,
+			hideCompleted && 'is-hide-completed'
+		);
+		const shouldShowHideCompletionUI = currentTask > 0;
```
iuravic: - also just enabled the `HideCompletedUI`, in case some steps are already completed:

```suggestion
		const numberOfCompleted = this.props.children.filter( child => child.props.complete ).length;
		const shouldShowHideCompletionUI = numberOfCompleted > 0;
```

assets/src/components/checklist/index.js
```diff
@@ -0,0 +1,73 @@
...
+		return (
+			<div className={ classes } { ...otherProps }>
+				<div className="muriel-checklist__header">
+					<div className="muriel-checklist__header-main">
+						<ProgressBar completed={ currentTask } total={ Children.count( children ) } displayFraction label={ progressBarText } />
```
iuravic: -  updated the `ProgressBar` to use `numberOfCompleted` as the numerator:

```suggestion
						<ProgressBar completed={ numberOfCompleted } total={ Children.count( children ) } displayFraction label={ progressBarText } />
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -31,43 +34,83 @@ class SubscriptionsWizard extends Component {
...
 		}
 	}
 
 	/**
-	 * Render the example stub.
+	 * componentDidMount.
+	 */
+	componentDidMount() {
+		this.setState( { currentTask: this.getNextTask( 0, this.state.completedTasks) } )
+	}
```
iuravic: - just to keep things clean, extracted this into a separate function:

```suggestion
	}

	/**
	 * Starting from the startingIndex, determine the next task to go to.
	 *
	 * @param int startingIndex
	 * @param Array completedTasks
	 * @returns int {number}
	 */
	getNextTask = ( startingIndex, completedTasks ) => {
		let nextTaskIndex = 0;
		while ( completedTasks.indexOf( nextTaskIndex ) !== -1 ) {
			nextTaskIndex++;
		}

		return nextTaskIndex;
	}
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -31,43 +34,99 @@ class SubscriptionsWizard extends Component {
...
+	 * @returns int {number}
+	 */
+	getNextTask = ( startingIndex, completedTasks ) => {
+		let nextTaskIndex = 0;
+		while ( completedTasks.indexOf( nextTaskIndex ) !== -1 ) {
+			nextTaskIndex++;
+		}
+
+		return nextTaskIndex;
+	}
```
iuravic: - here added a handler when the task is complete (besides updating the `currentTask`, now also adding it into the `completedTasks` Array):

```suggestion
	}

	/**
	 * Upon completing the task, add its index to this.state.completedTasks, and increase this.state.currentTask.
	 *
	 * @param int taskIndex
	 */
	completeTask = ( taskIndex ) => {
		let completedTasks = [ ...this.state.completedTasks ];
		if ( completedTasks.indexOf( taskIndex ) === -1 ) {
			completedTasks.push( taskIndex );
			completedTasks.sort();

			this.setState( { completedTasks } );
		}

		this.setState( { currentTask: this.getNextTask( 0, completedTasks ) } );
	}
```

assets/src/wizards/subscriptions/index.js
```diff
@@ -31,43 +34,99 @@ class SubscriptionsWizard extends Component {
...
 	 */
 	render() {
-		const { inputTextValue1, inputTextValue2, selectValue1, selectValue2 } = this.state;
+		const { inputTextValue1, inputTextValue2, selectValue1, selectValue2, currentTask } = this.state;
```
iuravic: - now using `completedTasks` in the `render()`:

```suggestion
		const { inputTextValue1, inputTextValue2, selectValue1, selectValue2, currentTask, completedTasks } = this.state;
```

assets/src/wizards/subscriptions/index.js
iuravic: **This conversation/segment is a dupe, please ignore.**

assets/src/wizards/subscriptions/index.js
iuravic: **This comment is a dupe, please ignore.**

assets/src/components/checklist/index.js
jeffersonrabb: Nice catch! Fixed in https://github.com/Automattic/newspack-plugin/pull/58/commits/b05f14cbad553c5fb8cd24f76905a039959878b8

assets/src/components/task/style.scss
jeffersonrabb: There was a typo in the Prettier configuration file and a couple of differences in `.editorconfig` between this repo and https://github.com/Automattic/newspack-blocks. I fixed these in https://github.com/Automattic/newspack-plugin/pull/58/commits/3391656768aff8f011fcebd3fe57ea9cf8e28d80 and https://github.com/Automattic/newspack-plugin/pull/58/commits/c6759ad2a972789f35a3c74ff338f231f4679e57, then applied Prettier in https://github.com/Automattic/newspack-plugin/pull/58/commits/d39eae5021f81f996113093baaa41e091bc00548 and now I think these should be resolved.

iuravic:
Looks really nice now! Both the `completed` and the `active` props work great, and the whole thing is even simpler&clearer now. Thank you for your patience!

jeffersonrabb:
> One non-blocking thing we'll want to think about: After a task is completed, how will the user know what that task was (e.g. so they can know what 'Edit' links to)?

I think this would just be a matter of more descriptive `completedTitle`s. I used "All Set!" for all of them, which is pretty unhelpful!

PR #60, opened 2019-05-14
TextControl's CSS width changed from hard-coded px to percentage-based

iuravic:
### All Submissions:

Updated the TextControl's CSS width from hard-coded pixel value to a percentage-based value.

The TextControl are demoed in the Subscription Wizzard. One can change the value of `.muriel-card`'s `max-width` value by using the Inspector, and check that the component styling changes accordingly and behaves properly.

### Other information:

### DIFF
--- a/assets/src/components/button/button.test.js
+++ b/assets/src/components/button/button.test.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Button from './';
+
+describe( 'Button', () => {
+	describe( 'basic rendering', () => {
+		it( 'should render a Button element with a value of "Continue"', () => {
+			const button = shallow( <Button isPrimary>Continue</Button> );
+			expect( button.render().hasClass( 'muriel-button' ) ).toBe( true );
+		} );
+	} );
+	describe( 'rendering default', () => {
+		it( 'should render a Button element with a class of is-default', () => {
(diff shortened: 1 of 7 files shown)


iuravic:
Aah that's absolutely right! Thank you for both giving it a test run, and for the excellent explanation, makes perfect sense! :) I now removed the `max-width`s, and it seems to scale fine, both for small and large `card`s.

iuravic:
Claudiu, I did another update in my last commit, the ["Better onClick handler"](https://github.com/Automattic/newspack-plugin/pull/60/commits/44ee2bb16f63b5843e442847d18c28a372669684) (44ee2bb1).

Following the same pattern as with `className` setting, where we acknowledge both the `className` set by the parent, and join it with the Component's internal `className`s, so have I now done with the `onClick` handler. The point is -- before this change, if someone were to set the `onClick` handler on the Component in the Wizzard, the component's internal `onClick` would "overwrite" it. I've done the simplest thing here, now both the component's internal handler is executed, just as the one set from the Parent (Wizzard) (if it exists).

I'll apply this same pattern to the SelectComponent -- if you take a look [here](https://github.com/Automattic/newspack-plugin/pull/62/files#diff-56ebb05a01f78c24c8608213a454266cR80), I was actually careful enough to do it for the `onChange`, but not for the `onClick` events/handlers.

PR #62, opened 2019-05-14
Add Muriel-styled Select dropdown component

### LINKED ISSUE #21: Components: Select
Standard select dropdown, but with Muriel UI.

Example usage:
add-subscription

Design:
dropdown
dropdown-2

Existing implementations:
- https://github.com/WordPress/gutenberg/tree/master/packages/components/src/select-control
- https://github.com/woocommerce/woocommerce-admin/tree/master/packages/components/src/dropdown-button
- https://github.com/Automattic/wp-calypso/tree/master/client/components/forms/form-select

iuravic:
### All Submissions:

Closes #21  .

The Select components in different states added to the Subsciptions Wizzard page.

### Other information:

### DIFF
--- a/assets/src/components/selectControl/index.js
+++ b/assets/src/components/selectControl/index.js
@@ -71,7 +71,7 @@ const SelectControl = withFocusOutside(
 		 */
 		render() {
 			const { isFocused } = this.state;
-			const { value, disabled, className, onChange } = this.props;
+			const { value, disabled, className, onChange, ...otherProps } = this.props;
 			const isEmpty = ! value;
 			const isActive = isFocused && ! disabled;
 			const classes = murielClassnames( "muriel-select", this.getClassName( disabled, isEmpty, isActive ), className );
@@ -79,9 +79,9 @@ const SelectControl = withFocusOutside(
 			return (
 				<BaseComponent
 					className={ classes }
-					{ ...this.props }
 					onClick={ () => this.handleOnClick() }
 					onChange={ value => this.handleOnChange( onChange, value ) }
+					{ ...otherProps }
 				/>
 			);
 		}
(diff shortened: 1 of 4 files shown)


assets/src/components/selectControl/index.js
```diff
@@ -0,0 +1,90 @@
...
+		 */
+		render() {
+			const { isFocused } = this.state;
+			const { value, disabled, onChange } = this.props;
+			const isEmpty = ! value;
+			const isActive = isFocused && ! disabled;
+
+			return (
+				<BaseComponent
+					className={ murielClassnames( "muriel-select", this.getClassName( disabled, isEmpty, isActive ) ) }
```
claudiulodro: `className` should be handled using the pattern [here](https://github.com/Automattic/newspack-plugin/pull/61/files#diff-8a670ed255f49ef836328cad94e50247R24). Otherwise, setting a `className`  overrides all of the classnames.

e.g. try this:
```
<SelectControl
	className="testing"
	label="Select with value"
	value={ selectValue1 }
	options={ [
		{ value: '1st', label: 'First' },
		{ value: '2nd', label: 'Second' },
		{ value: '3rd', label: 'Third' },
	] }
	value={ selectValue1 }
	onChange={ value => this.setState( { selectValue1: value } ) }
/>
```

You can see that our classes get removed and the only class is 'testing':

Screen Shot 2019-05-14 at 12 00 35 PM

assets/src/components/selectControl/style.scss
```diff
@@ -0,0 +1,130 @@
+/**
+ * Select dropdown component styles.
+ */
+
+.muriel-select {
+	position: relative;
+	width: 100%;
+	max-width: 440px;
```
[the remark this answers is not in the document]
iuravic: Thank you for catching that, again! Applying the same solution here now.

assets/src/components/selectControl/index.js
iuravic: Great thinking! Thank you! Updated now.

I also love how the spread operator is used to handle the `...otherProps` in that example, so I added this change as well.

Issue #55, opened 2019-05-13
Component: Required plugins


claudiulodro:
We'll need a component to use in our wizards that takes a list of required plugins, and tries to install/activate all of the required plugins. It should probably display some sort of spinner while this is going on with a generic message (e.g. "Preparing required plugins"). If the plugins are unable to be installed/activated (e.g. because an uninstalled premium plugin is required), it should display and/or return some sort of useful error message. Some new handling will need to be implemented in the Plugin Manager class to account for premium plugins that cannot be installed, and likely some new functionality will be needed to process a list of required plugins (unless that is done all in JS using our existing plugin management endpoints).

#### Example use case:
1. When the Subscriptions wizard is first loaded, this component checks to make sure WooCommerce, WooCommerce Subscriptions, etc. are installed/active.
2. If the required plugins are not installed/active, it should install and activate them where possible.
3. If WooCommerce Subscriptions is not installed, it cannot be automatically installed since it's a premium plugin. The Wizard should display a useful error message and prevent the user from going forward. If plugin requirements are met and all plugins are installed/activated, the wizard should proceed as normal.

#### Design

TBD. @sonjaleix, thoughts?

sonjaleix:
Here is a first draft of various states of the required plugin "cards" within the wizard. We might not necessarily need the yellow warning label for outdated plugins, if we're trying to utilize auto-updating of plugins.
<img width="519" alt="Screenshot 2019-05-24 11 07 31" src="https://user-images.githubusercontent.com/4823744/58345432-945c4a00-7e15-11e9-9eca-a8b5c0c1a293.png">

claudiulodro:
#69

PR #64, opened 2019-05-17
Subscriptions Wizard

### LINKED ISSUE #53: Wizard: Subscriptions
A wizard for creating and managing subscriptions. This consists of a frontend UI for the screens, and backend API endpoints/handling. For this initial implementation, we can assume that the appropriate plugins are already installed and configured. This wizard will need to use the Subscriptions wizard classes we are currently using for the component demos, so those will need to be moved somewhere else.

Here are a couple screens:
3  Marketing Tools - Split
4  Marketing Tools - Split

Further designs are available in Figma. 

We'll need to investigate a good way of handling the global 'Allow members to name their price' checkbox, as that's normally a meta field on a product. Likely a filter exists we can hook into.

claudiulodro:
### All Submissions:

Closes #53 .
Supersedes #15. 

This is a rather large PR, however it lays the groundwork for all future wizards, so it should be straightforward after this.

Changes:
- Moved the components demos into a new Components Demo wizard (this is mostly why the changed line count is so large)
- Implement the Subscriptions Wizard
- Some small bug fixed for issues blocking development of this feature (I've pointed these out in comments below)

One wizard has a simple pattern (in my opinion):
- On the backend, each wizard maintains its own endpoints in the associated PHP Wizard class.
- On the frontend, each screen is self-contained and manages its own data. The main wizard class exposes a changeScreen method to the screens to enable changing to different screens when buttons are clicked and stuff.

Not implemented in this initial version:
- Loading UI (spinner, etc.) while waiting for API requests. We should discuss how we want loading to behave in general. There is a Muriel design for a loading spinner.
- Plugin requirements handling. See #55. For now **make sure you have WooCommerce, WooCommerce Subscriptions, and WooCommerce Name Your Price active.**

1. `npm ci; npm run clean; npm run build:webpack`.
2. Go to Newspack > Components Demo and verify that's still working.
3. Go to Newspack > Subscriptions.
4. Add and save subscriptions.
5. Test editing a subscription.
6. Test the "Allow members to specify donation amount" checkbox.
7. Click on the small, round image to see a subscription on the frontend.
8. Test deleting a subscription.

Screen Shot 2019-05-17 at 10 09 49 AM
Screen Shot 2019-05-17 at 10 10 24 AM

### Other information:


assets/src/components/checkboxControl/style.scss
```diff
@@ -30,4 +30,11 @@
 		margin-left: 2em;
 		font-size: 16px;
 	}
+
+	.components-base-control__help {
+		margin-left: 58px;
+		font-style: normal;
+		margin-top: 0;
+		color: $muriel-gray-600;
+	}
```
claudiulodro: The help prop was unstyled for our checkboxes, causing it to go under the checkbox which looked ugly. I've added styling for it to make it look like our designs.

assets/src/components/selectControl/index.js
```diff
@@ -83,6 +83,7 @@ const SelectControl = withFocusOutside(
 
 			return (
 				<BaseComponent
+					value={ value }
```
claudiulodro: Since `value` was extracted from props above, it wasn't getting passed to BaseComponent, meaning you couldn't properly set a value on selects. I've not investigated whether the same issue happens for `disabled`, but it likely has the same problem. cc @iuravic

assets/src/components/textControl/style.scss
```diff
@@ -30,7 +30,7 @@
 		color: $muriel-gray-500;
 	}
 
-	input[type="text"] {
+	input[type="text"], input[type="number"] {
```
claudiulodro: `number` inputs were not getting our styles.

includes/wizards/class-subscriptions-wizard.php
```diff
@@ -37,6 +39,327 @@ class Subscriptions_Wizard extends Wizard {
...
 	protected $capability = 'edit_products';
 
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		add_action( 'rest_api_init', [ $this, 'register_api_endpoints' ] );
+		add_filter( 'woocommerce_product_data_store_cpt_get_products_query', [ $this, 'handle_only_get_newspack_subscriptions_query' ], 10, 2 );
```
claudiulodro: The Subscriptions Wizard only displays and manages products created through the Subscriptions Wizard. I think that's the way to go because it would be super confusing/breaking if we pulled in swag or other such products. This should keep things working nicely.

phpcs.xml
```diff
@@ -30,7 +30,7 @@
 	<rule ref="PHPCompatibilityWP"/>
 	<config name="testVersion" value="5.6-"/>
 
-	<arg name="extensions" value="php,js"/>
```
claudiulodro: Tweak: I've removed the js check from phpcs.xml because phpcs freaks out on React code and thinks that e.g. `<div>` is using greater/less-than operators, so it generates hundreds of style errors. Additionally, now that we have prettier working, we probably don't need this. We can re-approach after this is merged if we want to add it back.

claudiulodro:
@philipjohn can you paste your system status report (WooCommerce > Status) so I can reproduce your environment?

I'll split out the PR into multiple ones tomorrow also; that should make it easier to review.

jeffersonrabb:
> I'll split out the PR into multiple ones tomorrow

Since this one might be broken into separate PRs I'll hold off on line-by-line review for now, but I'll note some broad architecture questions here. As context, I'm trying to take a long view and think about establishing SPA patterns that we can use for 100% of our wizards, so if any of these points seem like pedantic over-engineering, that's kind of where my head's at.

1) It seems to me we have a controller (`index.js`) and two views—a **list** (`manageSubscriptionsScreen.js`) and a **detail** (`editSubscriptionScreen.js`). If this is an accurate description of the structure, we probably want all state management and API interaction (essentially anything related to model) in the controller, and the views should be  concerned with display concerns only. To achieve this, we would lift state from the controllers to the controller as described here https://reactjs.org/docs/lifting-state-up.html.  All of the API interactions would be moved to `index.js` and the list/detail views would be stripped down to basically just a `render()` function and a few event handlers. The views would receive data as props and pass changes up to the controller via `onChange` events.

2)  I'm also wondering if simple state will be sufficient for out needs, or if we might need to introduce Redux (https://redux.js.org/) as a state container? Per [this](https://reactjs.org/docs/faq-state.html#should-i-use-a-state-management-library-like-redux-or-mobx), maybe not?

3) I don't think Components should be stored in state as you have [here](https://github.com/Automattic/newspack-plugin/pull/64/files#diff-9e828026d7448da7b3ee49ae591cc14bR27). Explanation: https://shripadk.github.io/react/docs/interactivity-and-dynamic-uis.html#what-shouldnt-go-in-state. 

4) Thinking through ☝️ I was trying to figure out the best way to manage view hierarchies, and I wonder if we should be considering implementing a Router, e.g. https://github.com/ReactTraining/react-router?

Overall I found this doc useful: https://reactjs.org/docs/thinking-in-react.html

philipjohn:
Will send you the status report via DM...

Just to check I actually have this right, I'm using these two plugins:

* https://wordpress.org/plugins/yith-woocommerce-subscription/
* https://wordpress.org/plugins/yith-woocommerce-name-your-price/

Also, I'm deliberately playing dumb and doing _nothing_ to them and only interacting with the Newspack wizard.

philipjohn:
> I have installed and activated, but not configured in any way, WooCommerce, WooCommerce Subscriptions and WooCommerce Name Your Price. Are we sure there isn't some extra setup set before that class is available?

Just noting this was because I had the _wrong_ plugins and that it's all working lovely for me with the correct plugins.

claudiulodro:
See #73

PR #69, opened 2019-05-21
Plugin Installer Component

jeffersonrabb:
### All Submissions:

This PR adds a new component that provides a UI for installing plugins. The component takes one parameter, `plugins`, which is an array of the plugin slugs to display and manage. Assuming those plugins have been defined/whitelisted in https://github.com/Automattic/newspack-plugin/blob/8cced0e613fafaa6aef086ff13f4c6acb11d8f22/includes/class-plugin-manager.php, the component displays information about each one in a separate `ActionCard`. The UI allows for installation/activation (these are treated as a single action with no distinction), deinstallation, and one-click installation of all listed plugins. If installation fails, an error notification will be posted. I added a fake plugin that is guaranteed to fail as a demonstration of the notifications. 

1. Check out the branch.
2. `npm ci && npm run clean && npm run build:webpack`
3. Navigate to Newspack->Components Demo, observe Plugin Installer at the top of the page with four plugins listed (AMP, WooCommerce, Yoast SEO, "Fake Plugin")
4. Click `Install` button for individual plugins and verify that they are successfully installed.
5. When clicking `Install` for Fake Plugin, observe that installation fails and an error notification is posted with the text "Plugin in not found."
6. Hover over cards and click `Uninstall`, verify that plugins are deactivated successfully. (Note they will not be deleted, just deactivated)
7) Deactivate all plugins, then click `Install All`. Verify that all plugins are installed and activated, except for Fake Plugin which will fail.

Screen Shot 2019-05-30 at 6 35 47 PM

### Other information:


claudiulodro:
Regarding the checkboxes that select the plugins to install, is there any situation where you would only check some of the boxes? Seems like all the boxes would always be checked, right? The feature gated by the plugin installer won't work with only half of the requirements installed! :P

jeffersonrabb:
> Regarding the checkboxes that select the plugins to install, is there any situation where you would only check some of the boxes? Seems like all the boxes would always be checked, right? The feature gated by the plugin installer won't work with only half of the requirements installed! :P

Yeah, this is a good question. In the specific case of subscriptions, perhaps Name Your Price would be optional? But more generally I wonder about future SPAs where there might actually be choices to make between different plugins, which is why I left the door open for pick-and-choose installation. I could easily be talked out of it.

claudiulodro:
No that makes a lot of sense. Good thinking. 

We can just force a checkbox always-checked for a hard plugin requirement, so we have more control over things with your approach

philipjohn:
We're kinda making the user take a decision that they don't _need_ to take here, I think. We should be able to decide for them what plugins should be installed and active based on what features they want.

I.e. when they set up a subscription system, we ask all the relevant question about the format etc and then the wizard can simply say "Installing required plugins... Activating required plugins... Configuring plugins..." or something even less WordPress-y.

If we know what they want to do, I don't think we need them to make any plugin decisions - that's part of the power of Newspack.

iuravic:
I to tested, and all the actions do and work as expected! 👍

jeffersonrabb:
@claudiulodro per @sonjaleix's comment about `"There is a new version available."` messaging, do we have any way with the current API to verify if the plugin is up to date?

claudiulodro:
Plugin updated status is not currently in our API. It should be somewhat straightforward to get that info from `get_site_transient( 'update_plugins' )` and add it into `Plugin_Manager::get_managed_plugins` if we want that info.

jeffersonrabb:
> Is there any situation where we want people deactivating/uninstalling plugins in a wizard? When we use this component in the wizards it's all "You're missing this stuff; please install it", right? In those situations we may not want to let people uninstall/deactivate through this. Maybe have that ability enabled by a prop?

None I can think of. The standard wp-admin plugin screen is the better way if deactivation is required. TBH I mostly put the uninstall option in to make this component easier to test. I like the idea of making deletion a prop with default of no, just in case a use case becomes clear later. I'll add this.

jeffersonrabb:
> How do you envision programmatically using this component in a Wizard? e.g. A user gets to the Subscriptions wizard, needs to install some plugins, and either installs the required plugins or doesn't. Do we need an onAllPluginsInstalled prop callback or something to let us know when to block/enable progressing through the wizard?

We definitely need an a callback like that (I'll probably call it `onComplete`) which can be used to proceed—I'll add. A tougher question is how to handle it the second time the user lands on the wizard. One approach would be that the wizard loads the plugin loader component every time, and if the component determines that all plugins are installed it immediately fires `onComplete` and the wizard proceeds. It may be more efficient, however, for the wizard itself to determine if plugin installation is needed and skip rendering this component entirely when this is is the case. Which approach would you prefer @claudiulodro ?

jeffersonrabb:
> For plugins that are installed but not active, is 'Install' a good action phrase? Should it be 'Activate'? (These plugins will have the inactive status in the API response when getting their info) Similarly, is 'Uninstall', the best action phrase when we're actually deactivating? Just curious; if it's the current way for simplicity, I'm OK with that.

Curious to get @sonjaleix's $0.02 on this. I would prefer to use one verb only, but since Install and Activate have precise meanings related to plugins and they may actually be wrong in certain cases, maybe we're better off finding a new verb? Maybe just "Use?"

claudiulodro:
> We definitely need an a callback like that (I'll probably call it onComplete) which can be used to proceed—I'll add. A tougher question is how to handle it the second time the user lands on the wizard. One approach would be that the wizard loads the plugin loader component every time, and if the component determines that all plugins are installed it immediately fires onComplete and the wizard proceeds. It may be more efficient, however, for the wizard itself to determine if plugin installation is needed and skip rendering this component entirely when this is is the case. Which approach would you prefer @claudiulodro ?

I like the `immediately fires onComplete` approach, that way we can keep all the plugin logic in this component. 👍

sonjaleix:
> > Is there any situation where we want people deactivating/uninstalling plugins in a wizard? When we use this component in the wizards it's all "You're missing this stuff; please install it", right? In those situations we may not want to let people uninstall/deactivate through this. Maybe have that ability enabled by a prop?
> 
> None I can think of. The standard wp-admin plugin screen is the better way if deactivation is required. TBH I mostly put the uninstall option in to make this component easier to test. I like the idea of making deletion a prop with default of no, just in case a use case becomes clear later. I'll add this.

Agreed, that this feature is great for testing, but probably not needed for real-life usage. Unless you guys can think of any conflicts / error messages where a "deletion / deactivation" is a helpful or necessary path forward. 


> > For plugins that are installed but not active, is 'Install' a good action phrase? Should it be 'Activate'? (These plugins will have the inactive status in the API response when getting their info) Similarly, is 'Uninstall', the best action phrase when we're actually deactivating? Just curious; if it's the current way for simplicity, I'm OK with that.

> Curious to get @sonjaleix's $0.02 on this. I would prefer to use one verb only, but since Install and Activate have precise meanings related to plugins and they may actually be wrong in certain cases, maybe we're better off finding a new verb? Maybe just "Use?"

That's a great point and I agree that a "neutral" non-technical term would be better. Let's go with "Use" for now and see how that flows. And let's use "Deactivate" instead of Delete. Since that's the actual correct term for what it does and there are no two actions the action could have.

jeffersonrabb:
> I like the immediately fires onComplete approach, that way we can keep all the plugin logic in this component. 👍

`onComplete` added in https://github.com/Automattic/newspack-plugin/pull/69/commits/c144e12303b78fe2d030daebbdedae805f2e8638

jeffersonrabb:
> Agreed, that this feature is great for testing, but probably not needed for real-life usage. Unless you guys can think of any conflicts / error messages where a "deletion / deactivation" is a helpful or necessary path forward.

`canUninstall` option added in https://github.com/Automattic/newspack-plugin/pull/69/commits/b08ab68c01219f3d8d232abb184d4cfeb1324805. There are now two instances of the component on  the demo page. The upper one has deactivation buttons, the lower does not.

jeffersonrabb:
> That's a great point and I agree that a "neutral" non-technical term would be better. Let's go with "Use" for now and see how that flows. And let's use "Deactivate" instead of Delete. Since that's the actual correct term for what it does and there are no two actions the action could have.

Updated all of buttons and state terms in https://github.com/Automattic/newspack-plugin/pull/69/commits/bb779be6f1bc3a595c185056ea5344a03331b34a:

Install/activate button: "Use"
Deactivate button: "Deactivate"
Install/activate all: "Use All"
Installation in progress: "Setting up..."
Deactivation in progress: "Deactivating"
Plugin is installed: "In Use"

jeffersonrabb:
> Plugin updated status is not currently in our API. It should be somewhat straightforward to get that info from get_site_transient( 'update_plugins' ) and add it into Plugin_Manager::get_managed_plugins if we want that info.

I think we should leave plugin version check and notification for a different PR, since this will involve changes to the API, not the component.

sonjaleix:
Looks good to me too! 
It seems the individual components for this still have some of the old verbiage. Not sure if that needs to be updated, but wanted to point it out (e.g. "Installing...", "Install", etc.).

jeffersonrabb:
> It seems the individual components for this still have some of the old verbiage. Not sure if that needs to be updated, but wanted to point it out (e.g. "Installing...", "Install", etc.).

I'm confused about this. @sonjaleix where are you seeing the old text?

sonjaleix:
> I'm confused about this. @sonjaleix where are you seeing the old text?
@jeffersonrabb I've updated my local instance with master (since this components has been merged) and I see the updated copy on the wizard, but below the wizard on the "Components Demo" screen it shows each of the component instances and there the copy is not updated yet.

PR #88, opened 2019-06-05
DRAFT Ad Management Wizard

### LINKED ISSUE #82: Wizard: Ads
Many publishers have a need to serve advertisements on their site. They may do this through an existing service like Google AdSense or Ad Manager.

Publishers should find it easy to connect ad services they use from within the WordPress dashboard. They should be able to step through a series of steps where they can connect existing accounts, or set up new ones, and begin implementing ads on their site right away.

There are many ads plugins around but a review (pamTN9-fj-p2) found that most were either too cumbersome, or lacked necessary features without needing a paid upgrade.

We are providing configuration and setup through wizards so it will be a more natural UX for there to be an additional wizard for ads.

philipjohn:
### All Submissions:

Provides a simple wizard for setting up and managing placement of ads. Users can select their chosen ad network from an initial set of three. Two of the networks then allow for choosing pre-defined placements, borrowing from the Jetpack Ads set up. The third (AdSense) uses SiteKit.

This initial draft includes the UI elements that can be stepped through as the user would do, but with no backend functionality to make the changes. This PR is to invite comment and review on the current composition of the wizard and the user experience.

Closes #82

1. Apply the patch
2. Navigate to Newspack > Advertising
3. Select one of the ad networks and observe the set up flow.

### Other information:


philipjohn:
Oh! I did rebase before creating this, but didn't spot the moved wizards dir. All in the right place now.

> Does Site Kit let people configure their ad placements in a similar way to the WordAds placement settings? Just curious how close the feature parity is.

At the moment, I've been unable to get that far with SiteKit but will be checking back in on the site I'm using to test to see if it's properly connected now.

sonjaleix:
@philipjohn  and I met to flesh out the ads wizard. He demoed the first iteration of the wizard.
Here are my notes: 

We'll likely need to adjust the UI pattern for the ad provider selector. I'll work on a better solution after I've gained some insights into whether newsrooms would use just one ad provider or if there might be cases where multiple providers are used.

Will we need to add a "required plugins" wizard within the ad setup flow?
For WordAds, no. We assume Jetpack is enabled.
For AdSense, required plugin setup of Sitekit. The user gets taken to the Sitekit configuration wizard within WP Admin.
For AdManager we don't yet know how this will work in the backend. For phase 1 the user might have to copy/paste the HTML code from AdManager into our wizard. Ideally this could also be configured through Sitekit in the WP Admin.

sonjaleix:
@philipjohn I ran into an odd issue with the ad wizard not loading resulting in a blank screen. This was due to having an ad blocker enabled in my browser. I discovered the error through the JS console. Ideally we'd try prevent this issue – more opaque naming might help as per @jeffersonrabb suggestion.

jeffersonrabb:
> more opaque naming might help

I installed AdBlock (Chrome extensions) and tried renaming the wizard. Here's what did and did not work:

`advertising` (original name): 👎 
`adtech`: 👎 
`ads`: 👎 
`money`: 👍 
`setupads`: 👍

philipjohn:
My despise of ad blockers grows.... Can you try again now please @sonjaleix ? Thanks for the tests @jeffersonrabb - I went with that name change.

claudiulodro:
I wouldn't worry about ad blockers and this feature too much, as it doesn't make sense to **set up ads with ad blocker enabled**. :P

sonjaleix:
Sorry @philipjohn still same issue. See [error messages](https://www.dropbox.com/s/gjpcsb6ut17uopt/Screenshot%202019-06-13%2015.02.45.png?dl=0).

PR #107, opened 2019-06-23
React Router

jeffersonrabb:
### All Submissions:

This PR introduces React Router (https://reacttraining.com/react-router/), replacing all existing approaches to screen navigation in wizards. This branch is fully merged into `add/react-router+plugin-requirements`, so this PR need not be merged on its own. I'm opening it mostly so the implementation of routing can be assessed without distractions. 

To test, try out the Subscription and Subscription Onboarding wizards, which should function exactly as before. Note the changing URL as you use them.

Also note the use of URL Parameters in the Subscriptions Wizard (e.g. paths like `/wp-admin/admin.php?page=newspack-subscriptions-wizard#/edit/282`), and programmatic routing after a new subscription is saved for the first time (the URL should change from `/create` to `/edit/{id}`).

I opted to use hash routing (`<HashRouter/>`) to avoid conflicting with wp-admin paths—the only part of the URL that will change is to the right of the #.

### DIFF
--- a/assets/components/src/action-card/index.js
+++ b/assets/components/src/action-card/index.js
@@ -34,6 +34,7 @@ class ActionCard extends Component {
 			className,
 			title,
 			description,
+			href,
 			notification,
 			notificationLevel,
 			actionText,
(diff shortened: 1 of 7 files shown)


jeffersonrabb:
Superseded by https://github.com/Automattic/newspack-plugin/pull/109, closing.

PR #106, opened 2019-06-21
Wizard Component + React Router

jeffersonrabb:
This is an experimental branch for discussion...

PR #104, opened 2019-06-14
Wizard and WizardScreen components for managing common wizard tasks

claudiulodro:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds `Wizard` and `WizardScreen` components and migrates the `Subscriptions` and `SubscriptionsOnboarding` wizards to use those components. These components should let us build wizards without needing to re-implement a bunch of common functionality each time.

- The `WizardScreen` component renders one screen of a wizard, with optional primary button and optional tertiary button.
- The `Wizard` component manages a collection of child `WizardScreen` components, displays the correct wizard screen, displays errors if necessary, and handles required plugins.

Here is an example of a super simple 3-screen wizard (taken from the components demo):
```
<Wizard 
	activeScreen={ activeWizardScreen } 
	requiredPlugins={ [ 'jetpack' ] }
	requiredPluginsCancelText={ __( 'Back to checklist' ) }
	onRequiredPluginsCancel={ () => console.log( 'Checklist' ) }
>
	<WizardScreen
		identifier='test-wizard-1'
		completeButtonText={ __( 'Continue to 2' ) }
		onCompleteButtonClicked={ () => this.setState( { activeWizardScreen: 'test-wizard-2' } ) }
		subCompleteButtonText={ __( 'Back to checklist' ) }
		onSubCompleteButtonClicked={ () => console.log( 'Checklist' ) }
	>
		<FormattedHeader
			headerText={ __( 'WizardScreen Component 1' ) }
			subHeaderText={ __( 'This is #1 wizard screen' ) }
		/>
	</WizardScreen>
	<WizardScreen
		identifier='test-wizard-2'
		completeButtonText={ __( 'Continue to 3' ) }
		onCompleteButtonClicked={ () => this.setState( { activeWizardScreen: 'test-wizard-3' } ) }
		subCompleteButtonText={ __( 'Back to 1' ) }
[...]

PR #109, opened 2019-06-23
React Router + Wizard Higher Order Components

jeffersonrabb:
### All Submissions:

This PR aims to achieve the goals of https://github.com/Automattic/newspack-plugin/pull/104 using a few different constructs. The functionality should be almost identical to 104, with a few minor differences noted below.

1) Routing 
All of the home-grown Wizard navigation has been replaced by routing, using [React Router](https://reacttraining.com/react-router/). I created a separate PR (not meant to  be merged) with only the routing piece, to make it a bit easier to assess how it works: https://github.com/Automattic/newspack-plugin/pull/107. Originally we had thought that Routing was a future nice-to-have, but a couple of factors changed my mind:

- The routing work in [104](https://github.com/Automattic/newspack-plugin/pull/104) is pretty deep and complicated (all the `identifier` stuff), which made this feel like a now-or-never kind of moment for routing.
- As part of the recent `Handoff` button work, the last visited Newspack URL is persisted so that the user can return from other plugins to exactly where they left off. Routing allows this to work with much greater precision, e.g. being able to return to the edit page of a specific Subscription.
- There's been discussion that the entire plugin might one day be one single page app as opposed to a series of different ones. This transition will be much easier if we are already using a standard routing strategy. 
- Routing improves the overall UI in a few ways, e.g. browser back button and history.
- React Router appears to be a widely used, well supported library. @iuravic has used it previously and spoke highly of it. 

2) Higher Order Components
Instead of the `Wizard` and `WizardScreen` components, I created two higher-order components: `withWizard` and `withWizardScreen`. These basically do  what the components did, wrapping components in another one which inject props, alter the render, and provide additional functionality. 

`withWizard` provides plugin requirement checking, error handling, fires `onWizardReady` in the wrapped component, and adds an optional button at the bottom.

`withWizardScreen` renders `<FormattedHeader/>`, wraps the screen in a `<Card/>` and renders an optional `<Button/>` at the bottom. 

## Differences From 104

1) In PR 104, `WizardScreen` renders primary and secondary buttons. Since the secondary button is generally going to be used to return to the relevant checklist screen, I opted to move this to `Wizard`. So each HOC now has one built-in button.

2) With routing introduced, `<Button/>` elements now frequently require an `href` rather than an `onClick` event. Rather than taking separate attributes for events and URLs, I added a utility `buttonProps` which will create a set of props for a `<Button/>` element. If the value is a function the prop will be an `onClick` event. If the value is a string it will be an `href`.

3) The `Wizard`/`WizardScreen` demo on Components Demo has been removed entirely. To make a demo of the new structure we would need to create a few new blank components for that purpose only, which felt wasteful. 

4) For reference, the following files have been copied from 104 verbatim:
`assets/components/src/with-wizard-screen/style.scss`
`assets/wizards/subscriptions/style.scss`
`includes/class-plugin-manager.php`
`includes/wizards/class-subscriptions-wizard.php`

## Two Failed Experiments
In the spirit of completeness, here are two other approaches I tried which didn't work out:

1) Initially I attempted to take PR 104 as-is and implement routing entirely within the `Wizard` and `WizardScreen` components, thus leaving the implementations in the wizards untouched. This work is here: https://github.com/Automattic/newspack-plugin/pull/106. I got pretty far with it, but it ultimately became utterly convoluted. There were lots of situations where I needed to change routes programmatically from components where this wasn't expected behavior and I ended up with something pretty messy... Eventually I abandoned the approach in favor of a "normal" implementation of React Router.

2) I experimented with keeping the `<HashRouter>` and `<Switch>` wrapping in `withWizard`. This would have been nice because each Wizard's render would only need to contain a series of `<Route/>`s, and the Plugin Management UI could be injected automatically. The approach worked, but broke the React `ref` I'm using to call `onWizardComplete` after plugin management is complete. I'm not positive why this was, but ultimately abandoned the approach. All that this means is that `<HashRouter>`, `<Switch>`, and `{ pluginRequirements }` must be included in every Wizard's render, which isn't a terribly great sacrifice.

1. `npm ci && npm run clean && npm run build:webpack -- --watch`
2. Test all functionality in the Subscription Onboarding and Subscriptions wizards. These should be effectively identical to how they work on `master`
3. Observe URL changes to verify routing. 

### Other information:


iuravic:
@jeffersonrabb, these are all minor things I found. I haven't neither "requested changes" nor "approved" yet -- would love to hear you thoughts on what you think should or should not be updated.

assets/wizards/subscriptionsOnboarding/index.js
```diff
@@ -58,288 +64,180 @@ class SubscriptionsOnboardingWizard extends Component {
...
+		);
 	}
 }
+
 render(
-	<SubscriptionsOnboardingWizard />,
+	createElement( withWizard( SubscriptionsOnboardingWizard), {
+		buttonText: __( 'Back to checklist' ),
+		buttonAction: newspack_urls[ 'checklists' ][ 'memberships' ],
+	} ),
```
claudiulodro: [This wizard has a requirement of the WooCommerce plugin](https://github.com/Automattic/newspack-plugin/pull/104/files#diff-07e22c86f90e246a09451508cbf0db5eR20), otherwise we get fatal errors as soon as the wizard is loaded:

> Unrecoverable error
> The WooCommerce plugin is not installed and activated. Install and/or activate it to access this feature.

assets/wizards/subscriptions/views/manageSubscriptionsScreen.js
```diff
@@ -71,31 +62,9 @@ class ManageSubscriptionsScreen extends Component {
...
-				>
-					{ __( "I'm done adding" ) }
-				</Button>
 			</div>
 		);
 	}
 }
 
-export default ManageSubscriptionsScreen;
+export default withWizardScreen( ManageSubscriptionsScreen );
```
claudiulodro: This should have the `noBackground` prop on `withWizardScreen`, otherwise there are white cards on a white background:

localhost_wp-admin_admin php_page=newspack-subscriptions-wizard

assets/wizards/subscriptions/index.js
```diff
@@ -129,33 +155,116 @@ class SubscriptionsWizard extends Component {
...
+						path="/edit/:id"
+						render={ routeProps => {
+							return (
+								<EditSubscriptionScreen
+									headerText={ __( 'Edit subscription' ) }
+									subHeaderText={ __( 'You are editing an existing subscription' ) }
+									subscription={ subscriptions[ routeProps.match.params.id ] || {} }
+									onChange={ this.onSubscriptionChange }
+									onClickSave={ subscription =>
+										this.saveSubscription( subscription ).then( () => this.refreshSubscriptions() )
```
claudiulodro: After saving a subscription, we should be dropped off on the Subscriptions Management screen (`#/`). I was dropped off on the Edit Subscription screen (`#/edit/:id`) after saving. Not a really big deal, but right now it forces a weird flow where you have to Save and then Cancel after saving to go to the Manage Subscriptions screen.

claudiulodro:
> some behavior not yet finished -- if Stripe was not enabled (inspector/checkbox togle control not turned on, and API key not entered), and then the "Save"/"Finish" is clicked, it still gives the red text message at the top "Publishable Key and Secret Key are required to use Stripe."

This was a logic error in the code I wrote when implementing the Stripe form validation. I also noticed this issue when testing. I'll have it solved in a new PR shortly.

jeffersonrabb:
> a design layout question: should the "Set up Stripe" header and "Stripe is the recommended..." description line in the Subscriptions Onboarding Wizard be located inside the Card component (currently it's above the card):

This was a mistake, fixed in: https://github.com/Automattic/newspack-plugin/pull/109/commits/b8dacb4e9d40ff3580242e9ad16a1657c1e2603c

jeffersonrabb:
> not sure whether this is very important, but just mentioning -- the "Save" button has a bit smaller top margin:

Yup, this happened as a result of changes to which Component renders the button. This fixes and achieves the same effect: https://github.com/Automattic/newspack-plugin/pull/109/commits/214ece3569135556b5ec012c07bc87708a391f1f

jeffersonrabb:
> and also the button label is changed to "Finish" (

Done in https://github.com/Automattic/newspack-plugin/pull/109/commits/957a1d729b48fd3af636cd0882610c808d77c92d

jeffersonrabb:
> here on the Create screen, when finished creating a new Subscription, and clicked "Save", it keeps the user on the Edit screen ?page=newspack-subscriptions-wizard#/edit/25 until the user clicks "Cancel". While on the master, when a Subscription is saved it returns the user to the list of Subcriptions. Which one of these 2 behaviors is OK?

This has been changed so that after the first Save of a new Subscription, the user is returned to the list view: https://github.com/Automattic/newspack-plugin/pull/109/commits/4569b99cc23afa9d38487d0710bc1083ca0214ab

jeffersonrabb:
> just to point out that there is a slight difference in design layout in the list of Subscriptions. To me the the new version seems much better, I just wanted to point this difference out.

This regression is fixed in https://github.com/Automattic/newspack-plugin/pull/109/commits/ae3d9296287c1f5d42557bdece9c4329d4264444. You may be right about liking the other way better, but for the purposes of this PR I'm shooting for fairly precise visual parity just to be sure none of these new directions create future obstacles. We should examine the design question separately.

assets/wizards/subscriptionsOnboarding/index.js
jeffersonrabb: Oops, don't know how that went missing... Fixed in https://github.com/Automattic/newspack-plugin/pull/109/commits/da38f2a2596cedc67477eca70c7d47b4d7bd67fd

assets/wizards/subscriptions/views/manageSubscriptionsScreen.js
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-plugin/pull/109/commits/ae3d9296287c1f5d42557bdece9c4329d4264444

jeffersonrabb:
> This change needs to happen to the Edit Subscription screen also. When saving a new subscription, you are now correctly returned to the Manage Subscriptions screen, but not when saving an existing subscription.

Done! https://github.com/Automattic/newspack-plugin/pull/109/commits/30911114485e4a758ce46502d8b6df717fd7b128. I'm going to leave it open for the night to let @iuravic have another look, then will land it first thing tomorrow unless something comes up.

jeffersonrabb:
Thank you @claudiulodro and @iuravic for the reviews!

includes/wizards/class-subscriptions-wizard.php
```diff
@@ -7,41 +7,33 @@
...
  */
 class Subscriptions_Wizard extends Wizard {
-
 	/**
 	 * The slug of this wizard.
 	 *
 	 * @var string
 	 */
 	protected $slug = 'newspack-subscriptions-wizard';
-
```
philipjohn: Minor, but I think these missing spacing lines make the code harder to read.

includes/wizards/class-subscriptions-wizard.php
```diff
@@ -171,35 +155,57 @@ public function register_api_endpoints() {
...
+	 *
+	 * @return bool | WP_Error True on success, WP_Error on failure.
+	 */
+	protected function check_required_plugins_installed() {
+		if ( ! function_exists( 'WC' ) || ! class_exists( 'WC_Subscriptions_Product' ) || ! class_exists( 'WC_Name_Your_Price_Helpers' ) ) {
+			return new WP_Error(
+				'newspack_missing_required_plugin',
+				esc_html__( 'The required plugins are not installed and activated. Install and/or activate them to access this feature.', 'newspack' ),
+				[
+					'status' => 400,
```
philipjohn: A nitpick of pedantic proportions: Instead of 400 (Bad Request), how about [404 (Not Found)](https://httpstatuses.com/404). It's not that the client sent a bad request, but that the server couldn't find the thing the client is looking for.

PR #111, opened 2019-06-24
Fix key validation logic in Stripe setup

claudiulodro:
### All Submissions:

Originally noticed/doicumented while testing #109. 

The logic that validates the Stripe keys was flawed. When Test Mode is enabled and valid test keys are entered, it would throw an error if valid publish keys were not also entered. That is incorrect behavior, as publish keys should not be validated if test mode is active.

0. Go to the Stripe setup in the Subscriptions Onboarding wizard. Enable Stripe.
1. Uncheck Test Mode, delete any values in the live key inputs, recheck Test Mode, enter some values for test keys. Save. It should save without throwing an error.
2. Check Test mode, delete any values in the test key inputs, uncheck Test Mode, enter some values for live keys. Save. It should save without throwing an error.
3. Verify errors are still thrown when a required field is missing on the Stripe settings.

### Other information:

### DIFF
--- a/includes/wizards/class-subscriptions-onboarding-wizard.php
+++ b/includes/wizards/class-subscriptions-onboarding-wizard.php
@@ -370,7 +370,7 @@ class Subscriptions_Onboarding_Wizard extends Wizard {
 					'level'  => 'notice',
 				]
 			);
-		} elseif ( ! $this->api_validate_not_empty( $args['publishableKey'] ) || ! $this->api_validate_not_empty( $args['secretKey'] ) ) {
+		} elseif ( ! $args['testMode'] && ( ! $this->api_validate_not_empty( $args['publishableKey'] ) || ! $this->api_validate_not_empty( $args['secretKey'] ) ) ) {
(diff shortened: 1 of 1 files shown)


claudiulodro:
Not sure what happened with the "The response is not a valid JSON response" error, especially if you can't reproduce it. That error means that there was some sort of unexpected output in the API response messing up the JSON. It could be that a notice was thrown during the API call or something, but I can't reproduce it either.

> while this modal was displayed, neither the "Return to Checklist" button, nor the "x" modal button didn't perform any actions (they didn't close the modal, or redirect to the Checklist). So if this seems like a good opportunity, we might want to double check how this modal is put in place and whether its buttons are assigned proper actions. 

That modal implementation wasn't fully completed when this PR was opened. Now that the router/wizard components stuff is merged into master, [the modal in master should actually do stuff](https://github.com/Automattic/newspack-plugin/pull/109/files#diff-dea40b85ed64817cd968390a00610990R92). :)

PR #112, opened 2019-06-25
Improved handling for plugins that Newspack can't install

claudiulodro:
### All Submissions:

This PR improves the handling for plugins that can't be installed by Newspack because the plugin URL's aren't public. This could be premium plugins, Site Kit, etc. Attempting to install one of those plugins through the API or Plugin_Manager will fail and provide a useful error about where to get the plugin.

Plugins without `Download` info in the `Plugin_Manager` are plugins that cannot be installed. Information about where to get the plugin is pulled from the plugin's `PluginURI` info.

1. `npm run clean; npm run build:webpack`
2. In the Components Demo, try and install Site Kit. Observe it fails with a useful error:
localhost_wp-admin_admin php_page=newspack-components-demo

In the future once the plugin installer component is fine-tuned to be more user friendly, we can probably hide the "Use" button and just display a "This plugin can't be installed" message instead, so the user doesn't need the installation to fail before finding out.

### Other information:

### DIFF
--- a/assets/wizards/componentsDemo/index.js
+++ b/assets/wizards/componentsDemo/index.js
@@ -99,7 +99,7 @@ class ComponentsDemo extends Component {
 					/>
 					<Handoff
 						className="is-centered"
-						plugin="google-site-kit-wp"
+						plugin="google-site-kit"
 					/>
 					<Handoff
 						className="is-centered"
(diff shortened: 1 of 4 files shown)


claudiulodro:
Unit tests need updating, as the plugin manager can now only install plugins by slug that it knows.

claudiulodro:
This should be ready for review/merge now. Here is a summary of the changes:
- Added the download URL for Site Kit. You can test it works in the Plugin Installer component demo.
- WooCommerce Subscriptions is in the Plugin Installer component demo for testing failed installation of premium plugins.
- Updated the unit tests to use the AMP plugin instead of Hello Dolly. They take slightly longer now because it's a larger plugin, so we might want to update the tests in the future once we have a stand-alone .org plugin available that is smaller (Akismet?).

PR #114, opened 2019-06-25
Setup Wizard

jeffersonrabb:
### All Submissions:

This is the Set Up Wizard, the first thing a user encounters in Newspack. There's a lot in this PR—timing was too tight to effectively break it down into component parts. Apologies in advance! 

If you have not completed Set Up, you will be redirected to the Wizard (`/wp-admin/admin.php?page=newspack-setup-wizard#/`) when you try to go to the Dashboard. After fully completing set up you will be able to access the dashboard again. (To re-test this, you may have to go into the DB and remove `newspack_setup_complete` from the options table. I've added Setup to Newspack's WP-admin menu, so there's a way back in after completing.

The Wizard itself is a set of four screens: Welcome, About, Newsroom, Configure Plugins. After advancing from the Welcome screen the plugin installation begins, using `PluginInstaller` with the `asProgressBar` prop. As installation proceed, the user is asked to answer a few questions.

The questions they answer give us their "Publication Profile". Note: the Country and Currency fields use functions duplicated from WooCommerce. Unlike the Subscriptions Wizard we can't use Woo's functions since the plugin won't necessarily be installed at this point.

On the third screen, the button is disabled until plugin installation competes. When they are all installed and activated the user can advance to the Plugin Configuration screen.

Here there are two `PluginLinkCard`s for Jetpack and Sitekit. These will have a check mark if the plugin has been configured. (For the moment Site Kit will always be marked as configured, pending completion of https://github.com/Automattic/newspack-plugin/pull/134). Jetpack will be checked if the site has been connected to WP.com.

Clicking either of the Cards triggers `Handoff` flow. This PR contains a change to how Handoff works, where the URL to return to is set via API endpoint just before redirecting, which insures that the URL hash is recorded which leads to accurate round trips... After clicking any of the `Handoffs`, try clicking `Back to Newspack` and make sure it takes you to the exact screen you came from. 

The `Start Configuration` button should be a `Handoff` to the first of the two to not be configured.

After both are configured, the button text changes to `Finish`. Clicking this updates the DB so that the dashboard becomes accessible, and redirects the browser to dashboard. 

Since this one is a pretty complex flow, I would advise testing by studying the code, then simply working through the wizard in as many  ways as you can.

### Other information:


jeffersonrabb:
Initial UI questions for @sonjaleix. 

Note that the pagination and back link are missing here—I'm developing them in a separate branch.

1) How does the width look? Normal `Card` components have a `max-width` of 440px. For this Wizard the value is increased to `680px`.

2) When do we handle automatic (invisible to user) plugin configuration? Do we do this immediately as part of installation or is there a step where they click a button? e.g. AMP

3) When do we handle Handoff-style configuration?  e.g. Jetpack, Site Kit? (Site Kit is currently not installed by this Wizard, but that's due to a technical issue still being worked on)

sonjaleix:
@jeffersonrabb I am getting an error when I'm trying to build this PR. 
To your questions: 
> How does the width look? Normal Card components have a max-width of 440px. For this Wizard the value is increased to 680px.

We should implement a unified width for all wizards. I would recommend to look at the new Muriel design grid and land on an 8-column width for desktop. I've mocked this up in Figma for us.

> When do we handle automatic (invisible to user) plugin configuration? Do we do this immediately as part of installation or is there a step where they click a button? e.g. AMP

This should be run automatically in the background. We could indicate through a running progress bar on the second screen (see designs). Something like this: https://giphy.com/gifs/dsLZmo8rZOuOAgMr7o

> When do we handle Handoff-style configuration? e.g. Jetpack, Site Kit? (Site Kit is currently not installed by this Wizard, but that's due to a technical issue still being worked on)

What do you mean by "handoff-style configuration"? Do you mean, how do we guide the user to setup Jetpack for example after the initial Newspack onboarding setup flow?

sonjaleix:
Upon reviewing the prototype on local, I ran into the following things: 
- Some of the items in the country and currency selectors don't display special characters currectly
- You can't select the first item in the dropdown selectors (see: http://www.giphy.com/gifs/ZElwLEW7GHcX6LYa8x)

Other than that it's looking good!

jeffersonrabb:
> * Some of the items in the country and currency selectors don't display special characters currectly

This is fixed in https://github.com/Automattic/newspack-plugin/pull/114/commits/a3496376a048ab89b7bc3825fa1298fff4eb3838

jeffersonrabb:
> * You can't select the first item in the dropdown selectors

I'm not seeing this problem. Which browser are you using?

sonjaleix:
> I'm not seeing this problem. Which browser are you using?

I was using Chrome on Mac, but I just tested in Safari and Firefox and am finding the same bug.

sonjaleix:
> Some of the items in the country and currency selectors don't display special characters currectly
> This is fixed in a349637

I just pulled the latest version and I'm still seeing special characters in the country and currency dropdowns. 
Screenshot 2019-06-26 11 47 54
Screenshot 2019-06-26 11 47 40

assets/wizards/setup/index.js
```diff
@@ -0,0 +1,275 @@
...
+						<Route
+							path="/newsroom"
+							render={ routeProps => (
+								<Newsroom
+									headerText={ __( 'Tell us about your Newsroom' ) }
+									subHeaderText={ __(
+										'The description helps set the stage for the step content below'
+									) }
+									buttonText={ __( 'Continue' ) }
+									buttonAction="#/configure-plugins"
```
claudiulodro: This should also have `onClick: () => this.updateProfile()` like the address settings, right? Currently these settings don't save.

assets/wizards/setup/components/plugin-link-card/style.scss
```diff
@@ -0,0 +1,50 @@
...
+		position: absolute;
+		top: 14px;
+		width: 16px;
+		svg {
+			width: 16px;
+			height: 16px;
+			color: $muriel-white;
+		}
+	}
+	&.is-configured {
```
claudiulodro: Suggestion: Can we make it clearer that a plugin has been configured? Adding the word 'Configured' next to the check (and maybe decreasing the opacity further) would help. When I first went through the wizard I assumed the small green check meant the plugin was configured, but I wasn't super sure.

assets/wizards/setup/views/newsroom/style.scss
```diff
@@ -0,0 +1,13 @@
+.newspack-setup-wizard_newsroom-screen_plugin_group {
+	.muriel-checkbox {
+		margin-left: 2em;
+		margin-right: 2em;
+		float: left;
+    	min-width: calc( 50% - 4em );
```
claudiulodro: Mixed spaces and tabs on this line.

assets/wizards/setup/views/welcome/index.js
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * Render.
+	 */
+	render() {
+		return (
+			<div className="newspack-setup-wizard__welcome">
+				<p>{ __( 'The following wizard will help you set up Newspack.' ) }</p>
+				<div className="newspack-setup-wizard_image_container" />
+				<p>
+					'Newspack brings together the power of a suite of plugins. We will install the following
```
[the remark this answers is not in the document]
claudiulodro: Also this content should be run through localization.

assets/wizards/subscriptionsOnboarding/index.js
```diff
@@ -224,7 +224,7 @@ class SubscriptionsOnboardingWizard extends Component {
 }
 
 render(
-	createElement( withWizard( SubscriptionsOnboardingWizard, [ 'woocommerce' ] ), {
```
claudiulodro: Not sure why the WooCommerce requirement was removed. I get `Unrecoverable error: The WooCommerce plugin is not installed and activated. Install and/or activate it to access this feature.` if I try accessing the wizard without having WC installed.

includes/api/class-plugins-controller.php
```diff
@@ -210,7 +210,12 @@ public function get_item( $request ) {
 		}
 
 		$managed_plugins = Plugin_Manager::get_managed_plugins();
-		return rest_ensure_response( $managed_plugins[ $slug ] );
+
+		$plugin = $managed_plugins[ $slug ];
+
+		$plugin['Configured'] = \Newspack\Configuration_Managers::is_configured( $slug );
```
claudiulodro: Good idea 👍

includes/configuration_managers/class-configuration-managers.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Newspack;
+
+use \WP_Error;
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * List and instantiate configuration managers.
+ */
+class Configuration_Managers {
```
claudiulodro: This class was a good idea. 👍 I think we'll get a lot of use out of it.

includes/util.php
```diff
@@ -76,3 +76,454 @@ function newspack_install_activate_theme() {
...
 }
+
+/**
+ * Get full list of currency codes. Copied from https://github.com/woocommerce/woocommerce/blob/master/includes/wc-core-functions.php
+ *
+ * @return array
+ */
+function newspack_currencies() {
+	$currencies = array(
+		'AED' => __( 'United Arab Emirates dirham', 'woocommerce' ),
```
claudiulodro: We probably want `newspack` as the textdomain, right?

includes/util.php
```diff
@@ -76,3 +76,454 @@ function newspack_install_activate_theme() {
...
+}
+
+/**
+ * Get full list of country codes. Copied from https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-countries.php
+ *
+ * @return array
+ */
+function newspack_countries() {
+	$countries = array(
+		'AF' => __( 'Afghanistan', 'woocommerce' ),
```
claudiulodro: Same comment as above about the textdomain.

includes/wizards/class-setup-wizard.php
```diff
@@ -0,0 +1,144 @@
...
+	 *
+	 * @var string
+	 */
+	protected $slug = 'newspack-setup-wizard';
+	/**
+	 * The capability required to access this wizard.
+	 *
+	 * @var string
+	 */
+	protected $capability = 'edit_products';
```
claudiulodro: Probably want `install_plugins` and/or `manage_options` instead of `edit_products`.

includes/wizards/class-setup-wizard.php
```diff
@@ -0,0 +1,144 @@
...
+		update_option( NEWSPACK_SETUP_COMPLETE, true );
+		return rest_ensure_response( [ 'complete' => true ] );
+	}
+
+	/**
+	 * Enqueue Subscriptions Wizard scripts and styles.
+	 */
+	public function enqueue_scripts_and_styles() {
+		parent::enqueue_scripts_and_styles();
+		wp_enqueue_media();
```
claudiulodro: We don't need `wp_enqueue_media` unless there's an image uploader (it enqueues the media modal JS). Sorry, I probably should have documented that better.

includes/wizards/class-setup-wizard.php
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * If initial setup is incomplete, redirect to Setup Wizard.
+	 */
+	public function redirect_to_setup() {
+		if ( get_option( NEWSPACK_SETUP_COMPLETE ) ) {
+			return;
+		}
+		try {
+			$screen = get_current_screen();
```
claudiulodro: `get_current_screen` will never throw an exception. You can [see the source here](https://developer.wordpress.org/reference/functions/get_current_screen/) (it's only like 6 lines).

jeffersonrabb: Removed in https://github.com/Automattic/newspack-plugin/pull/114/commits/1d9b54030ad995992a90a1cdc8d224079aded128

I did this only because of this scary line in this docs: https://codex.wordpress.org/Function_Reference/get_current_screen

```
This function is defined on most admin pages, but not all. Thus there are cases where is_admin() will return true, but attempting to call get_current_screen() will result in a fatal error because it is not defined. One known example is wp-admin/customize.php.
```

includes/class-profile.php
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	/**
+	 * Retrieve one option, prefixing the option name for Newspack profile.
+	 *
+	 * @param string $key The unprefixed option name.
+	 * @return string The option value.
+	 */
+	public static function newspack_get_option( $key ) {
+		return get_option( NEWSPACK_PROFILE_OPTION_PREFIX . $key );
```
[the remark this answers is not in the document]
jeffersonrabb: Good call. I cleared the options rows and tried again, and now we see the correct labels in all the inputs. https://github.com/Automattic/newspack-plugin/pull/114/commits/e575bea6eca3f9a8c794eb92f0bfad56b73be86c

assets/wizards/setup/views/newsroom/index.js
```diff
@@ -0,0 +1,141 @@
...
+					] }
+				/>
+				<SelectControl
+					label={ __( 'Content focus' ) }
+					value={ content_focus }
+					onChange={ value => updateProfile( 'content_focus', value ) }
+					options={ [
+						{ label: __( 'Local News' ), value: 'local' },
+						{ label: __( 'National News' ), value: 'national' },
+						{ label: __( 'Topical' ), value: 'topical' },
```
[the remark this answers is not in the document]
jeffersonrabb: Added in https://github.com/Automattic/newspack-plugin/pull/114/commits/13646ddade71d03893dbd3fb2ff605f61dd530ff. Note that all of these fields will be edited for content at some point.

assets/wizards/subscriptionsOnboarding/index.js
jeffersonrabb: This was almost certainly a mistake on my part, probably to test something. Sorry, and good catch! https://github.com/Automattic/newspack-plugin/pull/114/commits/033095342e5b98cbdf3066a6b545f4c05e858f88

includes/wizards/class-setup-wizard.php
jeffersonrabb: Good to know. https://github.com/Automattic/newspack-plugin/pull/114/commits/40ab8d905ccc751eb37e269dffbd00ccba19cfe6

includes/wizards/class-setup-wizard.php
jeffersonrabb: Good call: https://github.com/Automattic/newspack-plugin/pull/114/commits/74493922971504fc6879cd64ac9aba0bc11d4e73

includes/util.php
jeffersonrabb: Yeah, good catch: https://github.com/Automattic/newspack-plugin/pull/114/commits/229ca041d293900d763d9a93de3eee43e9bf4b0f

includes/util.php
jeffersonrabb: https://github.com/Automattic/newspack-plugin/pull/114/commits/229ca041d293900d763d9a93de3eee43e9bf4b0f

assets/wizards/setup/views/newsroom/style.scss
jeffersonrabb: https://github.com/Automattic/newspack-plugin/pull/114/commits/4bac47f93e8ebdce0927666070714808fd5fe9ce

assets/wizards/setup/index.js
jeffersonrabb: Good eye, I probably never would have caught this! https://github.com/Automattic/newspack-plugin/pull/114/commits/a8e730eb198c3ba8ac3b36ed2855f4dfd221cc57

assets/wizards/setup/views/welcome/index.js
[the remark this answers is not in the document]
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-plugin/pull/114/commits/ccfb5e3be777e6ae3f4ed42c1ed5f4a93e1c590a and https://github.com/Automattic/newspack-plugin/pull/114/commits/cdb400af7d0a705512c5cc7ee89416630d716516

jeffersonrabb:
Note, this was just rebased to bring the Dashboard changes in...

assets/wizards/setup/components/plugin-link-card/style.scss
sonjaleix: Good point @claudiulodro . I've updated the design to this so there is a separate link in each plugin card. This will help indicate that they can be clicked individually: 
Screenshot 2019-06-28 16 12 34

And this is what it would look like if one of the plugins has been configured: 
Screenshot 2019-06-28 16 01 55

assets/wizards/setup/index.js
```diff
@@ -0,0 +1,275 @@
...
+									buttonDisabled={ ! installationComplete }
```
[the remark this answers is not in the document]
sonjaleix: I agree, we'll need to communicate to the user what is happening and why they can't progress. (Hopefully the plugin installation process is completed before the user is done filling out the forms in most cases.)

We could look at a few options here. Let's start with testing your suggestion: using explanatory text in the "inactive" button. Another option could be adding a tooltip, though I think that's not a very common pattern on an action button. For A11y, we might need to add screenreader text since inactive buttons are often skipped by screen readers.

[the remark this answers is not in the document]
jeffersonrabb: I've done a version of this in https://github.com/Automattic/newspack-plugin/pull/114/commits/4e0ac93145eb2b47f0e67ef050532be7d7c03067

assets/wizards/setup/components/plugin-link-card/style.scss
jeffersonrabb: This is done in https://github.com/Automattic/newspack-plugin/pull/114/commits/4e0ac93145eb2b47f0e67ef050532be7d7c03067

sonjaleix: Looks great! Approved.

jeffersonrabb:
Thank you both for digging in on this one!

PR #121, opened 2019-06-26
Handoff Banner Style Change

jeffersonrabb:
### All Submissions:

Per a suggestion from @sonjaleix this PR adds margins around the Handoff banner based on this screenshot:

Screenshot 2019-06-20 16 52 40

2. Navigate to `Newspack->Components Demo`
3. Click any of the `Handoff` buttons.
4. Observe the new banner styles. 

### Other information:

### DIFF
--- a/assets/components/src/button/style.scss
+++ b/assets/components/src/button/style.scss
@@ -50,6 +50,7 @@
 		background: none;
 		border: none;
 		color: $muriel-gray-300;
+		cursor: pointer;
 		text-decoration: underline;
 		&:hover {
 			text-decoration: none;
--- a/assets/wizards/handoff-banner/style.scss
+++ b/assets/wizards/handoff-banner/style.scss
@@ -6,11 +6,11 @@
 	background-size: 32px 32px;
 	box-shadow: 0 2px 2px #cccccc;
 	display: flex;
-	margin: 0 0 0 -20px;
+	margin: 20px 0 0 0;
 	min-height: 52px;
 	padding: 0;
 	position: relative;
-	width: calc( 100% + 20px );
+	width: calc( 100% - 20px );
 	.jetpack-pagestyles & {
 		margin-left: 0;
 		width: 100%;
--- a/assets/wizards/handoff-banner/style.scss
+++ b/assets/wizards/handoff-banner/style.scss
@@ -1,4 +1,9 @@
 #newspack-handoff-banner {
+	display: block;
+	padding: 0;
+	position: relative;
+}
+.newspack-handoff-banner {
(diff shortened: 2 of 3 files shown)


claudiulodro:
Not directly relevant to this PR, but I also observed that clicking the 'Dismiss' button leaves an empty bar at the top of the screen (The parent `#newspack-handoff-banner` element that the bar is rendered within should get removed). Also, after dismissing, refreshing the page brings the bar back (is this the expected behavior or should it be a persistent dismiss?). 

An easy solution for both of these issues would be to make the 'Dismiss' button a regular link that goes to the current page with a query param (e.g. `/wp-admin/admin.php?page=wc-settings&dismiss_newspack_handoff`). In the `persist_current_screen` method, if `dismiss_newspack_handoff` is in `$_GET`, clear the `NEWSPACK_HANDOFF` option. We can keep doing it the current way if you'd like; this is just how I'd do it. :)

A small, non-blocking style issue I noticed is that the 'Dismiss' button has the default cursor, instead of the "clickable link" cursor.

jeffersonrabb:
> On screens with Screen Meta Boxes (e.g. screen options dropdown), the handoff banner buds up against it, we should try keep the margin top alive there.

This is a bit tough—I wasn't able to find a solution that works all the time. I added some logic that handles most  of the cases where those buttons collide with our bar (here https://github.com/Automattic/newspack-plugin/pull/121/commits/c4191091cfcbd2b489fe11dbd5bd86f42c10f3a1) but as we implement more Handoffs we may find other cases that need special handling. I'd say let's deal with them as they come up.

jeffersonrabb:
> Not directly relevant to this PR, but I also observed that clicking the 'Dismiss' button leaves an empty bar at the top of the screen (The parent `#newspack-handoff-banner` element that the bar is rendered within should get removed). Also, after dismissing, refreshing the page brings the bar back (is this the expected behavior or should it be a persistent dismiss?).

I fixed the main dismiss problem in https://github.com/Automattic/newspack-plugin/pull/121/commits/cd238cfc7e954ad7bb7ed53db03403b3e95a073b. As for updating the database so the banner stays hidden, I think we should hold on that for a minute until we have a full understanding of how we're using Handoffs. It raises some tricky questions - does one dismissal hide the banner forever, or just until the next handoff is clicked? Should the dismissal be per-user? Etc. So, I propose we leave it for now and implement later if appropriate.

sonjaleix:
This is good to go from my side. Agreed on holding off on a decision on the "dismiss" issue.

jeffersonrabb:
Thank you!

PR #126, opened 2019-06-26
Add/plugin configurator

jeffersonrabb:
### All Submissions:

Introduces the `Configuration_Manager` abstract class and two subclasses, for AMP and Jetpack. Ultimately, there will be a subclass for every plugin that requires any configuration. Currently the AMP Configuration Manager switches to AMP strict mode when `configure()` is called. The Jetpack Configuration Manager shows how `is_configured()` can be used to determine if Jetpack has been connected. 

This also introduces a new endpoint: `newspack/v1/plugins/{plugin-slug}/configure`. This endpoint will install and activate the plugin, then if a configuration class has been specified it will be loaded and `configure()` called. If the plugin is already active, this endpoint won't fail like `/activate` does. Instead the error will be ignored and the plugin configured normally.

Testing can be a little tricky. In one approach, you could throw some PHP somewhere and try the calls out. This might look like this:

```
require_once NEWSPACK_ABSPATH . '/includes/configuration_managers/class-jetpack-configuration-manager.php'; 
$config_manager = new Jetpack_Configuration_Manager();
if ( $config_manager->is_configured() ) { 
  echo '<h1>Jetpack Configured</h1>';
}
```

The `<h1>` will be shown only if Jetpack is installed and connected.

Alternatively, edit the `PluginInstaller` component and change the `activate` endpoint to `configure` here: https://github.com/Automattic/newspack-plugin/blob/master/assets/components/src/plugin-installer/index.js#L78. Run the build script, then go to the Components Demo and try installing AMP. To verify the change, go into the AMP dashboard and switch the mode to Reader or Transitional. Then delete AMP. Go to Components Demo and install AMP. It should  install and activate normally, and after this is done you should see that the mode has been changed to Standard.

### Other information:

### DIFF
--- a/includes/configuration_managers/class-configuration-manager.php
+++ b/includes/configuration_managers/class-configuration-manager.php
@@ -1,0 +1,96 @@
+<?php
+/**
+ * Plugin configuration manager, abstract class.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+use \WP_Error;
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Provide an interface for configuring and querying the configuration of Newspack-managed plugins.
+ */
+abstract class Configuration_Manager {
+	/**
+	 * The slug of the plugin.
+	 *
+	 * @var string
+	 */
+	protected $slug = '';
+
+	/**
+	 * Retrieve data for this plugin.
+	 *
+	 * @var object
+	 */
+	protected function get_plugin_data() {
+		$managed_plugins = Plugin_Manager::get_managed_plugins();
+		try {
+			$plugin = $managed_plugins[ $this->slug ];
+		} catch ( Exception $e ) {
+			return new WP_Error( 'newspack_plugin_not_found', __( 'The plugin is not found.', 'newspack' ) );
+		}
+		return $plugin;
+	}
+
+	/**
+	 * Determines whether the plugin is installed and activated.
+	 *
+	 * @return bool Plugin activation state.
+	 */
+	public function is_active() {
+		$plugin = $this->get_plugin_data();
+		$status = isset( $plugin['Status'] ) ? $plugin['Status'] : null;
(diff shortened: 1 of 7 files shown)


includes/configuration_managers/class-jetpack-configuration-manager.php
```diff
@@ -0,0 +1,49 @@
...
+	 */
+	public $slug = 'jetpack';
+
+	/**
+	 * Get this wizard's name.
+	 *
+	 * @return string The wizard name.
+	 */
+	public function configure() {
+		return true;
```
claudiulodro: Nitpicks: The docblock for `configure` in the abstract class says it shouldn't return anything. Also the docblock for this method is clearly from some other method. :)

includes/configuration_managers/class-configuration-manager.php
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * Retrieve data for this plugin.
+	 *
+	 * @var object
+	 */
+	protected function get_plugin_data() {
+		$managed_plugins = Plugin_Manager::get_managed_plugins();
+		try {
+			$plugin = $managed_plugins[ $this->slug ];
+		} catch ( Exception $e ) {
```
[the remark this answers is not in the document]
jeffersonrabb: Reworked this in https://github.com/Automattic/newspack-plugin/pull/126/commits/6791213f9057dd7beb22eeb13fcd54bb02742fca

includes/configuration_managers/class-jetpack-configuration-manager.php
jeffersonrabb: Sloppy, sloppy... https://github.com/Automattic/newspack-plugin/pull/126/commits/ad64293d40396d7e251d9a4411e1e4137fe3f398

jeffersonrabb:
Great, thank you!

PR #132, opened 2019-06-27
Alternate Plugin Installer UI

jeffersonrabb:
### All Submissions:

The Setup Wizard installs plugins using different UI from the normal `PluginInstaller` component. It uses a `ProgressBar`, and installation begins automatically rather than requiring a `Button` click. To avoid code duplication, this PR brings this UI and functionality to the `PluginInstaller` component if it has the `asProgressBar` prop. Also, the `Use All` `Button` is hidden until plugin data has been retrieved, which will remove the flash of incorrect UI that appears in cases where the component determines that all plugin requirements have been met and hides itself.

2. Deactivate and remove these plugins: WooCommerce, AMP, Yoast, Site Kit
3. Navigate to `Newspack->Components Demo`
4. Click `Show Plugin Installer w/Progress Bar`.
5. Observe a `ProgressBar` appears and fills in as the four plugins are installed.

### Other information:

### DIFF
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -13,7 +13,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies.
  */
-import { ActionCard, Button } from '../';
+import { ActionCard, Button, ProgressBar } from '../';
 import './style.scss';
 
 const PLUGIN_STATE_NONE = 0;
@@ -35,7 +35,10 @@ class PluginInstaller extends Component {
 
 	componentDidMount = () => {
 		const { plugins } = this.props;
-		this.retrievePluginInfo( plugins );
+		this.retrievePluginInfo( plugins ).then( () => {
+			const { asProgressBar } = this.props;
+			if ( asProgressBar ) this.installAllPlugins();
+		} );
 	};
 
 	componentDidUpdate = prevProps => {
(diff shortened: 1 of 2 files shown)


assets/components/src/plugin-installer/index.js
```diff
@@ -187,15 +203,15 @@ class PluginInstaller extends Component {
 								notificationLevel="error"
 							/>
 						);
+						<Button
```
claudiulodro: The `Button` is after the `return`, so it will never get rendered? It never showed up in my testing.

jeffersonrabb: This is correct actually. The `<Button>` is only needed in the normal non-`asProgressBar` context, since `asProgressBar` causes the plugins to begin installing as soon as the Component mounts. It was confusing that I made this change in the same PR—ideally I would have done this one separately.

claudiulodro: I mean it will never render in any context (it's within the `map` after the `return`):
```
slugs.map( slug => {
    return (
	    <ActionCard/>
    );
    <Button/>
})
```
You can see no button on e.g. the Subscriptions Wizard:
localhost_wp-admin_admin php_page=newspack-subscriptions-wizard (1)

jeffersonrabb: My mistake, you are correct. This should fix: https://github.com/Automattic/newspack-plugin/pull/132/commits/014bc4313c4b0b163c57314901ebe36c909e3611

jeffersonrabb:
Good catch, and thanks for the review.

PR #134, opened 2019-06-27
Site Kit configuration manager

claudiulodro:
### All Submissions:

This PR adds a configuration manager for the Site Kit plugin. We can't actually really manage the configuration of the plugin, but we can get information about how configuration is going.

Since Site Kit requires a real connection to Google to set up, it's somewhat difficult to test this. This is the way I did it, but it's 100% janky:

1. For testing an unconnected site, I used my local environment.
2. For testing a connected site I created a new site on https://jurassic.ninja, installed Site Kit and did the Site Kit setup.
3. Copy + paste the `Site_Kit_Configuration_Manager` class into a blank file. Remove the `extends Configuration_Manager` and add an `is_active` method so it can be a stand-alone class.
3. In the Plugin Editor on the jurassic.ninja site, add the stand-alone class somewhere, hook into the `init` action and test the methods. e.g.
```
$config = new Site_Kit_Configuration_Manager();
var_dump( $config->is_configured() );
var_dump( $config->get_module_info( 'adsense' ) );
var_dump( $config->is_module_configured( 'adsense' ) );
});
```

### Other information:


jeffersonrabb:
Quick note about the testing - I had to require the class so the complete testing code I used was:

```
include_once NEWSPACK_ABSPATH . 'includes/configuration_managers/class-site-kit-configuration-manager.php';
			$config = new Site_Kit_Configuration_Manager();
			var_dump( $config->is_configured() );
			var_dump( $config->get_module_info( 'adsense' ) );
			var_dump( $config->is_module_configured( 'adsense' ) );
		});
```

jeffersonrabb:
`is_configured()` seems to be returning true even after disconnecting Site Kit from my Google Account. I also tried deleting and reinstalling, and after reinstallation `is_configured()` immediately returns true again.

claudiulodro:
My understanding from looking through the source code is that this is the correct behavior. [Each user that wants to access the settings needs to connect](https://github.com/google/site-kit-wp/blob/develop/includes/Core/Authentication/Clients/OAuth_Client.php#L329). **Connecting controls whether you can access the settings, not whether Site Kit is configured.** 

For example, you can try setting up the Google Analytics module. When GA is set up and running, you should have a `ga` object available in the JS console and some script from `https://www.googletagmanager.com` in the page source. Disconnect and you should still see these.

If you go to **Site Kit > Settings > Admin Settings > Reset Site Kit** you can disconnect Site Kit for real, and I've just added logic to handle when that happens in https://github.com/Automattic/newspack-plugin/pull/134/commits/5bcbc8dfed28eddd99d1d79e04aa655ab936180c

I think we will need to be able to tell whether the user is connected (because they can't set up any modules if they're not connected). I've added a method for this in https://github.com/Automattic/newspack-plugin/pull/134/commits/c61b6d7a4be85ada2f256d2aa74528b1fb0177e8

PR #140, opened 2019-06-30
Improved Plugin Loading Flow

jeffersonrabb:
### All Submissions:

This PR improves the plugin installer loading UI in a few ways:

1) Previously wizards with required plugins always defaulted to the `/` route after plugin installation. This has been reworked here so that a wizard can begin  on a deeper route but still perform plugin installation check.
2) Even if all plugins have been installed, there will be a flash of the "Required plugins" header before the initial plugin data is retrieved. Replacing `onComplete` with a more sophisticated `onStatus` event makes it possible to hide this header.
3) "Retrieving plugin information" UI with `Spinner` added to `PluginInstaller`.

1. On `master` branch navigate directly to `/wp-admin/admin.php?page=newspack-subscriptions-onboarding-wizard#/stripe`. Observe a flash of the `Required plugins` header. Observe that after plugin requirements are satisfied, the Wizard redirects to the `/` route.
2. Switch to `fix/plugin-loading-flow`, and run `npm ci && npm run clean && npm run build:webpack`
3. Navigate again to `/wp-admin/admin.php?page=newspack-subscriptions-onboarding-wizard#/stripe`
4. Observe the headers are suppressed, "Retrieving plugin information" UI is shown, and the Wizard ultimately lands on the correct route.

### Other information:

### DIFF
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -165,6 +165,12 @@ class PluginInstaller extends Component {
 		}
 		return (
 			<div>
+				{ ! pluginInfo.length && (
+					<div className="newspack-plugin-installer_waiting">
+						<p>{ __( 'Retrieving plugin information...') }</p>
+						<Spinner />
+					</div>
+				) }
 				{ pluginInfo &&
 					slugs.length > 0 &&
 					slugs.map( slug => {
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -29,7 +29,7 @@ class PluginInstaller extends Component {
 	constructor() {
 		super( ...arguments );
 		this.state = {
-			pluginInfo: [],
+			pluginInfo: {},
 		};
 	}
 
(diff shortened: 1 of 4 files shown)


assets/components/src/plugin-installer/index.js
```diff
@@ -167,6 +165,12 @@ class PluginInstaller extends Component {
 		}
 		return (
 			<div>
+				{ ! pluginInfo.length && (
```
claudiulodro: The spinner is always active in my testing, even after loading is complete:
localhost_wp-admin_admin php_page=newspack-subscriptions-wizard (2)

This is because once plugin info is retrieved, the `pluginInfo` field is an object, not an array (so `length` is `undefined`, and `! length` is always `true`).

jeffersonrabb: Good catch! This should fix it up: https://github.com/Automattic/newspack-plugin/pull/140/commits/6ea3d64d59a584df6f7ea8992aab11a1be90ab22

PR #113, opened 2019-06-25
Wizard Pagination/Navigation

jeffersonrabb:
An experiment in implementing pagination and navigation for Wizards.

PR #142, opened 2019-07-01
Wizard Pagination

jeffersonrabb:
### All Submissions:

This PR adds the `WizardPagination` component and uses it in the Set Up Wizard. The component will render a BACK button in the upper left corner and "Page {n} of {total}" UI in the upper right corner. 

Screen Shot 2019-07-01 at 11 17 02 AM

`WizardPagination` should be added anywhere within a `HashRouter`. The use of `withRouter` will cause errors if used outside of a router. The component requires one prop: `routes`. This should be an array of the page routes, in order. In an earlier iteration (https://github.com/Automattic/newspack-plugin/pull/113) I played with deriving the array of routes programmatically, but recent work on the plugin requirements handling illustrated that there will be `Routes` found that should not be included in pagination, so I opted to specify them manually.

One UI question for @sonjaleix: the presence of a right arrow by the pagination suggests clicking to advance is possible. This would be quite complex to build in, since Wizards often have additional logic attached to their next buttons, so simply advancing to the next route would be bad.

2. Navigate to `/wp-admin/admin.php?page=newspack-setup-wizard`
3. Step through the Set Up Wizard.
4. Observe the pagination advancing as you go.
5. Click the BACK button and verify you return to the previous route (or to the dashboard when on page 1.)

### Other information:

### DIFF
--- a/assets/components/src/index.js
+++ b/assets/components/src/index.js
@@ -16,3 +16,4 @@ export { default as Task } from './task';
 export { default as TextControl } from './text-control';
 export { default as withWizard } from './with-wizard';
 export { default as withWizardScreen } from './with-wizard-screen';
+export { default as WizardPagination } from './wizard-pagination';
--- a/assets/components/src/wizard-pagination/index.js
+++ b/assets/components/src/wizard-pagination/index.js
@@ -1,0 +1,47 @@
+/**
+ * Wizard pagination
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { __ } from '@wordpress/i18n';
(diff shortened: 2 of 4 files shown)


jeffersonrabb:
> My only small concern, especially for the onboarding wizard, is that the pagination might collide a little with the NP logo on smaller screens. So we might want to consider pushing the screen content down a little if we have pagination on a screen. Also something we could table for now and re-explore later on.

I'm worried about this too. One solution might be to simply hide the pagination UI at smaller screen widths - it's not critical that this be seen. I'm going to 🚢 and we can discuss whether this makes sense after.

PR #164, opened 2019-07-03
Loading UI and use loading UI in Subscriptions wizards

### LINKED ISSUE #76: Pattern: Loading
Originally mentioned at https://github.com/Automattic/newspack-plugin/pull/73#discussion_r287789671.

We need a design/code pattern for blocking the screen while an API call is happening. This could be a loading spinner (component?), or something else. When the screen starts loading the UI blocks and communicates that loading is in progress, and when the screen is finished loading the UI unblocks.

claudiulodro:
### All Submissions:

Closes #76 .

This PR adds functionality for handling loading to the `withWizard` component. While loading the screen is overlayed with an element that prevents users from clicking on anything and a loading UI is displayed. I've implemented this UI based on the [Material design guidelines for indeterminate progress bars](https://material.io/design/components/progress-indicators.html#) and the Muriel designs for progress bars. They're currently not available as a stand-alone component, but if that's desired let me know.

To make usage simple, I've defined a new `apiFetch` method on `withWizard` that automatically handles loading UI. Usage is the same as `apiFetch`, you just load up `apiFetch` from `withWizard` instead of `@wordpress/api-fetch`. It can handle multiple concurrent API calls, and will stop loading once they are all resolved. @jeffersonrabb please let me know if you see any issues in the `apiFetch` wrapper method in `withWizard` or with this approach.

I've updated the two Subscriptions wizards to use this api/loading style for demonstration purposes. After this is merged I'll update the other wizards in a separate PR.

### Screenshot

localhost_wp-admin_admin php_page=newspack-subscriptions-onboarding-wizard (2)

1. `npm run clean; npm run build:webpack`
2. Use Subscriptions Onboarding and Subscriptions wizards.
3. Observe loading UI. Test that buttons/etc. are blocked from being clicked on while loading is happening. Other than that, things should work as normal.

### Question

localhost_wp-admin_admin php_page=newspack-subscriptions-wizard (3)

@sonjaleix Does this look OK? I wasn't able to come up with a really nice-looking way of doing the loading on screens that don't have the card background. Do you have a better suggestion?

### Other information:

### DIFF
--- a/assets/components/src/with-wizard/index.js
+++ b/assets/components/src/with-wizard/index.js
@@ -6,7 +6,9 @@
  * WordPress dependencies
  */
 import { Component, createRef, Fragment } from '@wordpress/element';
+import { Spinner } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
 
 /**
  * Internal dependencies.
@@ -28,6 +30,7 @@ export default function withWizard( WrappedComponent, requiredPlugins ) {
 			this.state = {
 				complete: null,
 				error: null,
+				loading: 0,
 			};
 			this.wrappedComponentRef = createRef();
 		}
@@ -141,6 +144,43 @@ export default function withWizard( WrappedComponent, requiredPlugins ) {
 		};
 
 		/**
+		 * Begin loading.
+		 */
+		startLoading = () => {
+			this.setState( state => ( {
+				loading: state.loading + 1
+			} ), () => { console.log( this.state.loading + ' items loading' ); } );
+		}
+
+		/**
+		 * End loading.
+		 */
+		doneLoading = () => {
+			this.setState( state => ( {
+				loading: state.loading - 1
+			} ), () => { console.log( this.state.loading + ' items loading' ); } );
+			
+		}
+
+		/**
+		 * Replacement for core apiFetch that automatically manages wizard loading UI.
+		 */
+		apiFetch = args => {
+			this.startLoading();
+			return new Promise( ( resolve, reject ) => {
+				apiFetch( args )
+					.then( response => {
+						this.doneLoading();
+						resolve( response );
+					} )
+					.catch( error => {
+						this.doneLoading();
+						reject( error );
+					} );
+			} );
+		}
+
+		/**
 		 * Render a Route that checks for plugin installation requirements, and redirects to '/' when all are done.
 		 *
 		 * @return void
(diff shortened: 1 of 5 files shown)


assets/components/src/with-wizard/index.js
```diff
@@ -140,6 +143,43 @@ export default function withWizard( WrappedComponent, requiredPlugins ) {
...
+			this.setState( state => ( {
+				loading: state.loading - 1
+			} ) );
+			
+		}
+
+		/**
+		 * Replacement for core apiFetch that automatically manages wizard loading UI.
+		 */
+		apiFetch = args => {
```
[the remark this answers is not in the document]
claudiulodro: Good idea. I've updated to `wizardApiFetch` in https://github.com/Automattic/newspack-plugin/pull/164/commits/a1f6e777b71479409c6cbcf78f5d4a71bcd0f489.

claudiulodro:
> In the case of the Subscription Onboarding Wizard the height of the Wizard jumps after the fields are retrieved since the SelectControls are hidden if they have no options. We can fix this here (and in any similar situation) by changing options attributes to something like this: options={ countrystateFields.length ? countrystateFields : [ {} ]. This file (locationSetup.js) isn't part of this PR so I wasn't sure how to leave it as an inline code suggestion.

Good suggestion. I've made the update in https://github.com/Automattic/newspack-plugin/pull/164/commits/872634b38c335d7133f1fac41d1856b7a18a8368.

> For Subscriptions Wizard there are a couple of problems. Initially the Wizard has no height so the blue animation occurs over a very short empty Wizard as shown in this screenshot. We could mitigate this by setting a min-height for Wizards, but then there will be a noticeable jump after the content loads, and also the Wizard will switch from having a white background (for loading) to no background when ready. I'm not sure there's any good solution, but curious if @claudiulodro or @sonjaleix have any ideas?

I've updated the loading on wizards with no backgrounds in https://github.com/Automattic/newspack-plugin/pull/164/commits/c75e11c6442690b767e5d9ffe587fbfa70434914 It looks much nicer 👍 

localhost_wp-admin_admin php_page=newspack-subscriptions-wizard (4)

There isn't a good solution for the "jumping" issue that will work in all situations. I think we'll have to be OK with a little jumping on certain screens. It will mostly be an issue on screens that render a list of something (subscriptions, plugins, etc.), since they go from 0 elements rendered to >=0 elements rendered.

> Finally, it would be great to include the PluginInstaller in the loading sequence. I tinkered with ways to do this where loading begins at a value of "1", and gets decremented after PluginInstaller is done. This works fine, but introduces the same visual problems described in the second one — a noticeable jump when loading completes and a switch from white background to no background. Any ideas?

I took a shot at this in https://github.com/Automattic/newspack-plugin/pull/164/commits/29d8d0dfb2606ce8ebaf140e24ebf53b164f499f. It'll show the loading bar while loading the plugin info. I think it looks natural now that the white BG issue has been solved (see above screenshot).

assets/components/src/with-wizard/style.scss
```diff
@@ -1,4 +1,64 @@
...
+			top: 0;
+			width: 10%;
+			height: 4px;
+			background-color: $primary-color;
+			animation: loading 2s linear infinite;
+		}
+
+		&.muriel-wizardScreen__no-background {
+			&:after {
+				background: #f1f1f1; /* WP admin bg color */
```
jeffersonrabb: How about transparent to be safe?

```suggestion
				background: transparent;
```

claudiulodro: I want it to "fade out" the stuff behind the BG, so it needs a real BG.

PR #165, opened 2019-07-03
Configure Plugins After Install

jeffersonrabb:
### All Submissions:

This PR switches the endpoint `PluginInstaller` uses from `/newspack/v1/plugins/{plugin-slug}/activate` to `/newspack/v1/plugins/{plugin-slug}/configure`. The configure endpoint will install, activate, and finally run the plugin's Configuration Manager's `configure()` method. At this point, the only actual change here is that AMP will be initially set to `Standard` mode on installation.

1. If AMP is installed, go to AMP's settings page and set Mode to "Reader" or "Transitional". 
2. From the plugins page, uninstall AMP.
3. Go to Newspack->Set Up and run through the wizard. 
4. Return to the AMP settings page and verify that mode is now "Standard."

### Other information:

### DIFF
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -80,7 +80,7 @@ class PluginInstaller extends Component {
 	installPlugin = slug => {
 		this.setInstallationStatus( slug, PLUGIN_STATE_INSTALLING );
 		const params = {
-			path: `/newspack/v1/plugins/${ slug }/activate/`,
+			path: `/newspack/v1/plugins/${ slug }/configure/`,
 			method: 'post',
 		};
 		return apiFetch( params )


philipjohn:
I tested this on my live test site, and it didn't work as expected. This is what I ended up doing:

1. Switched to Reader mode for AMP
2. Disabled all plugins except Newspack
3. Went to Newspack > Setup
4. Ran through the wizard (note that JP/SK both showed as configured as soon as they were activated, which makes sense)
5. Clicked the Finish button
6. Observed that AMP was active
7. Visited AMP settings page and found it still set to Reader mode

philipjohn:
@jeffersonrabb I've added you as an admin on my test site so you can see for yourself

jeffersonrabb:
> I tested this on my live test site, and it didn't work as expected. This is what I ended up doing:

Hmm, I just tested twice more and got the expected result both times. Just to confirm, did you run the build script first? `npm run build:webpack`? The change is in a Wizard's JS,  so if this wasn't run the change wouldn't be picked up.

Issue #171, opened 2019-07-08
Grid components aren't aligned properly 


jffng:
**Describe the bug**
Styles targeting the `#wpcontent` containing element push the newspack container off center by 10px and 20px, depending on the size of the viewport. 

This is related to #94 and #130. 

**To Reproduce**
1. Download the Muriel grid chrome extension (https://chrome.google.com/webstore/detail/muriel-grid/heabkjiofmaphebjodoflpglpkdojpmg?hl=en)
2. Navigate to any page with a `muriel-card` component e.g.  `Newspack -> Components demo`
3. Toggle the grid using the extension 

**Screenshots**
![newspack-muriel-grid-before](https://user-images.githubusercontent.com/5375500/60845543-a702c680-a1aa-11e9-9ebe-abf4d418a84e.png)

**Expected behavior**
The card should align to the Muriel grid.

jeffersonrabb:
I believe this is fully addressed in https://github.com/Automattic/newspack-plugin/pull/172

PR #172, opened 2019-07-09
Use CSS grid to implement the Muriel grid

jffng:
### All Submissions:

Introduces an approach to address #171 using CSS grid to size and position components to the Muriel grid.

Before:
image

After: 
image

2. Navigate to `Newspack -> Components demo`
3. Verify that the components align to the grid at all viewport sizes

Note that these changes attempt to scope this to the `Components demo` page only. Thus having to add the `muriel-grid-item` classname to each card component in the components demo.

### Other information:

### DIFF
--- a/assets/components/src/card/style.scss
+++ b/assets/components/src/card/style.scss
@@ -26,7 +26,7 @@
 		padding: 0;
 		max-width: 682px;
 	}
-	@media screen and ( max-width: 1024px ) {
+	@media screen and ( max-width: 1040px ) {
 		max-width: calc( 642px - 58px );
 		padding-left: 29px;
 		padding-right: 29px;
@@ -54,7 +54,7 @@
 		grid-column: 3 / span 8;
 		margin: 0;
 		margin-bottom: 10px;
-		@media screen and ( max-width: 1024px ) {
+		@media screen and ( max-width: 1040px ) {
 			max-width: none;
 			grid-column: 2 / span 6;
 			&.muriel-card__no-background {
--- a/assets/components/src/card/style.scss
+++ b/assets/components/src/card/style.scss
@@ -49,4 +49,25 @@
 			width: 100%;
 		}
 	}
+	&.muriel-grid-item {
+		max-width: none;
+		grid-column: 3 / span 8;
+		margin: 0;
+		margin-bottom: 10px;
+		@media screen and ( max-width: 1024px ) {
+			max-width: none;
+			grid-column: 2 / span 6;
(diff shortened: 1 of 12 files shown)


jffng:
Ah yeah. The grid actually uses a `1040px` breakpoint, made that fix and pushed. 

> Did you scope to the Components page purely for demo purposes? After this PR ships, would you suggest we apply the muriel-grid-item styling to all Card components?

Yes and yes. I wanted to make sure we were comfortable with the approach -- it requires a grid container class to be applied to wherever the card components are used.

I went with a mixin so we could just apply it to existing container classes as necessary. But if it turns out that cards are used everywhere, maybe we should use a utility class instead to avoid a bunch of duplicate css. I could work on this next and see how it goes.

claudiulodro:
It might be easiest to just add the `muriel-grid-item` class to [the Card component.](https://github.com/Automattic/newspack-plugin/blob/master/assets/components/src/card/index.js#L26)

jffng:
@claudiulodro totally. It would just require a containing element to have grid styles also.

Just pushed a change that implements this using a grid container component. Let me know if that's an acceptable approach.

sonjaleix:
@jffng This is looking great now. I'm getting one alignment issue for some of the action cards on mobile: 
Screenshot 2019-07-10 12 58 03

Otherwise all components are perfectly aligning to the grid in all viewports.

jffng:
@sonjaleix thanks for catching that, pushed a fix.

sonjaleix:
Thanks @jffng! The issue with the action card on mobile looks great now!

I'm not sure if you've applied the changes only to the components screen. If so, we might want to merge or see how they apply to the other screens and wizards too. 
We still need to test how this looks for the Setup screens and individual wizard.

One other specific thing we will probably have to adjust, is the odd grid on the Newspack dashboard: /wp-admin/admin.php?page=newspack

See grid dashboard (homescreen) blocks and how they align in the grid in design: 
Screenshot 2019-07-10 15 33 41
Screenshot 2019-07-10 15 33 30

claudiulodro:
I've updated the wizards and checklists to use the grid in https://github.com/Automattic/newspack-plugin/pull/172/commits/4e58ddfc960b982d322b99d1cd134beb0aa5d8b4. They should all line up nicely now. 

The only exception is the dashboard which is going to need some CSS fine-tuning to play nicely with the grid:
localhost_wp-admin_admin php_page=newspack (2)

I can handle it after this is merged.

PR #190, opened 2019-07-18
Unified Ad Wizard

jeffersonrabb:
### All Submissions:

This PR is a complete refactor of the Ad wizards, based on a variety of new directions in UI design and code organization.

<details>
1. Main screen. Lists all currently supported Ad Services. Enabling AdSense or Ad Manager simply sets an option. Enabling WordAds activates the Jetpack module. If Jetpack plan is too low, a notification will be shown.

Screen Shot 2019-07-24 at 9 33 35 AM

2. Global Settings. This is where predefined slots can be filled. Note that there is no implementation of this in the theme yet, so for now these settings will not affect the published site in any way. For each placement, once enable there will be a dropdown allowing selection of the enabled services. If Google Ad Manager is selected a second Select appears containing all defined Ad Units. The WordAds configure link also goes to this screen, since this is the only WordAds configuration necessary. 

Screen Shot 2019-07-24 at 9 34 29 AM

3. AdSense setup. This is a `Handoff` to Site Kit, pretty much unchanged from the AdSense wizard currently on `master`..

Screen Shot 2019-07-24 at 9 34 43 AM

4. Ad Manager Ad Units. This is where ad units are listed, edited, deleted, and created. More or less unchanged from the current Google Ad Manager wizard on `master`.

Screen Shot 2019-07-24 at 9 34 51 AM

5. Ad Unit Detail. Create or edit Google Ad Manager Ad Units here. This UI is largely unchanged from the Ad Manager wizard on `master.` There is now support for AMP/non-AMP codes. Also, the name of the ad service is saved with each unit (`google_ad_manager`), which means this can easily be extended to support other similar ad services.

Screen Shot 2019-07-24 at 9 34 55 AM

6. Global code. Google Ad Manager header code. 

Screen Shot 2019-07-24 at 9 34 58 AM

</details>

2. Newspack Ads will need to be cloned into `wp-content/themes`, checked out to `update/extend-model`, and build (`npm ci && npm run build:webpack`)
3. The screens are described in detail above, so it's probably best to play around with them and see what comes up.
4. Use the sample ad codes at p2-pamTN9-kv to test

### Other information:

This includes two open PRs which will need to ship before this can: 
- https://github.com/Automattic/newspack-plugin/pull/203
- https://github.com/Automattic/newspack-plugin/pull/202

As well as https://github.com/Automattic/newspack-ads/pull/6


claudiulodro:
> When enabling AdWords, clicking "Configure" shouldn't go to the intermitten screen, it should bring the user directly to Site Kit AdSense configuration

The tricky part here is that we need to do the required plugin check before going to Site Kit, and we also want to set the "Back to Newspack" banner. For the short-term, I'm OK with the extra click (it would def. be ideal to simplify that in the future, though).

jeffersonrabb:
> When enabling AdWords, clicking "Configure" shouldn't go to the intermitten screen, it should bring the user directly to Site Kit AdSense configuration

Per @claudiulodro's [note](https://github.com/Automattic/newspack-plugin/pull/190#issuecomment-515546381) we'll need to extend `ActionCard` to support `Handoff`-style behavior. Very doable, but I'd prefer to address in a separate PR. So I'd suggest we move forward on this one without that and add afterwards.

jeffersonrabb:
> 4\. The "Add an ad unit" screen should not have the tabs. This is one level deeper in the hierarchy and we don't want to distract the user from the task at hand with the tabs.

Fixed in https://github.com/Automattic/newspack-plugin/pull/190/commits/250a5953ad5a08c20879ba2460313ba19e186268

jeffersonrabb:
> 5\. The "Ad an ad unit" screen should have at least the title field required. A user can currently save an ad unit without a title.

Done in https://github.com/Automattic/newspack-plugin/pull/190#pullrequestreview-266243086. You'l now get an error notification if you attempt to Save an Ad Unit with a blank title.

jeffersonrabb:
Evened out a bunch of the top spacing in https://github.com/Automattic/newspack-plugin/pull/190/commits/53022a88549076a0fb8356875d48d0ea714861ce

PR #191, opened 2019-07-19
Fixes Select Component initial value is always the first option

### LINKED ISSUE #83: Bug: Initial select value is always the first option
When using the `SelectControl` component, the initial value for the `select` element upon page load uses the value of the first child `option` element, rather than having no - or an empty - value. The _appearance_ of the select is that it has a default option, but this is not the case and so the normal behaviour of the browser is to default to choosing the first `option` for the value.

**To Reproduce**
1. Open the Components Demo page in WP Admin
2. Using the console, observe the values for the 3 three drop downs as follows:
3. `document.getElementById("inspector-select-control-0").value` shows "2nd"
4. `document.getElementById("inspector-select-control-1").value` shows "1st"
5. `document.getElementById("inspector-select-control-0").value` shows "1st"
6. Add the following code after the second `SelectControl` component in `componentsDemo/index.js`:

```
{ selectValue2 == '1st' && (
	<p>1st is selected!</p>
) }
{ selectValue2 == '2nd' && (
	<p>2nd is selected!</p>
) }
```

7. Reload the page and observe there is no text after the second drop down
8. From the second drop down select "First"
9. Observe that there is still no text below the drop down
10. Select "Second" from the drop down
11. Observe that the text "2nd is selected!" appears below the drop down
12. Select "First" from the drop down
13. Observe that the text "1st is selected!" appears below the drop down.

When we selected "First" in step 8 we should have seen text appear. We didn't because the value was already "1st" and so state did not change even though we selected that first item. Only by changing state in 10 and back again in 12 are we able to see the text we were expecting.

iuravic:
### All Submissions:

As detected by issue #83 , _"... when using the SelectControl component, the initial value for the select element upon page load uses the value of the first child option element, rather than having no - or an empty - value"_.

Here this is fixed by adding a first `disabled` option with the label "- Select -" and empty `value`. 

However, since the `SelectControl as BaseComponent` from`'@wordpress/components` doesn't support the `disabled` attribute for the `<option>` element (see [in codebase](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/select-control/index.js#L51)), now switched to using a manual `<select>` HTML element.

[This PR](https://github.com/WordPress/gutenberg/pull/15976) should eventually update the BaseComponent, but since a PR process could be relatively lengthy, and our publishers are experiencing the bug right now, currently switched the component to use a pure `<select>` element with a `<label>`.

Also as a minor change, in the Components Demo wizard, updated more meaningful labels and fixed the wrong use of onChange to handle event as argument, rather than value.

When now clicking on the Select component (specifically relevant for a Select with no previous selections), a first disabled option "- Select -" is present, enabling correct onClick of the first option:

image

Closes #83 .

1. Pull and build `nvm use 10.10.0 && npm ci && npm run clean && npm run build:webpack -- --watch`.
2. Visit the Components demo page at `/wp-admin/admin.php?page=newspack-components-demo`.
3. Verify that the three demoed Select Components are initially visually sound.
4. Check that the second Select component with label "Label for Select with no selection" now accepts selection of the first option with a single click, and that on clicking away the selection persists/sticks.
5. Test click the remaining two Select components.

### Other information:


iuravic:
Thanks for all the comments! Great thinking! 👍I agree, that's a smarter solution, since it will make switching to the updated `@wordpress/components` a breeze.

To handle both your comments, I made the disabled "0th option" just a regular option, provided by the developer or not. This seems simplest after all since that's the same way [the PR](https://github.com/WordPress/gutenberg/pull/15976/files#diff-c9a371eb4ddfb61b5746698586b13ee8R37) is handling it; and also, we don't have to use new props.

It would be great if you could please give it another test run.

iuravic:
Thank you for noticing all this!
Now updated.

iuravic:
@jeffersonrabb thank you so much for your patience! Reformatted the stylesheet as well, and it's up to the formatting standards now. Merging!

PR #194, opened 2019-07-20
Tabbed Navigation

jeffersonrabb:
### All Submissions:

This PR creates a `<TabbedNavigation>` component and implements it in `withWizardScreen`. This will be a common navigation style used in most (if not all) wizards.

Screen Shot 2019-07-20 at 1 15 56 PM

Tabbed Navigation isn't actually used in any Wizards in this PR, so the best way to test is to add the following prop to both Wizard Screens in the Subscription Onboarding wizard and give it a try. You should see results like the screenshot above:

```
tabbedNavigation={ [
  {
    label: __( 'Step 1' ),
    path: '/',
    exact: true,
  },
  {
    label: __( 'Step 2' ),
    path: '/stripe',
  },
] }
```

### Other information:

### DIFF
--- a/assets/components/src/index.js
+++ b/assets/components/src/index.js
@@ -13,6 +13,7 @@ export { default as NewspackLogo } from './newspack-logo';
 export { default as PluginInstaller } from './plugin-installer';
 export { default as ProgressBar } from './progress-bar';
 export { default as SelectControl } from './select-control';
+export { default as TabbedNavigation } from './tabbed-navigation';
 export { default as Task } from './task';
 export { default as TextControl } from './text-control';
 export { default as withWizard } from './with-wizard';
--- a/assets/components/src/tabbed-navigation/index.js
+++ b/assets/components/src/tabbed-navigation/index.js
@@ -1,0 +1,52 @@
+/**
+ * Tabbed Navigation
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { Component } from '@wordpress/element';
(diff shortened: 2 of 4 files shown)


jeffersonrabb:
> Just curious, have you explored any ways we could automate the tabbed navigation so we wouldn't have to "manually" manage the tabs? For example, `withWizard` could automatically get the tabs info from child `Route` elements or something? If there isn't a good automated solution, feel free to merge.

Yeah, this bugged me too and I played around with this at first. The problems I ran into had to do with UI flexibility, not technical limitations per se. The main conceptual problem is that tabbed navigation will generally be a hand-picked subset of the available routes, so an automated solution would limit flexibility too much. Also it's going to be touch to lock in a system for which views need which tabbed navs, so I kind of figured it was best to leave it ultra-flexible. If you have a clearer vision I'd love to discuss. I won't merge just yet pending your response.

claudiulodro:
I see. That makes sense. Thanks for clarifying. I agree with your opinion: let's go with the more flexible route. It's not a big deal to manually manage the tabs from a central variable or something:

```
tabbedNavigation={ [
	{
		label: __( 'Step 1' ),
		path: '/',
		exact: true,
	},
        ...
] }

<Wizard1 tabbedNavigation={ tabbedNavigation } />
<Wizard2 tabbedNavigation={ tabbedNavigation } />
...
```

Go ahead and merge. 👍

PR #195, opened 2019-07-20
Toggle Group

jeffersonrabb:
### All Submissions:

Creates `<ToggleGroup>` component. This component is a title and description with a toggle to the left. The children of the element are hidden unless the toggle is checked. The use case for this component is the Advertising Wizard view where ad placements are switched on and defined:

Advertising 05

Screen Shot 2019-07-20 at 5 18 49 PM

1. `npm ci && npm run clean &&  npm run build:webpack`
2. Navigate to the Components Demo
3. Verify the appearance of the toggle group at the top. Toggle the switch and verify that additional content is revealed.

### Other information:

### DIFF
--- a/assets/components/src/index.js
+++ b/assets/components/src/index.js
@@ -16,6 +16,7 @@ export { default as SelectControl } from './select-control';
 export { default as TabbedNavigation } from './tabbed-navigation';
 export { default as Task } from './task';
 export { default as TextControl } from './text-control';
+export { default as ToggleGroup } from './toggle-group';
 export { default as withWizard } from './with-wizard';
 export { default as withWizardScreen } from './with-wizard-screen';
 export { default as WizardPagination } from './wizard-pagination';
--- a/assets/components/src/toggle-group/index.js
+++ b/assets/components/src/toggle-group/index.js
@@ -1,0 +1,45 @@
+/**
+ * Toggle group
+ */
+
+/**
+ * WordPress dependencies.
+ */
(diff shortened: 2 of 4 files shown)


claudiulodro:
The code for this looks solid. My main concern is [according to the usage guidelines](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/checkbox-control#when-to-use-checkboxes), checkboxes are the way to go in these sorts of situations, not toggles. Thoughts, @sonjaleix?

sonjaleix:
> The code for this looks solid. My main concern is according to the usage guidelines, checkboxes are the way to go in these sorts of situations, not toggles.
That's a fair question. Toggles (switches) and Checkboxes are very similar in their function. Material Design describes them as follow:

> Switches toggle the state of a single setting on or off. They are the preferred way to adjust settings on mobile.

> Checkboxes allow the user to select one or more items from a set. Checkboxes can be used to turn an option on or off.

In our particular case we are asking the user to "turn on/off" one or multiple individual functionality (e.g. using a specific advertising provider). Even though all of the providers can be seen as a "set", they operate independently from each other. I therefore think Toggles (Material calls them Switches) are the better UI for this.

Reference: 
https://material.io/design/components/selection-controls.html#switches
https://material.io/design/components/selection-controls.html#checkboxes

claudiulodro:
Thanks for looking into it. It looks like toggles should work well for our use-case. 👍

PR #203, opened 2019-07-24
Action Card Badges

jeffersonrabb:
### All Submissions:

Action Cards will often be used to list plugins and products, and in some cases we'll need to indicate a variety of special statuses for these e.g.: PREMIUM, REQUIRES UPGRADE, etc. This PR introduces a `badge` prop on the `ActionCard` component, which adds such a badge.

Screen Shot 2019-07-23 at 10 25 12 PM

Closes # .

1. Navigate to Components Demo.
2. Observe `ActionCard` example with badge.

### Other information:

### DIFF
--- a/assets/components/src/action-card/style.scss
+++ b/assets/components/src/action-card/style.scss
@@ -124,9 +124,9 @@
 	/* Badge */
 
 	.newspack-action-card-badge {
-		border: 1px solid #606A73;
+		border: 1px solid $muriel-gray-500;
 		border-radius: 3px;
-		color: #606A73;
+		color: $muriel-gray-500;
 		font-size: 9px;
 		font-weight: bold;
 		margin: 0 0 0 2em;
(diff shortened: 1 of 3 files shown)


assets/components/src/action-card/style.scss
```diff
@@ -120,4 +120,18 @@
 	&:hover button.newspack-action-card__secondary_button {
 		visibility: visible;
 	}
+
+	/* Badge */
+
+	.newspack-action-card-badge {
+		border: 1px solid #606A73;
+		border-radius: 3px;
+		color: #606A73;
```
claudiulodro: Not sure if this is something we have access to like the Muriel colors, but if so this color is available through Gutenberg colors: https://github.com/Automattic/jetpack/blob/22297f2ecbc43ebc425dadb1e83f025ca4daff7e/extensions/shared/styles/gutenberg-colors.scss#L16

jeffersonrabb: After discussing with @sonjaleix changed to Muriel Gray 500 in https://github.com/Automattic/newspack-plugin/pull/203/commits/443e11df7d863fcbfdf74c54b3b749a535c974cc

PR #216, opened 2019-07-27
Plugin Installer UI Improvements

### LINKED ISSUE #187: Adjust Required plugin wizard
Based on some design feedback we received from the Muriel team, it's a good idea to simplify the required plugins wizard UI. _Related to component #55_ 

**Proposed changes:** 
- Remove installation action links for each individual plugin (since all plugins will need to be installed in order to use the feature)
- More visually indicate already installed / activated plugins with a checkbox icon
- Keep the inline "Setting up..." indicator with spinner
- Only have one primary CTA on this screen to install all plugins
- We'll keep the error / warning handling as already implemented (see screenshot below)

**Updated UI (design [in Figma](https://www.figma.com/file/CFxMAXFD0DxNFcVkDkQw4gLb/Newspack-Plugin-v1?node-id=455%3A40)):**
Screenshot 2019-07-16 13 05 36

**Various card states incl. error handling:**
Screenshot 2019-07-16 13 10 34

jeffersonrabb:
### All Submissions:

This PR contains a series of UI improvements to the `PluginInstaller` component, described in https://github.com/Automattic/newspack-plugin/issues/187.

- Individual plugin install links have been removed. Now, the only action is the `Install` button at the bottom, which installs all of them.
- Uninstall UI has been removed. (@sonjaleix this was not requested in the issue but I assumed that the thinking would be identical as for removing individual plugin installation UI)
- Blue checkbox UI indicates success, replacing the words "In Use."

Screen Shot 2019-07-27 at 12 13 37 PM

Closes #187.

1. `npm run build:webpack`
2. Navigate to Components Demo page and try out the example Plugin Installer.
3. Go to WordPress plugins page and deactivate or uninstall some of the plugins in the example, and try it again.
4. Bonus: Uninstall other required plugins and try various Wizards to verify they work just like before.

### Other information:

### DIFF
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -215,7 +215,7 @@ class PluginInstaller extends Component {
 							className="is-centered"
 							onClick={ this.installAllPlugins }
 						>
-							{ __( 'Use All' ) }
+							{ __( 'Install' ) }
 						</Button>
 					) }
 			</div>
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -9,6 +9,7 @@ import apiFetch from '@wordpress/api-fetch';
 import { Component } from '@wordpress/element';
 import { Spinner } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
+import { Dashicon } from '@wordpress/components';
 
 /**
  * Internal dependencies.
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
> * Padding on error messages is inconsistent. This may have been a long-standing issue I only just noticed. :) I think a little bit of padding at the top of the error is all it needs.

Good catch! This should fix: https://github.com/Automattic/newspack-plugin/pull/216/commits/223f525f2cacf66a5072aa68ac86556fdb6613ee

jeffersonrabb:
> We previously avoided using 'Install' as wording on the primary button because the button also activates installed plugins, so 'Install' isn't quite 100% correct. 'Install and activate' might be a clearer wording?

I tend to think we should leave it for now and possibly revisit later? We had been using "Use" to get around this problem but personally that started to grate on me, and I kind of prefer the straightforwardness of "Install," which is what you typically see in UI like this. @sonjaleix what do you think?

jeffersonrabb:
Thank you both!

PR #217, opened 2019-07-27
Button Component Styling

### LINKED ISSUE #177: Adjust Muriel button style
We need to adjust the all button styles based on feedback from the Muriel team in regards to the width of most of our buttons. 

All the button styles we've implemented (primary, secondary, tertiary) are too wide. Buttons should not have a fixed or max width, the width of a button should rather be determined by its content (the text within) and the padding to the left and right of it.

_Sidenote: We're working on future forward components, and the Muriel components have not all been finalized. So the button and other components will likely have to be refined in the future. In the meantime Material Design is our source of truth._ 

Here is what already exists, though it doesn't fully match our use case for buttons, nor the final Muriel button design: http://muriel-docs.bradgriffith.me/components/button

**How to update the buttons:**
Based on Material Design (our ultimate truth), let's adjust our button styles as follows: 
Padding: 16px on top/bottom, 32px left/right
Font-size: 16px
Depending on what we're using overall, we might want to use "em" instead of pixels for better scaling.

Let's leave the rest of the button style as is.

_Current implementation:_
Screenshot 2019-07-26 12 55 16

_Updated style:_
Screenshot 2019-07-26 12 57 04

jeffersonrabb:
### All Submissions:

This PR makes a few adjustments to the styling of the `Button` component based on feedback from the Muriel team, described [here](https://github.com/Automattic/newspack-plugin/issues/177). 

Screen Shot 2019-07-27 at 3 23 15 PM

_Note: The approach to centering `Button`s has been refactored. Originally centering was achieved by applying the `is-centered` class, which set the button's display to `block` and left/right margins to auto. The PR removes widths for the component, so this became problematic. Originally I tried setting the `Button` width to `fit-content`, but later removed this approach because of [insufficient browser support](https://caniuse.com/#feat=intrinsic-width). In the new approach, `Button`s with the `is-centered` class will be wrapped in a `<div>` with `text-align: center;`, which feels like a better approach. I'm uneasy about changing React's rendering based on a class, so in a future PR I plan to switch to using an `isCentered` prop, and apply this to all `Button`s that currently have the `is-centered` class. I opted not to do this here because it will dramatically increase the Files Changed, making this PR more difficult to review._ 

Closes #177.

1. `npm run build:webpack`
2. Navigate to Components Demo, observe the new styling of all `Button`s. 
3. Click through the existing Wizards and make sure everything looks solid.

### Other information:

### DIFF
--- a/assets/components/src/button/button.test.js
+++ b/assets/components/src/button/button.test.js
@@ -30,8 +30,7 @@ describe( 'Button', () => {
 	describe( 'rendering centered primary', () => {
 		it( 'should render a Button element with a class of is-primary and class of is-centered', () => {
 			const button = shallow( <Button isPrimary className="is-centered">Continue</Button> );
-			expect( button.render().hasClass( 'is-primary' ) ).toBe( true );
-			expect( button.render().hasClass( 'is-centered' ) ).toBe( true );
+			expect( button.render().hasClass( 'muriel-button-is-centered-wrapper' ) ).toBe( true );
 		} );
 	} );
 } );
--- a/assets/components/src/button/index.js
+++ b/assets/components/src/button/index.js
@@ -16,15 +16,18 @@ import murielClassnames from '../../../shared/js/muriel-classnames';
 import './style.scss';
 
 class Button extends Component {
-
 	/**
 	 * Render.
 	 */
 	render( props ) {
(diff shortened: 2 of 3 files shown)


jeffersonrabb:
Thank you!

PR #218, opened 2019-07-29
Setup Wizard UI Improvements

### LINKED ISSUE #200: Adjust Newspack Setup (onboarding) screens as per feedback
We received some good design feedback for our onboarding flow. Here are the proposed changes: 

**Setup landing screen:**
- Add an explicit note that the plugin setup flow will install core plugins in the background. We want to be transparent with our users and communicate what is about to happen. 
Screenshot 2019-07-23 13 16 56

**About Publication and More about Newsroom screens:**
- We're removing the "plugin installation progress bar" at the bottom of the screen for the following reasons: 
-- It draws too much attention, while not being utilitarian
-- It was confused for a stepper
-- It's unclear to the users on first sight if it's ok to progress to the next step
Screenshot 2019-07-23 13 18 46
Screenshot 2019-07-23 13 19 02

**Adding a conditional plugin installation progress screen:*
The user would only land on this screen in these cases: 
- The plugin installation process has not been completed, in that case we're showing the progress bar. 
- There was an error during the plugin installation process. In that case we display the error message (ideally based on the plugin). 

I suggest to implement a design similar to the Required Plugins screen. This will make it easy to see progress and display error messages / warnings as needed. The continue button will be inactive until the process is complete.
Screenshot 2019-07-23 13 27 33

**Jetpack and Site Kit configuration handoff screens:**
The feedback has revealed that the current implementation (below for reference) is confusing as it offers options for both plugins to be configured. The proposed update is to split this screen into two separate screens / steps. If one of the plugins is already setup and configured the screen won't show. 
Screenshot 2019-07-23 15 58 13
Screenshot 2019-07-23 15 58 32

_Current implementation for reference:_
Screenshot 2019-07-23 16 00 46

**Note:** With all of these conditional screens comes the challenge of adjusting the stepper number. I don't have a good solution for this. Maybe someone on the technical side has one.

Updated designs can be found [in Figma here](https://www.figma.com/file/CFxMAXFD0DxNFcVkDkQw4gLb/Newspack-Plugin-v1?node-id=683%3A0).

jeffersonrabb:
### All Submissions:

This branch addresses all of the feedback listed in https://github.com/Automattic/newspack-plugin/issues/200. The issue is quite detailed so here I'll only list the things that were done a little differently than described:

- On screen 1, the "Not Right Now" button links to the root of the WordPress dashboard. Since this is the very first thing a Newspack customer sees, this seemed like the right place to send them if they want out.
- If all plugins have installed successfully by the time screen 4 is reached, the wizard advances directly to screen 5. Screen 4 (plugin installer details) is only shown if there are incomplete or failed installations.
- The stepper shows all 6 screens, including the optional plugins list screen. I toyed with dynamically changing the total number of screens but this seemed like a bad UI pattern that will confuse the user.
- On the Jetpack and Site Kit configuration screens, the button will read "Configure {Plugin}" or "Continue" if set up is complete. I played with skipping these screens if the plugin is already configured, but found that to also be confusing UI. After configuring a plugin, the handoff link will return the user to the same setup page, although they will see checkmark UI indicating success and a Continue button.

<details>

</details>

Blocked by these PRs:

https://github.com/Automattic/newspack-plugin/pull/216
https://github.com/Automattic/newspack-plugin/pull/217

Closes #200 .

Setting up Jetpack and Site Kit require the test site have a public URL, so testing is ideally done on a real site or local site using Ngrok.

1. `npm run build:webpack`
2. Manually deactivate and delete the Setup plugins: Yoast, AMP, Site Kit, Jetpack, PWA.
3. Navigate to Newspack->Setup and run through the flow.

### Other information:


rickybanister:
There was great debate during the design of the dotcom onboarding flow about steppers and whether they're helpful or a hinderance. In dotcom's case we have to rely on usability testing and analytics. In our case we can simply ask charters for their opinion. I'd say it's likely fine to omit the stepper to start and only broach adding dynamism to it if absolutely necessary. Setting up your website if you website is the entire crux of your business is likely to take some time and we probably don't need to worry about folks getting bored while doing so (in our specific case).

I understand @sonjaleix's opinion regarding skipping items that are already configured, but perhaps it's a good opportunity to briefly mention to customers that Newspack and Jetpack have a very symbiotic relationship, with us relying on many of Jetpack's features. We could do this in different ways—instead of presenting a Jetpack set up screen during a linear flow we could perhaps instead show a 'security' menu item on the Newspack dashboard with a 'provided by Jetpack' subheading or similar.

jeffersonrabb:
A few technical and UI notes about the way this was put together:

- I'm skeptical of changing the number of steps in a stepper in a way that could confuse the user. If we adjust the screen count dynamically, a user first sees 6 screens, then after installing plugins—5, then after Jetpack—4, and so on. The benefits we gain from hiding completed screens may be lost by presenting an ever-changing screen count, which could lead to "where'd they go?" reactions.
- As currently built, all Wizards acquire their initial state via an API call. This means there is a delay while this is retrieved. This delay means the number of screens may change immediately after state is retrieved, and that the screen they begin on may be hidden dynamically and push them to the next screen. Both lead to an undesirable FOUC-esque effect. Ultimately there is a technical solution to this where we provide initial state on page load, but this will require refactoring and doesn't seem worth pursuing until a later stage of development when we can focus exclusively on refinement. 

Because of both of these considerations, my suggestion is that we land this as-is and address dynamic screen visibility in a later phase. I can open an issue capturing this to guide future work. 

@rickybanister what do you think?

jeffersonrabb:
Note: rebased...

rickybanister:
@jeffersonrabb I agree with that. Let's get the substance of the work merged and circle back for the finesse bits. I'd much rather stub in all the functionality now and worry about things like paging, illustration, copy later as we can see more of the forrest.

PR #222, opened 2019-07-31
Donations: Fix typo in dependency reference

philipjohn:
The build can fail with the following error due to the uppercase D:

> ERROR in ./assets/wizards/donations/index.js
Module not found: Error: Can't resolve './views/DonationSettingsScreen' in '/home/philipjohn/vagrants/newspack/www/newspack/public_html/wp-content/plugins/newspack-plugin/assets/wizards/donations'
 @ ./assets/wizards/donations/index.js 23:0-68 190:31-53

This change fixes that by making sure the reference has a lowercase 'd'

### DIFF
--- a/assets/wizards/donations/index.js
+++ b/assets/wizards/donations/index.js
@@ -11,7 +11,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
This is blocking testing on Digital Ocean instances, so I'm going to land it.

PR #225, opened 2019-07-31
Donation form processing

claudiulodro:
### All Submissions:

Blocked by #199. The relevant commit for this PR is https://github.com/Automattic/newspack-plugin/commit/f1087e0dafb3350fddd9bb2a6bba4a9160564806

This PR adds processing for the Donations block (or other forms that use compatible fields). When the form is submitted, the correct product gets added to the cart at the appropriate price and the user is redirected to the Checkout page.

1. Check out https://github.com/Automattic/newspack-blocks/pull/64 if it hasn't been merged yet. Set up your donation settings and insert a Donation block somewhere.
2. On the frontend, pick how you'd like to donate, and click the Donate Now button in the block. You should be dropped off on the Checkout page. Scroll down to order details and verify the order details match what you submitted. 

Checkout page styling will be in a future PR; the order details along with a link to change your selected donation will be at the top in the future. In the meantime, if you need to change your donation, go to `{siteURL}/cart` and remove it there.

Screen Shot 2019-07-31 at 11 43 34 AM

### Other information:


includes/class-donations.php
```diff
@@ -103,6 +114,24 @@ public static function get_donation_settings() {
...
+			}
+
+			if ( 'subscription' === $child_product->get_type() ) {
+				if ( 'year' === $child_product->get_meta( '_subscription_period', true ) ) {
+					$settings['products']['year'] = $child_id;
+				} else {
+					$settings['products']['month'] = $child_id;
+				}
+			} else {
+				$settings['products']['once'] = $child_id;
```
jeffersonrabb: Is there a scenario where the grouped product is edited directly in WooCommerce (i.e. not by using Newspack's Wizard) and we end up with unexpected child products? Based on this risk, wondering if we should avoid catch-all `else` statements, both for determining the `once` product and the `month`?

includes/class-donations.php
```diff
@@ -275,4 +304,59 @@ protected static function update_donation_product( $args ) {
 			$child_product->save();
 		}
 	}
+
+	/**
+	 * Handle submission of the donation form.
+	 */
+	public static function process_donation_form() {
+		$donation_form_submitted = filter_input( INPUT_GET, 'newspack_donate', FILTER_SANITIZE_NUMBER_INT );
```
jeffersonrabb: Doesn't this create a possibility that an unauthorized form submission could be interpreted as coming from the Donate block? One the one hand, this wouldn't be a significant problem since all we're doing is populating the cart, but it still could be seen as a problem. Could we solve  this with a nonce strategy?

claudiulodro: We could add a nonce, but it wouldn't really help much. The main reason I didn't worry about it is because [WooCommerce doesn't put a nonce on its generic `add-to-cart` handler](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-form-handler.php#L726). e.g. anyone can put a link to `example.com?add-to-cart={product_id}` anywhere. There isn't any real security risk (other than slightly inconveniencing someone), and it lets people easily use the links in articles, emails, etc. (e.g. 'Click here to donate $10 now!')

includes/class-donations.php
```diff
@@ -20,6 +20,13 @@ class Donations {
 	const DONATION_UNTIERED_SUGGESTED_AMOUNT_META = 'newspack_donation_untiered_suggested_amount';
 	const DONATION_TIERED_META                    = 'newspack_donation_is_tiered';
 
+	/**
+	 * Initialize hooks/filters/etc.
+	 */
+	public static function init() {
+		add_action( 'wp_loaded', [ __CLASS__, 'process_donation_form' ] );
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/225/commits/dee3f380a7946e03c70931e2d01d76d8ba1bcc6e

includes/class-donations.php
jeffersonrabb: OK sounds good.

includes/class-donations.php
claudiulodro: Good call. I've added better checks in 1cecd64 to more accurately determine whether a product should be included.

claudiulodro:
> How will we handle login before checkout?I tested incognito and get the following when adding the item to the cart:

There's a setting you need to enable in the WC > Accounts & Privacy settings:
Screen Shot 2019-08-05 at 12 17 02 PM


That setting will be part of a suite of settings I need to automate with the WC configuration manager after everything is merged and our configuration has settled down somewhat.

claudiulodro:
> It appears this only works the first time through the flow. If you add donation to the cart, hit the back button and add a different item to the cart only the first one will be in it.

I've changed it to clear donations from the cart before adding a donation to the cart in https://github.com/Automattic/newspack-plugin/pull/225/commits/4b926c8909681fb97ceae0b88b4e31d4fdd7bc5f. This should solve the issue, and I think it's a better flow because nobody wants to have 2 donation products in their cart on purpose.

I'll use `remove_donations_from_cart`  again in a future PR when I add a "Change your selected donation" (or similar) link at the top of the checkout page.

PR #228, opened 2019-08-03
WIP: Unified Revenue Wizard

jeffersonrabb:
### All Submissions:

Work In Progress to pull together Reader Revenue Onboarding Wizard, Donations Wizard, and soon [Membership Landing Page Wizard](https://github.com/Automattic/newspack-plugin/pull/173) into a single unified Wizard.

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -424,7 +424,12 @@ class Donations {
 			'comment_status' => 'closed',
 			'ping_status'    => 'closed',
 		];
-		return wp_insert_post( $page_args );
+
+		$page_id = wp_insert_post( $page_args );
+		if ( is_numeric( $page_id ) ) {
+			self::set_donation_page( $page_id );
+		}
+		return $page_id;
 	}
 
 	/**
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -387,7 +387,7 @@ class Donations {
 	 * @return int Post ID of page.
 	 */
 	public static function create_donation_page() {
-		$revenue_model = Reader_Revenue_Onboarding_Wizard::get_revenue_model();
+		$revenue_model = 'donations';
 
 		$intro           = esc_html__( 'With the support of readers like you, we provide thoughtfully researched articles for a more informed and connected community. This is your chance to support credible, community-based, public-service journalism. Please join us!', 'newspack' );
 		$content_heading = esc_html__( 'Donation', 'newspack' );
@@ -487,7 +487,7 @@ class Donations {
 			$page_id = get_option( self::DONATION_PAGE_ID_OPTION, 0 );
 		}
 		if ( ! $page_id || 'page' !== get_post_type( $page_id ) ) {
-			return false;
+			$page_id = self::create_donation_page();
 		}
 
 		return [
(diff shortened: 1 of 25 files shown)


claudiulodro:
Thanks for tackling this!

rickybanister:
After some testing of this I think we will want to hide the `Plan Name` and `Add Image` fields in the form for the time being since they are not rendered in the Donations block. 

It might also be more intuitive to reorder the elements in the form like this:

Screen Shot 2019-08-05 at 5 56 11 PM

rickybanister:
The latest commit (removing title and image) looks good. I notice a slight jump in the height of the inputs when you toggle from single to three fields.

It looks like this rule is causing it:

image

We need to add margin-top:0 to the default single field as well I think. Or fix something float related.

rickybanister:
And also I'd prefer the toggle to go above the thing it's controlling (per my screenshot above) if possible.

claudiulodro:
> And also I'd prefer the toggle to go above the thing it's controlling (per my screenshot above) if possible. Unless that's a Material thing I don't know about.

This was changed because with that hierarchy it wasn't super clear whether the "Suggested donation amount" field does anything when the toggle is off: https://github.com/Automattic/newspack-plugin/pull/199#issuecomment-514802859

rickybanister:
That make sense @claudiulodro, as discussed we can revisit at a later date and consider a radio option instead of the toggle (if any customers find the UI confusing).

includes/class-donations.php
```diff
@@ -379,5 +380,122 @@ public static function process_donation_form() {
...
+		$page_args = [
+			'post_type'      => 'page',
+			'post_title'     => __( 'Support our publication', 'newspack' ),
+			'post_content'   => $page_content,
+			'post_excerpt'   => __( 'Support quality journalism by joining us today!', 'newspack' ),
+			'post_status'    => 'draft',
+			'comment_status' => 'closed',
+			'ping_status'    => 'closed',
+		];
+		return wp_insert_post( $page_args );
```
claudiulodro: Currently this creates a page every time the wizard is loaded; it should set the donation page after creating it.

```suggestion
		$page_id = wp_insert_post( $page_args );
		if ( is_numeric( $page_id ) ) {
			self::set_donation_page( $page_id );
		}
		return $page_id;
```

jeffersonrabb: Great catch. https://github.com/Automattic/newspack-plugin/pull/228/commits/23e95964e09ad03f9a5cda8b6a617da329944843

jeffersonrabb:
> In `Donations::get_donation_default_settings`, the `name` field should be updated from `''` to `__( 'Donation', 'newspack' )` otherwise the generated products have names like `:Yearly` since we've gotten rid of the name input.

Done in https://github.com/Automattic/newspack-plugin/pull/228/commits/e28d7857ce7437d3d63d3ce3c7811d6430d2566d

jeffersonrabb:
> One non-blocking UI issue I noticed is that at around 1000px-wide screens (when the tab contents go to two lines) the one-line active tabs are not lined up properly:

This should be resolved in https://github.com/Automattic/newspack-plugin/pull/228/commits/16b4341d006898e596860cabcd01ab0479eee001

claudiulodro:
Fixed GH download issue in https://github.com/Automattic/newspack-plugin/pull/228/commits/affc9c26a9cdd44ee4bc7b9193f43f193d0321b0. We'll most likely want to find a better way at some point in the future.

PR #239, opened 2019-08-13
Hook up global ad injection for Google Ads

claudiulodro:
### All Submissions:

This PR connects the dots between Newspack Plugin, Newspack Ads, and Newspack Theme to make the global ad functionality work. 

There will be a little styling needed in the theme after this PR is merged to make things centered and look nice. The search results/archive global ads need some further scoping to see what those settings are exactly supposed to do, so they'll be addressed in a follow-up PR in the future.

1. Set up an ad unit in Newspack Ads.
2. Test setting the unit in above header/below header/above footer and previewing on the frontend. Currently it should look something like this:

Below header:
Screen Shot 2019-08-13 at 9 30 58 AM

Above header:
Screen Shot 2019-08-13 at 9 30 32 AM

Above footer:
Screen Shot 2019-08-13 at 9 30 03 AM

### Questions
Do we want the ability to set global ads in the top and bottom of the sidebar? Those seem like common ad placements on sites and aren't currently supported (but can be manually done with the custom html widget). Maybe an ad widget?

### Other information:

### DIFF
--- a/includes/wizards/class-advertising-wizard.php
+++ b/includes/wizards/class-advertising-wizard.php
@@ -65,6 +65,9 @@ class Advertising_Wizard extends Wizard {
 	public function __construct() {
 		parent::__construct();
 		add_action( 'rest_api_init', [ $this, 'register_api_endpoints' ] );
+		add_action( 'before_header', [ $this, 'inject_above_header_ad' ] );
+		add_action( 'after_header', [ $this, 'inject_below_header_ad' ] );
+		add_action( 'before_footer', [ $this, 'inject_above_footer_ad' ] );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/wizards/class-advertising-wizard.php
```diff
@@ -466,9 +469,9 @@ private function get_placement_data( $placement ) {
 		$option_value = json_decode( get_option( self::NEWSPACK_ADVERTISING_PLACEMENT_PREFIX . $placement, '' ) );
 
 		$defaults = array(
-			'adUnit'  => '',
-			'enabled' => false,
-			'service' => '',
+			'ad_unit'  => '',
```
claudiulodro: This was a bug I noticed. Everywhere else `ad_unit` not `adUnit` is expected.

PR #241, opened 2019-08-13
Ad Picker Improvements

jeffersonrabb:
### All Submissions:

This PR addresses a situation where only one Ad Service is enabled and because of this the service cannot be selected, making it impossible to view the Ad Unit select. This addresses the problem in two ways:

- If only one Ad Service is enabled, the data returned by the API will have that service automatically selected for all placements.
- The first item in the Select form elements is no longer disabled. This means it is possible to make a selection when there is only one item. It also means the Ad Unit can be unset, which was previously impossible.

1. `npm run build:webpack`
2. Navigate to Advertising Wizard.
3. Turn on only Google Ad Manager. 
4. Click Global Settings tab, switch on one or more placements.
5. Observe that the Ad Provider select is automatically set to Google Ad Manager and the Ad Unit select is visible.
6. Switch on a second Ad Service.
7. On Global Settings switch on another placement.
8. Observe that the select has two elements and a selection can be made normally.

### Other information:

### DIFF
--- a/assets/wizards/advertising/views/placements/ad-picker/index.js
+++ b/assets/wizards/advertising/views/placements/ad-picker/index.js
@@ -21,9 +21,8 @@ class AdPicker extends Component {
 	adUnitsForSelect = adUnits => {
 		return [
 			{
-				label: __( 'Select an ad unit' ),
+				label: __( '- Select an ad unit -' ),
 				value: null,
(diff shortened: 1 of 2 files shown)


jeffersonrabb:
Thanks for reviewing!

Issue #188, opened 2019-07-16
Plugin Installation Failures Under Poor Network Conditions


jeffersonrabb:
**Describe the bug**
While helping @iamstevebeatty run through Newspack's Set Up Wizard we encountered some plugin installation issues. We were working on his local VVV instance using an Ngrok-supplied public URL. At the time his network was extremely spotty and everything was running painfully slowly. During the Set Up Wizard, several plugin installation API requests timed out and failed. When trying to run the Wizard again, these failed immediately with the error `The plugin directory already exists`. The only way were able to resolve was by manually deleting the plugin directories and re-running the Setup Wizard.  

**To Reproduce**
This is a tough one to repro. I've tried a few times playing with Network Throttling but it's still very tough to achieve. 

**Expected behavior**
At minimum we'll need a strategy for handling `The plugin directory already exists` errors, even if the best we can do is inform the user and offer instructions for cleaning up.

Plugin installation should retry and/or message on failure. Currently this doesn't happen in the Setup Wizard.

Issue #97, opened 2019-06-11
Installing/Activating a plugin in a separate tab breaks Plugin Installer flow.


claudiulodro:
**Describe the bug**
Originally reported at https://github.com/Automattic/newspack-plugin/pull/86#issuecomment-500390719.

Installing/Activating a required plugin in a separate tab will cause errors to be shown in the Plugin Installer and prevent proceeding to the next screen.

**To Reproduce**
1. Deactivate AMP. Go to the Components Demo. Observe that 'Use' is next to the AMP plugin on the Plugin Installer component.
2. In a separate tab, activate the AMP plugin using the regular WP plugins screen.
3. In the original tab, click 'Use' to try and activate the AMP plugin. Observe a "The plugin is already active" error displays and the Plugin Installer fails.

**Screenshots**
<img width="483" alt="Screen Shot 2019-06-11 at 8 20 51 AM" src="https://user-images.githubusercontent.com/7317227/59285344-b2ea7f80-8c22-11e9-9222-0bc1bafbb922.png">

**Expected behavior**
I think either the `PluginInstaller` component should ignore `newspack_plugin_already_active` and `newspack_plugin_already_installed` errors, or we shouldn't throw those errors in the `Plugin_Manager` (just return success if someone tries to re-activate an active plugin).

adekbadek:
Not reproducible anymore.

PR #249, opened 2019-08-28
Install The Newspack Theme

jeffersonrabb:
### All Submissions:

This PR adds the Newspack Theme to the list of plugins and takes care of handling it as an unusual case. The theme is installed as part of the Setup Wizard. The PR also includes a stub of a configuration manager for the theme, which at present only sets the colors to be default (which is  the theme default anyway). This configuration manager will eventually be the main interface between the Site  Design wizard and the Theme. The PR also contains some minor UI/text changes to the Setup Wizard to indicate that the theme is being installed as well as plugins.

1. Check out this branch, run `npm run build:webpack`
2. Deactivate and delete the theme.
3. Run through the Setup Wizard. 
4. Verify that the theme has been installed and is the active theme. 

### Other information:

### DIFF
--- a/assets/wizards/setup/index.js
+++ b/assets/wizards/setup/index.js
@@ -23,7 +23,7 @@ import './style.scss';
 import { HashRouter, Redirect, Route, Switch } from 'react-router-dom';
 import { pickBy, includes, forEach } from 'lodash';
 
-const REQUIRED_PLUGINS = [ 'jetpack', 'amp', 'pwa', 'wordpress-seo', 'google-site-kit' ];
+const REQUIRED_PLUGINS = [ 'jetpack', 'amp', 'pwa', 'wordpress-seo', 'google-site-kit', 'newspack-theme' ];
 
 /**
  * Wizard for setting up ability to take payments.
@@ -174,7 +174,7 @@ class SetupWizard extends Component {
 						exact
 						render={ routeProps => (
 							<Welcome
-								buttonText={ __( 'Install core plugins' ) }
+								buttonText={ __( 'Install core plugins and Newspack theme' ) }
 								buttonAction={ {
 									href: '#/about',
 									onClick: () => this.updateProfile(),
@@ -184,7 +184,7 @@ class SetupWizard extends Component {
 								profile={ profile }
 								notice=<p>
 									{ __(
-										'Clicking “Get Started” will install core Newspack plugins in the background.'
+										'Clicking “Get Started” will install core Newspack plugins and the theme in the background.'
 									) }
 								</p>
 							/>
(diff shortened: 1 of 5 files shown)


jeffersonrabb:
When the Newspack theme is initially installed the API returns status of `uninstalled`. A page refresh corrects this, but this glitch throws off the Setup Wizard, which won't complete. This does not occur if the theme is installed but inactive—it happens only if the theme is being installed and activated. 

The problem seems to lie in https://github.com/Automattic/newspack-plugin/pull/249/files#diff-044b9390f497ea129e19496244893c6eR191, where `wp_get_theme()` returns false immediately after https://github.com/Automattic/newspack-plugin/blob/2ff253f84882fa6782f6d12fbd114017c33861fa/includes/util.php#L42-L78 has been executed. This can be verified by following these steps:

- Add `error_log( 'Current theme URI: ' . $plugin_uri );` immediately below this line https://github.com/Automattic/newspack-plugin/blob/2ff253f84882fa6782f6d12fbd114017c33861fa/includes/class-plugin-manager.php#L192
- `tail` the PHP error log. 
- Delete the Newspack theme.
- Run the Setup wizard and observe the log.

claudiulodro:
I've simplified the check so that it doesn't require the theme to be fully initialized and loaded to work (it just checks the active stylesheet): https://github.com/Automattic/newspack-plugin/pull/249/commits/6b1f25b838a254cab8c67c8064a900eb6901f0be

This has solved the issue in my testing.

jeffersonrabb:
After running the Setup Wizard I'm sometimes seeing two instances of the Newspack theme in the themes admin. There will be two folders in the theme directory, one with the correct name (`newspack-theme`) and one with a string appended (`newspack-theme-iov549`).  Also see this PHP warning in the log:

`PHP Warning:  rename(/srv/www/wordpress-simple/public_html/wp-content/themes/newspack-theme-iov549,/srv/www/wordpress-simple/public_html/wp-content/themes/newspack-theme): File exists in /srv/www/newspack-repos/newspack-plugin/includes/util.php on line 68`

jeffersonrabb:
Looking a bit deeper, these problems stem from a few React problems with the Setup Wizard. The key is the handling of the `hidden` prop in `withWizardScreen`:

https://github.com/Automattic/newspack-plugin/blob/master/assets/components/src/with-wizard-screen/index.js#L39-L45

If `hidden` changes during the lifecycle of the wizard, the content will be re-rendered. In the Setup wizard this means that if the 4th screen is reached before all plugin installation is complete, `PluginInstaller` will be re-mounted. This causes a second API request to be made. The failed & double theme installations occur because two request are in flight at the same time which is likely to bork just about any plugin or theme installation—one will fail since the directory already exists. This creates a race condition, which explains the difficulty we've had reliably reproducing  this issue. 

Since it's unrelated to the work in this PR I'm going to merge, and will open a new PR with fixes for the Setup wizard. 

Likely related to https://github.com/Automattic/newspack-plugin/issues/188, possibly https://github.com/Automattic/newspack-plugin/issues/97

PR #254, opened 2019-09-01
Fixes To Setup Wizard Plugin Installation

jeffersonrabb:
### All Submissions:

Fixes a few issues with the Setup wizard that caused sporadic installation failures. The issues relate to the rendering of the `InstallationProgress` view which contains the `PluginInstaller` component. In the `Welcome` route (`/`) `InstallationProgress` isn't rendered at all. Then, in all subsequent views it's rendered hidden, with one exception: if installation is not yet complete the `Newsroom` view's Continue button leads to a visible version of the `InstallationProgress` screen. Due to problems in the handling of the `hidden` prop, this visible version of `InstallationProgress` caused `PluginInstaller` to re-mount, which led to two competing plugin installation API calls at the same time. In this situation the first API request succeeds, but the second one fails because the plugin directory already exists. This failure is reported in the UI and in some cases leads to duplicate installations.

This PR fixes these problems in a few ways:

- Correct `withWizardScreen`'s handling of the `hidden` prop so that the content is not re-rendered when the prop changes from true to false. 
- Change `InstallationProgress`'s route so it is rendered on all screens of the Setup wizard (usually hidden). 

Related to problems discussed in https://github.com/Automattic/newspack-plugin/pull/249
Possibly resolves https://github.com/Automattic/newspack-plugin/issues/188, although we may want to investigate hardening the APIs so they assess and handle the case where the plugin directory already exists.

1. On `master` branch, deactivate and delete several plugins. Also activate a theme other than Newspack and delete the Newspack theme.
2. Navigate to the Setup wizard, and click through the buttons on the first three screens as quickly as possible. The goal is to get to the `InstallationProgress` screen before the plugins finish installing.
3. Observe installation progress on the `InstallationProgress` screen. One or more should fail. With `The plugin directory already exists.` error message. This may need to be tried a few times to repro. In some cases, there won't be an error but the blue checkbox will fail  to appear. 
4. Do the same flow, but after clicking the blue button on the first view, click the "back" button in the upper left corner and then click through all blue buttons. Since the plugin installer is not rendered in the `/` route, this should also lead to some duplicate API hijinx.
5. Switch branches to `fix/plugin-installer-double-mount`, run `npm run build:webpack`, and try all of the above flows again. They should succeed consistently. 

### Other information:

### DIFF
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -30,6 +30,7 @@ class PluginInstaller extends Component {
 		super( ...arguments );
 		this.state = {
 			pluginInfo: {},
+			installationInitialized: false,
 		};
 	}
 
@@ -42,9 +43,13 @@ class PluginInstaller extends Component {
 	};
 
 	componentDidUpdate = prevProps => {
-		const { plugins } = this.props;
+		const { autoInstall, plugins } = this.props;
+		const { installationInitialized } = this.state;
 		if ( plugins !== prevProps.plugins ) {
 			this.retrievePluginInfo( plugins );
+		}
+		if ( autoInstall && ! installationInitialized ) {
+			this.installAllPlugins();
 		}
 	};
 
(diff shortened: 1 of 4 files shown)


jeffersonrabb:
Thanks for the review!

PR #267, opened 2019-09-07
Add notice if Disqus is run without compatibility plugin

claudiulodro:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds a notice to the admin screen if the user is running Disqus without the AMP-compatibility plugin. The notice has a link that installs and activates the Disqus-AMP-compatibility plugin when clicked:

Screen Shot 2019-09-07 at 7 31 29 AM
Screen Shot 2019-09-07 at 7 31 43 AM

If this seems like a good approach, we can do the same treatment for TablePress and the TablePress-AMP plugin.

### How to test the changes in this Pull Request:

1. Install and activate Disqus plugin. Observe notice at the top of screen with warning.
2. Click link in notice to install AMP-compatibility plugin. Observe successful installation.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #268, opened 2019-09-11
Remove DocumentCloud AMP compatibility tweak because it's fixed in AMP

claudiulodro:
See https://github.com/ampproject/amp-wp/pull/3209

I've kept the AMP_Enhancements class because of https://github.com/Automattic/newspack-plugin/pull/267, and we'll likely still need it for other situations.


jeffersonrabb:
https://github.com/ampproject/amp-wp/releases/tag/1.3.0 has been released, so we can merge this now.

PR #274, opened 2019-09-13
Unified Engagement Wizard

jeffersonrabb:
### All Submissions:

This  is a stub Engagement Wizard, with navigation UI and basic functionality for most features. 

#### Newsletters

Screen Shot 2019-09-12 at 11 40 51 PM

This is a combination of  the  existing "Set Up Mailchimp" and "Set Up Newsletter Subscriptions Block" Wizards. The upper `Card` provides description and link to connect a Mailchimp account on WPCOM. The lower `Card` manages installation of the WooCommerce Mailchimp extension and allows input of Mailchimp API key. The lower section will only be shown if WooCommerce is installed and active. Note that the API Key is not persisted, which is the status quo in the "Set Up Mailchimp" Wizard.

#### Social

This page contains an `ActionCard` with Handoff to the Jetpack Publicize feature.

Screen Shot 2019-09-12 at 11 51 32 PM

#### Commenting

Commenting has a three-item submenu. `Disqus` and `The Coral Project` pages will install the respective plugins, and then show an `ActionCard` with handoff to their dashboards. `WordPress Discussion` is an `ActionCard` that simply links to Discussion settings. This isn't a true Handoff because of some complexities linking to a page that isn't associated with an installed plugin. I elected not to resolve this issue yet, since it's possible that instead of sending to the Discussion dashboard this screen will be responsible for configuring it. 

Screen Shot 2019-09-12 at 11 59 31 PM

#### UGC

The UGC area is placeholder.

Screen Shot 2019-09-13 at 12 10 27 AM

This PR also adds a notification (based on https://github.com/Automattic/newspack-plugin/pull/267) if Disqus plugin is active but Newspack AMP Disqus is not. Between the automated installation of the two together and this notification, this should supersede https://github.com/Automattic/newspack-plugin/pull/267.

1. Checkout the branch and run  `npm run build:webpack`
2. Navigate to Engagement Wizard.
3. Click through all pages, verify that their appearance and behavior matches the descriptions and screenshots above. 

### Other information:


includes/wizards/class-engagement-wizard.php
```diff
@@ -95,12 +89,15 @@ public function register_api_endpoints() {
 	 * @return WP_REST_Response with the info.
 	 */
 	public function api_get_connection_status_settings() {
-		$configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'jetpack' );
-		return rest_ensure_response( $configuration_manager->get_mailchimp_connection_status() );
+		$jetpack_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'jetpack' );
+		$wc_configuration_manager      = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'woocommerce' );
+		$response                      = $jetpack_configuration_manager->get_mailchimp_connection_status();
+		$response['wcConnected']       = $wc_configuration_manager->is_active();
```
claudiulodro: [`$response` will be a `WP_Error` if connection is not good](https://github.com/Automattic/newspack-plugin/blob/583fe2687032b51ab57ee3a1bf41bc740ac7a762/includes/configuration_managers/class-jetpack-configuration-manager.php#L55), so this will cause fatal errors if you try and use a `WP_Error` like an array.

```suggestion
		if ( ! is_wp_error( $response ) ) {
			$response['wcConnected']   = $wc_configuration_manager->is_active();
		}
```

assets/wizards/engagement/views/commenting-native/index.js
```diff
@@ -0,0 +1,35 @@
...
+ * Native Commenting Screen
+ */
+class CommentingNative extends Component {
+	/**
+	 * Render.
+	 */
+	render() {
+		return (
+			<ActionCard
+				title={ __( 'WordPress Commenting' ) }
```
claudiulodro: Would be cool to make this a real handoff in the future. It's hard to get back to the wizard without the "Back to Newspack" banner at the top of the screen (I'm so used to having it available!). Not a blocker for this PR; we'll address when we build this section out properly.

assets/wizards/engagement/views/social/index.js
```diff
@@ -0,0 +1,41 @@
...
+class Social extends Component {
+	/**
+	 * Render.
+	 */
+	render() {
+		const { connected, connectURL } = this.props;
+
+		return (
+			<ActionCard
+				title={ __( 'Jetpack Publicize' ) }
```
claudiulodro: Publicize and Social Icons are both on the Jetpack settings page this links. We should probably also call attention to the Social Icons feature here. Sites can customize the style of the buttons and whether they want them on posts or not. Not a blocker for this PR, though; we'll address when we fully build out this section.

includes/class-admin-plugins-screen.php
```diff
@@ -257,6 +261,10 @@ public function unsupported_plugin_notifications() {
 			$messages[] = __( 'The Newspack Theme is not currently active.' );
 		}
 
+		if ( $disqus_notification ) {
+			$messages[] = __( 'You are using the Disqus comment system without the Newspack Disqus AMP plugin. The comments form will not display on AMP pages. Correct this ' ) . '<a href="/wp-admin/admin.php?page=newspack-engagement-wizard#/commenting/disqus">' . __( 'here' ) . '.</a>';
```
claudiulodro: Good solution. I like this approach. 👍

includes/wizards/class-engagement-wizard.php
jeffersonrabb: Good catch https://github.com/Automattic/newspack-plugin/pull/274/commits/5a88e2de4628711f7aaaa4b4f1cb2a2ff8a6b7f1

assets/wizards/engagement/views/commenting-native/index.js
jeffersonrabb: Yes, I was just being lazy. Done in https://github.com/Automattic/newspack-plugin/pull/274/commits/3c65ae84eeb354f764cb2dbf53f7e1b24492e248. `wordpress-settings-discussion` is defined as a plugin, with the parameter `WPCore`. That param indicates that the plugin is actually just a WP core feature, and treats it accordingly.

jeffersonrabb:
Thanks for reviewing!

PR #275, opened 2019-09-13
Back Button

jeffersonrabb:
### All Submissions:

Adds a back button to the upper left corner of all Wizards. The icon is the `screenoptions` Dashicon: https://developer.wordpress.org/resource/dashicons/#screenoptions. The button links to the Dashboard. This UI should be present on all Wizards except for Setup,  which uses a linear  approach to navigation and pagination in the upper left & right corners.

Screen Shot 2019-09-13 at 12 40 37 AM

1. Check out the  branch and run `npm run build:webpack`
2. Navigate to all Wizards, verify the  appearance of the icon, click and verify it links to Dashboard.
3. Run through the Setup Wizard and verify that this works as it did before.

### Other information:

### DIFF
--- a/assets/components/src/wizard-pagination/index.js
+++ b/assets/components/src/wizard-pagination/index.js
@@ -7,6 +7,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
+import { Dashicon } from '@wordpress/components';
 
 /**
  * External dependencies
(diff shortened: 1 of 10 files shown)


rickybanister:
I think it's worth adding the text 'Dashboard' to the right of the icon to create a little extra affordance and bigger hit state. 

👍

rickybanister:
@claudiulodro makes a very good point. When looking at this through a straw, the back button seemed more appropriately located up in that corner, but it does feel a tad redundant. Also since the dashboard is currently mostly a gathering place for settings, rather than a frequently-visited area for stats and things, I think we could drop the button. The logo linking to the dash is already a nice improvement. Maybe we archive this PR so if the need arises later we can resurrect?

jeffersonrabb:
> Maybe we archive this PR so if the need arises later we can resurrect? 

Agreed. Closing for now...

PR #278, opened 2019-09-26
Update Primary Color for better accessibility

thomasguillot:
### All Submissions:

I updated the primary colour (`#2a7de1`) to make it more accessible since the contrast ratio was failing the WCAG AA tests. It's also based on a [Pantone colour](https://www.pantone.com/color-finder/2386-C) which makes it easier for printing.

* Original contrast ratio: 4.1
* New contrast ratio: 5.4

__Changes:__
* Switch `#2a7de1` to `#2d68c4` and introduce new darker/lighter primary colors
* Improve some of the colours of the buttons
* Update Newspack logo with new colour (and simplify its shape)

__Colour update example:__
Screenshot 2019-09-27 at 13 12 13

1. Go through the different screens of the plugins.
2. Do you notice the new blue colour for the logo, buttons, links?

### Other information:


pablohoneyhoney:
I don't have an issue with the color change. One of the initial agreed conditions of the logo change was to keep the original color. Aesthetically, the darker tone feels more serious and institutional, which is probably not the desire effect. The original seems fresher. 
Another path is to take a pass at secondary blues for legibility in texts and CTAs but keep the brand color as is.

I don't think a color exercise was done around this blue, primary colors, secondary, UI color schemes, etc. Worth the exercise now with @thomasguillot in the house : )

jeffersonrabb:
> Another path is to take a pass at secondary blues for legibility in texts and CTAs but keep the brand color as is.

I'd be ok with this direction, where we leave the logo untouched but change the primary color everywhere else it's used. @thomasguillot are you OK with this approach? I put the PR on hold pending discussion of this point.

thomasguillot:
My main concern with having one blue for the logo and another blue for the UI is that it would be the only Automattic brand behaving like this.

I can get a ratio contrast of 4:51 with #1f75dd which passes the AA test. The difference between the original blue is very subtle. Would that work better @pablohoneyhoney?

Screenshot 2019-09-30 at 10 24 24
before (#2a7de1) / after (#1f75dd)

pablohoneyhoney:
> My main concern with having one blue for the logo and another blue for the UI is that it would be the only Automattic brand behaving like this.
The fact that this happens doesn't mean is correct : )
We generally lack richness in brand and UI leaving our experiences flat. And this would be a good opportunity to add more depth to Newspack both as a brand and as an interface.

pablohoneyhoney:
I actually don't dislike that #1f75dd, but I'd test it more in real scenarios and scales, and I still think we need to add more depth to Newspack.

thomasguillot:
Closing this as the change of the Primary Color will require more work/research/testing. I will do a separate PR with just the subtle changes on the buttons that this PR had.

thomasguillot:
See #280 for the more subtle changes of this PR

PR #279, opened 2019-09-26
Update the CSS of the Components for more consistency

thomasguillot:
### All Submissions:

Small CSS tweaks to all the components to bring them closer to the Calypso components, including:
* Reworked the general font-size/line-height
* Reduced padding of the buttons
* Modified focus behaviour of input/select
* Unified paddings/margins

1. Run the setup wizard
2. Check the dashboard and all the pages
3. Look at the Components demo page

### Other information:


rickybanister:
This looks pretty great @thomasguillot, could you share a bit of the 'why' behind these changes (what was inconsistent etc) just so we're all on the same page and don't repeat mistakes going forward.

assets/components/src/button/style.scss
```diff
@@ -7,16 +7,17 @@
...
 
 	&.is-button {
 		background-color: $muriel-white;
 		border-color: #cccccc;
-		border-radius: 3px;
+		border-radius: 4px;
 		height: auto;
-		padding: 16px 32px;
+		line-height: 21px;
+		padding: 7px 14px 9px;
```
jeffersonrabb: The change in `Button` padding is almost certainly the most noticeable aspect of this PR. Originally the buttons were designed to Muriel spec, including the fairly significant padding. I'm all for trimming them down, but would like to know where the new padding comes from? Is it purely an aesthetic improvement or does it draw from any particular design framework?

thomasguillot:
> This looks pretty great @thomasguillot, could you share a bit of the 'why' behind these changes (what was inconsistent etc) just so we're all on the same page and don't repeat mistakes going forward.

@rickybanister my main "concern" was the inconsistency between some of the components. There were different font-sizes with different line-heights and paddings, no real vertical rhythm.

assets/components/src/button/style.scss
thomasguillot: I copied the padding from Calypso... Thinking about it now, it doesn't make much sense since our vertical rhythm is currently based on a 21px grid 🤦‍♂ 

If we want to keep the large buttons, I think we can have 2 types of buttons: "default" (42px) and "large" (63 or even 84px).

assets/components/src/card/style.scss
```diff
@@ -9,12 +9,12 @@
 	border-radius: 3px;
 	box-shadow: 0px 1px 1px rgba( 0, 0, 0, 0.25 );
 	margin: 0 auto 10px auto;
-
-	max-width: calc( 682px - 62px );
-	padding: 31px;
+	max-width: calc( 682px - 48px );
```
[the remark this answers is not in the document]
thomasguillot: Removed in 5c0a14b

assets/components/src/button/style.scss
thomasguillot: See: 2c2d78d

PR #299, opened 2019-10-29
Fake Newspack

jeffersonrabb:
### All Submissions:

This branch creates starter content so that after running the Newspack Setup Wizard a brand new site comes pre-populated with placeholder content and is immediately ready to test. This is useful for demo-ing Newspack and giving new and prospective users a way to try it out. Here are the elements of starter content:

- These categories  are created:  `'Featured', 'Sports', 'Entertainment', 'Opinion', 'News', 'Events', 'Longform', 'Arts', 'Politics', 'Science', 'Tech', 'Health'`
- 40  posts are created, each with title and body text randomly generated by https://www.lipsum.com and a randomly selected featured image imported from https://picsum.photos.
- The 40 posts are assigned one of the starter categories, three posts per category. This approach insures that there are enough posts in each category to populate homepage blocks.
- Homepage is created, with a great layout @laurelfulford designed containing numerous Homepage Articles blocks. 
- The theme is set up with Style Pack 3, Solid Header Background, Simplified Header. The Newspack logo is installed in the media library and designated as the theme logo.

The starter content is created in the final screen of the Setup Wizard, which makes numerous (43 to be exact) API requests, one at a time. These are done sequentially to avoid making too many API requests for Lipsum and stock photos in too short of a time, and to avoid locking situations which can cause the Newspack API calls to fail.

Screen Shot 2019-11-03 at 11 03 04 AM

Previously, if `WP_DEBUG` is true, the Setup Wizard does not require  that Jetpack and Site Kit be connected. Jurassic Ninja instances come with `WP_DEBUG` on, which makes demo-ing  the Setup Wizard tricky. To get around this, the flag has been changed to `WP_NEWSPACK_DEBUG`. Also, this branch introduces a Newspack Settings page, with a  checkbox to enable this flag. 

Screen Shot 2019-11-03 at 11 02 46 AM

There are two ways to test this branch:

#### Local Instance

1) On a local instance, delete all posts, pages, and categories. 
2) Ideally remove the Newspack theme, or at least try to reset all theme mods. 
3) Run through the Setup Wizard and on the final screen click `Install Starter Content`. 
4) After the Wizard completes, verify the categories, posts, homepage content, and theme mods.

#### Jurassic Ninja

1) Create a release from this branch `npm run release`. 
2) Spin up a new Jurassic Ninja instance with no special options. http://jurassic.ninja/
3) When the site is ready, go to plugins and install Newspack from the release created in step one (in `assets/release/`)
4) Before running Setup, navigate to Settings->Newspack and check `Debug Mode`. This will save the trouble of activating/connecting Jetpack and Site Kit on the test site.
5) Run through the Setup Wizard and on the final screen click `Install Starter Content`. 
6) After the Wizard completes, verify the categories, posts, homepage content, and theme mods.

### Other information:


includes/class-starter-content.php
```diff
@@ -0,0 +1,368 @@
...
+
+		$newspack_featured_image_position = $newspack_featured_image_positions[ wp_rand( 0, 2 ) ];
+
+		if ( $newspack_featured_image_position ) {
+			add_post_meta( $post_id, 'newspack_featured_image_position', $newspack_featured_image_position );
+		}
+
+		$attachment_id = self::add_featured_image( $post_id );
+
+		$categories = get_categories(
```
claudiulodro: There are a few issues with the category-setting logic (see the Count column):

Screen Shot 2019-11-03 at 12 11 32 PM


I think a much simpler way of finding the correct category to set would be like:

```
$categories = get_categories([
	'orderby'    => 'count',
	'order'      => 'ASC',
	'hide_empty' => false,
	'exclude'    => '1', // Exclude 'Uncategorized' which has ID of 1.
	'number'     => 1,
]);
$category_to_set = $categories[0];
```
This should get the category containing the least amount of posts, so it will always evenly distribute the posts among the categories.

jeffersonrabb: Good call - this is much more elegant. Change made  in https://github.com/Automattic/newspack-plugin/pull/299/commits/e631432bfa9966033fc981500088d98843938f37

PR #318, opened 2019-11-26
Designate our supported S3 integration

claudiulodro:
### All Submissions:

This PR marks The Human Made S3 integration as our recommended S3 integration for people that need S3. 

They don't provide a nice zip like we do on their releases, so we need to tag a specific version (which is not ideal). We can re-approach this in the future and see what we can do about it; there should be something we can do with the GitHub API.

1. Verify S3 plugin shows up in the list of Newspack plugins on the Plugins screen. It can be installed through the plugins list, however once installed the installed plugin won't show up in the Newspack list because of the `-2.2.1` appended to the folder (it will be in the regular plugins area). That's going to require some adjustments to the installer to sort out, so we can handle those in a separate PR.

### Other information:

### DIFF
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -245,6 +245,14 @@ class Plugin_Manager {
 				'PluginURI'   => 'http://wordpress.org/extend/plugins/co-authors-plus/',
 				'Download'    => 'wporg',
 			],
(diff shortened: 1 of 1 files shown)


adekbadek:
Closing, this seems quite stale.

PR #329, opened 2019-12-15
TextControl: Update the component to simplify it and mostly rely on BaseComponent with custom styles

thomasguillot:
### All Submissions:

As part of the redesign of the plugin (see #322) I reworked/simplified the `TextControl` to behave in a similar way as Core, with the Newspack styles.

__Before:__

1 before

__After:__
2 after

I also took the opportunity to rework the Money Input by having the currency next to the input.

__Before:__

3 before

__After:__

4 after

_Note: I'm aware of a margin issue in the location setup. I plan to fix this in #322._

1. Check the Components Demo, Setup, Reader Revenue. You should see the current text inputs
2. Switch to this branch
3. Recheck the different pages. Do you see the new text inputs?

### Other information:

### DIFF
--- a/assets/components/src/text-control/index.js
+++ b/assets/components/src/text-control/index.js
@@ -5,82 +5,28 @@
 /**
  * WordPress dependencies
  */
-import { TextControl as BaseComponent, withFocusOutside } from '@wordpress/components';
+import { TextControl as BaseComponent } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
-import murielClassnames from '../../../shared/js/muriel-classnames';
 import './style.scss';
 
-const TextControl = withFocusOutside(
+/**
+ * External dependencies
+ */
+import classNames from 'classnames';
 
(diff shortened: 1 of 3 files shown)


thomasguillot:
> Looks great, especially the currency display. Tested in latest Chrome/Safari/Firefox/Edge, IE11, iOS, all looks good. The only thing that I noticed is in iOS when the text input is active the page zooms slightly which pushes the right edge of the input offscreen. This doesn't seem to occur on `master`. However, I think this may be expected/correct behavior, and in fact there may be something weird on `master` that is suppressing the behavior? Even if this is a problem to be fixed it needn't hold anything up, I just wanted to mention it.

It should be fixed in 9db6ac5. Can you confirm this?

jeffersonrabb:
> It should be fixed in [9db6ac5](https://github.com/Automattic/newspack-plugin/commit/9db6ac5b5bfde38b7713943dd2b94fab83da9e25). Can you confirm this?

Confirmed, this fixes it.

PR #350, opened 2019-12-23
Image Upload: Update component to use Newspack's `Button`

thomasguillot:
### All Submissions:

* Use Material Icon for add/remove image
* Uses `isTertiary` button

1. Go to Performance > Site Icon
2. The current button is easily missable as it doesn't look like a button.
3. Switch to this branch
4. Do you see a `isTertiary` button with an icon? Does the button work?

### Other information:

### DIFF
--- a/assets/components/src/button/style.scss
+++ b/assets/components/src/button/style.scss
@@ -16,6 +16,13 @@
 		line-height: 24px;
 	}
 
+	&.is-small {
+		font-size: 14px;
+		height: auto;
+		line-height: 24px;
+		padding: 3px 12px;
+	}
+
 	&.is-button {
 		background: white;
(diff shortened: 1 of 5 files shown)


thomasguillot:
@jeffersonrabb I have reworked this a bit, see https://cloudup.com/cPlvxUItF-j

thomasguillot:
> The Add Icon UI is a little misleading because it feels like an upload hit area, but is actually just a button. My first reaction was to drag an asset onto the button, which didn't have the expected result. @claudiulodro would there be an easy fix to make this a true upload target? If this change isn't straightforward we may be better off switching back to a more button-like UI and looking into this feature later.

@jeffersonrabb that's a good point.
I've remove the style in a6a58af

PR #371, opened 2020-01-07
Redirect To Setup Wizard After Activation

jeffersonrabb:
### All Submissions:

After activating the Newspack plugin, redirect to the Setup Wizard. This will happen only once.

1. Deactivate and reactivate the Newspack Plugin.
2. Verify the redirect to the Setup Wizard.
3. Navigate away without completing setup, and verify the redirect does not happen again.

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -40,10 +40,12 @@ final class Newspack {
 	public function __construct() {
 		$this->define_constants();
 		$this->includes();
+		add_action( 'admin_init', [ $this, 'admin_redirects' ] );
 		add_action( 'admin_menu', [ $this, 'remove_all_newspack_options' ], 1 );
 		add_action( 'admin_notices', [ $this, 'remove_notifications' ], -9999 );
 		add_action( 'network_admin_notices', [ $this, 'remove_notifications' ], -9999 );
 		add_action( 'all_admin_notices', [ $this, 'remove_notifications' ], -9999 );
+		register_activation_hook( NEWSPACK_PLUGIN_FILE, [ $this, 'activation_hook' ] );
 	}
 
 	/**
@@ -52,6 +54,7 @@ final class Newspack {
 	private function define_constants() {
 		define( 'NEWSPACK_VERSION', '0.0.1' );
 		define( 'NEWSPACK_ABSPATH', dirname( NEWSPACK_PLUGIN_FILE ) . '/' );
+		define( 'NEWSPACK_ACTIVATION_TRANSIENT', '_newspack_activation_redirect' );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-newspack.php
```diff
@@ -40,10 +40,12 @@ public static function instance() {
 	public function __construct() {
 		$this->define_constants();
 		$this->includes();
+		add_action( 'admin_init', [ $this, 'admin_redirects' ] );
 		add_action( 'admin_menu', [ $this, 'remove_all_newspack_options' ], 1 );
 		add_action( 'admin_notices', [ $this, 'remove_notifications' ], -9999 );
 		add_action( 'network_admin_notices', [ $this, 'remove_notifications' ], -9999 );
 		add_action( 'all_admin_notices', [ $this, 'remove_notifications' ], -9999 );
+		register_activation_hook( NEWSPACK_PLUGIN_FILE, [ $this, 'activation_hook' ] );
```
thrijith: Please correct me if I am mistaken, just had a minor query with usage of `register_activation_hook` here,  will this work as expected on wpcom env based on the [VIP Go coding standards](https://wpvip.com/documentation/vip-go/code-review-blockers-warnings-notices/#plugin-registration-hooks)  docs it is not supported.

claudiulodro: Good catch! 

I don't believe this will work as expected on VIP Go, but Newspack anyways requires a developer to do a little special preparation for a site running on that specific environment before it can be effectively used. (Since VIP Go doesn't allow access to the file system, all desired plugins need to be pre-installed in the `plugins` directory.) Since a more advanced setup is already required for VIP Go, I think it's OK if this helpful redirect fails.

Alternately, something like this could work:
```
// The NEWSPACK_SETUP_COMPLETE check is for backwards compatibility, so existing users don't get redirected to the setup screen next time they go to the admin.
add_action( 'admin_init', function() {
  if ( ! get_option( 'newspack_setup_visited' && ! get_option( NEWSPACK_SETUP_COMPLETE ) ) {
    update_option( 'newspack_setup_visited', true );
    wp_safe_redirect( admin_url( 'admin.php?page=newspack-setup-wizard' ) );
  }
});
```

I think that has its own set of issues, though.

jeffersonrabb: I'd be inclined to leave this as-is, especially since it's a non-critical feature. If the hook fails on VIP Go the end result would be a slightly more cumbersome setup sequence, which is not a terribly great risk.  FWIW `register_activation_hook` seems to be used in most of the larger plugins Newspack integrates, including Jetpack, Site Kit, and AMP.

thrijith: Thanks for the info @claudiulodro and @jeffersonrabb

PR #375, opened 2020-01-09
Ad Wizard Improvements

jeffersonrabb:
### All Submissions:

This PR needs to be tested in tandem with https://github.com/Automattic/newspack-ads/pull/15. The two together bring a number of improvements to the Google Ad Manager integration. 

- Users no longer need to copy and paste raw code from the GAM dashboard, which was an awkward and tedious task.
- There is no longer a requirement that the user have the `unfiltered_html` capabilitiy.
- The risk of breaking a site by pasting slightly incorrect code is now removed.
- These changes pave the way for future iterations where Newspack synchronizes programmatically with a Google Ad Manager account.
- Users can now assign multiple sizes to a single Ad Unit.
- There's now a Widget. Ad Units can be placed in three ways: 1) pre-defined ad placements, 2) Gutenberg block, 3) Widget.

Ideally testing takes place with a Google Ad Manager account, but as an alternative @philipjohn has provided many sample GAM codes in pamTN9-kv-p2. This new work makes no attempt to convert old ad units and header code, so it's probably best to clear old material out and start fresh.

@thomasguillot, the multiple size interface is new UI and could use some finesse.

1. Navigate to Newspack->Advertising, switch on Google Ad Manager. Click "Configure," then the "Global Code" tab. Here you enter the Network Code, which can be found in the upper left of the GAM dashboard. It is also the first path component of all slots in the sample codes. 
2. Click the "Individual Ad Units" tab, and click "Add." 
3. "Ad Unit Name" is purely a label, it is never used in code. If this is left blank, the value of "Ad Unit Code" will be used for both.
4. "Ad Unit Code" is the machine-readable (slug-like) version of the Ad Unit title that GAM generates. You can get this code in the GAM dashboard by navigating to Inventory->Ad Units, clicking on any unit's title, and locating the "Code" field. In the sample code, this is the second path component in any slot definition.
5. Next click to add a size, which reveals a Width/Height pair. Enter values for both. Try adding multiples sizes, and try deleting some. Verify all of this UI behaves as expected, and that values are saved properly.
6. Try a Pre-defined ad placement. Navigate back to the front page of the Advertising Wizard then click the "Global Settings" tab. Toggle on one of the regions, set Ad Provider to Google Ad Manager, and select one of the Ad Units.
7. Verify the unit appears on published pages in both AMP and non-AMP requests.
8. Try the Gutenberg block. Edit a post, add the "Ad Unit" block. Select an ad unit, and verify that the block resizes to the unit's size (if the unit has multiple sizes, the first one will be used).
9. Publish and view the page, and verify the unit appears in AMP and non-AMP requests.
10. Try the Widget. Navigate to Appearance->Wizards, and add the Newspack Ad Unit Widget to a sidebar. Select an Ad Unit, and verify the unit appears in both AMP and non-AMP requests.
11. Probably wise to test with multiple units, deployed in different ways. Also, try some Ad Units with multiple sizes and some with only one.

### Screenshots

Screen Shot 2020-01-09 at 11 07 38 PM
Screen Shot 2020-01-09 at 11 07 59 PM
Screen Shot 2020-01-09 at 11 08 50 PM

### Other information:

### DIFF
--- a/includes/wizards/class-advertising-wizard.php
+++ b/includes/wizards/class-advertising-wizard.php
@@ -116,17 +116,17 @@ class Advertising_Wizard extends Wizard {
 		// Update header code.
 		register_rest_route(
 			'newspack/v1/wizard/',
-			'/advertising/service/(?P<service>[\a-z]+)/header_code',
+			'/advertising/service/(?P<service>[\a-z]+)/network_code',
 			[
 				'methods'             => \WP_REST_Server::EDITABLE,
-				'callback'            => [ $this, 'api_update_header_code' ],
-				'permission_callback' => [ $this, 'api_permissions_check_unfiltered_html' ],
+				'callback'            => [ $this, 'api_update_network_code' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
 				'args'                => [
-					'service'     => [
+					'service'      => [
 						'sanitize_callback' => [ $this, 'sanitize_service' ],
 					],
-					'header_code' => [
-						// 'sanitize_callback' => 'esc_js', @todo If a `script` tag goes here, esc_js is the wrong function to use.
+					'network_code' => [
+						'sanitize_callback' => 'sanitize_text_field',
 					],
 				],
 			]
@@ -159,25 +159,6 @@ class Advertising_Wizard extends Wizard {
 				'args'                => [
 					'service' => [
 						'sanitize_callback' => [ $this, 'sanitize_service' ],
-					],
-				],
-			]
-		);
-
-		// Update header code.
-		register_rest_route(
-			'newspack/v1/wizard/',
-			'/advertising/service/(?P<service>[\a-z]+)/header_code',
-			[
-				'methods'             => \WP_REST_Server::EDITABLE,
-				'callback'            => [ $this, 'api_update_header_code' ],
-				'permission_callback' => [ $this, 'api_permissions_check' ],
-				'args'                => [
-					'service'     => [
-						'sanitize_callback' => [ $this, 'sanitize_service' ],
-					],
-					'header_code' => [
-						// 'sanitize_callback' => 'esc_js', @todo If a `script` tag goes here, esc_js is the wrong function to use.
 					],
 				],
 			]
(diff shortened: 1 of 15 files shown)


philipjohn:
Re the second point above, after some testing I've found the following:

* AMP: The larger (468x60) banner appears
* Non-AMP: Whichever size appears first in the list shows

So it seems that in the non-AMP code we need to be particular about which size we place first in the array passed to `defineSlot()`. We probably want it to be the largest version, as that follows what the AMP code does.

includes/wizards/class-advertising-wizard.php
```diff
@@ -222,22 +203,24 @@ public function register_api_endpoints() {
...
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'code'       => [
 						'sanitize_callback' => 'sanitize_text_field',
 						'validate_callback' => [ $this, 'api_validate_not_empty' ],
 					],
-					'ad_code'     => [
-						// 'sanitize_callback' => 'esc_js', @todo If a `script` tag goes here, esc_js is the wrong function to use.
+					'sizes'      => [
+						// 'sanitize_callback' => 'sanitize_text_field', TODO: Correct sanitization for this field.
```
claudiulodro: Can we sanitize to make sure it's an array where each element is numeric or `''`? Does that meet all the use-cases?

jeffersonrabb: Added sanitization in https://github.com/Automattic/newspack-plugin/pull/375/commits/2009e4e7e49be060444557367edb86fb7abb4712. It's identical to sanitization being performed in Newspack Ads, but there's certainly no harm in doing it twice.

jeffersonrabb:
> 1. It's possible to not add any sizes when adding an ad unit, which means the ad unit doesn't show

This is resolved in https://github.com/Automattic/newspack-plugin/pull/375/commits/c710737b1d2ce9c21ef155502c3661ec6ddbd3bf and https://github.com/Automattic/newspack-plugin/pull/375/commits/dbaa0549cbd829e4eda3961a7066f5949ac544a1. New Ad Units initially have one size, and the deletion icons are only present if there are multiple sizes.

jeffersonrabb:
> We probably want it to be the largest version, as that follows what the AMP code does.

In https://github.com/Automattic/newspack-ads/pull/15/commits/b3b31d2f75befe81dd0db0c31949f04c06fff259 we're now ordering the sizes from widest to narrowest. As it stands, the sorting is done purely by width, ignoring the height.

PR #380, opened 2020-01-13
Wizard: (Re)add loading animation

thomasguillot:
### All Submissions:

In order to get visual feedback when clicking a button, this PR adds back the loading animation.

Go to the advertising wizard. And toggle the different ads providers

### Other information:

### DIFF
--- a/assets/components/src/with-wizard/index.js
+++ b/assets/components/src/with-wizard/index.js
@@ -243,28 +243,30 @@ export default function withWizard( WrappedComponent, requiredPlugins, options )
 							</a>
 						) }
 					</div>
-					<WrappedComponent
-						pluginRequirements={ requiredPlugins && this.pluginRequirements() }
-						clearError={ this.clearError }
-						getError={ this.getError }
-						setError={ this.setError }
-						startLoading={ this.startLoading }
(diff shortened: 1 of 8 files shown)


jeffersonrabb:
What's the thinking behind the switch to a one-time animation vs. infinite as it was originally?

thomasguillot:
> What's the thinking behind the switch to a one-time animation vs. infinite as it was originally?

@jeffersonrabb the original animation wasn't going from 0 to 100%. Instead it was doing a strange bouncing effect so it made sense to have in using `infinite`.

Now, it's a straightforward 0 to 100. And to me, it doesn't make sense to see the animation more than once since it should reach the 100 when it's fully loaded.

jeffersonrabb:
> And to me, it doesn't make sense to see the animation more than once since it should reach the 100 when it's fully loaded.

I'm frequently seeing the animation complete long before the loading finishes, which leaves the UI grayed out with no animation. This is on a local development instance so of course it's unusually slow, but this could occur on production as well:

https://cloudup.com/cfq_5u4YZjW

thomasguillot:
@jeffersonrabb In bdcccf9 I changed the animation to `infinite`. Is it any better?

PR #381, opened 2020-01-13
Fix the Notice RawHTML object conflict

thomasguillot:
### All Submissions:

Makes `RawHTML` as an option.
Can be set by adding `rawHTML` to a Notice.

__Before:__

1 before

__After:__

2 after

1. If a notice uses something like `ExternalLink` as part of its content. It will look broken (see screenshot above).
2. Switch to this branch. And check all the notices.

### Other information:

### DIFF
--- a/assets/components/src/notice/index.js
+++ b/assets/components/src/notice/index.js
@@ -53,7 +53,7 @@ class Notice extends Component {
 			<div className={ classes }>
 				{ noticeIcon }
 				<div className="newspack-notice__content">
-					{ !! rawHTML ? <RawHTML>{ noticeText }</RawHTML> : noticeText }
+					{ rawHTML ? <RawHTML>{ noticeText }</RawHTML> : noticeText }
 				</div>
 			</div>
 		);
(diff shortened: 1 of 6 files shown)


assets/components/src/notice/index.js
```diff
@@ -52,7 +52,9 @@ class Notice extends Component {
 		return (
 			<div className={ classes }>
 				{ noticeIcon }
-				<RawHTML className="newspack-notice__content">{ noticeText }</RawHTML>
+				<div className="newspack-notice__content">
+					{ !! rawHTML ? <RawHTML>{ noticeText }</RawHTML> : noticeText }
```
adekbadek: nitpick: in a ternary expression, the condition will be treated as a boolean anyway
```suggestion
					{ rawHTML ? <RawHTML>{ noticeText }</RawHTML> : noticeText }
```

Also, I wonder if the rendering condition is necessary - a string in `RawHTML` will not cause any issues 🤔

thomasguillot: Fixed in 194104e

PR #388, opened 2020-01-17
Ensure a single instance of react-router-dom is used

adekbadek:
### All Submissions:

`assets/components` is a small npm package on it's own, and as such it may contain a `node_modules` folder. In that case, if a node module was imported by a component in `assets/components` and one in, say, `assets/wizards`, they would end up with separate copies of the module. This caused issues with `react-router-dom`.
This was resolved by proxying `react-router-dom` as a module in `assets/components` - a single source of truth for this external module. 

1. run `npm i` in `assets/components` to create `node_modules` there
2. build the plugin JS (`npm run build`) and navigate to a wizard (e.g. Analytics)
3. everything should work as expected

### Other information:


adekbadek:
I agree on both - fixed in https://github.com/Automattic/newspack-plugin/pull/388/commits/a9fbf4d8d236b3e6e62a9cd40ba265cc152560e9

adekbadek:
Agreed again, with one change - the proxied module has to stay in `assets/components` - otherwise if the components package is used outside of the plugin, it would not have the proxied module/s included
https://github.com/Automattic/newspack-plugin/pull/388/commits/77ab7ae5d01992378d75f657b04120982ee706ff

PR #390, opened 2020-01-18
Child Theme Installer

jeffersonrabb:
### All Submissions:

This PR moves theme management into a dedicated class (`Theme_Manager`), and adds functionality to install and activate the five Newspack child themes. 

There is now a test for installing each child theme, so in theory running tests may be all you need. To test theme installation in PHP, add this snippet to `newspack.php`:

```
add_action(
	'init',
	function() {
		\Newspack\Theme_Manager::install_activate_theme( 'newspack-theme' );
	}
);
```

Then, refresh the Themes page (you will actually need to refresh twice because the newly installed theme won't show as installed until the next request). Try this with all child themes and verify each is installed an activated successfully:

`newspack-theme`, `newspack-scott`, `newspack-nelson`, `newspack-katharine`, `newspack-sacha`, `newspack-joseph`

### Other information:

### DIFF
--- a/includes/class-theme-manager.php
+++ b/includes/class-theme-manager.php
@@ -47,7 +47,7 @@ class Theme_Manager {
 			);
 		}
 
-		$theme_url = 'https://github.com/Automattic/newspack-theme/releases/latest/download/{$slug}.zip';
+		$theme_url = "https://github.com/Automattic/newspack-theme/releases/latest/download/{$slug}.zip";
 
 		$theme_object = wp_get_theme( $slug );
 		if ( ! $theme_object->exists() ) {
--- a/includes/class-theme-manager.php
+++ b/includes/class-theme-manager.php
@@ -25,9 +25,9 @@ class Theme_Manager {
 		'newspack-theme', // Default.
 		'newspack-scott', // Style Pack 1.
 		'newspack-nelson', // Style Pack 2.
-		// 'newspack-TK', Style Pack 3.
+		'newspack-katharine', // Style Pack 3.
 		'newspack-sacha', // Style Pack 4.
-		// 'newspack-TK', Style Pack 5.
+		'newspack-joseph', // Style Pack 5.
 	];
 
 	/**
--- a/includes/class-theme-manager.php
+++ b/includes/class-theme-manager.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * Newspack theme manager
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+use \WP_Error;
+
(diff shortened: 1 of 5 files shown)


includes/class-theme-manager.php
```diff
@@ -0,0 +1,81 @@
...
+ * Class for managing installation/activation of Newspack theme and child themes.
+ */
+class Theme_Manager {
+
+	/**
+	 * Array of valid Newspack theme slugs.
+	 *
+	 * @var array
+	 */
+	public static $theme_slugs = [
```
jeffersonrabb: @laurelfulford do you have names for the two remaining child themes yet (Style Packs 3 & 5)?

laurelfulford: Sorry for the late reply! They are, respectively, Newspack Katharine and Newspack Joseph.

includes/class-theme-manager.php
```diff
@@ -0,0 +1,99 @@
...
+			if ( is_wp_error( $success ) ) {
+				return $success;
+			} elseif ( $success ) {
+				// Make sure `-master` or `-1.0.1` etc. are not in the theme folder name.
+				// We just want the folder name to be the theme slug.
+				$theme_object    = $upgrader->theme_info();
+				$theme_folder    = $theme_object->get_template_directory();
+				$expected_folder = $theme_object->get_theme_root() . '/' . $slug;
+				if ( $theme_folder !== $expected_folder ) {
+					rename( $theme_folder, $expected_folder ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_rename
```
claudiulodro: `get_template_directory` will return the path to the parent theme if a child theme is being used. This causes a warning like this when trying to install a child theme that isn't installed:

```
Warning:  rename(/srv/www/newspack/public_html/wp-content/themes/newspack-theme,/srv/www/newspack/public_html/wp-content/themes/newspack-nelson): Is a directory
```

It's trying to rename the existing `newspack-nelson` directory to `newspack-theme`, and luckily throwing a warning and not succeeding. (`$theme_folder` will always be `'newspack-theme'` and `$expected_folder` will be the child theme slug)

The solution would be to [use `get_stylesheet_directory`](https://developer.wordpress.org/reference/functions/get_template_directory/#usage) instead of `get_template_directory` I think?

jeffersonrabb: Great catch, and the suggested fix seems to do the trick: https://github.com/Automattic/newspack-plugin/pull/390/commits/7b1d6f43b6273df6e357bb4f1ff7ff2ca0538eae

PR #404, opened 2020-01-23
WebPreview: layering fix; renderButton prop

adekbadek:
### All Submissions:

1. Renders the modal in a div appended to the body, this way there is more control over layering and the modal will always be on top of the admin UI. As it was before has worked fine in components preview, but when used in editor sidebar (https://github.com/Automattic/newspack-popups/pull/36) the modal was below the admin UI.
2. A new prop is introduced: `renderButton`. This lets the caller of this component render it's own button. This was used in https://github.com/Automattic/newspack-popups/pull/36 to disable the button until the post is saved. In general it enables total control over the `WebPreview`'s trigger styling (e.g. it might be a link, not a button).

1. Open components preview 
2. Click the "Preview Newspack Blog" button 
3. Observe that a model opens, and is placed towards the bottom of body element in DOM
4. Close the modal 
5. Click the "Preview Newspack Blog" link. The result should be the same as after clicking the button

### Other information:

### DIFF
--- a/assets/components/src/web-preview/index.js
+++ b/assets/components/src/web-preview/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies.
  */
-import { Component, Fragment } from '@wordpress/element';
+import { Component, Fragment, createPortal } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
@@ -30,11 +30,33 @@ class WebPreview extends Component {
 	state = {
 		device: 'desktop',
 		loaded: false,
-		previewVisibility: false,
+		isPreviewVisible: false,
+		modalDOMElement: null,
 	};
 
+	/**
+	 * Create a div and append it to the body, so that the modal is on the top layer
+	 */
+	componentDidMount() {
+		const modalDOMElement = document.createElement( 'div' );
+		document.body.appendChild( modalDOMElement );
+		this.setState( { modalDOMElement } );
+	}
+
+	/**
+	 * Cleanup
+	 */
+	componentWillUnmount() {
+		if ( this.state.modalDOMElement ) {
+			document.body.removeChild( this.state.modalDOMElement );
+		}
+	}
+
+	/**
+	 * Add or remove applicable body classes
+	 */
 	componentDidUpdate() {
-		if ( this.state.previewVisibility === true ) {
+		if ( this.state.isPreviewVisible === true ) {
 			document.body.classList.add( 'newspack-web-preview__open' );
 		} else {
 			document.body.classList.remove( 'newspack-web-preview__open' );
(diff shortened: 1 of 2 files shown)


assets/components/src/web-preview/index.js
```diff
@@ -30,95 +30,152 @@ class WebPreview extends Component {
...
-		previewVisibility: false,
+		isPreviewVisible: false,
+		modalDOMElement: null,
 	};
 
+	/**
+	 * Create a div and append it to the body, so that the modal is on the top layer
+	 */
+	componentDidMount() {
+		const modalDOMElement = document.createElement( 'div' );
```
jeffersonrabb: The [React Portals doc](https://reactjs.org/docs/portals.html) suggests creating the element in the constructor and adding to the DOM in `componentDidMount`. It seems like your approach is going to be effectively identical, but I'm wondering if the official approach should be adhered to here?

jeffersonrabb: Any thought that we should try to reuse the same `div` if it has already been created by a different instance of `WebPreview`? In certain situations we could end up with a large number of unused `div`s sitting around, which may be harmless but we should at least consider the possibility.

assets/components/src/web-preview/index.js
```diff
@@ -30,95 +30,152 @@ class WebPreview extends Component {
...
+			 * Inversion of control - let the caller render
+			 * the button that will trigger the modal
+			 */
+			renderButton,
+		} = this.props;
+
+		return (
+			<Fragment>
+				{ renderButton ? (
+					renderButton( { showPreview: this.showPreview } )
```
jeffersonrabb: Excellent idea, I wish I had done this the first time around 😄. Do we even need the `<Button>` fallback? There are currently no uses of `WebPreview` on master yet, so we could adjust unmerged code to reflect the approach where the button UI is always provided.

assets/components/src/web-preview/index.js
```diff
@@ -30,95 +30,152 @@ class WebPreview extends Component {
...
+		modalDOMElement: null,
 	};
 
+	/**
+	 * Create a div and append it to the body, so that the modal is on the top layer
+	 */
+	componentDidMount() {
+		const modalDOMElement = document.createElement( 'div' );
+		document.body.appendChild( modalDOMElement );
+		this.setState( { modalDOMElement } );
```
jeffersonrabb: Seems possibly weird to maintain the DOM element in state. Would this be opportunity for a React Ref? (It could be that Portal+Ref is not a good combination, just thinking aloud)

adekbadek: That's true! I've changed it so the DOM element is a class field (as it is in the React docs)

assets/components/src/web-preview/index.js
adekbadek: IMO a default/fallback is a good idea - we can assume the trigger will usually be the Button

assets/components/src/web-preview/index.js
adekbadek: Agreed :) Also that single identifiable div will be less cryptic in devtools:

image

adekbadek: I don't know what would be the practical difference. But when introducing use of single div ([comment below](https://github.com/Automattic/newspack-plugin/pull/404#discussion_r370108298)) I've noticed that if one instance should be aware of the div mounted by the previous one, it has to happen in `componentDidMount`

PR #431, opened 2020-02-07
Add Color Picker component

thomasguillot:
### All Submissions:

This PR introduces a new Newspack component: `ColorPicker`

It's based on a combination of `@wordpress/components`'s `ColorPalette` and `ColorIndicator` 

color-picker

1. Switch to this branch
2. Check the Components demo

### Other information:


assets/components/src/color-picker/index.js
```diff
@@ -0,0 +1,117 @@
...
+	state = {
+		color: '#ffffff',
+	};
+
+	/**
+	 * Render.
+	 */
+	render() {
+		const { hasDefaultColors, label, suggestedColors, ...otherProps } = this.props;
+		const { color } = this.state;
```
jeffersonrabb: For components generally the value should be a prop, not state, so that the container can set and respond to changes. To accomplish this:

- Drop all the state management.
- Make `color` a `prop`.
- Make `onChange` a prop too, and for `ColorPalette`'s onChange use: `onChange={ onChange }`
- In Components Demo, add three new properties to state: `color1` `color2` `color3`
- Then for each `ColorPicker` component add attributes like this: `color={ color1 } onChange={ color => this.setState( { color1: color } ) }` (using `color2` and `color3` for the 2nd and 3rd)
- I'd also add defaultProps to the Component so it doesn't error out if there is no `onChange` set:

```
ColorPicker.defaultProps = {
       onChange: () => null,
       color: '#ffffff',
};
```

assets/components/src/color-picker/index.js
```diff
@@ -0,0 +1,117 @@
...
+		const defaultColors = [
+			{
+				name: __( 'navy' ),
```
jeffersonrabb: The `name` of the colors don't seem to be displayed anywhere. Are they required and/or should they be displayed?

assets/components/src/color-picker/style.scss
```diff
@@ -0,0 +1,164 @@
+/**
```
jeffersonrabb: I'm seeing the circle around selected color slightly off-center:
 
Screen Shot 2020-02-08 at 6 25 50 PM

assets/components/src/color-picker/style.scss
```diff
@@ -0,0 +1,164 @@
...
+ * Color Picker
+ */
+
+@import '../../../shared/scss/colors';
+@import '~@wordpress/base-styles/colors';
+
+.newspack-color-picker {
```
jeffersonrabb: I know the Components Demo is a bit of a strange case, but the `ColorPicker` examples look a bit odd filling the full width. Do you think we should put a `max-width` on the component? Or do you prefer to leave that up to the container to determine?

assets/components/src/color-picker/style.scss
thomasguillot: Really strange one as this did not happen in the editor. Should be fixed in https://github.com/Automattic/newspack-plugin/pull/431/commits/f82101e4c23ca6353c29ebe8f3e7231e05e1b8bb

assets/components/src/color-picker/index.js
thomasguillot: It's for accessibility reason

assets/components/src/color-picker/style.scss
thomasguillot: I'd prefer not to add a max-width to the actual component but instead leave that up to the layout/container when it's being used.

assets/components/src/color-picker/index.js
thomasguillot: Not sure if this is what you meant: https://github.com/Automattic/newspack-plugin/pull/431/commits/e0840c4789519c4042ab7ed02bd29c6a41d385c9 😅

thomasguillot:
> Left one last note about default props, then we should be good to go.

@jeffersonrabb changes made in https://github.com/Automattic/newspack-plugin/pull/431/commits/29792e006b9f4ac9ab7ce6b9a5a6f81065dcce01

PR #458, opened 2020-02-28
WP.com Stripe Endpoints

jeffersonrabb:
Use WPCOM endpoints to protect Stripe keys from exposure.

### All Submissions:

Moves endpoints for Stripe API activity to WP.com, to avoid risk of API key exposure. Complete testing instructions in `D39588-code`.  

### Other information:

### DIFF
--- a/includes/wizards/class-payment-wizard.php
+++ b/includes/wizards/class-payment-wizard.php
@@ -7,9 +7,9 @@
 
 namespace Newspack;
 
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
> It can be as part of this PR or as a follow-up PR, but we should also remove the stripe library composer installs and update the release script if needed since we don't need anything from the `vendor` directory in releases any more.

Yes good call, I just forgot. I'll do it in a separate PR to avoid triggering a new review.

Issue #446, opened 2020-02-26
Starter Content generation erroring out


adekbadek:
**Describe the bug**

On the final step of the setup wizard, user has the possibility to generate some starter content. The posts are not created though, because the requests to lipsum.com made here time out:

https://github.com/Automattic/newspack-plugin/blob/fb98e3a7f6c35b4b28029182e7c2ef9d32c2308f/includes/class-starter-content.php#L285-L298

I've reproduced it locally and on a fresh JN site. 
The solution might be to use another service (like loripsum.net) or a PHP package to generate the strings locally. 
Note that all of this has implications on E2E testing, which will need a deterministic placeholder content, but that's another subject.

**To Reproduce**
1. Visit `/wp-admin/admin.php?page=newspack-setup-wizard#/starter-content` page
2. Click on "Install Starter Content"
3. Observe the errors in the console
4. Observe that the posts are not created

**Expected behavior**

The starter content should be generated.

PR #459, opened 2020-02-28
Fix Starter Content Lorem Ipsum Generation

jeffersonrabb:
### All Submissions:

The starter content feature recently began failing because the API we use to generate lorem ipsum (https://www.lipsum.com/) often times out. This PR switches to https://loripsum.net/ (thank you @adekbadek for the suggestion!)

Closes https://github.com/Automattic/newspack-plugin/issues/446

1. On test instance, delete all posts.
2. Navigate to Settings->Newspack and click "Setup" link.
3. Step through Set Up Wizard. At the end, click to create Starter Content.
4. After Wizard completes, check that ~40 posts were created with appropriate lorem ipsum title and body text.

### Other information:

### DIFF
--- a/includes/wizards/class-setup-wizard.php
+++ b/includes/wizards/class-setup-wizard.php
@@ -219,7 +219,7 @@ class Setup_Wizard extends Wizard {
 		$theme_mods = get_theme_mods();
 
 		foreach ( $theme_mods as $key => &$theme_mod ) {
-			if ( in_array( $key, $this->$media_theme_mods ) ) {
+			if ( in_array( $key, $this->media_theme_mods ) ) {
 				$attachment = wp_get_attachment_image_src( $theme_mod )[0];
 				$theme_mod  = [
 					'url' => is_array( $attachment ) ? $attachment[0] : null,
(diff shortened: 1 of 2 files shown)


includes/wizards/class-setup-wizard.php
```diff
@@ -219,7 +219,7 @@ public function api_retrieve_theme() {
 		$theme_mods = get_theme_mods();
 
 		foreach ( $theme_mods as $key => &$theme_mod ) {
-			if ( in_array( $key, $this->$media_theme_mods ) ) {
+			if ( in_array( $key, $this->media_theme_mods ) ) {
```
claudiulodro: Good catch, I noticed errors in the logs today from this. 👍

jeffersonrabb:
> Lorem ipsum dolor sit amet

Ah interesting - I didn't notice that. I think we should land this nonetheless so we have a working release, and I'll immediately explore the composer package which seems like a better approach.

adekbadek:
@jeffersonrabb with e2e testing in mind, we should have a deterministic starter content, so that we could do visual regression testing on themes (with starter content). 
Some variety could be introduced by treating each iteration index as a deterministic "seed" for the dummy content (e.g. `create_post(0)` -> `Lorem ipsum…`, `create_post(1)` -> `Other ipsum…`).

PR #463, opened 2020-02-29
Use php-loremipsum For Starter Content Text

jeffersonrabb:
### All Submissions:

The starter content feature generates lorem ipsum text using third party APIs. This approach is slow and failure prone. In this branch the [php-loremipsum](https://github.com/joshtronic/php-loremipsum) library is installed and used instead. 

From the php-loremipsum docs:

> The first string generated will always start with the traditional "Lorem ipsum dolor sit amet, consectetur adipiscing elit". Subsequent strings may contain those words but will not explicitly start with them.

This is problematic because it means all titles will be identical, and the point of starter content is to display a random, realistic-looking body of content. To beat this, this branch strips out the identical opening bits of generated lorem ipsum to make each string unique. In the case of paragraphs, the first one is removed which guarantees the opening will be unique.

1. Switch to this branch, run `composer update`.
2. Go to `/wp-admin/admin.php?page=newspack-setup-wizard#/starter-content`, and click Install Starter Content.
3. Verify the wizard completes successfully.
4. Verify 40 new posts are created, each with unique Latin titles.
5. Verify the posts contain 5 paragraphs of lorem ipsum, which appears to be different from post to post.

### Other information:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
 	},
 	"require"    : {
 			"composer/installers": "~1.6",
-			"stripe/stripe-php": "^7.25"
+			"stripe/stripe-php": "^7.25",
+		"joshtronic/php-loremipsum": "^1.0"
 	},
 	"require-dev": {
 		"brainmaestro/composer-git-hooks": "^2.6",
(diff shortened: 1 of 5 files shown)


jeffersonrabb:
> We can probably remove the `stripe` library from `composer.json` since it's not used for anything any more.

I did that in https://github.com/Automattic/newspack-plugin/pull/464, specifically https://github.com/Automattic/newspack-plugin/pull/464/commits/a56da1dee24c60bb4055bbff583cd3e1b8f7bdb4

PR #507, opened 2020-04-06
Allow Composer path to be overridden

davisshaver:
If the Newspack plugin is managed as a composer dependency, the Composer path might not align with the main plugin path. This pull requests maintains current behavior while adding a way to specify which directory is used for the Composer autoload file.


adekbadek (approved):
Thanks and sorry for the wait!

PR #524, opened 2020-05-05
Analytics events management

claudiulodro:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds a Google Analytics events management framework and wizard. We can refine it further, but this should be a good prototype right now. If the idea/architecture/ux looks good, we can refine it further.

*Disclaimer: Haven't actually tested much with a real site to verify analytics get recorded correctly. I'll do QA before switching this PR out of draft. The architecture has been my main concern so far.*

### How to test the changes in this Pull Request:

1. You should see a new Events tab in the Analytics wizard.

Screen Shot 2020-05-05 at 2 35 54 PM

I'm not a huge fan of those category labels, but that's what the NCS doc said to use for the names. Might be good to put something more readable for the event categories.

2. To simulate Site Kit AMP analytics, go to an AMP post and add the following snippet somewhere:

```

add_action( 'wp_footer', function(){
	$gtag_amp_opt = array(
		'vars'            => array(
			'gtag_id' => '123456',
			'config'  => array(
				'123456' => array(
					'groups' => 'default',
					'linker' => array(
						'domains' => array( 'https://newspack.test' ),
					),
				),
			),
		),
		'optoutElementId' => '__gaOptOutExtension',
	);

	$gtag_amp_opt_filtered = apply_filters( 'googlesitekit_amp_gtag_opt', $gtag_amp_opt );
	?>
	<amp-analytics type="gtag" data-credentials="include">
		<script type="application/json">
			<?php echo wp_json_encode( $gtag_amp_opt_filtered ); ?>
		</script>
	</amp-analytics>
	<?php

});

```

3. In the page source you should see something like:

```
<amp-analytics type="gtag" data-credentials="include">
   <script type="application/json">
[...]

PR #531, opened 2020-05-07
Analytics events framework

claudiulodro:
### All Submissions:

Supersedes #524 

This PR is like #524 but without the UI parts (which are probably unnecessary anyways). 
 
**Discussion**: Do we want to make these events [non-interaction events](https://support.google.com/analytics/answer/1033068#NonInteractionEvents)? I think this is what was causing issues with bounce rates on some sites, because it doesn't count as a bounce in GA if an event is sent. E.g. if a user scrolls down 25% of the way through an article then leaves or gets shown a popup then leaves, my understanding is it won't count it as a bounce since an interaction event fired. Thoughts?

1. Connect Site Kit and analytics.
2. In AMP mode, scroll down an article and click on social buttons. You should see events logged in GA.
3. In non-AMP mode, scroll down an article and click on social buttons. You should see events logged in GA. (In my testing I don't though for some reason. Everything is firing and sending the event correctly as far as I can tell, though. Even manual calls to `ga( 'send', 'foo', 'bar', 'baz' );` in the dev console don't show any events for me)

Screen Shot 2020-05-07 at 11 59 07 AM

### Other information:


adekbadek:
Added some changes:
1. `value` to scroll events, as per [NTG spec](https://newsinitiative.withgoogle.com/training/states/ntg/assets/ntg-playbook.pdf#page=245) (0a365d6f4b5d317514efb3d9523e5c78e0c9699c)
2. fixed non-AMP event reporting – Site Kit uses `gtag` instead of `analytics.js` (3a6981d6c88bf5d5eed60ba504a3e064c0ff66c7)
3. fixed scroll event reporting on non-AMP (3b095b9e8eabb71ef1d782a95c27fcc271d82a27)

adekbadek:
> Do we want to make these events non-interaction events?

I think so, but NCI folks will probably have a definitive answer.

claudiulodro:
Thanks for jumping in, @adekbadek. Your changes look great!

includes/class-analytics.php
```diff
@@ -0,0 +1,245 @@
...
+	 * @param array $event Event info. See 'get_events'.
+	 */
+	protected static function output_js_click_event( $event ) {
+		?>
+		<script>
+			( function() {
+				const elementSelector = '<?php echo esc_attr( $event['element'] ); ?>';
+				const elements        = document.querySelectorAll( elementSelector );
+
+				for ( const element of elements ) {
```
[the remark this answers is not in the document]
adekbadek: There's this trick for converting a `NodeList` to a plan old array:
```
const elements = [...document.querySelectorAll( elementSelector )];
```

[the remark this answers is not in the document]
adekbadek: There's this trick for converting a `NodeList` to an array:
```js
const elements = Array.prototype.slice.call( document.querySelectorAll( elementSelector ) );
```

includes/class-analytics.php
```diff
@@ -0,0 +1,245 @@
...
+	/**
+	 * Inject event listeners on non-AMP pages.
+	 */
+	public static function inject_non_amp_events() {
+		if ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() ) {
+			return;
+		}
+
+		$sitekit_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'google-site-kit' );
+		if ( false && ! $sitekit_manager->is_module_active( 'analytics' ) ) {
```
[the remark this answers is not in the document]
claudiulodro: Good catch! I forgot that short-circuit in there. https://github.com/Automattic/newspack-plugin/pull/531/commits/6cc0ed7807ebf01c99667e1f7bb5021042fbf8f5

includes/class-analytics.php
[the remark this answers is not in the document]
claudiulodro: Good call and good suggestions! https://github.com/Automattic/newspack-plugin/pull/531/commits/20eab910ada47fc6d38ed4b575f32321731d2e4c

includes/class-analytics.php
```diff
@@ -0,0 +1,245 @@
...
+			],
+			[
+				'id'             => 'socialShareClickedTwitter',
+				'on'             => 'click',
+				'element'        => 'a.share-twitter',
+				'amp_element'    => 'amp-social-share[type="twitter"]',
+				'event_name'     => 'social share',
+				'event_label'    => 'twitter',
+				'event_category' => 'NTG social',
+			],
```
adekbadek, line 58: cross-referencing [`<amp-social-share>`](https://amp.dev/documentation/components/amp-social-share/)'s `type`s and [NTG guidelines](https://newsinitiative.withgoogle.com/training/states/ntg/assets/ntg-playbook.pdf#page=245), we're missing `whatsapp` and `linkedin`. Is there a reason there are only `facebook` and `twitter` here?

includes/class-analytics.php
```diff
@@ -0,0 +1,245 @@
...
+					if ( ( ( scrollPos / documentHeight ) * 100 ) >= scrollPercent ) {
+						eventSent = true;
+						gtag(
+							'event',
+							'<?php echo esc_attr( $event['event_name'] ); ?>',
+							{
+								event_category: '<?php echo esc_attr( $event['event_category'] ); ?>',
+								event_label: '<?php echo esc_attr( $event['event_label'] ); ?>',
+								value: scrollPercent
+							}
```
adekbadek, line 233: `'non_interaction': true` should be [added here](https://developers.google.com/analytics/devguides/collection/gtagjs/events#send_non-interaction_events), as we agreed to send scroll events as [non-interaction events](https://support.google.com/analytics/answer/1033068#NonInteractionEvents). One thing to check is whether the `amp-analytics` events will be sent as non-interaction events.

includes/class-analytics.php
claudiulodro: This was intended to be the framework rather than a full implementation of the spec, so I just did a few elements as examples but I can add those.

includes/class-analytics.php
claudiulodro: Yes I had it on my list but forgot. Good catch!

claudiulodro: Added support in https://github.com/Automattic/newspack-plugin/pull/531/commits/c0763b1c3079771315ffe80d98dd97caa78150c0. By default events will be non-interactive with ability to mark them as interactive, which seems a safe default.

includes/class-analytics.php
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/531/commits/5835e54b97a7bf369ca28be303fee3094ea139cb 

Note: The Jetpack WhatsApp share button isn't AMP-compatible so it won't show up in AMP, but when it's compatible we will already be prepared. :)

adekbadek:
3. "I assume reaching the bottom of the content area should count as 100% since everything below isn't content." – good point. Since `amp-analytics`'s [scroll trigger](https://amp.dev/documentation/components/amp-analytics/#scroll-trigger) does not support a selector, I guess we'd have to compute what percentage of the page's height does article content constitute.

jeffersonrabb:
> 1. Regarding the non-interaction logic, wouldn't a `click` event always be an interactive event and a `scroll` always be a non-interactive? I'm wondering if this logic could be entirely based on the `on` key without needing the added specificity of the `interactive` key?

This is done in https://github.com/Automattic/newspack-plugin/pull/531/commits/90655bb9c2ab9c8e152324076b027292cda8c9eb

jeffersonrabb:
> 1. Since `amp-analytics`'s [scroll trigger](https://amp.dev/documentation/components/amp-analytics/#scroll-trigger) does not support a selector, I guess we'd have to compute what percentage of the page's height does article content constitute. Does that make sense?

Discussed with @adamsilverstein who is looking into this as a potential AMP improvement. This shouldn't hold back this PR, though.

claudiulodro:
Latest version looks good. I'm merging, and then let's address any lingering scroll tweaks or things like that in follow-up PRs if needed.

IMO if there isn't a good solution for the 100% read trigger, we can always just trigger it at like 90%. It might be a little off, but I don't think there's a meaningful difference between users that read 95% of an article and users that read 100% of an article. :)

adamsilverstein:
Note that amp-analytics events should now support a trigger on an arbitrary element, see https://github.com/ampproject/amphtml/pull/29172

adekbadek:
Yep, planned as #670

PR #534, opened 2020-05-15
Add tickets list in Support wizard

adekbadek:
### All Submissions:

Add a tickets list view to the Support wizard.

image

1. Set up a [WPCOM app for OAuth](https://developer.wordpress.com/apps)
2. Ensure support-related env variables are present (as per README.md)
3. Visit the Support wizard in the Newspack plugin
4. Authenticate, if necessary, with a WPCOM account 
5. Observe a list of open and closed tickets tickets, or a message informing about no tickets

### Other information:

### DIFF
--- a/assets/wizards/support/components/withWPCOMAuth.js
+++ b/assets/wizards/support/components/withWPCOMAuth.js
@@ -1,0 +1,50 @@
+/* global newspack_support_data */
+
+/**
+ * WordPress dependencies
+ */
+import { Fragment, Component } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { Notice, Button } from '../../../components/src';
+import { saveReturnPath } from '../utils';
+
+const { WPCOM_ACCESS_TOKEN } = newspack_support_data;
+
+const withWPCOMAuth = WrappedComponent => {
+	return class WithWPCOMAuth extends Component {
+		componentDidMount() {
+			if ( ! WPCOM_ACCESS_TOKEN ) {
+				saveReturnPath();
+			}
+		}
+		render() {
+			return (
+				<Fragment>
+					{ WPCOM_ACCESS_TOKEN ? (
+						<WrappedComponent token={ WPCOM_ACCESS_TOKEN } { ...this.props } />
+					) : (
+						<Fragment>
+							<Notice
+								noticeText={ __(
+									'Click the button below to authenticate using a WordPress.com account.',
+									'newspack'
+								) }
+							/>
+							<div className="newspack-buttons-card">
+								<Button href={ newspack_support_data.WPCOM_AUTH_URL } isPrimary>
+									{ __( 'Authenticate', 'newspack' ) }
+								</Button>
+							</div>
+						</Fragment>
+					) }
+				</Fragment>
+			);
+		}
+	};
+};
+
+export default withWPCOMAuth;
--- a/assets/wizards/support/components/withWPCOMAuth.js
+++ b/assets/wizards/support/components/withWPCOMAuth.js
@@ -19,27 +19,28 @@ const withWPCOMAuth = WrappedComponent => {
 	return class WithWPCOMAuth extends Component {
 		state = {
 			isInFlight: false,
+			shouldAuthenticate: true,
 		};
 		componentDidMount() {
-			if ( ! WPCOM_ACCESS_TOKEN ) {
-				saveReturnPath();
-			}
-
-			this.setState( { isInFlight: true } );
-			apiFetch( {
-				path: `/newspack/v1/wizard/newspack-support-wizard/validate-access-token`,
-			} )
-				.then( () => {
-					this.setState( { isInFlight: false } );
+			if ( WPCOM_ACCESS_TOKEN ) {
+				this.setState( { isInFlight: true } );
+				apiFetch( {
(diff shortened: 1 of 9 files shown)


adekbadek:
Sorry for that, some errors were not handled. Should be fixed by a9caf46169a180bbdd447be5e64dc12c510ee6b8

philipjohn:
The tickets this pulls in are based on the current user. Will it only retrieve tickets where the current user is the requester? Or will it also include any tickets the user is CC'd on?

I ask because I can see multiple people from the same publishers wanting to see all the tickets. That makes authentication more difficult, unless we can include all CC'd tickets too.

adekbadek:
> I now get the same error but on a different line:

@philipjohn I've added error handling there in 7becbb2911cbb54c2cee1bbace3caefc81df86d9 – you should now see the error in the frontend. What does it say?

> Will it only retrieve tickets where the current user is the requester? 

Exactly. But I'll explore the possibility of adding an "enhanced support history" or something endpoint to WPCOM.

philipjohn:
Okay, that's working now, but I don't see my tickets. I just have the "no open tickets found". The test ticket I created was initially set to "New" which makes sense, so I escalated it to Newspack and saved it as "Open" but it still doesn't show in my tickets view.

The email address on the ticket is admin@local.test which matches the user account I'm logged in with, and I have chat authenticated successfully (but I supposed that's connected to my a8c email).

adekbadek:
Yeah, you should change that WP user email. The ticket is submitted as the current user email, but tickets are retrieved as the WPCOM authenticated user.

jeffersonrabb:
The no tickets state has an errant "0" in it:

Screen Shot 2020-05-19 at 6 23 57 AM

jeffersonrabb:
A few notes on the ticket display:

- Can we show the total number of replies, and the date of the most recent one?
- If user action is required, can that be flagged? Possibly brought to the top?
- Can the ticket name be a link to the ticket in Zendesk?
- Show the ticket status? If the only options are `New`, `Open`, and `Closed` then the handling of Closed tickets already covers this, but if there are more status options it might be good to display.
- Is it possible to show an Excerpt, or Description of the ticket?
- Following on @philipjohn's comment about showing all tickets associated with the site, we would want to list the user who filed it.
- For a later iteration, search functionality would be great. Once we start using Zendesk for pre-launch tasks, the list could quickly become overwhelming, and a search feature could make it manageable.
- Also for later, we may need pagination.

adekbadek:
@jeffersonrabb 

> The no tickets state has an errant "0" in it:

That should be fixed in 128fb754a72b6f5ca872cc7a44cef545116e088e

> A few notes on the ticket display

The current implementation displays all data available from the WPCOM's `/support-history` endpoint. Changing that endpoint (or creating a new one) might take a while, so I'd propose to file those changes as an issue – WDYT?

jeffersonrabb:
> Changing that endpoint (or creating a new one) might take a while, so I'd propose to file those changes as an issue – WDYT?

Absolutely. Let's get this PR done as-is and address the rest later. Could you add the changes from https://github.com/Automattic/newspack-plugin/pull/520 into this PR? It's difficult to test the new HOC without token invalidation working.

adekbadek:
>  Could you add the changes from #520

Done in f323038282b3d363ae9ec06b26d682307336a090

adekbadek:
@jeffersonrabb I've opened #537 with your suggestions, besides this one:

> Can the ticket name be a link to the ticket in Zendesk?

The customer does not have access to Zendesk, so that would not be very useful. Or am I missing something?

jeffersonrabb:
> The customer does not have access to Zendesk, so that would not be very useful. Or am I missing something?

No you're right. In testing this PR I received confirmation emails from Zendesk which included links to the tickets. I missed that these are in the links only because I'm an agent, which makes perfect sense.

adekbadek:
> Something from #520 didn't make it through — after mangling the key or disconnecting the WP.com app I get this, with no option to reconnect

Thanks for catching that! Fixed in 4dde4c10ad39b82b8a418bc1fe0286eec3402d22

> the Chat screen throws a large number of this JS error

It will throw this error if you change wizard tab after initializing chat. This is a bug in `happychat-client` – it does not clean up event listeners after closing :/

adekbadek:
@philipjohn having realised that WPCOM API does return the Zendesk status, I've updated the UI to show "Action required" tickets separately in 73c8e647b47c6d3e596b2a29033d221374da15d1

image

philipjohn:
> The ticket is submitted as the current user email, but tickets are retrieved as the WPCOM authenticated user.

I'm still not able to see my tickets. Here's what I did:

1. On my local test site, newspack.test I logged in as admin (email admin@local.test)
1. Created a ticket
1. Checked the ticket was open (it's no. 3010865)
1. Checked the ticket view - no tickets
1. Changed the user account email address to the one associated with my WP.com account (`philipjohna8c`)
1. Refreshed the ticket view - no tickets
1. Changed the requester on the ticket to my A8c email address to exactly match both my local WP user account and the WP.com user account
1. Refreshed the ticket view - no tickets

I'm worried that if there is a mismatch between the WP user account email and the WP.com account email folks aren't going to see their tickets, and I know for sure that there is inconsistency with some publishers.

adekbadek:
Got it. I've changed it so that the authenticated WPCOM user is used as the ticket requester in 22119089d1c1edc15a7de696d5bc671f93295e9b

PR #549, opened 2020-06-05
SalesForce integration

dkoo:
### All Submissions:

Adds a Salesforce config view to the Reader Revenue wizard which allows admin users to connect a Salesforce org via the SF [OAuth 2.0 web server flow](https://help.salesforce.com/articleView?id=remoteaccess_oauth_web_server_flow.htm&type=5).

When the Newspack site is successfully connected to a Salesforce org, a WooCommerce webhook is generated which will trigger a data sync every time a donation order is created. This webhook is deleted if the Salesforce config is reset.

When an order is created, the webhook fires a request which includes order and billing details. This data is then synced to the connected Salesforce org via the Salesforce REST API. Lead records are looked up by email address and if there's an existing Lead record in SF, donor details are added to that record. If there isn't an existing Lead record, one is created. More details on the SalesForce API here: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_what_is_rest_api.htm

Screen Shot 2020-06-05 at 3 35 18 PM

At this stage, this is a proof-of-concept implementation, so I'd like to request feedback on the overall approach as well as the UI and where in the wizard it lives, etc. before I consider it ready for a full review/merge. We may want to split this out into an entirely new wizard (Customer Retention?), but for now, the existing Reader Revenue wizard seemed like the best home for this.

I have not tested this in browsers other than Chrome so far, nor written automated tests for this functionality. So far I've only tested it in a fully setup Newspack site, not as part of a new site setup.

1. Pull and build this branch to a local Newspack site environment. Run `npm install && npm run dev` to rebuild JS.
2. Visit the Reader Revenue wizard to find the Configure SalesForce page.
3. Sign up for a new SalesForce Developer Org: https://developer.salesforce.com/signup
4. Once your org has been created, follow these instructions to create a new Connected App: https://help.salesforce.com/articleView?id=connected_app_create.htm
  - In short, go to Setup > Apps > App Manager > New Connected App and fill out the form, making sure to enable OAuth settings with the following capabilities: api, refresh_token, offline_access, web
  - Make sure to paste the URL for your Configure SalesForce page into the "Redirect URI" field. This whitelists the URL for use in the OAuth flow, otherwise SalesForce will block access to the login/allow screens in the next few steps.
5. Note the Consumer Key and Consumer Secret strings that are generated in the created Connected App page.
6. In the Configure SalesForce page, enter your Connected App's Consumer Key and Secret into the text fields. The "Connect" button should become available after both fields contain values.
7. After clicking "Connect," you should be redirected to a SalesForce login page (if not already logged in with the same browser), or to an "Allow Access?" page. Click Allow to grant access.
8. You will be redirected back to the wizard page on the Newspack site with an authorization code appended to the URL. The app should automatically fetch API access and refresh tokens using the auth code, then clear the `code` param from the URL.
9. In theory, the site can now use the `access_token` to make requests to the SalesForce org via REST API, or use the `refresh_token` to fetch new `access_token` values.

### Other information:

### DIFF
--- a/assets/wizards/readerRevenue/index.js
+++ b/assets/wizards/readerRevenue/index.js
@@ -37,7 +37,7 @@ class ReaderRevenueWizard extends Component {
 				locationData: {},
 				stripeData: {},
 				donationData: {},
-				salesforceData: {}
+				salesforceData: {},
 			},
 		};
 	}
@@ -109,7 +109,7 @@ class ReaderRevenueWizard extends Component {
 		countryStateFields: data.country_state_fields,
 		currencyFields: data.currency_fields,
 		donationPage: data.donation_page,
-		salesforceData: data.salesforce_settings
+		salesforceData: data.salesforce_settings,
 	} );
 
 	/**
@@ -135,7 +135,7 @@ class ReaderRevenueWizard extends Component {
 			stripeData,
 			donationData,
 			donationPage,
-			salesforceData
+			salesforceData,
 		} = data;
 
 		const tabbedNavigation = [
@@ -260,9 +260,10 @@ class ReaderRevenueWizard extends Component {
 												client_secret: '',
 												access_token: '',
 												refresh_token: '',
-												instance_url: ''
+												instance_url: '',
 											};
-											this.setState( { data: {  ...data, salesforceData: defaultSettings } } );
+
+											this.setState( { data: { ...data, salesforceData: defaultSettings } } );
 											return this.update( 'salesforce', defaultSettings );
 										}
 
@@ -270,10 +271,19 @@ class ReaderRevenueWizard extends Component {
 
 										if ( client_id && client_secret ) {
 											this.update( 'salesforce', salesforceData );
-											window.location.assign( `https://login.salesforce.com/services/oauth2/authorize?response_type=code&client_id=${ encodeURIComponent( client_id ) }&client_secret=${ encodeURIComponent( client_secret ) }&redirect_uri=${ encodeURIComponent( window.location.href ) }` );
+											window.location.assign(
+												`https://login.salesforce.com/services/oauth2/authorize?response_type=code&client_id=${ encodeURIComponent(
+													client_id
+												) }&client_secret=${ encodeURIComponent(
+													client_secret
+												) }&redirect_uri=${ encodeURIComponent( window.location.href ) }`
+											);
 										}
 									} }
-									buttonDisabled={ ! salesforceIsConnected && ( ! salesforceData.client_id || ! salesforceData.client_secret ) }
+									buttonDisabled={
+										! salesforceIsConnected &&
+										( ! salesforceData.client_id || ! salesforceData.client_secret )
+									}
 									onChange={ _salesforceData =>
 										this.setState( { data: { ...data, salesforceData: _salesforceData } } )
 									}
--- a/assets/wizards/readerRevenue/index.js
+++ b/assets/wizards/readerRevenue/index.js
@@ -15,13 +15,14 @@ import { __ } from '@wordpress/i18n';
  */
 import HeaderIcon from '@material-ui/icons/AccountBalanceWallet';
 import LoyaltyIcon from '@material-ui/icons/Loyalty';
+import GroupAddIcon from '@material-ui/icons/GroupAdd';
 
 /**
  * Internal dependencies.
  */
 import { withWizard } from '../../components/src';
 import Router from '../../components/src/proxied-imports/router';
-import { Donation, LocationSetup, StripeSetup, RevenueMain } from './views';
+import { Donation, LocationSetup, StripeSetup, RevenueMain, SalesForce } from './views';
 
 const { HashRouter, Redirect, Route, Switch } = Router;
 
@@ -36,6 +37,7 @@ class ReaderRevenueWizard extends Component {
 				locationData: {},
 				stripeData: {},
 				donationData: {},
+				salesforceData: {}
 			},
 		};
 	}
@@ -52,7 +54,6 @@ class ReaderRevenueWizard extends Component {
 		const { setError, wizardApiFetch } = this.props;
 		return wizardApiFetch( { path: '/newspack/v1/wizard/newspack-reader-revenue-wizard' } )
 			.then( data => {
-				console.log( this.parseData( data ) );
 				return new Promise( resolve => {
 					this.setState(
 						{
@@ -108,6 +109,7 @@ class ReaderRevenueWizard extends Component {
 		countryStateFields: data.country_state_fields,
 		currencyFields: data.currency_fields,
 		donationPage: data.donation_page,
+		salesforceData: data.salesforce_settings
 	} );
 
 	/**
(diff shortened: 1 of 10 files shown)


assets/wizards/readerRevenue/views/salesforce/index.js
```diff
@@ -0,0 +1,161 @@
...
+						)
+					}
+
+					{
+						error && <Notice noticeText={ __( error ) } isWarning />
+					}
+
+					{
+						isConnected ?
+						<Notice noticeText={ __( 'Your site is connected to SalesForce.' ) } isSuccess />
```
adekbadek, line 118: Note: should be `Salesforce`, with a minuscule `f` ;)

assets/wizards/readerRevenue/views/salesforce/index.js
```diff
@@ -0,0 +1,161 @@
...
+						value={ client_id }
+						disabled={ isConnected }
+						onChange={ value => {
+							onChange( { ...data, client_id: value } );
+						} }
+					/>
+					<TextControl
+						label={ __( ( isConnected ? __( 'Your' ) : __( 'Enter your' ) ) +' SalesForce Consumer Secret' ) }
+						value={ client_secret }
+						disabled={ isConnected }
```
adekbadek, line 145: Since these fields are disabled when connected, how would a user deal with a re-connection, in case something goes wrong and they need to reset the access token?

dkoo: @adekbadek The button action changes to a reset action once there's an established connection. Right now the label is "Clear Connection" but maybe it should be "Reset Connection" for clarity. Clicking that button clears all tokens and credentials from the WP side and lets you enter credentials and start the OAuth flow again. I suppose ideally there should also be a "refresh" action which lets you keep the Consumer Key/Secret and request new tokens, but I'll need to think about the UI for that.

dkoo: I changed this button to "Reset" so it's clearer what happens when you click it.

dkoo:
@jeffersonrabb @adekbadek Another update: after running into a lot of trouble when testing the OAuth flow with Jeff yesterday, I implemented some additional validation checks and error messaging to handle the cases we ran into. Changes in the last few commits include:

* If you try to visit the login page at Salesforce with invalid `client_id` or `client_secret` credentials, Salesforce blocks access with an ugly plain-text error page. Now, before we redirect to that page, we test the URL with a silent GET request and only redirect if it responds with a `200` status code, meaning that the URL will resolve if we redirect to it. If the status code is anything other than `200`, we still save the credentials as entered, but show an error message to the user instead of trying to redirect to a dead end.
* After getting an OAuth authorization code from Salesforce, we need to execute another POST request to get the access and refresh tokens that actually let us interact with the REST API. If this request fails for whatever reason, we now show an error message to the user informing them that they need to try connecting again.
* I tested the flow with a brand-new Salesforce org, and found that it took about an hour before I could connect, so I updated the help copy in the wizard to reflect this.
* One thing that I couldn't quite resolve is that while a new org is still propagating, some requests to the login page with correct credentials would resolve while others wouldn't, so at times the validation test would pass even though redirecting would still fail. This happened with both a brand-new org, and a new Connected App within the same org about an hour later, although the new org only took about 20 minutes before I could connect instead of a full hour.
* I also tested with the callback URL set to only the main domain of my site, and confirmed that it doesn't work—it needs to match the wizard page's exact URL. This was just to confirm that we can't simplify the instructions to the user.

dkoo:
@jeffersonrabb @adekbadek I fixed one more bug related to insufficient admin user permissions when the webhook is delivered. This seems to be working now, so moving this PR to review.

assets/wizards/readerRevenue/index.js
```diff
@@ -119,11 +122,70 @@ class ReaderRevenueWizard extends Component {
...
+
+		// Otherwise, attempt to establish a connection with Salesforce.
+		const { client_id, client_secret } = salesforceData;
+
+		if ( client_id && client_secret ) {
+			const loginUrl = `https://login.salesforce.com/services/oauth2/authorize?response_type=code&client_id=${ encodeURIComponent(
+				client_id
+			) }&client_secret=${ encodeURIComponent( client_secret ) }&redirect_uri=${ encodeURIComponent(
+				window.location.href
+			) }`;
```
adekbadek, line 156: tip: `addQueryArgs` from `@wordpress/url` can be user to create URLs without the need for string interpolation and encoding.

includes/class-salesforce.php
```diff
@@ -0,0 +1,365 @@
...
+		// Get webhook id.
+		$webhook_id = get_option( self::SALESFORCE_WEBHOOK_ID, 0 );
+
+		// If we're establishing a new connection and don't have an existing webhook, create a new one.
+		if ( empty( $webhook_id ) && ! empty( $args['refresh_token'] ) ) {
+			self::create_webhook();
+		}
+
+		// If we're resetting, let's delete the existing webhook, if it exists.
+		if ( ! empty( $webhook_id ) && '' === $args['refresh_token'] ) {
```
adekbadek, line 102: When the webhook is **triggered**, I'm getting `Undefined index: refresh_token` error on this line.

assets/wizards/readerRevenue/index.js
dkoo: Didn't know that, thanks for the hint!

includes/class-salesforce.php
dkoo: Good catch—I now check `isSet( $args['refresh_token' ] )` first, so I think that should suppress the error.

dkoo:
> In the wizard, there's no indication that the connection's broken – it'd be nice to validate credentials on component mound, so the user can see the status of the connection.

@adekbadek I added a check after we've established that a connection has been made to verify that the `refresh_token` is still valid. If it isn't, we show an error message stating that we can't validate the connection. To test:

* After successfully connecting the Newspack site, revoke the connection in the SF UI or delete the connected app.
* Refresh the wizard page in the Newspack site. You should see an error message like the following:

Screen Shot 2020-06-18 at 2 12 51 PM

adekbadek:
@dkoo I confirm it works as described 👍

dkoo:
@adekbadek @jeffersonrabb I tested and confirmed that the connection flow and webhook data sync works with the Nonprofit Success Pack as well as a "standard" Salesforce account. If you'd like to test, too, you can sign up for a 30-day free trial of the NPSP edition of Salesforce here: https://www.salesforce.org/nonprofit/get-started/

One more thing I wanted to call attention to is that currently, I'm just storing donation data as a JSON string in the Lead's "description" field: https://github.com/Automattic/newspack-plugin/pull/549/files#diff-c0559cb9ce53cfcc0923caef5afff4e9R186

If you have any suggestions for how this could be better stored, I'm open to changing this behavior. Otherwise, I'd say this is ready for a merge pending your final testing and approval. Thanks!

includes/class-salesforce.php
dkoo: @adekbadek I don't see this error in my logs anymore—can you please confirm?

dkoo:
> Great work! I left a few non-essential notes, all about translation text domains. With or without, this is ready to 🚢

I think I got all of them, thanks! FYI, a lot of strings not in this changeset are also missing the text domain, so maybe something to mop up in a future PR. Merging!

PR #557, opened 2020-06-18
Analytics: add NTG tracking events for newsletters

dkoo:
### All Submissions:

Add additional NTG events as described in the [NTG Playbook](https://newsinitiative.withgoogle.com/training/states/ntg/assets/ntg-playbook.pdf), in order to achieve full NCI support. This will close Issue #559.

Note that this functionality depends on https://github.com/Automattic/newspack-popups/pull/156 in the newspack-popups plugin.

Closes # .

1. Create an inline or overlay campaign containing a Mailchimp newsletter form block. Ensure that it's Active and viewable in a non-logged-in state (e.g. Active and On Every Page).
2. Open a post or page in a new incognito window.
3. Open the Network tab in dev tools. Filter by "ntg" to see only new requests containing "ntg".
4. Scroll down to the campaign (or wait until the overlay campaign appears). The event should fire only once and only when any part of the element first becomes visible within the viewport. Observe an NTG event request in the Network tab that looks something like this (important part is the `NTG%20newsletter&ea=newsletter%20modal%20impression%201` string: https://www.google-analytics.com/collect?v=1&_v=j83&aip=1&a=2091212415&t=event&_s=5&dl=http%3A%2F%2Fnewspack-dkoo.ngrok.io%2Fnon-consequat-commodo-eros-justo-vulputate-potenti%2F&ul=en-us&de=UTF-8&dt=Non%20consequat%20commodo%20eros%20justo%20vulputate%20potenti%20-%20Newspack&sd=30-bit&sr=1792x1120&vp=1303x1018&je=0&ec=NTG%20newsletter&ea=newsletter%20modal%20impression%201&el=Non%20consequat%20commodo%20eros%20justo%20vulputate%20potenti&_u=CCCACUABB~&jid=&gjid=&cid=305040532.1592848697&tid=UA-169138974-1&_gid=1339759271.1592848697&gtm=2ou6a0&z=2058936163
5. Enter a fake email address and submit the form. Observe an NTG event request in the Network tab that looks something like this (important part is the `NTG%20newsletter&ea=newsletter%20signup` string): https://www.google-analytics.com/r/collect?v=1&_v=j83&aip=1&a=2091212415&t=event&_s=6&dl=http%3A%2F%2Fnewspack-dkoo.ngrok.io%2Fnon-consequat-commodo-eros-justo-vulputate-potenti%2F&ul=en-us&de=UTF-8&dt=Non%20consequat%20commodo%20eros%20justo%20vulputate%20potenti%20-%20Newspack&sd=30-bit&sr=1792x1120&vp=1303x1018&je=0&ec=NTG%20newsletter&ea=newsletter%20signup&el=success&_u=CCCACUABB~&jid=1215251838&gjid=1120199689&cid=305040532.1592848697&tid=UA-169138974-1&_gid=1339759271.1592848697&_r=1&gtm=2ou6a0&z=1924980601
6. If you've connected your local Newspack environment to Site Kit and a Google Analytics account, after several hours you should see these events being reported in the analytics dashboard.

### Other information:

### DIFF
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -18,6 +18,7 @@ class Analytics {
 	 * Constructor
 	 */
 	public function __construct() {
+		add_filter( 'render_block', [ __CLASS__, 'add_amp_visibility_pixel' ], 10, 2 );
 		add_filter( 'googlesitekit_amp_gtag_opt', [ __CLASS__, 'inject_amp_events' ] );
 		add_action( 'wp_footer', [ __CLASS__, 'inject_non_amp_events' ] );
 	}
@@ -120,22 +121,40 @@ class Analytics {
 				'event_category' => 'NTG newsletter',
 			],
 			[
-				'id'             => 'newsletterImpression',
+				'id'             => 'newsletterImpressionOverlay',
 				'on'             => 'visible',
 				'amp_element'    => '.newspack-newsletter-prompt-overlay',
 				'element'        => '.wp-block-jetpack-mailchimp form',
-				'event_name'     => 'newsletter modal impression 1', // 1: overlay/lightbox
+				'event_name'     => 'newsletter modal impression overlay', // 1: overlay/lightbox campaign
 				'event_label'    => get_the_title(),
 				'event_category' => 'NTG newsletter',
+				'visibilitySpec' => [
+					'totalTimeMin' => 500,
+				],
 			],
 			[
-				'id'             => 'newsletterImpression',
+				'id'             => 'newsletterImpressionInline',
 				'on'             => 'visible',
 				'amp_element'    => '.newspack-newsletter-prompt-inline',
 				'element'        => '.wp-block-jetpack-mailchimp form',
-				'event_name'     => 'newsletter modal impression 2', // 2: inline prompt
+				'event_name'     => 'newsletter modal impression inline', // 2: inline prompt campaign
 				'event_label'    => get_the_title(),
 				'event_category' => 'NTG newsletter',
+				'visibilitySpec' => [
+					'totalTimeMin' => 500,
+				],
+			],
+			[
+				'id'             => 'newsletterImpressionContent',
+				'on'             => 'visible',
+				'amp_element'    => '.entry-content > .newspack-visibility-wrapper',
+				'element'        => '.wp-block-jetpack-mailchimp form',
+				'event_name'     => 'newsletter modal impression content', // 3: in-content prompt (not a campaign)
+				'event_label'    => get_the_title(),
+				'event_category' => 'NTG newsletter',
+				'visibilitySpec' => [
+					'totalTimeMin' => 500,
+				],
 			],
 		];
 
@@ -143,6 +162,27 @@ class Analytics {
 		 * Other integrations can add events to track using this filter.
 		 */
 		return apply_filters( 'newspack_analytics_events', $events );
+	}
+
+	/**
+	 * Filters block HTML content server-side.
+	 * Lets us extend core/Jetpack blocks without breaking the editor experience.
+	 * This filter callback adds an invisible 1x1 <amp-img> pixel to the block HTML
+	 * that can be used to track their visibility using the AMP visibility trigger.
+	 *
+	 * @param string $block_content The HTML for the block to be filtered.
+	 * @param array  $block Block data for the block to be filtered.
+	 * @return string Modified HTML for the block to be rendered.
+	 */
+	public static function add_amp_visibility_pixel( $block_content, $block ) {
+		$blocks_to_filter = [ 'jetpack/mailchimp' ];
+
+		if ( ! in_array( $block['blockName'], $blocks_to_filter ) ) {
+			return $block_content;
+		}
+
+		$filtered_content = '<amp-layout class="newspack-visibility-wrapper">' . $block_content . '</amp-layout>';
+		return $pixel . $filtered_content;
 	}
 
 	/**
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -352,7 +352,7 @@ class Analytics {
 		?>
 		<script>
 			( function() {
-				window.viewedElements = window.viewedElements || [];
+				window.newspackViewedElements = window.newspackViewedElements || [];
 				window.newspackCheckVisibility = window.newspackCheckVisibility || function( el ) {
 					var rect = el.getBoundingClientRect();
 
@@ -369,9 +369,8 @@ class Analytics {
 
 				var reportEvent = function() {
 					for ( var i = 0; i < elements.length; ++i ) {
-						if ( window.newspackCheckVisibility( elements[i] ) && -1 === window.viewedElements.indexOf( elements[i] ) ) {
-							console.log( 'viewed', elements );
-							window.viewedElements.push( elements[i] );
+						if ( window.newspackCheckVisibility( elements[i] ) && -1 === window.newspackViewedElements.indexOf( elements[i] ) ) {
+							window.newspackViewedElements.push( elements[i] );
 
 							gtag(
 								'event',
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -113,7 +113,7 @@ class Analytics {
 				'id'             => 'newsletterSignup',
 				'amp_on'         => 'amp-form-submit-success',
 				'on'             => 'submit',
-				'element'        => '#mailchimp_form',
+				'element'        => '#mailchimp_form, .wp-block-jetpack-mailchimp form',
 				'event_name'     => 'newsletter signup',
 				'event_label'    => 'success',
 				'event_category' => 'NTG newsletter',
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -101,11 +101,22 @@ class Analytics {
 				'on'             => 'scroll',
 				'event_name'     => '100%',
 				'event_value'    => 100,
-				'event_label'    => get_the_title(),
+				'event_label'    => 'success',
 				'event_category' => 'NTG article milestone',
 				'scrollSpec'     => [
 					'verticalBoundaries' => [ 100 ],
 				],
+			],
+
+			// Newsletters.
+			[
+				'id'             => 'newsletterSignup',
+				'amp_on'         => 'amp-form-submit-success',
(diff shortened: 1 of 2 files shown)


dkoo:
@adekbadek (cc @jeffersonrabb) Getting this started as a draft with just one event, to make sure I'm on the right track. I have questions which I could use your help on:

* In general, once I've implemented both AMP and non-AMP events, what's the best way to test whether they're working correctly?
* For the Newsletter Signup event: I notice that the `#mailchimp_form` element is not actually an `<amp-form>` in my dev tools. Will the `amp-form-submit-success` event still work? If not, how do we target the regular non-AMP form on the AMP page instead?
* For Newsletter Prompt (the next event I'm looking at), you're supposed to trigger an event when the prompt is seen.
  - What's the best way to identify a Campaign that is related to Newsletters? Should we enforce adding a `Newsletter` category for Campaign post types that relate to newsletters, so we can trigger the NTG event only when the campaign is a newsletter prompt?

Will probably have other questions as I do the other events.

adekbadek:
---
> In general, once I've implemented both AMP and non-AMP events, what's the best way to test whether they're working correctly?

Go to [GA web UI](https://analytics.google.com/analytics/web/) and visit Behavior->Events:
image

From here you can query all the reported events. 

---
> For the Newsletter Signup event: I notice that the #mailchimp_form element is not actually an <amp-form> in my dev tools. Will the amp-form-submit-success event still work? If not, how do we target the regular non-AMP form on the AMP page instead?

Maybe simply `"form"` as `selector` will work?

---

> What's the best way to identify a Campaign that is related to Newsletters? 

Here's a similar case: Campaign analytics checks if there's a form in the campaign content to determine if conversion rate is applicable

https://github.com/Automattic/newspack-plugin/blob/900644b40eebf95d17fb9dfbb780102225b4ee9b/includes/popups-analytics/class-popups-analytics-utils.php#L235-L237

jeffersonrabb:
> For Newsletter Prompt (the next event I'm looking at), you're supposed to trigger an event when the prompt is seen.

@adekbadek it seems there is going to be some overlap between Campaign analytics and NTG recommendations, which leads to a few questions:

1) Should we consider dropping our existing Campaign events in favor of NTG's, for consistency's sake?
2) If yes, will it be possible for the Campaign Analytics Wizard to aggregate the old and new tagging?
3) Should all Campaign-related events be implemented in the Campaigns plugin vs. here in Newspack Plugin?

jeffersonrabb:
> * In general, once I've implemented both AMP and non-AMP events, what's the best way to test whether they're working correctly?

Another tip for testing GA: open Chrome DevTools and go to Network tab, "All." Type `google-analytics` in the filter. As events fire, you should see requests to `https://google-analytics.com/collect` for each one. This should occur in AMP or non-AMP. 

Also, this part probably isn't news to you but just in case... to toggle between AMP and non, set the AMP plugin to "Transitional" mode. There will be an option to switch between the two versions in the admin bar. Or, you can simply append `?amp`to the end of the URL to see the AMP version.

adekbadek:
@jeffersonrabb:

> it seems there is going to be some overlap between Campaign analytics and NTG recommendations […] Should we consider dropping our existing Campaign events in favor of NTG's, for consistency's sake?

I don't think so  – NTG events are different than Campaigns events. Campaigns analytics is data about _any_ impression/conversion on a campaign. NTG data is about a couple of specific kinds of impressions/conversions, not necessarily as a part of a campaign. Say we have a campaign promoting an election liveblog – no NTG event captures that. In other words, I think there's some, but not enough, consistency between Campaign- and NTG events.

dkoo:
@adekbadek @jeffersonrabb Maybe it's my ignorance of how AMP works, but I can't for the life of me get the newsletter signup NTG event to fire. I have this trigger rendering in the last `<amp-analytics>` block along with all the other NTG events:

```
"newsletterSignup": {
	"on": "amp-form-submit-success",
	"request": "event",
	"selector": "#mailchimp_form",
	"vars": {
		"event_name": "newsletter signup",
		"event_label": "success",
		"event_category": "NTG newsletter"
	},
	"extraUrlParams": { "ni": 1 }
}
```

I've confirmed that it looks very similar to this trigger that IS being fired, from the popups plugin:

```
"popupFormSubmitSuccess": {
	"on": "amp-form-submit-success",
	"request": "event",
	"selector": "#lightbox1095194634 form:not(.popup-action-form)",
	"vars": {
		"event_name": "Form Submission",
		"event_label": "Newspack Announcement: Inline Campaign 3 (325)",
		"event_category": "Newspack Announcement"
	}
},
```

But the NTG trigger just doesn't fire. I tried changing the `selector` to just `form` and it still didn't fire. Is there some binding that I'm missing here? I did confirm that if I paste the NTG trigger with selector `#mailchimp_form` to the newspack-popups plugin in the same place the "Form Submission" trigger is created, it does get fired, but how can we ensure that it gets fired as part of the `class-analytics.php` class?

dkoo:
Confirming that viewing the page while not logged out lets the trigger fire correctly.

However, now I'm seeing a console error: `error.js:201 Element " #mailchimp_form " is required to be an AMP element`

This form seems to come from the Jetpack plugin, so do we have control over that? I'm not sure the AMP event will get tracked if this error is thrown.

jeffersonrabb:
> Maybe it's my ignorance of how AMP works, but I can't for the life of me get the newsletter signup NTG event to fire.

@dkoo looks like there's a bug that breaks the form submission events, but only in non-AMP requests. Issues filed:

https://github.com/Automattic/newspack-popups/issues/151
https://github.com/Automattic/newspack-popups/issues/152

adekbadek:
@dkoo

> Confirming that viewing the page while not logged in lets the trigger fire correctly.

Yes, sorry about not warning you about that :) That's the difference between the implementation here and in campaigns plugin (newspack-popups) – the former uses a Site Kit hook, and Site Kit does not report data for logged-in users.

includes/class-analytics.php
```diff
@@ -101,12 +101,23 @@ public static function get_events() {
 				'on'             => 'scroll',
 				'event_name'     => '100%',
 				'event_value'    => 100,
-				'event_label'    => get_the_title(),
+				'event_label'    => 'success',
```
adekbadek, line 104: Is that intended? The event label for this article milestone events _is_ supposed to be the article title.

dkoo: Whoops, not intentional. Thanks for pointing out.

dkoo:
@adekbadek @jeffersonrabb I think I have these newsletter events (view and signup) working in both AMP and non-AMP pages. 

The only thing that I haven't been able to figure out: the AMP `visible` event requires that the element being tracked is an AMP-specific element (e.g. `amp-ad`, `amp-img`, etc.) and doesn't support arbitrary DOM elements like `div`. This is okay for Campaigns because they're `amp-lightbox` elements, but for a regular Gutenberg block, it's not wrapped in an AMP element, so I'm not sure how we can track visibility using the AMP visibility event. 

I did implement a simple visibility checker for non-AMP pages that would work for any DOM element—is it possible to reuse this for AMP as well? I don't see how since the events are tracked as JSON in an `amp-analytics` block instead of via JS.

If tracking just Campaigns is okay for a phase 1, I can ensure that the `signup` event only gets fired from Campaigns instead of both campaigns and standalone blocks. Please let me know your thoughts. Thanks!

includes/class-analytics.php
```diff
@@ -107,6 +107,42 @@ public static function get_events() {
...
+				'id'             => 'newsletterImpression',
+				'on'             => 'visible',
+				'amp_element'    => '.newspack-newsletter-prompt-inline',
+				'element'        => '.wp-block-jetpack-mailchimp form',
+				'event_name'     => 'newsletter modal impression 2', // 2: inline prompt
+				'event_label'    => get_the_title(),
+				'event_category' => 'NTG newsletter',
+				'visibilitySpec' => [
+					'totalTimeMin' => 500,
+				],
```
adekbadek, line 144: The timeout will not be handled in non-AMP – perhaps we should drop it, so that AMP and non-AMP reporting is as similar as possible?

dkoo: I removed it for now.

dkoo:
Not quite ready for merging yet, because the behavior isn't consistent in my testing. But https://github.com/Automattic/newspack-plugin/pull/557/commits/5853f937d1907f00fe7a05db2ae45176d643d7d1 adds a filter that wraps the Mailchimp block in an `amp-layout`, which lets us track visibility regardless of context. The challenges now are: 

1. The selectors for the overlay/inline modals don't seem to be working well. Both campaign types seem to get tracked as the same type regardless of the selector.
2. The `visibilitySpec` timeout seems to be needed, otherwise an impression is tracked for the overlay campaign even if hidden. The timeout ensures that the impression won't be tracked if it's not displayed.

includes/class-analytics.php
dkoo: As mentioned in another comment, this seems to be needed to avoid false impressions on hidden campaigns. I think the better solution would be to recreate the timeout logic for non-AMP events as well.

dkoo:
This is mainly working now, except for an issue where I want to use a `>` CSS selector in the non-AMP analytics JS, but escaping the output causes the `>` to be encoded to `&gt;`, which invalidates the JS. Here's an example of the offending line: https://github.com/Automattic/newspack-plugin/pull/557/commits/abf48ca9f6881005a3376140c566ae539ba9b581#diff-55af0c7c94b4a43e428b9ba731be0f2fR356

Currently this PR is unmergeable because this causes CircleCI to fail at the `lint-php` stage.

dkoo:
Also wanted to mention that there's an edge case where on an AMP page, if there are multiple instances of a class-based selector that should trigger a visibility event, only the first instance of the matched selector will be reported. This is still an open issue in the AMP project: https://github.com/ampproject/amphtml/issues/11380

If you create more than one Mailchimp block in the content of a post or page (NOT within multiple campaigns, which are targeted by unique ID), the `newsletter modal impression` event will only get reported for the first instance of the block rather than every instance.

dkoo:
#575 has a cleaner approach that couples with hooks from the Campaigns plugin. Closing this PR in favor of that branch.

adamsilverstein:
Please note for the AMP case, this PR merge https://github.com/ampproject/amphtml/pull/29172 should help with this issue. cc: @dkoo @jeffersonrabb

PR #567, opened 2020-06-25
feat: apply analytics events filter in block filter

adekbadek:


dkoo:
Pasting from a Slack thread for posterity:

1. For some reason, I'm getting some errors that are preventing analytics from being reported:
[AmpAnalytics <unknown id>] Error rewriting configuration:  https://www.googletagmanager.com/gtag/amp Error: XHR Failed fetching (https://www.googletagmanager.com/...): Failed to fetch
[AmpAnalytics <unknown id>] No request strings defined. Analytics data will not be sent from this page.
[AmpAnalytics <unknown id>] No triggers were found in the config. No analytics data will be sent.
Confirmed that this only happens when I check out this branch.

2. The render_block hook runs more times than the blocks are actually rendered on the page, resulting in duplicate triggers being added to the config. On a post where I had just two campaigns (an overlay and an inline) I counted this hook running 11 times, resulting in 11 triggers.

3. The NTG Playbook suggests reporting the context of the prompt and assigning a number, e.g. 1: overlay, 2: inline, 3: in-content. Even if we were to fix the render_block hook so that it only runs exactly as many times as the block appears on the page, if we rely on the hook for creating all triggers, how do we know whether the block is appearing in an overlay campaign, inline campaign, or in-content?

For reason 3 I'm still leaning toward creating the triggers for campaigns in the popups plugin. To me, it also makes sense to isolate any campaigns-related functionality in that plugin, don't you think? They are still using the analytics filter in the main plugin, so we can ensure that they're outputted in the same format. I'm still stumped on creating unique IDs for the in-content blocks, though. Do you have any idea how we might ensure that we're only adding as many amp-layout wrappers as there are instances of the block? And any idea how we can maybe not run the filter if the block is being rendered in a campaign? I tried testing get_post_type within the hook's callback thinking that it might output newspack_popups_cpt when being rendered in campaigns, but it seems inconsistent (outputs newspack_popups_cpt on pages, but post on posts).

adekbadek:
> 1. getting some errors that are preventing analytics from being reported

I'm getting these from time to time – but so far _not_ on this branch.

---

> 2. The `render_block` hook runs more times than the blocks are actually rendered on the page

Yeah, I haven't tested this with `newspack-popups` 🙈 I think what happens is that popups plugin calls `render_block` function once per block, and that results in multiple calls to the `render_block` hook. 

---

> 3. how do we know whether the block is appearing in an overlay campaign, inline campaign, or in-content?

That's tough! We'd have to know the block's context in the DOM.

---

The approach in this PR works for the in-content newsletter forms, but is incompatible with popups plugin.
What can be done is marking the newsletter blocks which are rendered in a popup and skipping them in this implementation. The popups plugin can add NTG events by as proposed in https://github.com/Automattic/newspack-popups/pull/159

I've drafted something [here](https://github.com/Automattic/newspack-popups/blob/b199d68ef74222f7e28419d7fbcd98e69177fc86/includes/class-newspack-popups-model.php#L301-L351) on popups side, and here (6477b812d84e8a38ca0308938cd2446a596688a3) on plugin side. 

This solution solves issue `3` but adds some complexity and still suffers from issue `2`.

adekbadek:
Closing in favour of #575

PR #568, opened 2020-06-26
Add non_interaction: true for scroll events

dkoo:
### All Submissions:

Adds `non_interaction: true` for events which are recommended to have that attribute in the [NTG Playbook](https://newsinitiative.withgoogle.com/training/states/ntg/assets/ntg-playbook.pdf). Also lets us pass a `non_interaction` property to events and output that value to the event JS or JSON as part of the `newspack_analytics_events` filter. Basically, any event that is tracked based on visibility, scroll depth, or anything that doesn't happen as a direct result of user action (such as a click or form submit) should be marked non-interactive in this way.

Relates to #557, where we also implemented this fix, but that PR is dragging so we wanted to ensure that it made it into the next release.

1. Open a page or post while not logged in.
2. View source for the page. Look for event triggers with `"event_category": "NTG article milestone"`.
3. These should have `"non_interaction": true` set on them. 

### Other information:


adekbadek:
(formatted the code in 8d739d3)

PR #575, opened 2020-07-01
feat: ntg mailing list events in amp

### LINKED ISSUE #559: NTG events: Newsletters
Add Newsletter-related Google Analytics events for AMP and non-AMP pages, as described in the [NCI Playbook](https://newsinitiative.withgoogle.com/training/states/ntg/assets/ntg-playbook.pdf).

Events to add:

* Newsletter signup
* Newsletter prompt

jeffersonrabb:
### All Submissions:

Adds NTG newsletter events (signup and visibility). Replaces #557. This approach uses the `render_block` hook to alter Mailchimp block markup for AMP analytics compatibility, and also uses new hooks from https://github.com/Automattic/newspack-popups/pull/164 to determine the context of the block (inline campaign, overlay campaign, or other).

Closes #559.

1. Check out https://github.com/Automattic/newspack-popups/pull/164 in the campaigns plugin or make sure the changes in there are merged to master and active in your local environment.
2. Create at least two different campaigns: 1.) Overlay, site-wide default, and 2.) Inline, Every Page. Both should be Active (not test mode) and both should contain a Mailchimp newsletter signup block.
3. Create a new page or post and insert a Mailchimp newsletter block into the content (or insert a Mailchimp block into an existing page or post's content). Ensure that AMP is enabled.
4. Open this page or post in a new incognito window.
5. Open the Network tab in dev tools. Filter by "ntg" to see only new requests containing "ntg".
6. Test visibility event: Scroll down to the inline campaign (or wait until the overlay campaign appears). The event should fire only once and only when any part of the Mailchimp block element first becomes visible within the viewport. Observe an NTG event request in the Network tab that looks something like this (important part is the `NTG%20newsletter&ea=newsletter%20modal%20impression%201` string: https://www.google-analytics.com/collect?v=1&_v=j83&aip=1&a=2091212415&t=event&_s=5&dl=http%3A%2F%2Fnewspack-dkoo.ngrok.io%2Fnon-consequat-commodo-eros-justo-vulputate-potenti%2F&ul=en-us&de=UTF-8&dt=Non%20consequat%20commodo%20eros%20justo%20vulputate%20potenti%20-%20Newspack&sd=30-bit&sr=1792x1120&vp=1303x1018&je=0&ec=NTG%20newsletter&ea=newsletter%20modal%20impression%201&el=Non%20consequat%20commodo%20eros%20justo%20vulputate%20potenti&_u=CCCACUABB~&jid=&gjid=&cid=305040532.1592848697&tid=UA-169138974-1&_gid=1339759271.1592848697&gtm=2ou6a0&z=2058936163
7. The number at the end of the `NTG%20newsletter&ea=newsletter%20modal%20impression%201` is significant. `1` = inline campaign, `2` = overlay campaign, `3` = other. Verify that this number in the request matches the context of the block you're viewing as it becomes visible.
8. Test form submit event: Enter an email address and submit the form. Observe an NTG event request in the Network tab that looks something like this (important part is the `NTG%20newsletter&ea=newsletter%20signup` string): https://www.google-analytics.com/r/collect?v=1&_v=j83&aip=1&a=2091212415&t=event&_s=6&dl=http%3A%2F%2Fnewspack-dkoo.ngrok.io%2Fnon-consequat-commodo-eros-justo-vulputate-potenti%2F&ul=en-us&de=UTF-8&dt=Non%20consequat%20commodo%20eros%20justo%20vulputate%20potenti%20-%20Newspack&sd=30-bit&sr=1792x1120&vp=1303x1018&je=0&ec=NTG%20newsletter&ea=newsletter%20signup&el=success&_u=CCCACUABB~&jid=1215251838&gjid=1120199689&cid=305040532.1592848697&tid=UA-169138974-1&_gid=1339759271.1592848697&_r=1&gtm=2ou6a0&z=1924980601
9. Disable AMP. Repeat the tests on the non-AMP version of the page.
10. Test other types of pages and posts, then re-test after dismissing the campaigns.
11. Finally, disable the campaigns and re-test the original page/post.
12. Look out for JS console errors or missing event reporting requests in all cases.

### Other information:

### DIFF
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -19,7 +19,7 @@ class Analytics {
 	 *
 	 * @var array
 	 */
-	public static $block_amp_events = [];
+	public static $block_events = [];
 
 	/**
 	 * An integer to indicate the context a block is rendered in, e.g. content|overlay campaign|inline campaign.
@@ -34,7 +34,7 @@ class Analytics {
 	public function __construct() {
 		add_filter( 'googlesitekit_amp_gtag_opt', [ __CLASS__, 'inject_amp_events' ] );
 		add_action( 'wp_footer', [ __CLASS__, 'inject_non_amp_events' ] );
-		add_filter( 'render_block', [ __CLASS__, 'prepare_blocks_for_amp_events' ], 10, 2 );
+		add_filter( 'render_block', [ __CLASS__, 'prepare_blocks_for_events' ], 10, 2 );
 		add_action( 'newspack_campaigns_before_campaign_render', [ __CLASS__, 'set_campaign_render_context' ], 10, 1 );
 		add_action( 'newspack_campaigns_after_campaign_render', [ __CLASS__, 'reset_render_context' ], 10, 1 );
 	}
@@ -167,14 +167,11 @@ class Analytics {
 	 * @param string $content The block content about to be appended.
 	 * @param object $block The full block, including name and attributes.
 	 */
-	public static function prepare_blocks_for_amp_events( $content, $block ) {
+	public static function prepare_blocks_for_events( $content, $block ) {
 		if ( is_admin() ) {
 			return $content;
 		}
-		$is_amp = function_exists( 'is_amp_endpoint' ) && is_amp_endpoint();
-		if ( ! $is_amp ) {
-			return $content;
-		}
+
 		switch ( $block['blockName'] ) {
 			case 'jetpack/mailchimp':
 				$content = self::prepare_jetpack_mailchimp_block( $content );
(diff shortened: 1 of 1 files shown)


dkoo:
Adding some details here on a weird edge case @jeffersonrabb and I discovered, because I haven't been able to determine why it's happening. To reproduce:

1. Enable Jetpack Related Posts and enable thumbnails if available (for convenience, this doesn't actually matter for reproducing the issue).
2. View a post in AMP mode that will display at least one campaign containing the Mailchimp block. Doesn't matter if it's overlay or inline.
3. Confirm that there are related posts appearing at the bottom of the post content.
4. If all three related posts have featured images, edit one and remove the featured image, then save that post. If at least one related post already lacks a featured image, skip to the next step.
5. Observe in the JS console several `Element not found` AMP errors.

This is happening because for some reason, the `the_content` hook runs twice for the main post if there's a related post that lacks a featured image rendered by the Jetpack Related Posts block. This causes our campaigns to be built twice, because they're built baed on `the_content`, although each campaign is only injected once into the actual content. But because the campaigns are built twice, the `render_block` hook runs twice, two unique IDs are generated, and two AMP analytics triggers are outputted, but only one ID actually matches an element that ends up in the rendered post content. The AMP error is then thrown for the duplicate trigger with a unique ID that doesn't match the rendered campaign.

For reference, here's the file in Jetpack that handles Related Posts: https://github.com/Automattic/jetpack/blob/master/modules/related-posts/jetpack-related-posts.php

I don't believe this bug should stop us from merging and releasing this feature, because 1.) it doesn't invalidate AMP for the page, 2.) it doesn't prevent the correct events from firing and getting reported, and 3.) it happens independently of the changes in this particular branch, so if we can find the cause, we should be able to push a solution it in a future PR.

jeffersonrabb:
> Adding some details here on a weird edge case @jeffersonrabb and I discovered, because I haven't been able to determine why it's happening.

@dkoo I suspect this is happening because of [this line](https://github.com/Automattic/jetpack/blob/master/_inc/lib/class.media-summary.php#L40), which is called [here](https://github.com/Automattic/jetpack/blob/master/modules/related-posts/jetpack-related-posts.php#L1326), but only if there is no featured image. We may be able to address by adding before and after actions similar to [these](https://github.com/Automattic/jetpack/blob/master/class.jetpack-post-images.php#L567-L586).

dkoo:
> @dkoo I suspect this is happening because of [this line](https://github.com/Automattic/jetpack/blob/master/_inc/lib/class.media-summary.php#L40), which is called [here](https://github.com/Automattic/jetpack/blob/master/modules/related-posts/jetpack-related-posts.php#L1326), but only if there is no featured image. We may be able to address by adding before and after actions similar to [these](https://github.com/Automattic/jetpack/blob/master/class.jetpack-post-images.php#L567-L586).

@jeffersonrabb Good call, but after stepping away from this problem for a few days I wonder if we might be able to solve this more simply. Would this do the trick? https://github.com/Automattic/newspack-popups/pull/165

Since the problem is due to the Campaigns insertion code running too many times when `the_content` fires more than expected, maybe we can just avoid re-running the insertion more than once there. In initial tests it avoids the errors while also still allowing the campaigns to render correctly with corresponding analytics code.

PR #611, opened 2020-07-31
Add custom events adding UI

### LINKED ISSUE #601: Enable reporting CTR on sidebar and navbar links
I'd be beneficial to a publisher to be able to see CTRs on links in navbar and sidebar. 

Example: a site displays a "Donate" link in the top navbar. A publisher would like to know how many people, out of all who seen it, have clicked the link. 

## Technical implementation

Basically it's about adding analytics events via [the `newspack_analytics_events` filter](https://github.com/Automattic/newspack-plugin/blob/9eaaf1842619110effca4a873b64ab6d85f67915/includes/class-analytics.php#L175) - an impression event, and a click event. 
The tricky part is choosing when to do it – sending these events for each link might be too much and spam the GA data. A simple solution would be to have a setting for that, off by default. A more granular solution would be to enable choosing exactly which links should have those events attached. 
We also might consider displaying the CTR data in Newspack UI.

adekbadek:
### All Submissions:

Adding custom events insertion UI. Only `click` and `submit` events are handled: 
- `visibility`, `ini-load` require the element to be AMP element,
- `scroll` requires some more UI for scroll parameters, can be implemented later.

Design considerations:

1. This is an advanced feature, misuse of which can brake sites – we may wanna add some warning there
2. Some fields in the form are not required, not sure if we have an established pattern for that
3. The "Active" field is of a different kind, maybe it should be displayed differently in the form? 

Closes #601

1. Open the Analytics Wizard, observe a new tab – "Custom Events"
2. Add a custom event, e.g. a click on `#search-toggle`
3. Visit the front end (as non-logged-in user), oberve the event being sent on action (look for requests to `https://www.google-analytics.com/r/collect` in Network tab) 
4. Test on both AMP and non-AMP versions
4. Mark the event as not active, observe it's not sent anymore

### Other information:

### DIFF
--- a/assets/components/src/text-control/index.js
+++ b/assets/components/src/text-control/index.js
@@ -39,8 +39,8 @@ class TextControl extends Component {
 		const { className, required, ...otherProps } = this.props;
 		const classes = classNames( 'newspack-text-control', className );
 		return required ? (
-			<div ref={ this.wrapperRef } className={ classes }>
-				<BaseComponent required={ required } { ...otherProps } />
+			<div ref={ this.wrapperRef }>
+				<BaseComponent className={ classes } required={ required } { ...otherProps } />
 			</div>
 		) : (
 			<BaseComponent required={ required } className={ classes } { ...otherProps } />
--- a/assets/components/src/text-control/index.js
+++ b/assets/components/src/text-control/index.js
@@ -6,7 +6,8 @@
  * WordPress dependencies
  */
 import { TextControl as BaseComponent } from '@wordpress/components';
-import { Component } from '@wordpress/element';
+import { Component, createRef } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
@@ -19,13 +20,31 @@ import './style.scss';
 import classNames from 'classnames';
 
 class TextControl extends Component {
+	constructor( props ) {
+		super( props );
+		this.wrapperRef = createRef();
+	}
+	componentDidMount() {
+		if ( this.wrapperRef.current ) {
+			const labelEl = this.wrapperRef.current.querySelector( 'label' );
+			if ( labelEl ) {
+				labelEl.setAttribute( 'data-required-text', __( '(required)', 'newspack' ) );
+			}
+		}
+	}
 	/**
 	 * Render.
 	 */
 	render() {
-		const { className, ...otherProps } = this.props;
+		const { className, required, ...otherProps } = this.props;
 		const classes = classNames( 'newspack-text-control', className );
-		return <BaseComponent className={ classes } { ...otherProps } />;
+		return required ? (
+			<div ref={ this.wrapperRef } className={ classes }>
+				<BaseComponent required={ required } { ...otherProps } />
+			</div>
+		) : (
+			<BaseComponent required={ required } className={ classes } { ...otherProps } />
+		);
 	}
 }
 
--- a/assets/components/src/text-control/style.scss
+++ b/assets/components/src/text-control/style.scss
@@ -11,7 +11,6 @@
 	line-height: 24px;
 	margin: 32px 0;
 
-
 	label {
 		color: $dark-gray-900;
 		font-size: inherit;
@@ -24,7 +23,7 @@
 				content: attr( data-required-text );
 				padding-left: 5px;
 				font-weight: normal;
-				opacity: 0.5;
+				color: $dark-gray-300;
 			}
 		}
 	}
--- a/assets/components/src/text-control/style.scss
+++ b/assets/components/src/text-control/style.scss
@@ -11,12 +11,22 @@
 	line-height: 24px;
 	margin: 32px 0;
 
+
 	label {
 		color: $dark-gray-900;
 		font-size: inherit;
 		font-weight: bold;
 		line-height: inherit;
 		margin: 0;
+
+		&[data-required-text] {
+			&::after {
+				content: attr( data-required-text );
+				padding-left: 5px;
+				font-weight: normal;
+				opacity: 0.5;
+			}
+		}
 	}
 
 	input[type='date'],
--- a/assets/wizards/analytics/views/custom-events/index.js
+++ b/assets/wizards/analytics/views/custom-events/index.js
@@ -133,7 +133,7 @@ class CustomEvents extends Component {
 					rawHTML
 					isInfo
 					noticeText={ `${ __(
-						'This is an advanced feature, read more about on our',
+						'This is an advanced feature, read more about it on our',
 						'newspack'
 					) } <a href="${ NEWSPACK_SUPPORT_URL }/analytics">${ __(
 						'support page',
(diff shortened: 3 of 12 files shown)


includes/wizards/class-analytics-wizard.php
```diff
@@ -311,4 +368,28 @@ public static function set_category_dimension( $request ) {
...
+					$event['id'] = uniqid();
+				}
+				return $event;
+			},
+			$request['events']
+		);
+		if ( update_option( self::$custom_events_option_name, wp_json_encode( $custom_events ) ) ) {
+			return [ 'events' => $custom_events ];
+		} else {
+			return new WP_Error( 'newspack_analytics', __( 'Error when setting category custom dimension.', 'newspack' ) );
```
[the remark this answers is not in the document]
adekbadek: Good catch, fixed in 17a4498

adekbadek:
> How badly can this feature break the site? Is it that your analytics would stop working, or can you actually break site functionality with this?

No, nothing bad can happen. It's more about the user not being surprised by this developer-like options. I think it will be enough to say "This is an advanced feature, read more about on our support page" – with docs link as you suggested:

image

> As an advanced section, it could also help if it's not the first/default tab. 

Good call, updated in 3bcb711a8af62e5cb4d40fb8d7faaba2b1b81e73

> I think we could keep it simple visually, but I'm not sure complicated doing something like this would be to do correctly (eg. have text, or a CSS class, be appended automatically based on the 'required' status):

Done in a pretty hacky way in 603f02145d83ddd57a780a72287a74af295efafd

> How is [active field] different from the other fields?

Should've been clearer about this, sorry :) It's different because "active" is not a property of a GA custom event, it's just for the use of the plugin. A user might want to have the custom event defined, but not reported temporarily – in such a case they can mark it as inactive, like a post draft. But it's not a big issue, just looked a bit odd in the modal:


> but the checkbox was disabled for me after creating new events

That's on me – the table row is clickable, and then you can click the checkbox in a modal.  I've changed it so there are "Edit" links in the table, and an icon instead of a disabled checkbox:

image

> Lastly, tiny nitpick, what do you think about changing the 'Non-interaction' label to 'Non-interaction event' (just to match Google's docs)?

Sure, updated in 3bcb711a8af62e5cb4d40fb8d7faaba2b1b81e73

adekbadek:
> How badly can this feature break the site? Is it that your analytics would stop working, or can you actually break site functionality with this?

No, nothing bad can happen. It's more about the user not being surprised by this developer-like options. I think it will be enough to say "This is an advanced feature, read more about on our support page" – with docs link as you suggested:

image

Note that the docs page is a draft at the moment.

> As an advanced section, it could also help if it's not the first/default tab. 

Good call, updated in 90f67a5515fc2083baa7e4cdefdb863c1642acd7

> I think we could keep it simple visually, but I'm not sure complicated doing something like this would be to do correctly (eg. have text, or a CSS class, be appended automatically based on the 'required' status):

Done in a pretty hacky way in 0a0fd687c3a58de4cd4d965bc28e87f82dfb228e

> How is [active field] different from the other fields?

Should've been clearer about this, sorry :) It's different because "active" is not a property of a GA custom event, it's just for the use of the plugin. A user might want to have the custom event defined, but not reported temporarily – in such a case they can mark it as inactive, like a post draft. But on a second thought, I don't think it's a big issue.

> but the checkbox was disabled for me after creating new events

That's on me – the table row was clickable, and the checkbox was in the way.  I've changed it so there are "Edit" links in the table, and an icon instead of a disabled checkbox:

image

> Lastly, tiny nitpick, what do you think about changing the 'Non-interaction' label to 'Non-interaction event' (just to match Google's docs)?

Sure, updated in 90f67a5515fc2083baa7e4cdefdb863c1642acd7

assets/components/src/text-control/index.js
```diff
@@ -19,13 +20,31 @@ import './style.scss';
...
 	 * Render.
 	 */
 	render() {
-		const { className, ...otherProps } = this.props;
+		const { className, required, ...otherProps } = this.props;
 		const classes = classNames( 'newspack-text-control', className );
-		return <BaseComponent className={ classes } { ...otherProps } />;
+		return required ? (
+			<div ref={ this.wrapperRef } className={ classes }>
+				<BaseComponent required={ required } { ...otherProps } />
```
laurelfulford, line 43: The current placement of `className` on require text controls causes `.newspack-text-control` not to be added to `div.components-base-control`, which makes the labels the wrong font size -- they're `13px` rather `14px`:

image

Moving the `className` to the `BaseComponent` seems to fix, but I'm not sure if this is a bad idea for a different reason:

```suggestion
			<div ref={ this.wrapperRef }>
				<BaseComponent className={ classes } required={ required } { ...otherProps } />
```

assets/components/src/text-control/style.scss
```diff
@@ -11,12 +11,22 @@
...
 		font-weight: bold;
 		line-height: inherit;
 		margin: 0;
+
+		&[data-required-text] {
+			&::after {
+				content: attr( data-required-text );
+				padding-left: 5px;
+				font-weight: normal;
+				opacity: 0.5;
```
laurelfulford, line 27: Minor nitpick, but suggest swapping the opacity here for one of the existing mid-greys used in the UI, like:

```suggestion
				 color: $dark-gray-300;
```

assets/wizards/analytics/views/custom-events/index.js
```diff
@@ -0,0 +1,280 @@
...
+						'newspack'
+					) }
+				</p>
+				<Notice
+					rawHTML
+					isInfo
+					noticeText={ `${ __(
+						'This is an advanced feature, read more about it on our',
+						'newspack'
+					) } <a href="${ NEWSPACK_SUPPORT_URL }/analytics">${ __(
```
laurelfulford, line 138: Another really minor nitpick: 

```suggestion
					) } <a target="_blank" href="${ NEWSPACK_SUPPORT_URL }/analytics">${ __(
```

assets/components/src/text-control/index.js
adekbadek: Done in 5ae9711750d7e43a80fd15b32bcfed700c7e849b

assets/components/src/text-control/style.scss
adekbadek: Done in 5ae9711750d7e43a80fd15b32bcfed700c7e849b

assets/wizards/analytics/views/custom-events/index.js
adekbadek: Done in 5ae9711750d7e43a80fd15b32bcfed700c7e849b

adekbadek:
@laurelfulford 

> It looks like those docs still need to be written

Yes, I've started a draft but will publish when these changes are released.

assets/components/src/text-control/index.js
laurelfulford: Looks good, thanks!

assets/components/src/text-control/style.scss
laurelfulford: Thanks!

assets/wizards/analytics/views/custom-events/index.js
laurelfulford: Thanks Adam!

PR #618, opened 2020-08-04
Disable the MC4WC campaign tracking cookie

philipjohn:
### All Submissions:

A performance concern is cookies that are set for logged out users. These cause variations in requests that lead to too many uncached responses. It seems that the Mailchimp for WooCommerce plugin sets a cookie for every visitor regardless of whether it's actually needed. This change stops that behaviour and helps to reduce the possibility of uncached page views.

Note that I'm introducing a new concept with this too - that of compatibility fixes for known plugins - so thoughts welcome on that as an approach if there's a better way.

1. Load any page of a Newspack site using the MC4WC plugin and observe that a `mailchimp_landing_site` cookie is set
2. Apply this PR, then clear all site data
3. Load the page again and observe that the cookie is no longer set

### Other information:


includes/compat/mailchimp-for-woocommerce.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Newspack setup
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+function newspack_mc4c_remove_cookie() {
```
adekbadek, line 10: Linter has [two issues](https://app.circleci.com/pipelines/github/Automattic/newspack-plugin/575/workflows/1376b9e3-6ca7-43c5-89c6-72aca978a5ed/jobs/2449/parallel-runs/0/steps/0-102) with this file.

includes/compat/mailchimp-for-woocommerce.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * Newspack setup
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
```
adekbadek, line 8: `namespace` declaration seems to make the function unavailable as `"newspack_mc4c_remove_cookie"` for the `add_action` hook:

```
PHP Warning:  call_user_func_array() expects parameter 1 to be a valid callback, function 'newspack_mc4c_remove_cookie' not found or invalid function name […]
```

Removing this line makes the feature work as described.

includes/compat/mailchimp-for-woocommerce.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Disable MC4WC's cookie on Newspack sites to avoid perf issues.
```
adekbadek, line 11: Perhaps to avoid confusion with the MC4WP plugin, the official name of the Mailchimp for WooCommerce plugin should be used? Here an in function name ("newspack_**mc4c**_remove_cookie")

includes/compat/mailchimp-for-woocommerce.php
philipjohn: Ah okay, I think I might not have tested that properly after I moved it from my initial test plugin, sorry. :facepalm:

includes/compat/mailchimp-for-woocommerce.php
philipjohn: Yeah, good call. I'll update the comments and fix that naming mistake too to make it clearer.

includes/compat/mailchimp-for-woocommerce.php
adekbadek: Note: it's still an issue, so I'm adding the changes needed label

adekbadek:
Looks like it won't be needed after all.

adekbadek (approved):

claudiulodro:
Thanks for revisiting this!

PR #622, opened 2020-08-05
Remove un-cacheable ajax call in AMP mode from WP GDPR Cookie Notice

claudiulodro:
### All Submissions:

This PR improves server-side performance of the WP GDPR Cookie Notice plugin. Previously in AMP mode, it would fire an un-cacheable ajax call on every request to determine whether to show the cookie notice. This change removes that call and makes it so that the cookie notice will be indiscriminately shown for users that haven't accepted it.

I've also added WP GDPR Cookie Notice to the list of managed plugins, since it was missing and it's our approved cookie notice solution.

0. Install and activate WP GDPR Cookie Notice. If you don't have `https` on your local environment, you'll need to modify [this line](https://github.com/felixarntz/wp-gdpr-cookie-notice/blob/master/src/cookie-notice/class-cookie-notice-amp-markup.php#L64) to `'checkConsentHref' =>  str_replace( 'http:', '', add_query_arg( 'action', self::AMP_CHECK_CONSENT_HREF_ACTION, admin_url( 'admin-ajax.php' ) ) )`, otherwise the notice won't work on your local environment.
1. Before applying the patch, in an incognito window with the network inspector open, view the site in AMP mode. Observe the notice sends an ajax request.
Screen Shot 2020-08-05 at 12 02 04 PM
 
2. Apply the patch. In an incognito window with the network inspector open, view the site in AMP mode. Observe no ajax request is sent. Accept the notice (this will send an ajax request to update preferences, but that should be OK). Verify the notice remembers that you've accepted on subsequent page loads.
Screen Shot 2020-08-05 at 12 02 36 PM

### Other information:

### DIFF
--- a/includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php
+++ b/includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Overrides Cookie_Notice_AMP_Markup to remove ajax call.
+ *
+ * @package Newspack
+ */
+
+use Felix_Arntz\WP_GDPR_Cookie_Notice\Cookie_Notice\Cookie_Notice_AMP_Markup;
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Modifies the <amp-consent> config for better performance on high-traffic sites.
+ */
+class Newspack_Cookie_Notice_AMP_Markup extends Cookie_Notice_AMP_Markup {
+
+	/**
+	 * Modifies the <amp-consent> config to eliminate the ajax call sent on every request.
+	 * The notice will then always show to people that haven't accepted it yet.
+	 *
+	 * @return array Data to pass to the `<amp-consent>` element.
+	 */
+	protected function get_consent_data() {
+		return [
+			'consentInstanceId' => self::AMP_INSTANCE,
+			'consentRequired'   => true,
+			'promptUI'          => 'wp-gdpr-cookie-notice',
+
+		];
+	}
+}
--- a/includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php
+++ b/includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php
@@ -5,9 +5,9 @@
  * @package Newspack
  */
 
-use Felix_Arntz\WP_GDPR_Cookie_Notice\Cookie_Notice\Cookie_Notice_AMP_Markup;
-
 namespace Newspack;
+
+use Felix_Arntz\WP_GDPR_Cookie_Notice\Cookie_Notice\Cookie_Notice_AMP_Markup;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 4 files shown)


includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * Overrides Cookie_Notice_AMP_Markup to remove ajax call.
+ *
+ * @package Newspack
+ */
+
+use Felix_Arntz\WP_GDPR_Cookie_Notice\Cookie_Notice\Cookie_Notice_AMP_Markup;
+
+namespace Newspack;
```
dkoo, line 10: The namespace declaration needs to go before the `use` statement, otherwise it triggers a fatal for me:

```
PHP Fatal error:  Uncaught Error: Class 'Cookie_Notice_AMP_Markup' not found in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-plugin/includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php:17
```

includes/amp-enhancements/class-wp-gdpr-cookie-notice-performance.php
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * AMP improvements for WP GDPR Cookie Notice.
+ *
+ * @package Newspack
+ */
+
+use Felix_Arntz\WP_GDPR_Cookie_Notice\Shortcodes\WordPress_Shortcode_Parser;
+use Felix_Arntz\WP_GDPR_Cookie_Notice\Settings\Plugin_Option_Reader;
+
+namespace Newspack;
```
dkoo, line 11: This one isn't triggering a fatal for me, for whatever reason, but my understanding is that namespace declarations should always come first, right?

includes/amp-enhancements/class-newspack-cookie-notice-amp-markup.php
claudiulodro: Good catch. Fixed in https://github.com/Automattic/newspack-plugin/pull/622/commits/c07874d2b9f226b174a69d83760e04a68a42bee3. I wasn't able to reproduce that issue. Just curious, what PHP version are you using?

dkoo: My environment is on PHP 7.4.1.

includes/amp-enhancements/class-wp-gdpr-cookie-notice-performance.php
claudiulodro: Yep, you're correct. This bit wasn't triggering an error because those namespaces aren't actually used any more (they were from an earlier iteration). :) I've removed them in https://github.com/Automattic/newspack-plugin/pull/622/commits/6db2a0c38e5543f3f972a9bb0acb441a6940a53e

claudiulodro:
@dkoo I've made the changes you requested. Thanks for the speedy review!

includes/amp-enhancements/class-wp-gdpr-cookie-notice-performance.php
dkoo: Makes sense, thanks!

PR #644, opened 2020-08-24
Append categories, tags data to WC order; sync w/ SF

### LINKED ISSUE #636: Salesforce: Report prior page information from WooCommerce checkout to SF sync
Add information about prior page's categories & tags to WC order and SF Opportunity. This way, a publisher can know what categories/tags are generating most donations.

adekbadek:
### All Submissions:

Closes #636

Also fixes two tiny issues: 
1. Checking of webhook exists before removing it
2. Removal of setting `HasOptedOutOfEmail` on a SF Contact. It looks like it would never be set, because an object in `$data['meta_data']` array would never have `mailchimp_woocommerce_is_subscribed` key – it would be the value of such object's `value` key. Moreover, if this field is not configured in SF, [a contact will not be created](https://www.petefreitag.com/item/730.cfm). Open for discussion, though. cc @dkoo 

1. Using a Donate block on a page which has at least one tag and a category, make a donation (e.g. set up a Campaign w/ Donate block)
2. Observe the tags and categories are added to the WC order ("Custom Fields" section of Order details)
3. Observe the Opportunity in SF (on a single Contact view, Opportunities are listed on the right side) – it should have categories an tags listed as "Lead Source"
4. Try also with donating from a page with no categories and tags, observe the information is not appended to a WC order, and that the "Lead Source" in SF lists categories and tags as "none"

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -27,6 +27,8 @@ class Donations {
 	public static function init() {
 		if ( ! is_admin() ) {
 			add_action( 'wp_loaded', [ __CLASS__, 'process_donation_form' ], 99 );
+			add_action( 'woocommerce_checkout_update_order_meta', [ __CLASS__, 'woocommerce_checkout_update_order_meta' ] );
+			add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'woocommerce_billing_fields' ] );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


dkoo:
> Removal of setting HasOptedOutOfEmail on a SF Contact. It looks like it would never be set, because an object in $data['meta_data'] array would never have mailchimp_woocommerce_is_subscribed key – it would be the value of such object's value key. Moreover, if this field is not configured in SF, a contact will not be created. Open for discussion, though. cc @dkoo

I'm okay with this—I was trying to capture the value of the `Subscribe to our newsletter, please` checkbox in WC's checkout page, but if it's causing issues, I don't think it's a super important piece of data to sync.

PR #655, opened 2020-09-07
Add author, word count, publish date custom dimensions

adekbadek:
### All Submissions:

Add additional custom dimensions.

1. On a site with GA configured, visit Analytics wizard, Custom Dimensions tab
2. Observe that a custom dimension in the table has a select with five options – four custom dimensions and a "none" option
3. If there are less than four custom dimensions, add some using the form in "Create a new custom dimension" section
4. Set the custom dimensions to their roles, so that all roles are covered
5. Visit an article as a non-logged in user – observe the custom dimensions are reported with correct values (look for a request to a `/collect` endpoint) as `cdN` parameters, where `N` is the index of the custom dimension
6. In the settings, observe that duplicate roles for custom dimensions are not possible (if dimension A is set to role X, setting dimension B to role X will set the role of custom dimension A to "none") 

### Other information:

### DIFF
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -93,7 +93,7 @@ class Analytics {
 				if ( 'publish_date' === $dimension_role ) {
 					self::add_custom_dimension_to_ga_config(
 						$dimension_id,
-						date_format( date_create( $post->post_date ), 'Y-m-d' )
+						get_the_time( 'Y-m-d H:i', $post->ID )
 					);
 				}
 			}
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -33,7 +33,7 @@ class Analytics {
 	 */
 	public function __construct() {
 		add_filter( 'googlesitekit_amp_gtag_opt', [ __CLASS__, 'inject_amp_events' ] );
-		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'handle_category_reporting' ] );
+		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'handle_custom_dimensions_reporting' ] );
 		add_action( 'wp_footer', [ __CLASS__, 'inject_non_amp_events' ] );
 		add_filter( 'render_block', [ __CLASS__, 'prepare_blocks_for_events' ], 10, 2 );
 		add_action( 'newspack_campaigns_before_campaign_render', [ __CLASS__, 'set_campaign_render_context' ], 10, 1 );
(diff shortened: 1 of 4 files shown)


includes/class-analytics.php
```diff
@@ -47,49 +47,85 @@ public function __construct() {
...
+					self::add_custom_dimension_to_ga_config(
+						$dimension_id,
+						count( explode( ' ', wp_strip_all_tags( $post->post_content ) ) )
+					);
+				}
+
+				if ( 'publish_date' === $dimension_role ) {
+					self::add_custom_dimension_to_ga_config(
+						$dimension_id,
+						date_format( date_create( $post->post_date ), 'Y-m-d' )
```
dkoo, line 96: Just a nitpick and not at all a blocker, but couldn't we just do `get_the_time( 'Y-m-d', $post->ID )`?

includes/class-analytics.php
```diff
@@ -47,49 +47,85 @@ public function __construct() {
...
+						$dimension_id,
+						get_the_author_meta( 'first_name', $author_id ) . ' ' . get_the_author_meta( 'last_name', $author_id )
+					);
 				}
-			);
+
+				if ( 'word_count' === $dimension_role ) {
+					self::add_custom_dimension_to_ga_config(
+						$dimension_id,
+						count( explode( ' ', wp_strip_all_tags( $post->post_content ) ) )
```
dkoo, line 89: This results in a word count that might not match the word count given by Gutenberg in the editor. Is it possible to use the [`@wordpress/wordcount`](https://developer.wordpress.org/block-editor/packages/packages-wordcount/) utility to get a number that matches?

includes/class-analytics.php
```diff
@@ -47,49 +47,85 @@ public function __construct() {
...
-				function ( $gtag_amp_opt ) use ( $categories_slugs, $category_dimension_id ) {
-					$tracking_id = $gtag_amp_opt['vars']['gtag_id'];
-					$gtag_amp_opt['vars']['config'][ $tracking_id ][ $category_dimension_id ] = $categories_slugs;
-					return $gtag_amp_opt;
+
+				if ( 'author' === $dimension_role ) {
+					$author_id = $post->post_author;
+					self::add_custom_dimension_to_ga_config(
+						$dimension_id,
+						get_the_author_meta( 'first_name', $author_id ) . ' ' . get_the_author_meta( 'last_name', $author_id )
```
dkoo, line 82: Should we use `get_the_author_meta( 'display_name', $author_id )` instead?

dkoo (changes requested):
Works well! Tested with AMP and non-AMP, and multiple admin users. I left a few nitpick requests for fetching the data to report, but overall looks good.

includes/class-analytics.php
dkoo: Or more specifically, use whatever method that utility is using so that we get the same word count as is shown in the editor.

includes/class-analytics.php
adekbadek: Good idea, changed in 34b956172e6b0c8b70c06eec93ff46ad987c57a0. Also added time, as suggested in #660

includes/class-analytics.php
adekbadek: Agreed, fixed in 69f974123cc2857eccf616d3ca6d0bec05d9b92c

includes/class-analytics.php
adekbadek: Word count could be a post meta, saved in the editor, but that an overkill in my opinion. There will be slight discrepancies, but they do not matter for the purposes of this feature.

dkoo: Yeah, I don't think it needs to be post meta...I see that the functions to count words in `@wordpress/wordcount` are pretty complex: https://github.com/WordPress/gutenberg/blob/master/packages/wordcount/src/index.js#L113

I was hoping to use whatever counting method they use, but maybe it's overkill to replicate it.

dkoo (approved):

PR #666, opened 2020-09-15
Ignore Yoast weight limit to prevent missing `og:image` tags

claudiulodro:
### All Submissions:

This PR fixes an issue where Yoast would not output `og:image` tags on certain posts. Here is the underlying cause:

1. We're not generating intermediate image thumbs any more (those have been offloaded to the Jetpack image CDN), so only the original full-size image gets uploaded.
2. Therefore, Yoast only finds the full-size image when generating `og:image` tags.
3. Yoast has a max size limit of 2MB for images.
4. Yoast can't know the file size of an image as served from Photon and is unaware of Photon in general, so it checks the original, full-size image size to determine whether the image is under the size limit. The Photon image is resized and compressed, so the size of the original and Photon-served image are wildly different.
5. Yoast thinks the image is too big to use for the `og:image` tag, so it doesn't output the tag.

I think the most straightforward solution here is to ignore the size check. Since Yoast doesn't provide any explanation to the user when it skips generating an `og:image` tag, it will also prevent confusion and reduce the support burden of dealing with questions around this issue.

1. [Get and upload this 4MB image](https://washingtoncitypaper.com/wp-content/uploads/2020/08/WEB_Bowser.5f3d3ff57bb32.png) as the featured image for a post.
2. Delete the intermediate sizes: `wp media regenerate 6950 --image_size=thumbnail` (Replace `6950` with the attachment ID of the image on your site). If you look in the `uploads` directory, you should see only the full-size image and a 150x150 thumbnail after running that command.
3. Before applying this patch, view the page source. Observe no `og:image` tag:
Screen Shot 2020-09-15 at 11 23 43 AM
4. Apply this patch. View the page source. Observe `og:image` tag:
Screen Shot 2020-09-15 at 11 23 21 AM
5. (optional) Do the above steps with Jetpack Image Accelerator active. Should get the same results.

### Other information:

### DIFF
--- a/includes/wizards/class-seo-wizard.php
+++ b/includes/wizards/class-seo-wizard.php
@@ -38,6 +38,7 @@ class SEO_Wizard extends Wizard {
 	public function __construct() {
 		parent::__construct();
 		add_action( 'rest_api_init', [ $this, 'register_api_endpoints' ] );
+		add_filter( 'wpseo_image_image_weight_limit', [ $this, 'ignore_yoast_weight_limit' ] );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jeffersonrabb (approved):
Nice sleuthing.

PR #674, opened 2020-09-25
feat: add AutocompleteTokenfield component

dkoo:
### All Submissions:

As suggested in https://github.com/Automattic/newspack-listings/pull/1#discussion_r492030845, adds the AutocompleteTokenfield component from the Homepage Posts block to newspack-components. This component provides an autocomplete search input field for use in the block editor.

After this PR is released and the NPM package updated, we should be able to remove the code from the `newspack-listings` and `newspack-blocks` repos and import the component from NPM instead.

Not sure how to test the NPM package functionality until this PR has been merged and the package updated on NPM, but you can test that the component works by importing it to an existing editor component in this repo and adding it to the render output, something like this:

```
<AutocompleteTokenField
	key="test"
	tokens={ [] }
	onChange={ value => console.log( value ) }
	fetchSuggestions={ value => console.log( value ) }
	fetchSavedInfo={ value => console.log( value ) }
	label={ 'Test autocomplete search field' }
	help={ __(
		'Begin typing post title, click autocomplete result to select.',
		'newspack-plugin'
	) }
/>
```

The field won't do anything or provide suggestions without actual fetch functions being passed to the props, but seeing the field and being able to interact with it should be enough to confirm that it works as an import from `newspack-components`.

### Other information:

### DIFF
--- a/assets/components/src/autocomplete-tokenfield/index.js
+++ b/assets/components/src/autocomplete-tokenfield/index.js
@@ -1,0 +1,184 @@
+/**
+ * External dependencies
+ */
+import { debounce } from 'lodash';
+
+/**
+ * WordPress dependencies
+ */
+import { Component } from '@wordpress/element';
+import { FormTokenField, Spinner } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+/**
+ * An multi-selecting, api-driven autocomplete input suitable for use in block attributes.
+ */
+class AutocompleteTokenField extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			suggestions: [],
+			validValues: {},
+			loading: this.isFetchingInfoOnLoad(),
+		};
+
+		this.debouncedUpdateSuggestions = debounce( this.updateSuggestions, 500 );
+	}
+
+	/**
+	 * If the component has tokens passed in props, it should fetch info after it mounts.
+	 */
+	isFetchingInfoOnLoad = () => {
+		const { tokens, fetchSavedInfo } = this.props;
+		return Boolean( tokens.length && fetchSavedInfo );
+	};
+
+	/**
+	 * When the component loads, fetch information about the tokens so we can populate
+	 * the tokens with the correct labels.
+	 */
+	componentDidMount() {
+		if ( this.isFetchingInfoOnLoad() ) {
+			const { tokens, fetchSavedInfo } = this.props;
+
+			fetchSavedInfo( tokens ).then( results => {
+				const { validValues } = this.state;
+
+				results.forEach( suggestion => {
+					validValues[ suggestion.value ] = suggestion.label;
+				} );
+
+				this.setState( { validValues, loading: false } );
+			} );
+		}
+	}
+
+	/**
+	 * Clean up any unfinished autocomplete api call requests.
+	 */
+	componentWillUnmount() {
+		delete this.suggestionsRequest;
+		this.debouncedUpdateSuggestions.cancel();
+	}
+
+	/**
+	 * Get a list of labels for input values.
+	 *
+	 * @param {Array} values Array of values (ids, etc.).
+	 * @return {Array} array of valid labels corresponding to the values.
+	 */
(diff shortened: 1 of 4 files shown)


jeffersonrabb:
(Adding @thomasguillot as a reviewer since this is a new component)

thomasguillot:
Is it going to be used within the Newspack Plugin? or is it just because it's convenient to then share it with Blocks and Listing?

If we are going to end up using it for Newspack Plugin then it needs styling. If it's just for conveniency then I'm not sure how I feel to add non-needed component to Newspack Plugin.

Also bear in mind that the style in Newspack Plugin is slightly different than in the editor/Gutenberg.

assets/components/src/autocomplete-tokenfield/index.js
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * Render.
+	 */
+	render() {
+		const { help, label = '', maxLength } = this.props;
+		const { suggestions, loading } = this.state;
+
+		return (
+			<div className="autocomplete-tokenfield">
```
adekbadek, line 168: The classname should be prefixed with `newspack-` as in the other components.

adekbadek (changes requested):

dkoo:
> Is it going to be used within the Newspack Plugin? or is it just because it's convenient to then share it with Blocks and Listing?

My understanding is that because we're using it in both Blocks and Listings, it demonstrates reusability that warrants putting it in the NPM package. I don't know of any plans to use it in the main plugin yet, but it's certainly a future possibility. Other thoughts from @adekbadek?

assets/components/src/autocomplete-tokenfield/index.js
dkoo: Done in b29c0ce.

assets/components/src/autocomplete-tokenfield/index.js
```diff
@@ -0,0 +1,184 @@
...
+				<FormTokenField
+					value={ this.getTokens() }
+					suggestions={ suggestions }
+					onChange={ tokens => this.handleOnChange( tokens ) }
+					onInputChange={ input => this.debouncedUpdateSuggestions( input ) }
+					label={ label }
+					maxLength={ maxLength }
+				/>
+				{ loading && <Spinner /> }
+				{ help && <p className="autocomplete-tokenfield__help">{ help }</p> }
```
adekbadek, line 178: The classname should also be updated here ;)

assets/components/src/autocomplete-tokenfield/style.scss
```diff
@@ -0,0 +1,32 @@
+.autocomplete-tokenfield {
```
adekbadek, line 1: The classname should also be updated here, and on line 22

adekbadek (changes requested):

adekbadek:
> Is it going to be used within the Newspack Plugin? or is it just because it's convenient to then share it with Blocks and Listing?

We have no other place to put shared JS code, and the package is called `newspack-components`. Not using `AutocompleteTokenField` in Newspack Plugin might be an argument for separating `newspack-components` to a repository, but I'm not sure it's worth the effort.

adekbadek (approved):

assets/components/src/autocomplete-tokenfield/index.js
dkoo: Doh!

assets/components/src/autocomplete-tokenfield/style.scss
dkoo: Doh!

dkoo:
@thomasguillot Do you have any other thoughts or objections about this change? We already have three instances in other plugins where this could be used:

* Newsletters
* Blocks
* Listings

This is a really useful component that I see getting more use in other Newspack plugins. I also can see the argument for making `newspack-components` its own repository, but I'll defer to you and @adekbadek on that.

thomasguillot (approved):
@dkoo let's add it to the plugin for now and we can revisit `newspack-components` later on if we decide to.

PR #679, opened 2020-10-07
fix: increase timeout of Salesforce API requests to 30s

dkoo:
### All Submissions:

Increases the timeout limit of Salesforce API requests to 30 seconds, from the default of 5.

This should help avoid intermittent timeout errors due to slow request roundtrips with Salesforce that result in unsynced or partially synced data.

Also shouldn't noticeably affect front-end or admin performance because the requests are triggered by webhooks.

1. Test the Salesforce donation sync functionality and watch the webhooks logs in **WooCommerce > Status > Logs**. 
2. There shouldn't be any timeout errors on sync attempt.

### Other information:

### DIFF
--- a/includes/class-salesforce.php
+++ b/includes/class-salesforce.php
@@ -240,6 +240,7 @@ class Salesforce {
 				'headers' => [
 					'Authorization' => 'Bearer ' . $salesforce_settings['access_token'],
 				],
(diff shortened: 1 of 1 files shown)


claudiulodro (approved):
The timeout increase makes sense, since we're relying on a third-party REST API within our REST API. The code looks good, though I haven't tested it because I'm not set up with Salesforce yet.

PR #688, opened 2020-10-28
feat: allow gam scripts in amp pages

jeffersonrabb:
### All Submissions:

This PR creates the basic infrastructure for AMP Plus — an approach where AMP Dev Mode is used to add non-AMP features to AMP pages. The resulting page will be invalid AMP and therefore excluded from the AMP cache and Google search results carousel, however nearly all of the performance advantages of AMP are still present (tree-shaking, minification, various JS libraries, etc). 

The first experiment is Google Ad Manager. The integration allows use of non-AMP Ad Manager embeds in an AMP page. 

1. Switch Newspack Ads to this branch: https://github.com/Automattic/newspack-ads/pull/81
2. Edit `wp-config.php` and add this line:
```
define( 'NEWSPACK_AMP_PLUS_CONFIG', [ 'gam' ] );
```
3. Add some GAM ads to the site. 
4. Put AMP plugin in Standard mode.
5. Visit a page on the site, append `ampplus` query param to the url (e.g. `https://mywpsite.com/?ampplus`).
6. Verify the page has all the characteristics of an AMP page (minified CSS, AMP scripts) but the ads are not instances of `amp-ad`. 
7. Remove the `ampplus` query param, verify the page is now valid AMP and contains `amp-ad` components.

### Other information:

### DIFF
--- a/includes/class-amp-enhancements.php
+++ b/includes/class-amp-enhancements.php
@@ -29,7 +29,6 @@ class AMP_Enhancements {
 				return $gtag_opt;
 			}
 		);
-		add_filter( 'should_use_amp_plus', [ __CLASS__, 'amp_plus_mode' ] );
 		add_filter( 'amp_content_sanitizers', [ __CLASS__, 'amp_content_sanitizers' ] );
 	}
 
@@ -39,15 +38,12 @@ class AMP_Enhancements {
 	 * @param  string $context The context for which AMP plus should be assessed.
 	 * @return bool Should AMP plus be applied.
 	 */
-	public static function amp_plus_mode( $context = null ) {
-		if ( ! isset( $_GET['ampplus'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-			return false;
+	public static function should_use_amp_plus( $context = null ) {
+		$should = false;
+		if ( isset( $_GET['ampplus'] ) && defined( 'NEWSPACK_AMP_PLUS_CONFIG' ) && is_array( NEWSPACK_AMP_PLUS_CONFIG ) && in_array( $context, NEWSPACK_AMP_PLUS_CONFIG ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			$should = true;
 		}
-		if ( ! defined( 'NEWSPACK_AMP_PLUS_CONFIG' ) ) {
-			return false;
-		}
-		$config = (array) NEWSPACK_AMP_PLUS_CONFIG;
-		return in_array( $context, $config );
+		return apply_filters( 'should_use_amp_plus', $should, $context );
 	}
 
 	/**
@@ -56,7 +52,7 @@ class AMP_Enhancements {
 	 * @param array $sanitizers The array of sanitizers, 'MyClassName' => [] // array of constructor params for class.
 	 */
 	public static function amp_content_sanitizers( $sanitizers ) {
-		if ( self::amp_plus_mode( 'gam' ) ) {
+		if ( self::should_use_amp_plus( 'gam' ) ) {
 			require_once NEWSPACK_ABSPATH . 'includes/amp-sanitizers/class-amp-sanitizer-gam.php';
 			$sanitizers['AMP_Sanitizer_GAM'] = [];
 		}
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
@westonruter I'd be curious to hear your thoughts on this implementation and the "AMP Plus" concept in general.

jeffersonrabb:
> This appears to be the same idea that was explored in Jetpack recently

Yep, this work is largely inspired by/based on https://github.com/Automattic/jetpack/pull/16901. Is the AMP plugin filter you proposed in that comment actually planned or is it speculative at this point?

westonruter:
It's still speculative. It ended up not being needed for those Jetpack features since they could all be implemented in an AMP-compatible way. So I didn't move forward with that filter in the plugin at that time. 

If it is important for your use case, then we can continue to investigate.

The important thing is for the `amp` attribute to be removed when dev mode is being forced. Otherwise,.the site will get tons of search console complaints about invalid AMP.

jeffersonrabb:
> It's still speculative. It ended up not being needed for those Jetpack features since they could all be implemented in an AMP-compatible way. So I didn't move forward with that filter in the plugin at that time.
> 
> If it is important for your use case, then we can continue to investigate.
> 
> The important thing is for the `amp` attribute to be removed when dev mode is being forced. Otherwise,.the site will get tons of search console complaints about invalid AMP.

In this PR the mixed rendering only occurs if a config constant is set and a query parameter is added to the URL. This will allow us to do some functionality testing and comparative performance analysis without disrupting production sites. If the outcome of the analysis is positive, then we'll look to lift the protections and use this in production. At that point we would need a way to remove the `amp` attribute. Should I open an issue in the AMP plugin repo?

westonruter:
I've filed an issue: https://github.com/ampproject/amp-wp/issues/5549.

includes/class-amp-enhancements.php
```diff
@@ -29,6 +29,38 @@ function ( $gtag_opt ) {
 				return $gtag_opt;
 			}
 		);
+		add_filter( 'should_use_amp_plus', [ __CLASS__, 'amp_plus_mode' ] );
```
claudiulodro, line 32: This logic that determines whether we're in AMP plus or not is sort-of hard to follow between the two plugins (Newspack Plugin and Newspack Ads) and I think it'll get confusing as we add more contexts.

I'd recommend a slightly different approach:

```
public static function should_use_amp_plus( $context = null ) {
    $should = false;

    if ( isset( $_GET['ampplus'] ) && defined( 'NEWSPACK_AMP_PLUS_CONFIG' ) && is_array( NEWSPACK_AMP_PLUS_CONFIG ) && in_array( $context, $config ) ) {
        $should = true;
    }

    return apply_filters( 'should_use_amp_plus', $should, $context );
}
```

I'm also not sure what the behavior of this should be when no context is provided (e.g. `should_use_amp_plus()`). `null` will never be contained in `NEWSPACK_AMP_PLUS_CONFIG`, so that'll always return `false`.

claudiulodro (changes requested):
This works great! I left one note about the structure of the code, but otherwise good stuff.

Do you want to keep the `$_GET['ampplus']` flag? It'll be impossible to actually run this feature on a site if we do, because no URL will ever have that flag. :)

includes/class-amp-enhancements.php
claudiulodro: Related: https://github.com/Automattic/newspack-ads/pull/81#discussion_r520210980

jeffersonrabb:
> Do you want to keep the `$_GET['ampplus']` flag?

The intention was to get this into production so that we could do comparative performance testing of "full" AMP vs. AMP Plus. However, if you plan to use this sooner rather than later, this should be reconsidered. What do you think?

claudiulodro:
Let's do the performance testing then we can remove the GET flag next week. 👍  One week is no big deal.

includes/class-amp-enhancements.php
jeffersonrabb: Much nicer, thanks for the idea! https://github.com/Automattic/newspack-plugin/pull/688/commits/9463501b5fd7ab01e8f8b616ff5cc47c54df0411

claudiulodro (approved):
Looks great!

PR #689, opened 2020-10-29
Add segmentation UI in Campaigns wizard

adekbadek:
### All Submissions:

Segmentation UI for Campaigns plugin.

_Follow testing steps in https://github.com/Automattic/newspack-popups/pull/289_ 

### Other information:

### DIFF
--- a/assets/wizards/popups/views/segmentation/SingleSegment.js
+++ b/assets/wizards/popups/views/segmentation/SingleSegment.js
@@ -1,0 +1,114 @@
+/**
+ * WordPress dependencies.
+ */
+import { useEffect, useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { find } from 'lodash';
+
+/**
+ * Internal dependencies.
+ */
+import { Button, TextControl, CheckboxControl, Router } from '../../../../components/src';
+
+const { NavLink, useHistory } = Router;
+
+const SegmentSettingSection = ( { title, children } ) => (
+	<div className="newspack-campaigns-wizard-segments__section">
+		<h3>{ title }</h3>
+		<div className="newspack-campaigns-wizard-segments__section__content">{ children }</div>
+	</div>
+);
+
+const SegmentsList = ( { segmentId, wizardApiFetch } ) => {
+	const [ name, setName ] = useState( '' );
+	const [ min_posts, setMinPosts ] = useState( 0 );
+	const [ max_posts, setMaxPosts ] = useState( 0 );
+	const history = useHistory();
+
+	const isSegmentValid = name.length > 0;
+
+	const isNew = segmentId === 'new';
+	useEffect( () => {
+		if ( ! isNew ) {
+			wizardApiFetch( {
+				path: `/newspack/v1/wizard/newspack-popups-wizard/segmentation`,
+			} ).then( segments => {
+				const foundSegment = find( segments, ( { id } ) => id === segmentId );
+				if ( foundSegment ) {
+					setName( foundSegment.name );
+					setMinPosts( foundSegment.configuration.min_posts );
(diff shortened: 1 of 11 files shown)


assets/wizards/popups/views/segmentation/SingleSegment.js
```diff
@@ -0,0 +1,114 @@
...
+				if ( foundSegment ) {
+					setName( foundSegment.name );
+					setMinPosts( foundSegment.configuration.min_posts );
+					setMaxPosts( foundSegment.configuration.max_posts );
+				}
+			} );
+		}
+	}, [ isNew ] );
+
+	const saveSegment = () => {
```
dkoo, line 46: @adekbadek do you think we need some light validation here to ensure that `max_posts` is greater than `min_posts` if both are enabled?

assets/wizards/popups/views/segmentation/SingleSegment.js
```diff
@@ -0,0 +1,114 @@
...
+import { useEffect, useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { find } from 'lodash';
+
+/**
+ * Internal dependencies.
+ */
+import { Button, TextControl, CheckboxControl, Router } from '../../../../components/src';
+
+const { NavLink, useHistory } = Router;
```
dkoo, line 13: After running `npm ci && npm run build`, I keep getting a `TypeError: useHistory is not a function` error that prevents me from viewing the Campaigns wizard screens.

dkoo (commented):
@adekbadek I wasn't able to get very far with testing due to the TypeError I mentioned below. I think something might be up with `package-lock.json`? I also tried deleting this and `node_modules` and running `npm install`, but after that I get an error with the `format` import from `@wordpress/date` in `SegmentsList.js`.

assets/wizards/popups/views/segmentation/SingleSegment.js
adekbadek: Can you also try resetting the npm packages in `assets/components`?

dkoo: That fixes the `TypeError`, but I still get the `format` error:

```
Uncaught (in promise) TypeError: Cannot read property 'format' of undefined

SegmentsList.js:72
```

adekbadek: d52a2cc0388f0ecc0a418b6c4b054a67701e1003 should fix it, let me know!

assets/wizards/popups/views/segmentation/SingleSegment.js
dkoo: I see the "save" button is disabled in this case—thanks!

dkoo: Whoops, unresolving because I was wrong (was just missing a title)

dkoo (commented):
This is working for me now. I have one suggestion which is more of a UX question than an issue, but currently it's possible to create segments which will never display a campaign—thinking it might be a quick win to avoid that situation with some front-end validation. Otherwise, looks great to me.

adekbadek:
> currently it's possible to create segments which will never display a campaign—thinking it might be a quick win to avoid that situation with some front-end validation

Good idea, but I think this can be tackled separately, as AFAIK we don't have a consistent strategy for showing form validation errors in the plugin. 
@thomasguillot is there one? How should validation errors be displayed in this UI?

assets/wizards/popups/views/segmentation/SingleSegment.js
adekbadek: (responded in [review comment](https://github.com/Automattic/newspack-plugin/pull/689#issuecomment-720647687))

dkoo (approved):

Issue #704, opened 2020-11-12
Campaigns wizards redesign


thomasguillot:
<img width="1160" alt="v5 - init" src="https://user-images.githubusercontent.com/177929/98974507-e1057a80-250c-11eb-8235-a91b1429562b.png">

<img width="1160" alt="v5 - stuff actived" src="https://user-images.githubusercontent.com/177929/98974510-e19e1100-250c-11eb-84cb-05d61525271a.png">

adekbadek:
All wizards redesign done in #705, only Campaigns wizards redesign left in this issue

thomasguillot:
Partially done and closing this issue since the design is evolving anyway.

Issue #697, opened 2020-11-09
New onboarding UI: Step 0 (Welcome) and Step 1 (Settings)


dkoo:
Based on @thomasguillot's redesigns. Please reassign to me or @adekbadek when designs are ready to be worked on.

thomasguillot:
Figma file has been updated with latest tweaks.

PR #705, opened 2020-11-24
Wizards redesign and cleanup

adekbadek:
### All Submissions:

Part of the redesign outlined in #704. Updates:

1. Wizard navigation redesign.
2. Removal of secondary navigation – it was used only in Engagement wizard's Commenting tab, but all of it's contents can fit into one screen. A secondary navigation is an overkill.
3. Removal of `Checklist` component and PHP handlers – these were not used. 
4. Multiple refactors and smaller code & design improvements

1. Visit all the wizards
2. Observe that the functionality is the same, but layout has changed (according to the design in #704)

### Other information:


thomasguillot:
I pushed a few changes https://github.com/Automattic/newspack-plugin/pull/705/commits/cccaf802e197fec76bc79295b159001d73264fc5

I think there are a bunch of stuff that we can get rid of (maybe in a separate PR).

Like the concept of a `card` we don't really need it anymore. Same with the `grid`.

assets/components/src/plugin-installer/index.js
```diff
@@ -177,15 +176,6 @@ class PluginInstaller extends Component {
 			? __( 'Activate' )
 			: __( 'Install' );
```
dkoo, line 177: This is true on `master` too, but it seems heightened to me with the new design. And maybe now's a good time to change it. 

1. (See the Disqus section for what I'm talking about.) If there's more than one plugin to be installed, should this primary action button say "Install All" instead of just "Install"? With just "install" it's not immediately clear to me how the action differs from clicking "Install" next to one of the individual plugins.

Screen Shot 2020-11-30 at 3 13 34 PM

2. (See the Coral section for what I'm talking about.) If there is is only one plugin to be installed, do we even need the action button, since it will do the same thing as clicking "Install" next to the single plugin?

dkoo (changes requested):
This is looking good, just left a few minor questions which @thomasguillot may want to weigh in on.

assets/components/src/tabbed-navigation/style.scss
```diff
@@ -6,63 +6,48 @@
...
-		margin: 0 0 8px;
-		padding: 0;
-		text-align: center;
-
-		@media screen and ( min-width: 744px ) {
-			flex: 1;
-			margin: 0 2px;
+		// margin-bottom: 10px;
+		width: 100%;
+		@media screen and ( min-width: 500px ) {
```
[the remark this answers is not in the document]
thomasguillot: > @thomasguillot is your intention to use the "dropdown" mobile styles you mocked up for the new onboarding screens, that look like this?

It is, yes, but I'm not too worried about it for now.

assets/components/src/plugin-installer/index.js
thomasguillot: I'm not even sure we need the buttons here.

adekbadek: Removed in 1d4690bd37b54196720977c93e4805951c56b53e

assets/components/src/with-wizard-screen/index.js
```diff
@@ -27,106 +14,73 @@ import classnames from 'classnames';
...
+							<>
+								<TabbedNavigation items={ tabbedNavigation } />
+							</>
+						) }
+					</div>
+				</div>
+
+				<div className={ classnames( 'newspack-wizard newspack-wizard__content', className ) }>
+					{ content }
+					<div className="newspack-buttons-card">
```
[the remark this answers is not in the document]
adekbadek: Good idea, done in 5705408e7e51ab416e6baf5c8af287f5ccd0e370

adekbadek:
@thomasguillot 

> I pushed a few changes cccaf80

Is the white background on the Dashboard intentional?

image

thomasguillot:
> Is the white background on the Dashboard intentional?

For now yes.

To be honest this will need a complete redesign at some point (and the Dashboard might evolve anyway) so it's fine for now.

dkoo (approved):
Changes since my last review look good from a technical standpoint. I'll let @thomasguillot approve from a design perspective. Thanks!

Once this lands I'll also continue working on #697 (I had begun but some of the work I started doing was redundant with this branch).

adekbadek:
> I'll let @thomasguillot approve from a design perspective.

I assume https://github.com/Automattic/newspack-plugin/pull/705#event-4056253696 is an approval ;)

image

PR #706, opened 2020-11-25
Campaigns Wizard: display campaign ID

adekbadek:
### All Submissions:

https://github.com/Automattic/newspack-popups/pull/321 makes it possible to insert campaigns using a shortcode. To make it easier, the wizard will display a campaign ID (which is to be inserted into a shortcode)

1. Visit the Campaigns Wizard, observe that on the options popover visible after clicking the three dots icon, Campaign ID is shown 

### Other information:


jeffersonrabb (approved):
Good idea. If the shortcode approach becomes common we could also add UI to display the campaign's shortcode, to save a step.

PR #707, opened 2020-11-26
Campaigns Wizard: favorite category segmentation

adekbadek:
### All Submissions:

UI for adding  a favourite-category based segmentation. 

Do not merge until https://github.com/Automattic/newspack-popups/pull/322 is merged.

1. Switch Campaign plugin to `add/handle-category-segment` branch (https://github.com/Automattic/newspack-popups/pull/322)
1. Visit the Campaigns Wizard, Segmentation tab. Create or edit a segment
2. Observe a new section: "Category Affinity". It should list all categories, and if there are more than 5, allow for list showing/hiding the whole list with a button
3. Assign a favourite category to the segment, and create a campaign targeted to that segment
4. Verify that the campaign is shown only if the reader did most of their reading in the selected category
5. Now create a segment with multiple favourite categories
6. Verify that the campaign is displayed only if the reader did most of their reading in one of the selected categories
7. Create a segment combining favourite categories with other conditions, like minimum amount of posts read
8. Verify the campaign is shown only of all segment conditions are met
9. Create a segment without favourite categories, verify it works as expected (just to make sure an empty favourite categories set will not break things)

### Other information:


jeffersonrabb (changes requested):
> Assign a favourite category to the segment, and create a campaign targeted to that segment

I tested with an overlay campaign, and it only worked if `Sitewide Default` is toggled on or if the Campaign has the same category as the segment assigned. I understand why this is, since the campaign wouldn't be rendered on the page at all, but as a user it was unexpected. I would want to be able to create multiple category affinity overlays and have them all work by simply assigning the segment, without having to worry about also assigning categories to the campaign. Any thoughts on how to resolve? I also get this could mean we start having lots and lots of campaign markup on posts, most of which will never be shown.

assets/wizards/popups/views/segmentation/SingleSegment.js
```diff
@@ -130,16 +133,49 @@ const SegmentsList = ( { segmentId, wizardApiFetch } ) => {
...
+				<SegmentSettingSection
+					title={ __( 'Category Affinity', 'newspack' ) }
+					description={ __( 'Most read categories of reader.', 'newspack' ) }
+				>
+					{ take(
+						siteCategories.map( category => {
+							const selected = segmentConfig.favorite_categories;
+							const id = String( category.id );
+							return (
+								<CheckboxControl
```
jeffersonrabb, line 155: Ideally this UI would use `<CategoryAutocomplete />` like the category filters in the Inline and Overlay screens. This would improve UI consistency and also avoid massively long category lists.

jeffersonrabb (changes requested):

assets/wizards/popups/views/segmentation/SingleSegment.js
adekbadek: Good idea, changed in a848ab73f2f5a3db55d7e39e7820800b6643061f

adekbadek:
> I would want to be able to create multiple category affinity overlays and have them all work by simply assigning the segment

Interesting! I didn't think of that, but it's a very valid user story. I see two possible solutions:

1. Treat segmentation as a third display condition for overlay campaigns – alongside sitewide default and chosen category.
2. Change the meaning of "sitewide default" setting – so that it would be "sitewide default for segment", and instead of a toggle it would be a select in the UI.

> this could mean we start having lots and lots of campaign markup on posts, most of which will never be shown.

Yes, I can't see a way around that since the segmentation is a client-side condition :/

jeffersonrabb:
> * Treat segmentation as a third display condition for overlay campaigns – alongside sitewide default and chosen category.
> * Change the meaning of "sitewide default" setting – so that it would be "sitewide default for segment", and instead of a toggle it would be a select in the UI.

Let's get @thomasguillot's opinion on this, since it's going to impact the UI. I'm wondering if the segmentation features might give us an opening to rethink sitewide default, which has always been an awkward and tough-to-understand setting.

thomasguillot:
> > * Treat segmentation as a third display condition for overlay campaigns – alongside sitewide default and chosen category.
> > * Change the meaning of "sitewide default" setting – so that it would be "sitewide default for segment", and instead of a toggle it would be a select in the UI.
> 
> Let's get @thomasguillot's opinion on this, since it's going to impact the UI. I'm wondering if the segmentation features might give us an opening to rethink sitewide default, which has always been an awkward and tough-to-understand setting.

Are we overthinking this? I thought we kind of wanted to minimise the overlay campaigns since they are not best practice.

jeffersonrabb:
> I thought we kind of wanted to minimise the overlay campaigns since they are not best practice.

I don't think this is established at all. Overlays tend to be highly effective, although they risk alienating the reader. A better way to think of this is that we want to encourage and/or enforce best practices in overlay usage, not that we want to minimi(z|s)e their use.

adekbadek:
> A better way to think of this is that we want to encourage and/or enforce best practices in overlay usage, not that we want to minimi(z|s)e their use.

Makes sense. 

Of the two options I've proposed I think the first makes more sense. The default is still the default, but alongside campaign-category we introduce a reader-category-segment display condition. The latter would take precedence over the former. 

Given three overlay campaigns: 

1. A – sitewide default
2. B – with "Sports" category set
3. C – with a "Politics category affinity" segment set

On a post of "Sports" category visited by a reader with "Politics" category affinity, campaign C will be displayed. For a reader with different category affinity, B campaign will be displayed.

WDYT @jeffersonrabb @thomasguillot ?

jeffersonrabb:
Thinking more on this, the discussion of how to handle Sitewide default is out of scope of this PR. As all of the segmentation features come into focus we'll probably have to give more thought to questions of campaign precedence and fallbacks anyway, and we can wrap the sitewide default question into that. Let's merge this and move that conversation elsewhere...

jeffersonrabb (approved):

PR #709, opened 2020-11-26
Campaigns Wizard: display segment reach

adekbadek:
### All Submissions:

Display segment reach – what percentage of recorded audience the segment covers.

Do not merge before https://github.com/Automattic/newspack-popups/pull/324 is merged.

1. Switch Campaign plugin to `add/segment-size` branch (https://github.com/Automattic/newspack-popups/pull/324)
1. Visit the Campaigns Wizard's Segmentation tab and edit or create a segment
2. Observe the segment reach is displayed above the buttons section
3. Make some changes to the segment, observe the reach is updated dynamically, and that the update does not block the UI

### Other information:

### DIFF
--- a/assets/wizards/popups/views/segmentation/SingleSegment.js
+++ b/assets/wizards/popups/views/segmentation/SingleSegment.js
@@ -1,9 +1,10 @@
 /**
  * WordPress dependencies.
  */
-import { useEffect, useState } from '@wordpress/element';
+import { useMemo, useEffect, useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
-import { find } from 'lodash';
+import apiFetch from '@wordpress/api-fetch';
+import { find, debounce } from 'lodash';
 
 /**
  * Internal dependencies.
@@ -39,6 +40,8 @@ const SegmentsList = ( { segmentId, wizardApiFetch } ) => {
 	const updateSegmentConfig = key => value =>
 		setSegmentConfig( { ...segmentConfig, [ key ]: value } );
 	const [ name, setName ] = useState( '' );
+	const [ isFetchingReach, setIsFetchingReach ] = useState( { total: 0, in_segment: 0 } );
+	const [ reach, setReach ] = useState( { total: 0, in_segment: 0 } );
 	const history = useHistory();
 
 	const isSegmentValid = name.length > 0;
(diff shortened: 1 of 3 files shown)


assets/wizards/popups/views/segmentation/SingleSegment.js
```diff
@@ -130,17 +146,29 @@ const SegmentsList = ( { segmentId, wizardApiFetch } ) => {
...
 						label={ __( "Hasn't donated", 'newspack' ) }
 					/>
 				</SegmentSettingSection>
 			</div>
+
+			{ reach.total > 0 && (
+				<div style={ { opacity: isFetchingReach ? 0.5 : 1 } }>
+					{ __( 'This segment would reach', 'newspack' ) }{' '}
+					<b>
+						{ Math.round( ( reach.in_segment * 100 ) / reach.total ) }
```
jeffersonrabb, line 165: Could we show the actual segment size as well as the percentage of all users?

jeffersonrabb (changes requested):

assets/wizards/popups/views/segmentation/SingleSegment.js
adekbadek: I was wondering about that, but I think it might be misleading – the actual audience size will be different, so this is just an approximation. For "real" numbers the users should go to GA. WDYT?

jeffersonrabb: Why would the actual audience size be different?

adekbadek: Only last 30 days worth of data is kept – based on that we can assume the percentage (e.g. 10% of them were subscribers, so it's likely that 10% of _all_ readers are), but when presenting an "actual segment size", we'd have mention that it's just based on last 30 days' visits. By "real" numbers I meant that whole record, not just last 30 days.

jeffersonrabb: Ok, not a big deal. We can revisit if anyone questions it.

jeffersonrabb (approved):
Looks good and works well. I have mild concerns which needn't block the PRs but I'll note them here in case they come later:

- As noted in a comment we may hear that users expect to see the numeric counts of segment sizes, not just the percentages.
- For a publication with a large body of data in the `newspack_campaigns_events` table the database operation to retrieve and process them all could be quite a whopper. It's only called from `wp-admin`, and then only when playing around with segments, but worth keeping an eye on in case this leads to scale issues.

PR #713, opened 2020-12-01
Fix freudian slip

adekbadek:
Updating React is always a good idea, but this is about something else.


jeffersonrabb (approved):
I was wondering about this! Seemed like an unusually bold naming convention, but then you're an unusually bold developer...

PR #726, opened 2020-12-03
NRH Integration

jeffersonrabb:
### All Submissions:

This adds a "first class" [News Revenue Hub](https://fundjournalism.org/) integration in the Reader Revenue wizard. For sites using NRH checkout, we install [Newspack NRH Checkout](https://github.com/Automattic/newspack-nrh-checkout), which redirects the Donate block's form submission to NRH checkout with the appropriate post data. For this flow to work, these sites must have all of the Newspack donation plugins activated (WooCommerce, WooCommerce Subscriptions, WooCommerce Stripe Gateway, WooCommerce Name Your Price), which places an unnecessary burden on these sites since none of these are actually used. In addition to a cleaner setup for NRH customers, this branch allows these sites to operate without these plugins. In an upcoming PR I'm going to add support for cross-domain tracking so that NRH customers can capture analytics events on their donation page, and the UI work in this PR creates a logical home for this feature. Finally, this PR paves the way for integrations of other reader revenue tools that are in use on some Newspack sites.

On the first screen of the wizard a new select allows choosing between Newspack (WooCommerce) and News Revenue Hub approaches. If "News Revenue Hub" is selected, a few specific fields will be shown and none of the WC-dependent features will be. If "Newspack" is selected the flow will be roughly the same as it has been all along, with the exception of plugin installation. Originally, all of the WooCommerce-related plugins were required to use this wizard. Now only Newspack Blocks is required, and the WC ones are installed only after selecting Newspack.

There are also a few changes to the Donate block. If NRH Checkout is selected and the WooCommerce plugins are disabled, price tiers in the block must be configured manually. The "Configure Manually" toggle is hidden in this scenario and forced to be on. The only feature which cannot yet be configured manually is the currency symbol. Since all Newspack/NRH customers are U.S. based, I hard-coded the dollar sign. In the future we can make this selectable.

Test along with https://github.com/Automattic/newspack-blocks/pull/642

Closes https://github.com/Automattic/newspack-blocks/issues/623, https://github.com/Automattic/newspack-blocks/issues/518

To test you will need a News Revenue Hub client ID. You can get one by contacting me or from `Settings->Newspack NRH Checkout` on any NRH-enabled site. The key to testing this is ensuring that checkout works correctly in all five of these scenarios:

1. Regression one: Normal Newspack approach, after switching to this branch and doing nothing in the Reader Revenue wizard.
2. Regression two: NRH checkout using Newspack NRH Checkout after switching to this branch and doing nothing in the Reader Revenue wizard.
3. With all of the WooCommerce plugins disabled, go to Reader Revenue wizard, select Newspack and go through the setup steps. Ideally, start by resetting all Newspack options, which can be done in Settings->Newspack by checking "Reset Newspack" and clicking "Save Changes."
4. Remove the Newspack NRH Checkout plugin. After setting up WooCommerce donations, switch to NRH Checkout, and ensure it works properly. 
5. Disable WooCommerce plugins. You will need to set price tiers in the Donate block manually, then verify everything is still working.

If NRH checkout is selected and WooCommerce disabled, the Donate block will only appear if configured manually. 

### Other information:


thomasguillot (changes requested):
Screenshot 2020-12-04 at 15 38 15

Not really a fan of a select here. Especially because I can click on "Select platform* which returns an error.

Instead I'd suggest to switch to radio buttons and set it by default to "Newspack". I'd also suggest to move the Donations and Salesforce to its own tabs (it looks a bit odd/ugly but it will be "fixed" once we have the new design merged)

Screenshot 2020-12-04 at 15 46 50

jeffersonrabb:
@thomasguillot Here's the problem with making Newspack the default setting. If Newspack is the selection, then we need to install all of those WooCommerce plugins before the Wizard can be used. However, if NRH is the selection then you probably don't want them installed (I say "probably" only because theoretically WooCommerce could be used to sell merchandise on a site using NRH for donations, but it's a edge case for sure). So, if we make Newspack the default, that requires that we install those plugins before the wizard can be used.

jeffersonrabb:
> Especially because I can click on "Select platform* which returns an error.

This is fixed in https://github.com/Automattic/newspack-plugin/pull/726/commits/0b6cbb30c4a38dd5bf8242f41afe7e285c0cfa85

includes/class-donations.php
```diff
@@ -329,7 +341,7 @@ protected static function update_donation_product( $args ) {
 	 */
 	public static function remove_donations_from_cart() {
 		$donation_settings = self::get_donation_settings();
-		if ( ! $donation_settings['created'] ) {
+		if ( ! $donation_settings['created'] || is_wp_error( self::is_woocommerce_suite_active() ) ) {
```
dkoo, line 344: This line is throwing a warning if WooCommerce is active, but the Reader Revenue platform is set to NRH:

```
Cannot modify header information - headers already sent by (output started at /srv/users/user53d5f4dc/apps/user53d5f4dc/public/wp-content/plugins/newspack-plugin/includes/class-donations.php:349)
Undefined index: products in /srv/users/user53d5f4dc/apps/user53d5f4dc/public/wp-content/plugins/newspack-plugin/includes/class-donations.php on line 349
```

includes/wizards/class-reader-revenue-wizard.php
```diff
@@ -528,6 +595,19 @@ protected function fetch_all_data() {
 	public function api_get_donation_settings() {
 		$required_plugins_installed = $this->check_required_plugins_installed();
 		if ( is_wp_error( $required_plugins_installed ) ) {
+			if ( 'nrh' === get_option( NEWSPACK_READER_REVENUE_PLATFORM ) ) {
+				return rest_ensure_response(
+					[
+						'created'                 => true,
+						'suggestedAmounts'        => [ 0, 0, 0 ],
```
dkoo, line 602: Could we hardcode some suggested tiers here, maybe? Or even give users the option to set those tiers in Reader Revenue when the NRH UI is active?

includes/wizards/class-reader-revenue-wizard.php
```diff
@@ -508,16 +550,41 @@ public function api_check_salesforce_token() {
 	 * @return Array
 	 */
 	protected function fetch_all_data() {
+		$platform                 = get_option( NEWSPACK_READER_REVENUE_PLATFORM );
```
dkoo, line 553: Is there any chance this option could be undefined at this point, and would that cause any issues with no default value? It looks like this endpoint would basically return the default `$args`, right?

dkoo (changes requested):
I'm not sure whether this would be a real-world scenario, but this extreme edge case tripped me up a bit in testing. To replicate:

1. Set your platform as Newspack and install all required WooCommerce plugins.
2. In a new session, click Donate now! from a Donate block to land on the WC checkout page, but don't complete the transaction (leaving an open item in the cart).
3. Set the platform as NRH but _don't_ deactivate the WooCommerce plugins.
4. In the same session where you got to the WC checkout page, go back to a Donate block and click Donate now!
5. Observe a blank page with PHP errors:

Screen Shot 2020-12-04 at 3 38 59 PM

I think this is a fairly unlikely scenario in the real world. It seems to only happen if the user has already a WC item in the cart, then the platform changes to NRH and the user tries to donate again. The cart item removal at https://github.com/Automattic/newspack-plugin/pull/726/files#diff-535ccef3a7598eb102f05f7818878649003a34a753b0d742adcf7acd3da6e977R344 seems to throw the warnings and the redirection to NRH never happens.

includes/wizards/class-reader-revenue-wizard.php
jeffersonrabb: Since this function is only used by the Reader Revenue Wizard, I think the worst case scenario would be some unusual results in the later screens of the wizard, but definitely worth fixing. Done in https://github.com/Automattic/newspack-plugin/pull/726/commits/6d9aaafc91100ca9f384650e953b2f931d5cf122

includes/wizards/class-reader-revenue-wizard.php
jeffersonrabb: Good idea. [This change](https://github.com/Automattic/newspack-plugin/pull/726/commits/9405be3dcc67eca654f2c0ad89830f7621e3e8b1) provides defaults to the block, and [this commit](https://github.com/Automattic/newspack-blocks/pull/642/commits/4d58bd2f5fdf87f5f58b7cb92d1ef97f3f5d92a4) causes the block to set the default prices if the tiers haven't already been manually defined.

https://github.com/Automattic/newspack-blocks/pull/642/commits/4d58bd2f5fdf87f5f58b7cb92d1ef97f3f5d92a4

includes/class-donations.php
jeffersonrabb: I wasn't able to reproduce this myself, but it makes total sense. https://github.com/Automattic/newspack-plugin/pull/726/commits/9405be3dcc67eca654f2c0ad89830f7621e3e8b1 makes the donation data provided for NRH and WooCommerce much more similar, which should resolve all the warnings.

thomasguillot (changes requested):
> Here's the problem with making Newspack the default setting. If Newspack is the selection, then we need to install all of those WooCommerce plugins before the Wizard can be used. However, if NRH is the selection then you probably don't want them installed (I say "probably" only because theoretically WooCommerce could be used to sell merchandise on a site using NRH for donations, but it's a edge case for sure). So, if we make Newspack the default, that requires that we install those plugins before the wizard can be used.

Ok this makes sense, I didn't think of this...

I'd still like to separate the select from the other settings. Maybe something like:

 ^ h2 might not be needed on init

jeffersonrabb:
> I'd still like to separate the select from the other settings. Maybe something like:

@thomasguillot implemented your suggestions in https://github.com/Automattic/newspack-plugin/pull/726/commits/b301b8a689cfd3815bb60bb3da87be7685c69e89. I took your earlier suggestion that Salesforce and Donations be given their own tabs, and got rid of the `ActionCard`s. I also did away with the first-time flow which steps through the screens one at a time. This was confusing, error-prone, and didn't feel like it was adding much to the experience.

dkoo (changes requested):
Latest changes for the default suggestions and avoiding the PHP warnings look good! Just one more thing I noticed:

There are four required plugins in order to use the "Newspack" platform: WooCommerce, WC Name Your Price, WC Subscriptions, and WC Stripe Gateway. However, Stripe Gateway isn't included in the UI for managing plugins. So if you have the first three plugins installed and activated but not Stripe Gateway, you get this screen with no other tabs:

Screen Shot 2020-12-07 at 11 23 12 AM

Seems like we just need to add Stripe Gateway to the wizard component's plugin requirements.

jeffersonrabb:
> Seems like we just need to add Stripe Gateway to the wizard component's plugin requirements.

Good catch. Fixed in https://github.com/Automattic/newspack-plugin/pull/726/commits/f993109279093ef29f6893ee84b91d8ac052ef8f. Note: I also merged `master` which now includes https://github.com/Automattic/newspack-plugin/pull/705, so the wizard will look substantially different the next time you check it.

dkoo (approved):
Works now! The new UI looks good :)

thomasguillot (changes requested):
This looks great.

A couple of things:

Screenshot 2020-12-08 at 08 32 00

I'd move the Platform to the last menu item. It should be selected only if the user just pick its revenue platform. If not, it should default to Donations (Newspack) or Settings (NHR) -- unless we're missing plugins, we should default to Platform (still last tab). The reason why I don't think it should be the first screen is because it's unlikely that a user will change this once it's set.

Screenshot 2020-12-08 at 08 35 27

Frankly I'm not sure we need the Activate/Install primary button, especially if there is only 1 plugin to activate/install as it'd be redundant with the "Activate/Install O"

jeffersonrabb:
> I'd move the Platform to the last menu item. It should be selected only if the user just pick its revenue platform. If not, it should default to Donations (Newspack) or Settings (NHR) -- unless we're missing plugins, we should default to Platform (still last tab). The reason why I don't think it should be the first screen is because it's unlikely that a user will change this once it's set.

These  changes are made, and I've removed the ACTIVATE button from the plugin installer. @thomasguillot could you take another look?

thomasguillot (approved):
Looks good 🚢

jeffersonrabb:
@dkoo sorry to do this to you but could you give this one last look? @thomasguillot's design requests required a few additional changes in https://github.com/Automattic/newspack-plugin/pull/726/commits/a45cf526c8572cdb13e2b2667d3c06669d2ff677, https://github.com/Automattic/newspack-plugin/pull/726/commits/acc2befe3f257f7bd06a25be24cbdb2ada4911eb, and https://github.com/Automattic/newspack-plugin/pull/726/commits/120f54ab05dd9ce5183da7b2712e4e061355c336

dkoo (approved):
Changes look good and work well. Conflicts need to be resolved before merging.

Also wanted to note:

> It should be selected only if the user just pick its revenue platform

This is not the case. Currently, after selecting the platform, the UI automatically navigates to the platform's default tab (Donation for Newspack, Settings for NRH). I'm totally fine with this, but it doesn't sound like what @thomasguillot described, right?

jeffersonrabb:
> Also wanted to note:
> 
> > It should be selected only if the user just pick its revenue platform
> 
> This is not the case. Currently, after selecting the platform, the UI automatically navigates to the platform's default tab (Donation for Newspack, Settings for NRH). I'm totally fine with this, but it doesn't sound like what @thomasguillot described, right?

Fixed up the redirects in https://github.com/Automattic/newspack-plugin/pull/726/commits/98df697d8c2b8e99178d10d4bb7138cc208e2fd1.

dkoo (approved):
Changes approved!

PR #731, opened 2020-12-09
Update SelectControl and Popover style since WP 5.6

thomasguillot:
### All Submissions:

Completely rework the SelectControl
Remove icons to the Campaigns Popover and fix the SelectControl

Also minor update to:
* ActionCardSections
* FormTokenField
* StyleCard
* TextControl
* TextareaControl

1. Check all the different wizards, especially Campaigns

### Other information:


assets/components/src/select-control/index.js
```diff
@@ -13,19 +13,17 @@ import { Component } from '@wordpress/element';
...
- */
-import classNames from 'classnames';
-
 class SelectControl extends Component {
 	/**
 	 * Render.
 	 */
 	render() {
 		const { className, ...otherProps } = this.props;
-		const classes = classNames( 'newspack-select-control', className );
```
adekbadek, line 27: This breaks select styling tweaks [here](https://github.com/Automattic/newspack-plugin/blob/dce2eb123d38511fa681f935a79937cf51fd286e/assets/wizards/analytics/views/custom-dimensions/index.js#L109) and [here](https://github.com/Automattic/newspack-plugin/blob/dce2eb123d38511fa681f935a79937cf51fd286e/assets/wizards/popups/components/popup-popover/index.js#L91), especially since the base component ignores the `className` prop. Why is this change needed?

adekbadek (changes requested):

assets/components/src/select-control/index.js
thomasguillot: How is this broken? The className is still added to the `BaseComponent`.

Screenshot 2020-12-10 at 10 53 57
Screenshot 2020-12-10 at 12 53 15

adekbadek: > The className is still added to the BaseComponent

That's true – my bad. But still, this change in `className` prop handling makes it harder to style the selects, since the added class name is not applied to the top level component. 
In the Analytics wizard, removing the margins on selects now requires assigning negative margins to the BaseComponent, which creates unnecessary complexity. 
Why was the change in passing the `className` prop introduced? It's customary to pass such a prop to the top-level React element.

thomasguillot: Gutenberg introduced a bunch of wrappers around the actual `Select` with some opinionated CSS that couldn't be targeted with the old `<BaseComponent className={ classes } { ...otherProps } />`. Is there a better way to handle this?

_Note: with the redesign I will need to revisit all the different screens and add more changes anyway, including the Analytics wizard_

adekbadek: > Is there a better way to handle this?

What's the goal here? 
I can see that (in this repo) the only instances of passing a `className` to a `SelectControl` are the ones I've linked to in the first comment. They can be fixed by keeping the classname passing to the top-level component, while still adding a wrapper div, like so: https://github.com/Automattic/newspack-plugin/pull/736

thomasguillot: I don't know what I was thinking 🙃  I merged #736

adekbadek (approved):

PR #733, opened 2020-12-10
Update newspack-popups constant name

adekbadek:
Update constant name after changes in https://github.com/Automattic/newspack-popups/pull/343


jeffersonrabb (approved):

dkoo (approved):
Fixes a fatal, so this and https://github.com/Automattic/newspack-popups/pull/343 must be released simultaneously.

PR #737, opened 2020-12-11
Update npm packages

thomasguillot:
### All Submissions:

This PR updates the different NPM packages to their latest verisons

1. Does it break anything? 😄

### Other information:

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "homepage": "https://newspack.blog",
   "devDependencies": {
     "@automattic/calypso-build": "^6.3.0",
-    "@babel/core": "^7.4.0",
+    "@babel/core": "^7.12.10",
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.3.4",
     "@semantic-release/changelog": "^5.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -132,7 +132,8 @@
     "react-dropzone": "^10.2.1",
     "react-rte": "^0.16.1",
     "recharts": "^1.8.5",
-    "whatwg-fetch": "^3.0.0"
+    "whatwg-fetch": "^3.0.0",
+    "wp-prettier": "git+https://github.com/Automattic/wp-prettier.git#wp-prettier-1.18.2"
```
adekbadek, line 136: Why is this here? We're [already using `wp-prettier` as `prettier`](https://github.com/Automattic/newspack-plugin/blob/82fc7fddd423021bb22238167d0f6fc54d08813a/package.json#L49)

adekbadek (changes requested):

package.json
thomasguillot: 🤔  no idea, I simply ran `npm update`. Should I just remove the line?

adekbadek: Confirmed, that's probably added because of that `prettier` aliasing 🤷‍♀️

adekbadek (approved):

PR #739, opened 2020-12-11
feat: new onboarding menu UI (v3)

### LINKED ISSUE #697: New onboarding UI: Step 0 (Welcome) and Step 1 (Settings)

dkoo:
### All Submissions:

Implements new onboarding UI design as mocked up by @thomasguillot.

Currently, this PR only implements the design changes in step 0 (Welcome screen) and the redesigned pagination UI. Pending a decision on the new onboarding steps, I've kept the onboarding steps the same as they currently are, except for renaming the titles and routes in the cases where the new steps have exact analogues in the existing steps ("About your publication" becomes "Settings", "Site Design" becomes "Design", etc.).

Mobile (I've interpreted the design mockup as a dropdown, but please let me know if this is incorrect):
Screen Shot 2020-12-14 at 11 53 55 AM

Desktop:
Screen Shot 2020-12-14 at 11 54 17 AM

Closes #697.

1. Install the plugin on a new WP site: checkout this branch and run `npm run release:archive` to build a ZIP file that can be installed on a new test site.
2. Go through the onboarding steps. The Welcome screen should display a Newspack logo at the top, and there should no longer be a "Not right now" secondary button here.
4. Proceed through all onboarding steps and confirm that the correct step is highlighted in the new UI in both desktop and mobile viewports.
5. The onboarding process should be the same as on `master`, with the exception that "About your publication" is now called "Settings", "Installation..." is now called "Integrations" and "Site Design" is now called "Design" per the new design mockups.
6. Since we've changed the structure of some route data: Test that if trying to navigate to an invalid onboarding route (e.g. change the hash at the end of the URL to some gibberish), you get redirected back to the Welcome screen.

### Other information:


dkoo:
@thomasguillot As discussed, updated the components per your feedback and will hold this PR until updated wizard designs are ready.

dkoo:
Reopening this PR since @thomasguillot confirmed that the redesigned components affected by this branch are not going to change.

adekbadek (changes requested):
Found one issue - on "Design" step, the menu is not visible: 

image

dkoo:
> Found one issue - on "Design" step, the menu is not visible:

Whoops, fixed in f79c1f6

assets/components/src/wizard-pagination/index.js
```diff
@@ -22,38 +16,93 @@ import './style.scss';
...
-					</Button>
-				</div>
-			</Fragment>
-		);
-	}
-}
+	};
+
+	return (
+		<Fragment>
```
adekbadek, line 47: No need for a `Fragment` since it wraps a single `div`

assets/components/src/wizard-pagination/index.js
```diff
@@ -22,38 +16,93 @@ import './style.scss';
...
+									classes.push( 'active' );
+								}
+
+								if ( index < currentIndex ) {
+									classes.push( 'complete' );
+								}
+
+								return (
+									<>
+										<li key={ index } className="newspack-wizard-pagination__step">
```
adekbadek, line 79: The `key` prop should actually be on the `Fragment` which is the parent of this `li`

adekbadek (changes requested):
Sorry for not catching these in the initial review

assets/components/src/wizard-pagination/index.js
dkoo: Good call, fixed in 01aedc7

assets/components/src/wizard-pagination/index.js
dkoo: True, fixed in 01aedc7

adekbadek (approved):

PR #741, opened 2020-12-15
Campaigns Wizard: "view as" UI

adekbadek:
### All Submissions:

_Note: merge after https://github.com/Automattic/newspack-popups/pull/349 is merged_

1. Switch to branch `add/view-from-group` (https://github.com/Automattic/newspack-popups/pull/349) of `newspack-popups`
2. Create at least two groups, at least two segments, and some related campaigns*
2. Visit the Campaigns Wizard, observe a new tab: "Preview", open it
3. Try out the previews in different configurations of segments and groups
4. Observe that the preview configuration is persisted after a page reload

\* My setup was: "subscribed", "unsubscribed" segments; "newsletter", "election" campaign groups; two "Newsletter" campaigns – a newsletter form segmented to unsubscribed and a "thank you" message for subscribers, an "Election" campaign

### Other information:


dkoo (changes requested):
This is already really powerful and pretty much exactly what I had in mind. There are a few things I found myself wanting during functional testing. CCing @thomasguillot for additional feedback since some of this touches on UX/design.

1. We should consider deprecating the "Test" mode with this PR. I think the new UI is much more useful and makes it obsolete.
2. Once I open the iframe, I see the campaigns I expect to see, matching the Segment and Groups I specified. However, if I navigate away from the first page that opens in the iframe, it seems to lose the params and I see only Test campaigns thereafter.
3. If we can resolve (2) so that the params are passed through for the life of the iframe, I think we can do away with separate buttons to "Preview" vs. "Preview on a post." We could have it open to a post by default, but users can always navigate to archive pages and the homepage if they want to test behavior there.
4. The homepage looks off to me in the iframe. It looks like it's loading with the archive template instead of the usual homepage + homepage posts layout? I think this is also happening in master with the single post previews, but I'm not sure why.
5. The "None" option for Segments isn't doing what I expected:
  - Current behavior: It loads "Test" campaigns (which as mentioned, I don't think is necessary anymore).
  - What I expected: It should show all active campaigns that don't have an assigned segment. Right now I don't think it's possible to preview these campaigns with the UI.
6. I think the "None" option for Segment should be called "All Readers," to match the UI in the single campaign edit view.
7. I think there should be an option to include draft campaigns. Maybe a dropdown that lets you select a.) only published campaigns (default), b.) only draft campaigns, or c.) both published and draft campaigns. Draft campaigns should still be shown based on the selected segment and groups.
8. Would love to see a bit of copy in the Preview tab explaining what this is and what each option does. Maybe something like: 

> Preview mode lets you view your site as a reader in a selected segment, or campaigns in selected groups. Choose a reader segment and/or one or more groups, choose whether to preview published or draft campaigns (or both), then click the "Preview" button to view your site as a reader in that segment. Only the campaigns that match the selected segment and group(s) will be shown.

Open to discussing any of these suggestions!

adekbadek:
> 1. We should consider deprecating the "Test" mode

That's the intention, eventually – but I'd prefer to tackle that in a separate PR

> 2. Once I open the iframe, I see the campaigns I expect to see, matching the Segment and Groups I specified. However, if I navigate away from the first page that opens in the iframe, it seems to lose the params and I see only Test campaigns thereafter.

Can't reproduce – the `view_as` param is appended to each anchor's `href` on each iframe load. How exactly do you navigate in this case?

> 3. I think we can do away with separate buttons to "Preview" vs. "Preview on a post." 

That's right, done in 081bb329d228e743248c0fd7e25cb95fb53c4d90

> 4. The homepage looks off to me in the iframe.

Yeah, for some reason using the `query_vars` filter breaks the homepage layout. Fixed in https://github.com/Automattic/newspack-popups/pull/349/commits/d2c5398fabd08a36b91029e996e69942f414e191

> 5a. The "None" option for Segments isn't doing what I expected […] loads "Test" campaigns

Yeah, that's not right – fixed in https://github.com/Automattic/newspack-popups/pull/349/commits/fcd6aafe08f44221318c3b21269e2433c0a3d878 & d2f6adb37f393df19e06fdcd709da63b1b6d3f23

> 5b. The "None" option for Segments isn't doing what I expected […] It should show all active campaigns that don't have an assigned segment. 

With the change of wording to "All readers", I disagree. IMO with "All readers" and no selected groups, the preview should display the site as it would look like for a regular non-logged-in user.

> 5c. It should show all active campaigns that don't have an assigned segment. Right now I don't think it's possible to preview these campaigns with the UI.

I think such a preview would be misleading – it'd be a preview of a very improbable configuration of campaigns. 

> 6. I think the "None" option for Segment should be called "All Readers," 

> 8. Would love to see a bit of copy in the Preview tab explaining what this is and what each option does.

Good ideas, done in 8611d0b8784f230ba2346d717e8dde571b017d34

> 7. I think there should be an option to include draft campaigns.

Agreed, but I'd prefer to add that in a separate PR.

thomasguillot:
For the Groups, I suggest we display them all (with a max-height and scrollable to avoid a massive list) and can be selected via a checkbox. I don't find the autocomplete extremely useful. It can be hard sometimes to remember all the Groups someone might have.

Also, I'd move the "Preview" tab before "Settings".

dkoo:
> Can't reproduce – the view_as param is appended to each anchor's href on each iframe load. How exactly do you navigate in this case?

I just clicked through article links from the homepage while the iframe was open. 

Now I'm not seeing the campaigns I expect to—I see that the iframe seems to have the correct query params in the URL, but I'm only seeing campaigns I would expect to see as a regular non-logged in user who hasn't donated or subscribed. Tested in a couple of browser to confirm, but I now see this same group of campaigns no matter which segments/groups I selected. Not seeing any console or PHP errors that would explain why.

> With the change of wording to "All readers", I disagree. IMO with "All readers" and no selected groups, the preview should display the site as it would look like for a regular non-logged-in user.

Yeah, that's what I meant: "the preview should display the site as it would look like for a regular non-logged-in user." I guess "All readers" isn't really accurate in either case, then, and it should just be called "no segment" both here and in the single edit UI?

> I think such a preview would be misleading – it'd be a preview of a very improbable configuration of campaigns.

Here again what I meant was the same thing as the above: "the preview should display the site as it would look like for a regular non-logged-in user." So that includes campaigns with no segment assigned as well as any campaigns that would match a first-time, non-logged-in reader who hasn't donated or subscribed.

dkoo:
> That's the intention, eventually – but I'd prefer to tackle that in a separate PR
> Agreed, but I'd prefer to add that in a separate PR.

Sounds good for both of these. Will defer!

dkoo:
> Now I'm not seeing the campaigns I expect to—I see that the iframe seems to have the correct query params in the URL, but I'm only seeing campaigns I would expect to see as a regular non-logged in user who hasn't donated or subscribed. Tested in a couple of browser to confirm, but I now see this same group of campaigns no matter which segments/groups I selected. Not seeing any console or PHP errors that would explain why.

This is resolved! I was behind in my Campaigns history.

> Once I open the iframe, I see the campaigns I expect to see, matching the Segment and Groups I specified. However, if I navigate away from the first page that opens in the iframe, it seems to lose the params and I see only Test campaigns thereafter.

This is also resolved with the latest Campaigns updates.

adekbadek:
> I guess "All readers" isn't really accurate in either case, then, and it should just be called "no segment" both here and in the single edit UI?

Huh, that's a very good point :D Change to "Default (no segment)" in 988378aa99213ba1a72d9dace9ea2ffff4329a9f and https://github.com/Automattic/newspack-popups/pull/349/commits/29e607d0e793000c813ebe6462f938975a3833a5

adekbadek:
@thomasguillot 

> For the Groups, I suggest we display them all (with a max-height and scrollable to avoid a massive list) and can be selected via a checkbox. I don't find the autocomplete extremely useful. It can be hard sometimes to remember all the Groups someone might have.

Interesting suggestion – added to https://github.com/Automattic/newspack-plugin/issues/754

> Also, I'd move the "Preview" tab before "Settings".

Done in 933d36cf9f62a19ed6a99fbcd2c7e2ab010b5f52

dkoo (approved):
Did another round of testing, and everything works as expected, now. The only (very minor) issue I still see is that `h3 `elements in Homepage Posts blocks look to be getting the wrong font size in the preview, but that's not a blocker:

Screen Shot 2020-12-18 at 11 24 08 AM

dkoo:
@adekbadek Sorry, I think Thomas had some requests for the UI. I'll move those to a new issue.

PR #751, opened 2020-12-17
Update is-loading animation

thomasguillot:
### All Submissions:

This PR updates the loading state animation. Moving away from the thin Material loading bar and instead using something fairly similar to Gutenberg with a "skeleton" loading.

new-anim

_Note: there's an issue with the Updates Wizard where the plugin requirement overwrites the loading... not sure what's going on cc @jeffersonrabb_

1. Switch to this branch
2. Navigate through the different Wizards (including the Setup)

### Other information:

### DIFF
--- a/assets/components/src/with-wizard/style.scss
+++ b/assets/components/src/with-wizard/style.scss
@@ -2,6 +2,7 @@
  * With Wizard
  */
 
+@import '~@wordpress/base-styles/colors';
 @import '../../../shared/scss/colors';
 
 // Reset Padding of the Admin Page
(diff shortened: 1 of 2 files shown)


jeffersonrabb:
> Note: there's an issue with the Updates Wizard where the plugin requirement overwrites the loading... not sure what's going on

The plugin loader isn't needed in the Updates wizard because there are no required plugins. I removed it in https://github.com/Automattic/newspack-plugin/pull/751/commits/e61fc8c71f758bcd7a01f84830c157271dc6779b. Each `ReleaseNotes` component loads separately, so there can't really be a global loading animation for this one.

jeffersonrabb (approved):

jeffersonrabb:
My only question about the new loading style is how it works on screens where API-based updates trigger when form elements such as toggles are changed. Example:

https://cloudup.com/cBQs4umsKVH

Does this feel like a problematic pattern?

thomasguillot:
> Does this feel like a problematic pattern?

@jeffersonrabb it does look a little bit odd because we don't have any abstract toggles but frankly it's better than the broken loading we currently have. If you look at the Site Design you could also argue that it also looks odd. I will remove 1 line from the skeleton so it doesn't look as text heavy as it is right now.

thomasguillot:
> I will remove 1 line from the skeleton so it doesn't look as text heavy as it is right now.

Done in https://github.com/Automattic/newspack-plugin/pull/751/commits/e02c43c7a694ed7e94ebea498a73ca00695c2745

Happy to merge this.

PR #764, opened 2020-12-30
refactor: campaign view and components rewritten as stateless components

jeffersonrabb:
### All Submissions:

Converts `PopupGroup`, `PopupPopover`, and `PopupActionCard` to stateless components. This branch is a pure refactor and there should be no functional changes whatsoever.

Verify the Campaigns Wizard Overlay and Inline tabs function exactly as they did before.

### Other information:

### DIFF
--- a/assets/wizards/popups/views/popup-group/index.js
+++ b/assets/wizards/popups/views/popup-group/index.js
@@ -5,7 +5,6 @@
 /**
  * WordPress dependencies.
  */
-import { Component } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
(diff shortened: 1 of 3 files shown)


assets/wizards/popups/views/popup-group/index.js
```diff
@@ -19,89 +18,75 @@ import { find } from 'lodash';
...
+	emptyMessage,
+	items: { active = [], draft = [], test = [], inactive = [] },
+	previewPopup,
+	setCategoriesForPopup,
+	setSitewideDefaultPopup,
+	publishPopup,
+	updatePopup,
+	segments,
+} ) => {
+	const getCardClassName = ( { key }, { sitewide_default } ) =>
```
dkoo, line 58: Non-blocker, just a question: any reason this can't also be declared outside the component scope alongside `descriptionForPopup` since we're passing all the data as args?

dkoo (approved):

assets/wizards/popups/views/popup-group/index.js
jeffersonrabb: Good idea. Done in https://github.com/Automattic/newspack-plugin/pull/764/commits/756c02eff3cddb85fa0cd0452309b5a4b0b1e6ab. This function is actually removed in another PR, but there's a similar one added which should also be placed out side of the component scope.

PR #770, opened 2020-12-31
feat: segmentation webview

jeffersonrabb:
### All Submissions:

This branch introduces `SegmentationPreview`, a new component in the Campaigns Wizard. The component combines logic from the `Preview` view and the `useStateWithPersistence` utility function, which will make it easier and more semantically clear to add the "view-as-segment" feature in multiple places. This also has the slight advantage of setting `localStorage` variables just before previewing, and unsetting them after the preview has been closed. 

This is a refactor, there should be no visible changes.

1. Verify the Preview feature in Campaigns Wizard works exactly as it did before. Test with no segment, segment set, segment + single campaign group, segment + multiple campaign groups.
2. Verify the `WebPreview` works normally in other contexts, e.g. the Site Design wizard.
3. Bonus: check localStorage before/during/after previewing as a segment. `newspack_campaigns-preview-segmentId` and `newspack_campaigns-preview-groupTaxIds` should be present only while the window is opening. Note that if the page is exited without closing the preview window the keys won't be purged, but this isn't much of a concern.

### Other information:

### DIFF
--- a/assets/wizards/popups/components/segmentation-preview/index.js
+++ b/assets/wizards/popups/components/segmentation-preview/index.js
@@ -1,0 +1,73 @@
+/**
+ * Segmentation Preview component.
+ * Extension of WebPreview with support for "view-as-segment" functionality.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { addQueryArgs } from '@wordpress/url';
+
+/**
+ * Internal dependencies.
+ */
+import { WebPreview } from '../../../../components/src';
+
+const SegmentationPreview = props => {
+	const postPreviewLink = window?.newspack_popups_wizard_data?.preview_post;
+	const frontendUrl = window?.newspack_popups_wizard_data?.frontend_url || '/';
+
+	const {
+		campaignGroups = [],
+		onLoad = () => {},
+		segment = '',
+		url = postPreviewLink || frontendUrl,
+	} = props;
+
+	const decorateURL = urlToDecorate => {
+		const params = {
+			view_as: [
+				`groups:${ campaignGroups.join( ',' ) }`,
+				...( segment.length ? [ `segment:${ segment }` ] : [] ),
+			].join( ';' ),
+		};
+		return addQueryArgs( urlToDecorate, params );
+	};
+
+	const onWebPreviewLoad = iframeEl => {
+		if ( iframeEl ) {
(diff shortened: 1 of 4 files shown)


assets/wizards/popups/components/segmentation-preview/index.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+import { addQueryArgs } from '@wordpress/url';
+
+/**
+ * Internal dependencies.
+ */
+import { WebPreview } from '../../../../components/src';
+
+const SegmentationPreview = props => {
+	const postPreviewLink = window?.newspack_popups_wizard_data?.preview_post;
```
dkoo, line 17: This is the first time I've seen [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in action. Always something new to learn!

assets/wizards/popups/components/segmentation-preview/index.js
```diff
@@ -0,0 +1,73 @@
...
+		const params = {
+			view_as: [
+				`groups:${ campaignGroups.join( ',' ) }`,
+				...( segment.length ? [ `segment:${ segment }` ] : [] ),
+			].join( ';' ),
+		};
+		return addQueryArgs( urlToDecorate, params );
+	};
+
+	const onWebPreviewLoad = iframeEl => {
```
dkoo, line 37: This correctly passes the params from page to page when I navigate within the preview window. I had trouble with this working before, so this is great.

dkoo (approved):
Tested, works as described, and actually improves the preview behavior in my local environment.

assets/wizards/popups/components/segmentation-preview/index.js
jeffersonrabb: I cannot claim credit :-), that's a lift from @adekbadek's [original Preview work](https://github.com/Automattic/newspack-plugin/pull/741/files#diff-b103111ab76a5a84285789bd2ae2cc0a5d4f29cdb5278802e195115d58cee273R30-R31)...

adekbadek:
@jeffersonrabb 

> This also has the slight advantage of setting localStorage variables just before previewing, and unsetting them after the preview has been closed.

The persistence was intentionally set up so that the chosen segment and group would still be there after the user has navigated away or reloaded the page. 
With the changes here, the localStorage values are never actually read, so why should they be set? If the persistence is not needed, then `beforeLoad` & `onClose` should be removed.

PR #771, opened 2020-12-31
feat: campaign group filter with preview

jeffersonrabb:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds preview functionality to the Campaigns Wizard Campaigns tab. At some point this functionality may make the Preview tab unnecessary, however the new approach doesn't support selecting multiple Campaign Groups so I'm leaving it untouched for now.

Dependent PRs:
~https://github.com/Automattic/newspack-plugin/pull/764~
~https://github.com/Automattic/newspack-plugin/pull/767~
~https://github.com/Automattic/newspack-plugin/pull/768~
~https://github.com/Automattic/newspack-plugin/pull/769~
~https://github.com/Automattic/newspack-plugin/pull/770~

Screen Shot 2021-01-01 at 9 37 19 AM

### How to test the changes in this Pull Request:

1. Be sure you have created some Campaign Groups and Segments.
2. Navigate to Campaign Wizard, first tab. 
3. Select a Campaign Group. Observe the "Segment to preview" `SelectControl` is shown. 
4. Select a Segment and click Preview, observe normal preview functionality.
5. Compare various preview scenarios in Preview and Campaigns tabs, verify the results are identical.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #769, opened 2020-12-31
feat: campaign group filter

jeffersonrabb:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds UI to filter campaigns by campaign group taxonomy.

Dependent PRs: 
~https://github.com/Automattic/newspack-plugin/pull/764~
~https://github.com/Automattic/newspack-plugin/pull/767~
~https://github.com/Automattic/newspack-plugin/pull/768~

Screen Shot 2021-01-04 at 8 30 56 AM

### How to test the changes in this Pull Request:

1. Make sure there are several Campaign Group taxonomy terms created, which can be done by editing any campaign and using the Campaign Groups sidebar panel.
2. In the Campaigns Wizard, observe `Campaign groups` `SelectControl`. Verify it contains all campaign groups. 
3. Change the selected campaign group, verify that visible campaigns are filtered correctly.
4. Change back to `All Campaigns`, verify all campaigns are shown again.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #768, opened 2020-12-31
feat: consolidate campaigns into one tab

jeffersonrabb:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Consolidates campaign management into one tab, removes `Overlay` and `Inline` tabs.

Dependent PRs:
~https://github.com/Automattic/newspack-plugin/pull/764~
~https://github.com/Automattic/newspack-plugin/pull/767~

Screen Shot 2020-12-31 at 5 38 46 AM

### How to test the changes in this Pull Request:

1. Observe there is only one campaign tab, verify it contains all campaigns and that filtering and campaign management works as expected.
2. Observe the placement is in the description. Options are `Inline|Overlay|Above header`

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #767, opened 2020-12-30
feat: campaign groups in popup action card

jeffersonrabb:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds UI to the Campaigns Wizard to control the Campaign Groups assigned to a Campaign. 

Dependent PRs:
~https://github.com/Automattic/newspack-plugin/pull/764~

This branch must be tested along with https://github.com/Automattic/newspack-popups/pull/363

Screen Shot 2020-12-30 at 6 32 09 PM

### How to test the changes in this Pull Request:

Switch `newspack-popups` to `feat/campaign-groups-in-popup-action-card`.

1. Make sure there are some Campaign Groups defined. This can be done by editing any campaign and adding campaign groups in the pane near the bottom of the sidebar. Note: at a later point we'll add UI to create Campaign Groups directly in the Campaigns wizard but this would be premature.
2. In the Campaigns wizard, click the hamburger icon on any Campaign. Observe a `Campaign Groups` autocomplete field.
3. Begin typing one of the campaign group names, observe autocomplete finds it. Click the name and observe it is added.
4. Refresh the page and observe the assigned campaign groups remain. 
5. Try removing a campaign group and observe this works as expected.
6. For Sitewide default Overlay Campaigns, observe that the hamburger tooltip reads `Campaign groups` instead of `Category filtering and campaign groups`. Click the hamburger and observe that only the Campaign Groups UI is visible.
7. Verify that the `Category filtering` fields work as they always have.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #775, opened 2021-01-02
Feat/current campaign group

jeffersonrabb:
### All Submissions:

Together with https://github.com/Automattic/newspack-popups/pull/364 adds the concept of the active campaign group. Only campaigns in the active group will be shown to end users. If there is no active group, no campaigns will be shown. 

**Note that this is a "breaking" change since it means legacy campaigns will stop working as soon as these changes take effect. I'm working on  few possible migration strategies to convert legacy campaign configurations to the new way, including this [CLI-based approach](https://github.com/Automattic/newspack-popups/tree/feat/cli-converter). We'll need to be careful not to deploy these branches until a migration solution is settled on.**

Screen Shot 2021-01-03 at 3 58 22 PM

Dependent PRs:
~https://github.com/Automattic/newspack-plugin/pull/771~
~https://github.com/Automattic/newspack-plugin/pull/769~
~https://github.com/Automattic/newspack-plugin/pull/768~
~https://github.com/Automattic/newspack-plugin/pull/767~

Switch `newspack-popups` to `feat/current-campaign-group`.

1. Make sure some Campaign Groups have been created. For now, this can only be done in the sidebar of the edit screen for a campaign. 
2. Navigate to Newspack->Campaigns->Campaigns. In the Campaign Groups `SelectControl` select a group. Observe an "Activate" button appears next to it.
3. Click to activate. Observe the label in the `SelectControl` now has "(Active)" after it. Observe there is a "Deactivate" button next to it.
4. Refresh the page, observe the Campaign Groups UI defaults to the active group. Observe this is also true when switching from another tab back to Campaigns.
5. Switch back to "All Campaigns." Assign various Campaign Groups to different campaigns. Note: the work adding Campaign Group UI to the popover is in https://github.com/Automattic/newspack-plugin/pull/767, which is merged into this branch.
6. Experiment with activating various Campaign Groups (including none) and verify in an incognito window that end users only see campaigns in the active group.
7. Verify that any individual campaign can be previewed, whether it is in the active group or not. Verify from both the Campaigns Wizard (click three dots on right->preview) and from within the editor sidebar.
8. Verify that other groups can be previewed successfully, whether active or not. Try this from both the Campaigns and Preview tabs.

### Other information:


assets/wizards/popups/views/popup-group/index.js
```diff
@@ -59,22 +61,36 @@ const descriptionForPopup = (
...
 
 	useEffect( () => {
 		apiFetch( {
 			path: '/wp/v2/newspack_popups_taxonomy?_fields=id,name',
 		} ).then( terms => setCampaignGroups( terms ) );
 	}, [] );
 
+	useEffect( () => {
+		if ( -1 === campaignGroup && activeCampaignGroup > 0 ) {
+			setCampaignGroup( activeCampaignGroup );
```
dkoo, line 90: I think we need to cast this term ID as a number here to ensure the `filteredByGroup` function always works:

```suggestion
			setCampaignGroup( parseInt( activeCampaignGroup ) );
```

I sometimes run into an issue where if I refresh the page or navigate to the Campaigns tab after setting an active group, the initial page load would show an empty list even if there were campaigns in the active group.

dkoo (changes requested):
Ran into one bug when testing this step:

> 4. Refresh the page, observe the Campaign Groups UI defaults to the active group. Observe this is also true when switching from another tab back to Campaigns.

Details are in the comment below.

At a higher level, I wanted to ask what the thinking is behind only allowing a single active group? Consider the following scenario:

* I create two groups to organize my various campaigns: 1.) Newsletter, and 2.) Donation.
* It's pledge week, and I now want campaigns from both groups to be displayed.

Is the only way to accomplish the above to create a _third_ group called "Newsletter AND Donation" and apply it to all campaigns from both groups, then make that the active group? This seems counterintuitive to me and might force editors to create a lot of redundant groups, depending on how they organize their campaigns.

Suggestion: what if each group could be toggled active or inactive, with inactive being the default? Then all groups that are marked active can be shown on the front-end? 

I think this feature warrants a bit of extra discussion because of the fact that it's a breaking change and might force publishers to reconsider how they're organizing and using their campaigns.

jeffersonrabb:
> Is the only way to accomplish the above to create a _third_ group called "Newsletter AND Donation" and apply it to all campaigns from both groups, then make that the active group? This seems counterintuitive to me and might force editors to create a lot of redundant groups, depending on how they organize their campaigns.

Campaigns can be added to multiple groups, which would be a much easier way to share a campaign across groups without creating new groups. 

> At a higher level, I wanted to ask what the thinking is behind only allowing a single active group? Consider the following scenario:
> 
> * I create two groups to organize my various campaigns: 1.) Newsletter, and 2.) Donation.
> * It's pledge week, and I now want campaigns from both groups to be displayed.
> 
> Is the only way to accomplish the above to create a _third_ group called "Newsletter AND Donation" and apply it to all campaigns from both groups, then make that the active group? This seems counterintuitive to me and might force editors to create a lot of redundant groups, depending on how they organize their campaigns.
> 
> Suggestion: what if each group could be toggled active or inactive, with inactive being the default? Then all groups that are marked active can be shown on the front-end?
> 
> I think this feature warrants a bit of extra discussion because of the fact that it's a breaking change and might force publishers to reconsider how they're organizing and using their campaigns.

I think you may be conflating expected use cases for _segments_ with _campaign groups_. "Newsletter" and "Donation" are more likely to be _segments_ than _campaign groups_. A campaign group is meant to represent the complete set of segments meant to be shown on a site at a given time, and is likely to contain a reasonably high number of campaigns (10?) each using a variety of segments. You might expect campaign groups to have names like "Production,"  "Experimental," maybe "Inactive." Overall I expect most sites will have very few campaign groups, perhaps only "active" and "inactive." Taking your example, you could have a Campaign Group called "Pledge Week," which is activated only for that specific period of time. Also, Campaigns can be members of multiple Campaign Groups, so rather than creating a Group called "Newsletter and Donation," you would just add the Campaigns to both groups, no need to create a hybrid group. Does that make sense?

assets/wizards/popups/views/popup-group/index.js
jeffersonrabb: I wasn't able to reproduce this on my local instance, but I _think_ I saw it on a production site I deployed this branch to. Here's the fix: https://github.com/Automattic/newspack-plugin/pull/775/commits/7b91839c9d8397081e1441c1996107bba6d22bf0

dkoo: Thanks!

assets/wizards/popups/views/popup-group/index.js
```diff
@@ -104,29 +126,98 @@ const PopupGroup = ( {
...
+						labelPosition="side"
+						disabled={ -1 === campaignGroups }
+					/>
+					{ campaignGroup > 0 && (
+						<SegmentationPreview
+							campaignGroups={ campaignGroup }
+							segment={ segmentId }
+							renderButton={ ( { showPreview } ) => (
+								<div className="newspack-campaigns__popup-group__filter-group-segmentation">
+									{ activeCampaignGroup !== campaignGroup && (
```
dkoo, line 159:
```suggestion
									{ +activeCampaignGroup !== campaignGroup && (
```

assets/wizards/popups/views/popup-group/index.js
```diff
@@ -104,29 +126,98 @@ const PopupGroup = ( {
...
+										<Button
+											isTertiary
+											isSmall
+											onClick={ () => manageCampaignGroup( campaignGroup ) }
+										>
+											{ __( 'Activate', 'newspack' ) }
+										</Button>
+									) }
+									{ campaignGroupExists && activeCampaignGroup === campaignGroup && (
```
dkoo, line 168:
```suggestion
									{ campaignGroupExists && +activeCampaignGroup === campaignGroup && (
```

dkoo (changes requested):
Thanks for explaining your vision of how groups should be used—definitely helps clarify things for me.

I see a couple of other places where the `activeCampaignGroup` var needs to be recast as well. Without these, in my site the "Activate" button never changes to "Deactivate".

assets/wizards/popups/views/popup-group/index.js
jeffersonrabb: I accidentally put this in a merge commit, but it's done in https://github.com/Automattic/newspack-plugin/pull/775/commits/99dd0a6a789c879b62a0e5a33643d1cda05aa0bb

assets/wizards/popups/views/popup-group/index.js
jeffersonrabb: Done in https://github.com/Automattic/newspack-plugin/pull/775/commits/99dd0a6a789c879b62a0e5a33643d1cda05aa0bb

jeffersonrabb:
On hold pending completion of https://github.com/Automattic/newspack-plugin/pull/799

dkoo:
Closing in favor of #804.

PR #777, opened 2021-01-04
Try/migrate campaigns

jeffersonrabb:
### All Submissions:

Adds UI to migrate v1 campaigns to v2. "V2" here means:

- Use campaign groups
- Do not use Test Mode
- Do not use Never frequency.

In the Campaigns Wizard, migrate UI will be shown only if the site has no Campaign Groups defined or no active group. If the user clicks "Upgrade Campaigns" the following will happen:

1. Two Campaign Groups are created, named "Active" and "Inactive"
2. Campaigns in Test Mode are assigned to the Inactive group.
3. All other Campaigns are assigned to Active group.
4. Campaigns with a frequency of "Never" are unpublished (returned to draft status). 
5. Inline and Above Header campaigns that had been in Test mode or had a Never frequency are given a frequency of "Every Page View."
6. Overlay campaigns that had been in Test mode or had a Never frequency are given a frequency of "Once A Day."
7. Activate the "Active" group.

Steps 5 & 6 may seem counter-intuitive, but since these campaigns have already been unpublished or assigned to the inactive group, they won't be displayed to end users without a decisive action by the admin.

Test with https://github.com/Automattic/newspack-popups/pull/366

1. Switch `newspack-popups` to `try/migrate-campaigns`
2. On `master` create a few campaigns, including some unpublished, some published with status "Never" and some in "Test Mode."
3. If you have Campaign Groups on your instance (and you probably do), you'll need to get rid of them:
4. SSH in to the instance. Run `wp term list newspack_popups_taxonomy --fields=term_id`. For each `term_id` run `wp term delete newspack_popups_taxonomy {term_id}`. Run `wp option delete newspack_popups_active_campaign_group`
5. Navigate to Campaigns Wizard, where you should see a `Notice` and upgrade button.
6. Click the button. Observe the operations listed above execute correctly.

### Other information:


dkoo:
Closing this since it's no longer relevant after #809.

PR #798, opened 2021-01-08
feat: quiet loading class

jeffersonrabb:
### All Submissions:

This branch introduces a "quiet" loading style. This style is applied when changing categories or any settings in the Popover. It also introduce a new animation, specific to the quiet loading style.

1. Go to Campaigns
2. Change the frequency or segment in the popover menu

### Other information:

### DIFF
--- a/assets/components/src/with-wizard/index.js
+++ b/assets/components/src/with-wizard/index.js
@@ -25,6 +25,7 @@ export default function withWizard( WrappedComponent, requiredPlugins ) {
 				complete: null,
 				error: null,
 				loading: requiredPlugins && requiredPlugins.length > 0 ? 1 : 0,
+				quietLoading: false,
 			};
 			this.wrappedComponentRef = createRef();
 		}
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
@thomasguillot this branch introduces a "quiet" loading style. This style is applied when changing categories or any settings in the `Popover`. The class is `newspack-wizard__is-loading-quiet`.

thomasguillot:
@jeffersonrabb in https://github.com/Automattic/newspack-plugin/pull/798/commits/4f3d36a90dd17d31fda72259bc9a6c3402888a7f I added a css animation specific to the quiet loading

assets/components/src/with-wizard/index.js
```diff
@@ -145,34 +146,47 @@ export default function withWizard( WrappedComponent, requiredPlugins ) {
...
 		 * Begin loading.
 		 */
-		startLoading = () => {
-			this.setState( state => ( {
-				loading: state.loading + 1,
-			} ) );
+		startLoading = quiet => {
+			if ( quiet ) {
+				this.setState( state => ( {
+					quietLoading: state.quietLoading + 1,
```
dkoo, line 152: I know this is true in `master` too, but what's the reasoning behind the type coercion and `0`/`1` instead of just using `true` and `false` for these state values? Seems to me like just sticking with booleans would be a bit more straightforward and not functionally different.

dkoo (approved):
I like it—thought it was a bit disruptive to show the whole content placeholder loading animation when changing something like segment and categories.

The below comment is a question out of curiosity, not a blocker.

assets/components/src/with-wizard/index.js
jeffersonrabb: I believe the idea is that there can be multiple concurrent API requests in flight simultaneously. By incrementing and decrementing a counter, the loading state will only be removed once all API requests finish up.

PR #776, opened 2021-01-03
Feat/deprecate test mode

jeffersonrabb:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Along with https://github.com/Automattic/newspack-popups/pull/365 this simplifies campaign settings in a few ways:

- Removes Test Mode. Recently added Preview and "preview as segment" functionality provides a richer approach and should replace Test Mode.
- Removes the "never" frequency. This setting is redundant as there are now better ways to keep campaign from being seen, including unpublishing and assigning to an inactive campaign group (or no campaign group).
- Adds "Unpublish" UI to the campaigns wizard, as a simpler alternative to the "never" frequency.

Dependent PRs:
https://github.com/Automattic/newspack-plugin/pull/775
https://github.com/Automattic/newspack-plugin/pull/771
https://github.com/Automattic/newspack-plugin/pull/769
https://github.com/Automattic/newspack-plugin/pull/768
https://github.com/Automattic/newspack-plugin/pull/767

**Note: this is a breaking change because campaigns with frequencies of "never" or test mode will no longer work. This shouldn't be deployed until a migration strategy has been implemented.**

### How to test the changes in this Pull Request:

Switch `newspack-popups` to `feat/deprecate-test-mode`. 

1. Observe Test mode UI and "Never" frequency have been removed from the editor sidebar and the Newspack Campaigns wizard pop-overs.
2. Observe a new "Unpublish" options in the Campaigns Wizard pop-over, for campaigns that are published.
3. Verify clicking Unpublish moves the campaign to the "Drafts" grouping, and causes the campaign to not appear in group previews or the front end.
4. Verify individual campaigns can still be previewed, whether published or not.
5. Verify group previews continue to work as expected.
6. Verify front end display of campaigns continues to work as expected.
7. Switch to `master` braches (in both repositories) and set several campaigns to use Test mode and/or Never frequency. 
8. Switch back to these branches, verify these campaigns do not display in previews or the front-end.

### Other information:

[...]

PR #809, opened 2021-01-12
feat: deprecate test mode and never frequency

jeffersonrabb:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/390 this simplifies campaign settings in a few ways:

- Removes Test Mode. Recently added Preview and "preview as segment" functionality provides a richer approach and should replace Test Mode.
- Removes the "never" frequency. This setting is redundant as unpublishing is now the preferred approach.

**~Note: this is a breaking change because campaigns with frequencies of "never" or test mode will no longer work. This shouldn't be deployed until a migration strategy has been implemented.~**

**Revised Note: This branch (more specifically the `newspack-popups` one) now contains migration logic.**

The migration approach examines each popup immediately after retrieval from the database. If the frequency is `test` or `never`, the campaign is unpublished. The frequency is set to `daily` for overlays and `always` for inline/above header.
 Theoretically this operation will occur at most once per site. The logic doesn't incur any processing cost, though, since the frequencies that trigger migration are wiped out forever during migration.

Replaces https://github.com/Automattic/newspack-plugin/pull/776, which had grown too out of date with a variety of recent changes. Also replaces https://github.com/Automattic/newspack-popups/pull/366.

Switch `newspack-popups` to `feat/deprecate-test-mode-two`. 

1. Observe Test mode UI and "Never" frequency have been removed from the editor sidebar and the Newspack Campaigns wizard pop-overs.
2. Verify front end display of campaigns continues to work as expected.

Verify campaign migration:

1. Switch to `master` branches (in both repositories).
2. Create an Overlay campaign in Test mode, published.
3. Create an Inline campaign with frequency Never, published.
4. Create an Inline campaign in Test mode, published.
4. Create an Above Header campaign in Test mode, published.
5. Create at least one inline campaign set to display normally, published.
6. Create one campaign, leave it unpublished.
7. View the site in a logged-in window. You should see the normal inline campaign and all the test mode ones.
8. View the site incognito - you should see only the normal inline campaign.
9. Switch to `feat/deprecate-test-mode-two` (`newspack-plugin` and `newspack-popups`)
10. View the site in a logged-in window. You should see no campaigns.
11. View the site in an incognito window. You should see only the normal inline campaign.
12. Navigate to the Campaigns Wizard. Observe all but the normal/inline one have been unpublished. 
13. Observe Above Header and Inline ones that were unpublished now have frequency Every Page.
14. Observe Overlay now has frequency Once A Day.

### Other information:


adekbadek (approved):
I took the liberty of fixing `sortPopups` method in 42089d266dd2f32be7ba189e4e436f3f418e8c44 (some popups were duplicated on the frontend), hope you don't mind 😬

dkoo (changes requested):
> Switch back to these branches, verify these campaigns do not display in previews or the front-end.

This is not what I saw—when I create a Test or "never" campaign on `master` then switch to this branch and https://github.com/Automattic/newspack-popups/pull/390, those campaigns are treated the same as published campaigns and are displayed on both the front-end and in previews as long as they match the reader segment.

jeffersonrabb:
Added fix for fatals if there are no active campaigns in https://github.com/Automattic/newspack-plugin/pull/809/commits/7075423fdb5c009e74a578377eb50ceaf6ac5088

adekbadek (approved):

dkoo (approved):
With the migration logic in https://github.com/Automattic/newspack-popups/pull/390, this is now working for me.

PR #810, opened 2021-01-12
Feat/nrh xdomain tracking

jeffersonrabb:
### All Submissions:

Adds [Google Analytics Cross Domain tracking](https://developers.google.com/gtagjs/devguide/linker.) for sites using News Revenue Hub checkout. For the feature to work, the Donate block needs to submit directly to the NRH checkout site — the `decorate_forms` feature won't work with the redirect approach currently on `master`. Because of this problem, the biggest part of this branch is a rewrite of the Donate block rendering function for sites with News Revenue Hub checkout. This also gets a bit complicated because of how the form inputs work. In the new implementation there's a hidden `amount` field which should contain the current amount accurately following a variety of form interactions, including:

- Initial values that reflect the default tab
- Clicking on the three price buttons
- Switching tabs
- Manually setting the amount after clicking the Other button

I noticed that the `https://checkout.fundjournalism.org/donateform` endpoint now seems to always redirect to `https://checkout.fundjournalism.org/memberform`, so I skipped implementing the logic to switch between the two endpoints (original implementation [here](https://github.com/Automattic/newspack-plugin/blob/master/includes/class-nrh.php#L60-L72)).

Test with https://github.com/Automattic/newspack-blocks/pull/671.

1. In the Reader Revenue wizard switch platform to News Revenue Hub. Use an Organization ID and Salesforce Campaign ID from a production site.
2. Google Analytics will need to be set up for cross-domain tracking to work.
3. Switch Newspack Blocks to `feat/attributes-in-donate-block-hook`
4. Create a few pages with the Donate block on them. Some should be AMP, some non-AMP. Some should have multiple Donate blocks on the same page. Use tiered and un-tiered, and experiment with different amounts and default tabs.
5. The easiest way to test that the hidden `amount` field is working is to change the type to `text` [here](https://github.com/Automattic/newspack-plugin/pull/810/files#diff-ae6055b360153ad3a025eaee30a5deade1801c95c3b03604c638a8372634118aR219) and [here](https://github.com/Automattic/newspack-plugin/pull/810/files#diff-ae6055b360153ad3a025eaee30a5deade1801c95c3b03604c638a8372634118aR318). This allows you to observe the value, and you can interact with the block in a variety of ways and verify the value is correct. Test in AMP and non-AMP, because the implementations are quite different.
6. Verify that the correct price and term is set when clicking through to the hub checkout page.
7. Verify the cross-domain tracking is working. There should be a `_ga` parameter appended to the URL that looks something like this: `_ga=2.257424207.531864944.1610926274-amp-V_WelaH4thIfJKrlpv_1HQ`
8. Bonus: add a button block to a page and link directly to an NRH checkout page. Verify the `_ga` parameter is appended to this URL too when clicked.
9. Verify none of this affects the behavior of the block on non-NRH sites.
10. Probably wise to verify the block works properly on sites that are still using the NRH checkout plugin and don't have NRH selected in the Reader Revenue wizard. Even though this is a configuration we should deprecate, we don't want to break donation on live sites.

### Other information:


jeffersonrabb:
@brianboyer could you take a look at this?

brianboyer:
Howdy. Didn't run through the full checklist (checking other sites, etc.), but I can verify that the linker appears to be working as designed. Notes follow.

**Error**: I do not see the Salesforce campaign ID that I set up in the reader revenue configuration, it should be on the URL. The organization ID does appear. (I see that there is also a campaign ID option on the configuration of the donate block. When I set that, it works, using what I entered for the donate block settings.)

**Error**: It appears that the linker parameter in the querystring is duplicated when you use the form on AMP pages. I see two "_gl" parameters. Links on AMP pages don't seem to have the same issue, and I don't see duplicates on regular HTML pages. 

**UX notes...**
I don't know how prevalent this is, but in my experience, an organization might use different Salesforce campaign IDs for different prompts on the site. For example, to understand the effectiveness of a particular popup. Personally, I'd prefer to use Google Analytics to determine the performance of different components on the site, but many people aren't familiar with event tracking and whatnot.

Related: I presume that the reader revenue settings here will only apply to the donate blocks? (And not, let's say, a donate button in the header, for which you'd just copy-paste the link.) If so, it might be better to make the settings configurable on the donate block itself. That way, everything is more parallel. You'd set the Org ID and the SF ID when you set up the prompt, similar to setting the URL for various buttons and other campaign things. (This would also open the door to the "different prompts get different SF IDs" use case above.) Or perhaps you do set Org ID and Campaign ID as a reader revenue setting, but it's marked as a *default* campaign id, that you can override at the donation block level? Hrm.

FWIW, it the form sends a bunch of unnecessary field values. I don't believe the "donation_value_..." fields are necessary.
https://checkout.fundjournalism.org/memberform
?org_id=montanafreepress&amount=15
&**donation_value_once**=180
&**donation_value_once_other**=180
&installmentPeriod=monthly
&**donation_value_month**=15
&**donation_value_month_other**=15
&**donation_value_year**=180
&**donation_value_year_other**=180
&_gl=1*1aj6kfw*_ga*YW1wLW9iSExQNzViUGh2U2xCOHlYUXM5ZlE.&_gl=1*1aj6kfw*_ga*YW1wLW9iSExQNzViUGh2U2xCOHlYUXM5ZlE.

jeffersonrabb:
> Error: It appears that the linker parameter in the querystring is duplicated when you use the form on AMP pages. I see two "_gl" parameters. Links on AMP pages don't seem to have the same issue, and I don't see duplicates on regular HTML pages.

@brianboyer could you clarify this? You say "AMP pages" in both cases, I assume one is meant to be "non-AMP pages"?

brianboyer:
> @brianboyer could you clarify this? You say "AMP pages" in both cases, I assume one is meant to be "non-AMP pages"?

Sorry for being unclear. I'm testing two things: forms and links (e.g. a donate button).

- Both forms and links are okay on HTML pages.
- Links on AMP pages are okay.
- Forms on AMP pages have the duplication issue.

includes/class-nrh.php
```diff
@@ -17,99 +17,451 @@ class NRH {
...
+				],
+				'input' => [
+					'type'  => [],
+					'name'  => [],
+					'value' => [],
+				],
+			];
+
+			$is_amp        = function_exists( 'is_amp_endpoint' ) && is_amp_endpoint();
+			$block_id      = 'wpbdbd_' . uniqid();
```
claudiulodro, line 152: Not a blocker or anything, just curious what does `wpbdbd` stand for? WordPress Block Donate Block Donate? :)

includes/class-nrh.php
```diff
@@ -17,99 +17,451 @@ class NRH {
...
+			$form_open = sprintf(
+				'<form method="GET" action="https://checkout.fundjournalism.org/memberform"><input type="hidden" name="org_id" value="%s" />',
+				esc_attr( $organization_id )
+			);
+
+			// phpcs:disable
+
+			ob_start();
+
+		if ( ! $is_amp ) : ?>
```
claudiulodro, line 170: I only noticed when I reached this line, but a lot of this function is indented weirdly. Above this, everything is indented too much. Below this, the matching `endif` doesn't line up with the opening `if ( ! $is_amp ) :`.

claudiulodro (approved):
It works great and the code looks good (that's some fancy work with the hidden `amount` field on AMP!). There are a number of issues with the white space in `render_nrh_donate_block`, but that's not blocking. Good work! 👍

includes/class-nrh.php
jeffersonrabb: Exactly - this isn't new in this branch, but at some point @laurelfulford and I realized we were adding a substantial amount of CSS weight by using verbose block names, which was pushing sites right up to the AMP CSS limit. Here's one PR where this change was implemented: https://github.com/Automattic/newspack-blocks/pull/184

includes/class-nrh.php
jeffersonrabb: Here's the saga. All of the weird indentation was done to comply with PHPCS rules, which get pretty odd when dealing with large blocks of markup. At some point I ended up adding PHPCS ignore rules, but left the indentation intact. In https://github.com/Automattic/newspack-plugin/pull/810/commits/1fa6e2950baaa5c2277d8ed8d295f6cd1a717c5b I was just able to resolve a couple remaining PHPCS issues and restored PHPCS. If you check it now you'll see that any change to the indentation will be flagged by PHPCS.

jeffersonrabb:
> FWIW, it the form sends a bunch of unnecessary field values. I don't believe the "donation_value_..." fields are necessary.

These fields aren't necessary for the submission but are for the behavior of the block - basically they're just benign noise in the POST data. As long as the extra fields aren't harmful, I'm inclined to leave them in place. @brianboyer is there any reason you're aware of why these fields might cause problems?

brianboyer:
I agree w/ your assessment that this behavior is probably benign. If you
want to be certain, ask Tristan at NRH.

jeffersonrabb:
> **Error**: It appears that the linker parameter in the querystring is duplicated when you use the form on AMP pages. I see two "_gl" parameters. Links on AMP pages don't seem to have the same issue, and I don't see duplicates on regular HTML pages.

After discussing with @adamsilverstein it seems this is a known Site Kit issue which has been addressed [here](https://github.com/google/site-kit-wp/issues/2668) and will be resolved in the 1.25.0 SK release. It also isn't harmful, just peculiar.

jeffersonrabb:
> **Error**: I do not see the Salesforce campaign ID that I set up in the reader revenue configuration, it should be on the URL. The organization ID does appear. (I see that there is also a campaign ID option on the configuration of the donate block. When I set that, it works, using what I entered for the donate block settings.)

Good catch. In https://github.com/Automattic/newspack-plugin/pull/810/commits/bfaab848a6b18153450705ebed3adfb2c2d35966 I've rewritten so the block's campaign is used if present. If not, it falls back to the global campaign set in the Reader Revenue wizard.

jeffersonrabb:
@claudiu could you give this one more look now that I've addressed @brianboyer's feedback about Salesforce ID?

claudiulodro (approved):
Latest version looks good 👍 The salesforce ID works well with a global ID and with a block-level ID. Good catch, @brianboyer!

PR #814, opened 2021-01-14
Update style card design

thomasguillot:
### All Submissions:

To be in line with the new onboarding design, this PR updates the style of `StyleCard` and the Theme Wizard. 

__After:__

1-after

__Before:__

2-before

1. Go to Site Design, check the old style
2. Switch to this branch and refresh wizard
3. Do you see the new style? Can you view demo/activate properly? 

### Other information:

### DIFF
--- a/assets/components/src/grid/style.scss
+++ b/assets/components/src/grid/style.scss
@@ -25,15 +25,10 @@
 
 	// Media Queries - 744px
 	@media screen and ( min-width: 744px ) {
-		// Columns 2 & 4
 		&__columns-2,
+		&__columns-3,
 		&__columns-4 {
 			grid-template-columns: repeat( 2, 1fr );
-		}
-
-		// Columns 3
-		&__columns-3 {
-			grid-template-columns: 1fr;
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


dkoo (changes requested):
@thomasguillot This is looking good! One thing I noticed which is related but outside the scope of this PR is that the demo staging sites for each theme seem to lack any content. This kind of looks like a mistake because when clicking "View Demo" you are presented with an empty shell of a page rather than a sample homepage or single page. It would be great if the homepages for the demo staging sites had some content to showcase.

I also have a few suggestions which I've pushed to this branch as 8f0c42e. These are just suggestions, so feel free to push more changes or revert this commit, and I'll approve after your response. To summarize:

* The tab indexing doesn't quite work for keyboard-only navigation in the site design wizard. Removing the `tabindex` attribute from the container `div` and adding them to each button instead allows you to navigate every "Activate" and "View Demo" button without a mouse, and using `:focus-within` along with the hover styles lets the buttons stay visible if you've tabbed over the buttons instead of hovering.
* Now that we have the "quiet" loading style implemented in #798, I think this is more appropriate for ops like updating the active theme or saving settings.

thomasguillot:
@dkoo Both are great suggestions that I didn't think about, thank you!

I removed some unused CSS, added the demo button when the theme is active (you might still want to see the demo...) and I removed the image conditional check since to me, it doesn't make any sense to use this component without an image 😅

dkoo (approved):
LGTM! 👍

PR #832, opened 2021-01-26
feat: segment priority UI and logic

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/416, introduces the concept of segment priority and the UI for managing it:

* Implements a drag-and-drop (and up/down button) UI for setting priority of segments
* If there were existing segments beforehand, priority is assigned automatically in the existing order, so that segments are presented at first in the same order as before enabling the feature.
* Removes sitewide default as an option and ignores it for purposes of inserting overlays
* Standardizes prompt insertion logic for previewing vs. front-end (no more additional logic for view-as)
* If multiple overlays or above header prompts are candidates for being inserted into a post or page, takes into account the priority of the assigned segments and shows only the one with the highest segment priority (with no segment always being last in priority)—note that "higher priority" means lower index value, with `0` being the highest possible priority
* ~Filters out the "Uncategorized" category from the wizard UI, because Popups treats it the same as no category~ This was done in #836.
* If there are multiple overlays or above-header prompts assigned to the same segment that share at least one category (or if they all lack any categories), shows a warning in each conflicting prompt's action card
* In these cases of conflict only the most recently created prompt will be shown in previews and on the front-end.

Closes https://github.com/Automattic/newspack-popups/issues/371.

#### Conflict UI and logic, no more sitewide default

1. On `master`, ensure there are multiple above-header prompts and overlay prompts assigned to the same segment, one of which is designated as the sitewide default. Try to spread them across multiple campaigns, with at least two sharing the same prompt. To start, make sure all have no category filtering applied.
2. Check out this branch and https://github.com/Automattic/newspack-popups/pull/416, and run `npm run build` in both repos.
3. Go to the Campaigns tab and confirm that the sitewide default is no longer labeled as such, and that all UI dealing with setting and unsetting the sitewide default no longer exists.
4. Confirm that the overlays assigned to the same segment and the above header prompts assigned to the same segment are shown with a warning like this (the warning will only be shown for prompts that are published):

Screen Shot 2021-02-02 at 3 54 33 PM

5. Assign a category to one of the overlays and/or Above Header prompts. Confirm that the conflict warning is only shown for prompts that share a segment and also share at least one category, or with no categories assigned. Remove the categories after confirming.
6. Preview the prompts so that all conflicting prompts are candidates for being displayed. Confirm that only one overlay and one above-header prompt is shown despite multiple prompts being candidates for display. Test previewing with various campaigns and segments.
7. Repeat step 5 on the front-end in a new Incognito session and confirm that you see the same prompts.

#### Segment priority UI and logic

1. Go to the Segments tab. Confirm that segments are now displayed with a sort UI (drag-and-drop handle, plus up/down buttons):

Screen Shot 2021-01-25 at 6 06 09 PM

2. Test reordering functionality with both drag-and-drop and the up/down buttons. After reordering segments manually and refreshing the tab (or leaving and coming back), the segments should remain in the order as set.
3. Ensure there are at least two segments with overlapping criteria. For example:
  - Segment A is New Readers with min articles 1
  - Segment B is Newsletter Subscribers with min articles 1 AND Newsletter Subscribers only
4. Set the priority so that Segment A has higher priority than Segment B.
5. In the Campaigns tab, assign Segment A to one overlay, and Segment B to another overlay. Confirm that the overlays no longer display the conflict warning. For now, ensure that both overlays have either no category or are set to "uncategorized" (which is the same thing, for our logic).
6. Preview both overlays and confirm that only the one assigned to Segment A is shown.
7. View in a new Incognito session on the front-end with `?utm_medium=email` appended to the URL, and confirm that only the overlay assigned to Segment A is shown.
8. In the Segmentation tab, reorder the priority so that Segment B is higher than Segment A.
9. Repeat steps 6-7 and confirm this time that only the overlay assigned to Segment B is shown.
10. Test that if the reader matches only the lower-priority segment, that segment's prompt is shown even though it has a lower priority. For example: visit an article in a new Incognito window, _without_ the `?utm_medium=email` param. The prompt for Segment A should be shown even though Segment B still has a higher priority, because the new session doesn't match the conditions for Segment B.
11. Assign the segments (or two other overlapping segments) to two different Above Header prompts and repeat steps 6-9 with those prompts.
12. Test the category overlay logic by assigning a category to one of the overlays. Preview and view the site on the front-end, and confirm that:
   - When viewing any post with a matching category, only the overlay with the matching category is shown
   - When viewing any post without a matching category, the overlay with the higher segment priority is shown
13. Test overlays and above-header prompts with:
   - One of each assigned to a segment, and the other assigned to Default/No Segment (confirm that if the reader matches segment criteria, prompts with an assigned segment are always shown over prompts with no segment in previews and on the front-end; and conversely, if viewing as a reader that doesn't match the segment criteria, the prompt with no segment is shown)
   - All assigned to Default/No Segment (conflicting prompts are shown with a warning, and only one of each placement type is shown in previews and on the front-end)
14. Continue testing with various segments and segmentation criteria, and a combination of overlay, above-header, and inline prompts. Confirm that the segmentation logic and inline prompt logic remain unchanged (e.g. you should still see all inline prompts regardless of segment priority, and you should only see prompts that match your reader segment when previewing or viewing as a front-end reader).
15. Test adding new segments, deleting segments, and starting with no segments.

### Other information:


dkoo:
@jeffersonrabb and @adekbadek, I'm setting this to "ready for review" so we can move forward with testing and reviewing against `master`, but I won't merge them if approved until we know what's happening with #833.

assets/wizards/popups/utils.js
```diff
@@ -8,6 +8,17 @@ export const isOverlay = popup =>
...
+ * Check whether the given popup is above-header.
+ *
+ * @param {object} popup Popup object to check.
+ * @return {boolean} True if the popup is a above-header, otherwise false.
+ */
+export const isAboveHeader = popup => 'above_header' === popup.options.placement;
+
+/**
+=======
+>>>>>>> master
```
adekbadek, line 21: Some conflict resolution comment to delete ;)

adekbadek (changes requested):

assets/wizards/popups/utils.js
dkoo: Whoops, weird how the linter didn't catch this. Fixed in eb09a1e.

assets/wizards/popups/views/popup-group/index.js
```diff
@@ -61,16 +55,61 @@ const descriptionForPopup = (
...
+				conflict.id !== campaign.id &&
+				isSameType( campaign, conflict ) &&
+				campaign.options.selected_segment_id === conflict.options.selected_segment_id &&
+				hasConflictingCategory
+			);
+		} );
+
+		if ( 0 < conflictingCampaigns.length ) {
+			warningMessages.push(
+				`${ __( 'Conflicts with:', 'newspack' ) } ${ conflictingCampaigns
```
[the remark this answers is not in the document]
dkoo: Good idea! How does this look? Updated in 6251ea5.

Screen Shot 2021-01-29 at 11 26 07 AM

adekbadek (approved):

assets/wizards/popups/components/prompt-action-card/index.js
```diff
@@ -23,7 +23,7 @@ const PromptActionCard = props => {
 	const [ categoriesVisibility, setCategoriesVisibility ] = useState( false );
 	const [ popoverVisibility, setPopoverVisibility ] = useState( false );
 
-	const { className, description, prompt = {}, segments } = props;
+	const { className, description, prompt = {}, segments, warning } = props;
 	const { id, edit_link: editLink, title, sitewide_default: sitewideDefault } = prompt;
```
adekbadek, line 27: Here and in a few other spots, sitewide default setting is still handled. Most notably, overlays are now marked as inactive in the UI:

image

assets/wizards/popups/index.js
```diff
@@ -337,18 +321,20 @@ class PopupsWizard extends Component {
...
-												campaignId={ campaignId }
-												createCampaignGroup={ createCampaignGroup }
-												deleteCampaignGroup={ deleteCampaignGroup }
-												duplicateCampaignGroup={ duplicateCampaignGroup }
-												renameCampaignGroup={ renameCampaignGroup }
-												prompts={ filterByCampaign( prompts, campaignId ) }
-												campaigns={ campaigns }
-												hasUnassigned={ filterByCampaign( prompts, 'unassigned' ).length }
-											/>
+											<CampaignsContext.Provider value={ prompts }>
```
adekbadek, line 324: Note: the use of context here creates a sort of source-of-truth duplication, since prompts (though filtered) are also passed in a prop to the `Campaigns` component. But this can be resolved in a separate PR, just flagging here.

adekbadek (changes requested):

dkoo:
@adekbadek Hate to do this to you, but could use another review of the changes since #833 got merged, since I had to fix a bunch of conflicts.

Also, @thomasguillot: I found myself having a lot of trouble discerning hierarchy and where one campaign ended and another began in the Campaigns list. Unlike other wizards, here we're now nesting Action Cards within other Action Cards, and I felt the extremely light borders and tight padding were making the components blend together visually. I made some accessibility-related tweaks in 1941dda to try to alleviate that a bit, but would you like to take a look and see if you have any better ideas?

assets/wizards/popups/index.js
dkoo: Ah, good point. Since this is the only other place `prompts` is passed directly as a prop, I just went ahead and refactored this in 98c7497. Now every child of `popups/index.js` that uses `prompts` directly uses it from the context data.

thomasguillot:
@dkoo please revert https://github.com/Automattic/newspack-plugin/commit/1941dda800b7629778e967b68657fe1fd3a2db89 and I'll think of a solution in a separate PR.

assets/wizards/popups/components/prompt-action-card/index.js
dkoo: I feel like I've fixed this 3x already but it keeps getting overwritten by other PRs! Fixed again in a9dcd85.

dkoo:
> @dkoo please revert 1941dda and I'll think of a solution in a separate PR.

Done! Thanks for taking a look.

adekbadek (approved):
😎

PR #833, opened 2021-01-26
feat: campaigns wizard overhaul

jeffersonrabb:
### All Submissions:

This PR is a substantial overhaul of the campaigns wizard. Here's an overview of what's changed:

1. Nomenclature has changed significantly. A "campaign" is now a term in the custom taxonomy (previously called "campaign groups"). A "prompt" is an instance of the custom post type (previously "campaign"). "Segments" have not changed.
2. The main Campaigns screen displays a set of prompts grouped by segment. The select at the top determines which set is shown. "Active Prompts" shows all published Prompts, regardless of taxonomy membership. Essentially this shows everything that could possibly appear to an end user. "Unassigned Prompts" shows all prompts which are not in any Campaigns, if there are any to show. All other items in the Select are Campaigns.
3. Prompts are grouped by Segment. There is a Preview link for each segment, which simplifies the UI for view-as-segment significantly. As a result of this simplification, the Preview screen has been removed.
4. Each Segment also has its own "Add New Prompt" button. When clicked, a `PopOver` shows icons for all placement options. Clicking an icon creates the Prompt and redirects to the Editor. Note: new prompts will be initially configured with the placement, segment, and campaign, however segment and campaign require https://github.com/Automattic/newspack-popups/pull/426 so it may be difficult to test. All set have "Add New Prompt" UI except for Unassigned Prompts.
5. The `ActionCard` for each Prompt shows all Campaigns it is in, as well as frequency and placement. The two `PopOvers` on the right side are largely unchanged, except for a few wording things.
6. In the upper right, the "Add New" button now creates a new campaign (previously this created a new Prompt). The UI to name the Campaign is in a `Modal.`
7. To the right of the Select are three vertical dots which open a `PopOver` with a variety of Campaign management tools. Note, these dots are only present when a Campaign is selected, not when Active or Unassigned Prompts are selected.
8. The first management feature allows bulk activation or deactivation of all the prompts in a campaign. Activating will publish all prompts in the campaign and un-publish all others (this is not a new feature in this branch). Deactivation simply un-publishes all the prompts in the campaign.
9. Duplicate opens a `Modal` to prompt for a new campaign name. The new campaign is then created, and all prompts in the original campaign are assigned to both. Note that the prompts themselves are not duplicated.
10. Rename opens a `Modal` where the current campaign can be reamed.
11. Archive uses term meta to assign a status to Campaigns. A campaign can be archived or unarchived. This only affects the display in the Select—Archived campaigns are displayed in a separate group below active ones. Note that archiving  campaign has no impact on the prompts within.
12. Delete deletes a campaign but has no impact on the prompts within it. 

### Screenshots:

Screen Shot 2021-02-01 at 1 00 38 AM

Screen Shot 2021-02-01 at 1 00 42 AM

Screen Shot 2021-02-01 at 1 00 48 AM

Testing requires `newspack-popups` be on https://github.com/Automattic/newspack-popups/pull/427. As noted above, a part of the Add New Prompt functionality requires https://github.com/Automattic/newspack-popups/pull/426 but it's not a critical feature. https://github.com/Automattic/newspack-popups/pull/425 addresses a minor edge case where a campaign assigned to a deleted segment shows up nowhere in the UI. 

So much is new in the branch that it's difficult to establish a simple testing sequence. The best approach is to functionally test the flows of setting up campaigns, prompts and managing them. 

A note about naming conventions in the code: many functions and variables have been renamed to match the new nomenclature, but there is still more to do. I limited renaming to the Newspack Plugin, leaving function names in Newspack Popups untouched, for now. After this lands we can go back and comprehensively fix all of the naming in both. 

### Other information:


jeffersonrabb:
Update: https://github.com/Automattic/newspack-popups/pull/425 and https://github.com/Automattic/newspack-popups/pull/426 have been merged so now all testing can be done with https://github.com/Automattic/newspack-popups/pull/427.

includes/wizards/class-popups-wizard.php
```diff
@@ -389,21 +470,23 @@ public function api_get_settings() {
...
 		];
 
 		if ( $newspack_popups_configuration_manager->is_configured() ) {
-			$response['popups']   = array_map(
-				function( $popup ) {
-					$popup['edit_link'] = get_edit_post_link( $popup['id'] );
-					return $popup;
+			$response['prompts']   = array_map(
+				function( $prompt ) {
+					$popup['edit_link'] = get_edit_post_link( $prompt['id'] );
```
adekbadek, line 482:
```suggestion
					$prompt['edit_link'] = get_edit_post_link( $prompt['id'] );
```
;)

assets/wizards/popups/index.js
```diff
@@ -224,14 +198,64 @@ class PopupsWizard extends Component {
...
+	};
+
+	deleteCampaignGroup = id => {
+		const { setError, wizardApiFetch } = this.props;
+		return wizardApiFetch( {
+			path: `/newspack/v1/wizard/newspack-popups-wizard/delete-campaign/${ id }`,
+			method: 'DELETE',
+			quiet: true,
+		} )
+			.then( this.updateAfterAPI )
```
adekbadek, line 258: After deletion, `history` should be updated - otherwise the UI is still displaying the deleted campaign's prompts, even though the top select is set to "Active prompts"

assets/wizards/popups/index.js
```diff
@@ -224,14 +198,64 @@ class PopupsWizard extends Component {
...
+
+	duplicateCampaignGroup = ( id, name ) => {
+		const { setError, wizardApiFetch } = this.props;
+		return wizardApiFetch( {
+			path: `/newspack/v1/wizard/newspack-popups-wizard/duplicate-campaign/${ id }`,
+			method: 'POST',
+			data: { name },
+			quiet: true,
+		} )
+			.then( this.updateAfterAPI )
```
adekbadek, line 247: Suggestion: update `history` so that after duplication the UI displays the new campaign

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+								key: 'header-campaigns',
+								name: __( 'Campaigns', 'newspack' ),
+								className: 'is-header',
+							},
+							...activeCampaigns.map( ( { term_id: id, name } ) => ( {
+								key: id,
+								name,
+								className: 'newspack-campaigns__campaign-group__select-control-group-item',
+							} ) ),
+							archivedCampaigns.length && {
```
adekbadek, line 178: If there are no archived campaigns, `0` is passed among options to the `CustomSelectControl`, which is not a valid option. Here's a trick for inline conditional array elements:

```js
...( archivedCampaigns.length
	? [
			{
				key: 'header-archived-campaigns',
				name: __( 'Archived Campaigns', 'newspack' ),
				className: 'is-header',
			},
	  ]
	: [] ),
```

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+					<CustomSelectControl
+						label={ __( 'Campaigns', 'newspack' ) }
+						options={ [
+							{ key: 'active', name: __( 'Active Prompts', 'newspack' ) },
+							{ key: 'unassigned', name: __( 'Unassigned Prompts', 'newspack' ) },
+							{
+								key: 'header-campaigns',
+								name: __( 'Campaigns', 'newspack' ),
+								className: 'is-header',
+							},
```
adekbadek, line 172: When there are no Campaigns, the UI looks like it's missing something – perhaps a "No Campaigns found" message could be displayed in the dropdown?

image

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+					<Button
+						isPrimary
+						isSmall
+						onClick={ () => {
+							setModalVisible( ! modalVisible );
+							setCampaignName( '' );
+							setModalType( MODAL_TYPE_NEW );
+						} }
+					>
+						{ __( 'Add New Campaign', 'newspack' ) }
```
adekbadek, line 243: Suggestion: some space could be saved by moving this button to segment header:

image

adekbadek (changes requested):

assets/components/src/plugin-installer/style.scss
```diff
@@ -7,7 +7,7 @@
 .newspack-card {
 	.newspack-plugin-installer__status {
 		align-items: center;
-		border-left-width: 4px;
+		box-shadow: inset 4px 0 0 $light-gray-700;
```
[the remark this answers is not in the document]
thomasguillot: Fixed: https://github.com/Automattic/newspack-plugin/pull/833/commits/780ef7e60c62f49cf4cbc8d28928489b55017310

assets/wizards/popups/utils.js
```diff
@@ -10,9 +15,86 @@ export const isOverlay = popup =>
...
+		bottom: __( 'Bottom Overlay', 'newspack' ),
+		inline: __( 'Inline', 'newspack' ),
+		above_header: __( 'Above Header', 'newspack' ),
+	}[ placement ];
+};
+
+const frequencyMap = {
+	once: __( 'Once', 'newspack' ),
+	daily: __( 'Once a day', 'newspack' ),
+	always: __( 'Every page', 'newspack' ),
```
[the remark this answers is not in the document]
jeffersonrabb: I like it too. Done in https://github.com/Automattic/newspack-plugin/pull/833/commits/59f98ce2fd5841f0cff03ec3d19270f2ed00c0e2. We'll need to make a similar change to the UI in `newspack-popups` here https://github.com/Automattic/newspack-popups/blob/cc04314c8986734e658613ef1f5120be6bfa2012/src/editor/FrequencySidebar.js#L24

assets/wizards/popups/components/prompt-action-card/index.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * Popup Action Card
```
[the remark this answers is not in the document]
jeffersonrabb: https://github.com/Automattic/newspack-plugin/pull/833/commits/42abf8565ab881495ff310c47b9970372a975334

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+		return {
+			key: 'active',
+			name: __( 'Active Prompts', 'newspack' ),
+		};
+	};
+	return (
+		<Fragment>
+			<div className="newspack-campaigns__campaign-group__filter-group-wrapper">
+				<div className="newspack-campaigns__campaign-group__filter-group-actions">
+					<CustomSelectControl
```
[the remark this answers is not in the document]
jeffersonrabb: In https://github.com/Automattic/newspack-plugin/pull/833/commits/16969a2e0a5e71b22907263b0680eb2f914870d2 `is-selected` class will be applied to the current selection. To verify this is functioning properly, I applied underlined text to this style, but @thomasguillot can remove this and figure out how the selected style should actually look.

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+	return __( 'New Campaign', 'newspack' );
+};
+
+const modalButton = modalType => {
+	if ( MODAL_TYPE_RENAME === modalType ) {
+		return __( 'Rename', 'newspack' );
+	} else if ( MODAL_TYPE_DUPLICATE === modalType ) {
+		return __( 'Duplicate', 'newspack' );
+	}
+	return __( 'Add', 'newspack' );
```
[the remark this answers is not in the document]
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-plugin/pull/833/commits/8b6576f608579305753fb3739d05464ed8c14f49

assets/wizards/popups/index.js
```diff
@@ -274,21 +298,30 @@ class PopupsWizard extends Component {
...
+									path="/campaigns/:id?"
+									render={ props => {
+										const campaignId = props.match.params.id;
+										return (
+											<Campaigns
+												{ ...popupManagementSharedProps }
+												campaignId={ campaignId }
+												prompts={ filterByCampaign( prompts, campaignId ) }
+												campaigns={ campaigns }
+												emptyMessage={ __( 'No Campaigns have been created yet.', 'newspack' ) }
```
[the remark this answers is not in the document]
jeffersonrabb: Removed in https://github.com/Automattic/newspack-plugin/pull/833/commits/46dc20974efd04e9a1eefa2dbb2627a876498dd5

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: Resolved in https://github.com/Automattic/newspack-plugin/pull/833/commits/c6c9d25d743703737e55cf59c069897432351a13 by hiding the `Campaigns` header row if none exist.

includes/wizards/class-popups-wizard.php
jeffersonrabb: Whoops. https://github.com/Automattic/newspack-plugin/pull/833/commits/0a18edc364e765623e6f37dcf4e303946965a133

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+					campaignId={ campaignId }
+					segments={ segments }
+					{ ...props }
+				/>
+			) ) }
+			{ prompts.length < 1 && -1 === campaignId && (
+				<p>{ __( 'No Campaigns have been created yet.', 'newspack' ) }</p>
+			) }
+			{ prompts.length < 1 && campaignId > 0 && (
+				<p>{ __( 'There are no Campaigns in this group.', 'newspack' ) }</p>
```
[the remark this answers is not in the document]
jeffersonrabb: Thinking a bit more on this, we don't really need the empty set UI. Even if there are no prompts, campaigns, or segments there will always be an Active Prompts screen with an empty area for the Default Segment. I've removed in https://github.com/Automattic/newspack-plugin/pull/833/commits/7cb557adf35e356d9a99776d53802adffc6f73cb

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,301 @@
...
+			{ groupBySegment( segments, prompts ).map( ( segment, index ) => (
+				<SegmentGroup
+					key={ index }
+					segment={ segment }
+					campaignId={ campaignId }
+					segments={ segments }
+					{ ...props }
+				/>
+			) ) }
+			{ prompts.length < 1 && -1 === campaignId && (
```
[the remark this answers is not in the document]
jeffersonrabb: This was removed in https://github.com/Automattic/newspack-plugin/pull/833/commits/7cb557adf35e356d9a99776d53802adffc6f73cb.

assets/wizards/popups/components/prompt-popovers/primary.js
```diff
@@ -83,16 +83,16 @@ const PrimaryPopupPopover = ( {
 					} }
 					className="newspack-button"
 				>
-					{ __( 'Unpublish', 'newspack' ) }
+					{ __( 'Deactivate', 'newspack' ) }
 				</MenuItem>
 			) }
```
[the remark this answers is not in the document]
jeffersonrabb: Good catch, this is a regression actually. Fixed in https://github.com/Automattic/newspack-plugin/pull/833/commits/c8215a900ca6ff3ef0a0854805da7edd2e5a2c58

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: Neat! https://github.com/Automattic/newspack-plugin/pull/833/commits/3dbf75e6ee52dc41c5e3a5d1dd78ee894710fb0d

assets/wizards/popups/index.js
```diff
@@ -224,14 +198,64 @@ class PopupsWizard extends Component {
...
+			method: 'POST',
+			data: { name },
+			quiet: true,
 		} )
-			.then( () => this.onWizardReady() )
+			.then( this.updateAfterAPI )
+			.catch( error => setError( error ) );
+	};
+
+	archiveCampaignGroup = ( id, status ) => {
```
dkoo, line 228: Feels a bit weird to me that archiving a campaign with active prompts won't deactivate the prompts. I get why that is (because the prompts might belong to other campaigns as well). But in many cases prompts will only belong to one campaign. Maybe we can ask the user with a modal if they want to also deactivate the prompts in the campaign? Might also be helpful when deleting a campaign.

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,319 @@
...
+									'newspack-campaigns__campaign-group__select-control-group-item',
+									+selectValue.key === +id && 'is-selected'
+								),
+							} ) ),
+							archivedCampaigns.length && {
+								key: 'header-archived-campaigns',
+								name: __( 'Archived Campaigns', 'newspack' ),
+								className: 'is-header',
+							},
+							...archivedCampaigns.map( ( { term_id: id, name } ) => ( {
```
dkoo, line 213: Wondering if we should show a label or style when viewing an archived campaign, so they're visually distinguishable from non-archived campaigns?

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,331 @@
...
+					>
+						{ __( 'Add New Campaign', 'newspack' ) }
+					</Button>
+					{ modalVisible && (
+						<Modal
+							title={ modalTitle( modalType ) }
+							isDismissible={ false }
+							className="newspack-campaigns__campaign-group__add-new-button__modal"
+						>
+							<TextControl
```
dkoo, line 284: Would love for the `TextControl` to be focused as soon as this modal renders.

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: I made this change in https://github.com/Automattic/newspack-plugin/pull/833/commits/01cd2fdd52c3b4de2c9ec72277a0580146300e5a, but curious what @thomasguillot thinks. If you don't like it, I can revert it easily.

dkoo (changes requested):
This is great—I definitely think this new UI offers more usability and convenience when dealing with campaigns (formerly groups), and I love the simplicity of a single click to preview a specific segment within a campaign.

I mainly have some suggestions which I found myself wanting as I tested on a new JN site—didn't run into any major issues except for the error when setting Sitewide Default on a pending/scheduled prompt as described below.

assets/wizards/popups/components/segment-group/index.js
```diff
@@ -0,0 +1,123 @@
...
+						className={ getCardClassName( item ) }
+						description={ descriptionForPopup( item ) }
+						key={ item.id }
+						prompt={ item }
+						{ ...props }
+					/>
+				) ) }
+			</Card>
+			{ prompts.length < 1 ? <p>{ __( 'No prompts in this segment yet.', 'newspack' ) }</p> : '' }
+			{ 'unassigned' !== campaignId && (
```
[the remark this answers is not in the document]
jeffersonrabb: Originally it was this way, but at the last minute I added the "Add New Prompt" buttons to the Active Prompts screen. Fundamentally I agree that its better to add prompts to campaigns, but I found it jarring to see no "Add New" button on the first screen you see. It felt weird to have to go hunting for that basic functionality.

assets/wizards/popups/components/prompt-popovers/primary.js
```diff
@@ -17,16 +17,16 @@ import { Popover, ToggleControl } from '../../../../components/src';
...
 	setSitewideDefaultPopup,
 	onFocusOutside,
 	publishPopup,
 	unpublishPopup,
 } ) => {
-	const { id, sitewide_default: sitewideDefault, edit_link: editLink, options, status } = popup;
+	const { id, sitewide_default: sitewideDefault, edit_link: editLink, options, status } = prompt;
 	const isDraft = 'draft' === status;
 	return (
 		<Popover
```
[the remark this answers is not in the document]
jeffersonrabb: I suspect this has been the case for a really long time! Fixed in https://github.com/Automattic/newspack-plugin/pull/833/commits/941abc5ec90444d940bac9927e6d44ebd01c94f7

assets/wizards/popups/components/segment-group/index.js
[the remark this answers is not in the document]
dkoo: I see, and that's a good point. Let's keep it around for now, in that case.

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: Would this be in the Select itself, like this?

Screen Shot 2021-02-01 at 7 16 31 PM

dkoo: That could work—I just found myself wanting to know at a glance somehow whether the campaign I'm looking at is archived.

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: This should handle the focus - https://github.com/Automattic/newspack-plugin/pull/833/commits/d03778500596268316c29a625e56ee9a183d8c4d

dkoo: Exactly what I wanted! Looks like it made the linter sad, though.

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: Done in https://github.com/Automattic/newspack-plugin/pull/833/commits/466d234ac68ec2098f19c591ee4e7abead3b14a1

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: Yup, linter is happy now: https://github.com/Automattic/newspack-plugin/pull/833/commits/0993b5a0b0999aeb652d65ee94f404d0f42b1943

assets/wizards/popups/components/segment-group/index.js
```diff
@@ -0,0 +1,123 @@
...
+					<PromptActionCard
+						className={ getCardClassName( item ) }
+						description={ descriptionForPopup( item ) }
+						key={ item.id }
+						prompt={ item }
+						{ ...props }
+					/>
+				) ) }
+			</Card>
+			{ prompts.length < 1 ? <p>{ __( 'No prompts in this segment yet.', 'newspack' ) }</p> : '' }
```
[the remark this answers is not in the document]
jeffersonrabb: Good idea, done in https://github.com/Automattic/newspack-plugin/pull/833/commits/ff3737723199f7c48835093bb87308f1c60c76ec

assets/wizards/popups/index.js
```diff
@@ -47,46 +38,54 @@ const tabbedNavigation = [
 		path: '/segmentation',
```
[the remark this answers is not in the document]
jeffersonrabb: That's a good idea - its more consistent: https://github.com/Automattic/newspack-plugin/pull/833/commits/199c9516160efa20c8c2da670ff0ab731dc628f9

assets/wizards/popups/components/segment-group/index.js
```diff
@@ -0,0 +1,123 @@
...
+		<Card isSmall className="newspack-campaigns__segment-group__card">
+			<h3 className="newspack-campaigns__segment-group__card__segment">
+				{ __( 'Segment: ', 'newspack' ) }
+				{ label }
+				<SegmentationPreview
+					campaignId={ [ campaignId ] }
+					segment={ id }
+					showUnpublished={ true } // Do we need a control for this?
+					renderButton={ ( { showPreview } ) => (
+						<Button isTertiary isSmall isLink onClick={ () => showPreview() }>
```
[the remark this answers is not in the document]
jeffersonrabb: This combined with [the suggestion to move the Add New button up](https://github.com/Automattic/newspack-plugin/pull/833#discussion_r567874088) leads to two fairly heavy buttons next to each other. I'm going to hold off for now, but if we end up reverting the button move we could revisit?

Screen Shot 2021-02-01 at 9 17 14 PM

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,319 @@
...
+										setCampaignName( '' );
+										setModalType( MODAL_TYPE_DUPLICATE );
+									} }
+									onRename={ () => {
+										setCampaignName( campaignData.name );
+										setModalVisible( true );
+										setModalType( MODAL_TYPE_RENAME );
+									} }
+									onUnarchive={ () => archiveCampaignGroup( campaignId, false ) }
+									hasUnpublished={ prompts.some( ( { status } ) => 'publish' !== status ) }
```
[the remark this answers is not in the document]
jeffersonrabb: You're right, but because this gets a bit more complex to resolve (the UI fix would just be a bandaid), why don't we wait for https://github.com/Automattic/newspack-plugin/pull/832

assets/wizards/popups/components/campaign-management-popover/index.js
```diff
@@ -0,0 +1,112 @@
...
+	<Popover
+		position="bottom right"
+		onFocusOutside={ () => dismiss() }
+		onKeyDown={ event => ESCAPE === event.keyCode && dismiss() }
+	>
+		<MenuItem onClick={ () => dismiss() } className="screen-reader-text">
+			{ __( 'Close Popover', 'newspack' ) }
+		</MenuItem>
+
+		{ hasUnpublished && (
```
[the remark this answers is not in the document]
jeffersonrabb: Seems reasonable. In https://github.com/Automattic/newspack-plugin/pull/833/commits/149d0097ac5bbb1e7eebeef37e2dc172bc3914db I've changed it so the Activate All link is present as long as there are some prompts. Deactivate All is visible only when some prompts are published.

assets/wizards/popups/components/segment-group/index.js
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const SegmentGroup = props => {
+	const { segment, campaignId } = props;
+	const [ modalVisible, setModalVisible ] = useState();
+	const { label, id, prompts } = segment;
+	return (
+		<Card isSmall className="newspack-campaigns__segment-group__card">
+			<h3 className="newspack-campaigns__segment-group__card__segment">
+				{ __( 'Segment: ', 'newspack' ) }
```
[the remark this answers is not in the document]
jeffersonrabb: Agreed: https://github.com/Automattic/newspack-plugin/pull/833/commits/9d77e145ab397fc25257a1684b2400048f17549c

assets/wizards/popups/index.js
jeffersonrabb: This change is made in https://github.com/Automattic/newspack-plugin/pull/833/commits/2d6d1d1997ddcb214a3dcf7ddc2856baabb8bd52. Requires https://github.com/Automattic/newspack-popups/pull/429

assets/wizards/popups/index.js
jeffersonrabb: Also done in https://github.com/Automattic/newspack-plugin/pull/833/commits/2d6d1d1997ddcb214a3dcf7ddc2856baabb8bd52. (Requires https://github.com/Automattic/newspack-popups/pull/429)

assets/wizards/popups/views/campaigns/index.js
```diff
@@ -0,0 +1,319 @@
...
+		wizardApiFetch( {
+			path: '/wp/v2/newspack_popups_taxonomy',
+			method: 'POST',
+			quiet: true,
+			data: {
+				name: term,
+				slug: term,
+			},
+		} )
+			.then( () => {
```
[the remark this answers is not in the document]
jeffersonrabb: A number of routing improvements are in https://github.com/Automattic/newspack-plugin/pull/833/commits/2d6d1d1997ddcb214a3dcf7ddc2856baabb8bd52. Requires https://github.com/Automattic/newspack-popups/pull/429

assets/wizards/popups/index.js
jeffersonrabb: I agree it feels weird, but I think this will require discussion. We could:

1) Unpublish all prompts within the campaign.
2) Unpublish all prompts that are in the campaign and are in no others.
3) If prompts are found in current and other campaigns, ask whether to unpublish them.
4) Do nothing (status quo).

None are perfect - what do you think?

jeffersonrabb:
Update: this branch now requires that `newspack-popups` be on https://github.com/Automattic/newspack-popups/pull/429 for testing.

assets/wizards/popups/index.js
dkoo: (3) feels like the best balance to me, but probably requires the most effort and might be beyond this PR's scope. Let's table this discussion for now and get BB's opinion for another potential PR.

assets/wizards/popups/components/segment-group/index.js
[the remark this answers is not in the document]
dkoo: I see what you mean—yes, let's table for now and revisit.

dkoo (approved):
Thanks for the improvements and feedback responses!

assets/wizards/popups/components/segment-group/index.js
```diff
@@ -0,0 +1,133 @@
...
+	return (
+		<Card isSmall className="newspack-campaigns__segment-group__card">
+			<h3 className="newspack-campaigns__segment-group__card__segment">
+				{ id ? __( 'Segment: ', 'newspack' ) : '' }
+				{ label }
+				<div>
+					<SegmentationPreview
+						campaignId={ [ campaignId ] }
+						segment={ id }
+						showUnpublished={ true } // Do we need a control for this?
```
adekbadek, line 68: It might be confusing to see draft prompt along with published ones - it's not _really_ a preview if a segment member will see only published prompts. But that might be a subject of another PR.

adekbadek (changes requested):
One small thing and we're there 💪

assets/wizards/popups/views/campaigns/index.js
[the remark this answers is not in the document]
thomasguillot: Styled in https://github.com/Automattic/newspack-plugin/pull/833/commits/5fa4c19c6d94f36c4908952884b0867649952582

assets/wizards/popups/views/campaigns/index.js
thomasguillot: @jeffersonrabb, I don't like it.

adekbadek (approved):

adekbadek (changes requested):
Whoops, approved accidentally.

assets/wizards/popups/views/campaigns/index.js
jeffersonrabb: Per @thomasguillot's UI feedback I've reverted the change in https://github.com/Automattic/newspack-plugin/pull/833/commits/ecfa5d1618edc117a87b50800083fde2971684ae

assets/wizards/popups/components/segment-group/index.js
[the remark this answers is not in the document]
jeffersonrabb: I committed this change by accident 😬 reverted in https://github.com/Automattic/newspack-plugin/pull/833/commits/cef4ac70dda03da3626f9597203e14bc00bd9fb1

assets/wizards/popups/index.js
jeffersonrabb: cc: @brianboyer

includes/wizards/class-popups-wizard.php
```diff
@@ -383,27 +479,33 @@ public function enqueue_scripts_and_styles() {
...
-			'settings' => [],
-		];
+		$response = array_merge(
+			[
+				'prompts'   => [],
+				'segments'  => [],
+				'campaigns' => [],
+				'settings'  => [],
+			],
+			$extras
```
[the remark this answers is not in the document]
jeffersonrabb: I added the casting in https://github.com/Automattic/newspack-plugin/pull/833/commits/57693f6e8a2246161d8155e9f4af26ae5563032a but I'm confused how that error would have occurred since all cases where `$extra` is used are explicitly arrays?

assets/wizards/popups/components/segment-group/index.js
jeffersonrabb: If I understand correctly, this would need to be a change in `newspack-popups`, which will require @thomasguillot to establish a UI pattern for placing admin-only informational tags on prompts. If true, I think we can file the issue and consider it a task for a new PR.

includes/wizards/class-popups-wizard.php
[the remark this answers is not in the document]
adekbadek: That's hard to trace, but I did get that error logged when loading Campaigns Wizard.

adekbadek (approved):

assets/wizards/popups/index.js
brianboyer: I agree that there is no perfect solution. To be honest, archiving campaigns makes a little less sense to me in a system where prompts aren't contained within a campaign.

We already have the ability to unpublish all prompts in the campaign, so I imagine the user's workflow would be: 1. Unpublish the prompts. 2. Archive the campaign.

Let's try option 4 and if it doesn't make sense, consider other paths?

assets/wizards/popups/components/segment-group/index.js
dkoo: I liked having the control to show or not show unpublished in our last UI iterations, but then I'm not sure how useful it is to publishers. With the control I don't think it's necessary to visually mark the drafts, though, because it becomes an action on the user's part.

thomasguillot (approved):

assets/wizards/popups/components/segment-group/index.js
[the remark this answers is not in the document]
brianboyer: Just ran into this, adding a prompt from the Active Campaigns view, and upon returning to that view, was confused that I couldn't see the prompt I just created. I agree that it's good to be able to take action when you first land on the page, but if I was confused, others will be too.

Sticking w/in the constraints of the current UI, I might change the name of "Unassigned prompts" to something more intuitive. "Assignment" isn't really a word we use anywhere. Perhaps the menu should look like:

- Currently active prompts (from all campaigns)
- Choose a campaign
  - Campaign one
  - Campaign two
  - Campaign three
- Prompts without campaigns

Same functionality, but a bit more descriptive. And moving "Prompts without campaigns" to the end makes more sense to me I think?

PR #847, opened 2021-02-03
feat: add segment descriptions to Campaigns and Segments tabs

dkoo:
### All Submissions:

Makes the descriptions for segment action cards in the Segments tab (and segment container headers in the Campaigns tab) more useful. Instead of showing the creation date, it now summarizes the configuration for each segment. This could get slightly redundant if the segments are named after the intended use, but I find it really useful to be able to see at a glance what each segment is without having to visit the edit screen.

Segments tab:

Screen Shot 2021-02-03 at 2 07 00 PM

Campaigns tab:

Screen Shot 2021-02-03 at 2 05 49 PM

Also creates a link from the segments' names in the Campaigns tab to each segment's edit page, for added convenience.

~Question: I tried to make the descriptions as terse as possible so as not to visually muddy the UI, but in some cases the phrasing differs from the option labels—should we ensure that the option names match?~ I think we should, and I implemented as a draft in #848.

1. Check out this branch, run `npm run build`
2. Visit the Campaigns and Segments tabs.
3. In both tabs, verify that each segment is accompanied by a short summary of its configuration.
4. In the Campaigns tab, verify that clicking on the segment name directs you to that segment's edit page.
5. Create new segments and edit existing segments, using a mix of every available configuration option in some segments and leaving some fields blank in others. Verify that the descriptions accurately portray the configuration for each segment, showing only the options that differ from the default.

Note that for segments with Category Affinity set, it will take a second or two for the category names to appear in the description. This is because the configuration saves category IDs only, so we need to fetch the names via REST.

### Other information:

### DIFF
--- a/assets/wizards/popups/views/segments/SegmentsList.js
+++ b/assets/wizards/popups/views/segments/SegmentsList.js
@@ -1,10 +1,9 @@
 /**
  * WordPress dependencies.
  */
+import { Tooltip, MenuItem } from '@wordpress/components';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
-import { format } from '@wordpress/date';
-import { Tooltip, MenuItem } from '@wordpress/components';
 import { ESCAPE } from '@wordpress/keycodes';
 
 /**
@@ -12,13 +11,13 @@ import { ESCAPE } from '@wordpress/keycodes';
  */
 import EditIcon from '@material-ui/icons/Edit';
 import DeleteIcon from '@material-ui/icons/Delete';
-
 import { Icon, moreVertical } from '@wordpress/icons';
 
 /**
  * Internal dependencies.
  */
 import { ActionCard, Popover, Button, Router } from '../../../../components/src';
+import { descriptionForSegment, getFavoriteCategoryNames } from '../../utils';
 
 const { NavLink, useHistory } = Router;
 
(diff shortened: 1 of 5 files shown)


assets/wizards/popups/views/segments/SegmentsList.js
```diff
@@ -131,6 +132,14 @@ const SegmentActionCard = ( {
 		resortSegments( target );
 	};
 
+	if ( 0 < segment.configuration?.favorite_categories?.length ) {
+		getFavoriteCategoryNames(
```
adekbadek, line 136: Here and in `SegmentGroup`, this function runs on each render, resulting in a ton of unnecessary requests. These calls should be wrapped in `useEffect` hooks.

adekbadek (changes requested):

assets/wizards/popups/views/segments/SegmentsList.js
dkoo: Good call—updated in 549fc70.

adekbadek (changes requested):
For a segment with multiple favorite categories set, I can only see one in the segment description. After each REST request to the category endpoint, it's name is updated.

dkoo:
> For a segment with multiple favorite categories set, I can only see one in the segment description. After each REST request to the category endpoint, it's name is updated.

Fixed in 0e251a9

adekbadek (approved):

PR #848, opened 2021-02-03
feat: reorganize segment UI

### LINKED ISSUE #828: Segmentation Problem: View Count/Referrer Conflict
For segments with a referrer defined, Articles Read and Articles Read In Session have no meaning. The segment applies only to the page viewed by clicking a link on the referrer site. This raises a few questions:

1) _Should_ the referrer be persisted? In other words, if a user clicks a link in a tweet, arrives at the site and navigates to a different post should they still be considered to be in the segment? If so, for how long? 

2) If the answer is no, we should adjust the UI to indicate this. This could be done by force-setting Articles Read and Articles Read In Session fields to "1" and disabling the inputs when a referrer is set. It also could be done by hiding these fields.

3) An alternate, more complex approach would be to define segment "types." If a segment is of type "referrer" then the visible fields change. 

4) Changes should be made to the segmentation logic in `newspack-popups` to ensure segments with a referrer and > 1 viewed values are handled.

@thomasguillot I'd be interested to hear your preferred approach.

dkoo:
### All Submissions:

Reorganizes the UI for editing segments by nesting the options into different "criteria" groups. Criteria can be enabled or disabled. When enabling a criteria group, we check for conflicting criteria and automatically disable any that were enabled. Currently the only conflicting criteria are Reader Engagement and Referrer Sources.

Note that if you have pre-existing segments with conflicting config options (e.g. referrers and min/max articles in session are already set), then this branch won't update those.

Screen Shot 2021-02-03 at 4 18 40 PM

Potentially closes #828. Some questions remain:

* Is this the right approach to address the concerns outlined in #828?
* Is this too blunt an instrument? Referrer source settings only _really_ conflict with session read settings, but in this UI enabling referrer settings will not allow you to set any engagement settings.
* This is a proof-of-concept, so I'd definitely like @thomasguillot to review and make changes as needed (IF we decide this is the right approach to address #828). I tried using only existing components so hopefully it's not a huge stretch.

1. Check out this branch, run `npm run build`.
2. In the Segments tab, create a new segment and enter a name in the Name field. Observe the new UI with all criteria disabled by default.
3. Enable the Reader Engagement and Reader Activity sections, and change some settings from the default. Confirm that the "Save" and "Cancel" buttons still behave as before (showing a warning if canceling without saving, and enabling the Save button only if the settings have changed from the default or previously saved config).
4. Enable Referrer Sources. Verify that the the Reader Engagement section is automatically disabled and that Reader Activity is unaffected. Enter some values in the Referrer fields.
5. Enable Reader Engagement again. Verify that Referrer Sources is automatically disabled.
6. Confirm that when re-enabling a section that you had previously changed settings for, your previous settings are persisted as long as you don't cancel or save the segment.
7. Ensure that all criteria sections have some fields filled out, even though only one of Reader Engagement or Referrer Sources will be able to be enabled. Note which one is disabled and click "Save" to save your segment. (Any options under disabled sections are discarded on save.)
8. Edit the segment you just saved and confirm that the criteria that had been enabled when you saved are automatically shown as enabled on load.
9. Enable the disabled criteria section and confirm that its config options were reset to the default.
10. Disable all criteria sections and confirm that the Save button is disabled (can't save if no criteria are enabled, which would match the default config).
11. Change some settings in the criteria section that had been disabled on your first save, then disable the section and re-enable the other two sections. Confirm that the Save button is disabled (can't save if the saved config would match the previously saved state, even if there are dirty options in a disabled section).

#### Bonus feature: min/max inputs

As a bonus, I fixed something that bothered me with the existing UI. Now, you can no longer enter a min value that's higher than a max value, and vice versa. To test:

1. Create or edit a segment and enable the Reader Engagement section.
2. Enter a max value for both Articles Read and Articles Read in Session.
3. Attempt to enter min values for both options that are higher than the max values. Confirm that it only lets you enter numbers that are equal to or lower than the max values, whether you use the checkboxes or text inputs.
4. Attempt to lower the max values to something lower than the min values. Confirm that it only lets you enter numbers that are equal to or greater than the min values, whether you use the checkboxes or text inputs.
5. Confirm that if min numbers are disabled or 0, you can enter any number in the max fields, and vice versa.

Ideally I would love to change this UI to something that's more appropriate for a range of numbers, but I'm not really sure what would be best. My first thought was something like the [Material range slider](https://material-ui.com/components/slider/#range-slider), but that would limit the max settings to a concrete number. Definitely open to ideas on this.

### Other information:


thomasguillot:
> This is a proof-of-concept, so I'd definitely like @thomasguillot to review and make changes as needed (IF we decide this is the right approach to address #828). I tried using only existing components so hopefully it's not a huge stretch.

@dkoo I have pushed a bunch of design-related commits 😊

dkoo:
Design updates look great—thanks @thomasguillot!

adekbadek (commented):
The looks and UI organisation are definitely much better. Here are the issues I've encountered:
1. The segment reach is not updated after a group is toggled – after setting e.g. min posts read value, the reach is updated, but after toggling the "Reader Engagement" group it's not. I'd expect it to revert back to 100% after all groups are toggled to "off"
2. After disabling all groups, then enabling "Referrer Sources", the "Save" button becomes active, even though the segment is effectively empty
3. The UI here is quite complex, so it'd make sense to add some JS tests. The project is set up with Jest & react-testing-library

dkoo:
> 1. The segment reach is not updated after a group is toggled – after setting e.g. min posts read value, the reach is updated, but after toggling the "Reader Engagement" group it's not. I'd expect it to revert back to 100% after all groups are toggled to "off"

Fixed in 88d44fa.

> After disabling all groups, then enabling "Referrer Sources", the "Save" button becomes active, even though the segment is effectively empty

Fixed in f6cf892.

> The UI here is quite complex, so it'd make sense to add some JS tests. The project is set up with Jest & react-testing-library

I might need some help with this. Also, `npm run test` is failing for me due to the following errors:

```
  ● ImageUpload › should render an add image button

Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

       6 | describe( 'ImageUpload', () => {
       7 | 	it( 'should render an add image button', () => {
    >  8 | 		const { getByText } = render( <ImageUpload /> );
         | 		                      ^
       9 | 		expect( getByText( 'Add image' ) ).toBeInTheDocument();
      10 | 	} );
      11 |

  ● ImageUpload › should render an image and remove button

    Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

      15 | 			url: 'https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg',
      16 | 		};
    > 17 | 		const { getByAltText, getByText } = render( <ImageUpload image={ image } /> );
         | 		                                    ^
      18 | 		expect( getByAltText( 'Upload preview' ) ).toBeInTheDocument();
      19 | 		expect( getByText( 'Remove image' ) ).toBeInTheDocument();
      20 | 	} );
```

adekbadek:
> Also, npm run test is failing for me

That's an issue with having multiple instances of React – if you have a `assets/components/node_modules` directory, remove that.

> I might need some help with this.

Sure, I've set up the test in c4f81eac960d1920ded2862653f96987f119520b – this should be a good overview. For debugging, a `screen.debug()` statement can be inserted in a test to print the current HTML. Jest & Testing Library have great docs, though at first it might be unclear where one ends and the other starts.

After clicking around more, I'm not 100% on board with the sections enabling/disabling idea. I love the new layout, but sections expanding & collapsing when another section is being edited feels like something is wrong.  
How about approaching it in a declarative way? We have two sections, which cannot be on at the same time. When one is on, the other should be disabled (greyed out, not collapsing). 
Example: user expands "Reader Engagement" section, "Referrer Sources" section becomes disabled. If they want to use the latter section, they have to turn off the former.

dkoo:
This is great—thanks for getting things started! I was a bit lost with the error and didn't know where to start, but this is a great help.

> How about approaching it in a declarative way? We have two sections, which cannot be on at the same time. When one is on, the other should be disabled (greyed out, not collapsing).

I'm on board with this idea. I'll try it out (and adjust tests for it).

dkoo:
@adekbadek is this enough to indicate that the section is disabled? Its options won't be factored into the Reach calculations, nor will any values be saved if the segment is saved in this state.
 
Screen Shot 2021-02-11 at 1 12 54 PM

adekbadek:
I've encountered an issue with the following scenario: enable Engagement, fill "min. posts", then enable Referrers, fill in an input there – both sections become disabled. Here's what I propose as a declarative solution: https://github.com/Automattic/newspack-plugin/pull/862. Simply add a notice & disable a section when the other one is enabled. WDYT?

(also, fixed a tiny error in bc51a884d5c5bd21477985d6a2c9f5e9d7f60a81)

adekbadek:
Closing in favor if #862

PR #850, opened 2021-02-04
chore: rename "default/no segment" to "everyone"

dkoo:
### All Submissions:

Renames the "Default/no segment" option to "Everyone" wherever it appears in the UI. Should be coupled with a corresponding change in the Campaigns plugin. Labeling this "On Hold" until we have that.

Try to find any holdover mentions of "Default/no segment"—they should all be replaced by "Everyone". UI, campaign, segment, and prompt behavior are otherwise unchanged.

### Other information:

### DIFF
--- a/assets/wizards/popups/components/prompt-popovers/secondary.js
+++ b/assets/wizards/popups/components/prompt-popovers/secondary.js
@@ -53,7 +53,7 @@ const SecondaryPromptPopover = ( {
 					onFocusOutside();
 				} }
 				options={ [
-					{ label: __( 'Default (no segment)', 'newspack' ), value: '' },
+					{ label: __( 'Everyone', 'newspack' ), value: '' },
 					...segments.map( ( { name, id: segmentId } ) => ( { label: name, value: segmentId } ) ),
 				] }
 				value={ selectedSegmentId }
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
Would this not give a false impression to the user, that a prompt segmented to this "segment" will be show to… every reader? This label makes it appear as an "override" more than a "default".

dkoo:
> Would this not give a false impression to the user, that a prompt segmented to this "segment" will be show to… every reader? This label makes it appear as an "override" more than a "default".

I see your point on this. @jeffersonrabb and @brianboyer, what do you think?

dkoo:
Thinking about this more, is it not the current behavior that prompts with "default/no segment" get shown to every reader regardless of which segment(s) they belong to? If that's not already the case then this behavior will be changing in https://github.com/Automattic/newspack-popups/pull/432 and this PR should be considered a companion to that PR.

adekbadek:
> Thinking about this more, is it not the current behavior that prompts with "default/no segment" get shown to every reader regardless of which segment(s) they belong to?

Huh, that's actually true. Ok. let's wait for https://github.com/Automattic/newspack-popups/pull/432 with that

dkoo:
Closing in favor of #853.

PR #856, opened 2021-02-10
fix: preview functionality by campaign/group

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/438, fixes previewing logic when previewing prompts within a single campaign, when previewing active prompts, or when previewing unassigned prompts.

Currently, when previewing a segment within a campaign, you will see all prompts assigned to that segment, even if they're not in the currently viewed campaign.

Also fixes a minor bug where the "Unassigned" option disappears if you're viewing the Campaigns tab while filtered by a campaign, even if there are unassigned prompts.

#### Previewing a single campaign

1. Create several ~campaigns~ (doh) prompts of all types (inline, overlay, above-header) and assign to a campaign. Make sure at least one prompt is unpublished (draft, pending or scheduled).
2. In the Campaigns tab, filter the view by your campaign.
3. Preview any segment containing prompts. Confirm that you see published and unpublished prompts:
  a. within the current campaign
  b. with the selected segment OR default/no segment, and
  c. with any segment that overlaps with the selected segment
4. Confirm that you do NOT see any prompts that are not assigned the currently viewed campaign.

#### Previewing active prompts

1. In the Campaigns tab with the same set of prompts, filter by Active Prompts.
2. Preview any segment. Confirm that you see only published prompts:
  a. with the selected segment OR default/no segment
  b. with any segment that overlaps with the selected segment, and
  c. within any campaign

#### Previewing unassigned prompts

1.  In the Campaigns tab with the same set of prompts, make sure at least a few prompts have no assigned campaign.
2. In the Campaign select dropdown, confirm that you see Unassigned Prompts no matter the current filter view.
3. Filter by Unassigned Prompts. Preview any segment and confirm that you see published and unpublished prompts:
  a. with the selected segment OR default/no segment
  b. with any segment that overlaps with the selected segment, and
  c. without any campaign
4. Confirm that you do NOT see any prompts that have an assigned campaign.

### Other information:

### DIFF
--- a/assets/wizards/popups/components/segment-group/index.js
+++ b/assets/wizards/popups/components/segment-group/index.js
@@ -55,6 +55,7 @@ const SegmentGroup = props => {
 	const [ categories, setCategories ] = useState( [] );
 	const { label, id, prompts } = segment;
 	const allPrompts = useContext( CampaignsContext );
+	const campaignToPreview = 'unassigned' !== campaignId ? parseInt( campaignId ) : -1;
 
 	useEffect( () => {
 		updateCategories();
@@ -86,9 +87,9 @@ const SegmentGroup = props => {
 					</span>
 				</a>
 				<SegmentationPreview
-					campaignId={ [ campaignId ] }
+					campaign={ campaignId ? campaignToPreview : false }
 					segment={ id }
-					showUnpublished={ true } // Do we need a control for this?
+					showUnpublished={ !! campaignId } // Only if previewing a specific campaign/group.
 					renderButton={ ( { showPreview } ) => (
 						<Button isTertiary isSmall isLink onClick={ () => showPreview() }>
 							{ __( 'Preview', 'newspack' ) }
--- a/assets/wizards/popups/components/segmentation-preview/index.js
+++ b/assets/wizards/popups/components/segmentation-preview/index.js
@@ -18,7 +18,7 @@ const SegmentationPreview = props => {
 	const frontendUrl = window?.newspack_popups_wizard_data?.frontend_url || '/';
 
 	const {
-		campaignGroups = [],
+		campaign = false,
 		onLoad = () => {},
 		segment = '',
 		showUnpublished = false,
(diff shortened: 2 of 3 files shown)


adekbadek (changes requested):
1. When previewing a segment with a prompt which has a category condition set, the prompt is not visible. After removing the category, the prompt is visible in the preview.
2. A prompt segmented to donors is showing up in the "Everyone" segment preview (both when previewing a campaign, and active prompts, regardless of whether the prompt is assigned to a campaign).

dkoo:
> 1. When previewing a segment with a prompt which has a category condition set, the prompt is not visible. After removing the category, the prompt is visible in the preview.

This is actually kind of working as intended—when previewing a prompt with category filtering, the prompt will only appear on posts that match its category. The problem is that the preview URL is just the latest post or the homepage of the site, and it's used for all preview links, so I'm not sure how we'd force it to open a post with matching category for these prompts. Or, do we fudge it and just show category-filtered prompts on any page when previewing?

> 2. A prompt segmented to donors is showing up in the "Everyone" segment preview (both when previewing a campaign, and active prompts, regardless of whether the prompt is assigned to a campaign).

This is fixed in a3a63c6 and https://github.com/Automattic/newspack-popups/pull/438/commits/83136ed4292c718572d2d684bea12d815472e238 in the other repo. Previously when previewing "Default/no segment" it assumed you were previewing as the base segment, which would match every segment and thus show all prompts. So this is a bit of a conceptual shift, although one that makes more sense to me and one that moves more toward the non-cascading segment behavior that we're planning to implement.

jeffersonrabb (approved):
LGTM

adekbadek:
> The problem is that the preview URL is just the latest post or the homepage of the site, and it's used for all preview links, so I'm not sure how we'd force it to open a post with matching category for these prompts. 

That would not be a complete solution: you can have a segment with prompts assigned to multiple categories.

> Or, do we fudge it and just show category-filtered prompts on any page when previewing?

That's what I propose https://github.com/Automattic/newspack-popups/pull/441

PR #860, opened 2021-02-11
feat: update style of the segments and prompts

thomasguillot:
### All Submissions:

* Move the "Add New Prompt" button to the Segment header section (original idea by @adekbadek)
* Rework the margins/paddings within a segment

1. Navigate to Campaigns
2. Switch to this branch

### Other information:

### DIFF
--- a/assets/components/src/action-card/style.scss
+++ b/assets/components/src/action-card/style.scss
@@ -61,7 +61,7 @@
 		}
 
 		& + & {
-			margin-top: -24px;
+			margin-top: -16px;
 		}
 
 		&.newspack-card__is-disabled,
--- a/assets/components/src/card/style.scss
+++ b/assets/components/src/card/style.scss
@@ -33,7 +33,7 @@
 		display: flex;
 		flex-wrap: wrap;
 		justify-content: space-between;
-		margin-bottom: -24px;
+		margin-bottom: -16px;
 
 		> * {
 			margin-bottom: 0 !important;
@@ -71,7 +71,7 @@
 		}
 
 		& + & {
-			margin-top: -24px;
+			margin-top: -16px;
 		}
 	}
 
--- a/assets/components/src/modal/style.scss
+++ b/assets/components/src/modal/style.scss
@@ -8,7 +8,6 @@
 .newspack-modal {
 	background: white;
 	border: 0;
-	box-shadow: 0 8px 16px rgba( black, 0.08 );
 	color: $dark-gray-300;
 	font-size: 14px;
 	line-height: 24px;
@@ -152,7 +151,7 @@
 // Overlay
 
 .newspack-modal__overlay {
-	background: rgba( $primary-500, 0.8 );
+	background: rgba( black, 0.7 );
 }
 
 // Small Screens
--- a/assets/components/src/popover/style.scss
+++ b/assets/components/src/popover/style.scss
@@ -130,11 +130,11 @@
 	}
 
 	&.is-from-bottom {
-		margin-bottom: 8px;
+		margin-bottom: 2px;
 	}
 
 	&.is-from-top {
-		margin-top: 8px;
+		margin-top: 2px;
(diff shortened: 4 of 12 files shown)


dkoo (changes requested):
I like the look of this a lot! It's much more compact and conveys the info more cleanly. I just have a few suggestions/questions:

* Now that we no longer have the extra padding for the + button at the bottom of each segment "container," the segments kind of bleed together visually. Could we add some more spacing between segments, or some other visual marker to more clearly separate them?

Screen Shot 2021-02-11 at 2 29 30 PM

* Jeff mentioned in another PR that moving the "Add New Prompt" button to the segment header, while it makes sense functionally, makes the header look visually heavy. What if we turned these two separate buttons into a popover menu hidden behind a menu icon (like the three vertical dots we've started using a lot throughout the Campaigns UI)?

* I would imagine most users won't be using this interface on mobile devices, but the header layout doesn't really collapse well at smaller viewports:

Screen Shot 2021-02-11 at 2 19 53 PM

While testing this, I had a few ideas about the "Add New Prompt" experience. Since that isn't really changing in this PR, I added them to a new issue here for your consideration: #861

thomasguillot:
@dkoo I made a bunch of changes, what do you think?

In https://github.com/Automattic/newspack-plugin/pull/860/commits/cc27526c6acd8fa855a693443f6d2ddaf699212e I moved the Preview button but it doesn't seem to work anymore. I'm not quite sure why and I could use your help 😊 🥺

dkoo:
> I made a bunch of changes, what do you think?

I like the style changes—these address everything I mentioned, thanks! The segment containers are a bit more spaced out and easier to distinguish at a glance, and the icon buttons make the segment headers look more visually balanced at all viewports.

> In cc27526 I moved the Preview button but it doesn't seem to work anymore. I'm not quite sure why and I could use your help 😊 🥺

Because of the way `SegmentationPreview` is written, it needs to be part of the initial parent render or it won't create the portal element it needs. At first I tried making the Popover element and its children always present but visible/not visible based on toggling the "more" button, but that breaks the `onFocusOutside` and escape keydown features to close the popover, which is another accessibility concern and just feels broken. So it looks like we can't easily nest it in a popover after all.

So what if instead of nesting the preview button in a popover, we turn the preview button itself into another icon? I did this in b096c8a and it seems to work again. I wasn't sure which icon would best represent "preview," so you may want to choose a better one. I also added a `flex: 1` value to the segment title so that in smaller viewports, the titles and icon buttons always take up the same amount space, like this:

Screen Shot 2021-02-16 at 4 06 54 PM

If the above looks good to you, feel free the swap out the icon, then we should have @adekbadek or @jeffersonrabb re-review since we both contributed code. Thanks again!

thomasguillot:
> So what if instead of nesting the preview button in a popover, we turn the preview button itself into another icon? I did this in [b096c8a](https://github.com/Automattic/newspack-plugin/commit/b096c8a511b3db11067b33b0b268287333b1c0f7) and it seems to work again. I wasn't sure which icon would best represent "preview," so you may want to choose a better one.

Interestingly, this is what I originally did, with the exact same icon 😊
Unfortunately, I think the icon is slightly out of place (I'll see if I can think of something else).
The idea of using a `Popover` was to introduce an "Edit Segment" button in the future.
Anyway... this will do for now.
Thank you very much for taking a look at it ✌️

dkoo (approved):
> Interestingly, this is what I originally did, with the exact same icon 😊

You know what they say about great minds 😁 

This looks good to me for now—perhaps there's a better way to implement the preview nested inside a popover menu, but it will probably require some more extensive refactoring.

dkoo:
@thomasguillot Is this ready to merge or are you holding it for something else?

thomasguillot:
> @thomasguillot Is this ready to merge or are you holding it for something else?

Not really, will merge it now 😄

PR #862, opened 2021-02-12
feat: reorganize segment UI

### LINKED ISSUE #828: Segmentation Problem: View Count/Referrer Conflict

adekbadek:
### All Submissions:

Improvements to segment editing UI.

Originally sourced from https://github.com/Automattic/newspack-plugin/pull/848.

Closes #828.

1. In Campaigns wizard, visit the Segments tab and click on a segment's name to edit it
2. Observe a new UI in which the segmentation criteria are compartmentalised to improve readability
3. Observe a notice in the "Referrer sources" informing about this criterion being applied to first page view only

### Other information:


dkoo (changes requested):
I appreciate that you're trying to make the treatment simpler, but there are few user-facing problems that I see with this approach:

* Removing the criteria updating from the `useEffect` with `segmentConfig` data means that all sections look disabled/collapsed when opening an existing segment. You have to expand each section to check whether it has values.
* When enabling both Reader Engagement and Referrer Sources, both sections get the warning and their controls become disabled. However, the disabled controls are still factored into reach calculations, and if you save the segment in this state, the settings are saved (meaning it's still possible to save an impossible segment). 
* Because the controls in both sections are disabled in the above scenario, if you enter values into each section, then enable both sections (triggering the warning/disabled controls), the only way to remedy the impossible segment is to first disable the section you want in order to clear the values in the section you don't want, then re-disable the cleared section, then re-enable the section you actually want.
* Not related to the above, but now clicking into an existing segment and trying to go back or click "Cancel" without changing any values will trigger the `isDirty` notice even though you haven't made any changes.

If we were to throw out any values in disabled controls when calculating reach and on saving, this might work, but that seems like it might cause user frustration since it's easy to trigger the disabled controls in both sections whenever both are enabled. The approach I was trying to implement in #848 allows you to overwrite one section with another—if you enable a conflicting section while another is already enabled, the newly enabled one remains editable and effectively resets the disabled section to default settings for reach calculations and saving.

adekbadek:
Thanks for the feedback! It's not only about making it simpler, but avoiding the kind of issues stemming from an imperative approach: e.g.: open Engagement, fill in min. articles read, open Referrer Sources, fill in an input there -> both sections are now disabled. 

I think we can make the declarative version work, it seems to me that this boils down to three issues:

1. Handling an existing segment (sections should be expanded)
2. Both conflicting sections enabled issue (see below)
3. Changes detected (`isDirty`) when expanding sections (both here & on #848)

> When enabling both Reader Engagement and Referrer Sources

How is this UI state possible? When one of these is expanded, the other is un-expandable (disabled).

image

dkoo:
> I think we can make the declarative version work, it seems to me that this boils down to three issues

I agree—if we can solve these issues, I think this approach could definitely work.

> How is this UI state possible? When one of these is expanded, the other is un-expandable (disabled).

That's not what I'm seeing locally. When I expand one section, the other gets a warning, but I'm still able to expand it. After that, both sections are expanded but become non-interactive:

Screen Shot 2021-02-17 at 10 39 56 AM

dkoo:
@adekbadek I figured out why my `ToggleControl` isn't getting disabled—in my local environment, I've disabled the Gutenberg plugin, and since core is running behind the plugin, I bet it has the wrong version of `ToggleControl`. When I activate the Gutenberg plugin, I see the behavior you describe.

How do we force our components to use the version in `node_modules` instead of depending on whether the plugin is installed?

adekbadek:
> So I believe the only remaining item is to figure out how to disable the ToggleControl if the site doesn't have the Gutenberg plugin enabled?

I've polyfilled the `disabled` prop handling in 0f3a8850e4c7e8d9259ef69d233ded7f1c68eef3 

I'll close the other PR since we agree on this approach.

adekbadek:
@jeffersonrabb Could you please review this PR, since both @dkoo and me have contributed?

dkoo:
Noting here (since I didn't mention before) that I updated all packages in `assets/components` to match the versions in the root plugin—I did this in an attempt to fix the issue I was seeing with `ToggleControl`. It didn't fix the issue, but I don't think there's any harm in bringing the dependencies in line with the root, right?

adekbadek:
> I don't think there's any harm in bringing the dependencies in line with the root, right?

Yeah I think it's ok.

dkoo:
@jeffersonrabb @adekbadek I had a pretty deep convo with @iamstevebeatty about the Campaigns UI today, and he had what I thought was a strong suggestion: rather than not showing the fields for disabled criteria sections, what if we showed all fields all the time, but only enabled the fields if their section is toggled on? 

This way, publishers can have a much better at-a-glance view of what criteria they can use to build a segment, without having to take any action:

Screen Shot 2021-03-25 at 1 37 07 PM

I thought this was a great idea, so I tried implementing it in 2683273 along with some minor style fixes. Also updated the Jest tests to match. WDYT?

adekbadek (commented):
These changes look good in principle but I have some second thoughts about the mutually-exclusive segments: https://github.com/Automattic/newspack-plugin/issues/828#issuecomment-808074094

dkoo:
Putting this "on hold" until we resolve the questions posed in https://github.com/Automattic/newspack-plugin/issues/828#issuecomment-808074094.

dkoo:
After some discussion in #828 we feel that the prescriptive toggling and disabling of Reader Revenue vs. Referrer Sources fields is perhaps too prescriptive. So the next steps for this PR are to update the components to remove the toggling and logic disabling certain fields when others are enabled, and instead show a lighter notice in the Referrer Sources section. Something like:

> Segments using Referrer Sources options will only apply to the first page visited during a session.

dkoo:
@adekbadek Updated in 9db7079. No more section toggling, much simplified tests, and a simple notice in Referrer Sources which is always active:

Screen Shot 2021-04-23 at 11 55 58 AM

adekbadek (commented):
Approval of the latest changes 👌 (I can't formally approve my own PR)

dkoo (approved):

Issue #852, opened 2021-02-05
Onboarding overhaul


adekbadek:
## Master thread of the [onboarding overhaul effort](https://github.com/Automattic/newspack-plugin/issues?q=is%3Aissue+is%3Aopen+label%3A%22%5BInitiative%5D+Onboarding%22+)

The initial changes are introduced in https://github.com/Automattic/newspack-plugin/pull/851. What's left:

- [x] Initial redesign, starter content (#851)
  - [x] Installation: retries on fail (#945)
  - [x] Installation: automatic redirect after installation step (https://github.com/Automattic/newspack-plugin/pull/945)
  - [x] Starter content removal (#864)
- [x] Settings: icon, title, tagline, and social accounts (https://github.com/Automattic/newspack-plugin/pull/863)
  - [x] Settings: [crop image after selection](https://github.com/Automattic/newspack-plugin/pull/863#pullrequestreview-590551296) ➡️ issue #1000
  - [x] Settings: [create social menu when social handles are set](https://github.com/Automattic/newspack-plugin/pull/863#pullrequestreview-590551296) (https://github.com/Automattic/newspack-plugin/pull/945)
- [x] Services step
- [x] Design: visual settings (#892)
  - [x] Design: Homepage layouts (#932)
  - [x] Design: custom fonts inputs (#999)
  - [x] Design: footer logo size (https://github.com/Automattic/newspack-plugin/pull/945)

PR #863, opened 2021-02-12
Setup Wizard: Settings step

adekbadek:
### All Submissions:

Settings step of the Setup Wizard.

Removes the `_newspack_profile_*` options, since they were not used anywhere anymore. Instead, WooCommerce `countrystate` & `currency` options are used. For this reason, WooCommerce is added as a required plugin during installation.  

Also, in a somewhat over-engineered way, this PR allows a component wrapped in a `withWizardScreen` HOC to render the wizard button. 

1. Install the plugin on a fresh WP site
2. Complete the first step (Settings) and verify the changes are saved: in Customizer, Reader Revenue wizard, and Yoast settings
3. Observe the site title in browser tab updating on "Site Title" input 😎

### Other information:


thomasguillot (changes requested):
I've pushed a bunch of changes.

One thing I'm not too sure about it the `ImageUpload` button. If you have a Site Icon, you click on "Replace" and then close the Modal, your Site Icon disappears.
Also, since we are requesting a square, it'd be nice to have a "Crop image" screen just like in the customizer.

For the social accounts, is it not possible to only have URLs and we somehow extract the twitter username for Yoast? Also, this doesn't seem to generate a Social Menu?

adekbadek:
> If you have a Site Icon, you click on "Replace" and then close the Modal, your Site Icon disappears.

Fixed in 5ddd8df8cb19fceca3386f5cb529b9211fdd38bf

> Also, since we are requesting a square, it'd be nice to have a "Crop image" screen just like in the customizer.

Added to https://github.com/Automattic/newspack-plugin/issues/852, the customizer flow is hard to reproduce here.

> For the social accounts, is it not possible to only have URLs and we somehow extract the twitter username for Yoast?

I'm not sure it's worth the added complexity - now we just mimic Yoast's settings.

> Also, this doesn't seem to generate a Social Menu?

No. Added to https://github.com/Automattic/newspack-plugin/issues/852

assets/components/src/image-upload/index.test.js
```diff
@@ -6,16 +6,6 @@ import ImageUpload from './';
...
-
-	it( 'should render an image and remove button', () => {
-		const image = {
-			id: 1234,
-			url: 'https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg',
-		};
-		const { getByAltText, getByText } = render( <ImageUpload image={ image } /> );
-		expect( getByAltText( 'Upload preview' ) ).toBeInTheDocument();
-		expect( getByText( 'Remove image' ) ).toBeInTheDocument();
+		expect( getByText( 'Upload' ) ).toBeInTheDocument();
```
dkoo, line 9: Should we add a test for the remove button and background image when there's an image prop?

assets/wizards/setup/views/settings/index.js
```diff
@@ -8,48 +8,134 @@ import apiFetch from '@wordpress/api-fetch';
...
+			data: { profile: profileData },
+		} );
+
+	useEffect( () => {
+		if ( typeof profileData.site_title === 'string' ) {
+			document.title = pageTitleTemplate.replace( '__SITE_TITLE__', profileData.site_title );
+		}
+	}, [ profileData.site_title ] );
+
+	const renderSetting = ( { options, label, key, type, placeholder, className } ) => {
```
dkoo, line 57: Nit: why not make this a React component?

assets/wizards/setup/views/settings/index.js
```diff
@@ -8,48 +8,134 @@ import apiFetch from '@wordpress/api-fetch';
...
+					{ renderSetting( { options: currencies, key: 'currency', label: __( 'Currency' ) } ) }
+				</Card>
+			</Grid>
+			<SectionHeader
+				title={ __( 'Social accounts', 'newspack' ) }
+				description={ __( 'Allow visitors to quickly access your social profiles', 'newspack' ) }
 			/>
+			<Grid columns={ 3 } gutter={ 32 } rowGap={ 16 }>
+				{ wpseoFields.map( seoField => (
+					<Fragment key={ seoField.key }>
```
dkoo, line 127: Is `Fragment` necessary since it only has one child?

includes/class-profile.php
```diff
@@ -70,7 +151,7 @@ public static function newspack_get_option( $key ) {
 	 * @param string $value The option value.
 	 */
 	public static function newspack_update_option( $key, $value ) {
-		update_option( NEWSPACK_PROFILE_OPTION_PREFIX . $key, $value );
+		update_option( self::$opt_prefix . $key, $value );
```
dkoo, line 154: Am I crazy or is `$opt_prefix` not defined anywhere?

dkoo (changes requested):
A couple of things:

Maybe an edge case, but if you reset a Newspack site that had been previously set up on an older version of the plugin, you might see the following for the social media links:

Screen Shot 2021-02-16 at 4 44 46 PM

> WooCommerce `countrystate` & `currency` options are used. For this reason, WooCommerce is added as a required plugin during installation.

WooCommerce was removed as a default/required plugin in #726 because several Newspack sites who use NRH complained about the overhead of having to have WC installed. I get why you added it back, but it seems like it might irk those sites again. This could be overthinking it, but what if we did something like:

- If WooCommerce is installed already when going through Newspack setup, get the currency/location info from WooCommerce.
- If it isn't installed (and it probably won't be on new sites before they choose to install to support the Newspack donation platform), then we store these as an option as before.
- When activating WooCommerce, we check for the existence of the location/currency options and update WooCommerce settings to match.

assets/wizards/setup/views/settings/index.js
adekbadek: _Something_ has to have the `key` prop. The key could be passed to the component rendered by `renderSetting`, but I think it's cleaner this way.

assets/wizards/setup/views/settings/index.js
adekbadek: Hm, it sort of is and isn't a real component - it's more of a switch statement lifted to a function. I'd keep it as it is.

assets/components/src/image-upload/index.test.js
adekbadek: 👌  Added in 967f3b788d6558a421a5f182162719228d3b9dff

includes/class-profile.php
adekbadek: Forgot to remove this method in ef196e588835342758b0ac3b17057b0faa7e7d08, fixed in 967f3b788d6558a421a5f182162719228d3b9dff

adekbadek:
> Maybe an edge case, but if you reset a Newspack site that had been previously set up on an older version of the plugin, you might see the following for the social media links:

That's odd, since these fields were now part of the setup wizard before. These might be errors - can you please inspect the return payload of the `/newspack/v1/profile` call?

> WooCommerce was removed as a default/required plugin in #726 because several Newspack sites who use NRH complained about the overhead of having to have WC installed.

Sorry, my comment was imprecise. It's still not _a required plugin_ (as it was before #726), just installed on setup. If a user wishes not to use it, they can remove it afterwards and go with any other solution. The point here is providing defaults.

assets/wizards/setup/views/settings/index.js
dkoo: Understood—I was thinking if `renderSetting` were a component, then that would get the `key`.

dkoo (approved):
> That's odd, since these fields were now part of the setup wizard before. These might be errors - can you please inspect the return payload of the /newspack/v1/profile call?

After advancing past the Settings screen, I can no longer replicate it, and I'm not seeing it in my local environment at all (which was obviously set up before checking out this branch), so let's defer until we see it happen again.

> Sorry, my comment was imprecise. It's still not a required plugin (as it was before #726), just installed on setup. If a user wishes not to use it, they can remove it afterwards and go with any other solution. The point here is providing defaults.

Ah, that makes sense to me. Thanks for clarifying.

adekbadek:
@thomasguillot Could you have another look?

thomasguillot:
@adekbadek all good for me 👍

thomasguillot (approved):

PR #868, opened 2021-02-17
fix: don't include site domain in linker

jeffersonrabb:
### All Submissions:

We've started seeing reports that clicking on the site logo leads to a redirect loop. This seems to be because the linker includes two domains `checkout.fundjournalism.org` as well as the site's domain. Because the site domain is included, the linker param is appended to links, and seems to cause  redirect loop on the home. This is generally reproducible in Safari, and Mobile Safari, but not Chrome. I haven't been able to establish a reliable way to reproduce the issue, but the change should be pretty benign. 

Verify the `linker` param no longer includes the site's URL.

### Other information:


claudiulodro (approved):
Thanks for the speedy fix!

PR #853, opened 2021-02-09
feat: only show prompts for the best-priority segment

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Along with https://github.com/Automattic/newspack-popups/pull/432, shows only prompts for the highest-priority segment you match as a reader, plus any prompts without an assigned segment (now labeled as "Everyone").

Also fixes the previewing functionality so that when previewing a segment, only prompts that belong to both the selected segment and the campaign/group you're viewing are displayed in the preview. Also fixes previewing for Active and Unassigned prompts. Behavior is now as follows:

* When previewing Active prompts, all published prompts matching the selected segment are shown in the preview.
* When previewing Unassigned prompts, all published and unpublished prompts without a campaign term that also match the selected segment are shown.
* When previewing a prompt in a campaign, all published and unpublished prompts with the matching campaign term that also match the selected segment are shown.

Includes changes from and therefore replaces #850.

Closes https://github.com/Automattic/newspack-popups/issues/445.

### How to test the changes in this Pull Request:

Follow testing instructions in https://github.com/Automattic/newspack-popups/pull/432.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #874, opened 2021-02-23
feat: allow multiple segments per prompt

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/451, allows multiple segments to be assigned to each prompt. When deciding whether to show a particular prompt to a reader, all assigned segments are taken into account and the prompt is shown if one of the assigned segments is the reader's highest-priority matching segment.

Depends on #853.

Closes https://github.com/Automattic/newspack-popups/issues/442.

#### Test behavior of existing prompts

This update should be non-destructive for existing prompts. But because it does depend on #853, we need to test that this branch's behavior is unchanged against that branch's behavior.

1. While on `try/non-cascading-priority ` in both this repo and the Campaigns repo, ensure that you have several segments with varying criteria and several prompts assigned to different segments.
2. Check out this branch and https://github.com/Automattic/newspack-popups/pull/451, run `npm run build` in both repos.
3. Without changing anything, view your site's front-end in a new incognito session and confirm that prompt display behavior is unchanged for all prompts with only a single segment assigned, or no segment ("everyone").
4. In the Campaigns tab, preview various existing segments and confirm that only prompts in that segment, plus prompts in "Everyone", are displayed in the preview. This behavior should be still be unchanged so far.

#### Test Campaigns wizard UI

Screen Shot 2021-02-22 at 7 18 49 PM

1. In the Campaigns tab, click on the cog icon for a prompt. Observe that the "Segments" field is now an autocomplete field instead of a select dropdown. Confirm that this field is prepopulated with the segment previously assigned to each prompt.
2. Click into the Segments field and start typing the name of another segment. Confirm that the field shows segment suggestions matching your input, only if the segment isn't already assigned to the prompt.
3. Add a few additional segments using the field. Confirm that after assigning multiple segments to the prompt, the prompt's action card now appears under every assigned segment's heading.
4. Remove a segment from the autocomplete field. Confirm that the prompt is removed from that segment's heading.
5. Change the filter view from Active Prompts to various campaigns or Unassigned Prompts. Repeat steps 1-4 and confirm that all still works as expected.

#### Test prompt editor sidebar UI

1. Click on a prompt to edit it, or create a new prompt.
2. Expand the "Segmentation Settings" sidebar panel. It should now look something like the below, with an autocomplete search field followed by checkboxes for each individual segment option. Confirm that for new prompts, "Everyone" is selected by default, while for existing prompts, the previously assigned segment(s) should be selected by default:

Screen Shot 2021-02-22 at 6 54 10 PM

3. Test enabling multiple segments (not "Everyone"). Confirm that you're able to select any number of non-Everyone segments. Update the prompt, refresh the page, and confirm that your selections are saved.
4. Enable the "Everyone" segment. Confirm that all other segment selections are cleared when "Everyone" is enabled. Confirm that if no other segments are selected, you're unable to clear the "Everyone" option.
5. Enable one or more other segments. Confirm that "Everyone" is cleared if at least one other segment is selected.
6. Start typing a segment name into the autocomplete field. Confirm that the field shows segment suggestions matching your input, only if the segment isn't already selected.
7. Click on a suggestion or select it with the arrow keys + enter. Confirm that on selection, the field is cleared and the corresponding segment's checkbox is enabled.
8. Type "Everyone" into the search field and select the suggestion. Confirm that on selection, Everyone is selected and all other segments are cleared.
9. Confirm that the help message ("this prompt will be shown...") updates itself when making selections and makes sense in the context of the selected options.

#### Test prompt display logic

The crucial difference is that a prompt with multiple assigned segments is displayed if one of its segments is the reader's highest-priority matching segment.

1. Make sure at least one active prompt has at least two segments with non-overlapping criteria. For example:
  - Segment A is New Readers, or min articles read 1, max articles read 3. Set this to a higher priority.
  - Segment B is Subscribers, or newsletter subscribers only. Set this to a lower priority.
2. Visit a post in a new incognito session. Confirm that the prompt is displayed because your highest-priority matching segment is "New Readers", which is assigned to the prompt.
3. In the same session, visit three other posts so that your max articles read count exceeds 3.
4. Confirm that on the fourth post view, the prompt is no longer displayed because you no longer match either assigned segment.
5. Visit another post and add `?utm_medium=email` to the URL (or otherwise subscribe to make yourself a newsletter subscriber). Confirm that the prompt is displayed once again because your highest-priority matching segment is now "Subscribers," which is also assigned to the prompt.

**Note:** If you've [enabled debugging](https://github.com/Automattic/newspack-popups/tree/master/api) in the Campaigns API, you can now see your highest-priority matching segment listed as `matching_segment` in the `debug` object when inspecting the `popups` XHR request in dev tools > Network.

#### Test category filtering, conflict warning, etc.

1. In the Campaigns wizard, confirm that a conflict warning is shown if multiple overlays or above-headers share at least one segment and category (or all have no category).
2. Test previewing and front-end display of category filtered prompts. Category-filtered prompts' display behavior should be unchanged as long as the reader's highest-priority segment is assigned to the prompt.
3. Test previewing and front-end display of prompts assigned to "Everyone." These prompts should still display for all readers, unless they're overlay or above header and there's at least one other overlay/above header prompt that has your highest-priority matching segment. For example, if you have an overlay with the New Readers segment assigned and another overlay assigned to Everyone, the prompt with New Readers should display in a new incognito session. If you deactivate the New Readers overlay, then the Everyone prompt should display. This behavior is unchanged from `try/non-cascading-priority`, but because we touched on the display logic, we should confirm.
4. Test previewing various segments and prompts in general.

### Other information:


assets/wizards/popups/components/segmentation-preview/index.js
```diff
@@ -28,6 +33,8 @@ const SegmentationPreview = props => {
 	const decorateURL = urlToDecorate => {
 		const view_as = segment.length ? [ `segment:${ segment }` ] : [ 'segment:everyone' ];
 
+		view_as.push( `cid:${ uniqueId( 'view_as_client_' ) }` );
```
adekbadek, line 36: Seems like there's something off with the `cid` URL param - the previewing navigation is still treated as a series of separate sessions. Maybe this `cid` value is recreated on navigation events?

assets/wizards/popups/components/segment-group/index.js
```diff
@@ -78,25 +79,94 @@ const SegmentGroup = props => {
...
+								{ __( 'Segment: ', 'newspack' ) }
+								{ label }
+							</Button>
+						) : (
+							label
+						) }
+					</h3>
 					<span className="newspack-campaigns__segment-group__description">
-						{ descriptionForSegment( segment, categories ) }
+						{ id ? descriptionForSegment( segment, categories ) : __( 'No segment', 'newspack' ) }
```
adekbadek, line 101: Suggestion: "All readers regardless of segment"

adekbadek (changes requested):

assets/wizards/popups/components/segment-group/index.js
dkoo: Good idea, updated in ccc3f17c.

assets/wizards/popups/components/segmentation-preview/index.js
dkoo: Ah yes, because we run `decorateUrl` on every anchor link in the preview window, and the spoofed `cid` was getting created as part of that function. Moving the `cid` creation outside of the function fixes it in 34eb565b.

adekbadek: Pointed out an issue with client ID spoofing here: https://github.com/Automattic/newspack-popups/pull/451#discussion_r583494755

dkoo: Removed all client ID spoofing from this PR and https://github.com/Automattic/newspack-popups/pull/451 for now. Let's handle that in #891 and https://github.com/Automattic/newspack-popups/pull/458.

adekbadek (approved):

PR #881, opened 2021-02-24
fix: duplication of segments when dragging while re-sorting

dkoo:
### All Submissions:

Fixes a duplication bug with segments resulting from the sort UI.

If you try to re-sort the segments list (either by dragging and dropping or by clicking the up/down arrow buttons) while a sort request is already in progress, it results in a malformed segments list with duplicate segment items. The duplicated items will have the same ID and configuration, so renaming or deleting one copy will result in the same action being applied to its copies.

This fix can't remedy the bad segments if the issue has already happened, but it will prevent the issue from happening going forward. The only fix for the issue if it's already happened is to delete all bad segment copies and rebuild them from scratch.

1. On `master`, go to Newspack > Campaigns > Segments and create several segments.
2. Re-sort the list by either dragging and dropping a segment or by clicking the up/down arrow buttons. Very quickly, while the view is still in a loading state, drag and drop another segment or click the up/down arrow buttons to move the segment. Observe that you now have multiple copies of the same segment in the list:

Screen Shot 2021-02-23 at 6 17 17 PM

3. Also observe in the Campaigns tab that the segment may appear multiple times:

newspack1 ngrok io_wp-admin_admin php_page=newspack-popups-wizard

4. Check out this branch and run `npm run build`.
5. Go back to the Segments tab and attempt to replicate step 2. Confirm that this is now impossible because UI interaction is blocked while the view is in a loading state.
6. Make sure to delete any bad segment copies resulting from step 2 and rebuild, if necessary.

### Other information:

### DIFF
--- a/assets/wizards/popups/views/segments/SegmentsList.js
+++ b/assets/wizards/popups/views/segments/SegmentsList.js
@@ -313,7 +313,10 @@ const SegmentsList = ( { wizardApiFetch, segments, setSegments, isLoading } ) =>
 				<h2>{ __( 'Audience segments', 'newspack' ) }</h2>
 				<AddNewSegmentLink />
 			</Card>
-			<div className="newspack-campaigns-wizard-segments__list" ref={ ref }>
+			<div
+				className={ 'newspack-campaigns-wizard-segments__list' + ( inFlight ? ' is-loading' : '' ) }
+				ref={ ref }
+			>
 				{ segmentsToShow.map( ( segment, index ) => (
 					<SegmentActionCard
 						deleteSegment={ deleteSegment }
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
While that resolves the issue described, dragging a segment is still possible while in the loading state - how about disabling the handle, too? Or, if that's not possible, the whole UI can be made non-interactive with a `pointer-events: none` CSS rule. 
Also, to make this more robust, the duplicated segments can be filtered out server-side.

dkoo:
> the whole UI can be made non-interactive with a pointer-events: none CSS rule.

Good suggestion, added in e2932117.

adekbadek (approved):

PR #885, opened 2021-02-24
Update the Footer to support a simpler style

thomasguillot:
### All Submissions:

The new `simple` prop added to `Footer` allows wizards to use the `Footer` without all the links and only display the logos of WordPress.com and Google News Initiative.

This is particularly useful for the Onboarding. Currently the `Footer` is being suppressed which is not ideal. With this PR we now have the 2 logos at the bottom.

1. Go through the Onboarding. Notice there isn't a Footer like on any other Wizard
2. Switch to this branch
3. Go through the Onboarding again, notice the logos at the bottom (no links)
4. Check a different wizard and see if you have the different links in the footer

### Other information:

### DIFF
--- a/assets/components/src/footer/index.js
+++ b/assets/components/src/footer/index.js
@@ -6,7 +6,6 @@
  * WordPress dependencies.
  */
 import { __ } from '@wordpress/i18n';
-import { Component } from '@wordpress/element';
 import { ExternalLink } from '@wordpress/components';
 
 /**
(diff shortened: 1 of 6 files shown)


assets/components/src/with-wizard/index.js
```diff
@@ -250,7 +250,7 @@ export default function withWizard( WrappedComponent, requiredPlugins ) {
 		 * Render.
 		 */
 		render() {
-			const { suppressFooter } = this.props;
+			const { footerSimple, footerSuppress } = this.props;
```
dkoo, line 253: is `footerSuppress` actually used anywhere? I couldn't find an instance of it being set, so I believe it will always be `false` and therefore can be removed from here and line 279, right?

assets/components/src/footer/index.js
```diff
@@ -14,78 +15,86 @@ import { ExternalLink } from '@wordpress/components';
...
-						<li key={ index }>
-							{ external ? (
-								<ExternalLink href={ url }>{ label }</ExternalLink>
-							) : (
-								<a href={ url }>{ label }</a>
-							) }
-						</li>
-					) ) }
-				</ul>
+class Footer extends Component {
```
dkoo, line 18: React best practice as I understand it is to use functional components whenever possible. Since it seems you only converted to a class-based components to extract the `simple` prompt, couldn't we keep this as a functional component and get the prop from args? e.g. 

`
const Footer = ( { simple } ) => {
`

dkoo (changes requested):
Looks good and works well, just a couple of minor nitpicks with the code below.

assets/components/src/footer/index.js
thomasguillot: I did this in order to have consistency with our other components but you're right, it's not needed 😊 
Fixed in https://github.com/Automattic/newspack-plugin/pull/885/commits/20fd6a98a041d10f4fe77f48082d6ac6ff5812ee

assets/components/src/with-wizard/index.js
thomasguillot: I kept it in case we were going to use it some day but it doesn't make sense to keep it (footer should always be present -- simple or not)
Removed in https://github.com/Automattic/newspack-plugin/pull/885/commits/972ed944203e8f4804ac2dffded3fa4a2c93d363

assets/components/src/footer/index.js
dkoo: Understood! Yeah, I think if we want to strive toward consistency across components we should refactor the class-based ones to be functional wherever we can, not the other way around. Thanks!

dkoo (approved):
Looks good now, thanks for the changes!

PR #886, opened 2021-02-24
feat: validated segmentation sort and error handling

### LINKED ISSUE #882: Add server-side validation of segments list before executing a re-sort
#881 illustrates that the new segment sorting UI has a weakness: because the sorting comes from the client-side render, it's possible to mess up your segments list pretty badly. Even after #881 lands (which fixes a bug where you can duplicate segments if you sort while a sort op is already in progress), some problems persist. For example:

1. Go to Newspack > Campaigns > Segments
2. Create a new segment and save.
3. Open a second tab in the same browser and go to Newspack > Campaigns > Segments again. Observe that the segments list is the same across both tabs.
4. In tab 1, delete the new segment.
5. In tab 2, without refreshing, drag and drop the new segment to another spot in the list.
6. Refresh tabs 1 and 2. Observe that the new segment has come back from the dead despite having been deleted in step 4.

As part of the `segmentation-sort` endpoint handler, we should validate the updated list of segments being passed against the existing list of segments from the options table, so that:

- The number of items is the same in both lists
- Each item in the updated list has a unique ID (no duplicates)
- Each item in the updated list has an ID that exists in the list from the options table

If any of the above conditions fail, we should abort the sorting operation and return a WP Error.

Maybe we should also consider fetching an updated segments list every time you start a sort op, and abort if the initial order doesn't match. This would help prevent a similar problem if you sort the list a certain way in one tab while another tab is open, then sort in the other tab (the order persisted to the database will match tab 2).

Open to other suggestions! 

**Additional context**
This should hopefully guard against most potential screw-ups due to bad info coming from the client-side UI being persisted in the database.

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/455, adds server-side validation of segmentation sort request params. This should prevent most instances of bad client-side data being persisted in the options table. If the Popups API endpoint returns an error, display the error in the UI.

Closes #882.

Confirm that you can no longer replicate the issue described in #882 either by deleting or adding segments in one tab but not the other.

If you try to sort with outdated data, the error messaging should look like this:

Screen Shot 2021-02-24 at 11 16 40 AM

Note that it's still possible to have two tabs open, re-sort in tab 1, then re-sort in tab 2, and have tab 2's sort order override tab 1's. But that's much less of a problem in my mind than creating malformed lists, deleting segments that had been previously added, or bringing back segments that had been deleted.

### Other information:

### DIFF
--- a/assets/wizards/popups/views/segments/SegmentsList.js
+++ b/assets/wizards/popups/views/segments/SegmentsList.js
@@ -16,7 +16,7 @@ import DeleteIcon from '@material-ui/icons/Delete';
 /**
  * Internal dependencies.
  */
-import { ActionCard, Card, Popover, Button, Router } from '../../../../components/src';
+import { ActionCard, Button, Card, Notice, Popover, Router } from '../../../../components/src';
 import { descriptionForSegment, getFavoriteCategoryNames } from '../../utils';
 
 const { NavLink, useHistory } = Router;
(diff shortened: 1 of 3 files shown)


jeffersonrabb (approved):
Works well!

PR #891, opened 2021-02-26
Replace CID cookie on preview triggering

adekbadek:
### All Submissions:

On preview, replace the Client ID cookie, so that a preview session is a fresh session.

1. Use `try/view-as-cookie-spoofing` branch of `newspack-popups` (https://github.com/Automattic/newspack-popups/pull/458)
2. Use the campaign/segment preview feature. Dismiss a campaign permanently and ensure you've seen another one, which is set to "once" frequency
3. Navigate the site in preview window, observe that the two prompts are not visible anymore
4. Close the preview, open it again - observe the prompts are visible again
5. Close the preview, logout, and visit the site - the prompts should behave as in a new session 

### Other information:

### DIFF
--- a/assets/wizards/popups/components/segment-group/index.js
+++ b/assets/wizards/popups/components/segment-group/index.js
@@ -113,10 +113,24 @@ const SegmentGroup = props => {
 								isQuaternary
 								isSmall
 								onClick={ () => {
-									cookies.remove( 'newspack-cid' );
-									cookies.set( 'newspack-cid', `preview-${ Date.now() }`, {
-										domain: window.location.host.replace( /[^\.]+\./, '.' ),
-									} );
+									if ( newspack_aux_data.popups_cookie_name ) {
+										// Remove cookies for all possible domains.
+										window.location.host
+											.split( '.' )
+											.reduce( ( acc, _, i, arr ) => {
+												acc.push( arr.slice( -( i + 1 ) ).join( '.' ) );
+												return acc;
+											}, [] )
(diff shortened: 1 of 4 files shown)


assets/wizards/popups/components/segment-group/index.js
```diff
@@ -110,7 +112,14 @@ const SegmentGroup = props => {
 							<Button
 								isQuaternary
 								isSmall
-								onClick={ () => showPreview() }
+								onClick={ () => {
+									cookies.remove( 'newspack-cid' );
```
dkoo, line 116: This isn't working for me, because the `domain` and `path` attributes set for this cookie don't match the `js-cookie` defaults. It works if I change it like so:

```
cookies.remove( 'newspack-cid', {
	domain: '.' + window.location.host,
	path: '/',
} );
cookies.set( 'newspack-cid', `preview-${ Math.round( Math.random() * 9999 ) }`, {
	domain: '.' + window.location.host,
	path: '/',
} );

console.log( cookies.get( 'newspack-cid' ) );
```

I think more testing is needed to ensure that these attributes remain constant across environments, though.

adekbadek: Good catch - ensured the cookie is removed regardless of domain in 57748ee1706fdd71316a835576e753da41a0b51f

dkoo (approved):
This works as described! Preview sessions now work as actual sessions, where frequency and dismissal rules behave as they do on the front-end.

PR #892, opened 2021-03-01
Onboarding: add Design step

adekbadek:
### All Submissions:

Add the Design step of the new Onboarding. Also, the UI is reused in the Design Wizard - this can be removed though. 

Some features to be tackled separately added to https://github.com/Automattic/newspack-plugin/issues/852

1. Visit the onboarding, Design step (`/wp-admin/admin.php?page=newspack-setup-wizard#/design`)
2. In different tabs, open the Design Wizard and the Customizer
3. Edit the options, obeserve they can be set & read in all three UIs
4. Observe the "Preview Site" floating button saves the settings & opens the preview - but it's available only in the onboarding 

### Other information:


thomasguillot (changes requested):
I haven't fully tested this yet but I've noticed an issue with the fonts.

We are only importing the normal weight (`400`). Can we also have italic and `700`?

`https://fonts.googleapis.com/css2?family=FONT+NAME:ital,wght@0,400;0,700;1,400;1,700&display=swap`

I tested this with PT Mono (which only have `400`) and Old Standard TT (which only has `400`, `400i`, and `700`) and it seems to work even if there are weights/styles missing.

thomasguillot:
I have a few more changes for the typography.

* Let's remove `System Font` for now as it doesn't work and it might require some tweaking in the theme
* Can we update the "fallback" font depending on the type? (`Display` and `Monospace` can fallback to `sans-serif` for now)

thomasguillot:
> * Let's remove `System Font` for now as it doesn't work and it might require some tweaking in the theme

It actually works since `https://fonts.googleapis.com/css?family=System+Font` returns a 404 but the fallback is wrong (`serif`) not sure we want a 404 though

adekbadek:
> We are only importing the normal weight (400). Can we also have italic and 700?
> Let's remove System Font for now as it doesn't work and it might require some tweaking in the theme
> Can we update the "fallback" font depending on the type?
 
All done in 5445cca0d10bbc47df1d56a96fe572541cb81d5d

assets/wizards/site-design/views/main/index.js
```diff
@@ -0,0 +1,275 @@
...
+			<SectionHeader
+				title={ __( 'Theme', 'newspack' ) }
+				description={ __( 'Activate a theme you like to get started', 'newspack' ) }
+			/>
+			<ThemeSelection theme={ themeSlug } updateTheme={ updateThemeSlug } />
+			<SectionHeader
+				title={ __( 'Colors', 'newspack' ) }
+				description={ __( 'Define your primary and secondary colors', 'newspack' ) }
+			/>
+			<Grid gutter={ 32 }>
```
dkoo, line 71: This UI is a bit confusing to me. In the Customizer, we at first have two options, "Default" and "Custom":

Screen Shot 2021-03-05 at 2 05 30 PM

Only when you select "Custom" do you see the color pickers.

In the new Onboarding and Wizard screens, the default behavior seems to be the same as "Default", but the color pickers both show `#fff`, which doesn't actually seem to be the case. Maybe we should have a toggle control to set "Default" vs. "Custom", and only show the pickers if the control is set to Custom? This would mirror the Customizer and make the default state a bit more understandable, IMO.

assets/wizards/site-design/views/main/index.js
```diff
@@ -0,0 +1,275 @@
...
+					onChange={ updateMods( 'primary_color_hex' ) }
+				/>
+				<ColorPicker
+					label={ __( 'Secondary' ) }
+					color={ mods.secondary_color_hex }
+					onChange={ updateMods( 'secondary_color_hex' ) }
+				/>
+			</Grid>
+			<SectionHeader
+				title={ __( 'Typography', 'newspack' ) }
```
dkoo, line 84: (Not a blocker) This is maybe more of a UX/design question than related to code, so maybe for @thomasguillot? I dig that we have the select dropdowns with Google Font options here! Much more user-friendly than entering font name strings. However, we do lose a bit of the flexibility that the Customizer UI provides (being able to use Typekit fonts, for example). Is there a middle ground here?

Screen Shot 2021-03-05 at 2 21 24 PM

Also, I love the user-friendliness of the dropdowns, but I found myself wanting to see what each font looks like. We don't need to get fancy with this, but maybe we can provide external links to the Google Fonts pages for selected fonts? e.g. https://fonts.google.com/specimen/Montserrat

A nice enhancement in the future might be to show an actual block of text with a heading and update the font dynamically based on the selections.

assets/wizards/site-design/views/main/index.js
```diff
@@ -0,0 +1,275 @@
...
+					}
+				/>
+			</Grid>
+			<ToggleControl
+				checked={ mods.accent_allcaps === true }
+				onChange={ updateMods( 'accent_allcaps' ) }
+				label={ __( 'Use all-caps for accent text', 'newspack' ) }
+			/>
+			<SectionHeader
+				title={ __( 'Header', 'newspack' ) }
```
dkoo, line 119: (Not a blocker) I know that the logo image preview is a preview of the logo only, but its place in the UI makes it look like it represents a preview of the header itself (especially because it updates with different color settings). Would be great if changing other options updated the preview as well. Something like:

* Alignment: We can update the `background-position` of the image upload component to left/center/right.
* Size (Header): We could update the dimensions of the image upload component itself.
* Size (Logo): We could update the dimensions of the image inside the image upload component to approximate the rendered size.
* Display Site Title/Tagline: These might be tricker to render with respect to the logo, but I really wanted to see them if we can figure out a way to show them easily on top of the image logo preview.

The above is definitely more of a nice-to-have and not a blocker.

assets/components/src/hooks/useOnClickOutside.js
```diff
@@ -0,0 +1,38 @@
+// https://usehooks.com/useOnClickOutside/
```
dkoo, line 1: Clever, I'm going to bookmark this site! 🙂

assets/wizards/site-design/views/main/index.js
```diff
@@ -0,0 +1,275 @@
...
+						height: '96px',
+						...( mods.footer_color === 'custom' && mods.footer_color_hex
+							? { backgroundColor: mods.footer_color_hex }
+							: {} ),
+					} }
+					image={ mods.newspack_footer_logo }
+					onChange={ updateMods( 'newspack_footer_logo' ) }
+				/>
+			</Grid>
+			{ hasPreview && (
```
dkoo, line 253: This "floating" preview button looked a bit lost and bolted on to me. I had a hard time finding at first, actually:

Screen Shot 2021-03-05 at 2 50 09 PM

Could we consider moving this to a simpler secondary-style button right next to the "Finish" button instead?

assets/wizards/site-design/views/main/utils.js
```diff
@@ -0,0 +1,138 @@
...
+			label,
+			value: label,
+		};
+		if ( isHeadingsOnly ? headingsOnly : true ) {
+			acc.push( selectOption );
+		}
+		return acc;
+	}, [] );
+
+export const getFontsList = headingsOnly =>
```
dkoo, line 20: (Not a blocker) Future enhancement idea? It would be great to source this list from the [Google Fonts API](https://developers.google.com/fonts/docs/developer_api).

assets/wizards/site-design/views/main/index.js
```diff
@@ -0,0 +1,275 @@
...
+				<div className="newspack-floating-button">
+					<WebPreview
```
dkoo, line 255: Something happened to the viewport buttons in the preview modal. The SVGs look tiny:

Screen Shot 2021-03-05 at 2 13 15 PM

dkoo (changes requested):
A couple of minor issues and some suggestions, feel free to move any to #852 if we want to discuss or tackle them separately.

assets/wizards/site-design/views/main/index.js
thomasguillot: > I dig that we have the select dropdowns with Google Font options here! Much more user-friendly than entering font name strings. However, we do lose a bit of the flexibility that the Customizer UI provides (being able to use Typekit fonts, for example). Is there a middle ground here?

In my mockup, I actually have a radio option to switch between Curated list and Custom font. But, I think for a v1, it's OK not to have the Custom font option since it's still accessible via the Customizer.

Screenshot 2021-03-09 at 10 07 49

Worth noting that with Newspack Lite, we will only offer the Curated list (current version).

>  found myself wanting to see what each font looks like. We don't need to get fancy with this, but maybe we can provide external links to the Google Fonts pages for selected fonts?
>
> A nice enhancement in the future might be to show an actual block of text with a heading and update the font dynamically based on the selections.

Here's what I think:

* We already have a "Preview Site" button, so the publication can check the font(s) they selected.
* In a potential v2, I pictured a small iframe with live changes applied to it so it'd be easier for a publication to see the changes
* Having an external link to the font or a random block of text doesn't give proper accuracy when it comes to font combination, layout, and context. I much prefer a "Preview Site" button 😊

assets/wizards/site-design/views/main/index.js
thomasguillot: No because it'd mean having to scroll all the way down to preview the site.

v2 with the iframe will give a much better visibility but I'm ok with the way it is right now for v1.

Having said that, I might play around with a full-width floating button.

assets/wizards/site-design/views/main/index.js
adekbadek: > the default behavior seems to be the same as "Default", but the color pickers both show #fff, which doesn't actually seem to be the case. 

How to reproduce that? When I set the colors to "Default" in the customizer, I can still see the primary & secondary colors in the  Onboarding and Wizard screens.

> Maybe we should have a toggle control to set "Default" vs. "Custom", and only show the pickers if the control is set to Custom? 

The idea here was to simplify color picking - instead of default vs custom there's just a color picker.

assets/wizards/site-design/views/main/index.js
adekbadek: Hm, that would duplicate the preview feature. I'd rather opt for a single complete preview than fragmented previews. WDYT @thomasguillot?

thomasguillot: > that would duplicate the preview feature. I'd rather opt for a single complete preview than fragmented previews

Agreed.

> Display Site Title/Tagline

Speaking of the Site Title/Tagline. Can we remove the toggles? (cc @adekbadek) If there's a logo then Site Title and Tagline should be set to "hidden". If a publication wants to display the Site Title and/or the Tagline alongside their logo, then they can always use the Customizer (fwiw it only represents ~8% of all our current publishers)

adekbadek: Done in a5b9b9020db7a1c98690e2b32c03feae3234d9f0

assets/wizards/site-design/views/main/index.js
adekbadek: Update the colors logic to handle the theme's custom vs. default colors in 7dcf5c5ddd90b24eb4cd859777c197718f3b782f - maybe that resolves the issue

includes/wizards/class-setup-wizard.php
```diff
@@ -240,12 +234,25 @@ public function api_retrieve_theme() {
...
+						'url' => is_array( $attachment ) ? $attachment[0] : null,
+					];
+				}
 			}
 		}
+		$theme_mods['theme_colors'] = get_theme_mod( 'theme_colors', 'default' );
+		if ( 'default' === $theme_mods['theme_colors'] ) {
+			$theme_mods['primary_color_hex']   = newspack_get_primary_color();
+			$theme_mods['secondary_color_hex'] = newspack_get_secondary_color();
+		}
```
dkoo, line 250:
```suggestion
		if ( 'default' === $theme_mods['theme_colors'] ) {
			$theme_mods['primary_color_hex']   = newspack_get_primary_color();
			$theme_mods['secondary_color_hex'] = newspack_get_secondary_color();
		} else {
			$theme_mods['primary_color_hex']   = get_theme_mod( 'primary_color_hex', newspack_get_primary_color() );
			$theme_mods['secondary_color_hex'] = get_theme_mod( 'secondary_color_hex', newspack_get_secondary_color() );
		}
```

assets/wizards/site-design/views/main/index.js
dkoo: > How to reproduce that?

Sorry I didn't get to posting replication steps. I can say that 7dcf5c5 fixes the issue partially for me, but the suggestion I made below fixes it completely. To test:

1. In the Customizer, choose "Custom Colors" and set a hex value for primary and secondary colors
2. In the Site Design page, observe that the primary color is shown, but the secondary shows `#fff`
3. Apply the suggested patch and refresh Site Design. Confirm that both custom colors are shown.
4. In the Customizer, reset the colors back to "Default."
5. In the Site Design page, confirm that the colors shown are now the theme's default colors.

dkoo (changes requested):
See suggested change for custom colors.

thomasguillot (changes requested):
Looks great. I have 2 more requests and that should be it 😊

1. Can we have the Site Title automatically added to the Copyright text input? If so, we should remove the info text as it wouldn't be needed anymore.

Screenshot 2021-03-12 at 13 14 22

2. Can we merge the 2 header background colour toggles?

Screenshot 2021-03-12 at 13 10 09

When enabled, the colour picker would default to the Primary colour (note if the user then changes the Primary, this should be updated as well, unless the user already picked a different colour)

Screenshot 2021-03-12 at 13 17 29

Screenshot 2021-03-12 at 13 18 43

Screenshot 2021-03-12 at 13 18 56

Screenshot 2021-03-12 at 13 19 17

thomasguillot:
Once https://github.com/Automattic/newspack-theme/pull/1302 is merged, we should add a size option for the footer logo

assets/wizards/site-design/views/main/index.js
adekbadek: Ah, that makes sense! Thanks, added in 0f75d5b36196dcf0225e13d83a53cfc7bdd5b586

adekbadek:
@thomasguillot 

> Can we merge the 2 header background colour toggles?

Done in a89ae8f853ef8055e0d77f7fd86c42b75c8b03b3. Note the header color will be forced to `'custom'`, so that it has the same result for all themes. Some Newspack themes use the primary color as the default, some a hard-coded default.

> When enabled, the colour picker would default to the Primary colour (note if the user then changes the Primary, this should be updated as well, unless the user already picked a different colour)

The header color will be set to the primary color initially. But when the primary color is changed, it will not be updated - this is not how this is handled in the themes (primary is not a fallback for the header color). I'd prefer to avoid a disconnect between how the colors are handled by the themes and this new UI - this would necessitate a lot of ugly workarounds.

adekbadek:
@thomasguillot 

> Can we have the Site Title automatically added to the Copyright text input? If so, we should remove the info text as it wouldn't be needed anymore.

Done in 1d60acd063fbb4d795986a6f55ab58e39e05a636

thomasguillot (approved):
LGTM, thank you @adekbadek

dkoo (approved):
Looks good now, thanks!

PR #898, opened 2021-03-04
feat: custom placements

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/467, creates a new system for placing prompts manually in post or page content. Prompts assigned to a custom placement will be rendered on the page and displayed according to segmentation rules (one prompt per segment per block).

This MVP provides three custom placements that can be placed in post content via a new Custom Placement block. Could be extended to allow editors to create and update their own custom placements, too, but this isn't included yet.

This is a doozy, so apologies in advance for the long testing instructions and if I missed any logic!

Closes https://github.com/Automattic/newspack-popups/issues/444.

#### "Manual" frequency deprecation

This feature will deprecate the "Manual Placement" frequency option and automatically migrate existing prompts with that setting to the new Custom Placements system. To test:

1. On `master` in this repo and Campaigns, create several "Manual Placement" prompts. Insert them into various posts and pages using shortcodes and verify that they are rendered on the front-end according to segmentation rules.
2. Check out this branch and https://github.com/Automattic/newspack-popups/pull/467, and run `npm run build` in both repos.
3. Visit the front-end where you placed the shortcodes and confirm that the prompts are still rendered where placed, according to segmentation rules.
4. In the Campaigns dashboard, confirm that all prompts that were previously assigned the "Manual Placement" frequency now have the following settings:
  - Frequency: "Until Dismissed", the default for new prompts
  - Placement: Custom Placement 1
5. Visit other posts and pages and confirm that these prompts are never inserted programmatically into post content.

#### Campaigns wizard updates

1. On any prompt action card, click the settings menu button (Screen Shot 2021-03-04 at 3 18 28 PM).
2. Confirm there's now a "Placement" select dropdown in the popover menu.
3. Confirm that the options in the Placement dropdown depend on the type of prompt: Overlays should show overlay options ( "Top", "Bottom" and "Center"), while Inline and Above Header prompts should show inline options ("Inline", "Above Header", and "Custom Placement" 1 through 3). This is a declarative way of ensuring that you don't accidentally end up with invalid settings by changing the prompt type from overlay to inline and vice versa.
4. Confirm that changing the setting in this dropdown saves the placement option after a page refresh.
5. Confirm that if the setting is changed to a Custom Placement option, the badge in the prompt's action card changes to say "Custom Placement."

Screen Shot 2021-03-04 at 3 32 34 PM

6. Confirm that if more than one prompt shares the same segment and category (or all have no category), a conflict warning is shown with each prompt:

Screen Shot 2021-03-04 at 3 46 36 PM

7. After changing the setting, edit a prompt and confirm that the updated placement setting is reflected in the Prompt Settings sidebar as well.

#### Creating Custom Placement prompts

1. In the Campaigns wizard, create a new prompt under any segment. Confirm that the "Manual Placement" option now says "Custom Placement."
2. Choose the "Custom Placement" option to create a new prompt. When the editor loads, expand the Prompt Settings sidebar panel and confirm that the "Inline Prompt" setting is enabled and the placement is set to "Custom Placement 1."
3. Toggle OFF the "Inline Prompt" setting. Confirm that Custom Placement options are no longer available in the Placement dropdown and the setting is automatically changed to "Center". (Only inline prompts can be placed manually.)
4. Create some content and publish the prompt.
5. Create several other Custom Placement prompts:
  - In the same Custom Placement
  - In other Custom Placements
  - With different and shared categories, and no category
  - With the same segments, and 
  - With different segments (make sure all segments are testable in incognito sessions)

#### Custom Placement block

1. Edit a page or post. In the block inserter, confirm that there's a new block called "Newspack Campaigns: Custom Placement" under the Newspack category.
2. In a different tab, edit a prompt and confirm that the new block does NOT appear in the block inserter (to avoid nesting prompts inside prompts).
3. Back in the page, add a Custom Placement block. The empty state should look like this:

Screen Shot 2021-03-04 at 3 31 50 PM

4. Update the page and view the front-end in a new incognito session. Confirm that the empty block with no custom placement option selected renders nothing.
5. In the editor, choose a custom placement. Confirm that the block updates itself to display a simple list of the active prompts that are assigned to that placement. Also confirm that you can click on each prompt's name to edit that prompt.

Screen Shot 2021-03-04 at 3 34 34 PM

6. Confirm that if you choose a custom placement that has no active prompts, the block will warn you and the front-end will render nothing.

Screen Shot 2021-03-04 at 3 35 39 PM

7. Choose a custom placement with some active prompts with no categories, and update the page.
8. View the page front-end while logged in as an admin. Inspect the page's markup in dev tools and confirm that the prompts are rendered in the content markup, but are not displayed under any circumstances (this matches behavior of all other types of prompts, which are never displayed to logged-in admin users).
9. View the page front-end in a new incognito session. Confirm the following logic:
  - A Custom Placement block will render markup for one active prompt per segment that belongs to the placement. (Verify in dev tools.)
  - Conversely, if multiple active prompts with the same segment are assigned to the same placement, only the most recently created prompt is rendered.
  - Prompts are displayed according to segmentation rules—they are displayed only if the reader's best-priority matching segment is assigned to the prompts. (This means that each block may render multiple prompts, but a maximum of one will be displayed.)

Note that it's possible to add multiple instances of the Custom Placement block to a single post or page. If the various instances share the same prompts, prompts may duplicated on the front-end.

#### Custom Placement block with categorized prompts

1. In the Campaigns wizard, assign at least one category to two Custom Placement prompts with the same segment, keeping the other Custom Placement prompts uncategorized. Confirm that a conflict warning is shown for prompts that share the same category, segment, and custom placement:

Screen Shot 2021-03-04 at 4 44 32 PM

2. Edit a post that shares the category you assigned to the prompts and insert a Custom Placement block. Select the Custom Placement that's assigned to the category-filtered prompts. Update the post.
3. In a new incognito session, make sure you match the segment assigned to the prompts. View the categorized post and confirm that you see only one (the most recent) prompt in the custom placement that has a matching category.
4. Add Custom Placement blocks with the same custom placement setting to other posts that don't share the same category as assigned to the prompts.
5. View those non-category-matching posts on the front-end and confirm that you only see prompts that don't have any category filtering set, that also match your reader segment. If the chosen custom placement doesn't have any uncategorized prompts, you should see nothing.

#### Previewing custom placement prompts

1. In the Campaigns wizard, preview various segments, including Everyone.
2. In the preview window, view some pages and posts with Custom Placement blocks and confirm that you see the prompts you expect based on the previewed segment and post category.

### Other information:


dkoo:
@adekbadek Boyer gave some feedback on the conflict messaging. I've pushed some changes to address that feedback in 322c46e and 668b9e7—could you take another look, please?

adekbadek (approved):
This messaging logic is getting complex, would use some tests.

dkoo:
> This messaging logic is getting complex, would use some tests.

Good idea—added some tests in b563a219.

assets/wizards/popups/components/segment-group/SegmentGroup.test.js
```diff
@@ -0,0 +1,302 @@
...
+
+		render(
+			<MemoryRouter>
+				<SegmentGroup { ...mockProps } />
+			</MemoryRouter>
+		);
+	} );
+
+	it( 'renders a conflict notice for uncategorized overlays in the same segment', async () => {
+		await waitForElement( () => screen.getByText( `Segment: ${ SEGMENT.label }` ) );
```
adekbadek, line 256: Why should the tests await anything? There are no async action to wait for, the tests pass with these lines commented out.

adekbadek (commented):
Perhaps a nit, but while the test is checking for the notice messages existence, it does not say much about the expectations for the code. The thing being tested here is: _will a message be displayed in two spots in the UI under certain conditions?_ So I'd expect the tests to check where are the notices rendered and how many of them are there. Just a remark, as there are multiple approaches to testing (I like the school that says tests should serve as a code specification - describing how the code should be used).

assets/wizards/popups/components/segment-group/SegmentGroup.test.js
dkoo: Removed in 01902e4b.

assets/wizards/popups/utils.js
```diff
@@ -197,7 +236,10 @@ const sharesSegments = ( segmentsA, segmentsB ) => {
 export const warningForPopup = ( prompts, prompt ) => {
```
adekbadek, line 236: There is no conflict message when prompts have no segment assigned:

image

adekbadek (changes requested):
One issue slipped through an earlier review

assets/wizards/popups/utils.js
dkoo: Ah, you're right—fixed in e6b29ed5

adekbadek (approved):

dkoo:
@adekbadek I reworked the tests in fa761cb and 2f93a7b, could you take another look, please?

assets/wizards/popups/components/segment-group/SegmentGroup.test.js
```diff
@@ -0,0 +1,462 @@
...
+			campaign: CAMPAIGN.campaignId,
+			segment: SEGMENT,
+		};
+		const { getByTestId } = render( <SegmentGroup { ...props } /> );
+		const notice1 = getByTestId( 'conflict-warning-1' );
+		const notice2 = getByTestId( 'conflict-warning-2' );
+
+		// Elements are in the document.
+		expect( notice1 ).toBeInTheDocument();
+		expect( notice2 ).toBeInTheDocument();
```
adekbadek, line 276: Nit - in testing-library, all `getBy*` functions fail the test if element is not found, so technically checking for existence is redundant.

adekbadek (approved):
Cool, these tests are signalling what the component does much better!

assets/wizards/popups/components/segment-group/SegmentGroup.test.js
dkoo: Good deal, removed in 459751e7!

PR #905, opened 2021-03-16
Fix starter content issues

### LINKED ISSUE #899: Starter Content: Removing it removes same-named categories
Note: This may not be a bug with how debug mode/the starter content is intended to work going forward. I ran into this setting up a staging site; I was running the wizard to make sure all of the appropriate plugins were installed and configured; I'm filing it as an issue compared to how the starter content used to work, as an optional step rather than a required one. 

**Describe the bug**
If you run the Newspack wizard with debug mode enabled, it will automatically install starter content. Clicking 'Remove Starter Content' works, but if you also happen to have any categories with the same names as the starter content -- like Features or Opinion -- they will also be removed. 

**To Reproduce**
1. Start on a test site with some content with a few of the starter content categories, like 'Opinion'. 
2. Switch the plugin to debug mode by adding `define( "WP_NEWSPACK_DEBUG", true );` to the `wp-config.php` file.
3. Run the Newspack Wizard.
4. At the end, click 'Remove Starter Content`.
5. Check your categories; note that the ones that overlap with the starter content are missing. 

**Screenshots**
Test site before running the setup wizard, with four posts under Opinion:

image

Categories after running setup wizard:

image

Categories after clicking "Remove Starter Content" -- my original five posts are there, but now they're all filed under Uncategorized, and the original Opinion category is gone:

image

**Expected behavior**
I'm not sure the best outcome here - of course, the starter content shouldn't add duplicate categories, either, so it makes sense that it didn't create a second Opinion category. 

Ideally the starter content wouldn't be installed unless triggered, or at least there would be a prompt in cases where the site isn't empty. 

The starter categories could also be replaced with something that's unlikely to actually be a category on a real site -- like Lorem Ipsum. The downside is it would make the content look "less real", but there wouldn't be the need to try to keep track of what categories already existed.

### LINKED ISSUE #903: Setup Wizard: Overwrites existing customizations
Note: this is a similar issue to #899, in that I'm not sure how the setup wizard is imagined to be used, and it's also related to the starter content. I could be misunderstanding it, and where it fits with our process of setting up sites for publishers. 

**Describe the bug**
If you run the setup wizard after customizing the site in any way, you'll overwrite some of your customization work -- specifically, the logo, the header settings, and your selected homepage. 

Right now I've been skipping the setup wizard due #899 to when Newspackifying a site, but there has now been a couple cases where it's run later, and work is undone. 

**To Reproduce**
1. Install the Newspack plugin and theme on a fresh site (edited to add: this recreates more of our usual staging site situation, not a brand-new site situation -- the themes are already in place).
2. Install and activate the Newspack Blocks plugin.
3. Go about customizing your site: add a logo, menu, and set up a homepage. Keep the header with the default settings (default height, default background).

Screen Shot 2021-03-12 at 8 04 16 AM

4. Navigate to WP Admin > Newspack and run the Setup Wizard.
5. View the site again: 

Screen Shot 2021-03-12 at 8 07 40 AM

Some settings (colours, fonts, menus, Custom CSS) are kept, and your homepage itself isn't entirely overwritten, a new page is created and replaces it under Customizer > Homepage Settings. Both your logo and header settings are overwritten, though.

**Expected behavior**
I would expect if a logo or front page is set, it wouldn't be overwritten.

For the header settings, it's a bit tougher -- in my case, they were using the default settings (which is a totally viable setup), so there's nothing really to check for. 

Again, this might be a case where we need some kind of prompt before adding the starter content to the site.

### LINKED ISSUE #904: Option for installing demo content
The new UI in #851 removed the option to not install demo content (now, demo content is always installed). We should add the option back because not all sites will need or want the demo content. This will let us avoid #899 for new sites until we have a proper fix for that issue.

Mockup courtesy @thomasguillot:

Screenshot 2021-03-12 at 17 06 38

adekbadek:
### All Submissions:

Closes #899 
Closes #903 
Closes #904

1. On a new site, add two categories: "Opinion", "Health"
2. Install the plugin, but don't do the setup just yet. Install & activate Newspack theme.
3. Add a custom logo via Customizer
4. Open the setup wizard. Observe that starter content creation is optional. Uncheck this option and install Newspack
5. Observe no starter content created
6. Go through the installation step again, this time with starter content creation
7. Observe starter content created, but original categories intact - all created categories have prefixed slugs
8. Observe the logo (in Customizer) intact 
9. Remove starter content (first enter debug mode by setting `WP_NEWSPACK_DEBUG` constant)
10. Observe the starter content removed and original categories unaffected

### Other information:

### DIFF
--- a/assets/wizards/setup/views/welcome/index.js
+++ b/assets/wizards/setup/views/welcome/index.js
@@ -21,9 +21,11 @@ import {
 	NewspackLogo,
 	ProgressBar,
 	withWizardScreen,
+	CheckboxControl,
 } from '../../../../components/src';
 
 const POST_COUNT = newspack_aux_data.is_e2e ? 12 : 40;
+const STARTER_CONTENT_REQUEST_COUNT = POST_COUNT + 3;
 
 const REQUIRED_SOFTWARE_SLUGS = [
 	'jetpack',
@@ -36,8 +38,6 @@ const REQUIRED_SOFTWARE_SLUGS = [
 	'newspack-newsletters',
 	'newspack-theme',
 ];
-
-const TOTAL = POST_COUNT + 4 + REQUIRED_SOFTWARE_SLUGS.length;
 
 const ERROR_TYPES = {
 	plugin_configuration: { message: __( 'Installation', 'newspack' ) },
@@ -53,9 +53,14 @@ const starterContentFetch = endpoint =>
 const Welcome = ( { buttonAction } ) => {
 	const [ installationProgress, setInstallationProgress ] = useState( 0 );
 	const [ softwareInfo, setSoftwareInfo ] = useState( [] );
+	const [ shouldInstallStarterContent, setShouldInstallStarterContent ] = useState( true );
 	const [ errors, setErrors ] = useState( [] );
 	const addError = errorInfo => error =>
 		setErrors( _errors => [ ..._errors, { ...errorInfo, error } ] );
+
+	const total =
+		( shouldInstallStarterContent ? STARTER_CONTENT_REQUEST_COUNT : 0 ) +
+		REQUIRED_SOFTWARE_SLUGS.length;
 
 	useEffect( () => {
 		document.body.classList.add( 'newspack_page_newspack-setup-wizard__welcome' );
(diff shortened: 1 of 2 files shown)


dkoo (approved):
This works as described. The only funny thing I saw is that after removing starter content, two starter content posts stuck around, even after reinstalling starter content and removing again. I think it's probably due to some hiccup from running setup multiple times (one time did error out so I had to refresh and try again), so I'm going to approve.

PR #913, opened 2021-03-19
Remove WooCommerce One Page Checkout from Newspack distribution

avcascade:
### All Submissions:

This PR removes WooCommerce One Page Checkout from the Newspack distribution. Once this change is shipped, One Page Checkout would no longer appear on the list of plugins included with Newspack. 

**Why remove One Page Checkout?**

Because no Newspack publishers are using it, and it's therefore not needed. 

Originally, we were thinking of having the Donate product page also be the Checkout page, but we didn't end up needing that since we created the Donate Block instead. 

Because WooCommerce One Page Checkout is a premium WooCommerce extension, it cannot be auto-updated. Its removal will make Newspack a little lighter and nimbler. 

Install the Newspack plugin and verify that One Page Checkout no longer appears as a plugin listed under Newspack (with the &mdash; prefix.) 

### Other information:

### DIFF
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -391,13 +391,6 @@ class Plugin_Manager {
 				'PluginURI'   => 'https://www.woocommerce.com/products/name-your-price/',
 				'AuthorURI'   => 'https://kathyisawesome.com',
 			],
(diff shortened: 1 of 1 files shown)


claudiulodro (approved):
Looks good. Thanks for the contribution :)

PR #914, opened 2021-03-19
Leaner GA config

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes #911.

### How to test the changes in this Pull Request:

1. On `master`, concoct a big GA config by [enabling NTG events](https://newspack.pub/support/analytics/#ntg-tagging), adding some custom events, and then making it so a ton of prompts is included on a page. You can also inflate the config size by setting a super long title on one of the prompts.
2. Visit a page and observe https://github.com/ampproject/amphtml/issues/32911 - if the GA config is larger than 10kB, the request to `https://www.googletagmanager.com/gtag/amp` should fail, triggering a `Error rewriting configuration` error. Observe the GA events triggered by prompts being loaded, seen, and dismissed, are not sent (query network requests with `collect`)
3. Switch to this branch and also `add/leaner-ga-config` on `newspack-popups`
4. Observe the issues from pt. 2 are gone (unless your config is _still_ >10kB after the changes 😱)
5. After ~24h the new data should be available to Campaigns Wizard's Analytics - verify the new data is there

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #920, opened 2021-03-29
Fix Campaigns Analytics reporting

adekbadek:
### All Submissions:

In #914, the GA report fetching was [changed to accommodate both legacy and new event category](https://github.com/Automattic/newspack-plugin/pull/914/files#diff-dcf593d710ae2fbb84ba284a1878a96389c35e092f59c5fd1cda39942dafb9fcL27-R30), but the [operator on the event category](https://github.com/Automattic/newspack-plugin/pull/914/files#diff-dcf593d710ae2fbb84ba284a1878a96389c35e092f59c5fd1cda39942dafb9fcR117) query was not updated, which results in a bad query.

The error would cause no events to show up after the latest release.

1. Visit the Campaigns Wizard, Analytics screen
2. You should see both old and new events on the chart

### Other information:


dkoo (approved):
👍 I confirmed that data recorded since the change is now showing (so no data was lost).

PR #931, opened 2021-04-07
Fix onboarding issues

adekbadek:
### All Submissions:

Fixes two issues.

1. On `master`, 
2. Visit the Design step of the onboarding. Observe that after clicking "Preview" you're redirected to Newspack dashboard
3. Visit the Services step of the onboarding. Observe the donation amounts and Stripe settings inserted there are not reflected in the Reader Revenue wizard
4. Switch to this branch, observe both issues resolved

### Other information:

### DIFF
--- a/includes/wizards/class-reader-revenue-wizard.php
+++ b/includes/wizards/class-reader-revenue-wizard.php
@@ -351,21 +351,18 @@ class Reader_Revenue_Wizard extends Wizard {
 	}
 
 	/**
-	 * Save Stripe settings.
+	 * Handler for setting Stripe settings.
 	 *
-	 * @param WP_REST_Request $request Request object.
-	 * @return WP_REST_Response Boolean success.
+	 * @param object $settings Stripe settings.
+	 * @return WP_REST_Response with the latest settings.
 	 */
-	public function api_update_stripe_settings( $request ) {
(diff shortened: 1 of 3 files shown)


includes/wizards/class-reader-revenue-wizard.php
```diff
@@ -402,7 +398,24 @@ public function api_update_stripe_settings( $request ) {
...
+	 *
+	 * @param WP_REST_Request $request Request object.
+	 * @return WP_REST_Response Response.
+	 */
+	public function api_update_stripe_settings( $request ) {
+		$params = $request->get_params();
+		$result = $this->update_stripe_settings( $params );
+
+		if ( is_wp_error( $result ) ) {
+			return rest_ensure_response( $result );
```
dkoo, line 415: Since we're still just returning the result in the case of a WP_Error, why even include this condition? Also, why the different namespacing on this line vs. line 418?

dkoo (approved):
Works as described, just have a non-blocking question below

includes/wizards/class-reader-revenue-wizard.php
adekbadek: That's right, removed this bit in 7157c007b690e225a3363810f76547a5a1e053b7

PR #936, opened 2021-04-16
Handle GAM ad units from Ads plugin

adekbadek:
### All Submissions:

New handling of ad units, since they are now integrated with GAM.

Follow testing steps in https://github.com/Automattic/newspack-ads/pull/124

### Other information:


dkoo (changes requested):
Great work! Just a couple of things below.

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -21,27 +26,74 @@ class AdUnits extends Component {
...
+								}
+								onSecondaryActionClick={ () => onDelete( id ) }
+								{ ...( is_legacy
+									? {
+											secondaryActionText: __( 'Delete' ),
+									  }
+									: {
+											href: `#${ service }/${ id }`,
+											actionText: __( 'Edit' ),
+											secondaryActionText: __( 'Archive' ),
```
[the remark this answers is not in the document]
adekbadek: Good idea - reworked a bit the action card in 9242cbac9a7a91fcfda0f64a0efefb8e82eddd0b

assets/wizards/advertising/index.js
```diff
@@ -368,88 +317,74 @@ class AdvertisingWizard extends Component {
...
-										onSave={ id =>
-											this.saveAdUnit( id ).then( () => {
-												routeProps.history.push( '/google_ad_manager' );
-											} )
-										}
-									/>
-								);
-							} }
-						/>
+								<Route
```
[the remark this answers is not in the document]
adekbadek: Fixed in 6ee27e35151af9f406f5c204215d65219e90185a

assets/wizards/advertising/views/ad-units/index.js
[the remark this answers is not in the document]
dkoo: The reworked UI works well functionally, but will re-test if you implement the changes @thomasguillot suggest.

assets/wizards/advertising/index.js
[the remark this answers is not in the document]
dkoo: Nice, looks fixed now.

dkoo:
> Could we move the Edit and Archive buttons to a `Popover`, similar to the Campaigns?

I second this change, and also would suggest that we move "Activate" and "Deactivate" actions to this popover instead of using the toggle switch (mainly for consistency, I do like the toggle switch but it threw me off for a second because I was expecting a pattern closer to the Campaigns popover).

adekbadek:
> Would it be possible to move the "Add New" button to the top, and the "Back" button to the header?
> Could we move the Edit and Archive buttons to a Popover, similar to the Campaigns?

I'd prefer to do that separately, these changes are substantial already. Also see https://github.com/Automattic/newspack-ads/pull/124#issuecomment-854763212

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -5,38 +5,167 @@
...
+											toggleChecked: adUnit.status === 'ACTIVE',
+											toggleOnChange: value => {
+												adUnit.status = value ? 'ACTIVE' : 'INACTIVE';
+												updateAdUnit( adUnit );
+											},
+									  } ) }
+								description={ () => (
+									<span>
+										{ gamConnectionStatus.can_connect && adUnit.is_legacy ? (
+											<i>{ __( 'Legacy ad unit.', 'newspack' ) }</i>
```
dkoo, line 140: Can we add a separator after this flag?

Screen Shot 2021-06-09 at 4 09 41 PM

Expected:

_Legacy ad unit_ | Sizes: `300x250`

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -5,97 +5,168 @@
...
+	wizardApiFetch,
+	service,
+	gamConnectionStatus = {},
+} ) => {
+	const warningNoticeText = `${ __(
+		'Please connect your Google account using the Newspack dashboard in order to use ad units from your GAM account.',
+		'newspack'
+	) } ${
+		Object.values( adUnits ).length
+			? __( 'The legacy ad units will continue to work, but they are not editable.', 'newspack' )
```
dkoo, line 40: Is this true anymore? Aren't legacy units now editable?

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -5,97 +5,168 @@
...
+		wizardApiFetch( {
+			path: '/newspack/v1/wizard/advertising/network_code/',
+			method: 'POST',
+			data: { network_code: networkCode },
+			quiet: true,
+		} );
+	};
+
+	return (
+		<Fragment>
```
dkoo, line 61: Is this extra wrapping `Fragment` necessary? Can't we return before the final render if `gamConnectionStatus.can_connect` is false? At the very least maybe we can stick with `<>` since we're not applying any props.

dkoo (changes requested):
Works well, just a few minor nitpicks below.

assets/wizards/advertising/views/ad-units/index.js
adekbadek: Fixed in bb65fde8276a12fd3e56c6c98c1a0cf5380c9ff0

assets/wizards/advertising/views/ad-units/index.js
adekbadek: Good catch, fixed in bb65fde8276a12fd3e56c6c98c1a0cf5380c9ff0

assets/wizards/advertising/views/ad-units/index.js
adekbadek: > Is this extra wrapping Fragment necessary? 

That's a list of children components 🤷‍♀️

> Can't we return before the final render if gamConnectionStatus.can_connect is false?

The legacy ad units have to remain editable, even if GAM is unavailable.

> At the very least maybe we can stick with <> since we're not applying any props.

Done in bb65fde8276a12fd3e56c6c98c1a0cf5380c9ff0

dkoo: Oh, I see what you mean. Thanks for explaining!

dkoo (approved):

thomasguillot (approved):

PR #940, opened 2021-04-23
feat: integrate Newspack Scheduled Post Checker into main plugin

### LINKED ISSUE #939: Integrate newspack-scheduled-post-checker plugin
@tmmbecker mentions that she's been installing the [newspack-scheduled-post-checker plugin](https://github.com/Automattic/newspack-scheduled-post-checker) on a lot of Newspack sites. Since it's such a small and useful tool, we should consider integrating it into this main Newspack plugin and deprecating the separate plugin repo.

Alternatively, we could keep the plugin separate but whitelist it as a managed plugin and let it be installed automatically as part of the Newspack product suite. This would require that we set up the separate repo with the requisite CI workflows.

I think the easiest solution is to bring over the code from the other plugin and integrate it here. Since both plugins are licensed under GPL2, there'd be no loss to the open source community. 

The alternative, if we'd like to keep the plugin separate, is to bring it up-to-date with our deployment and release management workflows, then whitelist it as part of the larger Newspack suite. This might be a more attractive option for larger or more complex plugins (or ones that not every Newspack site might need), but newspack-scheduled-post-checker is very small, very low-risk, and extremely useful for any site that publishes a lot of content frequently.

cc @claudiulodro @adekbadek

dkoo:
### All Submissions:

Pulls in code from the separate [Newspack Scheduled Post Checker plugin](https://github.com/Automattic/newspack-scheduled-post-checker) and integrates it into this main Newspack plugin. After this releases the separate repo can be scheduled (hah) for deprecation.

Note that the integrated version of the code is namespaced and has a slightly different constant name, so that it won't cause conflicts if a site already has the separate plugin installed and activated.

Closes #939.

1. On `master`, install and activate the [Newspack Scheduled Post Checker plugin](https://github.com/Automattic/newspack-scheduled-post-checker).
2. Using WP CLI, verify that the cron job is scheduled to run every five minutes: `wp cron event list --hook=newspack_scheduled_post_checker`
3. Check out this branch. Verify that the site behaves as expected with no PHP errors or warnings.
4. Also verify that `wp cron event list --hook=newspack_scheduled_post_checker` still shows the same scheduled cron job as before (this change won't overwrite or re-add the job if it already exists).
5. Deactivate the separate Newspack Scheduled Post Checker plugin.
6. Verify that `wp cron event list --hook=newspack_scheduled_post_checker` still shows a scheduled cron job, but with different `next_run_gmt` and `next_run_relative` values (once the other plugin is deactivated, this one will take over and make sure the cron job is scheduled).

### Other information:

### DIFF
--- a/includes/scheduled-post-checker/scheduled-post-checker.php
+++ b/includes/scheduled-post-checker/scheduled-post-checker.php
@@ -49,23 +49,7 @@ function nspc_run_check() {
 		]
 	);
 
-	if ( function_exists( 'spcl_log_event' ) ) {
-		$message = 'Newspack Scheduled Post Checker running';
-		$data    = [
-			'posts_with_missed_schedule' => $posts_with_missed_schedule,
-			'time'                       => $time,
-		];
-		spcl_log_event( $message, $data );
-	}
-
 	foreach ( $posts_with_missed_schedule as $post_id ) {
-		if ( function_exists( 'spcl_log_event' ) ) {
-			$message = 'Trying to publish post with missed schedule';
-			$data    = [
-				'id' => $post_id,
-			];
-			spcl_log_event( $message, $data );
-		}
 		check_and_publish_future_post( $post_id );
 	}
 }
--- a/includes/scheduled-post-checker/scheduled-post-checker.php
+++ b/includes/scheduled-post-checker/scheduled-post-checker.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Newspack Scheduled Post Checker
+ * Checks to make sure posts haven't missed their schedule, and publishes them if needed.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Scheduled_Post_Checker;
+
+defined( 'ABSPATH' ) || exit;
+
+define( 'NEWSPACK_SCHEDULED_POST_CRON_HOOK', 'newspack_scheduled_post_checker' );
+
+/**
+ * Set up the checking.
+ */
+function nspc_init() {
+	register_deactivation_hook( __FILE__, 'nspc_deactivate' );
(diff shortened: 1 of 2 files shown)


includes/scheduled-post-checker/scheduled-post-checker.php
```diff
@@ -0,0 +1,87 @@
...
+			'date_query'  => [
+				[
+					'before'    => $time,
+					'inclusive' => false,
+				],
+			],
+		]
+	);
+
+	if ( function_exists( 'spcl_log_event' ) ) {
```
adekbadek, line 52: Where does `spcl_log_event` come from? A comment would be handy here.

adekbadek (approved):

includes/scheduled-post-checker/scheduled-post-checker.php
dkoo: Great question, this comes directly from the original plugin. @claudiulodro, can you provide any context?

claudiulodro: All of the `spcl_` stuff is originally for a debug logger. We definitely don't want it in here now, since we know it works correctly and the `spcl_log_event ` function won't exist on any sites.

claudiulodro (changes requested):
Thanks for adding this @dkoo! We should remove the logger stuff, then this looks good to merge.

includes/scheduled-post-checker/scheduled-post-checker.php
dkoo: Ah, thanks for the context! I've removed all references to that function in ca5ffb4.

claudiulodro (approved):
LGTM. Thanks for making that revision.

PR #961, opened 2021-05-14
Add new ButtonCard component

thomasguillot:
### All Submissions:

This PR adds a brand new component: `ButtonCard` to be used in the Dashboard and the Onboarding (for the upcoming migration step).

1. Switch to this branch
2. Check the components demo
3. Check the dashboard

### Other information:

### DIFF
--- a/assets/components/src/button-card/style.scss
+++ b/assets/components/src/button-card/style.scss
@@ -1,0 +1,104 @@
+/**
+ * Button Card
+ */
+
+@import '../../../shared/scss/colors';
+@import '~@wordpress/base-styles/colors';
+
+.newspack-button-card {
+	border-radius: 3px;
+	border: 1px solid $light-gray-500;
+	box-shadow: inset 0 0 0 2px transparent;
+	cursor: pointer;
+	display: block;
+	margin: 32px 0;
+	padding: 32px;
+	text-decoration: none;
+	transition: background-color 125ms ease-in-out, border-color 125ms ease-in-out,
+		box-shadow 125ms ease-in-out;
+
+	&:active,
+	&:focus,
+	&:hover {
+		background-color: $light-gray-100;
+		color: $primary-600;
+		text-decoration: none;
+
+		.title,
+		.desc {
+			color: inherit;
+		}
+	}
+
+	&:focus {
+		border-color: $primary-500;
+		box-shadow: inset 0 0 0 2px $primary-500 !important;
+		outline: none !important;
+		position: relative;
+	}
+
+	.title,
+	.desc {
+		line-height: 24px;
+		transition: color 125ms ease-in-out;
+	}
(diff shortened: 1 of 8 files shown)


assets/wizards/dashboard/index.js
```diff
@@ -106,30 +101,28 @@ const Dashboard = ( { items } ) => {
 						</div>
 					) }
 					{ displayAuth ? (
-						<Card className={ classnames( 'newspack-dashboard-card', 'google-oauth2' ) }>
+						<>
 							{ userBasicInfo ? (
-								<div className="newspack-dashboard-card__contents">
-									<Icon icon={ plugins } />
-									<div className="newspack-dashboard-card__header">
+								<Card className="newspack-dashboard-card">
```
dkoo, line 106: Why a regular `Card` here—is it just so we can use the `strong` in the message? Wonder if we could adapt the `ButtonCard` to render `children` if it exists instead of (or in addition to) the `desc` prop?

assets/components/src/button-card/style.scss
```diff
@@ -0,0 +1,104 @@
...
+
+		svg {
+			fill: $primary-500;
+			transition: fill 125ms ease-in-out;
+		}
+	}
+
+	// Is Small
+
+	&.is-small {
```
dkoo, line 101: The `is-small` styles don't look much different from the regular styles. Wonder if we should make the type styles smaller as we do with the `is-small` version of the `ActionCard` component? For example, the card from Campaigns:

Screen Shot 2021-05-14 at 1 52 38 PM

vs. the new `ButtonCard` with small styles:

Screen Shot 2021-05-14 at 1 52 50 PM

dkoo (changes requested):
Looks good! Just a couple of minor nits below.

assets/components/src/button-card/style.scss
thomasguillot: We're not using isSmall so I kind of created a very basic style but you're right, it should look different from the regular size...  Anyway, I pushed some changes in https://github.com/Automattic/newspack-plugin/pull/961/commits/97d9c099313fbce70d700d145c869611f5fb2e72

I also added a `chevron` option... Why not?! 😂

dkoo: Nice, looks much more consistent now!

dkoo (approved):
Approving because my remaining question isn't a blocker if you don't think it's worth a refactor. Also need to resolve the conflicts before merging.

assets/wizards/dashboard/index.js
thomasguillot: It's not a link so it doesn't make sense to use a `ButtonCard`. It's also temporary and the Google oauth connection will be moved elsewhere.

PR #966, opened 2021-05-18
Replace WP User Avatar with Simple Local Avatars

philipjohn:
Replaces WP User Avatar (recently renamed to ProfilePress and _massively_ altered) with the much simpler Simple Local Avatars plugin by 10up as our recommendation.


claudiulodro (approved):
Installer works great. Thanks @philipjohn!

PR #971, opened 2021-05-19
Update gray colors using latest WordPress base-style

thomasguillot:
### All Submissions:

This PR updates @wordpress/base-styles to 3.4.4 which means variables like `$dark-gray-900` are gone... so I had to replace all the color variables with the new ones.

1. Switch to this branch
2. `npm i && npm rebuild node-sass && npm run build`
3. check all the various wizards (and let me know if you notice any issue!)

### Other information:


assets/components/src/preview-box/index.js
```diff
@@ -10,8 +10,8 @@ import './style.scss';
 
 const PreviewBox = ( { children } ) => (
 	<div className="newspack-preview-box">
-		<span className="newspack-preview-box__label">{ __( 'Preview', 'newspack' ) }</span>
-		<span className="newspack-preview-box__content">{ children }</span>
+		<div className="newspack-preview-box__label">{ __( 'Preview', 'newspack' ) }</div>
```
dkoo, line 13: Don't see this component in the demo page—should we add it?

assets/components/src/popover/style.scss
```diff
@@ -7,13 +7,13 @@
 
 .newspack-popover {
 	.components-popover__content {
-		border: 1px solid $light-gray-500;
+		border: 1px solid $gray-300;
```
dkoo, line 10: Would be nice to add an example of an action card with a popover to the components demo page.

dkoo (changes requested):
Didn't see any issues with the new styles. I noticed a couple of components are missing from the demo page—now might be a good time to add them.

assets/components/src/preview-box/index.js
thomasguillot: It's only used in 1 place, the SEO wizard. I don't think it's worth adding to the demo page.

assets/components/src/popover/style.scss
thomasguillot: Might be worth re-working the ActionCard to allow this kind of pattern since we are using it in multiple places now (Campaigns, Ad Units, ...)

dkoo (approved):
That might be a good idea. The current pattern is to include the popover in the action card's `actionText` prop [as done here](https://github.com/Automattic/newspack-plugin/blob/master/assets/wizards/popups/components/prompt-action-card/index.js#L39-L75), so we could potentially include an example using this pattern now and handle the refactor in a different PR, or we could just wait until the refactor. 

I'll approve for now to unblock this, but if you decide to add the current pattern now just send another review request. Thanks!

thomasguillot:
I've created a new issue for the ActionCard Popover: https://github.com/Automattic/newspack-plugin/issues/977

PR #975, opened 2021-05-21
feat: add multi-select capabillity to AutocompleteWithSuggestions

### LINKED ISSUE #954: Newspack Components: Extend AutocompleteWithSuggestions to handle multi-select
#952 adds a new component that includes an `AutocompleteTokenfield` with an accompanying list of suggestions. This allows selection of a single value from the list or the autocomplete input. A new feature for Newspack Listings needs this component to be able to handle multiple selections at once, like so:

2-new-state

In this use case, the list of suggestions looks and behaves more like a taxonomy select component. We should extend the `AutocompleteWithSuggestions` component to handle this as well if we provide certain props (perhaps `multiSelect={ true }` or something similar). If the required props are passed, then the component would render everything in the above screenshot that appears inside the modal, from the "4 listings selected" message and the optional data type selector (which could let you choose which post type or other data type to query), through the list of checkbox options to select multiple options.

dkoo:
### All Submissions:

This complete the `AutocompleteWithSuggestions` component to add multi-select capability. This is needed for the proposed child listings UI design in Newspack Listings.

Closes #954.

1. Navigate to the components demo page. Confirm there are now two instances of the `AutocompleteWithSuggestions` component: the first is single-select (unchanged from `master`), the second is multi-select with checkboxes instead of links.
2. Test the single-select UI; its behavior should be unchanged from `master`, except that the selected item is now shown at the top of the component separate from the autocomplete field, which is now only used for search functionality:

Screen Shot 2021-05-21 at 11 52 54 AM

3. Test the multi-select UI:
  - Post type selector: This dropdown lets you change which post type is shown in suggestions, both in the autocomplete field and in the suggestions list. (This can actually be used in both single-select and multi-select flavors of the component, but is only shown in multi-select to demonstrate that it's optional.)
  - Multi-select capability: You should now be able to select more than one item. Selected items appear at the top of the component, and are checked if displayed in the suggestions list. Checking an item from the list should add it to the selections, unchecking it should remove it. Typing a post title into the autocomplete field and selecting a suggestion should add the suggestion if it's not already selected, and remove it if it is.
  - "Clear all" feature: If more than one item is selected, a "Clear all" button will appear at the top of the component. Clicking this will remove all selections and return the component to an empty state.

Screen Shot 2021-05-21 at 12 00 08 PM

4. Test load more functionality: For both single-select and multi-select flavors, if there are more posts to suggest than the `suggestionsToFetch` prop (default: 10), a "Load more" button will appear at the end of the suggestions list. Clicking this button will load `suggestionsToFetch` more  posts until all posts of the selected post type are loaded.

### Other information:


adekbadek (approved):
Nice! I like how the default props make it easier to use the component.

dkoo:
@thomasguillot One thing I should mention is that I deviated slightly from your design mockup with the selected items here:

Screen Shot 2021-05-25 at 10 44 55 AM

In your mockup:

Screen Shot 2021-05-25 at 10 43 12 AM

In the implemented version, we group all selections together regardless of their post type. I found that grouping different post types into separate sections was causing the selection area to be very tall if more than one post type is selected, and also caused a lot of screen jitter when post type sections get added or removed.

dkoo:
@adekbadek I've been testing this component with a Newspack Listings feature and doing so has uncovered a few bugs/shortcomings in the component as you reviewed it. Here's a summary in case you want to re-test and/or re-review:

* 6da0a28 - This fixes a bug where if you pass a `fetchSuggestions` function as a prop to the component, it didn't receive info about the currently selected post type. This is important when passing the fetch function as a prop because the selected post type determines what the suggestions endpoint needs to be.
* 8bda922 - This fixes a similar set of bugs when using the autocomplete field with a passed `fetchSuggestions` prop. The `AutocompleteTokenfield` component was not passing the extra param for selected post type.

dkoo:
I've been using the component from this branch in Campaigns, Listings, and Blocks for several days now without issue, so I'm going to go ahead and merge/publish to NPM. If any further bugs or features need to be addressed, we can do that in additional and smaller PRs. Thanks to both of you for your reviews, @thomasguillot and @adekbadek!

PR #870, opened 2021-02-19
Onboarding: Services step; Newsletters wizard

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds the Services step of the onboarding process, as well as a Newsletters wizard. 

### How to test the changes in this Pull Request:

1. Switch `newspack-newsletters` to [`add/update-settings` branch](https://github.com/Automattic/newspack-newsletters/pull/407)
2. Visit the Setup wizard, Services step (`/wp-admin/admin.php?page=newspack-setup-wizard#/services`)
3. Observe four sections, which can be toggled
4. Edit the first section ("Reader Revenue"), observe the settings are reflected in the Reader Revenue Wizard. Edit it there, too, and observe the settings updated when visiting the Services step again
5. Same with the next section ("Newsletters"), which maps to the "Authoring" part of the Engagement wizard
6. Observe that the last two sections (ad related) are mutually exclusive
7. Edit the ad related sections, observe changes reflected in the Advertising wizard
8. Visit the Engagement wizard, change the settings of the Newsletters and observe them being reflected in the setting of the Newspack Newsletters plugin 
9. Deactivate the Newspack Newsletters plugin. Observe the Engagement wizard displays a plugin card instead of the settings interface

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #976, opened 2021-05-21
fix: prevent external links from ad dashboard UI

dkoo:
### All Submissions:

#870 changed the `ActionCard` component so that if it receives an `href` prop, it renders an `ExternalLink` to that `href` value. This resulted in a few Ads dashboard components rendering external links to internal UI pages, which is an annoying admin experience (clicking each link opens a new tab instead of loading the next view in a single-page app experience as expected). This change invokes the `Router` to turn those links into internal routes instead of external links.

1. In Newspack > Advertising, confirm that the Google Ad Manager action card no longer has the external link icon next to the "Configure" link:
 
Screen Shot 2021-05-21 at 2 53 48 PM

2. Confirm that clicking on either "Configure" or the "Google Ad Manager" title loads the GAM view with the list of ad units directly in the same tab, without refreshing the page.
3. In the Ads Units list, confirm that the ad unit title and the "Edit" link load the edit page for that ad unit in the same browser tab, without refreshing the page, and that the "Edit" link no longer has an external link icon.

### Other information:

### DIFF
--- a/assets/wizards/advertising/views/ad-units/index.js
+++ b/assets/wizards/advertising/views/ad-units/index.js
@@ -10,7 +10,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import { ActionCard, Router, withWizardScreen } from '../../../../components/src';
+import { ActionCard, withWizardScreen } from '../../../../components/src';
 
 /**
  * Advertising management screen.
--- a/assets/wizards/advertising/views/ad-units/index.js
+++ b/assets/wizards/advertising/views/ad-units/index.js
@@ -5,46 +5,46 @@
 /**
  * WordPress dependencies
  */
-import { Component, Fragment } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
-import { ActionCard, withWizardScreen } from '../../../../components/src';
+import { ActionCard, Router, withWizardScreen } from '../../../../components/src';
+
+/**
+ * Router component for managing single-page app nav.
+ */
+const { useHistory } = Router;
 
 /**
  * Advertising management screen.
  */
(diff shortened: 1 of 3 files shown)


assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -5,46 +5,46 @@
...
+				) }
+			</p>
+			{ Object.values( adUnits ).map( ( { id, name } ) => {
+				return (
+					<ActionCard
+						key={ id }
+						title={ name }
+						actionText={ __( 'Edit' ) }
+						titleLink={ `#${ service }/${ id }` }
+						onClick={ () => history.push( `${ service }/${ id }` ) }
```
adekbadek, line 40: Sidenote: a slightly different solution [is proposed here](https://github.com/Automattic/newspack-plugin/pull/936/files#diff-ebe9181a891d66ff9903af1ddc0652fcc33a11000450357f28e5a62b35410a44R73), where any internal links (relative links) are treated as such (not rendering an `ExternalLink` component).

adekbadek (approved):
Nice, thanks for the rewrite to function components!

assets/wizards/advertising/views/ad-units/index.js
dkoo: Oh, that's a good solution, too, and maybe a bit more DRY-esque. Should we merge this PR and move to your solution in #936, or do you want me to incorporate that here?

adekbadek: Up to you, the result will be the same and the difference is small.

dkoo: I implemented in a9d47e5—I think it's a cleaner solution. Can you please re-test once more?

adekbadek (approved):
🙌

PR #990, opened 2021-06-04
feat: allow some JS on AMP pages

adekbadek:
### All Submissions:

A different approach to "AMP Plus" than [forcing dev mode](https://github.com/Automattic/newspack-ads/pull/81) (#688). Here `amp_validation_error_sanitized` filters (h/t @westonruter) are used to allow some scripts to pass. When forcing dev mode, the `amp` attribute was not removed, so the site [would get Search Console errors about invalid AMP](https://github.com/Automattic/newspack-plugin/pull/688#issuecomment-719030677). With the approach in this PR, I see the `amp` attribute is removed automatically - I guess that's what AMP Plugin does when it serves invalid AMP.

Instructions in https://github.com/Automattic/newspack-ads/pull/132

### Other information:


westonruter:
> With the approach in this PR, I see the `amp` attribute is removed automatically - I guess that's what AMP Plugin does when it serves invalid AMP.

That's correct. When in Standard mode, the AMP plugin does not remove excessive CSS by default. Such invalid markup is not removed because it can cause broken styling, with no alternative non-AMP version to redirect to. So in the case of any validation error whose invalid markup is “kept”, the AMP plugin omits the `amp` attribute so that Search Console doesn't complain about invalid AMP. In this way, you'll get pages that are managed by the AMP runtime, so you'll get all those speed benefits, but you won't get prerendering or serving from the AMP cache. Since you already have good caching on origin this is less important for Newspack sites.

You can see the underlying logic here: https://github.com/ampproject/amp-wp/blob/ec3409f79315c830fb6df95bd3e25d1342e57ec5/includes/validation/class-amp-validation-manager.php#L1537-L1569

As part of the Bento AMP effort we're going to be expanding on this to make make this more of a blessed path, including having a UI to opt into a sandboxing mode where such invalid markup is not removed by default.

westonruter:
You might want to consider only adding the filters if `amp_is_canonical()`, that is if in Standard mode. In Transitional/Reader modes, AMP pages with kept invalid markup get redirected to the non-AMP version. This is probably not what you'd want.

includes/class-amp-enhancements.php
```diff
@@ -47,16 +59,16 @@ public static function should_use_amp_plus( $context = null ) {
...
-	 * @param array $sanitizers The array of sanitizers, 'MyClassName' => [] // array of constructor params for class.
+	 * @param bool|null $is_sanitized If null, the error will be handled. If false, rejected.
+	 * @param object    $error The AMP sanitisation error.
 	 */
-	public static function amp_content_sanitizers( $sanitizers ) {
-		if ( self::should_use_amp_plus( 'gam' ) ) {
-			require_once NEWSPACK_ABSPATH . 'includes/amp-sanitizers/class-amp-sanitizer-gam.php';
-			$sanitizers['AMP_Sanitizer_GAM'] = [];
+	public static function amp_validation_error_sanitized( $is_sanitized, $error ) {
+		if ( isset( $error['node_attributes'], $error['node_attributes']['amp-plus-allowed'] ) ) {
```
westonruter, line 77: To avoid extra validation errors from being emitted by the presence of this attribute, you may want to use a `data-` attribute instead.

adekbadek: Done in 3a1ee29bd5b6198c05e0282ffd7a69af681fc770 🙌

adekbadek:
> You might want to consider only adding the filters if amp_is_canonical(), that is if in Standard mode. In Transitional/Reader modes, AMP pages with kept invalid markup get redirected to the non-AMP version. This is probably not what you'd want.

Thanks! Done in 9cd196515d3bc8a6aec73df8329240b134e5e5cb

westonruter:
Another note: the validated URL screen currently isn't super well tailored for invalid AMP markup that is forcibly marked as kept. The UI will appear show that you can change the status from kept to removed, but after saving the change doesn't stick. As a quick fix for this I filed https://github.com/ampproject/amp-wp/issues/5389, but we need to do a more comprehensive revamp of the invalid markup statuses in light of Bento.

dkoo (approved):
The example in https://github.com/Automattic/newspack-ads/pull/132 works great with a `<script>` tag added directly to the document head. Noting here that it also may be possible to allow scripts enqueued via `wp_enqueue_script` by filtering the output of [`script_loader_tag`](https://developer.wordpress.org/reference/hooks/script_loader_tag/).

includes/class-amp-enhancements.php
```diff
@@ -29,34 +29,57 @@ function ( $gtag_opt ) {
...
 	 *
-	 * @param array $sanitizers The array of sanitizers, 'MyClassName' => [] // array of constructor params for class.
+	 * @return bool Is AMP plus mode configured.
 	 */
-	public static function amp_content_sanitizers( $sanitizers ) {
-		if ( self::should_use_amp_plus( 'gam' ) ) {
-			require_once NEWSPACK_ABSPATH . 'includes/amp-sanitizers/class-amp-sanitizer-gam.php';
-			$sanitizers['AMP_Sanitizer_GAM'] = [];
+	public static function is_amp_plus_configured() {
+		return defined( 'NEWSPACK_AMP_PLUS_ENABLED' );
```
dkoo, line 65: Not a blocker, but why not enforce a `true` value for this as with our other environment flags?

adekbadek: This should be `false` by default, though. It's still experimental until Google allows invalid AMP in the search carousel.

adekbadek: Good call, this way it can be set to `false` too. Done in 8cf319bed320d5811e4e88dc1c3133410cedf061

westonruter:
While not merged yet, you may want to check out this PR which adds support for “sandboxing levels” and facilities for allowing non-AMP markup: https://github.com/ampproject/amp-wp/pull/6546.

adekbadek:
Interesting, thanks! We'll keep an eye on that.

PR #999, opened 2021-06-09
feat(design-wizard): allow custom font importing

adekbadek:
### All Submissions:

Adds custom font importing feature to onboarding & design wizard.

1. In the design step onboarding of and the design wizard, 
2. Observe the "Typography" section allows to choose a "Custom fonts" option
3. Select the option, insert font URL, names, and fallback fonts, save
4. Observe the changes applied to the site, and viewable in the Customizer
5. Change back to "Curated list" and select from the curated fonts
4. Observe the changes applied to the site, and viewable in the Customizer

### Other information:


assets/wizards/site-design/views/main/index.js
```diff
@@ -118,32 +177,46 @@ const Main = ( {
 				title={ __( 'Typography', 'newspack' ) }
 				description={ __( 'Pick the font pairing to use throughout your site', 'newspack' ) }
 			/>
+			<RadioControl
```
dkoo, line 180: This may not be related to these changes, but it's the first time I'm noticing. The "selected" radio button style looks off to me (using WP 5.7.2 without the Gutenberg plugin, tested with Chrome and Firefox and replicated in local dev site and on a JN site):

Screen Shot 2021-06-09 at 4 36 49 PM

dkoo: I might also suggest a slightly different UI: instead of using two radio buttons, why not a single `ToggleControl`, e.g.

Screen Shot 2021-06-09 at 4 43 00 PM

dkoo: Also, if switching from custom to default/curated, it might be worth noting that your custom import code or URL will be lost after saving.

dkoo (changes requested):
Nice addition to the design wizard! I have a couple of small UI suggestions, and I ran into a strange style bug with the radio buttons.

assets/wizards/site-design/views/main/index.js
adekbadek: > The "selected" radio button style looks off to me 

Yes - I forgot to mention that. Could you take a look @thomasguillot ?

> I might also suggest a slightly different UI: instead of using two radio buttons, why not a single ToggleControl, e.g.

I second that - WDYT Thomas?

thomasguillot: > The "selected" radio button style looks off to me

Hmm yeah weird... fixed https://github.com/Automattic/newspack-plugin/pull/999/commits/3e7f0d3cd22336a04e145d74a40d3de6c6ec4e3c

thomasguillot: > I might also suggest a slightly different UI: instead of using two radio buttons, why not a single ToggleControl, e.g.

I used radio buttons on purpose as it felt more balanced visually. I also wanted to limit the user’s choice to one option whilst displaying both options: you can choose our curated list OR import your own custom fonts.

Screenshot 2021-06-10 at 15 37 03
Screenshot 2021-06-10 at 15 38 02

thomasguillot: >  if switching from custom to default/curated, it might be worth noting that your custom import code or URL will be lost after saving

I wouldn't worry too much about this. This setting tends to be set once, and forget about it. And, if a publication has its own custom fonts, it's unlikely they will play with the curated list.

thomasguillot (approved):

assets/wizards/site-design/views/main/index.js
dkoo: > I used radio buttons on purpose as it felt more balanced visually. I also wanted to limit the user’s choice to one option whilst displaying both options: you can choose our curated list OR import your own custom fonts.

Ok, I'll defer to you on this. One thing, though, do you think we can change the label "Curated list" to "Default fonts" instead? I think that would be more descriptive and it mirrors "Custom fonts" a bit more.

dkoo: > Hmm yeah weird... fixed 3e7f0d3

Confirmed, thanks!

thomasguillot: > Ok, I'll defer to you on this. One thing, though, do you think we can change the label "Curated list" to "Default fonts" instead? I think that would be more descriptive and it mirrors "Custom fonts" a bit more. (I also worry that the term "curated list" might be confused with the same term as used in Newspack Listings, for a completely different purpose!)

Done in https://github.com/Automattic/newspack-plugin/pull/999/commits/da426b542be4f3571d08ab50725fffa9c4458933

I removed "fonts" from the label since it felt redundant

dkoo: Looks better, thanks!

dkoo (approved):

Issue #861, opened 2021-02-11
[Campaigns] Add New Prompt UI


dkoo:
Some ideas I had about the experience of adding new prompts while testing #860. @thomasguillot, feel free to riff on this or close the issue if you think these are unnecessary.

* Maybe this is just my bad memory, but sometimes when I click on "Add New Prompt" I forget a moment later which segment I'm adding to. Could we add some info to this modal that specifies that we're adding a prompt to this specific segment in this specific campaign, maybe under the "Add New Prompt" title?

<img width="398" alt="Screen Shot 2021-02-11 at 2 34 22 PM" src="https://user-images.githubusercontent.com/2230142/107701775-995ece00-6c76-11eb-841f-fdbd60873478.png">

* This modal might also be a good place to add a dropdown to choose a campaign if you're adding from the Active Prompts view. And if you don't select a campaign first, the buttons are disabled—this way we can discourage creating new Unassigned prompts.

* Totally subjective, but in the Add New Prompt modal, the line spacing looks a little too loose to my eyes.

* Also totally subjective, but could the icon buttons and their labels be more visually integrated, perhaps by adopting a wider modal and a horizontally-oriented button? These could collapse to a single column for smaller viewports. We might also want to consider adding some short descriptions of what each placement is to help new users. e.g. something like this, but better-looking 😄 

<img width="613" alt="Screen Shot 2021-02-11 at 3 20 41 PM" src="https://user-images.githubusercontent.com/2230142/107705911-bdbda900-6c7c-11eb-8e72-27d674d8353e.png">

dkoo:
A follow-up. I had a convo with @iamstevebeatty about Campaigns today, and he has an alternative selection which I will present here for further discussion: rather than showing a single modal that only lets you choose the prompt's placement, what if the modal was more of a guided experience that makes you choose the most important settings for a new prompt? This would include:

1. "Where do you want the prompt to appear?" ➜ Placement dropdown, defaulting to "inline"
2. "How often do you want the prompt to appear?" ➜ Frequency dropdown, defaulting to "until dismissed" for inline/"once a day" for overlays
3. "Should the prompt be shown with specific categories?" ➜ Category filtering autocomplete, defaulting to none

Then, once the user clicks "OK", they get forwarded to the editor page as usual. BUT, if the options selected would result in a conflict with an existing active prompt, we could show a warning at this point before letting the user proceed, e.g. "This prompt will conflict with the following active prompt(s): [show a list]"

[...]

PR #1013, opened 2021-06-21
Update the style of the Modal component

thomasguillot:
### All Submissions:

I noticed the `Modal` felt slightly off, especially with the header and the various paddings. This PR restyle it to use a 64px padding around it. It also moves the title more inline with the content. For buttons, we are now using a `buttonsCard`.

I had to reorganise things in the Campaigns to make sure the changes were not breaking the "Add New Prompt" modal. And as an added bonus, I've tweaked the "Add New Prompt" modal to something similar to what was suggested in #861 

1. Check the various Modals (Components Demo, Campaigns, Analytics -- Custom Events, etc...)
2. Switch to this branch
3. Test again

### Other information:

### DIFF
--- a/assets/components/src/modal/style.scss
+++ b/assets/components/src/modal/style.scss
@@ -16,7 +16,7 @@
 	width: 100%;
 
 	@media screen and ( min-width: 600px ) {
-		border-radius: 4px;
+		border-radius: 3px;
 		max-width: 80%;
 		max-width: calc( 100% - 64px );
 	}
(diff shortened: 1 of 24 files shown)


assets/components/src/modal/style.scss
```diff
@@ -16,7 +16,7 @@
 	width: 100%;
```
dkoo, line 16: Adding `max-height: 100%` ( or `100vh`) lets the modal content scroll if it exceeds the height of the viewport:

```suggestion
	width: 100%;
	max-height: 100%;
```

dkoo (changes requested):
The new spacing looks good in Campaigns, components demo, and Analytics. However I noticed on **Analytics > Custom Events** the modal content exceeds the viewport size at smaller viewports. Tested the below and it seems to work.

Also, we'll want to publish this to the NPM package, right? I would love to figure out why the `@wordpress/icons` package [crashes](https://github.com/Automattic/newspack-popups/issues/556) when used in the NPM package before we publish the next version, though.

assets/components/src/modal/style.scss
thomasguillot: Fixed in https://github.com/Automattic/newspack-plugin/pull/1013/commits/a543fba520a2d7eaa83d76145f4de75a00e3dd5a

dkoo (changes requested):
Oh, I'm a fan of this new "Add New Prompt" modal. It does occur to me that we have one additional placement, though—Manual Only. Is it a concern that this isn't one of the options in the modal?

Screen Shot 2021-07-02 at 3 28 20 PM

I suppose it's already missing from the existing modal and having 7 options doesn't lend itself well to a neat grid, plus we don't necessarily want to encourage users to create a lot of manual-only prompts, so I'm okay with omitting it. Marking "request changes" so we make sure to discuss, but if you don't think it's necessary then I'll approve.

thomasguillot:
@dkoo I think we should add the "Manual Only" option. We should display all the available options. That being said, I'd suggest to do it in a separte PR since I will need to reorganise the Grid/Modal and I need to think about it 😊

dkoo (approved):

PR #1042, opened 2021-07-15
Fix fatal error if Newspack reader revenue is selected but WC is deactivated

### LINKED ISSUE #1011: Fatal Error when setting up Reader Revenue on a new site
**Describe the bug**
During setup for Reader Revenue, after selecting `Newspack`, a fatal error occurred, preventing any further steps from being taken or being able to edit anything in the Newspack wizard.

Screen Shot 2021-06-16 at 2 27 19 PM

**To Reproduce**
1. Created a new site with Knife
2. Enabled the Newspack debug constant in wp-config.php
3. Went through the onboarding flow for the site -- note that it asked for Stripe credentials, and I realized at that point that I only had a production public key for the publisher, and no secret key, so I proceeded without providing Stripe credentials.
4. Once I completed the initial pass through the onboarding, I clicked on the option for Reader Revenue. Once I selected Newspack from the drop-down, the error shown in the screenshot above popped up.
5. After going back to the Dashboard, every option on the Newspack Dashboard resulted in the same error.
6. Through the browser CLI, @tmmbecker checked the PHP errors and saw the following:

```
PHP Fatal error:  Uncaught Error: Class ‘WC_Payment_Gateways’ not found in /wordpress/plugins/newspack-plugin/1.45.0/includes/configuration_managers/class-woocommerce-configuration-manager.php:104
```

Based on this error, we realized that WooCommerce wasn't yet installed. Once I installed WooCommerce, the error went away.

_Additionally_, once WooCommerce was installed, I was able to set suggested donation amounts, but even after saving, they seemed to disappear. I realized this was because the additional premium plugin extensions (Name Your Price and Subscriptions) weren't installed either. 

**Expected Behavior**

Once choosing Newspack, the required plugin(s) would be installed automatically rather than generating an error message or silently failing to save.

claudiulodro:
### All Submissions:

This PR fixes an issue I ran into: on an existing site, if reader revenue is selected but WC is deactivated, there is a fatal error thrown:

Screen Shot 2021-07-15 at 8 56 00 AM
Screen Shot 2021-07-15 at 8 51 42 AM

Closes #1011 

1. Set up reader revenue using the Newspack reader revenue style
2. Deactivate WC. On the Reader Revenue wizard you should get the Unrecoverable Error modal.
3. Apply this patch. The Reader Revenue wizard should load.

### Other information:

### DIFF
--- a/includes/class-stripe-connection.php
+++ b/includes/class-stripe-connection.php
@@ -36,7 +36,6 @@ class Stripe_Connection {
 	 * Get Stripe data, either from WC, or saved in options table.
 	 */
 	public static function get_stripe_data() {
-		$stripe_data              = self::get_default_stripe_data();
 		$wc_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'woocommerce' );
 
 		if ( Donations::is_platform_wc() && $wc_configuration_manager->is_configured() ) {
--- a/includes/class-stripe-connection.php
+++ b/includes/class-stripe-connection.php
@@ -36,11 +36,12 @@ class Stripe_Connection {
 	 * Get Stripe data, either from WC, or saved in options table.
 	 */
 	public static function get_stripe_data() {
-		$stripe_data = self::get_default_stripe_data();
-		if ( Donations::is_platform_wc() ) {
+		$stripe_data              = self::get_default_stripe_data();
+		$wc_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'woocommerce' );
+
+		if ( Donations::is_platform_wc() && $wc_configuration_manager->is_configured() ) {
 			// If WC is configured, get Stripe data from WC.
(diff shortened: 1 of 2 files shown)


adekbadek:
@claudiulodro I can't reproduce, but if you can - please check if #1046 doesn't fix #1011. We might've doubled our efforts :)

claudiulodro:
It looks like a combination of #1046 and turning Newspack debug off solves the issue 👍 Thanks for clarifying! I hadn't realized we'd changed things so that Newspack debug bypasses plugin installers.

adekbadek:
> I hadn't realized we'd changed things so that Newspack debug bypasses plugin installers.

The reason is that when you're troubleshooting, you might upload a e.g. `newspack-plugin-locally-built` or something plugin and turn it on. Without that bypass, the wizards would bail since `newspack-plugin` would not be active.

PR #1052, opened 2021-07-19
feat/campaigns mgmt

### LINKED ISSUE #784: Option to view and restore deleted Campaigns
Currently, if you delete a campaign, it presumably gets sent to trash. But since we're hiding the regular WP posts list UI for campaigns, there's no way to see trashed campaigns to restore them. Should we consider adding a filter to the Campaigns wizard UI to show trashed campaigns, with the option to restore them to draft status?

cc @jeffersonrabb and @thomasguillot

### LINKED ISSUE #869: Prompt deletion & trashes, all prompts views
Slightly related issues:

1. **Prompt deletion**. As is: prompt deletion in the Campaigns Wizard offers no confirmation and omits trash. Should be: deletion should allow the user to confirm, and the deleted prompt should be moved to trash instead of deleted permanently.
2. **Trash view**. After the above is done, a "Trash" filter should be added to the view selection (same as https://github.com/Automattic/newspack-plugin/issues/786, not closing yet cause that might be too much for one PR).
3. **All prompts view**. An "All Prompts" view should be added tot the view selector.

adekbadek:
### All Submissions:

Closes #784
Closes #869 

1. Switch `newspack-popups` to `feat/trash-retrieval` branch (https://github.com/Automattic/newspack-popups/pull/580)
1. In the Campaigns wizard, observe two new sections - "All Prompts" and "Trash"
2. Observe that "All Prompts" displays all (untrashed) prompts, regardless of status or campaign
3. Delete a prompt via the popover handler - observe it moved to "Trash"
4. Restore the prompt, observe it's restored
5. Delete a prompt and then "Delete permanently" - observe that it's gone forever 

### Other information:

### DIFF
--- a/includes/wizards/class-popups-wizard.php
+++ b/includes/wizards/class-popups-wizard.php
@@ -114,6 +114,20 @@ class Popups_Wizard extends Wizard {
 		);
 		register_rest_route(
 			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/(?P<id>\d+)/restore',
+			[
+				'methods'             => \WP_REST_Server::EDITABLE,
+				'callback'            => [ $this, 'api_restore_popup' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+				'args'                => [
+					'id' => [
+						'sanitize_callback' => 'absint',
+					],
+				],
+			]
+		);
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
 			'/wizard/' . $this->slug . '/(?P<id>\d+)/duplicate',
 			[
 				'methods'             => \WP_REST_Server::READABLE,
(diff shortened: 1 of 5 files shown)


includes/wizards/class-popups-wizard.php
```diff
@@ -523,7 +537,7 @@ function( $prompt ) {
 					$prompt['edit_link'] = get_edit_post_link( $prompt['id'] );
 					return $prompt;
 				},
-				$newspack_popups_configuration_manager->get_prompts( true )
+				$newspack_popups_configuration_manager->get_prompts( true, true )
```
miguelpeixe, line 540: Non-blocking nit-picking: maybe it's time this becomes a configuration object?

miguelpeixe (approved):
Works as described! Left a non-blocking suggestion, you can merge if you judge it unnecessary.

includes/wizards/class-popups-wizard.php
adekbadek: I've found it cumbersome in PHP to use this pattern - too much boilerplatey checking for key existence etc. But feel free to propose a change - it's likely I'm not thinking about the most elegant implementation of it.

PR #1058, opened 2021-07-21
feat(amp): skip AMP on WC pages

### LINKED ISSUE #967: Newspack should turn off AMP on WooCommerce pages by default
The bug:
By default, when installing Newspack and installing WooCommerce and the recommended WooCommerce Stripe Gateway, payments won't work on WooCommerce. WooCommerce will display an error message that the client's browser does not support JavaScript, and the credit card number field will not accept input. The fix is to edit the affected page and turn off AMP on that page. (Unknown whether this affects all of Newspack or just sites using Newspack Katharine, but it affects the WooCommerce "Storefront" theme, too, so my guess is it affects them all.)

To reproduce:

1. Install Newspack, choose the Katharine child theme, plus install WooCommerce and the WooCommerce Stripe Gateway, create a Stripe account, etc.
2. Create a product.
3. Attempt to buy the product and observe that when you get to the Checkout page, you won't be able to make a payment. Observe also that the Checkout page says your browser doesn't support JavaScript or that it is turned off.
4. Edit the Checkout page and turn AMP off. Update the page.
5. Reload the Checkout page and observe that now you can enter your credit card number.

Screenshot:
payment wtf

Expected behavior:
Since NewsPack recommended that I install these plugins and I installed them through the Newspack plugin, I would expect NewsPack to edit the newly created WooCommerce pages for me upon installation to turn off AMP.
(decision tree coming!)
Or if there is some other setting that needs to be set or some other plugin that should be installed for this to just work with AMP turned on, I would expect Newspack to set or recommend installing that. (And if it requires a plugin and I choose not to install that plugin, Newspack should recommend I turn off AMP on WooCommerce pages)

Affected website:
https://outlookbythebay.com (click Subscribe if you want to test it, but I have already turned off AMP on affected pages so there's nothing to see there)

Other notes:
My notes are not as detailed about how pages other than the checkout page are affected, but I turned it off on another page (perhaps the cart?) when I saw an unhelpful error message that said something about contacting the website owner to try to make the error messages more helpful. There was a link in that error message which went to something about getting scripts to work better with the AMP plugin, so I turned it off on that page, too.

adekbadek:
### All Submissions:

Closes #967.

1. On `master`,
1. Ensure your Reader Revenue platform is Newspack
2. Ensure your WC checkout page has AMP **enabled** in the editor sidebar
2. Try to make a donation, observe the "Payment Info" section not interactive
3. Switch to this branch, observe the "Payment Info" section is interactive

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -46,6 +46,13 @@ class Donations {
 	];
 
 	/**
+	 * Cached status of the current request - is it a WC page.
+	 *
+	 * @var string
+	 */
+	private static $is_wc_page = null;
+
+	/**
 	 * Initialize hooks/filters/etc.
 	 */
 	public static function init() {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
🙌

Issue #630, opened 2020-08-12
Rethink Syndication Wizard and add support for Newspack RSS Enhancements


thomasguillot:
**Is your feature request related to a problem? Please describe.**
At the moment, the Syndication Wizard forces users to install Apple News and Facebook Instant Article. Should it be optional instead? We could always have the toggles turned on by default.
We could also add support here for [Newspack RSS Enhancements](https://github.com/Automattic/newspack-rss-enhancements).

**Screenshot**

![Screenshot 2020-08-12 at 15 28 21](https://user-images.githubusercontent.com/177929/90027996-0e921c00-dcb1-11ea-84ce-43f16306fb6d.png)

PR #1068, opened 2021-07-27
Add RSS Enhancement to Syndication Wizard

thomasguillot:
### All Submissions:

Adds Newspack RSS Enhancement to the list of plugins. It also makes Apple News and Facebook Instant Articles optional. Plus a small fix on the Plugin Installer.

Closes https://github.com/Automattic/newspack-plugin/issues/630

1. Go to Newspack > Syndication
2. Switch to this branch
3. Enable/Disable the various plugins

### Other information:


adekbadek (approved):

dkoo:
I was testing this at the same time as @adekbadek and it works as expected, but one thing I noticed (not sure if it's true in `master` too) is that if you toggle multiple switches at the same time, the UI goes a bit haywire. Wonder if we should disable other plugin cards on the page if any one is in a loading state? https://d.pr/i/GakKLB

adekbadek:
> one thing I noticed (not sure if it's true in master too) is that if you toggle multiple switches at the same time, the UI goes a bit haywire

That's true, should be fixed by 5e9e083036dd614d5122342df63161c6560f7857

PR #1094, opened 2021-08-11
Update PWA config to use defaults

claudiulodro:
### All Submissions:

This PR updates our PWA config to work better with the latest versions of the PWA plugin. Previously, we were throwing a lot of these errors/deprecation notices:

`
WP_Service_Worker_Caching_Routes::register was called <strong>incorrectly</strong>. The plugins configuration key is obsolete. Define Workbox plugin configuration at the top level
`

The latest version of PWA provides [a default offline browsing config out-of-the-box](https://github.com/GoogleChromeLabs/pwa-wp/wiki/Service-Worker#offline-browsing), and it's the same config we had in here, so I've just removed our config, and we'll use the default config. 

In the new version of PWA, offline browsing can be enabled/disabled using a checkbox on the Reading settings screen. So that people don't have to go into their settings and check the "Enable offline browsing" checkbox after our next Newspack release, I've added a filter that forces offline browsing to always be enabled. If there's a reason somebody wouldn't want offline browsing enabled, let me know, and we can revise that, but I couldn't think of any! :)

1. For PWA to work, you'll need a site with a valid `https` connection.
2. Open Chrome dev console. Go to Application > Service Workers. Unregister the existing service worker (top right corner in the screenshot)

Screen Shot 2021-08-11 at 10 06 36 AM

3. Visit a post. Either check the Offline checkbox in Application > Service Workers (top-middle in screenshot below) or turn your internet off. Refresh the page so you view the post with your internet turned off. You should see the post served. If you visit the homepage, you should see an Offline page.

Screen Shot 2021-08-11 at 10 07 49 AM

### Other information:

### DIFF
--- a/includes/class-pwa.php
+++ b/includes/class-pwa.php
@@ -7,7 +7,7 @@
 
 namespace Newspack;
 
-use \WP_Error, \WP_Service_Worker_Caching_Routes;
+use \WP_Error;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
🙌

PR #1095, opened 2021-08-12
Donations - Stripe as platform & refactor

adekbadek:
### All Submissions:

A change in architecture. Instead of treating direct Stripe donations as a possibility within NRH platform, Stripe is here set up as a separate platform. In effect, there are now three platforms:

1. WooCommerce ("Newspack" platform in Reader Revenue wizard) - may use multiple payment gateways
2. News Revenue Hub - redirects to NRH Checkout page
3. Stripe - which activates what was before "streamlined mode" of Newspack Blocks' Donate block. The donation happens on-site with minimal friction.

1. Switch `newspack-blocks` to `feat/donations-without-nrh` branch
2. Set up your site with all three platforms, one by one, and test the donation flow. 

### Other information:


dkoo (approved):
Tested and works as described. One thing that threw me off slightly as a tester is that if you have WooCommerce and the Stripe Gateway installed and configured with Stripe credentials, then set your Reader Revenue platform to Stripe and update credentials in Newspack, it's possible for these settings to get out of sync. Then when switch the platform to Newspack, they're synced. It's the difference in behavior that threw me off, but on reflection it does make sense since setting your platform to Newspack uses the WooCommerce Stripe implementation, where as using the Stripe platform does not.

adekbadek:
> One thing that threw me off slightly 

Yeah, it's because both platforms use Stripe, but for ease of migration to this new platform the settings will be cribbed from the last configuration, if they exist.

PR #1107, opened 2021-08-24
Increase PWA network timeout limit

claudiulodro:
### All Submissions:

This PR fixes an issue that appeared after https://github.com/Automattic/newspack-plugin/pull/1094: for a lot of sites, people were not getting the latest homepage when visiting the site. 

This is because the default PWA config has a 2-second network timeout, where if a page takes longer than 2 seconds to load, PWA will serve the offline-cached page. This default is there as encouragement for sites to make their pages fast, but it seems too heavy-handed for our users. :) This PR increases the timeout to 44 seconds, which is one second before the PHP process will timeout on our infrastructure.

h/t @westonruter 🙌 

1. You'll need a site with valid HTTPS to test this PR, otherwise PWA won't work.
2. Add this code snippet somewhere:

```
// Wait 5 seconds.
add_action( 'init', function(){ sleep(5); } );
```

3. Before applying this patch, visit the homepage to prime the offline cache then create a new post. Observe you get served a cached version of the homepage after 2 seconds, and the new post doesn't appear.
4. Apply this patch. Unregister your existing service worker under Chrome dev console > Application > Unregister (top-right)

Screen Shot 2021-08-24 at 11 12 06 AM

5. Visit the homepage to prime the offline cache. Create another new post. Visit the homepage again. Observe you get served an updated homepage with the new post after ~5 seconds.

### Other information:

### DIFF
--- a/includes/class-pwa.php
+++ b/includes/class-pwa.php
@@ -22,6 +22,8 @@ class PWA {
 	public static function init() {
 		// For backwards compatibility and because there's no reason not to, always and automatically enable offline browsing.
 		add_filter( 'pre_option_offline_browsing', '__return_true' );
+
+		add_filter( 'wp_service_worker_navigation_caching', [ __CLASS__, 'increase_network_timeout' ] );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Working as described! 🙌

PR #1119, opened 2021-08-31
feat(stripe): handle subscriber status in campaigns data update

adekbadek:
### All Submissions:

When a Stripe webhook comes in, the data sent to Campaigns plugin should also contain the subscription status.

Note: to test locally, you might have to bypass SSL check for the Campaigns API:

```php
add_filter('http_request_args', function($r) {
	$r['sslverify'] = false;
	return $r;
});
```

1. Create two Campaings segments, targeting donors and subscribers 
2. Create two prompts, set to each segment
3. Configure Reader Revenue to use Stripe as platform, and use AMP Plus mode (`define( 'NEWSPACK_AMP_PLUS_ENABLED', true );`)
4. Configure Newsletters plugin with either Mailchimp or Campaign Monitor as the ESP
5. Choose a list to allow the donors to subscribe to in Reader Revenue wizard
3. Add a Donate block somewhere
4. Ensure the site will receive Stripe webhooks (use `ngrok` or test on a publicly available site)
4. Make a donation while opting in to newsletter
5. Observe both prompts are displayed after navigating to another page (after a moment, so that webhooks comes through) - which means the user is identified as both donor and subscriber

### Other information:

### DIFF
--- a/includes/class-stripe-connection.php
+++ b/includes/class-stripe-connection.php
@@ -200,30 +200,12 @@ class Stripe_Connection {
 					}
 				}
 
-				// Update data in Campaigns plugin.
-				if ( isset( $customer['metadata']['clientId'] ) && class_exists( 'Newspack_Popups_Segmentation' ) ) {
-					$client_id = $customer['metadata']['clientId'];
-					if ( ! empty( $client_id ) ) {
-						$donation_data = [
-							'stripe_id'          => $payment['id'],
-							'stripe_customer_id' => $customer['id'],
-							'date'               => $payment['created'],
-							'amount'             => $amount_normalised,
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described except for:

> Observe both prompts are displayed after navigating to another page (after a moment, so that webhooks comes through) - which means the user is identified as both donor and subscriber

Due to [non-cascading segment matching](https://newspack.pub/newspack-campaigns-breaking-changes-v1-26-0/#non-cascading-segment-matching), you will only see prompts from the best-priority segment you match as a reader. I confirmed that when my Donors segment was the best-priority segment, I saw its prompt, and vice versa with the Subscribers segment.

PR #1135, opened 2021-09-10
Update WordPress dependencies

noahtallen:
### All Submissions:

Over at wp-calypso, we're updating to React 17. This means getting our peer dependencies updated to the right versions, and it turns out that `newspack-components` is a peer dep of `newspack-blocks`, which we sync to WordPress.com via Calypso.

This PR updates to React 17 and to the latest WordPress component packages. 

Also updates most NPM packages to the latest compatible versions.

Also specifies Node version support as `v16.11.1`, matching Calypso's Node support.

Also enables Dependabot to keep NPM and Composer packages up-to-date.

**Bonus:** Also fixes a minor pet peeve of mine, which is that this repo's `npm run release:archive` script previously built the plugin archive ZIP to the `assets/release` directory, whereas every other Newspack plugin repo builds to `./release` (in the root). Now this repo should also build to the root `./release` directory.

1. Using [NVM](https://github.com/nvm-sh/nvm), set your current Node version to v16.11.1 (matching Calypso's supported version)
2. Ensure `npm ci` runs
3. Ensure all NPM build, lint and test scripts still run (`build`, `dev`, `lint`, `test`)
4. Test newspack components (see Components Demo page)
5. Test newspack plugin (especially wizard pages)

### Possible issues/follow-up

- [x] ~Dependency `react-daterange-picker` is stale and doesn't appear to be updated anymore. It doesn't list react 17 support, but I suppose it probably supports it anyways.~ The package still works in the Campaigns Analytics page, so I don't think we need to update this immediately.
- [x] ~Build doesn't work (possible incompatibility with the node version I'm using)~ Should now work with Node v16.11.1

### Other information:


dkoo:
Looping in the @Automattic/newspack-product team for additional testing and feedback. I was able to run `npm ci && npm run build` locally using Node v15.6.0. Confirming that Node v12.* is unable to install dependencies. So this change would require us to update our Node/NPM versions for dev as well as dependencies all repos to match. This is not a bad thing and is probably overdue, but I'm not sure what our timeline for being able to do this is.

As for the CI "build" job failing, we probably need to update something in the CircleCI config to use Node v15+. cc @adekbadek as he knows most about our CI config.

dkoo:
@thomasguillot regarding this:

> Dependency `react-daterange-picker` is stale and doesn't appear to be updated anymore. It doesn't list react 17 support, but I suppose it probably supports it anyways.

It seems to me like this package is only used in the `DateRangePicker` Newspack component, but I'm not seeing this component being used anywhere. Can we deprecate it entirely? Or at least replace the third-party package with the [core `DateTime` component](https://developer.wordpress.org/block-editor/reference-guides/components/date-time/) instead?

noahtallen:
> I also still think it might be possible to use the core DateTime component using the isDayHighlighted prop.

Yeah, it's always nice to use core components when we can!

noahtallen:
> we probably need to update something in the CircleCI config to use Node v15

I just checked, and it looks like the `package-lock` for the repo was generated with Npm 7, but CircleCI is still on npm 6. It appears to use node 14, which is also working for me locally now.

noahtallen:
I experimented with a few approaches, but haven't found a solution yet.

.circleci/config.yml
```diff
@@ -13,6 +13,16 @@ jobs:
...
+      - run:
+          name: Npm 7
+          command: |
+            set +e
+            set +x
+            export NVM_DIR="$HOME/.nvm" && (
+              git clone https://github.com/nvm-sh/nvm.git "$NVM_DIR"
+              cd "$NVM_DIR"
+            ) && \. "$NVM_DIR/nvm.sh" --no-use
+            nvm install --latest-npm
```
noahtallen, line 25: we can probably ditch this, as it didn't solve the problem in CI

adekbadek: Yup, removed

adekbadek:
With the build problems resolved (#1142), we now can deal with JS tests & linting failing 🕵️‍♀️

thomasguillot:
> @thomasguillot regarding this:
> 
> > Dependency `react-daterange-picker` is stale and doesn't appear to be updated anymore. It doesn't list react 17 support, but I suppose it probably supports it anyways.
> 
> It seems to me like this package is only used in the `DateRangePicker` Newspack component, ~but I'm not seeing this component being used anywhere. Can we deprecate it entirely? Can we replace the third-party package with the [core `DateTime` component](https://developer.wordpress.org/block-editor/reference-guides/components/date-time/) instead?~
> 
> Edit: Nevermind, I see it's used in the Campaigns analytics view. Here's another package that's maybe descended from the one we're using that is still maintained: https://github.com/wojtekmaj/react-daterange-picker
> 
> I also still think it might be possible to use the core `DateTime` component using the `isDayHighlighted` prop.

I leave this up to @adekbadek 😊

noahtallen:
I updated jest and the eslint packages and added a mock for a missing window global. This solved the unit test problem.

The issue with linting is because for some reason, `eslint` cannot find all of the eslint plugins which `@wordpress/eslint-plugin` installs. For some reason, it isn't finding them as a transitive dependency of the wp eslint plugin, so I've added all of the plugins it complained about to our dev dependencies. That seems to make eslint happy

miguelpeixe:
Although tests are passing, I'm getting build errors with node `v16.7.0`. We need to update the test images so the node version matches our target version.

According to `docker run --rm circleci/php:7.2-node-browsers node -v`, the node version for the tests is `v14.15.2`.

The issue I'm having is node-sass `v4.0.3` not being available in the target version `v16.7.0`.

noahtallen:
Yeah, `node-sass` has been deprecated -- we had to replace it with plain `sass` in `wp-calypso` (see https://github.com/Automattic/wp-calypso/pull/52729)

I was testing with Node 14 locally, which seems to work okay. Which version to we officially support?

miguelpeixe:
Oh, I didn't know that happened! We might have to do the same.

My understanding is that we'd like `node` to match [`wp-calypso`'s version](https://github.com/Automattic/wp-calypso/blob/trunk/.nvmrc). But I'd like @dkoo and @adekbadek's inputs on this.

dkoo:
Yeah, I think we'll ideally want to match Calypso's version. So that means deprecating `node-sass` in a similar way for all of our repos. This is one instance where I can see the benefit of a monorepo. 😄

noahtallen:
hahaha, yeah, I've been updating loads of dependencies in calypso, and it seems like every update requires 5 other updates as well :facepalm:

noahtallen:
I do want to note that this is no longer blocking us. We just hid the peer dependency warnings since they didn't end up breaking anything. (Though it would be lovely to not have old transitive dependencies!) As a result, I have to prioritize other work now, and will leave this PR in the hands of the newspack team :)

dkoo:
Update: I think running `npm i --legacy-peer-deps` was the missing key here. After running that and then updating some more packages and downgrading others that were causing dependency conflicts, this is now building on Node v16.11.1 and passing all CI checks. A quick functional test has revealed no crashes so far.

adekbadek (changes requested):
Added a small fix 0d1e204f90e7e19b900d4d7a6a31bacbfaa89a8d. 
I've noticed that the images aren't loading (even though they do end up in `dist`). Probably related to the webpack major version change.

image

dkoo:
> Added a small fix 0d1e204f90e7e19b900d4d7a6a31bacbfaa89a8d. 

Good catch, thanks for the fix!

> I've noticed that the images aren't loading (even though they do end up in `dist`). Probably related to the webpack major version change.

Confirmed—will investigate. Is it just happening anytime a static asset is referenced in JS?

dkoo:
> I've noticed that the images aren't loading (even though they do end up in `dist`). Probably related to the webpack major version change.

This is patched in 3124859. Turns out this is a bug in `calypso-build`, not webpack or our config.

The major version bump to webpack v5 is a requirement for the latest `calypso-build@9.0.0`. Unfortunately, this release contains a bug that causes all asset paths to be outputted to the bundle relative to the root server path, not the bundle output path.

[See here](https://github.com/Automattic/wp-calypso/pull/56390) for details on the bug and the fix, which has not yet been released. [See here](https://github.com/Automattic/jetpack/pull/20972) for the patch by the Jetpack team which inspired our own patch. 

Once the Calypso team releases their fix, we should be able to update our package version and remove the patch. Until then, any repo that updates to `calypso-build@9.0.0` will need to implement this same patch in its own webpack config.

noahtallen:
> Once the Calypso team releases their fix

I'll discuss with the team about releasing calypso-build to npm 👍

dkoo:
> I'll discuss with the team about updating calypso-build on npm 👍

Awesome—thanks for the explanation, @noahtallen!

adekbadek (approved):
Getting JS console errors from a `/wp-includes/js/mediaelement/mediaelement-and-player.min.js?ver=4.2.16` file, but this does not have any bearing on this plugin's functionality. 

image

noahtallen:
> Once the Calypso team releases their fix

Calypso build v10 has now been released 👍

dkoo:
> Getting JS console errors from a `/wp-includes/js/mediaelement/mediaelement-and-player.min.js?ver=4.2.16` file, but this does not have any bearing on this plugin's functionality.
> 
> image

Does that only happen on this branch? And/or is it something we can replicate on a temp site for debugging?

dkoo:
> Calypso build v10 has now been released 👍

Confirmed in f950761 that v10 works and no longer requires the webpack config patch.

adekbadek:
> Does that only happen on this branch? And/or is it something we can replicate on a temp site for debugging?

@dkoo Yes. DM's you a temp site

assets/components/package.json
```diff
@@ -1,51 +1,51 @@
...
+		"url": "https://github.com/Automattic/newspack-plugin.git",
+		"directory": "assets/src/components"
+	},
+	"main": "dist/esm/index.js",
+	"publishConfig": {
+		"access": "public"
+	},
+	"dependencies": {
+		"@wordpress/base-styles": "^4.0.0",
+		"@wordpress/components": "^17.0.0",
```
noahtallen, line 27: I think new versions have come out in the meantime (v18 is available), so it might be worth updating again to avoid having to do it later

dkoo:
Couple of updates: [this error](https://github.com/Automattic/newspack-plugin/pull/1135#pullrequestreview-798739699) is replicable on brand new sites and seems somehow related to webpack and the importing of SCSS files. For example, on the Plugins screen, if you disable importing `./plugins-screen.scss` in `plugins-screen.js`, the error doesn't occur. I'm not seeing any functional errors as a result of it. When `mediaelement-and-player.min.js` is actually used (e.g. in the classic Widgets screen), the error also doesn't occur. I'm not sure where to go from here on this bug.

Also, when testing Newspack Components as a package imported to another repo (Newspack Blocks), I ran into errors related to CommonJS (`module is not defined`). Switching the main entrypoint for the package from CJS to ESM as in 3c8aca7 fixes the issue in Newspack Blocks, but I'm not sure of the wider ramifications for Calypso.

assets/components/package.json
dkoo: Like trying to hit a moving target! This PR implements Dependabot, so I might freeze the updates as they are right now since things are currently building and functional.

noahtallen: cool, that also sounds like a good plan!

> Like trying to hit a moving target

Yeah, it can be very difficult 😬

dkoo:
Another update! I've found that if I include `mediaelement-core` as a dependency in our `wp_enqueue_script` statements, it fixes [these JS errors](https://github.com/Automattic/newspack-plugin/pull/1135#pullrequestreview-798739699).

But this is frustrating: now I'm getting a new JS error that I haven't seen before when viewing the Campaigns wizard:

```
Uncaught TypeError: _interopRequireDefault is not a function
    <anonymous> webpack://newspack/./node_modules/dom-helpers/class/addClass.js?:8
    js https://test-mcmansion.newspackstaging.com/wp-content/plugins/newspack-plugin/dist/popups.js?ver=1636499874:1349
[...]
```

dkoo:
458beb8 fixes the `_interopRequireDefault is not a function` build error—which was coming from the [recharts](https://www.npmjs.com/package/recharts) package—by replacing recharts with [google-react-charts](https://react-google-charts.com/area-chart).

The Campaigns wizard should be restored now and all other wizard pages should still be working as expected.

However, I'm now getting failing phpunit tests in CI:

```
Error: The PHPUnit Polyfills library is a requirement for running the WP test suite.
If you are trying to run plugin/theme integration tests, make sure the PHPUnit Polyfills library (https://github.com/Yoast/PHPUnit-Polyfills) is available and either load the autoload file of this library in your own test bootstrap before calling the WP Core test bootstrap file; or set the absolute path to the PHPUnit Polyfills library in a "WP_TESTS_PHPUNIT_POLYFILLS_PATH" constant to allow the WP Core bootstrap to load the Polyfills.

If you are trying to run the WP Core tests, make sure to set the "WP_RUN_CORE_TESTS" constant to 1 and run `composer install` before running the tests.
Once the dependencies are installed, you can run the tests using the Composer-installed version of PHPUnit or using a PHPUnit phar file, but the dependencies do need to be installed whichever way the tests are run.
```

I think this is out-of-scope to fix in this PR, though, since it looks like it requires some migration on our part and seems unrelated to the changes in this branch. @Automattic/newspack-product please confirm.

.circleci/config.yml
```diff
@@ -76,7 +76,7 @@ jobs:
             composer install
             composer global require "phpunit/phpunit=5.7.*"
             rm -rf $WP_TESTS_DIR $WP_CORE_DIR
-            bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 latest
+            bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 5.8.1
```
adekbadek, line 79: `5.8.2` [introduces some tricky changes](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/) to the test framework

adekbadek (commented):
Fixed the PHPUnit thing in 145fd5e17cabbdddc2afd37ec1ec45245d96ac53

Left one question about `newspack-components` version, otherwise it works as expected.

assets/components/package.json
```diff
@@ -1,51 +1,51 @@
...
-    "ignore": [
-      "./node_modules"
-    ]
-  },
-  "scripts": {
-    "prepublishOnly": "transpile && node copy-styles.js",
-    "postpublish": "rm -r dist && rm -r shared"
-  }
+	"name": "newspack-components",
+	"version": "1.7.0",
```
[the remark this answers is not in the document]
dkoo: I changed it in anticipation of having to make a new release. Is that not required?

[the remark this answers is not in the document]
adekbadek: Ah, didn't think about it. Yeah, makes sense.

adekbadek (approved):

.circleci/config.yml
dkoo: Nice temporary fix!

PR #1122, opened 2021-09-02
feat(oauth): use a self-hosted proxy for authenication

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

The OAuth saga continues. With this approach, the OAuth proxy is hosted on an arbitrary site, not wordpress.com. 

### How to test the changes in this Pull Request:

1. Define `NEWSPACK_GOOGLE_OAUTH_PROXY` in the site's environment, ensure WPCOM is authorised (visit the Support wizard to check)
2. To start afresh and get a `refresh_token`, remove Newspack from [authorised Google apps](https://myaccount.google.com/permissions)
3. Also, remove the `_newspack_google_oauth` user meta, if it exists
4. Visit the Newspack dashboard and complete the OAuth flow by clicking on the "Google Connection" tile
5. Check back after at least an hour to see if the token will be refreshed - you should still see "Authorized Google as <email-address>" on the "Google Connection" tile
6. Test some re-authentication scenarios by re-doing steps 2 or 3 (separately) and authenticating again
7. Test as a WPCOM user who is not Newspack customer - you should not be able to complete the OAuth flow

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1145, opened 2021-09-16
feat/connections view

### LINKED ISSUE #984: "Connections" view
https://github.com/Automattic/newspack-plugin/pull/962 introduces Google connection, on top of connection to WPCOM. There should be a separate view to manage those. User should be able to:

- Connect their Google & WPCOM accounts 
- Revoke the connections
- Google is dependent on WPCOM, the UI should reflect that

adekbadek:
### All Submissions:

Adds a new wizard, called Connections, for managing all third-party authorisations.

@thomasguillot - as per [your suggestion](https://github.com/Automattic/newspack-plugin/issues/984#issuecomment-862321456), this is kept similar to the Integrations step of Setup flow:

image

Closes #984

1. On a fresh site, load the Newspack dashboard. Observe there's no "Connections" wizard available
2. Add the `NEWSPACK_WPCOM_CLIENT_ID` environment variable
3. Observe the Connections wizard is now available, and it allows the user to connect the site to WordPress.com
4. Add the `NEWSPACK_GOOGLE_OAUTH_PROXY` environment variable (see #1122 for more details)
5. Observe an additional connection option - Google
6. Observe that the Google option is disabled, with information that first WordPress.com should be connected
7. Authorise WordPress.com
8. Observe that now Google can be authorised - do it
9. Revoke permissions of both, observe the authorisation flow can be followed again

### Other information:

### DIFF
--- a/assets/components/src/footer/index.js
+++ b/assets/components/src/footer/index.js
@@ -18,7 +18,6 @@ const Footer = ( { simple } ) => {
 	const componentsDemo = window && window.newspack_urls && window.newspack_urls.components_demo;
 	const setupWizard = window && window.newspack_urls && window.newspack_urls.setup_wizard;
 	const resetUrl = window && window.newspack_urls && window.newspack_urls.reset_url;
-	const resetWpcomUrl = window && window.newspack_urls && window.newspack_urls.reset_wpcom_url;
 	const pluginVersion = window && window.newspack_urls && window.newspack_urls.plugin_version;
 	const removeStarterContent =
 		window && window.newspack_urls && window.newspack_urls.remove_starter_content;
@@ -54,12 +53,6 @@ const Footer = ( { simple } ) => {
 		footerElements.push( {
 			label: __( 'Reset Newspack', 'newspack' ),
 			url: resetUrl,
-		} );
-	}
-	if ( resetWpcomUrl ) {
-		footerElements.push( {
-			label: __( 'Reset WordPress.com Authentication', 'newspack' ),
-			url: resetWpcomUrl,
 		} );
 	}
 	if ( removeStarterContent ) {
--- a/assets/components/src/waiting/index.js
+++ b/assets/components/src/waiting/index.js
@@ -34,14 +34,13 @@ class Waiting extends Component {
 	 * Render
 	 */
 	render() {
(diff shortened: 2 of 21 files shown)


thomasguillot:
Brilliant!

miguelpeixe (changes requested):
I was not able to connect to Google even though I had the WordPress.com connection and the proper `NEWSPACK_GOOGLE_OAUTH_PROXY` defined. I tried with a local setup and a newly configured Newspack site using JN.

After attempting to connect, the API response to the `/oauth/google` endpoint remained `{"success":false}` without any further information and the screen stuck on loading:

image

Disconnecting from WordPress.com redirects to the dashboard homepage, is this intended? I expected to stay on the Connections view.
 
I also wonder if we can add the WordPress.com and Google connections to the "Integrations" onboarding step and also bring those connections to this view? Unify the connections/integrations component would also be better to maintain.

adekbadek:
Should be fixed now! The issue was with WPCOM token encoding.

> Disconnecting from WordPress.com redirects to the dashboard homepage, is this intended? I expected to stay on the Connections view.

Good idea, changed in 64c2953a4002e04c8581f057d959c4568e1aba38

> I also wonder if we can add the WordPress.com and Google connections to the "Integrations" onboarding step and also bring those connections to this view? Unify the connections/integrations component would also be better to maintain.

Good idea too, but I'd prefer to tackle that in a separate PR to unblock other things with this PR.

miguelpeixe (approved):
Thank you for revising, it is working as expected also with the support constants not present.

One last non-blocking suggestion, as part of the OAuth flow we are redirected to a page with the `csrf_token` and `access_token` parameters for validation/registration. While on this page we are presented with an incorrect state of the application. Would be best if presented with some sort of "Authorizing..." message to represent the temporary aspect of this screen.

adekbadek:
> While on this page we are presented with an incorrect state of the application. Would be best if presented with some sort of "Authorizing..." message to represent the temporary aspect of this screen.

There's a [waiting animation in this flow](https://github.com/Automattic/newspack-plugin/blob/493772a2cbb47f7e1b63903b982f36a97952986b/assets/wizards/connections/views/main/index.js#L24), but the with all the redirects it's not so smooth. I agree this could be improved, but it's not super-important.

PR #1146, opened 2021-09-16
Feat: Ads GAM custom targeting keys integration

miguelpeixe:
### All Submissions:

Improves notices for Ads GAM integration and adds custom targeting keys update.
Testing instructions are over at https://github.com/Automattic/newspack-ads/pull/172.

### Other information:

### DIFF
--- a/assets/wizards/advertising/index.js
+++ b/assets/wizards/advertising/index.js
@@ -258,11 +258,10 @@ class AdvertisingWizard extends Component {
 	prepareData = data => {
 		return {
 			...data,
-			adUnits:
-				data.ad_units.reduce( ( result, value ) => {
-					result[ value.id ] = value;
-					return result;
-				}, {} ) || [],
+			adUnits: data.ad_units.reduce( ( result, value ) => {
+				result[ value.id ] = value;
+				return result;
+			}, {} ),
 		};
 	};
 
--- a/assets/wizards/advertising/index.js
+++ b/assets/wizards/advertising/index.js
@@ -260,7 +260,7 @@ class AdvertisingWizard extends Component {
 		return {
 			...data,
 			adUnits:
-				data?.ad_units?.reduce( ( result, value ) => {
+				data.ad_units.reduce( ( result, value ) => {
 					result[ value.id ] = value;
 					return result;
 				}, {} ) || [],
--- a/assets/wizards/advertising/index.js
+++ b/assets/wizards/advertising/index.js
@@ -16,6 +16,7 @@ import { __ } from '@wordpress/i18n';
 import { withWizard } from '../../components/src';
 import Router from '../../components/src/proxied-imports/router';
 import { AdUnit, AdUnits, Placements, Services, Suppression } from './views';
+import './style.scss';
 
 const { HashRouter, Redirect, Route, Switch } = Router;
 const CREATE_AD_ID_PARAM = 'create';
@@ -258,10 +259,11 @@ class AdvertisingWizard extends Component {
 	prepareData = data => {
 		return {
 			...data,
-			adUnits: data.ad_units.reduce( ( result, value ) => {
-				result[ value.id ] = value;
-				return result;
-			}, {} ),
+			adUnits:
+				data?.ad_units?.reduce( ( result, value ) => {
+					result[ value.id ] = value;
+					return result;
+				}, {} ) || [],
 		};
 	};
 
--- a/assets/wizards/advertising/index.js
+++ b/assets/wizards/advertising/index.js
@@ -333,6 +333,7 @@ class AdvertisingWizard extends Component {
 										subHeaderText={ __( 'Monetize your content through advertising', 'newspack' ) }
 										adUnits={ adUnits }
 										service={ 'google_ad_manager' }
+										serviceData={ services.google_ad_manager }
 										onDelete={ id => this.deleteAdUnit( id ) }
 										buttonText={ __( 'Add an ad unit', 'newspack' ) }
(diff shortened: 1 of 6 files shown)


assets/wizards/advertising/views/services/index.js
```diff
@@ -68,6 +68,22 @@ const Services = ( { services, toggleService } ) => {
 				toggleOnChange={ value => toggleService( 'google_ad_manager', value ) }
 				titleLink={ google_ad_manager ? '#/google_ad_manager' : null }
 				href={ google_ad_manager && '#/google_ad_manager' }
+				notification={
+					google_ad_manager.error
+						? [ google_ad_manager.error ]
+						: google_ad_manager.created_targeting_keys?.length > 0 && [
+								__( 'Created custom targeting keys:' ) + '\u00A0',
+								google_ad_manager.created_targeting_keys.join( ', ' ) + '. \u00A0',
+								// eslint-disable-next-line react/jsx-indent
```
miguelpeixe, line 77: I've disabled lint here because the linter is requiring one less tab, which does not make sense to me.

assets/wizards/advertising/views/services/index.js
```diff
@@ -68,6 +68,22 @@ const Services = ( { services, toggleService } ) => {
 				toggleOnChange={ value => toggleService( 'google_ad_manager', value ) }
 				titleLink={ google_ad_manager ? '#/google_ad_manager' : null }
 				href={ google_ad_manager && '#/google_ad_manager' }
+				notification={
```
adekbadek, line 71: Suggestion: display the notice in the view with GAM ad units:

image

includes/configuration_managers/class-newspack-ads-configuration-manager.php
```diff
@@ -32,6 +34,23 @@ public function is_configured() {
 		return class_exists( 'Newspack_Ads_Model' );
 	}
 
+	/**
+	 * Update custom targeting keys.
+	 *
+	 * @return string[]|WP_Error Created custom targeting keys or error if it fails.
+	 */
+	public function update_custom_targeting_keys() {
```
adekbadek, line 42: Suggestion: we might need to do more setup chores on wizard load in the future. This could be a generic setup handler, with custom targeting key configuration as the single (for now at least) setup step. This could then be handled by `Newspack_Ads_Model`, so this plugin would not bother with the implementation details and would not have to be updated when new setup steps are added.

adekbadek (changes requested):

assets/wizards/advertising/views/services/index.js
miguelpeixe: Good idea! Implemented at e056691eb5c51965d0da6d25be6ad0502a539f62 

I'm displaying it in both places so the user doesn't miss it in case they refresh or leave before accessing the Ad Units view.

assets/wizards/advertising/index.js
```diff
@@ -258,10 +259,11 @@ class AdvertisingWizard extends Component {
 	prepareData = data => {
 		return {
 			...data,
-			adUnits: data.ad_units.reduce( ( result, value ) => {
-				result[ value.id ] = value;
-				return result;
-			}, {} ),
+			adUnits:
+				data?.ad_units?.reduce( ( result, value ) => {
```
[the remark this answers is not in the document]
miguelpeixe: This would be in the case of `get_ad_units()` returning error due to failed sync on `newspack-ads`. This was actually not necessary because the `prepareData()` method would never be called in this situation.

a5df3f3e5f7c1f3c54639082d64a39fa6ab7240f reverts this change and also fix error handling when `get_ad_units()` returns WP_Error instead of failing with exception, as it is handled currently.

assets/wizards/advertising/style.scss
```diff
@@ -0,0 +1,3 @@
+.newspack-notice {
+	margin: 0;
```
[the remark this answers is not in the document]
miguelpeixe: I copied this fix from [the Dashboard wizard](https://github.com/Automattic/newspack-plugin/blob/feat/ads-gam-targeting-keys-integration/assets/wizards/dashboard/style.scss#L27-L29) to improve how the notices are displayed when shown above the wizard header.

This is no longer necessary by https://github.com/Automattic/newspack-ads/commit/b1668ff7ee6a1854b00014fb5cefdb9c55e5c698 no longer returning a `notice` level error, leaving this wizard without any errors of this kind.

1a977dc705f9eff5c36eb869535328397086392f reverts this change.

assets/wizards/advertising/index.js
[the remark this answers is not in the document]
miguelpeixe: Update: I left some residual code, 9b74997810afd433b652e1ff58d955262854f8c1 reverts it completely.

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -99,6 +101,22 @@ const AdUnits = ( {
...
+				<Notice
+					noticeText={ [
+						__( 'Created custom targeting keys:' ) + '\u00A0',
+						serviceData.created_targeting_keys.join( ', ' ) + '. \u00A0',
+						<ExternalLink
+							href={ `https://admanager.google.com/${ serviceData.network_code }#inventory/custom_targeting/list` }
+							key="google-ad-manager-custom-targeting-link"
+						>
+							{ __( 'Visit your GAM dashboard' ) }
+						</ExternalLink>,
```
adekbadek, line 115: Nit: code repetition (also in `assets/wizards/advertising/views/services/index.js`), but we can live with that.

adekbadek (approved):

PR #1147, opened 2021-09-17
feat: integrate newspack-image-credits into main plugin and wizard UI

dkoo:
### All Submissions:

Integrates the separate [Newspack Image Credits plugin](https://github.com/Automattic/newspack-image-credits) into this plugin, which will render the separate plugin obsolete for Newspack sites.

Instead of showing the settings in the Media settings page, shows them in the Newspack Site Design > Settings wizard UI.

Also makes a minor stylistic change to the Image Uploader component to display the selected image at the proper scale instead of as a tiny background thumbnail. cc @thomasguillot to review/test this change.

**Note:** I toyed with the idea of refactoring the options that store these settings into proper theme mods so they can also be updated in the Customizer, like the other site design settings. But I worry that might cause disruption for sites that rely on the separate plugin, which uses options. I'm not sure refactoring is worth it—thoughts?

1. First, install and activate the [Newspack Image Credits plugin](https://github.com/Automattic/newspack-image-credits) if not already installed. Follow testing instructions in https://github.com/Automattic/newspack-image-credits/pull/9 so that you have some pre-existing images with test credits and settings before you test this PR.
2. Check out this branch, rebuild, and deactivate the Newspack Image Credits plugin.
3. Confirm that the images and settings you tested in step 1 continue to work as they did prior to checking out this branch.
4. Go to **WP admin > Newspack > Site Design > Settings**. Confirm there's a new section here with the Media Credits settings:

Screen Shot 2021-09-17 at 5 21 45 PM

5. Play with the settings here and save. Confirm that your settings save as expected, even after refreshing the page. Also confirm that your saved settings are reflected on the front-end as expected.
6. Reactivate the Newspack Image Credits plugin and confirm that having both plugins active won't cause any trouble in the admin or on the front-end (so we won't have to worry about immediately deactivating this plugin on all sites that might have it).

### Other information:

### DIFF
--- a/includes/class-newspack-image-credits.php
+++ b/includes/class-newspack-image-credits.php
@@ -1,0 +1,430 @@
+<?php
+/**
+ * Image credits for media library items.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+/**
+ * Newspack_Image_Credits class.
+ */
+class Newspack_Image_Credits {
+
+	/**
+	 * Meta key for the credit.
+	 */
+	const MEDIA_CREDIT_META = '_media_credit';
+
+	/**
+	 * Meta key for the credit URL.
+	 */
+	const MEDIA_CREDIT_URL_META = '_media_credit_url';
+
+	/**
+	 * Meta key for the credit organization.
+	 */
+	const MEDIA_CREDIT_ORG_META = '_navis_media_credit_org';
+
+	/**
+	 * Meta key for whether the image is distributable.
+	 */
+	const MEDIA_CREDIT_CAN_DISTRIBUTE_META = '_navis_media_can_distribute';
+
+	/**
+	 * Hook actions and filters.
+	 */
+	public static function init() {
+		add_filter( 'attachment_fields_to_save', [ __CLASS__, 'save_media_credit' ], 10, 2 );
+		add_filter( 'attachment_fields_to_edit', [ __CLASS__, 'add_media_credit' ], 10, 2 );
+		add_filter( 'get_the_excerpt', [ __CLASS__, 'add_credit_to_attachment_excerpts' ], 10, 2 );
+		add_filter( 'render_block', [ __CLASS__, 'add_credit_to_image_block' ], 10, 2 );
+		add_filter( 'wp_get_attachment_image_src', [ __CLASS__, 'maybe_show_placeholder_image' ], 11, 4 );
+	}
+
+	/**
+	 * Get media credit info for an attachment.
+	 *
+	 * @param int $attachment_id Post ID of the attachment.
+	 * @return array Credit info. See $output at the top of this method.
+	 */
+	public static function get_media_credit( $attachment_id ) {
+		$output = [
+			'id'             => $attachment_id,
+			'credit'         => '',
+			'credit_url'     => '',
+			'organization'   => '',
+			'can_distribute' => false,
+		];
+
+		$credit = get_post_meta( $attachment_id, self::MEDIA_CREDIT_META, true );
+		if ( $credit ) {
+			$output['credit'] = esc_attr( $credit );
+		}
+
+		$credit_url = get_post_meta( $attachment_id, self::MEDIA_CREDIT_URL_META, true );
+		if ( $credit_url ) {
+			$output['credit_url'] = esc_attr( $credit_url );
+		}
+
+		$organization = get_post_meta( $attachment_id, self::MEDIA_CREDIT_ORG_META, true );
+		if ( $organization ) {
+			$output['organization'] = esc_attr( $organization );
+		}
+
+		$can_distribute = get_post_meta( $attachment_id, self::MEDIA_CREDIT_CAN_DISTRIBUTE_META, true );
+		if ( $can_distribute ) {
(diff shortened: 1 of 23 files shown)


thomasguillot:
@dkoo I've pushed a few visual updates. The `ImageUpload` component broke the Design Wizard but It should now be fixed 😊

Also, the "Remove" button doesn't seem to work. Are you able to reproduce this?

dkoo:
> Also, the "Remove" button doesn't seem to work. Are you able to reproduce this?

Ah, good catch—fixed in 94164724!

includes/class-newspack-image-credits.php
```diff
@@ -0,0 +1,430 @@
...
+	/**
+	 * Add media credit to image blocks when possible.
+	 *
+	 * @param string $block_output HTML block output.
+	 * @param array  $block Raw block info.
+	 * @return string Modified $block_output.
+	 */
+	public static function add_credit_to_image_block( $block_output, $block ) {
+		// Only for core image blocks or Jetpack slideshow blocks.
+		if ( 'core/image' !== $block['blockName'] && 'jetpack/slideshow' !== $block['blockName'] ) {
```
adekbadek, line 221: Nit: these conditions are duplicated below, can be removed.

includes/class-newspack-image-credits.php
```diff
@@ -0,0 +1,430 @@
...
+
+	/**
+	 * Hook actions and filters.
+	 */
+	public static function init() {
+		add_filter( 'attachment_fields_to_save', [ __CLASS__, 'save_media_credit' ], 10, 2 );
+		add_filter( 'attachment_fields_to_edit', [ __CLASS__, 'add_media_credit' ], 10, 2 );
+		add_filter( 'get_the_excerpt', [ __CLASS__, 'add_credit_to_attachment_excerpts' ], 10, 2 );
+		add_filter( 'render_block', [ __CLASS__, 'add_credit_to_image_block' ], 10, 2 );
+		add_filter( 'wp_get_attachment_image_src', [ __CLASS__, 'maybe_show_placeholder_image' ], 11, 4 );
```
adekbadek, line 43: The placeholder image applies to the site icon, too. This image "placement" should not be affected by the replacement.

adekbadek (approved):
Added a note about the site icon being affected, but this can be tacked in a separate PR.

includes/class-newspack-image-credits.php
dkoo: I wasn't sure about that, but I think you're probably right since no site should be using a non-licensed third-party image as their site logo. 😆 Fixed in d2376bf but fixed the wonky logic in 07604e3.

includes/class-newspack-image-credits.php
```diff
@@ -322,14 +317,20 @@ function( $matches ) use ( $placeholder_image, $size ) {
 	 * @return array Filtered array of image data.
 	 */
 	public static function maybe_show_placeholder_image( $image_data, $attachment_id, $size, $icon ) {
-		if ( ! is_singular() && ! is_archive() ) {
```
dkoo: I'm not sure why I had this logic at first. It prevents the placeholder image from appearing in search results.

includes/class-newspack-image-credits.php
```diff
@@ -322,14 +317,20 @@ function( $matches ) use ( $placeholder_image, $size ) {
...
 			return $image_data;
 		}
 
-		$media_credit      = get_post_meta( $attachment_id, self::MEDIA_CREDIT_META, true );
-		$placeholder_image = self::get_settings( 'newspack_image_credits_placeholder' );
+		$site_icon       = get_option( 'site_icon' );
+		$custom_logo     = get_theme_mod( 'custom_logo' );
+		$custom_logo_alt = get_theme_mod( 'newspack_alternative_logo', '' );
+		$is_placeholder  = intval( $placeholder_image ) !== intval( $attachment_id );
+		$is_logo         = intval( $site_icon ) !== intval( $attachment_id ) && intval( $custom_logo ) !== intval( $attachment_id ) && intval( $custom_logo_alt ) !== intval( $attachment_id );
```
adekbadek: A logic issue, should be:

```php
		$is_placeholder  = intval( $placeholder_image ) === intval( $attachment_id );
		$is_logo         = intval( $site_icon ) === intval( $attachment_id ) || intval( $custom_logo ) === intval( $attachment_id ) || intval( $custom_logo_alt ) === intval( $attachment_id );
```

adekbadek (changes requested):

includes/class-newspack-image-credits.php
adekbadek: Beat me to it!

adekbadek (approved):

yogeshbeniwal:
@dkoo Does enabling "Can distribute" adds metadata or license information to post html? Tried it for a featured image, didn't notice any change in page source.

claudiulodro:
@yogeshbeniwal no, it's purely for internal newsroom use, so article writers know which images they can use in their articles, and it's mostly there for backwards compatibility (Largo also had that checkbox). It's not displayed on the frontend anywhere.

PR #1151, opened 2021-09-21
Fix/ads handle legacy network code

miguelpeixe:
### All Submissions:

Allow update and remove active Ads GAM credentials and implements https://github.com/Automattic/newspack-ads/issues/183.

Checkout this branch, run `npm run build` and continue at https://github.com/Automattic/newspack-ads/pull/184

### Other information:

### DIFF
--- a/assets/wizards/advertising/views/ad-units/index.js
+++ b/assets/wizards/advertising/views/ad-units/index.js
@@ -219,7 +219,7 @@ const AdUnits = ( {
 									{ __( 'Remove credentials', 'newspack' ) }
 								</Button>
 								<Button isSecondary onClick={ () => setIsRemoving( false ) }>
-									{ __( 'Dismiss', 'newspack' ) }
+									{ __( 'Cancel', 'newspack' ) }
 								</Button>
 							</Card>
 						</Modal>
--- a/assets/wizards/advertising/views/ad-units/index.js
+++ b/assets/wizards/advertising/views/ad-units/index.js
@@ -130,19 +130,21 @@ const AdUnits = ( {
 						noticeText={ __( 'Currently operating in legacy mode.', 'newspack' ) }
 						isWarning
 					/>
-					<ButtonCard
-						onClick={ () => {
-							credentialsInputFile.current.click();
-						} }
-						title={ __( 'Connect your GAM account', 'newspack' ) }
-						desc={ [
-							__(
-								'Upload your Service Account credentials file to connect your GAM account with Newspack Ads.',
-								'newspack'
-							),
-							fileError && <Notice noticeText={ fileError } isError />,
-						] }
-					/>
+					{ ! serviceData.status.incompatible && (
+						<ButtonCard
+							onClick={ () => {
+								credentialsInputFile.current.click();
+							} }
+							title={ __( 'Connect your GAM account', 'newspack' ) }
+							desc={ [
+								__(
+									'Upload your Service Account credentials file to connect your GAM account with Newspack Ads.',
+									'newspack'
+								),
+								fileError && <Notice noticeText={ fileError } isError />,
+							] }
+						/>
+					) }
 				</>
 			) }
 			{ serviceData.created_targeting_keys?.length > 0 && (
--- a/assets/wizards/advertising/views/ad-units/index.js
+++ b/assets/wizards/advertising/views/ad-units/index.js
@@ -19,6 +19,7 @@ import {
 	TextControl,
 	Button,
 	Card,
+	Modal,
 	Notice,
 	withWizardScreen,
 } from '../../../../components/src';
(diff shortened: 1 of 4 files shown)


assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -87,17 +91,19 @@ const AdUnits = ( {
...
+
 	return (
 		<>
 			{ serviceData.status.can_connect ? (
 				<>
 					{ isDisplayingNetworkMismatchNotice && (
 						<Notice
 							noticeText={ __(
-								'Your GAM network code is different than the network code the site was configured with. Editing has been disabled.',
+								'Your GAM network code is different than the network code the site was configured with.',
```
dkoo, line 103: Would love to see a bit more clarity on what this means, e.g.:

```suggestion
								'The GAM network code from your uploaded credentials differ from the previously saved network code. Legacy ad units associated with the prior network code will no longer work.',
```

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -167,6 +162,70 @@ const AdUnits = ( {
...
+									isDestructive
+									onClick={ () => {
+										removeGAMCredentials();
+										setIsRemoving( false );
+									} }
+								>
+									{ __( 'Remove credentials', 'newspack' ) }
+								</Button>
+								<Button isSecondary onClick={ () => setIsRemoving( false ) }>
+									{ __( 'Dismiss', 'newspack' ) }
```
dkoo, line 222: Just for consistency with other Newspack UI:

```suggestion
									{ __( 'Cancel', 'newspack' ) }
```

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -167,6 +162,70 @@ const AdUnits = ( {
...
+			<div className="flex items-end" style={ { margin: '-30px 0' } }>
+				<TextControl
+					className="mr2"
+					label={ __( 'Network code', 'newspack' ) }
+					value={ networkCode }
+					onChange={ setNetworkCode }
+					disabled={ serviceData.status.connected }
+				/>
+				<Card buttonsCard noBorder>
+					<Button onClick={ saveNetworkCode } isPrimary disabled={ serviceData.status.connected }>
```
dkoo, line 174: If you've uploaded credentials, maybe there's no reason to even show the disabled "Save" button? I would prefer to hide it in this case.

dkoo (approved):
Functionally, this solves the problem described in https://github.com/Automattic/newspack-ads/issues/183 really well. I have a few nitpicks about the UI itself, but since these are more subjective let's consider them non-blocking.

assets/wizards/advertising/views/ad-units/index.js
```diff
@@ -167,6 +162,70 @@ const AdUnits = ( {
...
+								isSecondary
+								isSmall
+							>
+								{ __( 'Upload new credentials', 'newspack' ) }
+							</Button>
+							<Button
+								onClick={ () => {
+									setIsRemoving( true );
+								} }
+								// isQuarternary
```
adekbadek, line 192: Line to be removed

adekbadek (commented):
I agree with @dkoo 's review

assets/wizards/advertising/views/ad-units/index.js
miguelpeixe: > Legacy ad units associated with the prior network code will no longer work.

This is not _exactly_ what happens. Because of [this change](https://github.com/Automattic/newspack-ads/pull/184/files#diff-727e5225ecc01afcf95934ec600179f9c67bb561963530a15966d643a18f99dbL328-L331), the ad units will continue to load on the frontend, but the requested network code will be the one coming from the credentials.

The likely result of this is the ad unit not working because it wouldn't find the appropriate combination of size and code to load from on the new network.

miguelpeixe: I suggest we change it to:

> Your GAM network code is different than the network code the site was configured with. Legacy ad units are likely to not load.

assets/wizards/advertising/views/ad-units/index.js
miguelpeixe: Thank you for catching that!

miguelpeixe:
Changes have been implemented at 9300755fe1debafa6b953a11ddc95c6cf3d408d3

assets/wizards/advertising/views/ad-units/index.js
dkoo: That sounds good to me—communicates an important nuance.

adekbadek (approved):

dkoo (approved):

PR #1157, opened 2021-09-22
fix: restore scroll for wizard screens navigation

miguelpeixe:
### All Submissions:

Wizard screens navigation using react-router is not restoring scroll position, which can be confusing.

This PR implements "scroll to top" to every pathname change for components `withWizardScreen`.

1. In the master branch, visit the Newspack Ads dashboard
2. Click on the Google Ad Manager "Configure", scroll down and click on the "Back to advertising options" button
3. Observe your scroll is not at the top of the page after the screen change
4. Switch to this branch, run `npm run build`
5. Repeat steps and observe you have been scrolled top, a more predictable behavior.

### Other information:

### DIFF
--- a/assets/components/src/with-wizard-screen/index.js
+++ b/assets/components/src/with-wizard-screen/index.js
@@ -1,19 +1,23 @@
 /**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+/**
  * WordPress dependencies.
  */
+import { useEffect } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
 import { Button, Handoff, Notice, TabbedNavigation, WizardPagination } from '../';
+import Router from '../proxied-imports/router';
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described, nice and simple solution, and much better than the current behavior. One tiny nitpick I have is that when hitting the browser's "forward" or "back" buttons, I would expect the scroll position to _not_ change, but I understand that this would require some overengineering in a single-page app.

miguelpeixe:
Thank you @dkoo! Good point! This is indeed the ideal behavior and react-router even has some [comments and suggestions over this](https://reactrouter.com/web/guides/scroll-restoration/generic-solution). I also don't think this is too much of an issue for now since we don't have too long scrolls at our wizards.

dkoo:
> This is indeed the ideal behavior and react-router even has some comments and suggestions over this. I also don't think this is too much of an issue for now since we don't have too long scrolls at our wizards.

Ah, that's interesting, thanks for sharing. I agree that it's not a big deal for our app!

PR #1165, opened 2021-09-28
fix: react router crash from useLocation in HOC

dkoo:
### All Submissions:

Avoids the use of the `useLocation` hook from #1157 in order to avoid an occasional crash due to using a hook inside a higher-order component.

1. On `master`,  navigate to `https://<siteurl>/wp-admin/admin.php?page=newspack-setup-wizard#/settings` and observe a JS crash with the following error: https://reactjs.org/docs/error-decoder.html/?invariant=321

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons: 1. You might have mismatching versions of React and the renderer (such as React DOM) 2. You might be breaking the Rules of Hooks 3. You might have more than one copy of React in the same app See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
```

2. Check out and build this branch, refresh, confirm that the crash no longer occurs.
3. Navigate through the various Newspack wizard screens and confirm no crashes occur and navigation still works as expected. Also confirm that the testing steps from #1157 are still valid.

### Other information:

### DIFF
--- a/assets/components/src/with-wizard-screen/index.js
+++ b/assets/components/src/with-wizard-screen/index.js
@@ -17,7 +17,7 @@ import Router from '../proxied-imports/router';
 import { buttonProps } from '../../../shared/js/';
 import './style.scss';
 
-const { useLocation } = Router;
+const { withRouter } = Router;
 
 /**
  * Higher-Order Component to provide plugin management and error handling to Newspack Wizards.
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Good catch! Fix the issue nicely 🙌

PR #1170, opened 2021-09-30
revert scroll restoration in wizards

adekbadek:
### All Submissions:

#1157 introduces a but in the WPCOM authentication flow (which #1165 does not fix). This PR reverts both changes, since the impact of the bug is severe and the feature introduced is not crucial. 

The scroll restoration added in #1157 might have to be implemented at a different level. The issue with the existing approach seems to be (didn't have time to investigate definitely) the assumption that any component wrapped with `withWizard` HOC will return a router. This is not true in the case of [Support wizard during authentication flow](https://github.com/Automattic/newspack-plugin/blob/fa45f45e4cd58e7d79cc88b018f4cc55a923c844/assets/wizards/support/index.js#L66).

1. On `master`, reset WPCOM auth and try to authenticate by visiting the Support wizard
2. Observe WP admin crashing on redirect
3. Switch to this branch, observe authentication flow working as expected.

### Other information:

### DIFF
--- a/assets/components/src/with-wizard-screen/index.js
+++ b/assets/components/src/with-wizard-screen/index.js
@@ -17,7 +17,7 @@ import Router from '../proxied-imports/router';
 import { buttonProps } from '../../../shared/js/';
 import './style.scss';
 
-const { withRouter } = Router;
+const { useLocation } = Router;
 
 /**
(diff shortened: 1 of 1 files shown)


dkoo (approved):
👍

PR #1173, opened 2021-09-30
feat/fivetran

adekbadek:
### All Submissions:

Adds an integration with [Fivetran](fivetran.com/), via a proxy.

1. Visit the Connections wizard, observe that there's only the WPCOM connection (and Google if configured)
2. Add the `NEWSPACK_FIVETRAN_PROXY` environment variable 
3. Reload the site, observe that there's a "Fivetran" section
4. Connect the services, observe the cards displaying the connector state (e.g. "connected, scheduled") after connection

### Other information:


includes/oauth/class-fivetran-connection.php
```diff
@@ -0,0 +1,237 @@
...
+	 *
+	 * @param string $path Path to append to base URL.
+	 */
+	private static function get_proxy_url( $path = '' ) {
+		if ( ! self::is_fivetran_configured() ) {
+			return false;
+		}
+		return add_query_arg(
+			[
+				'wpcom_access_token' => urlencode( base64_encode( WPCOM_OAuth::get_access_token() ) ),
```
miguelpeixe, line 88: Must check if value returned from `WPCOM_OAuth::get_access_token()` is not a `WP_Error`

miguelpeixe (changes requested):
If WPCom is not connected I get the following error on page load:

Captura de Tela 2021-10-15 às 16 05 03

I’d expect it wouldn’t attempt to connect without user request or interaction.

The Mailchimp, GAM and Facebook connections apparently succeed but don't immediately change the empty (`-`) status on the UI. It only showed as connected after a few refreshes (a few minutes in the case of Facebook). If the request has been made but the API is still processing, can we display an intermediate status?

Also, will there be a `Disconnect` button?

For reference, this is the resulting service, status and label relation after connecting all services:

image

PR #1192, opened 2021-10-19
feat(prompt settings): add prompt new placements and size options

kariae:
### All Submissions:

https://github.com/Automattic/newspack-popups/pull/624 adds new placements for overlay prompts and
size option, we need to add them to the settings wizard

1. If https://github.com/Automattic/newspack-popups/pull/624 is not merged, please pull it.
2. Checkout this PR.
3. Add a new overlay prompt, and save it.
4. From the campaigns list, click on the `prompt settings` icon next to the new prompt.
5. Observe that you can set the new placements (top-left, top-right, ...) and the overlay size.
6. Change the overlay placement and size, and save the settings.
7. Edit the prompt, and notice that the new placement and size were saved from the previous step.

### Other information:


assets/wizards/popups/utils.js
```diff
@@ -47,9 +45,15 @@ export const isCustomPlacement = popup => {
 export const isInline = prompt => ! isOverlay( prompt );
 
 const placementMap = {
-	center: __( 'Center Overlay', 'newspack' ),
 	top: __( 'Top Overlay', 'newspack' ),
+	'top-left': __( 'Top Left Overlay', 'newspack' ),
```
dkoo, line 49: Minor nitpick that applies to all of these new properties:

```suggestion
	'top_left': __( 'Top Left Overlay', 'newspack' ),
```

For JS properties, it's best to use an `_` or even just camelcase as you can't use dot notation (`objectName.property_name`) if the property name contains a hyphen. Since we already have `above_header` I think we can follow that format for all multi-word properties.

PR #1202, opened 2021-10-22
feat: integrate fluid to ad unit size control

miguelpeixe:
### All Submissions:

The current UI for editing ad unit sizes separates the fluid selection from the dropdown of sizes. Although it's handled differently on the API, it shares the same effect on the ad unit, which is registering a new size to allow creatives to display.

This PR changes the `AdUnitSizeControl` to accommodate having `fluid` as a selectable option and removes the checkbox.

1. Check out this PR and edit an Ad Unit on the Ads wizard
2. Make sure you can perform the following edits smoothly:
    1. Create a new size, select it as fluid
    2. Change the size to a fixed one
    3. Create another size and select it as fluid
    4. Remove the fluid size
    5. Create yet another size, select it as fluid, confirm you can change again to Custom and modify the values
3. Confirm that values already selected are not available on the dropdown options
4. Save an ad unit with multiple sizes (including fluid), reload, and confirm the sizes are as expected

### Other information:


thomasguillot:
Brilliant! Frankly we could just merge it as is. But I did notice one tiny issue:
Screenshot 2021-11-05 at 17 34 44
When adding a new size, I shouldn't be able to add the same size multiple times since it's redundant.

thomasguillot:
I'm happy the Design part 🙂

assets/wizards/advertising/components/ad-unit-size-control/index.js
```diff
@@ -29,50 +34,73 @@ export const DEFAULT_SIZES = [
...
+			JSON.stringify( value ) === JSON.stringify( size ) ||
+			! selectedOptions.find(
+				selectedOption => JSON.stringify( selectedOption ) === JSON.stringify( size )
+			)
+	);
+	const sizeIndex = isCustom
+		? -1
+		: options.findIndex(
+				size => value === size || ( size[ 0 ] === width && size[ 1 ] === height )
+		  );
```
adekbadek, line 57: I have a feeling this logic is overly complex, and it is hard to follow. I see a following bug but at first glance it's hard to find the culprit:

1. Start with an ad unit with two sizes: one fixed from the list and fluid
2. Change the fluid one (second size) to custom - observe that the fixed one can't be set to fluid (I expected it to be, so I can have a `[custom, fluid]` sizes list – correct me if that's a wrong assumption)
3. Delete the fixed size (first size) – the second size, which was custom, changes to fluid (I expected it possible to have just one size of custom value)

adekbadek (changes requested):
1.
> Confirm that values already selected are not available on the dropdown options

Not exactly true – after adding multiple sizes, without changing anything else, I end up with a list of same sizes. Each has the size of the first option.

image

assets/wizards/advertising/components/ad-unit-size-control/index.js
miguelpeixe: b55ef8f78b48efc453e68c49cd2c21cedfe7121d improves the readability of the `findIndex` operation.

The problem you were seeing is that the `Custom` selection does not initially trigger `onChange()` to the parent component, keeping the previously selected value until a custom width or height is filled. ed3f475101af75df748ab1fcbc3ee4312eb869d8 changes that behaviour.

miguelpeixe:
da5636bfa6ca88f7d1c5445befeed6205085c9f0 implements a `getNextAvailableSize()` so the `Add New Size` button doesn't create sizes with repeated values.

assets/wizards/advertising/views/ad-unit/index.js
```diff
@@ -32,12 +31,31 @@ class AdUnit extends Component {
...
+				: adUnitChangesOrKey;
+		onChange( { ...adUnit, ad_service: service, ...adUnitChanges } );
+	}
+
+	getSizeOptions() {
+		const { adUnit } = this.props;
+		const sizes = adUnit.sizes && Array.isArray( adUnit.sizes ) ? adUnit.sizes : [];
+		let sizeOptions = [ ...sizes ];
+		if ( adUnit.fluid ) {
+			sizeOptions = [ ...sizeOptions, 'fluid' ];
```
adekbadek, line 52: This makes the UI shift when a non-last ad unit size is changed to fluid size (this size jumps to the end of the list. This is minor issue though.

adekbadek (approved):

assets/wizards/advertising/views/ad-unit/index.js
miguelpeixe: We would have to send the index of the changed size and create the options from that. I think it's an extra complexity that is not worth the gained UX.

PR #1207, opened 2021-10-29
Parse.ly config manager

claudiulodro:
### All Submissions:

This PR automates some of the Parse.ly setup. The main field we need filled automatically is the Site ID field, then Parse.ly should "just work" on our sites.

The easiest way to test is probably this:
1.  Add the following code snippet somewhere:
```
	$request = new WP_REST_Request( 'POST', '/newspack/v1/plugins/wp-parsely/configure' );
	$response = rest_do_request( $request );
} );
```
2. Refresh the page. Parse.ly will likely crash on this first request because [the way it's built can't really handle if the plugin is activated after plugins have loaded](https://github.com/Parsely/wp-parsely/blob/develop/wp-parsely.php#L45-L51). Not sure if it's worth worrying about it, as it crashes after the configuration manager has run and seems like it needs a tricky refactor to solve the underlying issue. 🙁 
3. Parse.ly plugin should have installed and activated. Go to Settings > Parse.ly and observe the Site ID field is populated with the hostname:
Screen Shot 2021-10-29 at 9 54 08 AM

### Other information:

### DIFF
--- a/includes/configuration_managers/class-configuration-managers.php
+++ b/includes/configuration_managers/class-configuration-managers.php
@@ -64,6 +64,10 @@ class Configuration_Managers {
 			'filename'   => 'class-wordpress-seo-configuration-manager.php',
 			'class_name' => 'WordPress_SEO_Configuration_Manager',
 		],
+		'wp-parsely'            => [
+			'filename'   => 'class-parsely-configuration-manager.php',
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
👌

PR #1208, opened 2021-10-29
feat: newspack plugin settings component

miguelpeixe:
### All Submissions:

Implements a generic `PluginSettings` component that handles the standardized settings structure implemented by https://github.com/Automattic/newspack-ads/pull/213 and updates the **Placement** wizard page into a more generic **Global settings** page.

1. Check out this PR and https://github.com/Automattic/newspack-ads/pull/213, then visit Advertising -> Global settings
2. Scroll down and observe the default values displayed
3. Update those values, refresh page
4. Confirm values are updated.

### Other information:


dkoo (changes requested):
This is working well, except for a couple of minor things I ran into below:

assets/wizards/advertising/views/settings/settings-section/index.js
```diff
@@ -0,0 +1,105 @@
...
+				return TextControl;
+		}
+	};
+	const getControlType = setting => {
+		switch ( setting.type ) {
+			case 'int':
+			case 'integer':
+			case 'float':
+			case 'number':
+				return 'number';
```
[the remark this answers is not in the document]
miguelpeixe: This is true! As I went to implement a `<Notice />` on the section using `isNaN()` validation for `number` values, I learned that even though the string value is filled and visible on the input, it returns an empty string on the `onChange()` callback. Could this be an issue with the `TextControl` component from `@wordpress/components`?

assets/wizards/advertising/views/settings/index.js
```diff
@@ -144,9 +210,21 @@ class Placements extends Component {
 						/>
 					) : null }
 				</ActionCard>
+				<SectionHeader
+					title={ __( 'General Settings', 'newspack' ) }
+					description={ __( 'Configure display and advanced settings for your ads.', 'newspack' ) }
+				/>
+				{ Object.keys( settings ).map( sectionKey => (
+					<SettingsSection
```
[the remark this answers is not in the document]
miguelpeixe: The issue was on the https://github.com/Automattic/newspack-ads/pull/213 side. Deactivating was attempting to `update_option()` with `$value` `false`, which does not work for creating the option and is a [known bug](https://core.trac.wordpress.org/ticket/40007). https://github.com/Automattic/newspack-ads/pull/213/commits/81f6fa829f9d5d35e89203a33862d11d3172b578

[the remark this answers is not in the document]
dkoo: Works now, thanks!

assets/wizards/advertising/views/settings/settings-section/index.js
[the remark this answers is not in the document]
dkoo: This seems to be the behavior for `input` with `type="number"` in HTML in general, at least when I tested in both Chrome and FF. As such I think casting the empty string to `0` is probably fine.

dkoo (approved):
👍

miguelpeixe:
Thanks, @dkoo! I'll wait for @adekbadek review as well since he also requested to take a look :)

adekbadek (commented):
I just wanted to suggest moving the settings management code (fetching & updating) to this new `SettingsSection` component. If we're standardising (https://github.com/Automattic/newspack-ads/pull/213) a sort of "Newspack Settings API", then it makes sense to give this component more responsibility. It could just receive the settings' REST API endpoint as prop and take it from there.

miguelpeixe:
I really like this idea @adekbadek !

cf67e9286a8e31eac2e9c22296fa96ce4a4535a7 implements a new `PluginSettings` component, which is implemented to the Ads wizard with the following:

```jsx
<PluginSettings
  pluginSlug="newspack-ads"
  title={ __( 'General Settings', 'newspack' ) }
  description={ __( 'Configure display and advanced settings for your ads.', 'newspack' ) }
/>
```

I didn't move all of its logic inside the `SettingsSection` component because it handles the section of a broader scope of settings. Instead, the `PluginSettings` will be responsible for handling API calls and state management.

This change required 2 small updates because we are no longer using the `wizardApiFetch()` helper function:
 - Error handling, with a general `<Notice />` above sections
 - `inFlight` state management, which currently only disables inputs (can be improved but I think its ok for now)

I'd like your thoughts on this @adekbadek and @dkoo .

miguelpeixe:
a980833539d97cbc6a44d2d69c40664c2a2ad284 implements another small change to allow a partial update of a section since the settings API does not require the entire section data to be sent.

We are using the `data` parameter of the section update for the "section/module activation" and for this action to also trigger an update of unsaved changes would be unexpected behavior.

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -0,0 +1,101 @@
...
+		}
+		switch ( setting.type ) {
+			case 'checkbox':
+			case 'boolean':
+				return CheckboxControl;
+			default:
+				return TextControl;
+		}
+	};
+	const getControlType = setting => {
```
adekbadek, line 35: No need for this function to be defined in the component – can be moved out for better readability & performance.

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -0,0 +1,101 @@
...
+import { ActionCard, Grid, Button, TextControl, CheckboxControl, SelectControl } from '../';
+
+const SettingsSection = ( { settings, disabled, onChange, onUpdate } ) => {
+	const section = settings.find( setting => ! setting.key || setting.key === 'active' );
+	if ( ! section ) {
+		return null;
+	}
+	const activation = settings.find( setting => setting.key === 'active' );
+	const fields = settings.filter( setting => setting.key && setting.key !== 'active' );
+	const getControlComponent = setting => {
```
adekbadek, line 23: No need for this function to be defined in the component – can be moved out for better readability & performance.

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -0,0 +1,101 @@
...
+ */
+import { ActionCard, Grid, Button, TextControl, CheckboxControl, SelectControl } from '../';
+
+const SettingsSection = ( { settings, disabled, onChange, onUpdate } ) => {
+	const section = settings.find( setting => ! setting.key || setting.key === 'active' );
+	if ( ! section ) {
+		return null;
+	}
+	const activation = settings.find( setting => setting.key === 'active' );
+	const fields = settings.filter( setting => setting.key && setting.key !== 'active' );
```
adekbadek, line 22: The logic with `'active'` value of `setting.key` is pretty confusing. I would not readily understand how does `settings` prop shape look like (an array with objects with `key` key which must be `'active'` in order for the setting to be displayed?) if I wanted to use this component.
Off the top of my head, looking at the UI, I'd imagine `SettingsSection` taking following props (aside from read/write handlers):
- active state
- title
- subheader
- a list of fields


image

I'd suggest moving the settings-wrangling logic to the parent component, which deals with the [proposed API](https://github.com/Automattic/newspack-ads/pull/213).

assets/components/src/plugin-settings/index.js
```diff
@@ -0,0 +1,119 @@
...
+		return (
+			this.state.settings[ sectionKey ]?.reduce( ( map, setting ) => {
+				map[ setting.key ] = setting.value;
+				return map;
+			}, {} ) || {}
+		);
+	};
+
+	handleSettingChange = sectionKey => ( key, value ) => {
+		const sectionSettings = [ ...this.state.settings[ sectionKey ] ];
```
adekbadek, line 53: `this.state.settings[ sectionKey ]` is already an array, why should it be coerced into array like so?

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -0,0 +1,101 @@
...
+			disabled={ disabled }
+			title={ section.description }
+			description={ section.help }
+			toggleChecked={ activation ? activation.value : null }
+			hasGreyHeader={ !! activation }
+			toggleOnChange={ value => onUpdate( { [ activation.key ]: value } ) }
+		>
+			{ ( ! activation || activation.value ) && (
+				<Fragment>
+					<Grid columns={ fields.length === 3 ? 3 : 2 } gutter={ 32 }>
```
adekbadek, line 78: Suggestion: with a single field or with six fields two columns will be rendered (while it seems one and three would be intended). Modulo should rather be used (but only for >1 item counts.)

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -0,0 +1,101 @@
...
+ */
+import { Fragment } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { ActionCard, Grid, Button, TextControl, CheckboxControl, SelectControl } from '../';
+
+const SettingsSection = ( { settings, disabled, onChange, onUpdate } ) => {
```
adekbadek, line 16: There's already a [`Settings.SettingsSection` component](https://github.com/Automattic/newspack-plugin/blob/47237490b34e190d744b2333b7483e2805eb7c84/assets/components/src/settings/SettingsSection.js) in this repository, but it seems unused in favour of `SettingsCard`. _If_ this component was less tied to the plugin settings API (see the other comment below), it could replace the `Settings.SettingsSection` component, so it could be used more widely.

adekbadek (changes requested):

assets/components/src/plugin-settings/index.js
```diff
@@ -0,0 +1,119 @@
...
+			inFlight: false,
+			settings: {},
+			error: null,
+		};
+	}
+
+	fetchSettings = () => {
+		const { pluginSlug } = this.props;
+		this.setState( { inFlight: true } );
+		apiFetch( { path: `/${ pluginSlug }/v1/settings-list` } )
```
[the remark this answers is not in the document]
miguelpeixe: Done in https://github.com/Automattic/newspack-ads/pull/220 and 184a8161d6f0d88b944935285602565b00811846

assets/components/src/plugin-settings/index.js
```diff
@@ -0,0 +1,119 @@
...
+import { Component, Fragment } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies.
+ */
+import { SectionHeader, Notice } from '../';
+import SettingsSection from './SettingsSection';
+
+class PluginSettings extends Component {
```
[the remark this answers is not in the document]
miguelpeixe: Why should we avoid class components?

assets/components/src/plugin-settings/SettingsSection.js
miguelpeixe: Good call! Done in 927c817eff4eecd0890d8273935d63a5d65e8337

assets/components/src/plugin-settings/SettingsSection.js
miguelpeixe: Much agreed! Done in fb2b80d04188e0287f7e8e31e104de485ecd27e9

assets/components/src/plugin-settings/index.js
miguelpeixe: This is not meant for typecasting, the spread operator is commonly used for cloning. From [react docs](https://reactjs.org/docs/react-component.html#state):

> Never mutate this.state directly, as calling setState() afterwards may replace the mutation you made. Treat this.state as if it were immutable.

assets/components/src/plugin-settings/SettingsSection.js
miguelpeixe: The `PluginSettings.Section` component is more abstract now, since fb2b80d04188e0287f7e8e31e104de485ecd27e9.

I believe we can rearrange/deprecate components in a different PR. WDYT?

assets/components/src/plugin-settings/SettingsSection.js
miguelpeixe: Note that the sections are not necessarily modules with activation. Therefore, the `active` property can be `null` and the component should be able to handle it as well.

assets/components/src/plugin-settings/index.js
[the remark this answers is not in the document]
adekbadek: Good question! Mainly for consistency, simplicity, and readability; but also React as a project is embracing a more functional style (thinking here about hooks, mainly). I can't think of any reason to use class-based components: it's more code to write (and read) and more of a cognitive load to follow what happens in the code – but this might be personal preference.

assets/components/src/plugin-settings/SettingsSection.js
miguelpeixe: Update: ce9d1ccad8e370d5d0a6abdae017d9ee2cdf991f handles a remaining `fields` logic left on the abstract component.

assets/components/src/plugin-settings/index.js
adekbadek: Ah, got it. One of the caveats of class-based React ;)

assets/components/src/plugin-settings/SettingsSection.js
adekbadek: Better to do it here IMO, but it's non-blocking - so as you prefer 👌

assets/components/src/plugin-settings/index.js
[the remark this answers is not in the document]
miguelpeixe: Makes sense. Although I like the functional style, I still think class components improve readability and enforce good practices when a component contains business logic, which is usually called container components.

This is likely personal preference but also changes according to the rest of the stack. A redux-based application probably shouldn't have business logic inside react components and should favor function components to encourage simplification and the overall presentational aspect.

assets/components/src/plugin-settings/index.js
adekbadek: Ah, got it. One of the caveats of class-based React ;)

adekbadek (approved):

assets/components/src/plugin-settings/SettingsSection.js
miguelpeixe: I think there's already a lot going on in this PR 😅 
I'd rather have it as another PR, in case further discussion is required we'll have clean space for it.

miguelpeixe:
Thank you for the review @adekbadek! 🎉

PR #1217, opened 2021-11-05
feat(google-oauth): display notice if there's no refresh token

adekbadek:
### All Submissions:

~Google OAuth will not issue a refresh token if the OAuth app has already been authorised with another site.~ This is not true (see [comment](https://github.com/Automattic/newspack-plugin/pull/1217#issuecomment-963943790)). But still there might be cases when the refresh token is missing. 

1. Configure a site to use Google OAuth (`NEWSPACK_GOOGLE_OAUTH_PROXY` environment variable)
2. [Revoke](https://myaccount.google.com/permissions) the previous authentication, if present
3. Complete the OAuth flow in the Connections wizard, observe no error notice is displayed
4. ~Complete the OAuth flow on _another site_~  _Disable [this line](https://github.com/Automattic/newspack-plugin/blob/f05cc22d6f59f892ec4705be855a72accca31e1d/includes/oauth/class-google-oauth.php#L411) to mimick a refresh-token-less state_, observe an error notice prompting to revoke authentication displayed in the Connections wizard

### Other information:


adekbadek:
It turned out that there is a setting in the OAuth flow which enables [forcing refresh token issuance](https://stackoverflow.com/a/65702100/3772847). Our proxy has been updated, so this should not be possible anymore – but we can still get that PR in:
- refresh token might be missing for _some_ reason
- someone uses another proxy which does not force refresh token issuance

assets/wizards/connections/views/main/google.js
```diff
@@ -60,6 +60,25 @@ const GoogleOAuth = ( { setError, canBeConnected } ) => {
 	const [ inFlight, setInFlight ] = useState( false );
 	const handleError = res => setError( res.message || __( 'Something went wrong.', 'newspack' ) );
 
+	useEffect( () => {
+		if ( isConnected && ! userBasicInfo.has_refresh_token ) {
```
dkoo, line 64: Whenever I make changes to the `_newspack_google_oauth` option, when I `console.log ( userBasicInfo )` I see `undefined`, so `isConnected` is always false due to the lack of an `email` property.

dkoo (changes requested):
Perhaps I'm testing this wrong, but I can't get the error message to show up.

1. I completed authorization via OAuth
2. I tried editing the `_newspack_google_oauth` value to remove the `refresh_token` value
3. I tried updating the `_newspack_google_oauth` value to gibberish
4. I tried deleting the `_newspack_google_oauth` option entirely

In all cases, when visiting the Connections wizard, Google is simply shown as unconnected, with no error message.

adekbadek:
@dkoo 

> I tried editing the _newspack_google_oauth value to remove the refresh_token value

There must've been an issue with option encoding – the goal is to have an auth token without the refresh token.

> I tried updating the _newspack_google_oauth value to gibberish
> I tried deleting the _newspack_google_oauth option entirely

Both of these would not result in the intended error.

I've updated the PR description with a better way to trigger the error.

dkoo (changes requested):
@adekbadek Thanks for the updated testing steps! I can confirm that the error message appears when there's an access token but no refresh token.

Screen Shot 2021-11-18 at 11 25 07 AM

Note that I had to merge `master` manually in order to get this branch to build. After doing that I got a JS console error in the Connections page:

```
ReferenceError: can't access lexical declaration 'isConnected' before initialization
```

This is fixed locally by moving `isConnected` back above the `useEffect` calls, as another PR seems to bump it below them, resulting in the above error. I'd recommend doing another merge from `master` and making sure the branch can build and no errors occur before we merge back to `master`.

adekbadek:
@dkoo – Done. I've verified that `isConnected` is in the correct spot.

dkoo (approved):
🙌

PR #1221, opened 2021-11-09
Update button-card and button-group focus/pressed style

thomasguillot:
### All Submissions:

This PR simplifies the focus/pressed style of the `ButtonCard` and `ButtonGroup`, removing the `$primary-050` background colour and simply using a tick `$primary-500` border

1. Switch to this branch
2. Go to the Dashboard or the Components Demo screen
3. Test the various `ButtonCard`s
4. Go to Site Design
5. Test the various `ButtonGroup`s

### Other information:

### DIFF
--- a/assets/components/src/button-card/style.scss
+++ b/assets/components/src/button-card/style.scss
@@ -17,8 +17,20 @@
 	transition: background-color 125ms ease-in-out, border-color 125ms ease-in-out,
 		box-shadow 125ms ease-in-out;
 
-	&:active,
 	&:focus,
+	&.is-pressed {
+		background: white;
+		border-color: $primary-500;
+		box-shadow: inset 0 0 0 2px $primary-500 !important;
+		outline: none !important;
+		position: relative;
+	}
+
+	&.is-pressed:focus-visible {
+		box-shadow: inset 0 0 0 2px $primary-500, 0 0 0 1px white, 0 0 0 2px $primary-500 !important;
+	}
+
+	&:active,
 	&:hover {
 		background-color: $gray-100;
 		color: $primary-600;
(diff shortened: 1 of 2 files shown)


assets/components/src/button-group/style.scss
```diff
@@ -23,12 +23,17 @@
 		&.is-pressed {
 			position: relative;
 			z-index: 2;
-			background-color: $primary-050 !important;
+			background-color: white !important;
```
dkoo, line 26: Same question as with ButtonCard—would there ever be a scenario where `white` is an inappropriate background color?

That said, I couldn't actually find any scenario in our code that uses an `is-pressed` class on a Newspack button component, so I wonder if we even need these styles?

assets/components/src/button-card/style.scss
```diff
@@ -17,8 +17,20 @@
 	transition: background-color 125ms ease-in-out, border-color 125ms ease-in-out,
 		box-shadow 125ms ease-in-out;
 
-	&:active,
 	&:focus,
+	&.is-pressed {
+		background: white;
```
dkoo, line 22: It seems like the ButtonCard component is only used in the admin, so this _should_ be fine—is there any scenario where the the `white` might clash with another color that's configurable?

dkoo (approved):
Looks fine in my test site, but I did want to make sure setting the background to `white` in these cases isn't going to cause any issues?

assets/components/src/button-card/style.scss
thomasguillot: At this point, not that I can think of.

assets/components/src/button-group/style.scss
thomasguillot: It's currently not being used but I've planned to use it during the onboarding. And again, I don't think the background-color should be an issue

PR #1225, opened 2021-11-10
Refactor/ads global placements

miguelpeixe:
### All Submissions:

Refactors the UI and remove Ad Placements backend code from the main plugin.

More details on the Ad Placements system refactor are here: https://github.com/Automattic/newspack-ads/pull/219

### Design updates

Along with the refactor, this PR proposes two updates on our components base:

#### Section loading placeholder

A simplified version of the `.newspack-wizard__is-loading` specific for when a section is loading. It is currently implemented for both sections inside the **Advertising -> Placements** and **Advertising -> Settings** wizard:

image

#### Support for `actionContent`, instead of `actionText`, for an `<ActionCard />`.

Currently used for inserting a `<SelectControl />` as card action:

image

Follow test instructions at https://github.com/Automattic/newspack-ads/pull/219

### Other information:


thomasguillot:
@miguelpeixe I've pushed a few visual changes in https://github.com/Automattic/newspack-plugin/pull/1225/commits/249dccc8eaef499664f965b648937ce6cd2792ff

I also wonder whether "Ad Placements" should have its own menu item and "Global Settings" be simply renamed to "Settings"

miguelpeixe:
Thanks @thomasguillot ! I agree on a different menu item, 129663c39ec33f52989589802dfb22bfaec241a0 implements it.

My initial goal was to not let the Placements configuration get too unnecessarily lengthy and overshadow the Settings area. Now that they belong in different places, I think the way you standardized the action card is the better way to go.

miguelpeixe:
@thomasguillot I noticed you removed the `description` from the action card. If that was on purpose, can you explain a bit more the reasoning?

thomasguillot:
The descriptions were not adding much value and most of the cards didn’t
have any anyway.

miguelpeixe:
Every core placement have a description. The ones currently without are the widget areas. I think we should keep them, even if not for every card. The "Sticky" ad placement is the most problematic, we need to be clear that the Sticky placement is exclusive for mobile and is fixed on the bottom of the page.

thomasguillot:
> [Every core placement has a description](https://github.com/Automattic/newspack-ads/blob/feat/global-placements/includes/class-newspack-ads-placements.php#L211-L240). The ones currently without are the widget areas. I think we should keep them, even if not for every card. The "Sticky" ad placement is the most problematic, we need to be clear that the Sticky placement is exclusive for mobile and is fixed on the bottom of the page.

No problem, let's add it back. Would be great if we could add descriptions for the placements with widget areas then. So it's all consistent.

miguelpeixe:
@thomasguillot done in 1b81ec1fabc5f1b5d16ab16b89682bf6e253518a and https://github.com/Automattic/newspack-ads/pull/219/commits/56e9d3dbdfa0b604aa504dd33e4418e5b21a4da9

assets/wizards/advertising/views/placements/index.js
```diff
@@ -0,0 +1,213 @@
...
+	Notice,
+	SectionHeader,
+	SelectControl,
+	withWizardScreen,
+} from '../../../../components/src';
+
+/**
+ * Advertising management screen.
+ */
+class Placements extends Component {
```
dkoo, line 32: Nitpick since this is a new component: any particular reason to use a class-based component over a functional component?

dkoo (approved):
Works well, and like I said in https://github.com/Automattic/newspack-ads/pull/219 I love the simplification here and the idea of moving the heavy lifting to Newspack Ads.

What are some use cases you had in mind for the new `actionContent` prop in the `ActionCard` component? I notice you didn't actually use it after implementing.

assets/wizards/advertising/views/placements/index.js
miguelpeixe: I might be behind on the latest React best practices because @adekbadek [asked me the same thing here](https://github.com/Automattic/newspack-plugin/pull/1208#discussion_r745454591) 😅 

I find that without a better state management structure that would provide HoCs or custom hooks, the stateful, or container, class-based components are better. Without these utilities I find functional components tend to get a bit messy with all the business logic inside.

Still, I can change that. Especially now that it has been brought up again haha

miguelpeixe:
> What are some use cases you had in mind for the new actionContent prop in the ActionCard component? I notice you didn't actually use it after implementing.

I did, but @thomasguillot had other ideas 😅 . You can see the original idea on the second image of the PR description. Should we revert this now it's not in use?

assets/wizards/advertising/views/placements/index.js
miguelpeixe: 2dc6bbf358dcc4ec82af89f1ead9adc5ae3c15c4 refactors the Placements component

dkoo: It's definitely somewhat subjective, and the React team [does not plan to deprecate or stop support class-based components anytime soon](https://reactjs.org/docs/hooks-faq.html#do-i-need-to-rewrite-all-my-class-components). But they are definitely moving slowly toward a hooks-based future:

> [In the longer term, we expect Hooks to be the primary way people write React components.](https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both)

So while I agree that class-based components can sometimes be a bit more readable and easier to understand than hooks-based functional components, I think it's best that we move toward React's recommendation for _new_ components, at least. There's no real urgency to refactor existing class-based components unless we want to use hooks.

dkoo (approved):
Changes look good!

assets/wizards/advertising/views/placements/index.js
miguelpeixe: Thank you for the links! To be honest, after I refactored it to a functional component I liked the result more than I expected 😅 .

This will be even better after we get #1272 going for every wizard 🎉

PR #1247, opened 2021-11-12
feat: use newspack manager plugin for authentication

adekbadek:
### All Submissions:

Changes how the Google OAuth and Fivetran proxies authenticate. Not by passing a WPCOM token anymore, but with an API key.
Note that this requires corresponding changes in the proxies.

1. Ensure you have the option* set to the correct value
2. Observe the Google and Fivetran connections working as expected
3. Remove the option, observe the aforementioned connections are not available in the Connections wizard
4. Change the option value to an invalid one, observe the proxies return error and disallow connection
5. Disconnect WPCOM. Observe the other services can be connected regardless of WPCOM connection status

\* Name is defined in another part of the software, as `NEWSPACK_MANAGER_API_KEY_OPTION_NAME` env variable

### Other information:

---
  - https://app.asana.com/0/0/1201344639817758

### DIFF
--- a/assets/wizards/connections/views/main/fivetran.js
+++ b/assets/wizards/connections/views/main/fivetran.js
@@ -10,7 +10,7 @@ import apiFetch from '@wordpress/api-fetch';
 /**
  * Internal dependencies
  */
-import { Notice, ActionCard, Button } from '../../../../components/src';
+import { ActionCard, Button } from '../../../../components/src';
 
 /**
  * External dependencies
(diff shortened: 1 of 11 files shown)


miguelpeixe (changes requested):
The `apiFetch()` does not resolve the `inFlight` state in case of error. `setInFlight( false );` should be placed inside the `finally()` method of the promise for both Fivetran and Google.

The following check will probably have to change too, since WPCom connection is no longer required for the other connections.
https://github.com/Automattic/newspack-plugin/blob/master/includes/wizards/class-connections-wizard.php#L90-L97

adekbadek:
> The apiFetch() does not resolve the inFlight state in case of error. setInFlight( false ); should be placed inside the finally() method of the promise for both Fivetran and Google.

Fixed in 2557025dfc7dd839d27d6040311c61b938e292fa

> The following check will probably have to change too, since WPCom connection is no longer required for the other connections.

Updated in 9c5b7a9

miguelpeixe (approved):
🎉

adekbadek:
Note that these changes require updated code on the proxy site. Once changes from this PR are live, the proxy site should be updated.

PR #1263, opened 2021-11-17
chore: update composer lockfile

adekbadek:
Updates composer lockfile, I've merged https://github.com/Automattic/newspack-plugin/pull/1246 too hastily.

### To test

Observe composer packages install nicely (`composer install`)


dkoo (approved):
`composer install` works nicely for me 👍

PR #1267, opened 2021-11-17
fix/deps

adekbadek:
### All Submissions:

- move some packages from dev-dependencies to regular dependencies
- remove `@wordpress/blocks` - not used here

Nothing should change, a smoke test is enough.

### Other information:


package.json
```diff
@@ -22,13 +22,8 @@
     "@typescript-eslint/eslint-plugin": "^5.3.0",
     "@wordpress/api-fetch": "^5.2.4",
     "@wordpress/base-styles": "^4.0.0",
```
noahtallen, line 24: Usually base-styles and api-fetch are also normal dependencies, unless something is different about their use here

adekbadek: Good catch, updated in e45395de55831a59bdb378f6462e86e4a3387d51

noahtallen (approved):
Seems like a good idea to me :)

adekbadek:
I thought it's something wrong with this PR so I've rebased. But the culprit was on `master`. Awaiting https://github.com/Automattic/newspack-plugin/pull/1271 so the CI can pass here.

dkoo:
@dependabot rebase

miguelpeixe (approved):
`npm ci && npm run build` is running without issues, CI tasks are passing and there's no apparent issue navigating through the wizards.

dkoo:
@noahtallen @adekbadek all should be passing now.

PR #1305, opened 2021-11-26
Add/apple pay file

adekbadek:
### All Submissions:

A couple of changes to Stripe integration:

1. Exposing `is_currency_zero_decimal` method – this will be used by the Donate block in `newspack-blocks`
2. Adding necessary configuration for Apple Pay support – again this will be used by the Donate block
3. Fixing an issue where a new payment method was replacing the old one, instead of being added

1. **Payment method updating**
    1. On `master`, make two donations using different [test cards](https://stripe.com/docs/testing#cards) but the same email for the donor
    2. Observe the the donor has just one payment method on Stripe – the latest used card
    3. Switch to this branch and re-do step 1. Observe the donor now has two cards listed as payment methods 
1. **Apple Pay domain.** Configuring Apple Pay on Stripe requires two things: i) hosting a special file on the site, and ii) adding the domain on Stripe:
    1. Visit `<your-site>/.well-known/apple-developer-merchantid-domain-association` and observe that a file is downloaded
    1. A _definitive_ test would only be possible with a live Stripe account – let's skip that for the actual Apple Pay integration (coming up in `newspack-blocks`)

### Other information:


dkoo (changes requested):
The bugfix for multiple Stripe payment methods looks 👍 

When trying to visit `<site_url>/.well-known/apple-developer-merchantid-domain-association` I keep getting a 404, even after refreshing permalinks. This seems to happen before the `init` hook is executed, so I'm not seeing any file download.

adekbadek:
> When trying to visit <site_url>/.well-known/apple-developer-merchantid-domain-association I keep getting a 404, even after refreshing permalinks. This seems to happen before the init hook is executed, so I'm not seeing any file download.

I've installed the plugin from this branch on a fresh site and it [works as expected](https://golden-aardvark.jurassic.ninja/.well-known/apple-developer-merchantid-domain-association) (even before configuring Stripe as the Reader Revenue platform, which might be something to change). Could you try to reproduce the issue on a shareable site?

dkoo:
Ah, on a [fresh site](https://thin-marlin.jurassic.ninja/.well-known/apple-developer-merchantid-domain-association) the URL doesn't 404, but it's also not prompting me to download. Is this acceptable behavior? If so, I'll approve.

adekbadek:
> not prompting me to download

That depends on the browser – in my testing, Chrome just downloads the file while FF & Safari ask. Is that what you see?

dkoo (approved):
Since the 404 and the non-download issues seem related to my local environment and/or browser quirks, these may or may not be actual issues with Apple Pay. Let's move ahead with this and address any issues down the line once we're able to fully test with Apple Pay.

PR #1341, opened 2021-12-06
chore: fix auto-merge worfklow to grant write permissions

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

#1332 implemented an auto-merge workflow to automatically merge minor and patch update PRs from Dependabot. However, this workflow is [broken](https://github.com/Automattic/newspack-plugin/runs/4428099886?check_suite_focus=true) because the workflow does not grant the necessary write permissions to merge PRs. More details here: https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/

This PR updates the workflow to run on the suggested `pull_request_target` event, which grants write permissions to actions, but also only runs if the pull request has a "dependencies" label applied, which would indicate that the PR has been opened by a privileged user (in this case, Dependabot). This should guard against [potential security vulnerabilities](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) that might result in auto-merging arbitrary PRs.

### How to test the changes in this Pull Request:

Can't test, unfortunately—we'll need to wait until Dependabot opens a new minor or patch PR after this PR is merged.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1332, opened 2021-12-01
chore: implement auto-merge for dependabot minor/patch updates

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Should help make Dependabot PRs more automated. Will automatically merge minor and patch update PRs if all CI jobs pass. Major updates will still require a manual review and merge.

h/t @yogeshbeniwal for the idea.

Closes #1323.

### How to test the changes in this Pull Request:

Nothing to test. Once merged, we should be able to see the workflow in action in the Actions tab.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

Issue #1323, opened 2021-11-30
Use Dependabot Auto Merge


yogeshbeniwal:
Looks like everyday there are still quite a number of pull requests that require team member's significant time. Shall give a try to [Dependabot Auto Merge](https://github.com/marketplace/actions/dependabot-auto-merge). Any regression if any can always be caught while testing PRs by developers against master or while releasing new version as master is used as base.

PR #1348, opened 2021-12-08
fix: use access token instead of default token

dkoo:
### All Submissions:

The [auto-merge saga](https://github.com/Automattic/newspack-plugin/issues/1323) continues. #1332 implemented auto-merging for minor and patch PRs from Dependabot, and #1341 fixed write permissions for the auto-merge action. However, something is still [missing](https://github.com/Automattic/newspack-plugin/pull/1347#issuecomment-988379465):

Screen Shot 2021-12-07 at 6 11 17 PM

I believe this error is happening because the workflow uses the default [`secrets.GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication), which by default doesn't grant push permissions. Instead, we need to use a [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) with `public_repo` push permissions that has been created by a user with push access to the repo (in this case me). I've added this as an [encrypted secret](https://github.com/Automattic/newspack-plugin/settings/secrets/actions) to the repository with the name `DEPENDABOT_TOKEN`.

Let's hope this token can be used by the workflow to auto-merge branches as desired.

Can't test, must wait for the bots to do their thing.

### Other information:

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -13,4 +13,4 @@ jobs:
       - uses: ahmadnassri/action-dependabot-auto-merge@v2
         with:
           target: minor # includes patch updates.
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          github-token: ${{ secrets.DEPENDABOT_TOKEN }}


adekbadek (approved):
🤞

PR #1374, opened 2021-12-16
test: refactor unit tests to avoid redefined constant errors

### LINKED ISSUE #1373: WP_UNINSTALL_PLUGIN define error on tests
**Describe the bug**
Tests are failing due to [a known bug on WordPress core](https://core.trac.wordpress.org/ticket/44884). More than one call for the `delete_plugin()` function causes the error because the method defines the constant without verifying if it has already been defined.

**To Reproduce**
Run tests and observe the `Constant WP_UNINSTALL_PLUGIN already defined` error.

dkoo:
### All Submissions:

Refactors a couple of unit tests to avoid a `Constant WP_UNINSTALL_PLUGIN already defined` error that results in a failing test suite:

- Most of these errors can be avoided by containerizing the test executions by setting `$preserveGlobalState = false` and `$runTestInSeparateProcess = true` on a unit test class; together, these will force PHPUnit to reset WP to a default state before running each test.
- The one test that still fails after doing this is `test_plugin_install_uninstall_wporg`, because it attempts to run `uninstall_plugin` after that method has already been run in the `setUp` method. I marked this test as skipped to let the suite pass, but left it in place so we can remove the skip once the bug is fixed in Core.
- Because the `plugin-manager` and `api-plugins-controllers` test classes were both using the AMP plugin for testing, the latter was running into failures from AMP not being uninstalled before its tests execute due to the above changes. Using a different plugin (Jetpack) fixes that error.

Closes #1373.

👇 Unit tests should be passing.

### Other information:

### DIFF
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -728,64 +728,6 @@ class Plugin_Manager {
 	}
 
 	/**
-	 * Patch the core uninstall_plugin function to avoid `Constant WP_UNINSTALL_PLUGIN already defined` warning.
-	 *
-	 * @param string $plugin Path to the plugin file relative to the plugins directory.
-	 *
-	 * @return boolean|void True if a plugin's uninstall.php file has been found and included. Void otherwise.
-	 */
-	public static function uninstall_plugin( $plugin ) {
-		$file = plugin_basename( $plugin );
-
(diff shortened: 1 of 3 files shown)


kariae (approved):
@dkoo nice catch for using two plugins in the tests to avoid `WP_UNINSTALL_PLUGIN` definition issue. The tests are passing now.

PR #1407, opened 2021-12-29
change how we check post validity when preventing page deletion

kariae:
### All Submissions:

`get_post_status` is used to check the validity of posts when preventing accidental page deletion
which time out on some publishers' sites. This replaces it with `get_posts` (used internally in
`get_post_status`) but does the work in this case.

It's a bit hard to reproduce this as it was raised on some publishers sites (e.g. CCP) on the users' list page. `users.php` time out due to the use of `get_post_status` to get the validity of a post when checking user capabilities.

Replacing `get_post_status` by `get_post` fix the timeout. Making sure the `users.php` loads correctly should be enough normally to test this. 

### Other information:

### DIFF
--- a/includes/class-patches.php
+++ b/includes/class-patches.php
@@ -81,7 +81,7 @@ class Patches {
 		$post_id            = $args[0]; // First item is usually the post ID.
 
 		// If $post_id isn't a valid post, bail early.
-		if ( false === get_post_status( $post_id ) ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Couldn't replicate the timeout, but was able to confirm that this change still prevents deletion of essential pages, so I think this is 👍

kariae:
Thanks @dkoo 🙏

PR #1424, opened 2022-01-05
fix(ads): ad unit error handling

miguelpeixe:
### All Submissions:

Closes # .

1.
2.
3.

### Other information:

### DIFF
--- a/assets/wizards/advertising/index.js
+++ b/assets/wizards/advertising/index.js
@@ -50,9 +50,9 @@ class AdvertisingWizard extends Component {
 		this.fetchAdvertisingData();
 	};
 
-	updateWithAPI = requestConfig =>
-		this.props
-			.wizardApiFetch( requestConfig )
+	updateWithAPI = requestConfig => {
+		const fetchPromise = this.props.wizardApiFetch( requestConfig );
+		fetchPromise
 			.then(
 				response =>
 					new Promise( resolve => {
@@ -74,6 +74,8 @@ class AdvertisingWizard extends Component {
 					} )
 			)
 			.catch( this.props.setError );
+		return fetchPromise;
+	};
 
 	fetchAdvertisingData = ( quiet = false ) =>
 		this.updateWithAPI( { path: '/newspack/v1/wizard/advertising', quiet } );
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
This works, however, the UI only works as expected because the `saveAdUnit` method doesn't have a `catch` statement, which causes an "uncaught error" that prevents the UI from updating when an error happens. We should catch the error, then explicitly:

- Prevent the UI from navigating away from the single `AdUnit` edit screen
- Prevent the UI from updating the display of the ad unit with updates that didn't get applied
  * Right now, for example, if we caught the error, when you update the name of an ad unit and save, but the save results in an error, the name of the ad unit is updated in the Ad Units list even though the update wasn't successful.

This might be a @thomasguillot question and out of this PR's scope, but I'm not a fan of the "above-header" error message component. It kind of gets lost easily, especially when the UI navigates to a different view at the same time. Should we consider some kind of modal notice or a more prominently placed notice?

miguelpeixe:
The uncaught error on this level of the code is not preventing the UI from updating. The API error is handled in two ways:

1. Prevents route change by attaching it to the `saveAdUnit` promise success callback
2. Error caught and notice handled by the `updateWithAPI()` helper

The change proposed in this PR allows the `saveAdUnit()` callback to attach a route change on success, by changing the return of `updateWithAPI()`. This is also what is causing the uncaught error. We're no longer returning the Promise chained with the default catch, so it has to be handled again.

I now realize that this is not obvious so I'm changing the strategy to make the helper function *rethrow* the error after handling its top-level notice, which is more readable and produces the same desired effect.

If an empty catch (`.catch( () => {} )`) is added to `saveAdUnit()` the current behavior continues without throwing the error.

Updated in 2b076a3f370263edca79974b1e8dc46e68beb645

Is this too weird? I'm trying not to change too much of this architecture now because soon all of these API requests will live inside `newspack-ads` and the UI will end up being adjusted accordingly.

>  I'm not a fan of the "above-header" error message component.

Me too. I think a redesign of the top-level non-fatal wizard error deserves its own issue.

dkoo (approved):
> Is this too weird? I'm trying not to change too much of this architecture now because soon all of these API requests will live inside `newspack-ads` and the UI will end up being adjusted accordingly.

I think I get what you're saying. Since you're already planning to move these requests to the other repo, I think re-throwing the error is fine for now.

> Me too. I think a redesign of the top-level non-fatal wizard error deserves its own issue.

Issue filed here: #1441

miguelpeixe:
Thanks, @dkoo!

PR #1425, opened 2022-01-05
feat(salesforce): check duplicate site status and disable syncs if clone

dkoo:
### All Submissions:

This avoids potential bad syncs if a site with an active Salesforce connection is cloned for staging or development purposes.

Basically, in such a case, if a site with an active Salesforce connection is cloned, any active subscriptions will continue to generate recurring orders, which will then get synced to Salesforce. This could result in bad or duplicate data if both the live site and its clones are sending recurring donation data.

This PR prevents this from happening using the [same mechanism](https://github.com/woocommerce/-unused-woocommerce-subscriptions/blob/master/woocommerce-subscriptions.php#L1099) WooCommerce Subscriptions uses to put itself into [staging mode](https://woocommerce.com/document/subscriptions-handles-staging-sites/).

If you need a dev Salesforce account to test with, please DM me for credentials.

1. You need to simulate the conditions for a clone or staging site. To do so, you can either [set up WooCommerce + Subscriptions and a Salesforce integration on an actual live site](https://newspack.pub/support/reader-revenue/) and clone it, or on your local dev site, add this temporary filter somewhere in your code. This will create a mismatch between the URL reported by WC Subscriptions vs. WP's site URL. (You can add it to the same `class-webhooks.php` file if you need, or your theme's `functions.php`.)

```php
add_filter(
	'wc_subscriptions_site_url',
	function( $url ) {
			return 'https://foo';
	}
);
```

2. In Reader Revenue, set your Platform to Newspack and ensure you have a working Salesforce connection.
3. After this, make a donation on your site and wait a few minutes.
4. Go to WooCommerce > Status > Logs and look in the dropdown for a `webhooks-delivery` log for today. (If you don't see one, wait a bit longer and refresh.)
5. In the log, look for a webhook request with details matching your donation, and verify that the response was a 500 error like this:

```
[Response] => Array
	(
		[Code] => 500
		[Message] => Internal Server Error
		[Headers] => Requests_Utility_CaseInsensitiveDictionary Object
			(
				[data:protected] => Array
					(
						[access-control-allow-headers] => Authorization, X-WP-Nonce, Content-Disposition, Content-MD5, Content-Type
						[access-control-expose-headers] => X-WP-Total, X-WP-TotalPages, Link
						[allow] => POST, PUT, PATCH
						[content-type] => application/json; charset=UTF-8
						[date] => Wed, 05 Jan 2022 17:51:11 GMT
						[link] => <https://newspack-dkoo.ngrok.io/wp-json/>; rel="https://api.w.org/"
						[server] => nginx/1.16.0
						[x-content-type-options] => nosniff
						[x-distributor] => yes
						[x-powered-by] => PHP/7.4.1
						[x-robots-tag] => noindex
					)

			)

		[Body] => {"code":"newspack_salesforce_duplicate_site","message":"This site is a staging or development clone. Third-party syncs will not be processed.","data":null}
	)
```

6. In your connected Salesforce account, verify that the donation data was NOT synced.

### Other information:


includes/class-webhooks.php
```diff
@@ -20,20 +20,44 @@ class Webhooks {
...
+	protected $is_duplicate_site = false;
+
 	/**
 	 * Constructor.
 	 */
 	public function __construct() {
 		$this->salesforce_settings = Salesforce::get_salesforce_settings();
 
 		add_action( 'rest_api_init', [ $this, 'register_api_endpoints' ] );
+		add_filter( 'woocommerce_subscriptions_is_duplicate_site', [ $this, 'check_duplicate_site_status' ] );
```
claudiulodro, line 38: Might be worth bumping the priority on this to `999` or something. That way we're still good if something else hooks into this filter and returns that it's a duplicate site

dkoo: Good idea! Updated in 30f6525f.

dkoo:
Had to refactor this due to conflicts from changes in #1429. @Automattic/newspack-product this could use another review, please!

includes/class-salesforce.php
```diff
@@ -29,11 +29,36 @@ class Salesforce {
...
 	 * Add hooks.
 	 */
 	public static function init() {
 		add_action( 'rest_api_init', [ __CLASS__, 'register_api_endpoints' ] );
+		add_filter( 'woocommerce_subscriptions_is_duplicate_site', [ __CLASS__, 'check_duplicate_site_status' ], 999 );
+	}
+
+	/**
+	 * Uses WooCommerce Subscription's check to determine whether this site is a duplicate/clone site.
+	 * https://github.com/woocommerce/-unused-woocommerce-subscriptions/blob/master/woocommerce-subscriptions.php#L1099
```
adekbadek, line 50: 🤔 

```suggestion
	 * https://github.com/woocommerce/woocommerce-subscriptions/blob/master/woocommerce-subscriptions.php#L1132
```

adekbadek (approved):

PR #1427, opened 2022-01-06
feat(amp): enable disallowing explicitly kept scripts, for debugging

adekbadek:
### All Submissions:

Adds handling of a special parameter, for debugging AMP issues. Specifically, this feature will be useful for debugging potential issues caused by AMP Plus mode, which works by marking certain scripts as "kept" in the AMP plugin.

1. Enable AMP Plus mode* and add the Streamlined Donate block** on a page
2. Visit the page, observe the Donate block behaving as expected
3. Append `?newspack-disallow-amp-kept` to the URL, reload the page
4. Observe the block is not interactive, because JS scripts were not loaded

\* Set `NEWSPACK_AMP_PLUS_ENABLED` env. variable to `true`
\** Set Stripe as the Reader Revenue platform 

### Other information:

### DIFF
--- a/includes/class-amp-enhancements.php
+++ b/includes/class-amp-enhancements.php
@@ -72,6 +72,9 @@ class AMP_Enhancements {
 	 * @param object    $error The AMP sanitisation error.
 	 */
 	public static function amp_validation_error_sanitized( $is_sanitized, $error ) {
+		if ( isset( $_GET['newspack-disallow-amp-kept'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			return true;
+		}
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described! Just to clarify, we return `true` if the `?newspack-disallow-amp-kept` param is on the URL because we want to disable ALL kept scripts, including those "kept" via the AMP plugin UI?

adekbadek:
> Just to clarify, we return true if the ?newspack-disallow-amp-kept param is on the URL because we want to disable ALL kept scripts, including those "kept" via the AMP plugin UI?

Yes, that's correct. This will apply to scripts marked as "kept" by AMP Plus and any other means. This is based on the assumption that manual "keeping" of scripts is very rare, and that if scripts are "kept", it's because of AMP Plus. This provides a way to determine if an issue is caused by anything non-AMP.

PR #1429, opened 2022-01-06
refactor(salesforce): simplify Salesforce API requests

dkoo:
### All Submissions:

Refactors and simplifies Salesforce class methods to avoid code repetition.

This is a pure refactor and shouldn't result in any change in behavior. Smoke test the Salesforce integration (follow instructions on [this page](https://newspack.pub/support/reader-revenue/) to set up a connection, and DM me if you need a Salesforce dev account).

### Other information:


dkoo:
Test failure doesn't seem related to this changeset 🤔

includes/class-salesforce.php
```diff
@@ -28,6 +28,133 @@ class Salesforce {
...
+	 *
+	 * @param WP_REST_Request $request Request containing webhook.
+	 * @return WP_REST_Response|WP_Error The response from Salesforce, or WP_Error.
+	 */
+	public static function api_sync_salesforce( $request ) {
+		$args          = $request->get_params();
+		$order_details = self::parse_wc_order_data( $args );
+
+		if ( empty( $order_details ) ) {
+			return new \WP_Error(
```
miguelpeixe, line 65: Errors should also be returned with `\rest_ensure_response();`

includes/class-salesforce.php
```diff
@@ -28,6 +28,133 @@ class Salesforce {
...
+	 */
+	public static function register_api_endpoints() {
+		// Handle WooCommerce webhook to sync with Salesforce.
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/salesforce/sync',
+			[
+				'methods'             => \WP_REST_Server::EDITABLE,
+				'callback'            => [ __CLASS__, 'api_sync_salesforce' ],
+				'permission_callback' => '__return_true',
```
miguelpeixe, line 49: If this is only used as a WooCommerce webhook, is there a way to ensure this endpoint is not explored by other sources?

miguelpeixe (changes requested):
Smoke test is working and the code is much cleaner now! Left a couple of requests:

includes/class-salesforce.php
dkoo: Good call, updated in 6d4574c4.

includes/class-salesforce.php
dkoo: 395b52b validates the webhook ID sent as part of the request before executing the sync. It also handles both the new Salesforce API namespace (needed for settings-related endpoints in order to use the `PluginSettings` component) as well as the old API namespace.

dkoo:
@miguelpeixe fdb90b6 updates the admin UI to use the `PluginSettings` component. I had to extend this component a bit in 93ab2f1 to add an `afterUpdate` callback option to this component so that I could intercept data from the response in order to redirect the browser for the OAuth flow. Also added an `afterFetch` callback option, which could be similarly useful in other cases. Please let me know if you'd like to handle this in a different way.

includes/class-salesforce.php
```diff
@@ -28,6 +29,544 @@ class Salesforce {
...
+				'value'       => null,
+			],
+			[
+				'description' => __( 'Consumer Key', 'newspack' ),
+				'help'        => __( 'The connected app’s Consumer Key.', 'newspack' ),
+				'section'     => 'salesforce',
+				'key'         => self::SALESFORCE_CLIENT_ID,
+				'type'        => 'string',
+				'default'     => true,
+				'public'      => true,
```
miguelpeixe, line 216: The `public` attribute determines whether that setting value is safe to be available to the public.

miguelpeixe (changes requested):
Cool! 🎉 

The new settings class fits well, just left a tiny comment regarding the `public` setting attribute.

Good call on the `afterUpdate` and `afterFetch` props, will definitely be useful for other cases 👌 

In the near future, we should figure out the best way to have an abstract class for the PHP portion of the setting system. We'll now have Ads and the Salesforce implementation with duplicate efforts for maintaining and improving its functionality.

Thank you for the revisions!

includes/class-salesforce.php
```diff
@@ -28,6 +29,544 @@ class Salesforce {
...
+		}
+		return true;
+	}
+
+	/**
+	 * Check validity of webhook requester before processing the webhook callback handler.
+	 *
+	 * @param WP_REST_Request $request API request object.
+	 *
+	 * @return bool}WP_Error
```
miguelpeixe, line 155: A little typo here!

includes/class-salesforce.php
dkoo: Ah, thanks for clarifying! Changed this to `false` for all settings in bcde8d41.

includes/class-salesforce.php
dkoo: Oops, fixed in bcde8d41

dkoo:
> After pulling these 5 latest commits I had to reconnect Salesforce. Do you think all users will have to reconnect their Salesforce integration after this merge?

I was hoping this wouldn't be the case, but I'll do some testing and report back.

dkoo:
> After pulling these 5 latest commits I had to reconnect Salesforce. Do you think all users will have to reconnect their Salesforce integration after this merge?

I tested this locally and on Model Home by connecting to Salesforce using the latest release plugin, waiting a bit, and then updating the plugin with this branch. I wasn't able to replicate needing to reconnect—in both cases, the new UI showed the "connected" state as expected and the webhooks remained intact.

dkoo:
> In the near future, we should figure out the best way to have an abstract class for the PHP portion of the setting system. We'll now have Ads and the Salesforce implementation with duplicate efforts for maintaining and improving its functionality.

Great idea! Perhaps we could establish a base class in the main plugin that other classes can extend?

dkoo:
> After pulling these 5 latest commits I had to reconnect Salesforce. Do you think all users will have to reconnect their Salesforce integration after this merge?

UDPATE: I was able to replicate this locally, but it doesn't seem related to the update of this branch. Instead, I think the Consumer Secret might be subject to change if you update the Connected App's settings in Salesforce (e.g., if you add a new callback URL to enable the app on a new site). This is probably why I couldn't replicate it before—I tested updating the plugin on my local site, which was already connected, then added Model Home to the Connected App and repeated the test there. Only after completing the test on Model Home and coming back to my local site did I "lose" the connection on my local site. I had to re-paste the Consumer Key and Secret and save settings in order to re-establish the connection. So hopefully this shouldn't happen to publishers who wouldn't update their Connected App settings as part of this refactor, but we should probably put out a notice to the community to check their connections and reconnect, if necessary, after these changes are deployed.

miguelpeixe:
> So hopefully this shouldn't happen to publishers who wouldn't update their Connected App settings as part of this refactor, but we should probably put out a notice to the community to check their connections and reconnect, if necessary, after these changes are deployed.

Thank you for investigating deeper!

> Perhaps we could establish a base class in the main plugin that other classes can extend?

It was my initial thought but it also goes against the idea of Newspack plugins being able to work standalone in the near future. The javascript being available at `newspack-components` makes me wonder if a composer package is suitable for such a feature.

miguelpeixe (approved):

dkoo:
> The javascript being available at `newspack-components` makes me wonder if a composer package is suitable for such a feature.

That's a great idea! 🤔

PR #1450, opened 2022-01-18
feat: use PluginSettings components for Campaigns wizard

dkoo:
### All Submissions:

Refactors the **Campaigns > Settings** wizard page to use the `PluginSettings` component. Requires https://github.com/Automattic/newspack-popups/pull/777.

1. Make sure your Campaigns repo is checked out with https://github.com/Automattic/newspack-popups/pull/777.
2. Visit **Newspack > Campaigns > Settings**.
3. Confirm you see the new settings UI:

Screen Shot 2022-01-21 at 1 49 18 PM

4. Change various settings and save, then refresh the page. Confirm that settings are saved as expected.

### Other information:

### DIFF
--- a/assets/components/src/plugin-settings/SettingsSection.js
+++ b/assets/components/src/plugin-settings/SettingsSection.js
@@ -27,6 +27,8 @@ const getControlComponent = setting => {
 		case 'checkbox':
 		case 'boolean':
 			return CheckboxControl;
+		case 'select':
+			return SelectControl;
 		default:
 			return TextControl;
 	}
@@ -46,6 +48,8 @@ const getControlType = setting => {
 		case 'boolean':
 		case 'checkbox':
 			return 'checkbox';
+		case 'select':
+			return 'select';
 		default:
 			return null;
 	}
@@ -75,7 +79,7 @@ const SettingsSection = ( {
 		options:
 			setting.options?.map( option => ( {
 				value: option.value,
-				label: option.name,
+				label: option.name || option.label,
 			} ) ) || null,
 		value: setting.value,
 		checked: setting.type === 'boolean' ? !! setting.value : null,
--- a/assets/components/src/plugin-settings/SettingsSection.js
+++ b/assets/components/src/plugin-settings/SettingsSection.js
@@ -27,8 +27,6 @@ const getControlComponent = setting => {
 		case 'checkbox':
 		case 'boolean':
 			return CheckboxControl;
-		case 'select':
-			return SelectControl;
 		default:
 			return TextControl;
 	}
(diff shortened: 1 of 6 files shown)


assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -27,6 +27,8 @@ const getControlComponent = setting => {
 		case 'checkbox':
 		case 'boolean':
 			return CheckboxControl;
+		case 'select':
```
miguelpeixe, line 30: This case is not necessary. The use of `<SelectControl />` is determined by the existence of `setting.options`:

```js
	if ( Array.isArray( setting.options ) && setting.options.length ) {
		return SelectControl;
	}
```

It allows for the setting to preserve type validation.

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -46,6 +48,8 @@ const getControlType = setting => {
 		case 'boolean':
 		case 'checkbox':
 			return 'checkbox';
+		case 'select':
```
miguelpeixe, line 51: This is also not necessary since `SelectControl` is a `<select />` element, which doesn't accept the `type` attribute.

miguelpeixe (changes requested):
The settings are working well. I feel that the grouping of multiple checkboxes looks a bit cluttered. Maybe an adjustment of the `help` font size or margin could help? cc @thomasguillot

assets/components/src/plugin-settings/SettingsSection.js
dkoo: Fixed in 4eef95d6

assets/components/src/plugin-settings/SettingsSection.js
dkoo: Thanks for clarifying! Fixed in 4eef95d6

dkoo:
> I feel that the grouping of multiple checkboxes looks a bit cluttered. Maybe an adjustment of the help font size or margin could help?

I agree! Changing the font-size might help a bit, perhaps changing the color too? It still looks a bit off-kilter when the settings labels have different lengths, though:

Screen Shot 2022-01-25 at 10 19 20 AM

I'll defer to @thomasguillot on this, and I think this can be handled in a separate PR if needed too.

miguelpeixe (approved):
Thank you for the revisions!

I agree it can be handled in a different PR. I already see improvement on your screenshot and believe the difference in label length could be resolved by vertically aligning the help text to the label. We also need to see how that works with other controls' help texts.

dkoo:
Thanks for reviewing—I'll hold off on merging until https://github.com/Automattic/newspack-popups/pull/777 is ready to be merged as well.

dkoo:
769d572a updates some method names used to update Campaigns settings sections to avoid confusion with that plugin's standalone/alternative settings UI's methods, based on [this suggestion](https://github.com/Automattic/newspack-popups/pull/777#discussion_r802116322) in Campaigns.

miguelpeixe (approved):

PR #1451, opened 2022-01-18
Feat/1393 update plugin list

### LINKED ISSUE #1393: Update list of managed plugins to match new approved/required plugins list
Now that we've updated our [list of approved plugins](https://docs.google.com/spreadsheets/d/1cOxvYjn57lV2mNU7ExiP1PRJOERDKoyDfO1I5T46DqY/edit#gid=0) and settled on a new review process, we should update the managed plugin list in the Newspack plugin to match.

My recommendation would be for our [managed plugin list](https://github.com/Automattic/newspack-plugin/blob/master/includes/class-plugin-manager.php#L40) to be slimmed down considerably to only include core Newspack plugins (second sheet in the approved plugins list) and the ones marked as "recommended" in the first sheet.

This should help avoid the situation where new users of Newspack install a bunch of plugins we don't necessarily recommend/require and then these plugins either sit around either installed but inactive or, worse, active but not configured in a way that could lead to performance issues/plugin conflicts/etc.

I'm happy to put a PR together to update this list but wanted to make sure the criteria for required/recommended is clear and that this list feels right before we make the update.

Feedback welcome!

aschweigert:
### All Submissions:

Update plugin list to remove the less-recommended plugins. Add Gravity Forms. Make escaping and translation more consistent. Closes #1393 .

1. go to the plugins screen in wp-admin
2. verify that the plugins list is updated, plugins are installable, etc

### Other information:

* [N/A] Have you written new tests for your changes, as applicable?

---
  - https://app.asana.com/0/0/1201069149097200


dkoo (approved):
Nice, pared-down list. Thanks for implementing translation strings, as well!

PR #1467, opened 2022-01-25
fix: tooltip position in header and remove duplicated css

thomasguillot:
### All Submissions:

Since https://github.com/Automattic/newspack-plugin/pull/1457 there's a tiny bug in the header with the "Return to Dashboard"

I also noticed that we were duplicating the same css, not quite sure why. I've removed it and I couldn't see anything broken.

1. Check the header
2. Switch to this branch
3. Check the header again
4. Check all the various the various wizards

### Other information:


dkoo (approved):
Didn't see any issues resulting from removing the extra SCSS file.

This seems to be browser-specific, but there's a very slight hint of the white Newspack logo underneath the hover icon in Firefox:

Screen Shot 2022-02-01 at 12 41 27 PM

Couldn't replicate in Chrome or Safari so I'd say this is non-blocking.

thomasguillot:
> Didn't see any issues resulting from removing the extra SCSS file.
> 
> This seems to be browser-specific, but there's a very slight hint of the white Newspack logo underneath the hover icon in Firefox
> 
> Couldn't replicate in Chrome or Safari so I'd say this is non-blocking.

Well spotted and fixed in https://github.com/Automattic/newspack-plugin/pull/1467/commits/a47646dbfbfd9aa566df7cfd3fae3c2bff5b8d3d

This was due to:
- Retina screen
- Absolute position with -50%
- 1px negative margin (so divided by two you'd get 0.5px)

Basically, the Newspack icon underneath was slightly visible. Adding some extra padding to the Dashboard icon fully hides the Newspack icon

PR #1474, opened 2022-01-27
hotfix: disable author archives for subscribers

dkoo:
### All Submissions:

Avoids displaying author archive pages for WP users that don't have one of the defined allowed roles, by default:

```
[
	'administrator',
	'editor',
	'author',
	'contributor',
]
```

This array of roles is filtered as `newspack_user_roles_with_author_archives` so other plugins can add/remove roles as needed.

Currently, WP will generate author archive pages for all WP users, including subscribers and WooCommerce customers, which most Newspack publishers won't want.

1. Create several users: a Co-Authors Plus guest author, an admin user, an editor user, an author user, a contributor user, and a subscriber user.
2. On `master`, view the author archives for each author type (hint: look at the "View" link in WP admin). Observe that the subscriber gets an author archive like all the others.
3. Check out this branch. Confirm that all author archives still work for the CAP guest author, admin, editor, author, and contributor users, but that the subscriber archive is now a 404.

#### Handle outgoing staff writer scenario

Newspack publishers will often convert WP users for outgoing staff to "subscriber" users so that those users no longer have access to the WP admin. In this case we still want the archives to be accessible for those users. Since it's not possible in WP admin to assign subscribers as post authors, we can assume that any subscriber with a post count greater than 0 falls into this scenario.

1. Assign one of the non-subscriber WP users you created above to at least one post.
2. Edit the user in WP admin and change their role to "Subscriber."
3. View the author archive and confirm that the archive is visible with their post(s).
4. Reassign all of their posts to another user and refresh the archive. Confirm that it's now a 404.

### Other information:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.72.1-hotfix.1](https://github.com/Automattic/newspack-plugin/compare/v1.72.0...v1.72.1-hotfix.1) (2022-01-27)
+
+
+### Bug Fixes
+
+* disable author archive pages for non-staff users ([3ec9fc7](https://github.com/Automattic/newspack-plugin/commit/3ec9fc77fe12d6539560b4939dc9d3897f048935))
+
 # [1.72.0](https://github.com/Automattic/newspack-plugin/compare/v1.71.0...v1.72.0) (2022-01-25)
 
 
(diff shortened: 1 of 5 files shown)


miguelpeixe (approved):
🙌

PR #1453, opened 2022-01-19
Remove integrations from onboarding and add them to connections wizard

thomasguillot:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR removes the "Integrations" step during the onboarding. Since they became optional to carry on with the onboarding I didn't see any reason why we had to keep them here.

Instead, I moved them to the "Connections" wizard. I did some reorganisation and slightly tweaked some of the components.

connections-wizard

### How to test the changes in this Pull Request:

1. Switch to this branch
2. Go through the onboarding
3. Go through the connections

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #1481, opened 2022-02-01
fix(setup-wizard): donation data

adekbadek:
### All Submissions:

Fixes a Setup wizard issue.

Also updates the useWizardData hook, so it can use multiple wizards. Before, the hook was using the Reader Revenue wizard data only.

1. On `master`, visit the Setup wizard, "Services" step
2. Enable "Reader Revenue" and set some donation values (e.g. suggested donation amount)
3. Save by navigating to the next setup step
4. Open the Reader Revenue wizard and observe that the inserted values have not been updated
5. Switch to this branch, redo steps 2 & 3, observe the updated values in Reader Revenue wizard

### Other information:

---
  - https://app.asana.com/0/0/1201745193114476

### DIFF
--- a/assets/wizards/readerRevenue/views/donation/index.js
+++ b/assets/wizards/readerRevenue/views/donation/index.js
@@ -11,13 +11,14 @@ import { MoneyInput } from '../../components/';
 import { Card, Grid, Button, Notice, ToggleControl, Wizard } from '../../../../components/src';
 import { READER_REVENUE_WIZARD_SLUG } from '../../constants';
 
-export const DontationAmounts = ( { data = {} } ) => {
+export const DontationAmounts = () => {
+	const wizardData = Wizard.useWizardData( 'reader-revenue' );
 	const {
 		suggestedAmounts = [ 0, 0, 0 ],
 		suggestedAmountUntiered = 0,
 		currencySymbol = '$',
 		tiered = false,
-	} = data;
+	} = wizardData.donation_data || {};
 	const { updateWizardSettings } = useDispatch( Wizard.STORE_NAMESPACE );
 	const changeHandler = path => value =>
 		updateWizardSettings( {
@@ -73,7 +74,7 @@ export const DontationAmounts = ( { data = {} } ) => {
 };
 
 const Donation = () => {
-	const wizardData = Wizard.useWizardData();
+	const wizardData = Wizard.useWizardData( 'reader-revenue' );
 
 	const { saveWizardSettings } = useDispatch( Wizard.STORE_NAMESPACE );
 	const onSave = () =>
@@ -108,7 +109,7 @@ const Donation = () => {
 					</Card>
 				) }
 				<Card noBorder>
-					<DontationAmounts data={ wizardData.donation_data } />
+					<DontationAmounts />
 				</Card>
 			</Grid>
 			<div className="newspack-buttons-card">
--- a/assets/wizards/readerRevenue/views/donation/index.js
+++ b/assets/wizards/readerRevenue/views/donation/index.js
@@ -11,7 +11,7 @@ import { MoneyInput } from '../../components/';
 import { Card, Grid, Button, Notice, ToggleControl, Wizard } from '../../../../components/src';
 import { READER_REVENUE_WIZARD_SLUG } from '../../constants';
 
-export const DontationAmounts = () => {
+export const DonationAmounts = () => {
 	const wizardData = Wizard.useWizardData( 'reader-revenue' );
 	const {
 		suggestedAmounts = [ 0, 0, 0 ],
@@ -109,7 +109,7 @@ const Donation = () => {
 					</Card>
 				) }
 				<Card noBorder>
-					<DontationAmounts />
+					<DonationAmounts />
 				</Card>
 			</Grid>
 			<div className="newspack-buttons-card">
--- a/assets/wizards/setup/views/services/ReaderRevenue.js
+++ b/assets/wizards/setup/views/services/ReaderRevenue.js
@@ -2,13 +2,11 @@
  * External dependencies
  */
 import classnames from 'classnames';
-import { pick } from 'lodash';
+import { isEmpty } from 'lodash';
 
 /**
  * WordPress dependencies.
  */
-import apiFetch from '@wordpress/api-fetch';
-import { useEffect } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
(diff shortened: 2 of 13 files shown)


assets/wizards/readerRevenue/views/donation/index.js
```diff
@@ -11,13 +11,14 @@ import { MoneyInput } from '../../components/';
 import { Card, Grid, Button, Notice, ToggleControl, Wizard } from '../../../../components/src';
 import { READER_REVENUE_WIZARD_SLUG } from '../../constants';
 
-export const DontationAmounts = ( { data = {} } ) => {
+export const DontationAmounts = () => {
```
dkoo, line 14:
```suggestion
export const DonationAmounts = () => {
```

assets/wizards/readerRevenue/views/donation/index.js
```diff
@@ -108,7 +109,7 @@ const Donation = () => {
 					</Card>
 				) }
 				<Card noBorder>
-					<DontationAmounts data={ wizardData.donation_data } />
+					<DontationAmounts />
```
dkoo, line 112:
```suggestion
					<DonationAmounts />
```

assets/wizards/setup/views/services/ReaderRevenue.js
```diff
@@ -37,15 +29,9 @@ const ReaderRevenue = ( { configuration, onUpdate, className } ) => {
 				</p>
 			) : (
 				<>
-					<DontationAmounts
-						data={ configuration.donation_data || {} }
-						onChange={ donation_data => onUpdate( { donation_data } ) }
-					/>
+					<DontationAmounts />
```
dkoo, line 32:
```suggestion
					<DonationAmounts />
```

assets/wizards/setup/views/services/ReaderRevenue.js
```diff
@@ -2,33 +2,25 @@
...
  * WordPress dependencies.
  */
-import apiFetch from '@wordpress/api-fetch';
-import { useEffect } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
 import { DontationAmounts } from '../../../readerRevenue/views/donation';
```
dkoo, line 15:
```suggestion
import { DonationAmounts } from '../../../readerRevenue/views/donation';
```

dkoo (changes requested):
Ran into a few issues:

1. This isn't really related to this PR, but while we're working with the Services view during setup...if you skip setting up Jetpack in the prior screen, then the Newsletters action card is totally blank:

Screen Shot 2022-02-01 at 11 12 59 AM

Suggestion: Either hide this card if Jetpack isn't connected at this stage, or show a message saying that Jetpack needs to be connected first (Like the `Connect Jetpack in order to configure Mailchimp` message we show in the dashboard wizard page).

2. In the Services page during setup: if you enable some services, fill out some settings, then proceed to the Design step, then go back to Services, the services you previously toggled on are no longer toggled. Expanding them does reveal the settings you previously set are saved, but the "on" state for each service is not saved. Question: is the toggle even necessary for these Services cards? Toggling them on or off doesn't really serve any functional purpose beyond hiding unnecessary settings fields, right? I would ask if we could just omit the toggle controls and show all settings for each service by default. This would remove a step for setting these options during setup (cc @thomasguillot). If we keep the toggle controls, the state should be saved if we go back to the Services page from another step.

3. I was able to replicate this on two test sites in a row. I ran into this JS console error when trying to continue past the Services screen:

```
Object { code: "invalid_json", message: "The response is not a valid JSON response." }
index.js:55:15
```

I had to refresh the page to try again, after which I was able to re-enter my settings and continue, but after finishing setup and going to the Reader Revenue wizard, my settings weren't saved. I had toggled on the "Set exact monthly donation tiers" option and set values for all tiers, but none of these settings were saved. Stripe settings **were** saved, however. Going back to the `admin.php?page=newspack-setup-wizard#/services` page and repeating the steps a second time did not throw the JS console error and did successfully save the settings.

thomasguillot:
> 1. This isn't really related to this PR, but while we're working with the Services view during setup...if you skip setting up Jetpack in the prior screen, then the Newsletters action card is totally blank

Jetpack shouldn't be needed for this. This sounds like a bug to me.

> 2\. In the Services page during setup: if you enable some services, fill out some settings, then proceed to the Design step, then go back to Services, the services you previously toggled on are no longer on. Toggling them on again does reveal the settings you previously set, but the "on" state for each service is not saved. Question: is the toggle even necessary for these Services cards? Toggling them on or off doesn't really serve any functional purpose beyond hiding unnecessary settings fields, right? I would ask if we could just omit the toggle controls and show all settings for each service by default. This would remove a step for setting these options during setup (cc @thomasguillot). If we keep the toggle controls, the state should be saved if we go back to the Services page from another step.

Losing the toggle sounds once again like a bug to me. The idea behind the toggles was that some publishers might not want to enable these services. If you show them a bunch of empty field then they might end up confused with a feeling of being "forced" to use reader revenue, newsletters, etc...

assets/wizards/setup/views/services/ReaderRevenue.js
adekbadek: Done in 7c7b657

assets/wizards/setup/views/services/ReaderRevenue.js
adekbadek: Done in 7c7b657

assets/wizards/readerRevenue/views/donation/index.js
adekbadek: Done in 7c7b657

assets/wizards/readerRevenue/views/donation/index.js
adekbadek: Done in 7c7b657

adekbadek:
> if you skip setting up Jetpack in the prior screen, then the Newsletters action card is totally blank

There's no Jetpack setup, the "Integrations" screen was removed in #1453 🤔 After disabling Jetpack, I still do see the Newsletters section. After disabling Newspack Newsletters, I see a plugin installation prompt. That said, error and loading states were missing – added in ded9f242d7cc5e84b17d7c20de3557a8f042c495

>  In the Services page during setup: if you enable some services, fill out some settings, then proceed to the Design step, then go back to Services, the services you previously toggled on are no longer on. 

Fixed in 9e22307b79f805a30e179a674979eca5489c8e95. The code before tried to be too smart (trying to infer if the service is "active") and there was a disconnect between what determined the "enabled" state and how the "enabled" state was saved after toggling it in the UI. Now the toggle states just save options values.

Note that this change is technically non-backwards-compatible, but since the Setup wizard is meant to be ran once (after completion it's unavailable in the UI), I don't this this change should be marked as breaking.

> I ran into this JS console error when trying to continue past the Services screen

I could not reproduce that locally, also when using production-built JS files. Could you share more about the circumstances of this error (which services are enabled, which plugins active)? Also, does this happen specifically on this branch, or on `master` too?

dkoo:
> There's no Jetpack setup, the "Integrations" screen was removed in https://github.com/Automattic/newspack-plugin/pull/1453 🤔 After disabling Jetpack, I still do see the Newsletters section. After disabling Newspack Newsletters, I see a plugin installation prompt. That said, error and loading states were missing – added in https://github.com/Automattic/newspack-plugin/commit/ded9f242d7cc5e84b17d7c20de3557a8f042c495. With that in place, are you seeing an error with Jetpack disconnected?

Whoops, my mistake. I must have gotten used to just clicking past the Integrations step. 😆 I'm still seeing the same behavior on a new site with the latest changes, so here's a new issue: #1487

dkoo (changes requested):
> >  In the Services page during setup: if you enable some services, fill out some settings, then proceed to the Design step, then go back to Services, the services you previously toggled on are no longer on. 

> Fixed in 9e22307b79f805a30e179a674979eca5489c8e95. The code before tried to be too smart (trying to infer if the service is "active") and there was a disconnect between what determined the "enabled" state and how the "enabled" state was saved after toggling it in the UI. Now the toggle states just save options values.

> Note that this change is technically non-backwards-compatible, but since the Setup wizard is meant to be ran once (after completion it's unavailable in the UI), I don't think this change should be marked as breaking.

Confirmed—the services stay toggled on/off if you go back to that step, now. I agree about the non-breaking nature of the change since pretty much no one who has already set up their site will be going back to this page.

Still seeing the `invalid_json` error when saving non-default Reader Revenue options for the first time. Details sent via DM.

adekbadek:
> Still seeing the invalid_json error when saving non-default Reader Revenue options for the first time. Details sent via DM.

As discussed, the fatal was happening here:

> PHP Fatal error:  Uncaught Error: Call to a member function get_countries() on null in […]wp-content/plugins/newspack-plugin/includes/configuration_managers/class-woocommerce-configuration-manager.php:57

I still can't reproduce that, but I've added a check to ensure WC's countries handler is available in 9f1ce7fd0b98fb81a553c98ab53ae746f2e7b379

Also, found some API-relates issues with sending is-service-enabled payload (the booleans were sent as `"1"`/`""`) – fixed in 7551ab11c96f8b8ad655e7ac18c342d4df2c2326

adekbadek:
Added another check for WC'c countries handlers availability in 2b548544a73e779414dc4baabb1279fda43a3d3b

dkoo (approved):
Looks good after the latest changes! 👍

PR #1483, opened 2022-02-02
feat(stripe): add business location field

adekbadek:
### All Submissions:

Adds a location field for Stripe settings of Reader Revenue Wizard. This is needed for Stripe payment request button (aka Apple/Google Pay). 
Also, as suggested by @thomasguillot, adds reading default values for this new field and the currency field from WooCommerce's settings. 

1. Visit the Reader Revenue wizard, set Stripe as payment platform*
2. Navigate to "Stripe Settings" tab
3. Observe a new select input in the first section – "Where is your business based?", with a default value of United States
4. Change it to something else, reload the page, observe the setting was saved
5. Switch platform to Newspack (which will use WooCommerce) and set country & currency in Reader Revenue wizard, "Address" tab
6. Remove (or temporarily rename) `newspack_stripe_data` option, switch platform back to Stripe
7. Observe the country & currency fields are same as were set in the "Address" tab for Newspack platform

\* Enabling AMP Plus by setting `NEWSPACK_AMP_PLUS_ENABLED` env. var is necessary.

### Other information:


thomasguillot:
the country/currency is set up during the onboarding (but I think for WC), is there any way we can reuse them here?

Screenshot 2022-02-02 at 15 12 40
Screenshot 2022-02-02 at 15 13 19

adekbadek:
> the country/currency is set up during the onboarding (but I think for WC), is there any way we can reuse them here?

Great idea, done!

includes/reader-revenue/class-stripe-connection.php
```diff
@@ -81,14 +81,31 @@ public static function api_permissions_check( $request ) {
...
 	 */
 	public static function get_default_stripe_data() {
+		$location_code = 'US';
+		$currency      = 'USD';
+
+		// The instance might've used WooCommerce. Read WC's saved settings to
+		// provide more sensible defaults.
+		$wc_country          = get_option( 'woocommerce_default_country', false );
+		$valid_country_codes = wp_list_pluck( newspack_get_countries(), 'value' );
+		if ( in_array( $wc_country, $valid_country_codes ) ) {
```
miguelpeixe, line 91: The value coming from `woocommerce_default_country` comes with more specific location code separated by `:`. E.g.: `US:CA` for United States (US) - California and `BR:SP` for Brazil - São Paulo.

We should match the first part of the incoming value.

miguelpeixe (changes requested):
Working as described! Only one request:

includes/reader-revenue/class-stripe-connection.php
adekbadek: Fixed in 173e5c7d69cd48ca5371378522b22dfaeb093742, and added some tests.

miguelpeixe (approved):
👌

PR #1485, opened 2022-02-02
feat(salesforce): ability to manually sync WC orders to salesforce

dkoo:
### All Submissions:

~Not to merged before #1429.~ Ready for review! Extends the WooCommerce admin UI for orders to more conveniently show the orders' sync status with Salesforce, as well as allow admins to manually sync or resync single orders with a button click.

1. Set up a Salesforce connection on your test site (follow instructions on [this page](https://newspack.pub/support/reader-revenue/) to set up a connection, and DM me if you need a Salesforce dev account).
2. Make a test donation as a reader and wait a few minutes.
3. In WP admin > WooCommerce > Orders, find the order corresponding to your test donation.
4. In the Order Actions metabox, verify that you see a new "Salesforce sync status" message with the sync status of your order (will only appear if you have an active Salesforce connection). Also verify that if the order shows a "synced" status, that it's clickable to the Salesforce Opportunity record:

Screen Shot 2022-01-07 at 11 34 19 AM

5. Also observe that on successful or failed sync, a message is added to the "Order notes" metabox (this is to add some slightly more user-friendly sync logging outside of the WooCommerce webhook logs):

Screen Shot 2022-01-07 at 11 34 56 AM

7. In the Order Actions metabox, expand the "Choose an action..." dropdown. Confirm that there's a new action to "Sync order to Salesforce" (will only appear if you have an active Salesforce connection):

Screen Shot 2022-01-07 at 11 36 04 AM

8. Select the new option and click the [>] button to execute the action. For already-synced orders, this will resync the order to the existing Opportunity record in Salesforce. Only billing info and close date will be updated, as the other order data is not updateable in WooCommerce. For orders that have not been synced for whatever reason, it will create a new Opportunity record in Salesforce.
9. Log into Salesforce and confirm that the manual sync action successfully updated or created the opportunity record.

### Other information:

### DIFF
--- a/includes/class-salesforce.php
+++ b/includes/class-salesforce.php
@@ -117,7 +117,9 @@ class Salesforce {
 			[
 				'methods'             => \WP_REST_Server::READABLE,
 				'callback'            => [ __CLASS__, 'api_get_order_status' ],
-				'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
+				'permission_callback' => function() {
+					return current_user_can( 'edit_others_shop_orders' );
+				},
 			]
 		);
 
--- a/includes/class-salesforce.php
+++ b/includes/class-salesforce.php
@@ -334,7 +334,6 @@ class Salesforce {
 	/**
 	 * API endpoint for checking validity of a Salesforce refresh token.
 	 *
-	 * @throws \WP_Error Error message.
 	 * @return WP_REST_Response with the active status.
 	 */
 	public static function api_check_salesforce_connection_status() {
@@ -443,7 +442,6 @@ class Salesforce {
 	 * API endpoint for getting Salesforce API tokens.
 	 *
 	 * @param WP_REST_Request $request Request containing settings.
-	 * @throws \WP_Error Error message.
 	 * @return WP_REST_Response with the latest settings.
 	 */
 	public static function api_get_salesforce_tokens( $request ) {
@@ -452,9 +450,11 @@ class Salesforce {
 
 		// Must have a valid API key and secret.
 		if ( empty( $settings['client_id'] ) || empty( $settings['client_secret'] ) ) {
-			throw new \WP_Error(
-				'newspack_salesforce_invalid_credentials',
-				__( 'Invalid Consumer Key or Secret.', 'newspack' )
+			return \rest_ensure_response(
+				new \WP_Error(
+					'newspack_salesforce_invalid_credentials',
+					__( 'Invalid Consumer Key or Secret.', 'newspack' )
+				)
 			);
 		}
 
@@ -1149,7 +1149,6 @@ class Salesforce {
 	/**
 	 * Get a new Salesforce token using a valid refresh token.
 	 *
-	 * @throws \WP_Error Error message.
 	 * @return string The new access token.
 	 */
 	private static function refresh_salesforce_token() {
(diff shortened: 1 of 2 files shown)


includes/class-salesforce.php
```diff
@@ -106,6 +110,17 @@ public static function register_api_endpoints() {
...
 		);
 
+		// Fetch order sync status.
+		register_rest_route(
+			self::SALESFORCE_API_NAMESPACE,
+			'/order',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => [ __CLASS__, 'api_get_order_status' ],
+				'permission_callback' => '__return_true',
```
miguelpeixe, line 120: I believe this should use the same capability applied to order edit.

includes/class-salesforce.php
```diff
@@ -408,7 +452,10 @@ public static function api_get_salesforce_tokens( $request ) {
 
 		// Must have a valid API key and secret.
 		if ( empty( $settings['client_id'] ) || empty( $settings['client_secret'] ) ) {
-			throw new \Exception( 'Invalid Consumer Key or Secret.' );
+			throw new \WP_Error(
```
miguelpeixe, line 455: I don't think we should `throw` at API methods. Instead `return rest_ensure_response( new WP_Error() );`

includes/class-salesforce.php
```diff
@@ -524,13 +625,18 @@ public static function api_sync_salesforce( $request ) {
...
-			foreach ( $orders as $order ) {
+			foreach ( $orders as $order_item ) {
 				// Add the NPSP "Primary Contact" field if the Salesforce instance supports it.
 				if ( $is_npsp ) {
-					$order['npsp__Primary_Contact__c'] = $contact_id;
+					$order_item['npsp__Primary_Contact__c'] = $contact_id;
 				}
 
-				$opportunity_response = self::create_opportunity( $order );
+				$existing_opportunity = self::get_opportunity_by_order_id( $order_id, $order_item );
```
miguelpeixe, line 634: Calling this method here will make an external API request for every loop iteration. Should we be concerned about performance and rate limits?

miguelpeixe (changes requested):
It's working well! A few requests:

includes/class-salesforce.php
dkoo: Done in fa51798. 6342ebb2 also adds a missing return if the result of refreshing a token is an error.

includes/class-salesforce.php
dkoo: 99e08ed authenticates the request and d4e3a72 updates the capability to `edit_others_shop_orders`, which should be enough to verify that the user has permissions to access the order edit screen for any order.

includes/class-salesforce.php
dkoo: That's a great point! In my experience, even though each WooCommerce order object has an `orders` array representing all line items for that order, for most Newspack sites every order will only have one line item representing the total donation amount. I think we can revisit this if it becomes a problem. So far it hasn't been an issue for the biggest Newspack site using Salesforce.

miguelpeixe (approved):
Thank you for the revisions!

As I tested again, my newly placed order was not automatically synced although I had my site connected to Salesforce. I manually synced through the new UI.

It's non-blocking to me since it didn't happen when I tried a second time.

dkoo:
> As I tested again, my newly placed order was not automatically synced although I had my site connected to Salesforce. I manually synced through the new UI.

If you go to WooCommerce > Status > Logs, you might be able to see an error message for that failed sync. But the manual sync feature is meant to address exactly that scenario, since currently there's no way to manually retry if an order fails to sync automatically.

PR #1486, opened 2022-02-02
feat: enable AMP Plus for Jetpack Instant Search

miguelpeixe:
### All Submissions:

Allow Jetpack Search module to work with AMP Plus.

~~Note that the non-instant search widget filters will continue to not work. Its inputs are controlled through a [hardcoded script](https://github.com/Automattic/jetpack/blob/4ef58afbaba5396902194e5af699c9fe3e520318/projects/packages/search/src/widgets/class-search-widget.php#L544-L595) that we are unable to hook into since it's printed through the widget output.~~

~~Update: d54fd510167de718a8fd3a507affa63001055980 adds full support to the non-instant search widget. It adds AMP Plus attribute to the main `jetpack-search-widget` script and ports a customized script to handle the [sorting controls](https://github.com/Automattic/jetpack/blob/4ef58afbaba5396902194e5af699c9fe3e520318/projects/packages/search/src/widgets/class-search-widget.php#L544-L595).~~

Update: 325b0a3367b5559a3f444fc35cd8c2d5605ba583 adds a filter hook to the sanitization handler and parses the jetpack errors using IDs and texts. Some of the inline scripts do not have ID so it's parsing the script text to match.

1. Install this branch on a live or JN site
2. Enable the Jetpack Search module (ping me if you need assistance)
3. Make sure you have AMP installed with AMP Plus enabled
4. Visit the homepage, start typing on the search bar
5. Confirm the instant search is working as expected
6. Customize the search experience through the Jetpack dashboard and confirm the changes are applied

#### Non-instant search widget

With Instant Search disabled, the search widget behaves a bit differently with scripts that trigger search on input changes.

1. Keep the search module enabled and disable the instant search
2. Add the Jetpack search widget with extra filters to your search page sidebar
3. Visit the search page ( `?s=` )
4. Change sorting, select taxonomy terms and text search and confirm everything behaves as expected

### Other information:

### DIFF
--- a/includes/class-jetpack.php
+++ b/includes/class-jetpack.php
@@ -53,14 +53,14 @@ class Jetpack {
 	 */
 	public static function jetpack_modules_amp_plus( $is_sanitized, $error ) {
 		if ( isset( $error, $error['node_attributes'], $error['node_attributes']['id'] ) ) {
-			$has_any_id = array_reduce(
+			$script_has_matching_id = array_reduce(
 				self::$scripts_handles,
 				function( $carry, $id ) use ( $error ) {
 					return $carry || 0 === strpos( $error['node_attributes']['id'], $id ); // Match starting position so it includes `-js`, `-after` and `-before` scripts.
 				},
 				false
 			);
-			if ( $has_any_id ) {
+			if ( $script_has_matching_id ) {
 				$is_sanitized = false;
 			}
 		}
@@ -70,14 +70,14 @@ class Jetpack {
 			'JetpackInstantSearchOptions', // Jetpack Instant Search options.
 		];
 		if ( isset( $error, $error['text'] ) ) {
-			$has_any_text = array_reduce(
+			$script_has_matching_text = array_reduce(
 				$texts,
 				function( $carry, $text ) use ( $error ) {
 					return $carry || false !== strpos( $error['text'], $text );
 				},
 				false
 			);
-			if ( $has_any_text ) {
+			if ( $script_has_matching_text ) {
 				$is_sanitized = false;
 			}
 		}
--- a/includes/class-jetpack.php
+++ b/includes/class-jetpack.php
@@ -52,10 +52,6 @@ class Jetpack {
 	 * @return bool Whether the error should be rejected.
 	 */
 	public static function jetpack_modules_amp_plus( $is_sanitized, $error ) {
-		$texts = [
-			'jetpackSearchModuleSorting',  // Jetpack Search module sorting.
-			'JetpackInstantSearchOptions', // Jetpack Instant Search options.
-		];
 		if ( isset( $error, $error['node_attributes'], $error['node_attributes']['id'] ) ) {
 			$has_any_id = array_reduce(
 				self::$scripts_handles,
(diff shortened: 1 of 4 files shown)


includes/class-patches.php
```diff
@@ -25,6 +28,120 @@ public static function init() {
...
+			10,
+			3
+		);
+	}
+
+	/**
+	 * Reinject Jetpack Instant Search options with AMP Plus enabled.
+	 *
+	 * Jetpack uses `wp_add_inline_script()`, which does not allow for custom tag
+	 * attributes, so we have to inject it ourselves.
```
adekbadek, line 68: All frontend JS scripts pass via [`amp_validation_error_sanitized`](https://github.com/Automattic/newspack-plugin/blob/3a8cfa794806ee7d7b61ab4481587f51e631f60f/includes/class-amp-enhancements.php#L74) – this problem should be approached from there, as a [script-filter](https://github.com/Automattic/newspack-plugin/blob/3a8cfa794806ee7d7b61ab4481587f51e631f60f/includes/class-amp-enhancements.php#L81) (instead of enqueueing the scripts explicitly). The scripts used by this Jetpack's feature seem to have `id`s prefixed with `jetpack-instant-search-` and `wp-jp-i18n-`, e.g.:

image

(this is the value of the `$error` argument to `amp_validation_error_sanitized` filter)

adekbadek (changes requested):
Explicit enqueueing & copying scripts from JP codebase seems brittle – left a suggestion on how to tackle that in a more stable way.

includes/class-patches.php
miguelpeixe: Good call! I didn't know that was how AMP Plus was handled! 325b0a3367b5559a3f444fc35cd8c2d5605ba583 adds a filter hook to the sanitization handler and parses the jetpack errors using IDs and texts. Some of the inline scripts do not have ID so it's parsing the script text to match.

miguelpeixe: I also removed any of the i18n scripts from this PR because `wp-` scripts are already handled by default.

includes/class-jetpack.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Allow Jetpack modules scripts to be loaded in AMP Plus mode.
+	 *
+	 * @param bool|null $is_sanitized If null, the error will be handled. If false, rejected.
+	 * @param object    $error        The AMP sanitisation error.
+	 *
+	 * @return bool Whether the error should be rejected.
+	 */
+	public static function jetpack_modules_amp_plus( $is_sanitized, $error ) {
+		if ( isset( $error, $error['node_attributes'], $error['node_attributes']['id'] ) ) {
+			$has_any_id = array_reduce(
```
adekbadek, line 56: Nit: `$has_any_id` kinda threw me off – I'd expect something like `$script_has_matching_id`.

adekbadek (approved):
👌

includes/class-jetpack.php
miguelpeixe: Agreed! Updated both matching variables in 13da562ee8f60d11abbbc794853c3c09a4d650c9

adekbadek (approved):

PR #1499, opened 2022-02-09
fix(parsely): configuring Parse.ly

adekbadek:
### All Submissions:

Fixes an issue with Parse.ly config manager (#1207). Instead of checking if `Parsely` class has loaded, use WP core `is_plugin_active` function.

A full test would involve configuring the Parse.ly plugin via Newspack Manager, but I think the changes here is pretty self-explanatory.

### Other information:


claudiulodro (approved):
Good catch and thanks for the fix. This works well in my testing.

PR #1502, opened 2022-02-09
fix: lookup user by queried slug, not by login

dkoo:
### All Submissions:

Fixes a bug introduced in #1474. On author archive pages, the query looks for the user's slug, which is a sanitized form of the user login. For many users slug and login name will look the same, but for users whose slug differs from the login, the fix in #1474 will not work.

1. Create a subscriber user with spaces in the login name (e.g. "Subscriber User"). Ensure they're not assigned as an author to any posts.
2. On `master`, view their archive page and observe that the page does NOT 404 as it should since #1474.
3. Check out this branch and confirm that their archive page now 404s as expected.

### Other information:

---
  - https://app.asana.com/0/0/1201790554258200

### DIFF
--- a/includes/class-patches.php
+++ b/includes/class-patches.php
@@ -170,7 +170,7 @@ class Patches {
 		}
 
 		$author_name = $query->query_vars['author_name'];
-		$user        = get_user_by( 'login', $author_name );
+		$user        = get_user_by( 'slug', $author_name );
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Confirmed the issue gets fixed by this change

PR #1524, opened 2022-02-23
[HOLD] feat(popups): default segments UI

dkoo:
### All Submissions:

UI for https://github.com/Automattic/newspack-popups/pull/809.

See testing instructions in https://github.com/Automattic/newspack-popups/pull/809.

### Other information:


assets/wizards/popups/views/segments/SegmentsList.js
```diff
@@ -296,6 +308,26 @@ const SegmentsList = ( { wizardApiFetch, segments, setSegments, isLoading } ) =>
...
+		setInFlight( true );
+		setError( null );
+		wizardApiFetch( {
+			path: '/newspack/v1/wizard/newspack-popups-wizard/segmentation-defaults',
+			method: 'POST',
+			quiet: true,
+		} )
+			.then( _segments => {
+				setSegments( _segments );
+			} )
```
adekbadek, line 321: Simpler version, since `setSegments` is unary: 
```
.then( setSegments )
```

assets/wizards/popups/views/segments/SegmentsList.js
```diff
@@ -304,47 +336,93 @@ const SegmentsList = ( { wizardApiFetch, segments, setSegments, isLoading } ) =>
...
+						</Button>
+					</Card>
+				</Modal>
+			) }
+			<Button isPrimary onClick={ () => setShowDefaultModal( true ) }>
+				{ sprintf(
+					// Translators: button label for generating or resetting the default segments.
+					__( '%s default segments', 'newspack' ),
+					defaultSegments || defaultsCreated
+						? __( 'Reset', 'newspack' )
```
adekbadek, line 421: After deleting the generated segments, I see "Reset…" message:

image

adekbadek (changes requested):
This would use @thomasguillot 's input, but I think there's a problem with CTA hierarchy here:

image

The "Generate…" message should be more prominent IMO – perhaps centered and framed?

image

thomasguillot:
I agree with @adekbadek, there are too many Primary buttons.

Segments - Empty

dkoo:
I like the proposed design a lot better than my single button! One thing to note, there's discussion in https://github.com/Automattic/newspack-popups/pull/809 about automatically triggering the default segment creation if the site lacks any active prompts and segments, so that would mean the proposed UI would only really be shown for sites that are running active campaigns but haven't created any segments.

For sites that may have active campaigns and segments already, how should the button be presented? As a secondary button next to the "Add New Segment" button? I thought about putting a popover menu button next to it, but those components don't quite work in this context, outside of an action card component.

assets/wizards/popups/views/segments/SegmentsList.js
dkoo: My thinking is that it should say "Reset" in this case, since you had already generated the defaults but then deleted them.

assets/wizards/popups/views/segments/SegmentsList.js
```diff
@@ -304,47 +336,93 @@ const SegmentsList = ( { wizardApiFetch, segments, setSegments, isLoading } ) =>
...
+							{ __( 'OK', 'newspack' ) }
+						</Button>
+					</Card>
+				</Modal>
+			) }
+			<Button isPrimary onClick={ () => setShowDefaultModal( true ) }>
+				{ sprintf(
+					// Translators: button label for generating or resetting the default segments.
+					__( '%s default segments', 'newspack' ),
+					defaultSegments || defaultsCreated
```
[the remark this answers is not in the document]
dkoo: Ah, this is because the global we're reading is empty and the state gets cleared after unmounting the view. I added a check via API if the global is empty to handle this case: 6515f25

assets/wizards/popups/views/segments/SegmentsList.js
dkoo: 6515f25 moves `setDefaultsCreated( true );` to this promise handler so that the button label doesn't update if the initial creation fails.

dkoo:
@adekbadek @thomasguillot I pushed some changes in 18cbc23 to update the design as proposed. If we [decide](https://github.com/Automattic/newspack-popups/pull/809#issuecomment-1049120275) to auto-generate these default segments for new sites, I'll make that change here.

thomasguillot:
@dkoo I pushed a few more visual changes (https://github.com/Automattic/newspack-plugin/pull/1524/commits/20254be8f6030354df3d51bbe1dbef8890b90387)

wizard-updated

assets/wizards/popups/views/segments/SegmentsList.js
```diff
@@ -406,41 +406,37 @@ const SegmentsList = ( { wizardApiFetch, segments, setSegments, isLoading } ) =>
...
-							</div>
+							</Card>
 						</Card>
 					</Card>
 				</>
 			) }
 			{ showDefaultModal && (
 				<Modal
-					title={ __( 'Default segments', 'newspack' ) }
+					title={ __( 'Default Segments', 'newspack-popups' ) }
```
dkoo, line 431: I think the text domain here should be `newspack` since this UI lives in the main plugin, right?

thomasguillot:
Oops fixed:
https://github.com/Automattic/newspack-plugin/pull/1524/commits/18c5ced034594e45f486ef6ee93923e3f44d1fd7

dkoo:
It's been [decided](https://github.com/Automattic/newspack-popups/pull/809#issuecomment-1049735329) to automatically generate the default segments on WP init rather than via the UI as long as the site has no pre-existing segments. This means that @thomasguillot's lovely new UI will only ever be seen as a fallback in case this creation fails for whatever reason.

174f966 adds an additional check and error message if the required Campaigns config file is missing—I noticed that on non-Atomic sites, there's no visual warning in the wizard UI that we're missing the file and can't use Campaigns features.

adekbadek (approved):
Note: changes here depend on https://github.com/Automattic/newspack-popups/pull/809, these should be merged simultaneously.

dkoo:
Note to @thomasguillot: after some discussion we've decided we'd like to build this out into a more guided onboarding/wizard experience to guide users toward one of several sets of default segments, based on their input. @aschweigert will work on a suggested flow for this, but let's hold these changes until we can flesh that out a bit more.

dkoo:
Closing this as we've decided to go in a different direction for establishing defaults.

PR #1525, opened 2022-02-23
Update Ads and Campaigns ActionCards style

thomasguillot:
### All Submissions:

* Lightened `hasGreyHeader`
* Moved "Saved Settings" button to top region
* Updated the number of columns
* Removed `hasGreyHeader` from the placement's `ActionCard`s
* Updated `PlacementControl`'s label

__After__

Screenshot 2022-02-23 at 16 51 23

__Before__

Screenshot 2022-02-23 at 16 51 43

1. Switch to this branch
2. Go to the Ads Wizard
3. Test the Placements and Settings
4. Go to the Campaigns Wizard
5. Test the settings

### Other information:

### DIFF
--- a/assets/components/src/plugin-settings/SettingsSection.js
+++ b/assets/components/src/plugin-settings/SettingsSection.js
@@ -91,6 +91,14 @@ const SettingsSection = props => {
 			props
 		);
 	};
+	let columns;
+	if ( fields.length % 3 === 0 ) {
+		columns = 3;
+	} else if ( fields.length % 2 === 0 ) {
+		columns = 2;
+	} else {
+		columns = 1;
+	}
 	return (
 		<ActionCard
 			isMedium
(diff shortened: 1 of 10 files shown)


dkoo (approved):
I added the `hasGreyHeader` prop to the action card for Reader Revenue > Salesforce settings to match the Ads and Campaigns UI.

Non-blocking suggestion: the "Save Settings" button is very prominent now in the header. Should we consider disabling it until one of the settings in its card is changed?

Other than that, it works well and looks good 👍

thomasguillot:
> I added the `hasGreyHeader` prop to the action card for Reader Revenue > Salesforce settings to match the Ads and Campaigns UI.

Oh great, thank you @dkoo! I knew I forgot something ☺️ 

> Non-blocking suggestion: the "Save Settings" button is very prominent now in the header. Should we consider disabling it until one of the settings in its card is changed?

Good idea. I created an issue for this: https://github.com/Automattic/newspack-plugin/issues/1531

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -91,45 +87,53 @@ const SettingsSection = props => {
...
 		);
 	};
+	let columns;
+	if ( fields.length % 3 === 0 ) {
+		columns = 3;
+	} else if ( fields.length % 2 === 0 ) {
+		columns = 2;
+	} else {
+		columns = 1;
+	}
```
miguelpeixe, line 97: This calculation makes prime numbers (e.g. 5) always have 1 column. A better approach would be:

```suggestion
	let columns = 2;
	if ( fields.length % 3 === 0 ) {
		columns = 3;
	} else if ( fields.length === 1 ) {
		columns = 1;
	}
```

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -91,45 +87,53 @@ const SettingsSection = props => {
...
 			isMedium
 			disabled={ disabled }
 			title={ title }
 			description={ description }
 			toggleChecked={ active }
-			hasGreyHeader={ true }
+			hasGreyHeader={ active || hasGreyHeader }
 			toggleOnChange={ active !== null ? value => onUpdate( { active: value } ) : null }
+			actionContent={
+				( active || hasGreyHeader ) &&
```
miguelpeixe, line 108: It is not clear to me why the save button requires a grey header or a "toggleable" setting. This obligates every setting that is not toggleable to determine a style property for the functionality to work. Not having the grey header by default in such a case also results in an unexpected style:

image

thomasguillot: In your screenshot, you should have `hasGreyHeader`

miguelpeixe: You can see that [line 124 uses `activer || active === null`](https://github.com/Automattic/newspack-plugin/blob/a3649c8629c6a6ae81421bf7f4137a62e4036756/assets/components/src/plugin-settings/SettingsSection.js#L124) to determine if settings should be displayed, which is the same statement that should determine a grey header for this component.

dkoo: Is there ever a use case for the non-grey header style in @miguelpeixe's screenshot?

PR #1533, opened 2022-02-25
fix: logic for PluginSettings styles and functionality

dkoo:
### All Submissions:

I approved https://github.com/Automattic/newspack-plugin/pull/1525 perhaps prematurely. Based on @miguelpeixe's feedback in that thread, there are a couple of logical errors in the columns handling and the use of the grey header style prop to determine functionality. This PR addresses those concerns.

1. Test the instances of `PluginSettings` in the components demo, **Advertising > Settings**, **Campaigns > Settings**, and **Reader Revenue > Salesforce** contexts. 

### Other information:

### DIFF
--- a/assets/components/src/plugin-settings/SettingsSection.js
+++ b/assets/components/src/plugin-settings/SettingsSection.js
@@ -60,7 +60,7 @@ const SettingsSection = props => {
 		disabled,
 		onChange,
 		onUpdate,
-		hasGreyHeader,
+		hasGreyHeader = true,
 	} = props;
 	const getControlProps = setting => ( {
 		disabled,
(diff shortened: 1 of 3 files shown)


assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -102,10 +100,10 @@ const SettingsSection = props => {
 			title={ title }
 			description={ description }
 			toggleChecked={ active }
-			hasGreyHeader={ active || hasGreyHeader }
+			hasGreyHeader={ hasGreyHeader }
```
miguelpeixe, line 103: I wonder if we should have the `hasGreyHeader` property ever customizable since its state is basically determined by the state of other component's properties, which are processed internally (`active || active === null`).

miguelpeixe (changes requested):
Thank you for revising! 🙌

assets/components/src/plugin-settings/SettingsSection.js
dkoo: Oh, I see what you mean. Updated in 63ea672.

miguelpeixe (approved):
Thank you for revising! 👌 

@thomasguillot, do you think these changes are ok?

thomasguillot (approved):
👍

PR #1539, opened 2022-03-01
fix(design): header defaults in line with the theme

### LINKED ISSUE #1134: Design Wizard: Installing starter content changes header settings
**Describe the bug**
If you run the Newspack wizard and install the starter content _after_ you've made some customizations, some of them are overwritten. Specifically:

1. The header gets a background colour
2. The header switches to the 'short' layout
3. The homepage is replaced (similar to #1099), though the original page is still there, a new page is created and just replaces it in the customizer. 

I'm wondering if _not_ making the 'Starter Content' selected by default could help avoid this accidentally being run. 

**To Reproduce**
1. Start with a test site with the Newspack plugin, theme, and blocks installed.
2. Hop into the Customizer and make some changes; leave the header with the default BG, and default height. 
3. Run the Newspack Wizard, including installing starter content.
4. Don't change any of the design settings; if you look you may notice some of the header settings (aren't what you set (solid BG, short header)
5. View the design settings on the front-end and note the difference.

**Screenshots**
Wizard screenshot:

Screen Shot 2021-09-09 at 4 53 33 PM

How my site originally looked:
Screen Shot 2021-09-09 at 4 35 15 PM

After running the wizard:
image (4)

**Expected behavior**
All of my settings would be maintained (or at least the header ones; the homepage changing somewhat makes sense!)

adekbadek:
### All Submissions:

Closes #1134.

Makes `header_solid_background` and `header_simplified` theme mods `false` by default, as they are in the Theme:

- [`header_solid_background`](https://github.com/Automattic/newspack-theme/blob/f3b2347f3c9c9622738394093dc25f8dfa945311/newspack-theme/inc/customizer.php#L75-L82)
- [`header_simplified`](https://github.com/Automattic/newspack-theme/blob/f3b2347f3c9c9622738394093dc25f8dfa945311/newspack-theme/inc/customizer.php#L93-L100)

Observe #1134 is not reproducible

### Other information:

### DIFF
--- a/includes/class-starter-content.php
+++ b/includes/class-starter-content.php
@@ -154,8 +154,8 @@ class Starter_Content {
 				set_theme_mod( 'logo_size', 0 );
 			}
 		}
-		set_theme_mod( 'header_solid_background', true );
-		set_theme_mod( 'header_simplified', true );
+		set_theme_mod( 'header_solid_background', false );
+		set_theme_mod( 'header_simplified', false );
 		return true;
 	}
 


laurelfulford (approved):
This looks good to me! :shipit: 

I retried my original test, and confirmed the taller header/non-solid background options were kept. On a second site I also tried setting a solid background and shorter header, and confirmed that those options were also preserved when running through the wizard afterwards.

PR #1563, opened 2022-03-16
feat: allow segmentation by user login status

dkoo:
### All Submissions:

UI for the new segmentation criteria enabled by https://github.com/Automattic/newspack-popups/pull/825.

Closes https://github.com/Automattic/newspack-popups/issues/746.

See testing instructions in https://github.com/Automattic/newspack-popups/pull/825.

### Other information:


assets/wizards/popups/views/segments/SingleSegment.js
```diff
@@ -251,6 +253,31 @@ const SingleSegment = ( { segmentId, setSegments, wizardApiFetch } ) => {
 						] }
 					/>
 				</SettingsSection>
+				<SettingsSection title={ __( 'Login State', 'newspack' ) }>
```
miguelpeixe, line 256: Super non-blocking nit: isn't "Authenticated" a better word for this?

miguelpeixe (approved):

assets/wizards/popups/views/segments/SingleSegment.js
dkoo: Ah, I had the same thought but forgot to push the change in d04c57f. I think this describes the intent of the feature better since all users known to have a user account are part of the segment, regardless of login status (so really a user only has to log in once to be considered part of the segment):

* Setting name: **User Account**
* Options: **Has user account**, **Does not have user account**

Thoughts?

miguelpeixe: Makes sense, especially considering that once the session identifies authentication it maintains the "has user account" states after logging out.

miguelpeixe (approved):

PR #1595, opened 2022-04-08
fix(popups): analytics popups report label regex

dkoo:
### All Submissions:

https://github.com/Automattic/newspack-popups/pull/830 added segment names to the `event_label` strings sent with Google Analytics event reporting. This changed the label from a format like this:

```
Above Header: Spring campaign 2022 banner ad (112982)
```

to this:

```
Above Header: Spring campaign 2022 banner ad (112982) - Everyone
```

With the segment(s) coming after the post ID. However, the Campaigns analytics report expects the post ID in parentheses to be the last item in the label, so it's ignoring rows that have the segment names appended—basically all events after https://github.com/Automattic/newspack-popups/pull/830 got deployed. This PR amends the regex patterns to look for the post ID anywhere in the label, not just at the end, which should fix the reporting view.

It's hard to test this without having access to a site that's connected to GA with active prompts that have been running. However, if you have a test site that has any past events to report, you can mock the names by adding a hardcoded value on [this line](https://github.com/Automattic/newspack-plugin/blob/master/includes/popups-analytics/class-popups-analytics-utils.php#L299):

```php
$row['dimensions'][1] = 'Above Header: Spring campaign 2022 banner ad (112982) - Everyone';
```

1. On `master`, view the **Newspack > Campaigns > Analytics report**. Observe that with the mocked name including segments as above, no events are reported.
2. Check out this branch and confirm that the events are reported.

### Other information:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.80.1-hotfix.1](https://github.com/Automattic/newspack-plugin/compare/v1.80.0...v1.80.1-hotfix.1) (2022-04-08)
+
+
+### Bug Fixes
+
+* **popups:** analytics popups report label regex ([56e72f2](https://github.com/Automattic/newspack-plugin/commit/56e72f223cf3c211bac7cb94f37f35d7822136db))
+
 # [1.80.0](https://github.com/Automattic/newspack-plugin/compare/v1.79.1...v1.80.0) (2022-04-05)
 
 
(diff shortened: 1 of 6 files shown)


dkoo:
111b63a updates the regex to be a bit more specific by looking for post IDs in parentheses succeeded by either an end-of-line character OR a ` -` hyphen separator.

Note that it's possible someone could name their prompt `Prompt name (123) - subscribe` which will result in an event label like `Prompt name (123) - subscribe (postId) - segments` in which case the reporting may not be able to find the post ID correctly as it will match the `123` in the title instead. However, this seems to be an unlikely enough edge case to not worry much about.

miguelpeixe (approved):
Removing the `$` argument from the regex and using `array_pop()` does the trick! 👌

PR #1593, opened 2022-04-08
feat(stripe-donations): update WooCommerce on successful donation

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds WC synchronisation to Stripe donations processing.

### How to test the changes in this Pull Request:

1. Set up the site to use the streamlined Donate block – turn on AMP Plus* and choose "Stripe" as platform in Reader Revenue wizard
2. Ensure your site can ingest Stripe webhooks – it has to be publicly available and webhook has to be active
3. Activate WooCommerce and WooCommerce Stripe Gateway plugins
4. Make a donation using the Donate block
5. Observe an order is created in WC "Orders" view, 
    - with an order note saying what kind of donation is it
    - if opted in to newsletter, another order note should state it
    - there should be a link to payment in Stripe dashboard, below "Order … details" header
7. Try with a few different options and observe that the order data checks out:
    1. with/without newsletter opt-in
    2. with/without covering Stripe fees
    3. different amounts and frequencies
8. Set `newspack-blocks` to branch `add/donate-send-user-id`, and make a donation as a logged-in user. Observe the WC order is assigned to the user
 

\* add `define( 'NEWSPACK_AMP_PLUS_ENABLED', true );` in `wp-config.php`

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1599, opened 2022-04-11
feat(donations-stripe): integrate w/ woocommerce-memberships

adekbadek:
### All Submissions:

Adds integration with `woocommerce-memberships` plugin.

1. Switch to `add/handle-wc-memberships` branch of `newspack-blocks` plugin (https://github.com/Automattic/newspack-blocks/pull/1096)
2. Follow the setup from https://github.com/Automattic/newspack-plugin/pull/1593 (steps 1-3)
3. Activate `woocommerce-memberships` plugin and set up a membership plan* tied to donations
4. In a session as a non-logged-in visitor, confirm the content is restricted (as per membership plan) and make a donation
5. Observe the restricted content is now visible
6. Observe that a user account was created and you're now logged in by visiting `/wp-admin` page
7. In a different session, make a donation as an already logged-on user. Observe the restricted content is now visible as per plan

\* e.g. restrict all post content unless one of the Newspack Donations products was bought

### Other information:

### DIFF
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -16,6 +16,63 @@ defined( 'ABSPATH' ) || exit;
  */
 class WooCommerce_Connection {
 	/**
+	 * Get WC's Order Item related to a donation frequency.
+	 *
+	 * @param string $frequency Donation frequency.
+	 * @param number $amount Donation amount.
+	 */
+	private static function get_donation_order_item( $frequency, $amount = 0 ) {
+		$product_id = Donations::get_donation_product( $frequency );
+		if ( false === $product_id ) {
+			return false;
+		}
+
+		$item = new \WC_Order_Item_Product();
+		$item->set_product( wc_get_product( $product_id ) );
+		$item->set_total( $amount );
+		$item->set_subtotal( $amount );
+		return $item;
+	}
+
+	/**
+	 * If the site is using woocommerce-memberships, create a new user and a
+	 * membership, if the donation type calls for it.
+	 *
+	 * @param string $email_address Email address.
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
Not sure exactly why, but after the succesful purchase, the user account is created and authenticated but the order is not set to the customer ID and the user doesn't get access to the membership. Accessing `/my-account/orders/` I get `No order has been made yet.` and the same to `/my-account/members-area/`.

This is the membership plan config:

image

Don't know if related but with debug mode I get several warnings regarding deprecated usage of `is_offline()` and `is_500()` from the Membership plugin:

````
PHP Notice:  is_offline was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run. Before then, they always return false.
````

```
PHP Notice:  is_500 was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run. Before then, they always return false.
```

adekbadek:
@miguelpeixe can you confirm if the site is publicly available, so webhook endpoint can process Stripe request?

> Don't know if related but with debug mode I get several warnings regarding deprecated usage of is_offline() and is_500() from the Membership plugin:

Are you sure these are coming from WC Memberships? I can't see any use of these functions in the plugin code.

miguelpeixe:
> Are you sure these are coming from WC Memberships? I can't see any use of these functions in the plugin code.

The stack trace points to it:

image

It's mostly a deprecation warning that we probably shouldn't have to worry right now.

miguelpeixe:
> Are you sure these are coming from WC Memberships? I can't see any use of these functions in the plugin code.

The stack trace points to it:

image

It's mostly a deprecation warning that we probably shouldn't have to worry right now.

miguelpeixe (approved):

PR #1603, opened 2022-04-12
fix(popups): improve formatting of human-readable numbers

dkoo:
### All Submissions:

Small fix to improve the display formatting of human-readable "Seen" numbers in Campaigns Analytics according to the following rules:

* If the value is less than 10k, shows the absolute value formatted according to the local language via [`toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString).
* If the value is greater than 10k but less than 1M, shows a human-readable figure rounded to the nearest integer. e.g. 11,236 becomes `11K` and 157,963 becomes `158K`
* If the value is greater than 1M, shows a human-readable figure rounded to the hundredths. e.g. 1,353,288 becomes `1.35M` and 826,677,234 becomes `826.68M`.

Closes https://github.com/Automattic/newspack-popups/issues/835.

1. If your test site isn't connected to Google Analytics via Site Kit, you can force the Campaigns Analytics view to display by adding `return [];` at [this line](https://github.com/Automattic/newspack-plugin/blob/master/includes/popups-analytics/class-popups-analytics-utils.php#L166). 
2. Mock various numbers by altering [this line](https://github.com/Automattic/newspack-plugin/blob/master/assets/wizards/popups/views/analytics/Info.js#L26) like so:

```js
const { form_submissions, link_clicks } = keyMetrics;
const seen = 12345; // a number to test.
```

3. Test numbers in the singles, hundreds, thousands, tens of thousands, hundreds of thousands, and millions and confirm that the "Seen" number displayed is human-readable and makes logical sense according to the rules above.

### Other information:

### DIFF
--- a/assets/wizards/popups/views/analytics/Info.js
+++ b/assets/wizards/popups/views/analytics/Info.js
@@ -52,7 +52,7 @@ const Info = ( { keyMetrics, filtersState, labelFilters, isLoading, postEditLink
 				{ [
 					{
 						label: __( 'Seen', 'newspack' ),
-						value: humanNumber( seen, limitDecimals ),
+						value: 10000 < seen ? humanNumber( seen, limitDecimals ) : seen.toLocaleString(),
 						withSeparator: true,
 					},
 					{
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Nice! 💯

PR #1609, opened 2022-04-13
fix: crashes with autocomplete inputs in Campaigns wizard and CategoryAutocomplete

dkoo:
### All Submissions:

Fixes two JS errors that result in a hard crash when "selecting" an undefined item in the autocomplete fields in the Campaigns wizard UI. The change to `assets/components/src/category-autocomplete/index.js` will require a new NPM version release to fix in the NPM package.

Closes https://github.com/Automattic/newspack-popups/issues/810.

1. On `master`, visit the **Newspack > Campaigns** wizard. Click the settings icon to open the settings modal:

Screen Shot 2022-04-13 at 5 20 06 PM

2. In the Category, Tag, or Category/Tag Exclusion fields, start typing in a term and hit enter without having a term highlighted. Observe that the UI crashes with a JS console error.
3. Check out this branch and repeat, confirming that the UI does not crash and that hitting enter is harmless and doesn't result in changing the autocomplete input's value.

### Other information:

### DIFF
--- a/assets/wizards/popups/components/settings-modal/index.js
+++ b/assets/wizards/popups/components/settings-modal/index.js
@@ -178,9 +178,7 @@ const PromptSettingsModal = ( { prompt, disabled, onClose, segments, updatePopup
 								onChange={ tokens =>
 									setPromptConfig( {
 										options: {
-											excluded_categories: tokens
-												.filter( token => token?.id )
-												.map( token => token.id ),
+											excluded_categories: tokens.map( token => token.id ),
 										},
 									} )
 								}
@@ -198,7 +196,7 @@ const PromptSettingsModal = ( { prompt, disabled, onClose, segments, updatePopup
 								onChange={ tokens =>
 									setPromptConfig( {
 										options: {
-											excluded_tags: tokens.filter( token => token?.id ).map( token => token.id ),
+											excluded_tags: tokens.map( token => token.id ),
 										},
 									} )
 								}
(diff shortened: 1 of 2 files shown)


assets/wizards/popups/components/settings-modal/index.js
```diff
@@ -191,7 +198,18 @@ const PromptSettingsModal = ( { prompt, disabled, onClose, segments, updatePopup
 								hideHelpFromVision
 								taxonomy="tags"
 								value={ excludedTags || [] }
-								onChange={ tokens => setPromptConfig( { options: { excluded_tags: tokens } } ) }
+								onChange={ tokens => {
+									return setPromptConfig( {
+										options: {
+											excluded_tags: tokens.reduce( ( acc, token ) => {
+												if ( token?.id ) {
```
adekbadek, line 205: Why shouldn't the `CategoryAutocomplete` component do the filtering?

dkoo: The `onChange` function is passed as a prop to `CategoryAutocomplete`. I tried adding some filtering to the `handleOnChange` helper in the component, but it doesn't prevent the `undefined` error so the filtering is still required in the function passed as `onChange`.

That said, cc662e7 does handle the filtering in a slightly more readable way, and I retained the extra filtering in `CategoryAutocomplete` just in case.

adekbadek: If the filtering happens after the mapping in `handleOnChange`, the value passed to the `onChange` will always be valid (there will be no need for filtering in `onChange`):

```js
const allValues = tokens
	.map( token => ( 'string' === typeof token ? suggestions[ token ] : token ) )
	.filter( Boolean );
```

dkoo: Ah, you're right. Updated in 94a399e. I was putting `filter` before the `map`, which was still causing the `undefined` error.

adekbadek (approved):

PR #1653, opened 2022-05-18
Fix/ga4 on amp

adekbadek:
### All Submissions:

Adds GA4 support to AMP pages.

#### Questions 

- Is the ping to DoubleClick (`stats.g.doubleclick.net`) necessary? I'm assume it's analytics for GAM – cc @kmwilkerson 

1. Ensure a GA4 property is configured in Site Kit:

image

2. Load an AMP page in an incognito session, observe in your GA4 dashboard* (real-time view) that a visit has been recorded 

\* Search for the GA4 property in the [same old GA dashboard](https://analytics.google.com/analytics/web) 

### Other information:


includes/class-analytics.php
```diff
@@ -838,5 +839,41 @@ protected static function output_js_ini_load_event( $event ) {
 		</script>
 		<?php
 	}
+
+	/**
+	 * Add GA4 analytics support to AMP pages.
+	 */
+	public static function insert_ga4_analytics() {
+		if ( function_exists( 'is_amp_endpoint' ) && ! is_amp_endpoint() ) {
```
miguelpeixe, line 847:
```suggestion
		if ( ! function_exists( 'is_amp_endpoint' ) || ! is_amp_endpoint() ) {
```

If the function doesn't exist, should assume AMP is not installed and also return.

miguelpeixe (changes requested):
Working as described! Only one minor fix below

includes/class-analytics.php
adekbadek: This will have the same effect, if the function does not exist then the right-hand side of `&&` won't be evaluated. That's how it's done in [other](https://github.com/Automattic/newspack-plugin/blob/e89eda4e1b01f4799865b592b1bf9be14d0ee7e4/includes/class-nrh.php#L169) [places](https://github.com/Automattic/newspack-plugin/blob/e89eda4e1b01f4799865b592b1bf9be14d0ee7e4/includes/class-analytics.php#L562), so I'd prefer to keep it consistent.

miguelpeixe: But there's a logic error.

This is fine when evaluating `function_exists( 'is_amp_endpoint' ) && is_amp_endpoint()` to return if AMP is available and active. To have the opposite effect you should either `! ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() )` or `! function_exists( 'is_amp_endpoint' ) || ! is_amp_endpoint()`.

PR #1656, opened 2022-05-19
Setup Wizard: Remove theme selection from setup wizard

### LINKED ISSUE #1544: Remove theme selection from design setup tool
When walking through the Newspack setup tool (`/wp-admin/admin.php?page=newspack-setup-wizard#/design`), the user is asked to choose from one of Newspack's six themes. However, Newspack's staff designer is choosing the theme during our "Newspackification" process. So any choice there is likely to be replaced with a better option more suited to their needs.

The setup tool should default to the Newspack parent theme and not present alternatives to the publisher.

We could alternatively present more contextual information about what the strengths and raisons d'etre for each theme are, allowing the publisher to make an informed choice that is less likely to need to be overridden during Newspackification.

Or we could still default to the parent theme and ask the user, "Using the information you are about to provide, our staff designer will choose an appropriate theme for you. Do you wish to suggest a particular one?"

**Additional context**
Some publishers have been stymied by this step in the design process and have halted their progress, retreating to Slack to seek guidance from their TAM. Removal of this step would streamline and expedite this process.

laurelfulford:
### All Submissions:

This PR removes the theme selection step from the Setup Wizard (but keeps it on the regular Site Design screen).

Closes #1544

2. Run the Newspack Setup Wizard; confirm that when you get to the Design tab of the setup, it starts with the Homepage Patterns rather than theme selection:

image

3. Complete the wizard.
4. Navigate to Dashboard > Newspack > Site Design and confirm that the theme selection is still at the top. 

(As an aside, I'd appreciate some feedback on whether `! isPartOfSetup` was a good choice here -- it felt weird to me to test for a `false` since every other check in the file for `isPartOfSetup` was checking for `true`, but I might be overthinking this! 😅). 

### Other information:

### DIFF
--- a/assets/wizards/site-design/views/main/index.js
+++ b/assets/wizards/site-design/views/main/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { useEffect, useState } from '@wordpress/element';
+import { useEffect, useState, Fragment } from '@wordpress/element';
 import { alignCenter, alignLeft } from '@wordpress/icons';
 import { __ } from '@wordpress/i18n';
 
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
>  it felt weird to me to test for a false since every other check in the file for isPartOfSetup was checking for true

That's fine – `WebPreview` should be display only on setup, hence the positive check, and `ThemeSelection` _unless_ it's setup, hence the negative check.

PR #1669, opened 2022-05-27
fix: deactivate Salesforce syncing if RR platform is not Newspack

dkoo:
### All Submissions:

Fixes an issue where if you had previously set up a Salesforce integration on the Newspack Reader Revenue platform, then changed to a different platform that still uses WooCommerce (e.g. Stripe), the Salesforce webhooks would still fire and attempt to sync, but fail because the orders don't contain the expected data. This fixes the issue by deactivating the webhooks if the wrong Reader Revenue platform is active, which should help clean up the debug log.

1. On `master`, make sure WooCommerce and all required extensions are installed and active. Set your Reader Revenue platform to "Newspack" and set up a [Salesforce integration](https://newspack.pub/support/reader-revenue/salesforce/). Ping me if you need test credentials. 
2. Change your platform to Stripe and complete Stripe configuration. Make sure to create the required webhooks.
3. On the front-end, make a donation via Stripe.
4. After a few minutes, observe fatals in your debug.log indicating a failed sync. In the WP dashboard, go to **WooCommerce > Orders** and find the order corresponding to your Stripe donation. Observe a "Salesforce sync status" message notice in the Order Actions panel indicating a failed sync.
5. Check out this branch, repeat step 3. This time, confirm there are no PHP errors thrown after the WC order is created, and confirm that the "Salesforce sync status" messaging is no longer shown in WC order pages.

### Other information:

### DIFF
--- a/includes/class-salesforce.php
+++ b/includes/class-salesforce.php
@@ -857,6 +857,11 @@ class Salesforce {
 	 * @return void
 	 */
 	private static function create_webhook() {
+		// Failsafe in case WC isn't active.
+		if ( ! class_exists( 'WC_Webhook' ) ) {
+			return;
+		}
+
 		$webhook = new \WC_Webhook();
 		$webhook->set_name( 'Sync Salesforce on order checkout' );
 		$webhook->set_topic( 'order.created' ); // Trigger on checkout.
--- a/includes/class-salesforce.php
+++ b/includes/class-salesforce.php
@@ -41,11 +41,30 @@ class Salesforce {
 	 * Add hooks.
 	 */
 	public static function init() {
+		add_action( 'init', [ __CLASS__, 'platform_check' ] );
 		add_action( 'rest_api_init', [ __CLASS__, 'register_api_endpoints' ] );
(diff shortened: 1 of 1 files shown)


includes/class-salesforce.php
```diff
@@ -41,13 +41,41 @@ class Salesforce {
...
 		add_action( 'woocommerce_order_action_newspack_sync_order_to_salesforce', [ __CLASS__, 'wc_order_manual_sync_action' ] );
 	}
 
+	/**
+	 * Check if Newspack is the current Reader Revenue platform.
+	 * If not, disable syncing. If so and the required webhook
+	 * doesn't exist, create it.
+	 */
+	public static function platform_check() {
+		$is_newspack = self::platform_is_newspack();
```
adekbadek, line 58: The method is already implemented in `Donations` class.

```suggestion
		$is_newspack = Donations::is_platform_wc();
```

(this will have to change in a few more places here)

adekbadek (changes requested):

includes/class-salesforce.php
dkoo: Nice, I missed that. Updated in b591fde

adekbadek (approved):

PR #1676, opened 2022-05-31
Simplify components and use Gutenberg's

thomasguillot:
### All Submissions:

This PR removes some of the custom components and styles in order to use standard Gutenberg components.

Since more and more plugins are using Gutenberg components I thought it'd make sense for us to do the same. This will bring some consistency throughout the different plugins.

This should also prevent some unexpected bugs when Gutenberg updates a component and breaks our custom styles.

We are still using a bunch of custom components but I redesigned them to match Gutenberg's style.

__After:__

after 1
after 2
after 3

__Before:__

before 1
before 2
before 3

This is quite a big PR so please test ALL the different screens/settings

### Other information:


assets/components/src/category-autocomplete/style.scss
```diff
@@ -11,37 +11,23 @@
...
 	// Within a Grid
 
-	.newspack-grid & .newspack-form-token-field {
+	.newspack-grid & .newspack-form-token-field__input-container {
 		margin: 0;
 	}
 
 	// Loading state spinner
 	&__suggestions-spinner {
-		bottom: 0.5em;
```
dkoo, line 43: Not sure if this is desired, but the spinner is now appearing below the input:

Screen Shot 2022-06-09 at 1 00 47 PM

Previously, it appeared inside the input:

Screen Shot 2022-06-09 at 1 03 01 PM

dkoo: Speaking of the spinner, it has a weird grey background and the positioning seems slightly off in Gutenberg editor pages. For example, try the Author List block, or an Author Profile block and search for an author name using the autocomplete field.

Screen Shot 2022-06-09 at 1 06 47 PM
Screen Shot 2022-06-09 at 1 09 24 PM

Again, this only happens in Gutenberg editor pages. On the Components Demo page, the spinner looks fine in those contexts.

assets/components/src/color-picker/style.scss
```diff
@@ -7,283 +7,34 @@
...
-				border-color: $primary-500;
-				box-shadow: 0 0 0 2px $primary-500;
-			}
-		}
-
-		// Reset
+	&__expander {
+		box-shadow: inset 0 0 0 1px rgb( 0 0 0 / 20% );
+		border-radius: 2px;
+		height: 64px;
```
dkoo, line 18: Not sure if this is intentional, but the fixed height causes the color HEX label to appear off-center, vertically.

Screen Shot 2022-06-09 at 12 56 48 PM

assets/components/src/card/style.scss
```diff
@@ -21,11 +20,8 @@
 	&__buttons-card {
 		display: flex;
 		flex-wrap: wrap;
-		margin: 32px -8px 16px;
```
dkoo, line 24: I think removing these styles causes some strange alignment in this part of the Components Demo page. Not sure if it affects any real UI, though:

Screen Shot 2022-06-09 at 12 56 26 PM

assets/components/src/button/style.scss
```diff
@@ -2,373 +2,10 @@
...
-	color: $text;
-}
-
-body[class*='admin-color-'] {
-	.newspack-button {
-		border-radius: 3px;
-		font-size: 14px;
-		font-weight: bold;
-		height: auto;
-		line-height: 24px;
```
dkoo, line 20: The reduced/default line height for buttons causes some Action Card components to look a bit squished. Maybe we need some margin for these?

Screen Shot 2022-06-09 at 12 57 48 PM

assets/components/src/select-control/style.scss
```diff
@@ -2,114 +2,29 @@
...
-
-		select.components-select-control__input {
-			color: $gray-700;
-			border-color: $gray-300;
-			font-size: 16px;
-			height: 48px;
-			line-height: 24px;
-			max-width: 100%;
-			min-height: 48px;
+		&[multiple] {
```
dkoo, line 11: The selected items in a multiselect component stop just short of going full-width, and look somewhat strange. I think we may also want to add some similar padding so that the top of the first selection aligns with the text in the single-select component next to it:

Screen Shot 2022-06-09 at 1 22 50 PM

assets/components/src/checkbox-control/style.scss
```diff
@@ -1,78 +0,0 @@
...
-
-@import '../../../shared/scss/colors';
-@import '~@wordpress/base-styles/colors';
-
-.newspack-checkbox-control {
-	color: $gray-700;
-	display: flex;
-	font-size: 14px;
-	line-height: 24px;
-	margin: 32px 0;
```
dkoo, line 13: Maybe outside the scope of this PR, but removing the margins causes some Settings components to look strange. For example, in the Reader Revenue > Stripe Settings page:

Before:

Screen Shot 2022-06-09 at 1 26 08 PM

After:

Screen Shot 2022-06-09 at 1 25 29 PM

dkoo (changes requested):
For the most part, I love the simplification and the move toward using core WP styles where appropriate! I ran into a few transitional issues that mainly have to do with components that probably looked okay due to our custom styles, but now have some visual weirdness with the core styles. I tested by going into every wizard page of the Newspack dashboard, and also edited CPTs like prompts, listings, and newsletters.

I do see that we've removed a few components entirely, but these shouldn't cause any issue as I don't see any external repos importing these removed components from `newspack-components`:

* `ButtonGroup`
* `CheckboxControl`
* `TextareaControl`
* `ToggleGroup`

assets/components/src/color-picker/style.scss
thomasguillot: It is intentional. It's to match Core's [ColorPalette](https://wordpress.github.io/gutenberg/?path=/story/components-colorpalette--default) component

assets/components/src/card/style.scss
thomasguillot: This specific demo was always misaligned.
Screenshot 2022-06-10 at 10 06 11

Anyway... I fixed the alignment: https://github.com/Automattic/newspack-plugin/pull/1676/commits/983b3f2f0546b3823389598454f56d4eb5b303ce 😄

assets/components/src/checkbox-control/style.scss
thomasguillot: I've fixed the Stripe layout here: https://github.com/Automattic/newspack-plugin/pull/1676/commits/99b4ea41a3bd61f61dbd4a8e3ce7ad834240bb41

Did you notice more broken Settings components?

assets/components/src/category-autocomplete/style.scss
thomasguillot: Moving the spinner below the input is intentional but I might rework this.

I'm not sure what's happening with the author profile/list. I wonder if it's because of the new wp spinner component vs our newspack component. Might be worth testing again when we update newspack-components and update the npm package

thomasguillot: Spinner tweaked: https://github.com/Automattic/newspack-plugin/pull/1676/commits/89919197c2434145e8b003ee34d179fe0d609aa5

assets/components/src/checkbox-control/style.scss
thomasguillot: I noticed the Campaigns Settings were also broken: https://github.com/Automattic/newspack-plugin/pull/1676/commits/89919197c2434145e8b003ee34d179fe0d609aa5

assets/components/src/button/style.scss
thomasguillot: Fixed: https://github.com/Automattic/newspack-plugin/pull/1676/commits/789dc0f6d602a9502358f42c4662020880701412

assets/components/src/select-control/style.scss
thomasguillot: Fixed: https://github.com/Automattic/newspack-plugin/pull/1676/commits/8d757e8f66bc04c39db290e2212e30b7ac5a527c

Tbh, I'm don't think we even use a multiple select other than the demo page.

assets/components/src/button/style.scss
dkoo: I like the option to make the secondary button `destructive`, too 👍

assets/components/src/checkbox-control/style.scss
dkoo: Didn't notice any other broken settings components. Campaigns does look better now (the spinner positioning I highlighted in another comment came from Campaigns settings).

dkoo (approved):
🙌

assets/components/src/checkbox-control/style.scss
thomasguillot: Thank you @dkoo.

PR #1687, opened 2022-06-03
fix: oauth error handling

adekbadek:
### All Submissions:

Fixes handling of failed OAuth credentials saving.

1. On `master`, explicitly return a `WP_Error`* from `save_auth_credentials` – try to connect Google via OAuth and observe it failing silently
2. Switch to this branch, observe the error returned to the front-end

\* Comment out the CSRF condition

### Other information:

### DIFF
--- a/includes/oauth/class-google-oauth.php
+++ b/includes/oauth/class-google-oauth.php
@@ -276,11 +276,8 @@ class Google_OAuth {
 		if ( is_wp_error( $user_info_data ) ) {
 			return $user_info_data;
 		}
-		return \rest_ensure_response(
-			[
-				'user_basic_info' => $user_info_data,
-			]
-		);
+		$response['user_basic_info'] = $user_info_data;
+		return \rest_ensure_response( $response );
 	}
 
 	/**
--- a/includes/oauth/class-google-oauth.php
+++ b/includes/oauth/class-google-oauth.php
@@ -130,7 +130,13 @@ class Google_OAuth {
 
 		if ( $tokens['csrf_token'] !== $saved_csrf_token ) {
 			Logger::log( 'Failed saving credentials - CSRF token mismatch.' );
-			return new \WP_Error( 'newspack_google_oauth', __( 'Session token mismatch.', 'newspack' ) );
+			return new \WP_Error(
+				'newspack_google_oauth',
+				__( 'Session token mismatch.', 'newspack' ),
+				[
+					'status' => 403,
+				]
+			);
 		}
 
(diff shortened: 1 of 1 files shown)


includes/oauth/class-google-oauth.php
```diff
@@ -207,8 +213,14 @@ public static function api_google_auth_save_details( $request ) {
...
+		}
+		$auth_save_result = self::save_auth_credentials( $auth_save_data );
+		if ( is_wp_error( $auth_save_result ) ) {
+			return $auth_save_result;
 		}
-		if ( self::save_auth_credentials( $auth_save_data ) ) {
+		if ( $auth_save_result ) {
 			Logger::log( 'Credentials saved.' );
 			return \rest_ensure_response( [ 'status' => 'ok' ] );
 		} else {
```
miguelpeixe, line 226: This check that returns `Could not save auth data for user.` is a result of `add_option()` not updating the database. The only reason for that to happen is if the option already exists, which will never be the case due to `self::remove_credentials()` always being called.

includes/oauth/class-google-oauth.php
```diff
@@ -130,7 +130,13 @@ private static function save_auth_credentials( $tokens ) {
...
+			return new \WP_Error(
+				'newspack_google_oauth',
+				__( 'Session token mismatch.', 'newspack' ),
+				[
+					'status' => 403,
+				]
+			);
 		}
 
 		$auth                 = self::get_google_auth_saved_data();
```
miguelpeixe, line 142: We could take advantage of more checks here, to make sure that the incoming `$tokens` data is valid.

miguelpeixe (changes requested):

includes/oauth/class-google-oauth.php
adekbadek: `remove_credentials` might be removed from the method, and there might be a DB issue preventing the option from being saved. Returning the success status only if the option was saved is a more robust solution than assuming it succeeded.

includes/oauth/class-google-oauth.php
adekbadek: Good idea – removed the refresh token optionality, too. In 53c173178a8f6fe071131167c2aca1f3a0ddcd6b

adekbadek: That's right, reverted in a540601e33509e95fec6c5a0e9121d4aaa1f55d4

miguelpeixe (approved):

PR #1681, opened 2022-06-01
feat(analytics): automatically link GA4 with Site Kit

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds automatic setup of GA4 property with Site Kit, if the property exists in GA.

### How to test the changes in this Pull Request:

Test on a site with Site Kit configured and Universal GA connected, but GA4 not connected.  

1. Confirm that the page is not sending hits to the GA4 property on `master`*
1. Visit Site Kit's Analytics settings page (`/wp-admin/admin.php?page=googlesitekit-settings#/connected-services/analytics`) and observe that GA4 is connected 
3. Visit an AMP page and confirm the GA4 snippet is inserted (search for `GA4_MEASUREMENT_ID` in the page source) and GA4 property is hit**

\* Just one `POST` request with `pageview` in payload to `google-analytics.com` should be sent
\** So there should be two of these requests, one for the old Universal GA and one for GA4

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1698, opened 2022-06-08
feat(analytics): automatically link GA4 with Site Kit

adekbadek:
### All Submissions:

Adds automatic setup of GA4 property with Site Kit, if the property exists in GA. This is a PR to `alpha`, so it's included with the next release.

_Initially submitted as https://github.com/Automattic/newspack-plugin/pull/1681._

Test on a site with Site Kit configured and Universal GA connected, but GA4 not connected.  

1. Confirm that the page is not sending hits to the GA4 property on `master`*
1. Visit Site Kit's Analytics settings page (`/wp-admin/admin.php?page=googlesitekit-settings#/connected-services/analytics`) and observe that GA4 is connected 
3. Visit an AMP page and confirm the GA4 snippet is inserted (search for `GA4_MEASUREMENT_ID` in the page source) and GA4 property is hit**

\* Just one `POST` request with `pageview` in payload to `google-analytics.com` should be sent
\** So there should be two of these requests, one for the old Universal GA and one for GA4

### Other information:

### DIFF
--- a/includes/plugins/google-site-kit/class-googlesitekitanalytics.php
+++ b/includes/plugins/google-site-kit/class-googlesitekitanalytics.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Google Site Kit integration class.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Modules\Analytics_4;
+use Google\Site_Kit\Modules\Analytics_4\Settings;
+use Google\Site_Kit\Core\Modules\Module;
+use Google\Site_Kit\Core\Authentication\Clients\Google_Site_Kit_Client;
+use Google\Site_Kit_Dependencies\Google\Service\GoogleAnalyticsAdmin as Google_Service_GoogleAnalyticsAdmin;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class extending Site Kit's Module, in order to easily access GA data via
+ * Site Kit's Analytics Admin service.
+ */
+class GoogleSiteKitAnalytics extends Module {
+	public function register() { // phpcs:ignore Squiz.Commenting.FunctionComment.Missing
+		return true;
+	}
+	public function setup_info() { // phpcs:ignore Squiz.Commenting.FunctionComment.Missing
+		return true;
+	}
+
+	/**
+	 * Set up the Analytics Admin service, so the module can use it.
+	 *
+	 * @param Google_Site_Kit_Client $client Google client instance.
+	 */
+	protected function setup_services( Google_Site_Kit_Client $client ) {
+		return array(
+			'analyticsadmin' => new Google_Service_GoogleAnalyticsAdmin( $client ),
+		);
+	}
+
+	/**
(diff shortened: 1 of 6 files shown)


includes/plugins/google-site-kit/class-googlesitekitanalytics.php
```diff
@@ -0,0 +1,84 @@
...
+use Google\Site_Kit\Core\Authentication\Clients\Google_Site_Kit_Client;
+use Google\Site_Kit_Dependencies\Google\Service\GoogleAnalyticsAdmin as Google_Service_GoogleAnalyticsAdmin;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class extending Site Kit's Module, in order to easily access GA data via
+ * Site Kit's Analytics Admin service.
+ */
+class GoogleSiteKitAnalytics extends Module {
```
miguelpeixe, line 23: If Site Kit is not installed and active, it throws a fatal error. There should be a proper check before importing this.

miguelpeixe (changes requested):
I was not able to test the setup of a new GA4. Setting up Site Kit with Analytics creates the GA4 automatically.

Do you recommend a way of testing the "automatic setup" outside of what Site Kit already provides?

includes/plugins/google-site-kit/class-googlesitekitanalytics.php
adekbadek: Done in 4768ceb6bbe31e7f62e8066f4f55b7e75ca292a8

adekbadek:
> I was not able to test the setup of a new GA4. Setting up Site Kit with Analytics creates the GA4 automatically.

You'd need a site set up before GA4 was available. You can remove the `googlesitekit_analytics-4_settings` option after setup to simulate that.

yogeshbeniwal:
@adekbadek From end user perspective, does this PR simply connects with existing GA 4 property for scenario where site kit is not connected to GA 4?

adekbadek:
@yogeshbeniwal Yes, that's it

includes/plugins/google-site-kit/class-googlesitekitanalytics.php
```diff
@@ -0,0 +1,84 @@
...
+			'analyticsadmin' => new Google_Service_GoogleAnalyticsAdmin( $client ),
+		);
+	}
+
+	/**
+	 * Return data needed to set up Site Kit's GA4 settings.
+	 *
+	 * @param string $account_id Account ID.
+	 */
+	public function get_ga4_settings( $account_id ) {
```
miguelpeixe, line 47: I worry this method, executed on `admin_init`, is very optimistic about a third-party structure. If `$this->get_service()` fails to return the expected object we'll have a fatal error on every dashboard page.

This could have `method_exists()` and `isset()` checks everywhere possible, just to be sure.

miguelpeixe (approved):
Working as expected. Left a non-blocking but important comment below.

includes/plugins/google-site-kit/class-googlesitekitanalytics.php
adekbadek: Good idea, done in 5885ce22d474aa9754786c755403abb87bf23b00

PR #1709, opened 2022-06-13
fix(reader-revenue): disable WC email if module will send email

### LINKED ISSUE #1699: Donations – two emails are sent (in a Stripe & WC setup)
## As is 

After a donation with Stripe as Reader Revenue platform, if WooCommerce is active, the donor will receive two emails: one from Newspack ("Thank You" subject) and one from WooCommerce ("Your … order is now complete").

## To be 

In this setup, only the Newspack email should be sent. WooCommerce email should be suppressed.

adekbadek:
### All Submissions:

Closes #1699

1. On `master`, activate WooCommerce and configure Stripe as Reader Revenue platform
2. Make a donation
3. Observe two emails are sent – one from Newspack ("Thank You" subject), and one from WC ("Your <site_title> order is now complete")
4. Switch to this branch, observe only the former is sent

### Other information:

### DIFF
--- a/includes/reader-revenue/class-stripe-connection.php
+++ b/includes/reader-revenue/class-stripe-connection.php
@@ -880,7 +880,7 @@ class Stripe_Connection {
 	 * @param bool $is_enabled True if enabled.
 	 */
 	public static function is_wc_complete_order_email_enabled( $is_enabled ) {
-		if ( Reader_Revenue_Emails::supports_emails() ) {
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
Works, but also suppresses the WC email if "Newspack" is the Reader Revenue platform. The email should only be suppressed if the platform is Stripe, right?

adekbadek:
Good catch! Fixed in 021bafc0f5fcb6adc3f858f03e47c2297bbdd9ee

dkoo (approved):
🙌

PR #1713, opened 2022-06-16
chore(deps): update newspack-scripts

adekbadek:
Similar to https://github.com/Automattic/newspack-theme/pull/1845, updates `prettier` and `stylelint` to latest versions:

- `newspack-scripts@4.0.0` was installed
- `wp-prettier` (aliased to `prettier`) were updated
- styles were migrated with `npx sass-migrator module assets/**/*.scss`
- styles were fixed with `npm run lint:scss -- --fix`
- JS was fixed with `npm run lint:js -- --fix` 

## How to test these changes

- take a look at the automated changes
- verify all jobs pass in CI


laurelfulford (changes requested):
For the CSS, I diff'd the files against the ones generated in trunk, and the only differences I noticed are a couple tweaks that have happened in the plugin since this PR was created.

The JavaScript was trickier because the generated version of some of the files have changed  a lot. But spot-checking the updates in GitHub, everything looks good, and the JavaScript files build without error. 

One thing I did notice in the Sass was some consistency differences with the colours -- in the theme and Newspack Blocks, the colours coming from the plugin/theme always use the namespace `colors`, but in this plugin they use `colors`, `colors2`, and `scss-colors` because `~@wordpress/base-styles/colors` is prefixed with `colors` when its used: 

```
@use '../../../shared/scss/colors' as colors2;
@use '~@wordpress/base-styles/colors';
```

```
@use '../../../shared/scss/colors' as scss-colors;
@use '~@wordpress/base-styles/colors';
```

I think this will get confusing in the long-term, especially when making edits from file to file. if possible, what do you think about giving the WordPress colours the unique namespace (`wp-colors`?) and using `colors` for the variables coming from the plugin? Something like that will make it consistent across files in the plugin, as well as consistent with the themes and Newspack Blocks. 

I'm happy to tackle the above if that would help! I can also take another look at this PR if you'd like to rebase it against master, just so it's more up to date. 

Just let me know if you have any questions about the above -- thanks!

adekbadek:
@laurelfulford – 

> what do you think about giving the WordPress colours the unique namespace (wp-colors?) and using colors for the variables coming from the plugin?

Good idea, done in 61c9e37

assets/wizards/advertising/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '~@wordpress/base-styles/colors';
+@use '~@wordpress/base-styles/colors' as wp-colors;
```
laurelfulford, line 1: I think it's safe to remove this, since none of the `wp-colors` variables are used in this file.

assets/wizards/site-design/views/main/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../../../shared/scss/colors';
+@use '../../../../shared/scss/colors';
```
laurelfulford, line 1: I think it should be safe to remove this, since none of the colour variables from the plugin are used in this file.

laurelfulford (changes requested):
Thanks @adekbadek! 

I did another pass and found some nitpicks -- I think only the first one (switching an `@import` to `@use`) is required, though. 

The rest are just cases where it looks like a file is being added with `@use`, but the variables coming from that file aren't actually being used -- so not a blocker at all! 

JavaScript looks good overall to me, and even with the one Sass `@import` remaining, there are no build issues (possibly because this is one case where the `@import` is not importing a file with Sass variables?). 

Just let me know if you have any questions about this at all -- thanks!

assets/blocks/reader-registration/editor.scss
```diff
@@ -1,4 +1,4 @@
-@import './style.scss';
+@import './style';
```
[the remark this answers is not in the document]
adekbadek: Done in fa8dc6699d7f3726ec14ef3057596338ce7d230d

assets/components/src/category-autocomplete/style.scss
```diff
@@ -2,7 +2,7 @@
  * Category Autocomplete
  */
 
-@import '../../../shared/scss/colors';
+@use '../../../shared/scss/colors';
```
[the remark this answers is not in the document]
adekbadek: Done in fa8dc6699d7f3726ec14ef3057596338ce7d230d

assets/components/src/info-button/style.scss
```diff
@@ -2,7 +2,7 @@
  * Info Button
  */
 
-@import '~@wordpress/base-styles/colors';
+@use '~@wordpress/base-styles/colors' as wp-colors;
```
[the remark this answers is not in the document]
adekbadek: Done in fa8dc6699d7f3726ec14ef3057596338ce7d230d

assets/wizards/advertising/style.scss
adekbadek: Done in fa8dc6699d7f3726ec14ef3057596338ce7d230d

assets/wizards/site-design/views/main/style.scss
adekbadek: Done in fa8dc6699d7f3726ec14ef3057596338ce7d230d

adekbadek:
PHP tests fail for network reasons, but the changes in this PR do not touch PHP. So nothing to worry about.

laurelfulford (approved):
Thanks @adekbadek! 

This all looks good to me 🚢  out of an abundance of caution I re-diffed the CSS files after these changes, they're still 1:1 (outside of the `currentcolor` case changes).

PR #1724, opened 2022-06-23
feat(reader-activation): registration block

miguelpeixe:
### All Submissions:

Implements a simple reader registration block integrated with a reader activation frontend js library. The block features should be later expanded to include:

 - Optional newsletter subscription through API provided by newspack-newsletters with:
   - Default list to assign the reader to, or;
   - Allow the reader to select from a list
 - Social login with Google and Facebook

image

image

## Reader Activation Frontend Library

This PR also introduces a small js library to allow easy communication between current and future reader activation front-end components on XHR requests.

Given the optimistic quality of the reader activation's auth intention, once a reader registers through the block, the `readerActivation.setReader( { email } );` is called regardless of authenticating or not. This method dispatches the `reader` event, which can be listened like this:

```js
readerActivation.on( 'reader', ( { detail: { email } } ) => {
	if ( email ) {
		console.log( "Reader updated", email );
	}
} );
```

The library also provides `readerActivation.off( 'reader', cb );` method to remove the listener and `readerActivation.getReader();` to retrieve the currently stored reader data.

In the near future, this library should also be used to update the component that links the user to authenticate or access the reader account page, providing dynamic and appropriate messaging given the authentication state.

## AMP Plus tweak

Similar to what has already been implemented for async/defer, this PR also introduces the support for enabling AMP Plus to a script by using:

```php
\wp_script_add_data( $handle, 'amp-plus', true );
```

This serves as a shortcut to properly add the `data-amp-plus-allowed` attribute to the script.

1. Confirm you have AMP with AMP Plus enabled and WooCommerce installed for the tests
2. Without `define( 'NEWSPACK_EXPERIMENTAL_READER_ACTIVATION', true );` on your `wp-config.php`, look for the Reader Registration block on a new draft and confirm it's not available, as it shouldn't without reader activation enabled.
3. Add `define( 'NEWSPACK_EXPERIMENTAL_READER_ACTIVATION', true );` and confirm you can add the block
4. Save the page, visit and register an email that doesn't exist on the database
5. Confirm the success message, visit Woo's `my-account` page and confirm you are logged in
6. Log out, register through the block again with the same email
7. Refresh the page and confirm the block does not render, as it identifies you as a registered reader
8. Visit Woo's `my-account` page and confirm you're prompted to authenticate
9. Disable AMP Plus, clear your cookies, register with pure AMP, and confirm you're redirected to a `?newspack_reader=1` page and you're registered (either with auth intention cookie or fully authenticated, depending on the email)

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -5,6 +5,11 @@ import { __ } from '@wordpress/i18n';
 import { TextControl, PanelBody } from '@wordpress/components';
 import { useBlockProps, InspectorControls } from '@wordpress/block-editor';
 
+/**
+ * Internal dependencies
+ */
+import './editor.scss';
+
 export default function ReaderRegistrationEdit( {
 	setAttributes,
 	attributes: { placeholder, label },
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -14,8 +14,6 @@ export default function ReaderRegistrationEdit( {
 	setAttributes,
 	attributes: { placeholder, label },
 } ) {
-	const defaultPlaceholder = __( 'Enter your email address', 'newspack' );
-	const defaultLabel = __( 'Register', 'newspack' );
 	const blockProps = useBlockProps();
 	return (
 		<>
@@ -24,13 +22,11 @@ export default function ReaderRegistrationEdit( {
 					<TextControl
 						label={ __( 'Input placeholder', 'newspack' ) }
 						value={ placeholder }
-						placeholder={ defaultPlaceholder }
 						onChange={ value => setAttributes( { placeholder: value } ) }
 					/>
 					<TextControl
 						label={ __( 'Button label', 'newspack' ) }
 						value={ label }
-						placeholder={ defaultLabel }
 						onChange={ value => setAttributes( { label: value } ) }
 					/>
 				</PanelBody>
@@ -38,8 +34,8 @@ export default function ReaderRegistrationEdit( {
 			<div { ...blockProps }>
 				<div className="newspack-reader-registration">
 					<form onSubmit={ ev => ev.preventDefault() }>
-						<input type="email" placeholder={ placeholder || defaultPlaceholder } />
-						<input type="submit" value={ label || defaultLabel } />
+						<input type="email" placeholder={ placeholder } />
+						<input type="submit" value={ label } />
 					</form>
 				</div>
 			</div>
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -3,7 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { TextControl, PanelBody } from '@wordpress/components';
-import { InspectorControls } from '@wordpress/block-editor';
+import { useBlockProps, InspectorControls } from '@wordpress/block-editor';
 
 /**
  * Internal dependencies
@@ -16,6 +16,7 @@ export default function ReaderRegistrationEdit( {
 } ) {
 	const defaultPlaceholder = __( 'Enter your email address', 'newspack' );
 	const defaultLabel = __( 'Register', 'newspack' );
+	const blockProps = useBlockProps();
 	return (
 		<>
 			<InspectorControls>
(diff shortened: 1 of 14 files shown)


assets/blocks/reader-registration/index.php
```diff
@@ -0,0 +1,149 @@
...
+	 * Fires after a reader is registered through the Reader Registration Block.
+	 *
+	 * @param string              $email   Email address of the reader.
+	 * @param int|false|\WP_Error $user_id The created user ID in case of registration, false if not created or a WP_Error object.
+	 */
+	\do_action( 'newspack_reader_registration_form_processed', $email, $user_id );
+
+	if ( \wp_is_json_request() ) {
+		if ( ! \is_wp_error( $user_id ) ) {
+			$message = __( 'Thank you for registering!', 'newspack' );
```
adekbadek, line 133: This should be customisable, but can be tackled in a separate PR. It also should be editable as blocks, not just as a string.

assets/blocks/reader-registration/edit.js
```diff
@@ -0,0 +1,48 @@
...
+ * Internal dependencies
+ */
+import './editor.scss';
+
+export default function ReaderRegistrationEdit( {
+	setAttributes,
+	attributes: { placeholder, label },
+} ) {
+	const defaultPlaceholder = __( 'Enter your email address', 'newspack' );
+	const defaultLabel = __( 'Register', 'newspack' );
```
adekbadek, line 18: There's no need to define `defaultPlaceholder` & `defaultLabel` – default values are already defined in `block.json`.

assets/blocks/reader-registration/index.php
```diff
@@ -0,0 +1,149 @@
...
+	if ( isset( $attrs['align'] ) && ! empty( $attrs['align'] ) ) {
+		$classes[] = 'align' . $attrs['align'];
+	}
+	if ( isset( $attrs['className'] ) ) {
+		array_push( $classes, $attrs['className'] );
+	}
+	if ( is_array( $extra ) && ! empty( $extra ) ) {
+		$classes = array_merge( $classes, $extra );
+	}
+	if ( ! empty( $attrs['hideControls'] ) ) {
```
adekbadek, line 101: There's no `hideControls` attribute on the block, what's it for here?

assets/blocks/reader-registration/index.php
```diff
@@ -0,0 +1,149 @@
...
+	</div>
+	<?php
+	return ob_get_clean();
+}
+
+/**
+ * Utility to assemble the class for a server-side rendered block.
+ *
+ * @param array $attrs Block attributes.
+ * @param array $extra Additional classes to be added to the class list.
```
adekbadek, line 86: This argument is not used, is it intended for future use?

assets/blocks/reader-registration/index.php
```diff
@@ -0,0 +1,149 @@
...
+add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\\enqueue_scripts' );
+
+/**
+ * Render Registration Block.
+ *
+ * @param array[] $attrs Block attributes.
+ */
+function render_block( $attrs ) {
+	if ( \is_user_logged_in() || Reader_Activation::get_auth_intention_value() ) {
+		return;
```
adekbadek, line 67: Empty render might be misleading, if the block is a part of a larger piece of UI (e.g. a prompt). I'd propose displaying the post-submission state by default ("Thank you for registering"), with a block attribute option to render empty if reader has identified themself.

assets/blocks/reader-registration/editor.scss
```diff
@@ -0,0 +1,34 @@
+@import './style.scss';
+
+.newspack-reader-registration {
```
adekbadek, line 3: It's weird that the block does not get theme's styles, but I'm not sure what we can do about it.

adekbadek (changes requested):

adekbadek:
Moving the discussion from https://github.com/Automattic/newspack-plugin/issues/1748 here. 

The input should have a help text, communicating that a login link will be sent to the supplied email address. The UI should clearly communicate that, so 

- reader is not confused when they get an email 
- reader is not looking for an option to get a login link sent

If that can be done in scope of this RP, then https://github.com/Automattic/newspack-plugin/issues/1748 can be closed.

assets/blocks/reader-registration/index.php
miguelpeixe: It was reminiscent from another code. Removed in 0a0185f68b77877154b7085d388f5232f37929bf

assets/blocks/reader-registration/index.php
miguelpeixe: No need, removed in 0a0185f68b77877154b7085d388f5232f37929bf

assets/blocks/reader-registration/editor.scss
miguelpeixe: I wondered about that too. I intend to continue investigating a better solution for the next step of this block.

assets/blocks/reader-registration/edit.js
miguelpeixe: You're right! Fixed in bce351eec2565c2624a380f7ae6abdae9347a17f

assets/blocks/reader-registration/index.php
miguelpeixe: Among other improvements, 1e72eede288a658f0c0cd54dabf18cb3f17acc23 implements the "Thank you for registering!" message to authenticated, pre-authenticated and non-JS form requests.

miguelpeixe:
1e72eede288a658f0c0cd54dabf18cb3f17acc23 implements a better handling of the form response.

The `process_form()` function sends a custom message to `send_form_response()` in case the reader already exists. I chose the message "Check your email for a confirmation link!" as it's a more generic tone and doesn't imply directly that a user account already exists. Am open to suggestions.

PR #1761, opened 2022-07-06
Stripe billing portal link

### LINKED ISSUE #1739: My Account: update password
Adding a password update section in WC's "My Account" (`/my-account`) page.

### LINKED ISSUE #1740: My Account: cancelling subscription
Cancelling a subscription in WC's "My Account" (`/my-account`) page.

### LINKED ISSUE #1741: My Account: display billing history
Billing history section in WC's "My Account" (`/my-account`) page.

### LINKED ISSUE #1742: My Account: update payment details
Adding a payment details updating section in WC's "My Account" (`/my-account`) page.

### LINKED ISSUE #1782: My Account: passwordless flow & updating password
If a user logs in using a magic link, and they've never set a password, the "My Account" UI for password updating should be used to _set_ a password, not update it.

adekbadek:
### All Submissions:

Adds an integration with WooCommerce's "My Account" page when Stripe is set as Reader Revenue platform.

The new "Billing" page only contains a link to Stripe's billing portal, and it's pretty bland:

image

Any thoughts here, @thomasguillot ?

In the future, we might implement such a billing portal on-site (https://github.com/Automattic/newspack-plugin/issues/1764).

Closes #1742
Closes #1739
Closes #1740
Closes #1741
Closes #1782 

1. Switch Newspack Theme to https://github.com/Automattic/newspack-theme/pull/1875
1. Ensure that required WooCommerce plugins are active, Stripe is configured as the Reader Revenue platform, and site is reachable for webhooks 
3. Ensure that `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` flag is set to `true`
4. As a non-logged-in and non-existing user, make a donation
5. Observe that `/my-account` page redirects to "Account Details"
6. Observe three links are available in "My Account" nav – "Account Details", "Billing", and "Logout"
7.  Go to "Billing" and click the "Update payment details" link – observe it opens Stripe customer portal, where you can update the CC details, cancel and edit subscriptions, and see billing history
9. Observe that password reset link can be sent in "Account Details" part in order to set up or reset password

### Other information:


thomasguillot:
I think https://github.com/Automattic/newspack-plugin/issues/1764 will be the key. At the moment, it's fine as it is although I'd probably have "Update payment details,..." as text and add an external link with something like "Go to Stripe".

miguelpeixe:
How are these changes affecting #1739?

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * Filter "My Account" items, if Stripe is the donations platform.
+	 *
+	 * @param array $items Items.
+	 */
+	public static function my_account_menu_items( $items ) {
+		if ( ! Donations::is_platform_stripe() ) {
+			return $items;
+		}
+		$disabled_wc_menu_items = [ 'dashboard', 'downloads', 'members-area', 'subscriptions', 'edit-address', 'orders', 'payment-methods' ];
```
dkoo, line 50: One thing to mention...not many publishers are using it because it's still experimental, but the Newspack Listings plugin's "self-serve" features rely on the Orders and Subscriptions menus for managing purchased listings. We might want to discuss how both feature sets can properly coexist, but in the meantime we can simply leave Orders and Subscriptions menu items active by doing something like:

```php
		if ( class_exists( '\Newspack_Listings\Products' ) && \Newspack_Listings\Products::is_active() ) {
			$disabled_wc_menu_items = array_values(
				array_filter(
					$disabled_wc_menu_items,
					function( $key ) {
						return 'orders' !== $key && 'subscriptions' !== $key;
					}
				)
			);
		}
```

or by applying a filter to the `$disabled_wc_menu_items` variable so the Listings plugin and others can allow their required menu items to remain.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * Add the necessary endpoints to rewrite rules.
+	 */
+	public static function add_rewrite_endpoints() {
+		if ( ! Donations::is_platform_stripe() ) {
+			return;
+		}
+		add_rewrite_endpoint( self::BILLING_ENDPOINT, EP_PAGES );
+		if ( ! get_option( '_newspack_has_set_up_custom_billing_endpoint' ) ) {
+			flush_rewrite_rules(); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.flush_rewrite_rules_flush_rewrite_rules
```
dkoo, line 141: For some reason this didn't work for me programmatically. Until I manually flushed permalinks, the Billing page was 404ing for me.

dkoo (changes requested):
This mostly works as described. I think the Account Details UI leaves a bit to be desired, but I can see that it's coming straight from WooCommerce, so until we do a wider overhaul it's okay to leave these. Just noting for future reference:

* We're authenticating a user without password in our registration flow, so it's strange to require the current password in order to set a new one. The "forgot password" flow works, but it's counterintuitive because you need to log out first to access it. In my mind we should replace this UI with the "forgot password" link to reset the password via email for everyone.
* If you enter partial info or weak passwords in the password change UI, the "Save" button is disabled, but there's no visual indication of that.
* The default display name description doesn't really make sense in our flow: " This will be how your name will be displayed in the account section and in reviews"
  - If there's a filter for this text, we should remove it entirely or change it to something more Newspack-appropriate, e.g. "This is how your name is displayed across <site name>."

A couple other minor issues:

dkoo:
> How are these changes affecting #1739?

@miguelpeixe This PR technically fulfills the requirement to let users set a password via WooCommerce's existing "Change password" UI, but that requires the user to know their current password first, which they won't after being authenticated via our registration flow. I think it would be cleaner to replace WC's "Change password" UI with the "Forgot password" link instead, if possible. The more I think about this the more it seems like a launch requirement for me since the current flow for new users to set a password is extremely counterintuitive.

We could leave #1739 open and tackle this in a new PR, perhaps?

adekbadek:
> How are these changes affecting #1739?

> We could leave #1739 open and tackle this in a new PR, perhaps?

https://github.com/Automattic/newspack-plugin/issues/1782 addresses the password & auth-intention flow issue. This PR does close #1739, or rather it's a non-issue since that functionality is provided by WC already.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
adekbadek: Hm, that weird. Are you sure this line was executed? And is there any other way to flush permalinks you're aware of?

dkoo: I don't know of another way to flush permalinks. At any rate, I tested on a new test site and didn't encounter this error, so let's consider it a non-issue. How I tested:

1. Create a new test site, install this PR, WooCommerce, and required extensions
2. Turn on required AMP Plus and Reader Activation flags in `wp-config.php`
3. Set up Stripe as the Reader Revenue platform
4. In a new session, made a recurring donation
5. Visited My Account > Billing, confirmed no 404

dkoo (changes requested):
One more thing, then I think this is good to merge.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -76,6 +79,42 @@ public static function my_account_menu_items( $items ) {
...
+	public static function handle_password_reset_request() {
+		if ( ! is_user_logged_in() ) {
+			return;
+		}
+		$nonce = filter_input( INPUT_GET, self::RESET_PASSWORD_URL_PARAM, FILTER_SANITIZE_STRING );
+
+		if ( $nonce ) {
+			$is_error = false;
+			if ( wp_verify_nonce( $nonce, self::RESET_PASSWORD_URL_PARAM ) ) {
+				$result  = retrieve_password( wp_get_current_user()->user_login );
```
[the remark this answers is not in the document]
adekbadek: Ugh, yeah. I don't think we can ever rely on `user_login` being a valid email, it will be escaped. Updated in 11aaeeffe49f27b985b29534e29f6461eedfb209 to use `user_email`.

dkoo (approved):
We're certain that WP users will always have a `user_email` property? If so, then ✅ !

PR #1754, opened 2022-07-01
feat(reader-activation): account link and auth form

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Implements a multi-strategy sign-in form and an always-visible responsive account/sign-in link.

**The form only signs in users identified as readers.**

If a reader is not fully authenticated, clicking on the link triggers the auth form. If the reader is authenticated, the link redirects to WC's "My Account" page.

The form button uses the theme's `primary_color` or Newspack's `#36f` as a fallback.

If a reader is pre-authenticated (auth intention), the form email input is automatically filled with the stored email.

## Desktop

The account link is displayed hooking into the Newspack theme's `tertiary-menu` location. In case the location doesn't have a menu assigned to it, it'll hook as a custom HTML using `fallback_cb`.

image

image

image

image

## Mobile

The account link is displayed hooking into the Newspack theme's [`newspack_header_after_mobile_toggle` custom action](https://github.com/Automattic/newspack-theme/pull/1875).

image

image

image

Closes #1746 
Closes #1749
Closes #1744

## Form next steps

 - Third-party auth
 - Newsletters integration for registration
 - Reader Activation settings management on the Engagement wizard for:
   - Toggling account link visibility
   - Configuring default messages (privacy policy)
  

### How to test the changes in this Pull Request:

1. Check out this branch and https://github.com/Automattic/newspack-theme/pull/1875
2. Make sure you have reader activation enabled and AMP plugin off (we'll later test AMP Plus)
3. Visit the homepage anonymously and confirm you see the "Sign In" link correctly placed
4. Test authenticating your admin account and confirm it displays an error
5. Test authenticating a reader through auth link and password and confirm the account link UI is updated
6. Click on the account link after authenticating and confirm you are redirected to the "My Account" page
7. Turn AMP on with AMP Plus
8. Repeat the previous steps and confirm the form works as expected
9. Visit the site anonymously on a mobile viewport
10. Test authentication and confirm the form behaves as expected

[...]

PR #1778, opened 2022-07-13
feat(registration-block): newsletter subscription

### LINKED ISSUE #1738: Registration block: allow for newsletter signup
A feature for the Registration Block (https://github.com/Automattic/newspack-plugin/pull/1724): when the user registers, they should be presented with a list of newsletters to sign up for at the same time.

miguelpeixe:
### All Submissions:

Implements newsletter subscription support for the Registration Block as provided by Newspack Newsletters ESP-agnostic API.

The block supports 2 style variations but the stacked version is prioritized on smaller viewports.

image

### Multiple newsletters with narrow width

This is also the default layout for the mobile viewport.

image

### Multiple newsletters with wider width

image

### Single newsletter

The checkbox label is editable to represent a more generic newsletter subscription messaging.

image

## Column layout

This layout is only displayed on the desktop viewport and with newsletter subscriptions enabled.

### Multiple newsletters

image

### Single newsletter

image

Closes #1738

1. Make sure you have AMP Plus and Reader Activation enabled (`define( 'NEWSPACK_EXPERIMENTAL_READER_ACTIVATION', true );`)
2. Also confirm you have newsletter subscription lists configured on Newspack Newsletters (Engagement wizard)
3. Place a Registration Block on a new page and enable Newsletter subscription
4. With one newsletter selected, confirm you can update the newsletter checkbox label through the block's rich text editor
5. Select multiple newsletters and confirm the UI is updated
6. Save the block using the default "Stacked" layout and test it on the frontend
7. In an incognito window, confirm the registration works and the email is subscribed to the selected lists on your ESP
8. Change your layout to "Column" and confirm the layout is as expected
9. Register again and confirm it also works as expected

Note: I intend to tackle the success message layout next, not in this PR.

#### Editable text

Confirm you can edit the privacy text, "Newsletters" title (displayed on Column layout with multiple newsletters), and Button label through the in-block rich text editor.

### Other information:


assets/blocks/reader-registration/index.php
```diff
@@ -63,12 +78,24 @@ function enqueue_scripts() {
...
 	$message    = '';
+
+	/** Setup list subscription */
+	if ( $attrs['newsletterSubscription'] && method_exists( 'Newspack_Newsletters_Subscription', 'get_lists_config' ) ) {
+		$list_config     = \Newspack_Newsletters_Subscription::get_lists_config();
+		$lists           = array_keys( $list_config );
+		$list_map        = array_flip( $lists );
+		$available_lists = array_values( array_intersect( $lists, $attrs['lists'] ) );
+		if ( empty( $available_lists ) ) {
+			$available_lists = [ $lists[0] ];
```
dkoo, line 92: If you don't have any lists active in the Engagement wizard, this will throw a PHP warning:

```
PHP Notice:  Undefined offset: 0 in /newspack/app/public/wp-content/plugins/newspack-plugin/assets/blocks/reader-registration/index.php on line 92
```

On the front-end, the same warning will be thrown and the box with the newsletter lists is present but empty:

Screen Shot 2022-07-18 at 4 04 22 PM

Expected behavior in both cases should be the same as if the "Enable newsletter subscription" option is disabled.

assets/blocks/reader-registration/edit.js
```diff
@@ -12,30 +21,197 @@ import './editor.scss';
...
+							<>
+								{ inFlight && <Spinner /> }
+								{ ! inFlight && ! Object.keys( listConfig ).length && (
+									<div style={ { marginBottom: '1.5rem' } }>
+										<Notice isDismissible={ false } status="error">
+											{ __( 'You must enable lists for subscription.', 'newspack-newsletters' ) }
+										</Notice>
+									</div>
+								) }
+								{ lists.length === 1 && (
```
dkoo, line 98: Non-blocking: I'm not sure we need to have the editable fields in both the WSYIWYG editor and the sidebar.

assets/blocks/reader-registration/edit.js
```diff
@@ -12,30 +21,197 @@ import './editor.scss';
...
+											placeholder={ __( 'Register', 'newspack' ) }
+											value={ label }
+											tagName="span"
+										/>
+									</button>
+								</div>
+								<div className="newspack-registration__privacy">
+									<RichText
+										onChange={ value => setAttributes( { privacyLabel: value } ) }
+										placeholder={ __( 'Newsletters title…', 'newspack' ) }
```
dkoo, line 208: This placeholder text should more accurately reflect the purpose:

```suggestion
										placeholder={ __( 'Terms & Conditions statement…', 'newspack' ) }
```

assets/blocks/reader-registration/edit.js
```diff
@@ -12,30 +21,197 @@ import './editor.scss';
...
+								</div>
+							) : null }
+							<div className="newspack-registration__main">
+								<div className="newspack-registration__inputs">
+									<input type="email" placeholder={ placeholder } />
+									<button type="submit">
+										<RichText
+											onChange={ value => setAttributes( { label: value } ) }
+											placeholder={ __( 'Register', 'newspack' ) }
+											value={ label }
```
dkoo, line 200: If `label` is empty, the button will show `Register` by default in the editor, but the button will display without any label on the front-end. We might consider showing `Register` on the front-end too, if this attribute is empty.

assets/blocks/reader-registration/view.js
```diff
@@ -4,54 +4,68 @@
...
-				headers: {
-					Accept: 'application/json',
-				},
-				body,
-			} ).then( res => {
-				submitElement.disabled = false;
-				res.json().then( ( { message, data } ) => endLoginFlow( message, res.status, data ) );
+			const endLoginFlow = ( message, status, data ) => {
+				const messageNode = document.createElement( 'p' );
+				messageNode.innerHTML = message;
```
dkoo, line 28: `innerHTML` might be a security concern—can we use `textContent` instead?

dkoo (changes requested):
Exciting to see this in action! Ran into a few minor issues below.

assets/blocks/reader-registration/index.php
miguelpeixe: Fixed in 43cc118a07d354986a0009eb0dd638cc5d9c1286

assets/blocks/reader-registration/edit.js
miguelpeixe: I agree! 155d24a06759eabc3e6c50ceacdc07953acb21ff removes the duplicate controls from the panel.

Which overrides are you referring to? If there's one list selected, its label should no longer be the list title and be a more direct generic newsletter subscription message. The same for the editable "Newsletters" title, which should only be available in the column layout and multiple newsletters.

assets/blocks/reader-registration/edit.js
miguelpeixe: Nice catch! d7780e72094a087908c1bd4be5a4f84d0dd49d27 implements handling default attributes

assets/blocks/reader-registration/edit.js
```diff
@@ -12,30 +21,205 @@ import './editor.scss';
...
+																	placeholder={ __( 'Subscribe to our newsletter', 'newspack' ) }
+																	value={ newsletterLabel }
+																	tagName="span"
+																/>
+															) : (
+																listConfig[ listId ]?.title
+															) }
+														</span>
+														{ displayListDescription && (
+															<span className="newspack-registration__lists__description">
```
dkoo, line 170: Now that this is in the WYSIWYG `RichText` editor only, it needs a placeholder in case the value is empty, otherwise it becomes 0px high and uneditable.

assets/blocks/reader-registration/index.php
dkoo: Editor looks good now, but on the front-end I still see this :ghost: form:

Screen Shot 2022-07-19 at 4 18 44 PM

assets/blocks/reader-registration/edit.js
dkoo: The editable label + description within the block is what I'm calling the "overrides", because they override the list name and description you set on the list itself in the Engagement wizard. I get the logic now, but I anticipate some initial publisher confusion about this.

dkoo (changes requested):
Getting close! A few minor issues remain.

assets/blocks/reader-registration/edit.js
miguelpeixe: I'm not sure I follow. The list description (or title) is not editable through `RichText`

assets/blocks/reader-registration/index.php
miguelpeixe: I was not able to reproduce that. My steps were:

1. Make sure there are lists enabled and with values on the wizard
2. Configure the block with all lists
3. Disable all lists on the engagement wizard
4. Visit rendered block on page:

image

While trying to understand what might have happened in your case, I pushed 8331ec05234dac04f00cc4dc6bbc245c0ee7df43. It forces a default value for the newsletter label in case there's only 1 newsletter selected. It was allowing the label to be left empty.

miguelpeixe:
After merging #1754 I had to change the strategy used for the dom load callback. `window.onload` gets overridden and `domReady()` is a more reliable strategy.

5825018bf5fb2aa194742977cf3036a7d3b1187b implements `domReady()` but I had to place the code twice for each script. Don't know why importing either through `@wordpress/dom-ready` or a custom `./utils/dom-ready` does not work and webpack renders `undefined` for it. Could use some help figuring this out, can be tackled separately though. cc @adekbadek

assets/blocks/reader-registration/edit.js
dkoo: When only a single list is enabled, the title and description are editable via `RichText`. If the description attribute is empty, the `RichText` field has zero height, so it's not possible to click into it in order to edit the text.

miguelpeixe: The description is never editable through `RichText`. When there's only 1 list, the title of the newsletter is replaced by a `newsletterLabel`, with the placeholder `Subscribe to our newsletter`.

The description remains the one configured and retrieved from the lists, if it doesn't exist it will remain empty.

assets/blocks/reader-registration/index.php
dkoo: Hmm, when I replicate using your steps now, I see this on the front-end:

Screen Shot 2022-07-20 at 1 52 15 PM

But since there are no lists enabled in the block, what does the checkbox actually do? Shouldn't we hide the whole section if there are no lists selected?

dkoo: I think I see the discrepancy in our tests! Here's what I did:

1. Make sure there are lists enabled with values in the wizard
2. Configure the block with NO lists selected
3. Keep all lists enabled in the Engagement wizard
4. Visit rendered block on page

miguelpeixe: If there are configured lists but none were selected for the block, it'll use the first from the configured lists:

https://github.com/Automattic/newspack-plugin/blob/b210853675231a6f240ec847fd1e0c328ab14c80/assets/blocks/reader-registration/index.php#L111-L113

You should probably be seeing an error notice on your block configuration as well.

miguelpeixe: I know realize that this is counter-intuitive and doesn't reflect the block seen on the editor, 50fc58acab5d6d5ef4e31bba284482278cf35ebb removes this fallback.

dkoo: I get it now, thanks for clarifying. The error message does appear, but in the block editor preview the block doesn't display the newsletters section like it does on the front-end. The render should match in both contexts.

Also, in my mind it's more intuitive to make the block behave as if newsletter signups are disabled in this situation, but that's subjective and therefore non-blocking.

dkoo: :coke:

miguelpeixe: :coke: on the intuition

dkoo (approved):

dkoo:
Approved, but just [this comment](https://github.com/Automattic/newspack-plugin/pull/1778#issuecomment-1190513280) about `domReady`. Feel free to continue working on this branch if you'd like.

miguelpeixe:
We can tackle it separately, let's get this to `master` so we can continue making things work together.

PR #1780, opened 2022-07-14
Feat/reader activation tweaks

adekbadek:
### All Submissions:

A few tweaks related to ongoing Reader Activation work:

- slight refactor of Registration block
- setting client ID cookie in absence of AMP
- Reader Activation fix for AMP Plus mode

1. On `master`, turn off AMP and Newspack Campaigns, ensure `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` env. variable is `true`
2. Visit the site in a new session, observe the `newspack-cid` cookie is not set
3. In JS console, observe `window.newspack_reader_activation_data` is not defined
4. Switch to this branch, observe the cookie is set (to a random string value) and `window.newspack_reader_activation_data` is defined
5. Turn on AMP and Newspack Campaigns, observe that `newspack-cid` cookie value is not overridden 

### Other information:

### DIFF
--- a/includes/class-reader-activation.php
+++ b/includes/class-reader-activation.php
@@ -284,6 +284,7 @@ final class Reader_Activation {
 		\wp_set_current_user( $user->ID );
 		\wp_set_auth_cookie( $user->ID, true );
 		\do_action( 'wp_login', $user->user_login, $user );
+		Logger::log( 'Logged in user ' . $user->ID );
 
 		return $user;
 	}
@@ -362,6 +363,8 @@ final class Reader_Activation {
 			/** Add default reader related meta. */
 			\update_user_meta( $user_id, self::READER, true );
 			\update_user_meta( $user_id, self::EMAIL_VERIFIED, false );
+
+			Logger::log( 'Created new reader user with ID ' . $user_id );
 
 			if ( $authenticate ) {
 				self::set_current_reader( $user_id );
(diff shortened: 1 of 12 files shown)


miguelpeixe (approved):

includes/class-reader-activation.php
```diff
@@ -54,15 +54,17 @@ public static function enqueue_scripts() {
 			$reader_email = \wp_get_current_user()->user_email;
 		}
 		\wp_localize_script(
-			$handle,
+			self::SCRIPT_HANDLE,
 			'newspack_reader_activation_data',
 			[
 				'auth_intention_cookie' => self::AUTH_INTENTION_COOKIE,
+				'cid_cookie'            => NEWSPACK_CLIENT_ID_COOKIE_NAME,
+				'nonce'                 => wp_create_nonce( 'wp_rest' ),
```
miguelpeixe, line 62: @adekbadek I overlooked this line. What is the purpose of this nonce? It won't behave as expected for cached pages.

adekbadek: This is necessary to [make an authenticated request from the front-end](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/#cookie-authentication). Not actually used in this PR, but [in the Google login PR](https://github.com/Automattic/newspack-plugin/pull/1781/files#diff-e9abbf90df0b524d7bfa6b07b4f954dd6092b1f061027d9a5faa6ec28f8be3a1R78-R86) (which was branched off this PR's branch, initially).

> It won't behave as expected for cached pages.

Can you elaborate on that?

miguelpeixe: The nonce value will be cached as well, defeating the purpose of a number used once.

Do we need a nonce for `/wp-json/newspack/v1/login/google/register`? Why would that be susceptible to a CSRF attack?

Issue #1772, opened 2022-07-13
Authentication: Identity Platform Azure AD B2C


yogeshbeniwal:
**Is your feature request related to a problem? Please describe.**
Publishers use various other supporting services and they all have their own local login authentication such as comments.

**Describe the solution you'd like**
With readers features and related data stored in WordPress, integrate logins for all additional platforms used by publishers in one location. i.e. Using [Azure AD B2C](https://azure.microsoft.com/en-us/services/active-directory/external-identities/b2c/)

**Describe alternatives you've considered**
Using WordPress as authentication provider #1408, this is much better (doesn't need extra service) but may need more dev effort.

Updating To Dos after testing Reader Revenue features released so far:

- [ ] Disable Email verification of users, user's email is verified at IDP level
- [ ] Password reset link should use IDP api or redirect to IDP hosted page to trigger password reset
- [ ] Sync account details change such as First Name, Last Name and Email with IDP

adekbadek:
This is out of scope for this project, perhaps a WP plugin could serve this functionality.

yogeshbeniwal:
@adekbadek with recent roadmap announcement for federated websites, does this feature request become in scope?

adekbadek:
No, it doesn't.

yogeshbeniwal:
@adekbadek Thank you. So how it will works with integration between multiple wordpress instances, users will be authenticated with first wordpress as provider and another as client? Asking because currently finalising central user management for multiple brands under one entity.

adekbadek:
That's a great question! We don't yet have an answer though. This will be worked in in first half of 2023, probably.

PR #1781, opened 2022-07-14
Registration block – login with Google

### LINKED ISSUE #1774: Authentication: Google sign-in
Google as sign-in method for the Registration block.

adekbadek:
### All Submissions:

Adds Google Login to the Reader Registration block. Also adds a `np_reader_login_method` user meta, to note the login method. 

image

This implementation reuses a lot of the code used for processing admin-related Google OAuth used for GAM connection. The differences are:

- only `auth/userinfo.email` scope is requested 
- credentials are not saved, since there will be no actions performed on behalf of the user

Closes #1774 

1. Without configuring Google OAuth proxy (no `NEWSPACK_GOOGLE_OAUTH_PROXY` defined), observe that the Registration block does _not_ feature a "Sign in with Google" button
2. Configure Google OAuth proxy by providing a `NEWSPACK_GOOGLE_OAUTH_PROXY` env. variable
3. Observe the Reader Registration block allows for signing in via Google
4. After being authenticated, observe the "My Account" page is accessible
5. Observe in the DB that the `np_reader_login_method` user meta value is `'google'`

### Other information:


yogeshbeniwal:
@adekbadek Is NEWSPACK_GOOGLE_OAUTH_PROXY only available to Newspack hosted sites only?

adekbadek:
Yes, it's a closed service.

yogeshbeniwal:
@adekbadek Follow up question. Does that mean login with Google feature won't be available for open source/self hosting?

adekbadek:
That's correct.

includes/oauth/class-google-login.php
```diff
@@ -0,0 +1,182 @@
...
+			if ( is_wp_error( $result ) ) {
+				return $result;
+			}
+			Reader_Activation::save_current_user_login_method( 'google' );
+
+			return \rest_ensure_response(
+				[
+					'email'   => $email,
+					'message' => __(
+						'Thank you for registering!',
```
dkoo, line 171: Non-blocking suggestion: If the reader is logging in with a new session after having already registered with the same email in the past, can we show a "Thank you for signing in" message instead?

dkoo (changes requested):
The login flow works well when Google OAuth is configured correctly. I ran into one hiccup on my local test site, where trying to get the Google auth URL returned a `Cannot verify the request is coming from a Newspack site` WP error, resulting in a 404. More details below.

assets/blocks/reader-registration/view.js
```diff
@@ -53,5 +65,41 @@ import './style.scss';
...
+						} ),
+					} ).then( res => {
+						res.json().then( ( { message, data } ) => endLoginFlow( message, res.status, data ) );
+					} );
+				};
+				fetch( '/wp-json/newspack/v1/login/google' )
+					.then( res => res.json() )
+					.then( url => {
+						if ( authWindow ) {
+							authWindow.location = url;
```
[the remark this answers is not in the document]
adekbadek: Fixed in 484a3068112e470aa8cac5903d53b6e213d09bde

includes/oauth/class-google-login.php
adekbadek: Good idea, done in d1dba77ee7506393b7bf4a2b74f6225e8bf9af4d

includes/oauth/class-oauth.php
```diff
@@ -25,6 +25,17 @@ public static function get_proxy_api_key() {
...
+	/**
+	 * Get unique ID for user or client.
+	 */
+	private static function get_unique_id() {
+		$id = Reader_Activation::get_client_id();
+		if ( ! $id ) {
+			$id = get_current_user_id();
+		}
+		return $id;
+	}
```
miguelpeixe, line 37: This method is not ensuring that the reader receives a unique ID and there's a security concern.

`$id` should only be the current user's ID if we can verify that the user is logged in, and IMO it should be prioritized over any other method.

`Reader_Activation::get_client_id();` is user input. Since this is performed during the request (uncached) it is safer to have this method generate a random value, set it as cookie, and later check the validity.

assets/blocks/reader-registration/view.js
```diff
@@ -53,5 +65,45 @@ import './style.scss';
 				res.json().then( ( { message, data } ) => endLoginFlow( message, res.status, data ) );
 			} );
 		} );
+
+		const googleLogin = container.querySelector( '#newspack-google-login' );
```
dkoo, line 69: It's possible for multiple login forms to exist on the same page, e.g. in a prompt and within page content. We should change this to handle `container.querySelectorAll( '.newspack-google-login' );` (note it's a class name instead of an ID) and use `addEventListener` instead of `onclick` in case we need to add additional event handlers or have greater control over adding/removing this one.

assets/blocks/reader-registration/view.js
```diff
@@ -53,5 +65,45 @@ import './style.scss';
...
+								}, 500 );
+							} else {
+								endLoginFlow();
+							}
+						}
+					} );
+			};
+		}
 	} );
 } )( window.newspackReaderActivation );
```
dkoo, line 109: I think there's a race condition happening here. Sometimes when I load the page, `window.newspackReaderActivation` is `undefined`, which you can confirm by console.logging the value inside this file's IFEE closure. When this happens, the flow is unavailable and nothing happens when you click the "Sign in with Google" button.

assets/blocks/reader-registration/view.js
```diff
@@ -20,12 +20,25 @@ import './style.scss';
...
+
 		const endLoginFlow = ( message, status, data ) => {
+			messageElement.innerHTML = '';
+			container.classList.remove( 'newspack-registration--in-progress' );
+
+			if ( ! message || ! status ) {
+				return;
+			}
 			const messageNode = document.createElement( 'p' );
 			messageNode.innerHTML = message;
```
dkoo, line 37: `innerHTML` is a potential security concern—we should use `textContent` instead.

assets/blocks/reader-registration/view.js
```diff
@@ -53,5 +65,45 @@ import './style.scss';
...
+									if ( authWindow.closed ) {
+										checkLoginStatus();
+										clearInterval( interval );
+									}
+								}, 500 );
+							} else {
+								endLoginFlow();
+							}
+						}
+					} );
```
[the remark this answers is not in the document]
dkoo: This means the SSO buttons should move inside the registration form, right?

includes/oauth/class-oauth.php
dkoo: > `Reader_Activation::get_client_id();` is user input. Since this is performed during the request (uncached) it is safer to have this method generate a random value, set it as cookie, and later check the validity.

Does this value even need to be directly associated with the reader's identity? Can't it just be a random, unique string that can be verified within the same session, in which case we could use a `nonce`?

assets/blocks/reader-registration/view.js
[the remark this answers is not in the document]
dkoo: Looking better, now—when you click the button and the plugin can't start the Google auth flow for whatever reason, it shows the error message inline and doesn't open the auth window at all. 👍

dkoo (changes requested):
I noticed a few additional issues with the JS this time around. @miguelpeixe's excellent points might also mean the form itself and the JS might need to be refactored further to support the other plugins' needs.

includes/oauth/class-oauth.php
miguelpeixe: It's my understanding that it's what it does. A custom CSRF protection implementation uses this unique value to validate the request after it returns from the proxy/third party.

We must be careful on generating those, though. Including nonces. The initial state is anonymous and subjected to cache. This random value must be generated on the request and preferably by the backend to ensure uniqueness, otherwise we may get duplicate session tokens.

assets/blocks/reader-registration/index.php
```diff
@@ -180,9 +201,15 @@ function process_form() {
 		return send_form_response( $user_id );
 	}
 
+	$user_logged_in = false !== $user_id;
+
+	if ( $user_logged_in ) {
+		Reader_Activation::save_current_user_login_method( 'registration-block' );
```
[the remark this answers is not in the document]
adekbadek: That's for analytics – in the future we might send this as a custom dimension for GA, or just display in a local dashboard.

yogeshbeniwal:
@adekbadek Can we consider making login with google available for open source usage. Site kit plugin do it by making proxy site available in public domain. If that's not feasible, then making reader login compatible with Auth0 will do the job, created a issue #1772 for further discussion on this option.

assets/blocks/reader-registration/view.js
adekbadek: Done in 5867cfa2bef7bff21be6dbd49c1a77fc5c7f72de

assets/blocks/reader-registration/view.js
adekbadek: Done in 51ce912ecfec7ef3d2f38fbcf2ebc9570ece6dfe & 5030ffe46d38975f1d0a6bb5989e899e6b5b2842

assets/blocks/reader-registration/view.js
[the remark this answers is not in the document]
adekbadek: 1. The form and 3rd party login are fundamentally different flows. The form submission handler expects an email in the payload, which can't be supplied by the 3rd party flow, because it's not user input. 
2. Any side-effect, such as newsletter subscription, should be done with WP hooks. I've updated the code here to handle newsletter lists signup in bcb43117c0e9cfac71eea7f9b6b4db88645fde43, with a [corresponding change in Newsletters](https://github.com/Automattic/newspack-newsletters/pull/895)

includes/oauth/class-oauth.php
adekbadek: (nonces and caching discussion [has started here](https://github.com/Automattic/newspack-plugin/pull/1780#discussion_r922332663))

miguelpeixe: Although the mentioned discussion touch similar topics, the `nonce` used at `/wp-json/newspack/v1/login/google/register` the `$id` used at `Reader_Activation::get_client_id();` have different goals, right?

adekbadek: 1. _Nonces vs. caching_: removed nonce use in 9965c61c98cecd649d5d366a8c99f126b130fe43. I was originally introduced to authenticate the registration request, but after all the changes in the meantime, this seems not necessary anymore. The request to `/login/google/register` is fine without the nonce. 
2. _Unique id for 3rd party data transient_: I've changed it so it tries for user ID, then then the client ID cookie, and then as a last resort, the session id. [VIP-Coding-Standards discourages the use of `session_id`](https://github.com/Automattic/VIP-Coding-Standards/blob/e1eb7c9dbf93ebcbb20a610c6babe4b74725551d/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php#L168), but I'm not sure why. Session ID seems like an idea candidate here.

adekbadek:
> Can we consider making login with google available for open source usage.

We've no such plans ATM. 

> […] making reader login compatible with Auth0 will do the job, created a issue https://github.com/Automattic/newspack-plugin/issues/1772 for further discussion on this option.

Auth0 looks interesting, but for now we're focusing on handling data as a first party.

yogeshbeniwal:
> Auth0 looks interesting, but for now we're focusing on handling data as a first party.

Auth0 helps to keep data as first party + streamline user accounts across multiple platforms used by publishers. I have tested its plugin and it works well.

dkoo:
@yogeshbeniwal thanks for the suggestion, but that's not within the scope of this current work. Please feel free to file an issue for us to investigate as a separate item, though.

assets/blocks/reader-registration/view.js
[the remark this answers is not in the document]
miguelpeixe: After a more detailed discussion, I now agree that this is a good approach! 👍

assets/blocks/reader-registration/index.php
```diff
@@ -180,18 +180,38 @@ function render_block( $attrs ) {
...
+									<div></div>
+								</div>
+								<button class="newspack-registration__logins__google">
+									<?php echo file_get_contents( dirname( NEWSPACK_PLUGIN_FILE ) . '/assets/blocks/reader-registration/icons/google.svg' ); // phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+									<span>
+										<?php echo \esc_html__( 'Sign in with Google', 'newspack' ); ?>
+									</span>
+								</button>
 							</div>
+						<?php endif; ?>
```
miguelpeixe, line 204: This section should be inside `newspack-registration__inputs` to not be vertically centered on the flex layout of `newspack-registration__main`.

image

includes/class-reader-activation.php
```diff
@@ -737,10 +737,11 @@ public static function set_current_reader( $user_id ) {
 	 * @param string $email        Email address.
 	 * @param string $display_name Reader display name to be used on account creation.
 	 * @param bool   $authenticate Whether to authenticate after registering. Default to true.
+	 * @param object $metadata Any metadata to pass along to the action hook.
```
miguelpeixe, line 740:
```suggestion
	 * @param array  $metadata     Any metadata to pass along to the action hook.
```

Linting and proper variable naming, `object` is [something else](https://www.php.net/manual/en/language.types.object.php)

includes/class-reader-activation.php
```diff
@@ -737,10 +737,11 @@ public static function set_current_reader( $user_id ) {
...
 	 *
 	 * @return int|false|\WP_Error The created user ID in case of registration, false if the user already exists, or a WP_Error object.
 	 */
-	public static function register_reader( $email, $display_name = '', $authenticate = true ) {
+	public static function register_reader( $email, $display_name = '', $authenticate = true, $metadata = false ) {
```
miguelpeixe, line 744:
```suggestion
	public static function register_reader( $email, $display_name = '', $authenticate = true, $metadata = [] ) {
```

We can default this to an empty array, no reason to use boolean type here.

includes/class-reader-activation.php
```diff
@@ -821,10 +822,38 @@ public static function register_reader( $email, $display_name = '', $authenticat
 		 * @param bool           $authenticate  Whether to authenticate after registering.
 		 * @param false|int      $user_id       The created user id.
 		 * @param false|\WP_User $existing_user The existing user object.
+		 * @param false|any      $metadata Metadata.
```
miguelpeixe, line 825:
```suggestion
		 * @param array          $metadata      Metadata.
```

Type and linting

includes/oauth/class-google-oauth.php
```diff
@@ -339,17 +338,34 @@ private static function authenticated_user_basic_information() {
...
 		} else {
 			Logger::log( 'Failed retrieving user info – invalid credentials.' );
 			return new \WP_Error( 'newspack_google_oauth', __( 'Invalid Google credentials. Please reconnect.', 'newspack' ) );
 		}
+	}
+
+	/**
+	 * Authenticated user's basic information.
+	 *
+	 * @return object|WP_Error Basic information, or error.
```
miguelpeixe, line 352:
```suggestion
	 * @return array|WP_Error Basic information, or error.
```

This method also returns `false`, returning an empty array would be a better practice to avoid type errors in the future.

miguelpeixe (changes requested):

adekbadek:
I've noticed that the changes bricked the admin OAuth flow, fixed in 57fd298

miguelpeixe (approved):
🎉 

The flow is working for both dashboard service connection and reader authentication.

adekbadek:
@dkoo uncovered a race condition. The Reader Registration block is initiated when the DOM is ready, but it relies on the Reader Activation front-end library, which might not be available just yet due to network conditions. The block should await a "reader activation ready" event to prevent a race condition.

miguelpeixe:
We can tackle it on a separate PR, this issue was not originated here. We must also address #1792

assets/blocks/reader-registration/view.js
dkoo: This will still only attach the event listener and flow for the first instance of the Registration block on the page. This should really be a `document.querySelectorAll`, and then a `for` loop to attach the event listener to each matching element.

assets/blocks/reader-registration/view.js
dkoo: More details in [comment below](https://github.com/Automattic/newspack-plugin/pull/1781#issuecomment-1192719034).

assets/blocks/reader-registration/view.js
dkoo: Looks good, thanks!

dkoo (changes requested):
> > The block should await a "reader activation ready" event to prevent a race condition.

> We most likely don't need this, the reader activation library functions are created synchronously. It's the script tag that has the `async` attribute.

Great point, but wouldn't the event solve the issue even if one or both `script` tags have an `async` attribute?

miguelpeixe:
> Great point, but wouldn't the event solve the issue even if one or both script tags have an async attribute?

Feels overkill if it's just the order of execution of a dependency. The script is not fetched through XHR to require such sophistication.

assets/blocks/reader-registration/view.js
miguelpeixe: Given that the selector is queried from `container`, it's not the case. The `container` will always have only one `googleLogin`.

I would, though, avoid using an ID for this. If the page happens to render multiple blocks it'll create an ID conflict.

dkoo: > Given that the selector is queried from `container`, it's not the case. The `container` will always have only one `googleLogin`.

You're absolutely right! 51ce912 & 5030ffe did in fact change this ID to a class name, so I think we're good to go here.

dkoo (approved):

PR #1773, opened 2022-07-13
feat(reader-activation): settings wizard

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

image

Implements basic global settings for Reader Activation features. They include:

 - Toggling reader activation entirely
 - Toggling the always-visible account link on the menu (false by default to not disrupt sites' UI on official launch)
 - Newsletter subscription text on registration for static auth modal
 - Terms and conditions text and URL, also for static auth modal

### How to test the changes in this Pull Request:

1. Check out this branch, visit the front page and confirm the account link is not visible
2. Visit the Engagement wizard and toggle the account link
3. Visit the front page and confirm the account link is rendered and working as expected
4. Change the default texts and save
5. Visit the site, open the Sign In modal and confirm the texts are applied for terms and conditions and newsletter selection while registering
6. Disable reader activation on the wizard and confirm the features become unavailable (reader registration and the blocks)

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1785, opened 2022-07-15
feat(registration-block): editable success state

### LINKED ISSUE #1768: Registration Block: customisable post-registration state
After a user registers using the Registration Block, a text message replaces the registration UI. This state should be customisable (using the block editor, not just text).

adekbadek:
### All Submissions:

Because this code introduces a `save` handler on the block, a warning will be shown in the editor. Since the relevant features (Reader Activation) are behind a flag and not officially released yet, I don't think this is an issue. 

Due to constraints of the block editor – a block can have only one block-editable area ("inner blocks") – there's only one success state. This means that the same message (with note about confirmation email) will be displayed regardless of whether the reader has an account or not. This is accurate, though, because the email will be sent anyway. Also, this way the response does not reveal if the email exists in the database. WDYT @miguelpeixe ? 

I'm not aware of a convention for editing block "states" – here's the switch implemented in this PR:

image

Please share your thoughts @thomasguillot 👆

Closes #1768.

1. On `master`, insert the Reader Registration block on a page
2. Switch to this branch, observe the block working as before on front-end 
3. Open the editor, observe a notice about needing block recovery – click the block recovery button and observe the recovery succeeding 
4. Save the post, observe it's still working as expected on front-end 
5. In the editor, observe a new sidebar panel – "Edited state" – with a "Initial"/"Success" toggle
6. Edit the success state by switching the edited state
7. On the front-end, observe the new success state is displayed after registering 

### Other information:


adekbadek:
I've noticed that there are actually two success states – "check your email" and "thank you for registering", but WP blocks allow only one inner-blocks container.

miguelpeixe:
I think this could be handled universally through settings placed on the Engagement wizard

miguelpeixe:
> Also, this way the response does not reveal if the email exists in the database.

We shouldn't be too worried about that. Anyone attempting to register a new account will know if the email exists in the database.

thomasguillot:
We could use something similar to the [ToggleGroupControl](https://wordpress.github.io/gutenberg/?path=/story/components-experimental-togglegroupcontrol--default) and have the states in the toolbar instead of in the sidebar.

Screenshot 2022-07-19 at 08 47 41

adekbadek:
@thomasguillot – that's how I've initially implemented it, but the quirk of "inner blocks blocks" is that the top-level-block toolbar is not visible when there's content inside. For example with columns block, clicking on the block selects the child block, and its toolbar is visible.

adekbadek:
@miguelpeixe –

> I think this could be handled universally through settings placed on the Engagement wizard (https://github.com/Automattic/newspack-plugin/pull/1773)

Yeah, the default value could be editable there.

thomasguillot:
> that's how I've initially implemented it, but the quirk of "inner blocks blocks" is that the top-level-block toolbar is not visible when there's content inside

Oh I see. A bit annoying.

Ok, so I wonder if we could display some sort of "inner toolbar" at the top when the block (and child blocks) are selected.

Screenshot 2022-07-19 at 11 14 15
Screenshot 2022-07-19 at 11 17 24

adekbadek:
@thomasguillot Good idea, don in f67274d ! Feels a bit weird with the toggle (suggesting an on/off state, not A/B switch), but gets the job done.

adekbadek:
> Could we switch to buttons?

Done in d8e5818dd06df69297bf89842d95f93fa5bdee54

> Also, can we have this "toolbar" visible only on focus?

Someday, maybe, but I don't think it's a blocker for now.

thomasguillot:
Pushed some small tweaks https://github.com/Automattic/newspack-plugin/pull/1785/commits/84c9862a2c7f2cd01b96f2c5993075507b40c69f

assets/blocks/reader-registration/edit.js
```diff
@@ -11,14 +11,31 @@ import { intersection } from 'lodash';
...
+	PanelBody,
+	Button,
+} from '@wordpress/components';
 
 /**
  * Internal dependencies
  */
 import './editor.scss';
 
+const variantOptions = [
```
miguelpeixe, line 35: NIT: `variant` does not feel like a proper name for this. How about `blockState`?

miguelpeixe (changes requested):
The rendered block does not currently reflect the editor:

| Editor | Rendered |
| --- | --- |
| image |  image |
| image | image |

The text color and position of privacy text are off. Also, the initial state of the success message shows as a separate paragraph but renders in one line.

assets/blocks/reader-registration/edit.js
adekbadek: Fixed in 2ae772000da5724b4a4589e2ee1f8132e7be2d2b

adekbadek:
> The text color and position of privacy text are off. Also, the initial state of the success message shows as a separate paragraph but renders in one line.

Fixed in 2ae772000da5724b4a4589e2ee1f8132e7be2d2b & a5714b5d28bca8137e36d1851f50fda225008a31

miguelpeixe (approved):
🎉

PR #1828, opened 2022-08-02
Fix/ga events server side

adekbadek:
### All Submissions:

Refactors how GA server-side event reporting is made. Adds custom events fired when:

- a reader registers via Reader Activation features
    - a custom event (has more contextual data than the NTG standard – login method and )
    - an NTG event, if configured
- a reader signs up for a newsletter - an NTG event, if configured 

Closes https://github.com/Automattic/newspack-issues/issues/235

1. Ensure your site is connected to GA. _On a local instance you can use the [`googlesitekit_allowed_tag_environment_types` hook](https://github.com/google/site-kit-wp/pull/5349) to make Site Kit behave as in production environment_
3. Ensure "News Tagging Guide custom events" are on in Analytics wizard → Custom Events (is on by default)
4. Register as a new reader using any of: Registration block, authorization modal triggered from site header, a newsletter signup, a donation
5. Observe two custom events is sent to GA: one with "Newspack Reader Activation", one with "NTG account" category
6. As a new reader, sign up for a newsletter using the Newsletters plugin' block. Observe a custom event sent with "NTG newsletter" category.

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -262,8 +262,9 @@ function process_form() {
 	if ( ! empty( $lists ) ) {
 		$metadata['lists'] = $lists;
 	}
-	$metadata['current_page_url'] = home_url( add_query_arg( array(), \wp_get_referer() ) );
-	$email                        = \sanitize_email( $_REQUEST['email'] );
+	$metadata['current_page_url']    = home_url( add_query_arg( array(), \wp_get_referer() ) );
+	$metadata['registration_method'] = 'registration-block';
+	$email                           = \sanitize_email( $_REQUEST['email'] );
 
 	$user_id = Reader_Activation::register_reader( $email, '', true, $metadata );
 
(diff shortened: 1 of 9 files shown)


dkoo (changes requested):
The new custom events are working as expected. However, after registration, if you continue navigating throughout the site analytics events are no longer sent. You can confirm this both by monitoring Real Time events in GA, or by monitoring for `collect` requests on the site—these requests are sent prior to registration/login but not afterward.

adekbadek:
This must be because of Site Kit's defaults. It will exclude logged-in users from analytics:

image

Can you confirm?

Added an issue: https://github.com/Automattic/newspack-plugin/issues/1838

dkoo (approved):
> This must be because of Site Kit's defaults. It will exclude logged-in users from analytics:

Ah, this was the cause of the issue. I cab confirm that turning off that setting allows events to get reported for logged-in users.

PR #1831, opened 2022-08-02
fix(google-auth): don't resolve form on popup error

### LINKED ISSUE #1825: Google auth blocked popup
If the Google auth popup is blocked the form is submitted incomplete.

The front end should detect if the Google auth flow was initiated correctly before continuing with any resolution strategy.

### Steps to Reproduce

1. Create a campaign prompt with the registration block
2. With an incognito session, make sure your browser will block popups
3. Click to sign in with Google and confirm the popup is blocked and the page is refreshed

miguelpeixe:
### All Submissions:

If something blocks the Google auth popup, the reader should handle the browser notice and try to click it again. This PR removes the "end flow" action in case the popup is not detected, in order to allow the reader to try again.

Closes #1825 

Check out this branch and confirm that #1825 is not reproducible.

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -299,8 +299,6 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 										clearInterval( interval );
 									}
 								}, 500 );
-							} else if ( googleLoginForm?.endLoginFlow ) {
-								googleLoginForm.endLoginFlow();
 							}
 						}
 					} );


miguelpeixe:
Don't know why reader revenue email tests are failing here.

dkoo (approved):
I'm going to approve this conditionally based on the code alone, with the caveat that I'm unable to replicate the original issue in any browser or any combination of browser settings. I don't think the change will result in any unwanted behavior, at least.

adekbadek (changes requested):
> This PR removes the "end flow" action in case the popup is not detected, in order to allow the reader to try again.

With this change, if `authWindow` is falsy, the auth view will be in a permanent disabled state. Can you share reproduction steps for the issue this resolves? As @dkoo, I was unable to reproduce.

miguelpeixe:
@adekbadek I was not able to reproduce it again either and you're right, the end flow must be called to restore the form availability.

To be on the safe side, I'm repurposing this PR to ensure that we'll never get the popup blocked. Even though we're not getting it blocked on our tests, it's common for browsers to block it when it's opened out of the thread of a user action (a `fetch()` callback).

5fbcb2c46f14b8e691270489f784915776c86d61 changes it to open on the click event thread and 87c4a18781fe5a4f5ba80ec768ed39459b769cb8 restore the end flow.

dkoo (changes requested):
I'm no longer seeing error messages if the operation fails—e.g. on my test site, I have an incomplete connection with Google, which results in a `Cannot verify the request is coming from a Newspack site.` error message in the JS console. When I click the "Sign in with Google" button, the auth window briefly opens, then immediately closes, but the error is visible in the console only, with no front-end indication of any failure. Weren't we displaying this error message previously?

miguelpeixe:
This change is intentional. To ensure the popup is not blocked by the browser (Safari, from my local test) we must trigger it's opening on the user click event thread, which means it's before the `fetch()` that validates the request callback.

dkoo:
Opening it first is fine, because the window does close itself if the request fails, but shouldn't we still show the error message if it does fail?

miguelpeixe:
On my local test the message is shown and persisted after the popup is closed:

image

dkoo (approved):
Strange, it does work for me now, but only on newly loaded pages or in a different browser. Must be some asset caching going on with my local test site's homepage?

miguelpeixe:
Let's wait for @adekbadek review. Thank you for taking another look here! :)

assets/reader-activation/auth.js
```diff
@@ -278,30 +278,36 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
...
+								if ( authWindow.closed ) {
+									checkLoginStatus( metadata );
+									clearInterval( interval );
+								}
+							}, 500 );
+						} else if ( googleLoginForm?.endLoginFlow ) {
+							googleLoginForm.endLoginFlow();
+						}
+					} )
+					.catch( () => {
```
adekbadek, line 308: Suggestion: handle the error in this case, too:

```suggestion
					.catch( error => {
						if ( googleLoginForm?.endLoginFlow ) {
							googleLoginForm.endLoginFlow( error.message );
						}
```

(to test, block the `/wp-json/newspack/v1/login/google` URL in devtools)

adekbadek (approved):
Looks good! Left one suggestion.

PR #1839, opened 2022-08-03
fix(reader-activation): handle modal conflict when auth is triggered …

### LINKED ISSUE #1835: [Reader Activation] On successful login, automatically suppress overlay prompts and registration blocks
**Describe the issue**
If you click the "Already have an account? Sign in" link from the registration overlay prompt, it opens a login modal over the overlay prompt. If you then log in via this modal, it successfully authenticates, but the overlay prompt is still visible with the "Register" CTA, but without the registration form. Any standalone Registration blocks on the page will display the same behavior as well.

Screen Shot 2022-08-02 at 5 22 08 PM

**To Reproduce**
1. In a new session, click the "Already have an account? Sign in" from the overlay prompt that appears on the first pageview.
2. Log in via the Sign In modal that appears on top of the overlay.
3. Observe the Registration blocks on the page still visible with "Register for free" CTAs, but without the forms to do so.

**Expected behavior**
Once the Sign In modal has been dismissed after a successful login, all Register blocks on the page should be hidden. 

Additionally, any overlay prompts should be suppressed as if closed, otherwise doing the above will result in an empty overlay prompt after logging in via the modal.

This will require changes in the Campaigns plugin as well.

adekbadek:
### All Submissions:

Closes #1835.

1. With Reader Activation enabled, set up an overlay prompt with a Registration block inside
2. Insert the Registration block on a page
3. Visit the page, click on "Sign In" in the "Already have an account? Sign in" message
4. Observe the overlay prompt is not displayed, and an auth modal is displayed
5. Close the auth modal, observe the prompt is visible again 
6. Open the auth modal again, and complete the authentication
7. Observe the prompt is not visible 
8. Observe the Registration block is hidden

The Registration block will display the success state for registered readers on subsequent pageviews. If we'd prefer it to be hidden, that's for another issue. WDYT @Automattic/newspack-product ?

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -70,6 +70,16 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 			form = initialForm;
 		}
 
+		let currentlyOpenOverlayPrompts = [];
+		const hideCurrentlyOpenOverlayPrompts = () =>
+			currentlyOpenOverlayPrompts.forEach( promptElement =>
+				promptElement.setAttribute( 'amp-access-hide', '' )
+			);
+		const displayCurrentlyOpenOverlayPrompts = () =>
+			currentlyOpenOverlayPrompts.forEach( promptElement =>
+				promptElement.removeAttribute( 'amp-access-hide' )
+			);
+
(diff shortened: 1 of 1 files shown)


adekbadek:
Agreed!

PR #1846, opened 2022-08-03
fix(reader-activation): remove async prop from library

### LINKED ISSUE #1845: Reader Activation: Google sign-in button race condition
**Describe the bug**

#1781 has a race condition that results in the "Sign in with Google" button sometimes being inoperable due to the required global not existing by the time the button is rendered. This is still replicable on the current `master`.

**To Reproduce**

1. On a site with Reader Activation enabled and a valid Google connection, set up a Register block in a prompt or in post content.
2. Refresh the page several times and observe that occasionally, but button remains disabled and unclickable.

miguelpeixe:
### All Submissions:

Remove the `async` prop from the reader activation front-end library to ensure it's available as a script dependency.

Closes #1845 

Confirm #1845 is not reproducible.

### Other information:

### DIFF
--- a/includes/class-reader-activation.php
+++ b/includes/class-reader-activation.php
@@ -87,7 +87,6 @@ final class Reader_Activation {
 				'authenticated_email'   => $authenticated_email,
 			]
 		);
-		\wp_script_add_data( self::SCRIPT_HANDLE, 'async', true );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This seems to fix the issue. 👍 Noting that this makes the Reader Activation script (~2kb minified) render-blocking, which may have minor performance implications.

PR #1847, opened 2022-08-03
feat(reader-activation): optimistic account link

miguelpeixe:
### All Submissions:

Implements a more optimistic experience for the header account link. If the reader is pre-authenticated it shows "Account" instead of "Sign In", just like an authenticated reader. Once you click on Account, you're prompted to sign in with a pre-filled email.

A successful authentication in this case redirects the reader to the account page, as it's expected by clicking on a link labeled "Account".

There's also a small tweak to not `endLoginFlow()` on `finally()` and use `catch()` instead, to not duplicate the action that is already fired at `.then()`.

1. With an existing reader account, make sure you have a password set
2. While logged out, use a registration block to pre-authenticate to this account 
3. Observe the header nav changed to "Account"
4. Click and confirm you see the "Sign In" modal with pre-filled email input
5. Sign in with the password and confirm you are redirected to the account page

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -73,6 +73,7 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 		const actionInput = form.querySelector( 'input[name="action"]' );
 		const emailInput = form.querySelector( 'input[name="email"]' );
 		const passwordInput = form.querySelector( 'input[name="password"]' );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works well, but also begs the question: shouldn't we treat the flow differently when a reader "registers" with an email address that's already been registered with an account? Perhaps TBD in a different PR.

PR #1849, opened 2022-08-04
feat: if registering an email that already has an account, show different message

dkoo:
### All Submissions:

Follow-up to https://github.com/Automattic/newspack-plugin/pull/1847#pullrequestreview-1061084857.

If you enter an email address that's already been registered into the Register block form, the `Reader_Activation` class recognizes the existing account and sends a magic link to the email address. However, on the front-end the messaging shows no indication that this has happened. This change displays a different message to the reader in this scenario, which can also be edited in a new, third edited state in the block editor:

Screen Shot 2022-08-03 at 6 40 54 PM

1. Check out this branch.
2. In the Register block, enter an email address that's already been registered and submit the form.
3. Confirm that the default message is displayed (as above in the editor screenshot).
4. Edit the success message in the block editor and repeat step 3. Confirm that your custom message is displayed:

Screen Shot 2022-08-03 at 6 44 10 PM

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -303,7 +303,7 @@ function process_form() {
 		[
 			'email'         => $email,
 			'authenticated' => $user_logged_in,
-			'existing_user' => false === $user_id,
+			'existing_user' => ! $user_logged_in,
 		]
 	);
 }
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -90,6 +90,7 @@ function render_block( $attrs, $content ) {
 		'newsletterLabel'  => __( 'Subscribe to our newsletter', 'newspack' ),
 		'haveAccountLabel' => __( 'Already have an account?', 'newspack' ),
 		'signInLabel'      => __( 'Sign in', 'newspack' ),
+		'signedInLabel'    => __( 'An account was already registered with this email. Please check your inbox for an authentication link.', 'newspack' ),
 	];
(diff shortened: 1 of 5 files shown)


assets/blocks/reader-registration/index.php
```diff
@@ -297,6 +303,7 @@ function process_form() {
 		[
 			'email'         => $email,
 			'authenticated' => $user_logged_in,
+			'existing_user' => false === $user_id,
```
adekbadek, line 306: Isn't it same as `! $user_logged_in`?

assets/blocks/reader-registration/index.php
```diff
@@ -90,6 +90,7 @@ function render_block( $attrs, $content ) {
 		'newsletterLabel'  => __( 'Subscribe to our newsletter', 'newspack' ),
 		'haveAccountLabel' => __( 'Already have an account?', 'newspack' ),
 		'signInLabel'      => __( 'Sign in', 'newspack' ),
+		'signedInLabel'    => __( 'An account was already registered with this email. Please check your inbox for an authentication link.', 'newspack' ),
```
adekbadek, line 93: Not an addition of this PR, but default attributes are already set in `block.json`. Why should they be repeated here, @miguelpeixe ?

adekbadek (approved):
Works as advertised, left some non-blocking comments.

assets/blocks/reader-registration/index.php
miguelpeixe: These are placed as default values in case the attribute is set as an empty string, which does not fall back to the configured default value.

PR #1854, opened 2022-08-04
feat: replace WooCommerce’s login form with our own

### LINKED ISSUE #1850: "My Account" page for anons – show Newspack auth instead of native login form
## As is 

With Reader Activation on, visiting `/my-account` page as an anonymous reader will result in a native WC login form. 

## To be

The login form should be replaced with Reader Activation auth view, the one displayed in a modal after clicking the header auth link.

dkoo:
### All Submissions:

Replaces the native WooCommerce login form with our own, utilizing the same markup as we use for the login modal (but with some additional styles to make it render inline).

To do so, ~we rely a bit on a hack to hijack the rendering of the WooCommerce login form template; when this template is about to be rendered, we replace with an empty template file and instead call our `render_auth_form` method to output the form markup. There may be a more elegant way to do this, but I couldn't find one in the short term (cc @claudiulodro if you have any suggestions).~ @miguelpeixe suggested a bit of a less-hacky way to do this by simply calling the `render_auth_form` method in the template itself.

Closes #1850.

1. Check out this branch.
2. In a new session, while not logged in, visit the My Account page to log in.
3. Confirm that you see our Reader Activation login form here instead of the usual native WC login form.
4. Confirm that all functionality works here as expected, and that once you log in successfully you're redirected to the My Account dashboard in the logged-in state.

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -85,13 +85,16 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 		const passwordInput = form.querySelector( 'input[name="password"]' );
 		const redirectInput = form.querySelector( 'input[name="redirect"]' );
 		const submitButtons = form.querySelectorAll( '[type="submit"]' );
+		const closeButton = container.querySelector( 'button[data-close]' );
 
-		container.querySelector( 'button[data-close]' ).addEventListener( 'click', function ( ev ) {
-			ev.preventDefault();
-			container.classList.remove( 'newspack-reader__auth-form__visible' );
-			container.style.display = 'none';
-			displayCurrentlyOpenOverlayPrompts();
-		} );
+		if ( closeButton ) {
+			closeButton.addEventListener( 'click', function ( ev ) {
+				ev.preventDefault();
+				container.classList.remove( 'newspack-reader__auth-form__visible' );
+				container.style.display = 'none';
+				displayCurrentlyOpenOverlayPrompts();
+			} );
+		}
 
 		const messageContentElement = container.querySelector(
 			'.newspack-reader__auth-form__response__content'
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -96,7 +96,10 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 				const redirectInput = container.querySelector( 'input[name="redirect"]' );
 				const reader = readerActivation.getReader();
 				let redirectAfterLogin = redirectInput && !! redirectInput.value;
-				emailInput.value = reader?.email || '';
+				if ( emailInput ) {
+					emailInput.value = reader?.email || '';
+				}
+
 				if ( accountLinks?.length ) {
 					accountLinks.forEach( link => {
 						/** If there's a pre-auth, signing in redirects to the reader account. */
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -95,7 +95,7 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 				const emailInput = container.querySelector( 'input[name="email"]' );
 				const redirectInput = container.querySelector( 'input[name="redirect"]' );
 				const reader = readerActivation.getReader();
-				let redirectAfterLogin = !! redirectInput.value;
+				let redirectAfterLogin = redirectInput && !! redirectInput.value;
 				emailInput.value = reader?.email || '';
 				if ( accountLinks?.length ) {
 					accountLinks.forEach( link => {
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -59,8 +59,122 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 			return;
 		}
 
+		let currentlyOpenOverlayPrompts = [];
+		const hideCurrentlyOpenOverlayPrompts = () =>
+			currentlyOpenOverlayPrompts.forEach( promptElement =>
+				promptElement.setAttribute( 'amp-access-hide', '' )
+			);
+		const displayCurrentlyOpenOverlayPrompts = () =>
+			currentlyOpenOverlayPrompts.forEach( promptElement =>
+				promptElement.removeAttribute( 'amp-access-hide' )
+			);
+
+		let accountLinks, triggerLinks;
+		const initLinks = function () {
+			accountLinks = document.querySelectorAll( '.newspack-reader__account-link' );
+			triggerLinks = document.querySelectorAll( '[data-newspack-reader-account-link]' );
+			triggerLinks.forEach( link => {
+				link.addEventListener( 'click', handleAccountLinkClick );
+			} );
+		};
+		initLinks();
+		/** Re-initialize links in case the navigation DOM was modified by a third-party. */
+		setTimeout( initLinks, 1000 );
+
+		/**
+		 * Handle reader changes.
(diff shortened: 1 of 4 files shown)


includes/templates/reader-activation/login-form.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Reader Activation login form.
+ * This is an empty file in order to trick WooCommerce's My Account login template
+ * into rendering nothing so that we can replace it with our own login form.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
```
miguelpeixe, line 10:
```suggestion
namespace Newspack;

Reader_Activation::render_auth_form( true );
```

Instead of rendering nothing, the method could be called here

includes/class-reader-activation.php
```diff
@@ -564,30 +580,33 @@ public static function render_auth_form() {
 		}
 		$terms_text = self::get_setting( 'terms_text' );
 		$terms_url  = self::get_setting( 'terms_url' );
+		$redirect   = $is_inline ? home_url( '/my-account/edit-account/' ) : '';
```
miguelpeixe, line 583:
```suggestion
		$redirect   = $is_inline ? \wc_get_account_endpoint_url( 'dashboard' ) : '';
```

includes/class-reader-activation.php
```diff
@@ -386,6 +386,11 @@ private static function get_element_class_name( ...$parts ) {
 	 * Setup nav menu hooks.
 	 */
 	public static function setup_nav_menu() {
+		$account_page = \wc_get_page_id( 'myaccount' );
+		if ( get_the_ID() === $account_page ) {
+			return;
+		}
```
miguelpeixe, line 392: I feel like this creates a UI inconsistency, almost like it's glitchy. Why would we hide the account link while on the page?

miguelpeixe (changes requested):
Some CSS could hide the page title like we already do for the redesigned account page:

image

includes/templates/reader-activation/login-form.php
dkoo: Great idea, that's much less hacky feeling! 4d6f9d2

includes/class-reader-activation.php
dkoo: I thought there would be little point in being able to open the login modal on the standalone login page. Additionally, doing so would require refactoring the JS a bit more to be able to handle multiple instances of the login form on the same page—not a big deal, but is it worth it for this feature?

dkoo:
> Some CSS could hide the page title like we already do for the redesigned account page:

Done in 6342ad1.

includes/class-reader-activation.php
```diff
@@ -555,6 +567,10 @@ public static function render_auth_form() {
 		}
 		// phpcs:enable
 
+		if ( $is_inline ) {
+			$classnames[] = 'newspack-reader__auth-form__inline';
```
miguelpeixe, line 571:
```suggestion
			$classnames[] = $class( 'inline' );
```

miguelpeixe (changes requested):
I missed another tweak and also looks like there's a disproportionate top margin from the Sign In form, WDYT?

image

includes/class-reader-activation.php
miguelpeixe: Fair point!

dkoo: You know, I rethought this and I think you were right. 9db42f5 restores the modal on My Account pages and does the JS refactoring handle multiple instances of the form on the same page.

includes/class-reader-activation.php
dkoo: Ooh, I misread what that function did. Much cleaner.

dkoo:
> I missed another tweak and also looks like there's a disproportionate top margin from the Sign In form, WDYT?

Agreed, 23f56e4 removes the padding from the content wrapper for inline forms.

assets/reader-activation/auth.js
```diff
@@ -54,299 +54,307 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
...
-		const submitButtons = form.querySelectorAll( '[type="submit"]' );
+			let currentlyOpenOverlayPrompts = [];
+			const hideCurrentlyOpenOverlayPrompts = () =>
+				currentlyOpenOverlayPrompts.forEach( promptElement =>
+					promptElement.setAttribute( 'amp-access-hide', '' )
+				);
+			const displayCurrentlyOpenOverlayPrompts = () =>
+				currentlyOpenOverlayPrompts.forEach( promptElement =>
+					promptElement.removeAttribute( 'amp-access-hide' )
+				);
```
miguelpeixe, line 83: This logic can live outside of the loop

assets/reader-activation/auth.js
```diff
@@ -54,299 +54,307 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
...
-						const labels = JSON.parse( link.getAttribute( 'data-labels' ) );
-						link.querySelector( '.newspack-reader__account-link__label' ).textContent =
-							reader?.email ? labels.signedin : labels.signedout;
-					} catch {}
+			let accountLinks, triggerLinks;
+			const initLinks = function () {
+				accountLinks = document.querySelectorAll( '.newspack-reader__account-link' );
+				triggerLinks = document.querySelectorAll( '[data-newspack-reader-account-link]' );
+				triggerLinks.forEach( link => {
+					link.addEventListener( 'click', handleAccountLinkClick );
```
miguelpeixe, line 113: This too, and the `handleAccountLinkClick()` and `handleReaderChanges()` functions.

assets/reader-activation/auth.js
```diff
@@ -54,299 +54,307 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
...
-				);
-				fetch( '/wp-json/newspack/v1/login/google' )
-					.then( res => res.json().then( data => Promise.resolve( { data, status: res.status } ) ) )
-					.then( ( { data, status } ) => {
-						if ( status !== 200 ) {
-							if ( authWindow ) {
-								authWindow.close();
+			/**
+			 * Third party auth.
+			 */
```
miguelpeixe, line 290: The third-party handling too does not rely on a specific container.

miguelpeixe (changes requested):

assets/reader-activation/auth.js
dkoo: `initLinks` can live outside the loop, but both `handleAccountLinkClick` and `handleReaderChanges` functions rely on form fields inside the specific containers, no?

miguelpeixe: Oops! You're right, they mix account link and form actions.

dkoo: Ah, but they do need to get moved outside the loop (and then create their own internal loops to handle multiple forms) because otherwise the account links will only trigger the last instance of the form on the page. This will probably be the modal form, but b08d5bc refactors to ensure that all form instances are handled.

assets/reader-activation/auth.js
dkoo: Moved outside the loop in b08d5bc.

assets/reader-activation/auth.js
dkoo: Done in b08d5bc

miguelpeixe (approved):
Looks good now! 🙌 

Thank you for the revisions!

dkoo:
Thanks for your patience with the back and forth!

PR #1869, opened 2022-08-08
fix(oauth): csrf token lifespan

miguelpeixe:
### All Submissions:

Although we want this token to be short-lived, 60 seconds may not be enough if the user is not yet authenticated to the service and needs to go through 2FA.

This PR increases the expiration of the transient to 5 minutes.

1. With reader activation on, click sign in using Google
2. Before resolving the Google popup, wait 3 minutes
3. Resolve the prompt and confirm you are authenticated

### Other information:


dkoo (approved):
Good call, and 5 minutes should be a reasonable time period to allow for 2FA flows.

PR #1870, opened 2022-08-08
Update reader registration design

thomasguillot:
### All Submissions:

This PR tweaks the design a little bit. Mainly the typography/vertical rhythm to follow an 8px baseline. There's probably more to tweak but this is a first pass 🙂

I also moved the text for the magic link sign up, hopefully it'll be sufficient to fix https://app.asana.com/0/1202719992789620/1202723823068199/f

1. Test the front-end
2. Switch to this branch
3. Re-test

### Other information:

### DIFF
--- a/assets/reader-activation/auth.scss
+++ b/assets/reader-activation/auth.scss
@@ -1,3 +1,4 @@
+@use 'sass:color';
 @use '~@wordpress/base-styles/colors' as wp-colors;
 
 .newspack-reader {
@@ -58,7 +59,11 @@
 		justify-content: center;
 		align-items: center;
 		z-index: 999999;
-		font-size: 0.9em;
+		font-size: 0.8888em;
+
+		@media screen and ( min-width: 600px ) {
+			font-size: 0.8em;
+		}
 
 		&__visible {
 			display: flex !important;
@@ -75,21 +80,18 @@
 		&[data-form-status='200'] {
 			.newspack-reader__auth-form {
 				&__response {
-					background: rgba( 74, 184, 102, 0.1 );
 					text-align: center;
-					padding: 2.5rem;
-					margin: 1rem 0;
 					font-size: 1em;
 					&__icon[data-form-status='200'] {
 						display: flex;
-						margin: 0 auto 2rem;
+						margin: 0 auto 1em;
 						border-radius: 100%;
-						width: 50px;
-						height: 50px;
+						width: 40px;
+						height: 40px;
 						justify-content: center;
 						align-items: center;
-						color: #fff;
-						background: #4ab866;
+						color: white;
+						background: wp-colors.$alert-green;
 					}
 				}
 			}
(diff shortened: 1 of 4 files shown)


includes/class-reader-activation.php
```diff
@@ -597,7 +615,10 @@ public static function render_auth_form( $is_inline = false ) {
 							<?php _e( "We've recently sent you an authentication link. Please, check your inbox!", 'newspack' ); ?>
 						</p>
 						<p data-action="pwd">
-							<?php _e( 'Sign in below to verify your identity.', 'newspack' ); ?>
+							<?php \esc_html_e( 'Sign in below to verify your identity.', 'newspack' ); ?>
+						</p>
+						<p data-action="link">
+							<?php \esc_html_e( 'Get a link sent to your email address to sign in instantly without your password.', 'newspack' ); ?>
```
dkoo, line 621: "your" in this context implies that the user has a password, but this isn't always the case.

```suggestion
							<?php \esc_html_e( 'Get a link sent to your email address to sign in instantly without a password.', 'newspack' ); ?>
```

assets/reader-activation/auth.scss
```diff
@@ -140,51 +173,70 @@
...
+
+				> *:last-child {
+					margin-bottom: 0 !important;
+				}
 			}
 			input[type='email'],
 			input[type='password'] {
 				width: 100%;
 			}
-			[type='submit'] {
```
dkoo, line 184: Was this intentionally removed? This makes the submit button take on the theme's secondary color, which by default is quite a bit less prominent than the primary.

With primary color (`master`):
Screen Shot 2022-08-08 at 9 22 49 PM

With secondary color (this PR):
Screen Shot 2022-08-08 at 9 23 10 PM

assets/reader-activation/auth.scss
```diff
@@ -219,41 +269,48 @@
...
+				outline: 1.5px solid #111;
+				outline-offset: 2px;
 			}
 		}
 
 		&__terms-text {
-			font-size: 0.8em;
-			margin: 0.5rem 0;
-			color: #757575;
+			font-size: 0.625em;
```
dkoo, line 311: This text is now minuscule! I feel like it's a bit _too_ small at 10px.

dkoo (changes requested):
Definitely love the new placement of the magic link explanation text—it feels more consistent with the placement of the explanation in the password form and doesn't get lost as easily.

Other than that, I have some questions/suggestions about the other changes.

assets/reader-activation/auth.scss
thomasguillot: Yes, it's intentional. I wanted to match the theme's styles.
Also for now, we said we would use custom CSS to apply a red colour #DD3333 to these buttons.
That being said, I'd like to introduce a setting somewhere (Site Design? Reader Activation?) to control the background/text colour for them.

assets/reader-activation/auth.scss
thomasguillot: My bad, I wanted 11px. Fixed here: https://github.com/Automattic/newspack-plugin/pull/1870/commits/ec3611c059f10a0fb9bafcf2dc428b1cfca086dd

dkoo: It's still pretty small—if it were up to me I'd make it 13px to match the size of the other "supplemental" text like "Lost your password?" But that's totally subjective, so I won't block the PR for this.

dkoo (approved):

PR #1871, opened 2022-08-08
fix(google-auth): catch and display errors

miguelpeixe:
### All Submissions:

Make sure to catch and display OAuth potential errors for different steps of the flow.

1. Signing in with Google should continue to work as expected
2. Change line 330 of `assets/reader-activation/auth.js` to a 404 URL and recompile
3. Try to sign in again and confirm JS doesn't break and you see an error message

### Other information:


dkoo (approved):
Sometimes the errors themselves are not so helpful (e.g. `JSON.parse: unexpected end of data at line 1 column 1 of the JSON data`) but at least they're caught and displayed, resolving the UI.

miguelpeixe:
The JSON parsing error would occur only if the backend fails to deliver a JSON response, which I think should be considered a backend error, not a client.

PR #1857, opened 2022-08-05
fix(ga): cookie parsing

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Tiny bug w/ cookie parsing. The `_ga` cookie might have different values than previously expected.

### How to test the changes in this Pull Request:

1. Register on a site with Reader Activation & Site Kit w/ GA configured
2. manually set `_ga` cookie to value of `GA1.2.amp-12345`
3. On `master`, observe a `PHP Notice:  Undefined offset: 3` notice logged
4. Switch to this branch, observe no notice logged

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1874, opened 2022-08-09
fix: parse CID from _ga cookie if it only contains CID string

dkoo:
### All Submissions:

#1857 fixed cookie parsing if it only contains the version number, domain depth and CID string parts, but I noticed in some sessions the cookie string contains _only_ the CID, which still causes the parsing to fail. This PR handles that situation.

I'm not sure if the cookie value can contain other combinations of strings, so @adekbadek may have some insight into this.

1. Register on a site with Reader Activation & Site Kit w/ GA configured
2. manually set `_ga` cookie to value of `amp-12345`
3. On `master`, observe a `PHP Notice:  Undefined offset: 1` notice logged
4. Switch to this branch, observe no notice logged

### Other information:


miguelpeixe (approved):
Working as described!

I noticed the cookie only had the `CID` while accessing AMP pages anonymously.

PR #1876, opened 2022-08-09
feat: disable woocomerce welcome emails in favor of verification email

### LINKED ISSUE #1872: Reader Activation: replace WooCommerce's registration confirmation email with our verification email
**Current**

When registering an email address for the first time, the email address will receive a standard WooCommerce "Your account has been created!" confirmation email. In order to access sensitive My Account settings, they need to verify their email via Reader Activation's verification flow.

**Desired**

When registering an email address for the first time, the email address should receive the Reader Activation verification email instead of the WooCommerce confirmation email. This removes some friction to verifying the email address and also reduces confusion that might stem from the differently-branded WooCommerce emails.

dkoo:
### All Submissions:

Disables WooCommerce's standard, WC-branded welcome emails that are normally sent to new customers, so that we can send our own verification email instead.

Will not send a verification email to readers who register via SSO, which is considered a secure login method and doesn't require verification.

Closes #1872.

1. Check out this branch.
2. Register a new email address via the Register block (via direct email input, not SSO) and confirm that you receive a magic link email with a special subject line and message instead of WooCommerce's "Your Newspack account has been created!" email.
3. Visit the magic link URL in the same session where you registered the email address and confirm that you're redirected to the main My Account dashboard page with a message that your email has been verified. Also confirm that you're able to manage newsletter subscriptions at this point.
4. In a new session, register a new account via SSO and confirm that you do not receive a magic link email, and that you're immediately able to manage newsletter subscriptions in My Account > Newsletters.

### Other information:

### DIFF
--- a/includes/class-magic-link.php
+++ b/includes/class-magic-link.php
@@ -509,16 +509,6 @@ final class Magic_Link {
 			return false;
 		}
 
-		if ( ! Reader_Activation::is_reader_verified( $user ) ) {
-			Reader_Activation::set_reader_verified( $user );
-
-			if ( function_exists( '\wc_add_notice' ) ) {
-				\wc_add_notice( __( 'Thank you for verifying your account!', 'newspack' ), 'success' );
-			}
-		}
-
-		Reader_Activation::set_current_reader( $user->ID );
-
 		/**
 		 * Fires after a reader has been authenticated via magic link.
 		 *
--- a/includes/class-magic-link.php
+++ b/includes/class-magic-link.php
@@ -318,6 +318,7 @@ final class Magic_Link {
 	 *   Used to build wp_mail().
 	 *
 	 *   @type string $to      The intended recipient - New user email address.
+	 *   @type string $subject The subject of the email.
 	 *   @type string $message The body of the email.
 	 *   @type string $headers The headers of the email.
 	 * }
--- a/includes/class-magic-link.php
+++ b/includes/class-magic-link.php
@@ -311,17 +311,18 @@ final class Magic_Link {
 	 *
 	 * @param \WP_User $user        User to generate the magic link for.
 	 * @param string   $redirect_to Which page to redirect the reader after authenticating.
+	 * @param string   $subject The subject of the email.
+	 * @param string   $message     Message to show in body of email.
 	 *
 	 * @return array|\WP_Error $email Email arguments or error. {
 	 *   Used to build wp_mail().
 	 *
 	 *   @type string $to      The intended recipient - New user email address.
-	 *   @type string $subject The subject of the email.
 	 *   @type string $message The body of the email.
 	 *   @type string $headers The headers of the email.
 	 * }
 	 */
-	public static function generate_email( $user, $redirect_to = '' ) {
+	public static function generate_email( $user, $redirect_to = '', $subject, $message = '' ) {
 		if ( ! self::can_magic_link( $user->ID ) ) {
 			return new \WP_Error( 'newspack_magic_link_invalid_user', __( 'Invalid user.', 'newspack' ) );
 		}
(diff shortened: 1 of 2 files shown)


dkoo:
@miguelpeixe 23b9809 eschews the use of Newsletters' verification system in favor of magic links with customizable subjects and messages.

includes/class-magic-link.php
```diff
@@ -311,17 +311,18 @@ private static function generate_url( $user, $url = '' ) {
 	 *
 	 * @param \WP_User $user        User to generate the magic link for.
 	 * @param string   $redirect_to Which page to redirect the reader after authenticating.
+	 * @param string   $subject The subject of the email.
```
miguelpeixe, line 314:
```suggestion
	 * @param string   $subject     The subject of the email.
```

NIT: lint

includes/class-magic-link.php
```diff
@@ -311,17 +311,18 @@ private static function generate_url( $user, $url = '' ) {
...
+	 * @param string   $message     Message to show in body of email.
 	 *
 	 * @return array|\WP_Error $email Email arguments or error. {
 	 *   Used to build wp_mail().
 	 *
 	 *   @type string $to      The intended recipient - New user email address.
-	 *   @type string $subject The subject of the email.
```
miguelpeixe, line 319: This may have been by accident?

miguelpeixe (changes requested):
Thank you for revising the strategy for this! It's working great!

I have another take on the concept here. I think we can preserve the behavior of the magic link by exposing the `Magic_Link::generate_url()` and not introducing registration logic to it.

With `Magic_Link::generate_url()`, `Reader_Activation` can send the email and handle the redirect URL and success notice on its own.

includes/class-magic-link.php
dkoo: Yes, I put the `@param` line in the wrong spot at first.

dkoo:
> I have another take on the concept here. I think we can preserve the behavior of the magic link by exposing the `Magic_Link::generate_url()` and not introducing registration logic to it.

That's a good point, but I'd rather not replicate the logic of sending the actual magic link email in `Reader_Activation`. How about this? 49d3067 moves all logic related to email verification to `Reader_Activation`, but still uses methods from `Magic_Link` to build and send the email, allowing for arbitrarily overriding the subject line and message for the magic link email. That keeps `Magic_Link` logic abstract and clean from any reader registration/verification logic.

includes/class-magic-link.php
```diff
@@ -499,9 +509,6 @@ private static function authenticate( $user_id, $token ) {
 			return false;
 		}
 
-		Reader_Activation::set_reader_verified( $user );
-		Reader_Activation::set_current_reader( $user->ID );
-
```
miguelpeixe, line 504: It's a non-blocker, but I find it confusing that `authenticate()` doesn't authenticate.

includes/class-reader-activation.php
```diff
@@ -959,21 +966,26 @@ public static function is_reader_verified( $user ) {
...
-		$user = \get_user_by( 'id', $user_id );
 		if ( ! $user || \is_wp_error( $user ) || ! self::is_user_reader( $user ) ) {
 			return new \WP_Error( 'newspack_authenticate_invalid_user', __( 'Invalid user.', 'newspack' ) );
 		}
 
+		$user_id = \absint( $user->ID );
+		if ( empty( $user_id ) ) {
+			return new \WP_Error( 'newspack_authenticate_invalid_user_id', __( 'Invalid user id.', 'newspack' ) );
+		}
+
```
miguelpeixe, line 988: This check is superfluous, the user has already been validated on the previously.

miguelpeixe (changes requested):
> not replicate the logic of sending the actual magic link email in Reader_Activation

My proposal is based on the idea that it is not a magic link email, it's a verification email that borrows the magic link as the verifier, hence using just the `generate_url()` from `Magic_Link`.

It makes for a more solid understanding that account verification is a core feature and if we, in the future, decide that the URL should not be a magic link it will be more straightforward to detach and maintain.

This is a non-blocker, though! It's working as described 🎉 

Left just a couple of comments below.

includes/class-magic-link.php
miguelpeixe: This line should be restored

includes/class-reader-activation.php
dkoo: True, removed in 9d7f3ab.

includes/class-magic-link.php
dkoo: Hmm, that's a good point. 9d7f3ab moves the `Reader_Activation::set_current_reader` call back to the `Magic_Link::authenticate` method, but leaves the `set_reader_verified` call as a callback on the `newspack_magic_link_authenticated` hook. In my mind the verification is a side effect of authenticating via magic link, as you suggest.

dkoo:
> My proposal is based on the idea that it is not a magic link email, it's a verification email that borrows the magic link as the verifier, hence using just the `generate_url()` from `Magic_Link`.

> It makes for a more solid understanding that account verification is a core feature and if we, in the future, decide that the URL should not be a magic link it will be more straightforward to detach and maintain.

Ah, I think I get what you're saying. I think we should leave it as-is for now and refactor if needed with an entirely new class and set of features if we decide to decouple verification from magic links. Right now it's not so tightly integrated that we can't easily swap the usage of the single `Magic_Link::send_email` method for another.

includes/class-magic-link.php
dkoo: Whoops, I misunderstood what you were saying—restored in ae476dd.

miguelpeixe (approved):
🙌

PR #1881, opened 2022-08-11
refactor: "My Account" lock for unverified readers

miguelpeixe:
### All Submissions:

This PR refactors the approach for locking "My Account" access to unverified readers. The approach was changed to hook into the `woocommerce_account_content` instead of the combination of `wc_get_template` hook and URL redirection.

Fixes:
 - Log out URL should now work for unverified readers accessing the page
 - Strict reader check now allows for a non-RAS customer/subscriber to be interpreted as verified

The new test for the strict reader should pass.

### Non-RAS Subscriber

1. Create a regular subscriber and set a password
2. On a new session, authenticate with the subscriber
3. Confirm you can navigate the account page
4. Visit the Newsletters section and confirm you still have to verify for newsletters management

### RAS Reader

1. Create a new account through RAS (registration block)
2. Access My Account and confirm you have to verify your account
3. Try to access another endpoint ( `/my-account/newsletters` or `/my-account/orders/`)
4. Confirm you're still prompted to verify
5. Log out
6. Confirm you are successfully logged out

### Other information:

### DIFF
--- a/includes/class-reader-activation.php
+++ b/includes/class-reader-activation.php
@@ -275,15 +275,16 @@ final class Reader_Activation {
 	/**
 	 * Whether the user is a reader.
 	 *
-	 * @param \WP_User $user User object.
+	 * @param \WP_User $user   User object.
+	 * @param bool     $strict Whether to check if the user was created through reader registration. Default false.
 	 *
 	 * @return bool Whether the user is a reader.
 	 */
-	public static function is_user_reader( $user ) {
+	public static function is_user_reader( $user, $strict = false ) {
 		$is_reader = (bool) \get_user_meta( $user->ID, self::READER, true );
 		$user_data = \get_userdata( $user->ID );
 
-		if ( false === $is_reader ) {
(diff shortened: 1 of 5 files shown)


dkoo (approved):
This works better than the original method—thanks for the refactor.

I have a minor nitpick related to the use of the content filter instead of a redirect. Non-blocking because it's an edge case and unlikely to be seen by most regular users, but when you visit a non-root My Account page via direct URL input while not verified, it shows the correct page content asking you to verify but:

1. The URL is not updated
2. No menu item is highlighted

This also highlights a minor issue where My Account pages that are not normally accessible to RAS accounts via the My Account menu (orders, subscriptions) are still accessible via direct URL input. Perhaps a fix for another day.

miguelpeixe:
I think the issue you mention is unrelated to this PR. The redirection already in place before the "My Account" lock was just from the "My Account" dashboard to `edit-account`, so you could still access orders and other hidden account pages.

I personally think it's a non-issue. We're not hiding those pages for restrictive access or security, it's a UX choice.

PR #1884, opened 2022-08-12
feat: reader account deletion and ESP sync options

### LINKED ISSUE #1892: Synchronise account deletion
## As is 

If a WP user account is removed, the Newsletter ESP won't be notified. 

## To be 

If the WP user account is removed, this should be synced to ESP. This might have to be an optional feature, since this behaviour might not be desired.

### LINKED ISSUE #1893: Reader Activation – make ESP synchronisation optional
## As is 

The ESP-synchronisation features (adding contact, updating metadata) are implicit when Reader Activation is enabled. 

## To be 

The ESP-synchronisation should be optional, for publishers who do not wish to use the ESP as a CRM.

miguelpeixe:
### All Submissions:

Allow the reader to delete their own account through email confirmation.

image

image

image

This PR also implements ESP synchronization options:

 - Reader metadata synchronization
 - Delete from ESP on reader deletion

image

Question for @Automattic/newspack-product: the ESP synced deletion only occurs when readers self-delete their account (which triggers `Reader_Activation::delete_reader()`). The regular `wp_delete_user()`, which is used by admins on the WP dashboard, does not. Should we reconsider this?

1. Check out this branch and https://github.com/Automattic/newspack-newsletters/pull/929
2. Make sure you have RAS activated and Newsletters setup with ActiveCampaign
3. Visit the Engagement wizard and confirm the default options are set (sync and delete enabled)
4. Using a Registration Block, register a new account subscribed to the lists
5. Confirm the account is created on the ESP with RAS metadata
6. Access the "My Account" page, confirm the email, and scroll down to request account deletion
7. Click on the received email and confirm you access a confirmation page
8. Click to delete and confirm you're redirected to the sign-in form
9. Confirm on the ESP dashboard that the contact has been deleted as well
10. Disable synchronization and repeat steps 4-9, this time confirm the contact is subscribed to the list but:
    1. without RAS metadata
    2. preserved after reader deletion
12. Re-enable sync, change your ESP to Mailchimp, update your block to use Mailchimp's lists, and repeat steps 4-9

Closes #1892
Closes #1893

### Other information:

### DIFF
--- a/assets/wizards/engagement/views/reader-activation/index.js
+++ b/assets/wizards/engagement/views/reader-activation/index.js
@@ -131,7 +131,7 @@ export default withWizardScreen( () => {
 						<CheckboxControl
 							label={ __( 'Delete contact on reader deletion', 'newspack' ) }
 							help={ __(
-								"If the reader's email is verified, permanently delete contact from ESP on reader deletion. ESP synchronization must be enabled.",
+								"If the reader's email is verified, delete contact from ESP on reader deletion. ESP synchronization must be enabled.",
 								'newspack'
 							) }
 							checked={ !! config.sync_esp_delete }
--- a/assets/wizards/engagement/views/reader-activation/index.js
+++ b/assets/wizards/engagement/views/reader-activation/index.js
@@ -89,15 +89,6 @@ export default withWizardScreen( () => {
 					checked={ !! config.enabled_account_link }
 					onChange={ value => updateConfig( 'enabled_account_link', value ) }
 				/>
-				<TextControl
-					label={ __( 'Newsletter subscription text on registration', 'newspack' ) }
-					help={ __(
-						'The text to display while subscribing to newsletters on the registration modal.',
-						'newspack'
-					) }
-					value={ config.newsletters_label }
-					onChange={ value => updateConfig( 'newsletters_label', value ) }
-				/>
 				<Grid columns={ 2 } gutter={ 16 }>
 					<TextControl
 						label={ __( 'Terms & Conditions Text', 'newspack' ) }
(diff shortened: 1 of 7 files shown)


dkoo (changes requested):
So far so good, but I got stuck at step 10 ("Disable synchronization and repeat steps 4-9"). I couldn't actually disable synchronization—every time I unchecked the box and clicked "save settings", it came back.

miguelpeixe:
There was a bug due to a [core issue](https://core.trac.wordpress.org/ticket/40007). `update_option()` doesn't create the option if the value is empty, so it was stuck on the coded default value of `true`.

e74e85afd93684b8fa7f6f6b7807d476505b402f fixes it by casting the value to int before storing

adekbadek:
>  the ESP synced deletion only occurs when readers self-delete their account (which triggers Reader_Activation::delete_reader()). The regular wp_delete_user(), which is used by admins on the WP dashboard, does not. Should we reconsider this?

I think the contact should be deleted (if the feature is enabled) when the WP user is removed by any means – via core WP users view, WP CLI, etc. WDYT @dkoo ?

dkoo:
> I think the contact should be deleted (if the feature is enabled) when the WP user is removed by any means – via core WP users view, WP CLI, etc. WDYT @dkoo ?

I agree—not deleting them would mean lots of "orphaned" contacts in the ESP with no good way to see which ones were deleted at the source.

miguelpeixe:
Done in 80b9b125a00e6a3059ad3240a871733b665585e7 and https://github.com/Automattic/newspack-newsletters/pull/929/commits/4ed4eca4fb3565684ffffe5f7cf3c21f0641447b

includes/reader-revenue/templates/myaccount-edit-account.php
```diff
@@ -67,10 +68,24 @@
...
 	</a>
 </p>
 
+<hr />
+
+<h3><?php \esc_html_e( 'Delete Account', 'newspack' ); ?></h3>
+
+<p class="woocommerce-form-row woocommerce-form-row--wide form-row form-row-wide">
+	<a href="<?php echo '?' . \esc_attr( $newspack_delete_account_arg ) . '=' . \esc_attr( \wp_create_nonce( $newspack_delete_account_arg ) ); ?>">
+		<?php \esc_html_e( 'Email me a link for account deletion', 'newspack' ); ?>
```
dkoo, line 87: Non-blocking suggestion:

```suggestion
		<?php \esc_html_e( 'Request account deletion', 'newspack' ); ?>
```

assets/wizards/engagement/views/reader-activation/index.js
```diff
@@ -112,17 +103,53 @@ export default withWizardScreen( () => {
...
-				<ActiveCampaign
-					value={ { masterList: config.active_campaign_master_list } }
-					onChange={ ( key, value ) => {
-						if ( key === 'masterList' ) {
-							updateConfig( 'active_campaign_master_list', value );
-						}
-					} }
+				<hr />
+				<SectionHeader
+					title={ __( 'Email Service Provider and Newsletters Integration', 'newspack' ) }
```
dkoo, line 108: Non-blocking: this seems a bit wordy and redundant with the description below. How about:

```suggestion
					title={ __( 'Email Service Provider Settings', 'newspack' ) }
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -122,33 +127,170 @@ public static function handle_password_reset_request() {
...
+		$token           = isset( $_POST['token'] ) ? \sanitize_text_field( $_POST['token'] ) : '';
+		$transient_token = \get_transient( 'np_reader_account_delete_' . $user_id );
+		if ( ! $token || ! $transient_token || $token !== $transient_token ) {
+			wp_die( \esc_html__( 'Invalid request.', 'newspack' ) );
+		}
+		\delete_transient( 'np_reader_account_delete_' . $user_id );
+
+		\wp_delete_user( $user_id );
+
+		\wc_add_notice( __( 'Your account has been deleted.', 'newspack' ), 'success' );
```
dkoo, line 291: Nit: this message is never visible, since as soon as you confirm deletion of your account, you're dumped back to the login screen. Ideally we should be able to show notices in the login screen as well.

dkoo (changes requested):
Everything is working as expected, except the new feature to delete the contact whenever the user account is deleted by any means (not just user input). When I delete a user via WP admin, I'm not seeing the contact deleted in the ESP regardless of the "Synchronize reader to ESP" and "Delete contact on reader deletion" settings.

dkoo (approved):
As discussed and confirmed, the ESP deletion happens only if the reader has verified their account—even if the deletion request comes from an admin user deleting the user's account via WP admin or CLI. An option @miguelpeixe proposed if this behavior causes publisher confusion is to make the "delete" option a dropdown:

```
Delete contact on reader deletion:
- Never
- With verified email
- All readers
```

But let's hold on that until we decide it's necessary.

With that cleared up, the other comments in the review are related to wording and non-blocking, except maybe for [this one](https://github.com/Automattic/newspack-plugin/pull/1884#discussion_r947190076), which is a UX thing and therefore subjective.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
miguelpeixe: 11f8b180544a86025aefccf2f56abcd9130fb8ff fixes the notice strategy

dkoo (approved):

PR #1895, opened 2022-08-15
feat: new option for minimum donation amount

### LINKED ISSUE #1886: Reader Revenue: Increase minimum default for WooCommerce Donate products
The minimum price for all WooCommerce donate products should be set to $5.

dkoo:
### All Submissions:

Establishes a new donation setting in the Reader Revenue wizard for "minimum donation", with a default value of `5`. This setting applies to both WC and Stripe platforms and fails donate block form submissions if the donation amount submitted is less than this minimum. Also ensures that the minimum value is updated to `5` automatically for all sites, so that no publisher input will be required for the new default to take effect.

Closes #1886 and https://github.com/Automattic/newspack-blocks/issues/1211.

1. On `master` on a site with Reader Revenue platform set to Stripe, publish a page with a Donate block.
2. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1239.
3. In the Newspack dashboard, visit **Reader Revenue > Donations**. Confirm there's a new setting for minimum donation toward the bottom of this page with a default value of `5`. Don't update the setting yet.

Screen Shot 2022-08-15 at 12 36 31 PM

4. On the page with your Donate block, try to submit a donation with an amount less than $5. (You may have to use the "other" option and input the value yourself.) Confirm that the donation fails with an error message.
5. In the Donations settings page, change the minimum donation amount to another value. Confirm that:
  a. The values for other amounts (tiered and untiered) can't be set below the minimum amount.
  b. If increasing the minimum amount to a value above any other amounts, those amounts are updated to match the minimum value. (This is to ensure that you can't create a Donate block with suggested amounts that would be below the minimum donation, and trigger an error message.)
6. Save settings and repeat step 4. Confirm that the block now errors out at your new minimum value.
7. Edit the page with the block and toggle on the "Configure manually" option.
8. Confirm that there's a new block attribute setting for "Minimum donation" and that it behaves identically to the option in the Reader Revenue wizard (e.g. repeat step 5 in block settings).
9. Save the block and repeat step 4. Confirm that the block now errors out at the minimum value set in block attributes, instead of the value set in the Reader Revenue wizard.
10. Switch your Reader Revenue platform to "Newspack" (WooCommerce) and repeat steps 3–9. On a site using WC, the Donate block should error out before you get to the checkout page.

### Other information:

### DIFF
--- a/assets/wizards/readerRevenue/views/donation/index.tsx
+++ b/assets/wizards/readerRevenue/views/donation/index.tsx
@@ -53,7 +53,7 @@ type WizardData = {
 				};
 				currencySymbol: string;
 				tiered: boolean;
-				minimumDonation: number;
+				minimumDonation: string;
 		  };
 	donation_page: {
 		editUrl: string;
@@ -83,6 +83,9 @@ export const DonationAmounts = () => {
 		key: slug,
 		...FREQUENCIES[ slug ],
 	} ) );
+
+	// Minimum donation is returned by the REST API as a string.
+	const minimumDonationFloat = parseFloat( minimumDonation );
 
 	return (
 		<>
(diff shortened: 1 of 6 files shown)


assets/wizards/readerRevenue/views/donation/index.tsx
```diff
@@ -129,19 +132,34 @@ export const DonationAmounts = () => {
 											<MoneyInput
 												currencySymbol={ currencySymbol }
 												label={ __( 'Low-tier' ) }
-												value={ amounts[ section.key ][ 0 ] }
+												value={
+													amounts[ section.key ][ 0 ] >= minimumDonation
+														? amounts[ section.key ][ 0 ]
+														: minimumDonation
```
adekbadek, line 138: That's too heavy-handed. This does not allow input of an amount that starts with an integer lower than the minimum. So `12` will be discarded. I think an error message is a better idea than coercing the value to a minimum.

includes/wizards/class-reader-revenue-wizard.php
```diff
@@ -275,6 +275,12 @@ public function api_update( $request ) {
 			}
 			update_option( NEWSPACK_NRH_CONFIG, $nrh_config );
 		}
+
+		// Ensure that any Reader Revenue settings changed while the platform wasn't WC are persisted to WC products.
+		if ( 'wc' === $platform ) {
```
adekbadek, line 280:
```suggestion
		if ( Donations::is_platform_wc() ) {
```

adekbadek (changes requested):

assets/wizards/readerRevenue/views/donation/index.tsx
dkoo: Ah, that's an interesting point. Will refactor to show an error message instead.

dkoo: 0abcf88 adds warnings instead of coercing the values:

Screen Shot 2022-08-16 at 10 45 05 AM

assets/wizards/readerRevenue/views/donation/index.tsx
```diff
@@ -159,13 +192,35 @@ export const DonationAmounts = () => {
 								currencySymbol={ currencySymbol }
 								label={ section.staticLabel }
 								value={ amounts[ section.key ][ 3 ] }
+								min={
+									amounts[ section.key ][ 3 ] >= minimumDonationFloat
+										? amounts[ section.key ][ 3 ]
+										: minimumDonationFloat
+								}
```
adekbadek, line 199: `min` attribute should have the value of `minimumDonationFloat`, shouldn't it? And then the `error` attribute should compare the input value and the minimum. Currently the error message won't be displayed if "Untiered" selection is made.

adekbadek (changes requested):

assets/wizards/readerRevenue/views/donation/index.tsx
dkoo: Whoops, that was an oversight. 15143df should fix both of these issues.

adekbadek (approved):

PR #1852, opened 2022-08-04
Reorganise reader activation wizard with ActionCard

thomasguillot:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Instead of using a section header then a toggle, I combined all the settings into an ActionCard/Toggle card.

localhost_10004_wp-admin_admin php_page=newspack-engagement-wizard

### How to test the changes in this Pull Request:

1. Check the reader activation wizard
2. Switch to this branch
3. Refresh wizard
4. Test the various settings

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1910, opened 2022-08-17
feat: use reCAPTCHA to secure all Reader Activation-related forms

dkoo:
### All Submissions:

Implements reCAPTCHA v3 verification for all Reader Activation-related forms: registration, auth, subscribe (requires https://github.com/Automattic/newspack-newsletters/pull/935), and donation (requires https://github.com/Automattic/newspack-blocks/pull/1244).

As part of this PR, the wizard UI to enable reCAPTCHA and enter credentials have been moved from the **Reader Revenue > Stripe Settings** wizard to the **Engagement > Reader Activation** wizard. However, because reCAPTCHA is also used by the subscribe and streamlined donation blocks whether or not Reader Activation is enabled on the site, the reCAPTCHA script and methods can be enqueued and used even if Reader Activation features are not enabled.

Screen Shot 2022-08-16 at 6 46 28 PM

Also note that the UI for the reCAPTCHA settings may look out of place right now, but are styled to fit in with [this pending PR](https://github.com/Automattic/newspack-plugin/pull/1852).

Also implements the correct capitalization of reCAPTCHA in all UI.

1. On `master` for this plugin, Newsletters, and Blocks, confirm that reCAPTCHA is set up and working on your test site by following testing steps 3–7 from [this PR](https://github.com/Automattic/newspack-blocks/pull/1234).
2. Check out this branch, (https://github.com/Automattic/newspack-newsletters/pull/935) and https://github.com/Automattic/newspack-blocks/pull/1244. Confirm that the UI to enable reCAPTCHA and enter creds and all related values are automatically migrated from **Reader Revenue > Stripe Settings** to **Engagement > Reader Activation**.
3. Confirm that reCAPTCHA still works on streamlined donate blocks by setting the reCAPTCHA threshold (now on [this line](https://github.com/Automattic/newspack-plugin/compare/feat/recaptcha-for-all-forms?expand=1#diff-3ce140c40dae980330bca82b55eaf67a4f7abbbcc4f6609368fcf55ea696a7a7R24)) to `1.0` and trying to submit a donation. You should see an error message stating `User action failed captcha challenge`.
4. With the threshold still at `1.0`, confirm that reCAPTCHA now also works on login, subscribe, and register blocks.
5. Disable Reader Activation (either via the wizard option or by removing the environment variable from `wp-config.php`) and confirm that reCAPTCHA still works for streamlined donate and subscribe blocks, both of which don't require Reader Activation in order to function.
6. Re-enable Reader Activation, but disable reCAPTCHA via the wizard UI. Confirm that donation, login, subscribe, and register forms all work again without reCAPTCHA.

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -279,6 +279,15 @@ function process_form() {
 		);
 	}
 
+	// reCAPTCHA test.
+	if ( ! empty( $_REQUEST['captcha_token'] ) ) {
+		$captcha_token  = \sanitize_text_field( $_REQUEST['captcha_token'] );
+		$captcha_result = Reader_Activation::verify_captcha( $captcha_token );
+		if ( \is_wp_error( $captcha_result ) ) {
+			return send_form_response( $captcha_result );
+		}
+	}
+
 	// Note that that the "true" email address field is called `npe` due to the honeypot strategy.
 	// The honeypot field is called `email` to hopefully capture bots that might be looking for such a field.
 	$email = isset( $_REQUEST['npe'] ) ? \sanitize_email( $_REQUEST['npe'] ) : '';
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -9,6 +9,7 @@ namespace Newspack\Blocks\ReaderRegistration;
 
 use Newspack;
 use Newspack\Reader_Activation;
+use Newspack\Recaptcha;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -280,9 +281,9 @@ function process_form() {
 	}
 
 	// reCAPTCHA test.
-	if ( ! empty( $_REQUEST['captcha_token'] ) ) {
-		$captcha_token  = \sanitize_text_field( $_REQUEST['captcha_token'] );
-		$captcha_result = Reader_Activation::verify_captcha( $captcha_token );
+	if ( Recaptcha::can_use_captcha() ) {
+		$captcha_token  = isset( $_REQUEST['captcha_token'] ) ? \sanitize_text_field( $_REQUEST['captcha_token'] ) : '';
+		$captcha_result = Recaptcha::verify_captcha( $captcha_token );
 		if ( \is_wp_error( $captcha_result ) ) {
 			return send_form_response( $captcha_result );
 		}
--- a/assets/reader-activation/index.js
+++ b/assets/reader-activation/index.js
@@ -200,6 +200,32 @@ export function getAuthStrategy() {
 }
 
 /**
+ * Get a captcha token based on user input
+ */
+export async function getCaptchaToken( action = 'submit' ) {
+	return new Promise( ( res, rej ) => {
+		const { grecaptcha, newspack_reader_activation_data } = window;
+		if ( ! grecaptcha || ! newspack_reader_activation_data ) {
+			return res( '' );
+		}
+
+		const { captcha_site_key: captchaSiteKey } = newspack_reader_activation_data;
+		if ( ! grecaptcha?.ready || ! captchaSiteKey ) {
+			rej( 'Error loading the reCaptcha library.' );
+		}
+
+		grecaptcha.ready( async () => {
+			try {
+				const token = await grecaptcha.execute( captchaSiteKey, { action } );
+				return res( token );
+			} catch ( e ) {
+				rej( e );
+			}
+		} );
+	} );
+}
+
+/**
  * Initialize store data.
  */
 function init() {
@@ -222,6 +248,7 @@ const readerActivation = {
 	hasAuthLink,
 	setAuthStrategy,
 	getAuthStrategy,
+	getCaptchaToken,
 };
 window.newspackReaderActivation = readerActivation;
 
--- a/assets/reader-activation/index.js
+++ b/assets/reader-activation/index.js
@@ -202,7 +202,7 @@ export function getAuthStrategy() {
 /**
  * Get a captcha token based on user input
  */
-export async function getCaptchaToken( action = 'submit' ) {
+export function getCaptchaToken( action = 'submit' ) {
 	return new Promise( ( res, rej ) => {
 		const { grecaptcha, newspack_reader_activation_data } = window;
 		if ( ! grecaptcha || ! newspack_reader_activation_data ) {
@@ -214,13 +214,11 @@ export async function getCaptchaToken( action = 'submit' ) {
 			rej( 'Error loading the reCaptcha library.' );
 		}
 
-		grecaptcha.ready( async () => {
-			try {
-				const token = await grecaptcha.execute( captchaSiteKey, { action } );
-				return res( token );
-			} catch ( e ) {
-				rej( e );
-			}
+		grecaptcha.ready( () => {
+			grecaptcha
+				.execute( captchaSiteKey, { action } )
+				.then( token => res( token ) )
+				.catch( e => rej( e ) );
 		} );
 	} );
 }
--- a/includes/class-reader-activation.php
+++ b/includes/class-reader-activation.php
@@ -16,10 +16,12 @@ final class Reader_Activation {
 
 	const OPTIONS_PREFIX = 'newspack_reader_activation_';
 
-	const AUTH_READER_COOKIE    = 'np_auth_reader';
-	const AUTH_INTENTION_COOKIE = 'np_auth_intention';
-	const SCRIPT_HANDLE         = 'newspack-reader-activation';
-	const AUTH_SCRIPT_HANDLE    = 'newspack-reader-auth';
+	const AUTH_READER_COOKIE      = 'np_auth_reader';
+	const AUTH_INTENTION_COOKIE   = 'np_auth_intention';
+	const SCRIPT_HANDLE           = 'newspack-reader-activation';
+	const AUTH_SCRIPT_HANDLE      = 'newspack-reader-auth';
+	const RECAPTCHA_SCRIPT_HANDLE = 'newspack-recaptcha';
+	const RECAPTCHA_THRESHOLD     = 0.5;
 
 	/**
 	 * Reader user meta keys.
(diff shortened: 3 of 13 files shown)


assets/blocks/reader-registration/index.php
```diff
@@ -279,6 +279,15 @@ function process_form() {
 		);
 	}
 
+	// reCAPTCHA test.
+	if ( ! empty( $_REQUEST['captcha_token'] ) ) {
```
miguelpeixe, line 283: If this is not sent and reCaptcha is enabled it should fail. This allows for the captcha validation to be skipped.

includes/class-reader-activation.php
```diff
@@ -979,6 +1037,14 @@ public static function process_auth_form() {
 			);
 		}
 
+		// reCAPTCHA test.
+		if ( ! empty( $captcha_token ) ) {
```
miguelpeixe, line 1041: Same captcha check. We shouldn't skip if captcha is enabled.

miguelpeixe (changes requested):
Nice! It's working great for the Sign In modal for authentication and registration as well as the reader registration and donation blocks.

I feel like the reader activation class is getting a bit crowded and we can consider this as a non-core aspect of RAS. WDYT of moving the functionality to a `class-recaptcha.php`?

I was not able to get it working on the Newsletters' subscription block due to a JS error. I'll comment on it further on the PR.

dkoo:
> I feel like the reader activation class is getting a bit crowded and we can consider this as a non-core aspect of RAS. WDYT of moving the functionality to a `class-recaptcha.php`?

I can support that notion. What do you think is a good location for the wizard UI, though? Connections?

miguelpeixe:
> What do you think is a good location for the wizard UI, though? Connections?

I didn't think of moving its options, but Connections do make sense to me, being a third-party API. Maybe @adekbadek have another idea?

adekbadek:
I think Connections Wizard is a fine place for this. This service is universal for any forms, so it makes sense put it in a generic "third-party integrations" place.

dkoo:
Thanks for the discussion—will revisit this and move the UI to the Connections wizard.

includes/class-reader-activation.php
```diff
@@ -130,6 +162,9 @@ private static function get_settings_config() {
 			'sync_esp'                    => true,
 			'sync_esp_delete'             => true,
 			'active_campaign_master_list' => '',
+			'useCaptcha'                  => false,
+			'captchaSiteKey'              => '',
+			'captchaSiteSecret'           => '',
```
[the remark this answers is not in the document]
dkoo: Do you mean the option names should be like `reCAPTCHASiteSecret`?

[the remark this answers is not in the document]
miguelpeixe: No, the other options from this list are underscore case.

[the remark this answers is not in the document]
dkoo: Changed the settings to snake case in 52b21a5. Note that if you had previously tested this PR and already migrated the settings from the Stripe wizard, the settings won't be migrated from the old camelcase option names. The camelcase settings will still be migrated from Stripe settings if they exist.

To test this, check out `master` again, go to **Reader Revenue > Stripe Settings** and enter reCAPTCHA credentials in the wizard there, then check out this branch again to ensure that the settings are migrated to the Connections wizard.

assets/blocks/reader-registration/index.php
dkoo: Whoops should fix this. Now the check is for `can_use_captcha` instead of whether the token is empty.

includes/class-reader-activation.php
dkoo: 11e2969 should fix this, too.

includes/class-reader-activation.php
```diff
@@ -1300,5 +1366,80 @@ public static function get_from_email() {
...
+				[
+					'secret'   => $captcha_secret,
+					'response' => $captcha_token,
+				],
+				'https://www.google.com/recaptcha/api/siteverify'
+			)
+		);
+
+		// If the reCaptcha verification request fails.
+		if ( is_wp_error( $captcha_verify ) ) {
```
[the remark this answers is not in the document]
dkoo: Fixed in 71d67ba.

includes/class-reader-activation.php
```diff
@@ -1300,5 +1366,80 @@ public static function get_from_email() {
...
+		if ( ! $captcha_token ) {
+			return new \WP_Error(
+				'newspack_recaptcha_invalid_token',
+				__( 'Missing or invalid captcha token.', 'newspack' )
+			);
+		}
+
+		$captcha_secret = self::get_setting( 'captchaSiteSecret' );
+		$captcha_verify = wp_safe_remote_post(
+			add_query_arg(
```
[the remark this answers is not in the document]
dkoo: Fixed in 71d67ba.

includes/class-reader-activation.php
```diff
@@ -1300,5 +1366,80 @@ public static function get_from_email() {
...
+
+		if ( ! $captcha_token ) {
+			return new \WP_Error(
+				'newspack_recaptcha_invalid_token',
+				__( 'Missing or invalid captcha token.', 'newspack' )
+			);
+		}
+
+		$captcha_secret = self::get_setting( 'captchaSiteSecret' );
+		$captcha_verify = wp_safe_remote_post(
```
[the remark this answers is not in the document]
dkoo: Fixed in 71d67ba.

includes/class-reader-activation.php
```diff
@@ -68,29 +72,57 @@ public static function init() {
...
+
+			// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
+			wp_register_script(
+				self::RECAPTCHA_SCRIPT_HANDLE,
+				esc_url( 'https://www.google.com/recaptcha/api.js?render=' . $captcha_site_key ),
+				null,
+				null,
+				true
+			);
+			\wp_script_add_data( self::RECAPTCHA_SCRIPT_HANDLE, 'amp-plus', true );
```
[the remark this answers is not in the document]
dkoo: Added in  11e2969.

includes/class-reader-activation.php
```diff
@@ -68,29 +72,57 @@ public static function init() {
...
+	 * Does not require Reader Activation features to be active.
+	 */
+	public static function register_recaptcha() {
+		if ( self::can_use_captcha() ) {
+			$captcha_site_key = self::get_setting( 'captchaSiteKey' );
+
+			// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
+			wp_register_script(
+				self::RECAPTCHA_SCRIPT_HANDLE,
+				esc_url( 'https://www.google.com/recaptcha/api.js?render=' . $captcha_site_key ),
```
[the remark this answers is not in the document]
dkoo: Fixed in 71d67ba.

includes/class-reader-activation.php
```diff
@@ -68,29 +72,57 @@ public static function init() {
...
+	/**
+	 * Register the reCAPTCHA v3 script.
+	 * Does not require Reader Activation features to be active.
+	 */
+	public static function register_recaptcha() {
+		if ( self::can_use_captcha() ) {
+			$captcha_site_key = self::get_setting( 'captchaSiteKey' );
+
+			// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
+			wp_register_script(
```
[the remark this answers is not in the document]
dkoo: Fixed in 71d67ba.

includes/class-recaptcha.php
```diff
@@ -0,0 +1,292 @@
...
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class for reCAPTCHA integration.
+ */
+final class Recaptcha {
+	const RECAPTCHA_SCRIPT_HANDLE = 'newspack-recaptcha';
+	const RECAPTCHA_THRESHOLD     = 0.5;
```
miguelpeixe, line 17:
```suggestion
	const SCRIPT_HANDLE           = 'newspack-recaptcha';
	const SCORE_THRESHOLD         = 0.5;
```

Prefixing with `RECAPTCHA_` is a bit redundant inside the `\Newspack\Recaptcha` class.

includes/class-recaptcha.php
```diff
@@ -0,0 +1,292 @@
...
+	const RECAPTCHA_API_NAMESPACE = 'newspack/v1';
```
miguelpeixe, line 18: We already have a global constant for the `newspack/v1` endpoint called `NEWSPACK_API_NAMESPACE`

miguelpeixe (changes requested):
Thank you for the refactor! It's looking great! Left a couple of nits

assets/reader-activation/index.js
```diff
@@ -199,6 +199,32 @@ export function getAuthStrategy() {
...
+		grecaptcha.ready( async () => {
+			try {
+				const token = await grecaptcha.execute( captchaSiteKey, { action } );
+				return res( token );
+			} catch ( e ) {
+				rej( e );
+			}
+		} );
+	} );
+}
```
miguelpeixe, line 226: I think we should either `async/await` or `Promise`s here.

includes/class-recaptcha.php
dkoo: Renamed in 595102e (and https://github.com/Automattic/newspack-newsletters/pull/935/commits/00dbd1761e246f8cf3b672c47001ac1be3c75f4a and https://github.com/Automattic/newspack-blocks/pull/1244/commits/be6eafb5878e40750442b2a74bcbf423be12ac3f)

includes/class-recaptcha.php
dkoo: Ah, good call: 2bf452c

assets/reader-activation/index.js
dkoo: 28a5869 refactors to remove the use of `async`/`await` on the front-end, which seems to be a problem for some browsers if not polyfilled.

dkoo: f497ba0 also fixes `auth.js`—the form field processing needs to happen in a `finally` statement in order to avoid a race condition between getting the reCAPTCHA token and submitting the form data.

miguelpeixe (approved):
Thank you for the revisions! 🙌

PR #1919, opened 2022-08-18
fix: x scrollbar on "Sign In" modal

miguelpeixe:
### All Submissions:

Fix horizontal scroll being rendered on the Sign In modal.

image

1. On master, open the Sign In modal and observe the horizontal scroll as displayed above
2. Check out this branch and confirm the scrollbar is no longer there

### Other information:

### DIFF
--- a/assets/reader-activation/auth.scss
+++ b/assets/reader-activation/auth.scss
@@ -184,6 +184,7 @@
 			position: relative;
 			max-height: 100vh;
 			overflow-y: auto;
+			overflow-x: hidden;
 
 			transition: height 125ms ease-in-out;
 


dkoo (approved):
I'm not seeing the horizontal scrollbar or any element exceeding the modal, but this is probably a good idea regardless. Do you have any idea what's causing the overflow?

miguelpeixe:
I couldn't pinpoint the exact reason, but it seems related to the `overflow-y: auto;` rule and webkit (Chrome/Safari). I couldn't reproduce it in Firefox.

adekbadek:
The reason is the new honeypot input field:

image

We could set its width to 0 to be have a less hacky solution. On the other hand, I can't think of a reason to _not_ block y-overflow in that modal.

dkoo:
Aha, good find! I wasn't seeing this behavior because I'm using Firefox as my main dev browser. I think blocking the overflow is a good solution. 👍

miguelpeixe:
The `.newspack-reader__auth-form__content input[type=email]` width rule is overriding the `width: 1px;` from the `visuallyHidden` mixin

PR #1927, opened 2022-08-23
fix: if logging in from an overlay prompt, dismiss the prompt after login

dkoo:
### All Submissions:

The default Reader Activation strategy serves a Registration block inside of an overlay prompt. The Registration block offers readers with existing accounts the option to log in, which opens another login modal on top of the Registration overlay prompt. If you then log in via this login modal and close the modal after successfully authenticating, the overlay prompt becomes visible again, but with the Registration prompt hidden (because the reader is now authenticated).

This PR detects if a login modal has been opened from inside an overlay prompt, and if the reader's status becomes authenticated at the point they close the login modal, will keep the overlay prompt of origin hidden as well.

1. Check out this branch
2. Publish an overlay prompt containing a Registration block
3. In a new session, click the "Already have an account? Sign in" link from inside the overlay prompt
4. Observe the login modal appearing while hiding the overlay prompt
5. Close the login modal without interacting with it, confirm that the Registration overlay prompt is shown again
6. Click "Already have an account? Sign in" from inside the prompt again
7. Log into an existing account with a password
8. Close the login modal
9. Confirm that the overlay prompt does NOT become visible again

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -60,14 +60,22 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 		}
 
 		let currentlyOpenOverlayPrompts = [];
+		let overlayPromptOrigin = null;
 		const hideCurrentlyOpenOverlayPrompts = () =>
 			currentlyOpenOverlayPrompts.forEach( promptElement =>
 				promptElement.setAttribute( 'amp-access-hide', '' )
 			);
-		const displayCurrentlyOpenOverlayPrompts = () =>
-			currentlyOpenOverlayPrompts.forEach( promptElement =>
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Works as described, however I'd also expect the prompt to be dismissed after requesting a magic link:

image

After clicking the modal closing button above, the registration prompt should nor reappear.

dkoo:
> Works as described, however I'd also expect the prompt to be dismissed after requesting a magic link:

That's a good point—07a0ab9 makes it so that the prompt remains dismissed as long as the reader has submitted an email address (so if they log in or request a magic link).

adekbadek (approved):

PR #1482, opened 2022-02-02
feat(block-editor): utility to relink the editor close button to a wizard screen

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds a small utility to relink the full-screen close button in the block editor to a corresponding Newspack wizard page, for Newspack CPTs that have one.

T6X6AN

Closes #1205.

### How to test the changes in this Pull Request:

1. On `master`, edit a Campaigns prompt. Observe that the full-screen close button (see screenshot above) links to the standard WP admin post list UI for Prompts, which is normally hidden when the Newspack plugin is active.
2. Check out this branch, edit the prompt. Now confirm that the close button links to the Campaigns wizard page instead.
3. Temporarily deactivate the Newspack Plugin and edit the prompt again. Confirm that the close button once again links to the WP admin post list for Prompts.
4. Reactivate the Newspack Plugin. Edit another standard WP post type (like a post or page). Confirm that the close button links to the WP admin post list for that post type, as usual.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1929, opened 2022-08-24
Reader Activation: customizable verification email

adekbadek:
### All Submissions:

Two closely related changes:

- Redesign of the verification email
- Enabling customization of the verification email (sent when a reader registers)

Default email design:

image

1. Since this changeset contains a refactor of the Reader Revenue email, test the post-Stripe-donation email flow:
    1. Make a donation with Stripe configured as the RR platform, observe the email is sent
    2. Edit the post-donation email, observe it's sent as the updated version
3. With Reader Activation enabled, register as a new reader
4. Observe an email is sent, with the updated design 
5. Visit Engagement wizard, "Reader Activation" screen, "Emails" section
6. Edit the "Verification" email
7. In a new session, register as a new reader. Observe the updated email is sent

### Other information:


includes/templates/reader-activation-emails/verification.php
```diff
@@ -0,0 +1,174 @@
...
...resentation" style="width:100%;"><tbody><tr><td style="direction:ltr;font-size:0px;padding:0;text-align:center;"><!--[if mso | IE]><table role="presentation" border="0" cellpadding="0" cellspacing="0"><tr><td class="" style="vertical-align:top;width:600px;" ><![endif]--><div class="mj-column-per-100 mj-outlook-group-fix" style="font-size:0px;text-align:left;direction:ltr;display:inline-block;vertical-align:top;width:100%;"><table border="0" cellpadding="0" cellspacing="0" role="presentation" width="100%"><tbody><tr><td style="vertical-align:top;padding:12px;"><table border="0" cellpadding="0" cellspacing="0" role="presentation" width="100%"><tbody><tr><td align="left" style="font-size:0px;padding:0;word-break:break-word;"><div style="font-family:Georgia;font-size:16px;line-height:1.5;text-align:left;color:#000000;"><p>Once verified, you\'ll be able to edit your profile, set a password for easier access, and manage your newsletter subscriptions and billing information.</p></div></td></tr></tbody></table></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table><![endif]--></div></body></html>',
```
dkoo, line 173: You have a development URL hard-coded here in the `a` and `img` tags—this should populate with the site's domain.

dkoo: Also, not sure we can rely on the logo's path in the `uploads` folder to always exist across sites—we should populate the URL for the logo as defined in the Design wizard and/or Customizer.

includes/templates/reader-activation-emails/verification.php
```diff
@@ -0,0 +1,174 @@
...
+  <p>To manage your account, please verify your email address by clicking the link below:</p>
+  <!-- /wp:paragraph -->
+
+  <!-- wp:buttons -->
+  <div class="wp-block-buttons"><!-- wp:button {"style":{"color":{"background":"#dd3333"}}} -->
+  <div class="wp-block-button"><a class="wp-block-button__link has-background" href="*VERIFICATION_URL*" style="background-color:#dd3333">Verify my email</a></div>
+  <!-- /wp:button --></div>
+  <!-- /wp:buttons -->
+
+  <!-- wp:paragraph {"style":{"typography":{"fontSize":"14px"}},"textColor":"secondary-variation"} -->
```
dkoo, line 28: Not sure I like how the font size for this paragraph is slightly smaller than the rest of the email—it creates a weird vertical rhythm that looks unintentional.

Screen Shot 2022-08-24 at 1 24 12 PM

includes/emails/class-emails.php
```diff
@@ -0,0 +1,464 @@
...
+			return false;
+		}
+		$serialized_email = [
+			'label'        => $email_config['label'],
+			'description'  => $email_config['description'],
+			'post_id'      => $post_id,
+			// Make the edit link relative.
+			'edit_link'    => str_replace( site_url(), '', get_edit_post_link( $post_id, '' ) ),
+			'subject'      => get_the_title( $post_id ),
+			'from_name'    => get_post_meta( $post_id, 'from_name', true ),
```
dkoo, line 324: For Reader Activation-related emails, the from name and email address should match those [set in the `Reader_Activation` class](https://github.com/Automattic/newspack-plugin/blob/master/includes/class-reader-activation.php#L1289-L1323).

Non-blocking enhancement: perhaps these from details should apply to all customizable emails, and be configurable in the UI.

includes/templates/reader-activation-emails/verification.php
```diff
@@ -0,0 +1,174 @@
...
+return array(
+	'post_title'   => __( 'Your verification link', 'newspack' ),
+	'post_content' => '<!-- wp:site-logo {"align":"center"} /-->
+
+  <!-- wp:heading -->
+  <h2>Hi there!</h2>
+  <!-- /wp:heading -->
+
+  <!-- wp:paragraph -->
+  <p>To manage your account, please verify your email address by clicking the link below:</p>
```
dkoo, line 19: Shouldn't all boilerplate copy here be run through `__()`?

includes/class-magic-link.php
```diff
@@ -350,13 +348,7 @@ public static function generate_email( $user, $redirect_to = '', $subject, $mess
...
 		}
 
-		// If the message contains a magic link placeholder, populate the magic link there.
-		if ( false !== strpos( $message, self::MAGIC_LINK_PLACEHOLDER ) ) {
-			$message = str_replace( self::MAGIC_LINK_PLACEHOLDER, $magic_link_url, $message );
-		} else {
-			// Otherwise, append it to the end of the message.
-			$message .= $magic_link_url . "\r\n";
-		}
+		$message .= $magic_link_url . "\r\n";
```
dkoo, line 351: Probably for a new PR, but noting that all Reader Activation-related emails should be redesigned and put through this new system.

includes/reader-revenue/class-stripe-connection.php
```diff
@@ -960,7 +960,7 @@ public static function handle_donation( $config ) {
 						'metadata' => $payment_metadata,
 					]
 				);
-				if ( ! Reader_Revenue_Emails::can_send_email( Reader_Revenue_Emails::EMAIL_TYPE_RECEIPT ) ) {
+				if ( ! Emails::can_send_email( 'receipt' ) ) {
```
dkoo, line 963: Nit: Maybe `receipt` can remain a constant in case we need to change this or add other types of receipts in the future?

dkoo (changes requested):
A few issues and non-blocking suggestions below. Other items which don't pertain to specific parts of the code:

* Clicking the full-screen close button from the editor (WordPress logo in the upper-right corner) brings you to a standard WP post list UI for the Newspack Emails CPT. We should use [this hook](https://github.com/Automattic/newspack-plugin/pull/1482) to redirect to the proper Newspack wizard UI instead.
* The instructions shown in the "Instructions" sidebar panel still mainly related to the Stripe receipt emails, which might be confusing for Reader Activation-related emails. Perhaps this should be a part of each email type's config so that we can show different instructions based on email type?

includes/emails/class-emails.php
dkoo: Note: https://app.asana.com/0/1202719992789620/1202857616664481/f also suggests that the emails could be easily rejected if the from email address looks suspicious to receipients' email address servers. This further highlights that the sender details should be customizable, and default to the `no-reply@sitedomain.com` address we're using for other Reader Activation emails.

assets/wizards/engagement/views/reader-activation/index.js
```diff
@@ -86,24 +107,41 @@ export default withWizardScreen( () => {
...
 						help={ __( 'URL to the page containing the terms and conditions.', 'newspack' ) }
-						value={ config.terms_url }
-						onChange={ value => updateConfig( 'terms_url', value ) }
+						{ ...getSharedProps( 'terms_url', 'text' ) }
 					/>
 				</Grid>
 				<hr />
+				{ emails?.length > 0 && (
+					<>
+						<SectionHeader
```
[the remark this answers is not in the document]
thomasguillot: I have a few ideas (still WIP) but ultimately we should gather all the various emails into its own wizard.

Emails (reader activation)

thomasguillot:
Not for this PR but something to maybe keep in mind, I've been working on a simple clean design for all these emails (still WIP)

But here's what I was thinking:

Gmail - White Background

Gmail - Coloured Background

Explanations

assets/wizards/engagement/views/reader-activation/index.js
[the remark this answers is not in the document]
adekbadek: > I'm also not sure I like how the various emails are spread out across the different wizards based on what features they're related to.

> ultimately we should gather all the various emails into its own wizard

I disagree. I think messaging is a part of a specific feature, not a feature in itself. When a user is e.g. setting up their donations – choosing default amounts and linking the payment provider – they should see the emails in this same UI, and be able to edit the donation receipt email as part of this setup. Dumping all editable emails is IMO an engineer-first (as opposed to user-first) UI, where the view follows the data entity, not a user flow.

adekbadek:
@dkoo – 

> The instructions shown in the "Instructions" sidebar panel still mainly related to the Stripe receipt emails […] Perhaps this should be a part of each email type's config

[But it is!](https://github.com/Automattic/newspack-plugin/blob/90fb086cf18aac1fcf4e919c53b0f811ea1d16b2/assets/other-scripts/emails/index.js#L81-L96) This is how the instructions look like for the verification email:

image

includes/class-magic-link.php
adekbadek: I agree, just didn't want to make this PR too large.

includes/templates/reader-activation-emails/verification.php
adekbadek: Fixed by using template placeholders for site URL and the site logo, in 7d047492fce77e2c25c155853d151b01cff55ca7

includes/templates/reader-activation-emails/verification.php
adekbadek: Updated in bef2cd40ef8c60bda40973b81f8e509da80403d4

includes/emails/class-emails.php
adekbadek: Done in 95c582a5e98d4720c164d309a78eb11e16ad25cd. The name & email can be provided in the email config, which will make them uneditable:

image

If these are not provided in the config, these fields just act as before, setting post meta.

includes/templates/reader-activation-emails/verification.php
adekbadek: Done in 635bbfe599297087e2c17bcc1aaf3ecad1c84f02

includes/reader-revenue/class-stripe-connection.php
adekbadek: Done in 062709e97634372d7e48729465fe1ed4007009af

assets/wizards/engagement/views/reader-activation/index.js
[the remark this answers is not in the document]
dkoo: I'm open to either approach, but either way it sounds like this is a discussion that should happen outside of this PR.

dkoo:
> [But it is!](https://github.com/Automattic/newspack-plugin/blob/90fb086cf18aac1fcf4e919c53b0f811ea1d16b2/assets/other-scripts/emails/index.js#L81-L96) This is how the instructions look like for the verification email:

You're right! I must have been looking at the wrong tab.

includes/class-newspack.php
```diff
@@ -254,7 +256,12 @@ public static function restrict_user_access( $current_screen ) {
 	 * @param WP_Screen $current_screen Current WP_Screen object.
 	 */
 	public function wizard_redirect( $current_screen ) {
-		$post_type_mapping = [];
+		$post_type_mapping = [
+			Emails::POST_TYPE => [
+				'base' => 'edit',
+				'url'  => esc_url( admin_url( 'admin.php?page=newspack' ) ),
```
dkoo, line 262: This is better than landing on the WP CPT post list for emails, but I wonder if we should add some more logic to redirect to the specific wizard page based on the email type? Engagement > Reader Activation for the verification email, and Reader Revenue > Emails for the Stripe reciept. If we're not consolidating the emails into a single wizard then it feels weird to be dumped back to the main Newspack dashboard page after closing out of the editor.

dkoo (changes requested):
Thanks for the revisions! One more thing that feels kind of strange about the close button behavior.

includes/class-newspack.php
adekbadek: @thomasguillot suggested elsewhere to use a hybrid approach to the email post type – link to specific emails where applicable, but also have a single "all emails" view. This will enable discoverability, but also allow to e.g. update branding on all emails from one place. 

I think the issue with the redirect is non-blocking, and the redirect can be updated to point to the "all emails" view once we have it.

dkoo: I'm okay with punting this to a new PR when we have an "All Emails" view.

dkoo (approved):

PR #1946, opened 2022-08-29
feat(analytics): prevent sending NTG newsletter event if subscribing to master list

adekbadek:
### All Submissions:

Fixes a small issue.

1. Configure a site w/ ActiveCampagin and set a "Master List" in the Engagement wizard
1. On `master`, register w/out subscribing to any newsletter
2. Observe an "NTG newsletter" event is sent anyway*
3. Switch to this branch, observe the event is not sent

\* This will be logged if `NEWSPACK_LOG_LEVEL` is set to `>= 2`

### Other information:

### DIFF
--- a/includes/plugins/class-newspack-newsletters.php
+++ b/includes/plugins/class-newspack-newsletters.php
@@ -202,11 +202,16 @@ class Newspack_Newsletters {
 	public static function get_lists_without_active_campaign_master_list( $list_ids ) {
 		$master_list_id = Reader_Activation::get_setting( 'active_campaign_master_list' );
 		if ( is_int( intval( $master_list_id ) ) ) {
-			if ( ( $key = array_search( $master_list_id, $list_ids ) ) !== false ) { // phpcs:ignore Squiz.PHP.DisallowMultipleAssignments.FoundInControlStructure
-				unset( $list_ids[ $key ] );
-			}
+			return array_values( // Reset keys.
+				array_filter(
+					$list_ids,
+					function( $id ) use ( $master_list_id ) {
+						return $id !== $master_list_id;
+					}
+				)
+			);
 		}
-		return array_values( $list_ids ); // Reset keys.
+		return $list_ids;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/plugins/class-newspack-newsletters.php
```diff
@@ -195,8 +195,22 @@ function( $acc, $item ) {
...
-	 * Ensure the contact is always added to ActiveCampaign's selected master
-	 * list.
+	 * Get lists without the master list, if set.
+	 *
+	 * @param int[] $list_ids List IDs to filter.
+	 */
+	public static function get_lists_without_active_campaign_master_list( $list_ids ) {
+		$master_list_id = Reader_Activation::get_setting( 'active_campaign_master_list' );
+		if ( is_int( intval( $master_list_id ) ) ) {
+			if ( ( $key = array_search( $master_list_id, $list_ids ) ) !== false ) { // phpcs:ignore Squiz.PHP.DisallowMultipleAssignments.FoundInControlStructure
```
dkoo, line 205: Nit: it took me minute to understand this line—why not use `array_values( array_filter() )` here? Also, I think the wrapping `array_values` might be needed in case the master list isn't at the end of the array—I'm not sure `unset` will re-index the other numeric keys.

dkoo (changes requested):
One question below.

includes/plugins/class-newspack-newsletters.php
adekbadek: > why not use array_values( array_filter() ) here? 

Hm, I think both solutions are suboptimal in terms of readability – with `array_filter` after finding the element, you still need the key to unset it 🤷. So we can stick with this one.

Added key reindexing and some tests in 828b26d06e2a514e1bd6da58a177ddd7c0eca8a6

dkoo: Your solution works great with the addition of `array_values`. 👍 

With `array_filter` you wouldn't need to `unset` at all:

```php
$list_ids = array_values(
	array_filter(
		$list_ids,
		function( $list_id ) use ( $master_list_id ) {
			return $list_id !== $master_list_id;
		}
	)
);
```

dkoo (approved):

includes/plugins/class-newspack-newsletters.php
adekbadek: Oh, I see now. Refactored in 91a5122e61758694b5ff3d7a86fa45f94045705e. Thanks!

PR #1949, opened 2022-08-29
feat(registration-block): hidden input for subscription

miguelpeixe:
### All Submissions:

This PR implements an option so that the registration block with only 1 newsletter list can have the list input as a hidden field. This allows for the block to have newsletter subscription messaging.

image

1. Check out this branch and edit a Reader Registration Block
2. Confirm you see the new "Hide input and always subscribe" option
3. Select only one newsletter and confirm the input is no longer disabled
4. Enable the option
5. Confirm the list is hidden
6. Visit the frontend and confirm the list is hidden and you can register and subscribe to that list
7. Back to the editor, add another list to the block so it has 2
8. Confirm the button is disabled and the block is rendering the 2 lists inputs regardless of the stored option
9. Save and visit the frontend
10. Confirm both lists appear and you can subscribe to them

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -148,7 +148,7 @@ export default function ReaderRegistrationEdit( {
 									}
 								/>
 								<ToggleControl
-									label={ __( 'Hide input and always subscribe', 'newspack' ) }
+									label={ __( 'Hide newsletter selection and always subscribe', 'newspack' ) }
 									checked={ hideSubscriptionInput }
 									disabled={ inFlight || lists.length !== 1 }
 									onChange={ () =>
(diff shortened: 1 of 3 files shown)


assets/blocks/reader-registration/edit.js
```diff
@@ -142,6 +147,14 @@ export default function ReaderRegistrationEdit( {
 										setAttributes( { displayListDescription: ! displayListDescription } )
 									}
 								/>
+								<ToggleControl
+									label={ __( 'Hide input and always subscribe', 'newspack' ) }
```
adekbadek, line 151:
```suggestion
									label={ __( 'Hide newsletter selection and always subscribe', 'newspack' ) }
```

Also, IMO a help text should further describe this settings. Along the lines of "If a single list is selected in the options below, the block can hide the newsletter selection and always subscribe the new reader to the list".

adekbadek (approved):
Pre-approving, as only copy changes are needed.

PR #1954, opened 2022-08-30
fix(WooCommerce): error notice text color

miguelpeixe:
### All Submissions:

This PR ensures that WC error notices get the inherited text color instead of the color applied by WC (white).

1. On the master branch, sign in to a reader account and update the Newsletters preferences without making any change
2. Confirm the error message appears but the text is not visible (select the text to confirm you see "You must select newsletters to update.")
3. Check out this branch and confirm the text is now visible

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/style.scss
+++ b/includes/reader-revenue/my-account/style.scss
@@ -25,6 +25,7 @@
 		&--error,
 		&.woocommerce-error {
 			border: 1px solid #ff4364;
+			color: inherit;
 			&::before {
 				background: #ff000038;
 			}


miguelpeixe:
I wonder why yours have more consistent padding all around, current and suggested.

This is mine:

image

Looks like the site font family and sizing are creating this inconsistency. Possibly something related to line height vs. the actual intended padding/spacing. It may be best to address this in a separate PR.

dkoo:
> I wonder why yours have more consistent padding all around, current and suggested.

Oh, my doesn't—I just mocked up what it could look like if we targeted `ul` inside the error container and set the margins and padding intentionally. I have the changes locally so I can push up as a new PR after we merge this one.

miguelpeixe:
Sounds good!

My point is that your current (as-is) looks better than my current, but with the added padding it'll be more standardized either way.

PR #1959, opened 2022-08-31
feat: after account deletion message

adekbadek:
### All Submissions:

Adds a special page for handling after account deletion flow.

1. With Reader Activation set up, create and then delete an account as a reader
2. Observe that after deletion, this page appears instead of the login screen:

image

### Other information:

### DIFF
--- a/includes/reader-revenue/templates/myaccount-after-delete-account.php
+++ b/includes/reader-revenue/templates/myaccount-after-delete-account.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * My Account page for reader account deletion.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+?>
+
+<div>
+	<h1>
+		<?php echo esc_html__( 'Your account has been deleted.', 'newspack' ); ?>
+	</h1>
+	<p>
+		<?php echo esc_html__( "We're sorry to see you go. But hope you'll continue to follow our coverage.", 'newspack' ); ?>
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/templates/myaccount-after-delete-account.php
```diff
@@ -0,0 +1,24 @@
...
+
+defined( 'ABSPATH' ) || exit;
+?>
+
+<div>
+	<h1>
+		<?php echo esc_html__( 'Your account has been deleted.', 'newspack' ); ?>
+	</h1>
+	<p>
+		<?php echo esc_html__( "We're sorry to see you go. But hope you'll continue to follow our coverage.", 'newspack' ); ?>
```
dkoo, line 18:
```suggestion
		<?php echo esc_html__( "We're sorry to see you go. We hope you continue to enjoy our content and consider supporting us again in the future.", 'newspack' ); ?>
```

dkoo: Maybe link `supporting us again` to the donation page? Probably won't get much conversion that way, but never hurts.

dkoo (approved):
This works—pre-approving since there are only copy suggestions.

includes/reader-revenue/templates/myaccount-after-delete-account.php
adekbadek: That would mean assuming that a reader has donated before. And if someone deletes their account, it might not be the best place to pitch for financial support :D

PR #1963, opened 2022-08-31
fix: prefix WC My Account actions so we can decide which hooks to support

dkoo:
### All Submissions:

WooCommerce's My Account dashboard contains many action and filter hooks to allow other plugins and themes to extend and modify its UI. Our custom Reader Activation templates call those same hooks, but in most cases the modifications are probably undesired. This PR retains the hooks in place, but prefixes them with `newspack_`. The effect is that most modifications from other plugins won't take effect, but we can selectively allow certain hooks to run if needed by calling the original action hook on our prefixed ones, e.g.

```php
\add_action(
	'newspack_woocommerce_before_edit_account_form,
	function() {
		\do_action( 'woocommerce_before_edit_account_form' );
	}
);
```

1. Install and set up the [Mailchimp for WooCommerce plugin](https://wordpress.org/plugins/mailchimp-for-woocommerce/). This plugin appends an undesired newsletter opt-in checkbox to the My Account form.
2. Observe in My Account that the checkbox shows up under the account details fields:

Screen Shot 2022-08-31 at 1 20 41 PM

3. Check out this branch, refresh My Account, and confirm that the extra field no longer appears.

### Other information:

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -14,7 +14,7 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Recaptcha {
 	const SCRIPT_HANDLE  = 'newspack-recaptcha';
-	const THRESHOLD      = 0.5;
+	const THRESHOLD      = 0.25;
 	const OPTIONS_PREFIX = 'newspack_recaptcha_';
 
 	/**
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -14,7 +14,7 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Recaptcha {
 	const SCRIPT_HANDLE  = 'newspack-recaptcha';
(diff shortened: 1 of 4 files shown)


includes/class-recaptcha.php
```diff
@@ -14,7 +14,7 @@
  */
 final class Recaptcha {
 	const SCRIPT_HANDLE  = 'newspack-recaptcha';
-	const THRESHOLD      = 0.5;
+	const THRESHOLD      = 0.25;
```
adekbadek, line 17: Is this intended in this PR?

dkoo: Sorry, no. I was submitting so many test registrations that my submissions were starting to fail reCAPTCHA tests, so I had to bump down the threshold on my local site in order to continue testing. I guess that's a good sign it's working, though.

adekbadek (approved):

PR #1965, opened 2022-09-01
feat: custom messaging for reader without password

miguelpeixe:
### All Submissions:

This PR implements custom messaging for readers that have just created their account and are yet to set their first password:

image

The other "reset password" messaging inside the sent email has been updated to use "set a new password" to have a more agnostic feel to it.

image

1. Manually remove the saved posts from `newspack_rr_email` post type
2. Create a RAS account and click on the received verification link
3. On the My Account page, confirm the "Create a Password" section as above
4. Click the link to set a new password
5. Set the new password
6. Visit "My Account" and confirm it no longer displays the password creation messaging

### Other information:


aschweigert (commented):
Could we also contextually change the link text? Something like "Email a link to set my password" (for people who don't have a password yet) vs "Email a password reset link" (for people who do) would probably be a bit clearer. Then the copy above could be something like "In order to create your first password, click the link below to request a link to securely set a password" (and/or just add the link directly to the copy here so it avoids the potential ambiguity around which link below to click on, etc)

miguelpeixe:
The intention to let the user know it's a password reset regardless of creation is because there's more customization required that I find no practical to maintain:

image

It's best to resolve the confusion in the initial message and let the user know that the flow is a password reset.

aschweigert:
Ah, I see. What's the difficulty in modifying the copy in the reset email? just that it would be another template to maintain or is there a more technical reason?

miguelpeixe:
Because it'd be another template to maintain and the overall duplication of logic for the same goal. Password reset and password creation have no difference internally.

aschweigert:
I can appreciate the desire to not duplicate logic internally but these really are different things from a user's perspective. The potential for confusion here is high enough that I think the tradeoff is worth it if it means making the flow clearer for the end user.

miguelpeixe:
Since we still haven't created our own flow, the password reset form would remain unmodified with the page also titled "Reset Password" because WordPress core doesn't have such differentiation:

image

aschweigert:
Right, but the language they use ("enter new password", "save password", etc.) is still more accurate than the language we're using. A user creating a password is still creating a *new* password, so while this screen is still not ideal, I think we can live with that. The steps leading up to it though, we should try to make as clear as possible to avoid confusion.

miguelpeixe:
The email copy could be updated to have a more agnostic text. "Set a new password" instead of "Reset your password"

aschweigert:
Yeah, that's a good suggestion, maybe we just mirror the language used on the WP core page to avoid the word "reset" as much as possible for people who don't yet have a password, that's really the part that's most confusing.

miguelpeixe:
Sounds good! Thank you and sorry about the back and forth here!

aschweigert:
No worries, thanks for talking it through!

miguelpeixe:
PR description and code updated :)

includes/reader-activation/class-reader-activation.php
```diff
@@ -1221,9 +1266,14 @@ public static function register_reader( $email, $display_name = '', $authenticat
...
 			}
 
-			/** Add default reader related meta. */
+			/**
+			 * Add default reader related meta.
+			 */
 			\update_user_meta( $user_id, self::READER, true );
+			/** Email is not yet verified. */
 			\update_user_meta( $user_id, self::EMAIL_VERIFIED, false );
+			/** User hasn't set its own password yet. */
```
adekbadek, line 1275:
```suggestion
			/** User hasn't set their own password yet. */
```

adekbadek (approved):
Just one tweak in a comment needed.

PR #1967, opened 2022-09-02
feat: move already have account message above content

thomasguillot:
### All Submissions:

This PR moves the "Already have an account? Sign in" message next to the Title. It makes it more visible for existing users where to sign in.

__After__

after

__Before__

before

1. Switch to this branch
2. Test in the editor
3. Check front-end

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -206,33 +206,35 @@ export default function ReaderRegistrationEdit( {
 				{ editedState === 'initial' && (
 					<div className={ `newspack-registration ${ className }` }>
 						<form onSubmit={ ev => ev.preventDefault() }>
-							<div className="newspack-registration__have-account">
-								<p>
-									<RichText
-										onChange={ value => setAttributes( { haveAccountLabel: value } ) }
-										placeholder={ __( 'Already have an account?', 'newspack' ) }
-										value={ haveAccountLabel }
-										tagName="span"
-									/>{ ' ' }
-									<a href="/my-account" onClick={ ev => ev.preventDefault() }>
+							<div className="newspack-registration__header">
+								<RichText
+									onChange={ value => setAttributes( { title: value } ) }
+									placeholder={ __( 'Write title…', 'newspack' ) }
+									value={ title }
+									tagName="h2"
+								/>
+								<div className="newspack-registration__have-account">
+									<p>
 										<RichText
-											onChange={ value => setAttributes( { signInLabel: value } ) }
-											placeholder={ __( 'Sign In', 'newspack' ) }
-											value={ signInLabel }
+											onChange={ value => setAttributes( { haveAccountLabel: value } ) }
+											placeholder={ __( 'Already have an account?', 'newspack' ) }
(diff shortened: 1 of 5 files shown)


assets/blocks/reader-registration/style.scss
```diff
@@ -82,19 +95,35 @@
...
+		background: var( --newspack-secondary-color );
+		color: var( --newspack-secondary-contrast-color );
+		display: flex;
+		font-size: 1em;
+		justify-content: center;
+		left: 0;
+		line-height: 1.5;
+		min-height: 36px;
+		padding: 0 16px;
+		position: absolute;
```
adekbadek, line 108: If the Registration block is placed on a page (not in a prompt), this makes the banner stick to the top of the viewport.

assets/blocks/reader-registration/style.scss
```diff
@@ -190,6 +225,27 @@
 	}
 }
 
+/* Sepcific styles to apply to .newspack-popup when Reader Registration is displayed */
```
adekbadek, line 228: There's no conditional logic here – these styles will be applied to prompts regardless of their content:

image

adekbadek (changes requested):
I don't think it's a good idea to make this banner absolutely positioned. With the background color and placed above the email input, it's prominent enough.

assets/blocks/reader-registration/style.scss
thomasguillot: I completely reworked this in https://github.com/Automattic/newspack-plugin/pull/1967/commits/207127595e4426fea51c90cdc76c8bc5f0951fa7

- I didn't like having a "floating" banner at the top. Instead, I moved the messaging next to the block title. So it's similar to the Sign In modal.

assets/blocks/reader-registration/edit.js
```diff
@@ -206,15 +206,35 @@ export default function ReaderRegistrationEdit( {
...
-							<RichText
-								onChange={ value => setAttributes( { title: value } ) }
-								placeholder={ __( 'Block title…', 'newspack' ) }
-								value={ title }
-								tagName="h2"
-							/>
+							<div className="newspack-registration__header">
+								<RichText
+									onChange={ value => setAttributes( { title: value } ) }
+									placeholder={ __( 'Write title…', 'newspack' ) }
```
adekbadek, line 212: Nit: the tone here feels weird, how about 

```suggestion
									placeholder={ __( 'Add title…', 'newspack' ) }
```

like the WP post editor? And then same with the description input.

adekbadek (approved):
Added one wording suggestion.

assets/blocks/reader-registration/edit.js
thomasguillot: I copied the wording from the Cover block but I guess there are a lot of inconsistencies between blocks....
"Write content...", "Heading", "Add quote", "Content..."

Changed to "Add title" and "Add description" (no elipsis) in https://github.com/Automattic/newspack-plugin/pull/1967/commits/12c1a139669fd392e57d697f92c62cbab3096bfe

PR #1968, opened 2022-09-03
Improved Woo sync to ActiveCampaign

claudiulodro:
### All Submissions:

This PR syncs more data from WC customers to ActiveCampaign in an attempt to get parity with the sort of reader revenue data that syncs when using the Simplified Donate Block. It also adds some legacy handling and updating for Woo customers, so whenever they log in it'll sync the latest customer data.

The one bit I haven't been able to figure out from the codebase is how the name fields are synced to ActiveCampaign. I'm sure it's a small tweak, but I haven't got that working yet. Most of the other fields are synchronizing nicely, so the data should look the same in ActiveCampaign no matter what (Newspack-supported) reader revenue platform people are using.

1. Set up a site with an ActiveCampaign connection and Reader Revenue.
2. In an incognito window, go to `example.com/product/donate` and you should see the Woo Donate products Newspack creates.
3. Purchase a Monthly Donation. After checkout, in ActiveCampaign you should see the customer and synced data:
Screen Shot 2022-09-03 at 3 15 26 PM
4.  Delete the customer's contact in ActiveCampaign (to simulate a legacy Woo customer that hasn't been synced). Log out of the site. Log in as the customer. Verify the customer's data shows up in ActiveCampaign.
5. In separate incognito sessions, test with Yearly Donation and One-Time Donation. The data should sync for those also.

### Other information:

### DIFF
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -120,6 +120,11 @@ class WooCommerce_Connection {
 			return;
 		}
 
+		if ( self::CREATED_VIA_NAME === $order->get_created_via() ) {
+			// Only sync orders not created via the Stripe integration.
+			return;
+		}
+
 		$metadata_keys = Newspack_Newsletters::$metadata_keys;
 		$user_id       = $order->get_customer_id();
 		if ( ! $user_id ) {
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -24,7 +24,7 @@ class WooCommerce_Connection {
 	 */
 	public static function init() {
 		\add_action( 'admin_init', [ __CLASS__, 'disable_woocommerce_setup' ] );
-		
+
 		// WC Subscriptions hooks in and creates subscription at priority 100, so use priority 101.
 		\add_action( 'woocommerce_checkout_order_processed', [ __CLASS__, 'sync_reader_on_order_complete' ], 101 );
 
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/class-woocommerce-connection.php
```diff
@@ -24,7 +24,20 @@ class WooCommerce_Connection {
 	 */
 	public static function init() {
 		\add_action( 'admin_init', [ __CLASS__, 'disable_woocommerce_setup' ] );
-		\add_action( 'woocommerce_checkout_order_created', [ __CLASS__, 'register_reader' ] );
+		
+		// WC Subscriptions hooks in and creates subscription at priority 100, so use priority 101.
+		\add_action( 'woocommerce_checkout_order_processed', [ __CLASS__, 'sync_reader_on_order_complete' ], 101 );
+
+		\add_action( 'wp_login', [ __CLASS__, 'sync_reader_on_customer_login' ], 10, 2 );
```
claudiulodro, line 31: This does have some implications for the Simplified Donate Block, since whenever someone logs in it will re-sync the customer data that was synced from Stripe. I think that's probably a good thing though.

adekbadek (commented):
When testing I've noticed that it's more about data sync than activating the reader. Account creation as a side effect of donation will happen anyway, so these hooks should only deal with metadata updates. I've tweaked it in 5d06d608dccef846b819d230dc543c5c7ca59240. Let me know if it looks OK.

includes/reader-revenue/class-woocommerce-connection.php
adekbadek: Not necessarily a good thing :) The `payment_page` meta will be overridden when a "Stripe donor" logs in. I've added a condition so that only non-Newspack-Stripe order will use this sync in 1f5f72943acda5aa68bffe666b64c00e05da5893.

claudiulodro:
Much better! Thanks for the revisions! This works well in my testing and now the name comes through and everything.

adekbadek (approved):

PR #1500, opened 2022-02-09
Fix/theme mods setting

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes #1093.

When Newspack is set up, the parent theme is downloaded and set as theme. Then, when another theme is chosen in Setup (or Design) wizard, the "new" theme has to be downloaded first. When theme mods are set immediately after a theme is downloaded, the mods are not saved. For some reason, this can be resolved by calling `switch_theme` (again) in the function which calls `set_theme_mods`.

Also `TextareaControl` is tweaked a bit, so that the placeholder text does not look like it's the actual text.

### How to test the changes in this Pull Request:

Install the plugin on a fresh site and observe #1093 is not reproducible – make sure to switch theme in the "Design" step of the Setup wizard

On an existing site, the original issue can be reproduced this way, on `master`:

1. Set theme to Newspack Theme
2. Remove all other Newspack themes from `wp-content/themes`
3. Remove all Newspack theme mods (options with `theme_mods_newspack` prefix)
4. Load the Design wizard or Setup wizard, Design step
5. Set the theme to a different Newspack theme (so it will have to be downloaded)
6. Change primary and secondary colors 
7. Save in Design wizard or "Finish" in Setup wizard
8. Observe the colors were not saved

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1978, opened 2022-09-06
Site Design: add check to only change theme if not empty

### LINKED ISSUE #1976: Site Design: Changing 'settings' deactivates theme
**Describe the bug**
When you change an option on the Dashboard > Newspack > Site Design > Settings tab, the theme gets un-selected. This has happened on publisher sites, and I was able to recreate it on a Jurassic Ninja site and locally. On the Jurassic Ninja and localhost sites, only the Newspack plugin and Newspack Blocks plugins were active.

Changing settings on the Dashboard > Newspack > Site Design > Design tab does not cause the same issue.

**To Reproduce**

1. Navigate to Dashboard > Newspack > Site Design > Settings.
2. Change any of the settings.
3. Click 'Save'.
4. View the front-end of the site.

**Screenshots**

When you navigate to Appearance > Themes, no theme is selected: 

image

When you view the front-end, some of the theme files are loading (to render the HTML), but not the styles:

image

When you view the source, CSS files like 'Print' are still loading, but not the theme's main stylesheet -- instead, it has a blank directory for the main style.css file. This happens for the default theme, and all the child themes:

Screen Shot 2022-09-06 at 10 06 58 AM

**Expected behavior**

I would expect that the current theme would remain selected. My best guess is saving this screen also saves an empty value for the current theme, but I'm not sure where.

laurelfulford:
### All Submissions:

Best I can tell, [this `switch_theme()` function](https://github.com/Automattic/newspack-plugin/pull/1500/files#diff-b68d58272a59cbafbe7aae011ee40fb05e17f57faa5bf55a84836cb073898e0cR493) can pass an empty string as the theme value, causing a site's theme to get set to nothing. This happens when any setting is changed on the Dashboard > Newspack > Site Design > Settings tab, but does not cause issues on the Design tab. 

This PR adds an `! empty()` check on the theme, to make sure it's not empty before running `switch_theme()`. While this _seems_ to work in my testing, I'm not sure if it would be better to revisit the original approach in #1500 since there was some question around why it worked in the first place 🙂 

Closes #1976

To recreate the original issue:

1. Start on a site that's already been set up. 
2. Navigate to WP Admin > Newspack > Site Design > Settings.
3. Change any of the settings and Save.
4. View the front-end of the site and see that the CSS is basically not loading; if you navigate to WP Admin > Appearance > Themes, no theme will be applied. 

To test this fix:

1. Reset your site's theme if you haven't already. 
2. Apply this PR. 
3. Navigate to WP Admin > Newspack > Site Design > Settings.
4. Change any of the settings and Save.
5. View the front end and confirm that the site looks okay. 
6. Repeat the testing steps [here](https://github.com/Automattic/newspack-plugin/pull/1500) with this PR, and confirm that this change doesn't introduce any regressions.

### Other information:

### DIFF
--- a/includes/wizards/class-setup-wizard.php
+++ b/includes/wizards/class-setup-wizard.php
@@ -475,7 +475,9 @@ class Setup_Wizard extends Wizard {
 		// If the theme has to be installed, the set_theme_mod calls below will – for some reason – have no effect
 		// If there's a switch_theme call here, even though it's already called in Theme_Manager::install_activate_theme,
 		// correct mods will be saved.
-		switch_theme( $theme );
+		if ( ! empty( $theme ) ) {
+			switch_theme( $theme );
+		}
 
 		// Set homepage pattern.
 		if ( isset( $request['theme_mods']['homepage_pattern_index'] ) ) {


laurelfulford:
@adekbadek I'd really appreciate your eyes on this change, since you probably understand the original issue better than I do -- thanks!

adekbadek (approved):
Can't reproduce the issue*, but the added check is a good idea anyway. 

\* the value of `$theme` is not empty after saving settings

laurelfulford:
Thanks @adekbadek!

> Can't reproduce the issue...

I could only reproduce it on one of two localhost sites, so there might be more to reproducing it than I worked out 🤔 At least it's the kind of change that doesn't affect how it should work!

PR #1979, opened 2022-09-06
fix(emails): don't create post if pluggable functions are not available

miguelpeixe:
### All Submissions:

The `Emails::get_email_config_by_type( $type )` method creates the email post on demand if it's called and the email is not found. This can be problematic if the method is called before `wp-includes/pluggable.php` has not yet been loaded.

This PR prevents the method from creating the post if `is_user_logged_in()` is not defined, which means pluggable functions are not yet loaded.

This change shouldn't prevent the posts from being created because the method is called multiple times and the post should be created as soon as pluggable functions are available.

1. While on master, remove posts from the `newspack_rr_email` post type directly from the database
2. Refresh your dashboard page and confirm the error
3. Check out this branch, remove the posts from the database again
4. Refresh the dashboard and confirm there's no error and the email posts are created

### Other information:

### DIFF
--- a/includes/emails/class-emails.php
+++ b/includes/emails/class-emails.php
@@ -295,6 +295,8 @@ class Emails {
 	 *
 	 * @param string $type Type of the email.
 	 * @param int    $post_id Email post id.
+	 *
+	 * @return array|false The serialized email config or false if not available or supported.
 	 */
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Confirmed the bug and that this solution fixes it. 👍

PR #1936, opened 2022-08-26
Feat/stripe and wc subscriptions

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds basic synchronisation with WooCommerce Subscriptions. The Stripe subscriptions will be synchronised to WC Subscriptions, but won't yet be editable (cancellable) _from_ WooCommerce Subscriptions (will be in another PR). 

TODO: 

- [ ] subscription cancellation
- [ ] subscription reactivation
- [ ] subscription pausing 

### How to test the changes in this Pull Request:

1. Start with a site with Reader Activation configured, Stripe set as Reader Revenue Platform, and the WooCommerce suite (including WooCommerce Subscriptions) active. Ensure Stripe can reach your test site with a webhook call.
2. Create a new reader, e.g. by registering on the front-end
3. We'll use [Stripe's test clocks](https://stripe.com/docs/billing/testing/test-clocks) to travel in time - visit the [test clocks in Stripe dashboard](https://dashboard.stripe.com/test/test-clocks) and create a new one
4. Add a new customer in the test clock, and add a payment method to them
5. Edit the customer metadata and add the new reader WP user ID as `userId` metadata*
6. Still in Stripe's test clock, add a new monthly-donation-subscription to the customer
7. Visit WP Admin -> WooCommerce Subscriptions and observe a new Subscription has been created
8. _Advance time_ on the Stripe test clock by a month
9. Observe a subscription renewal happens – a new charge is made by Stripe, and a new order add in WooCommerce
10. Observe the order is associated with the WC Subscription

\* This is all because test-clock-customers can't be created as a result of the donation, as usual customers are. The test-clock-customers are on a different time plane and thus have to be created manually and linked to the Newspack reader

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1981, opened 2022-09-07
Stripe to WC sync CLI tool

adekbadek:
### All Submissions:

For various reasons, the data in WC might end up out of sync with Stripe. This PR adds a WP CLI command for synchronising data from Stripe to WC. 

Note that only WC Customers and WC Orders will be created/updated. Newspack does not yet integrate with WC Subscriptions (coming in https://github.com/Automattic/newspack-plugin/pull/1936). 

1. Set up a fresh site, with WC suite installed and linked to ta new Stripe instance
2. Install the whole WC suite (load Reader Revenue wizard to see what's missing)
3. Enable AMP Plus (`NEWSPACK_AMP_PLUS_ENABLED`), but don't enable Reader Activation   
4. Configure Reader Revenue to use Stripe as the platform
5. Visit the front-end and donate using the Donate block
6. Make a one-time donation
7. Observe a WC order is created, but not a WP user. The order is attributed to Guest customer.
8. Now head over to Stripe dashboard and delete the webhook, breaking the sync process. Don't load the Reader Revenue wizard now, cause this will re-create the webhook!
9. Make another donation on the site and confirm no WC order was created 
10. And now is the time for the fix: run `wp newspack stripe sync-customers-to-wc`
11. Observe the command output – it should inform you that:
    1.  two WC Customers were created 
    2. one Order was updated
    3. one Order was created
12. Observe in WC UI that there are two orders, two customer, and the orders are properly assigned 
13. You can also try messing up and fixing other scenarios:
    1. unlinking the WC customer from the order (set `_customer_user` post meta to `0` on the order post)
    2. deleting the WP user, who's linked to an order
    3. deleting the order that's linked to a user 

### Other information:

### DIFF
--- a/includes/reader-revenue/class-stripe-sync.php
+++ b/includes/reader-revenue/class-stripe-sync.php
@@ -1,0 +1,213 @@
+<?php
+/**
+ * Stripe Syncs.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Stripe Syncs.
+ */
+class Stripe_Sync {
+	/**
+	 * The final results object.
+	 *
+	 * @var array
+	 * @codeCoverageIgnore
+	 */
+	private static $results = [
+		'processed'         => 0,
+		'created'           => 0,
+		'found'             => 0,
+		'orders_updated'    => 0,
+		'orders_created'    => 0,
+		'skipped_customers' => 0,
+	];
+
+	/**
+	 * Initialize.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function init() {
+		\add_action( 'init', [ __CLASS__, 'wp_cli' ] );
+	}
+
+	/**
+	 * Process a Stripe customer.
+	 *
+	 * @param array $customer Customer object.
+	 * @param array $args Arguments.
+	 */
+	private static function process_stripe_customer( $customer, $args ) {
+		$email_address = $customer->email;
+		$is_dry_run    = false !== $args['dry-run'];
+
+		$all_charges = Stripe_Connection::get_customer_charges( $customer->id );
+
+		// Skip charges created by WC.
+		$charges = array_filter(
+			$all_charges,
+			function( $charge ) {
+				return ! isset( $charge->metadata->order_id );
+			}
+		);
+		if ( empty( $charges ) ) {
+			self::$results['skipped_customers']++;
+			return;
+		}
+
+		$wp_user = get_user_by( 'email', $email_address );
+		$user_id = false;
+		if ( $wp_user ) {
+			$user_id = $wp_user->ID;
+			self::$results['found']++;
+		} else {
+			// Create the WC Customer and update past orders (lookup is done by email).
+			if ( ! $is_dry_run ) {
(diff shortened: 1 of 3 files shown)


includes/reader-revenue/class-stripe-sync.php
```diff
@@ -0,0 +1,213 @@
...
+			$user_id = $wp_user->ID;
+			self::$results['found']++;
+		} else {
+			// Create the WC Customer and update past orders (lookup is done by email).
+			if ( ! $is_dry_run ) {
+				$full_name  = $customer->name;
+				$user_login = \sanitize_title( $full_name );
+				$user_id    = \wc_create_new_customer( $email_address, $user_login, '', [ 'display_name' => $full_name ] );
+				if ( is_wp_error( $user_id ) ) {
+					\WP_CLI::warning( __( 'Error processing customer', 'newspack' ) . ' ' . $email_address . ': ' . $user_id->get_error_message() );
```
claudiulodro, line 76:
```
Warning: Error processing customer [redacted]@mail.com: An account is already registered with that username. Please choose another.
```
There is a somewhat rare situation I ran into where if there are two people with the same name (e.g. two people named Joe Smith), this throws an error. I think a `get_user_by( 'login', $user_login )` check here could avoid that situation. I'm not sure the best fallback username though (maybe something email based or add some suffix to the username).

includes/reader-revenue/class-stripe-sync.php
```diff
@@ -0,0 +1,213 @@
...
+
+		if ( ! class_exists( 'WC_Stripe_Helper' ) || ! method_exists( 'WC_Stripe_Helper', 'get_order_by_charge_id' ) ) {
+			\WP_CLI::error( __( 'WC Stripe Gateway plugin has to be active.', 'newspack' ) );
+			return;
+		}
+
+		if ( ! function_exists( 'wc_create_new_customer' ) ) {
+			\WP_CLI::error( __( 'WooCommerce plugin has to be active.', 'newspack' ) );
+			return;
+		}
```
claudiulodro, line 171: I'd move these checks into the `$sync_to_wc` function, otherwise this will throw an error when using any other CLI commands  if WooCommerce/Stripe Gateway isn't active. And it would be impossible to run `wp plugin activate woocommerce` to fix the problem! :)

claudiulodro (changes requested):
This works great! Thanks for putting it together so quickly. Just a couple small things I noticed, but we can definitely use it as-is in the meantime if needed.

includes/reader-revenue/class-stripe-sync.php
adekbadek: Thanks for catching that! Amended in 2f32891ccd21d4c72c0396190409fadefb22db2c so that an existing username will get a unique-id suffix.

includes/reader-revenue/class-stripe-sync.php
adekbadek: Good catch! Fixed in 6914406b19b9b879cdf6ca01d7e9d3a1ca140343

claudiulodro (approved):
Latest version looks great! 👍

claudiulodro:
Merging this because I want to build off this class for the other CLI migrators

PR #1983, opened 2022-09-08
fix(wc-to-ac-metadata): disregard `0` next payment date

adekbadek:
### All Submissions:

When a WC Subscription is terminated, the next payment date is set to 0. We don't want to sync that – the next payment date should remain as it was in the event of cancellation.

1. Set up a site with WC Subscription, add and then terminate a subscription
2. On `master`, observe the `NP_Next Payment Date` metadata field is set to `0` in AC
3. Repeat on this branch, observe the `0` value is not sent

### Other information:


miguelpeixe (approved):
Confirmed that canceling the subscription does not update the `NP_Next Payment Date` field.

While testing this PR I noticed that manually canceling the subscription through Stripe's dashboard updates the `NP_Signup Page` from the original signup page (`/support-our-publication` in my case) to the `https://example.com/wp-json/newspack/v1/stripe/webhook`. Filed #1995 for this.

PR #1985, opened 2022-09-08
Adds a filter by Author in the Posts page

### LINKED ISSUE #1182: Filter backend posts by author
Many Newspack publishers work with large editorial staffs. The WordPress backend offers a means for filtering by authors (`/wp-admin/edit.php?author_name=kgagne`) or by post status (`/wp-admin/edit.php?post_status=draft&post_type=post`), but there's not an inbuilt mechanism for combining those search terms, such as all drafts by a particular author. The current solution is to manually combine the search parameters into a hand-crafted URL (`/wp-admin/edit.php?author_name=kgagne&post_status=draft&post_type=post`).

A dropdown that exists alongside core's existing the "Dates" and "Categories" dropdowns on the `edit.php` screen.

Dropdown:

Author selection:

[Author Filters](https://wordpress.org/plugins/author-filters/) (source of the above screenshots) is a compact plugin that suits this need perfectly. The only limitation is that it does not include Co-Authors Plus (CAP) guest authors in the dropdown. Though ideal, that feature that wouldn't be needed for an MVP.

**Additional context**
OSM has requested this functionality and is satisfied with the above plugin, but I think it would be a welcome feature for all publishers, since filtering by author alone is not as intuitive as it could be.

leogermani:
### All Submissions:

This PR adds a filter by Author to the posts page. See issue description for details.

It was inspired by the [Author FIlter](https://wordpress.org/plugins/author-filters/) plugin, but it was simplified and updated:

* No need for pre_get_posts filter
* replace usage of deprecated `who` param in favor of `capability`. Also, make it work with any custom post type
* i18n

Closes #1182 .

1. Test in a site with more than one author. Ideally with many authors
2. Go to wp-admin
3. Go to Posts
4. Confirm there's a new "Authors" dropdown in the Filters section
6. Test the filter, combined with others, and make sure it works as expected
7. Make sure it works well when you go to the "Mine" tab
8. If possible, test it with custom post types
9. NOTE: I'm concerned that this can degrade the speed of the page for sites with many (hundreds) of users... maybe we could add a limit for now before creating a proper solution?

Test with Co Authors Plus
- Install and activate Co Authors Plus
- Create one or more guest authors (Users -> Guest authors)
- Assign posts to guest authors
- Repeat the tests above and confirm the filter works
- Confirm that the filter does not show up when you select the "Mine" tab

Test disable flag
Add `define( 'NEWSPACK_DISABLE_AUTHORS_FILTER', true );` to your wp-config and make sure the filter is no longer initialized.

### Other information:


includes/author-filter/class-author-filter.php
```diff
@@ -0,0 +1,71 @@
...
+		$capability = $type_object->cap->edit_posts;
+
+		$args = array(
+			'show_option_all'  => __( 'All Users', 'newspack' ),
+			'orderby'          => 'display_name',
+			'order'            => 'ASC',
+			'name'             => 'author',
+			'capability'       => [ $capability ], // only users who can edit posts of this post type.
+			'include_selected' => true,
+			'exclude'          => [ get_current_user_id() ], // phpcs:ignore WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude
```
adekbadek, line 61: What's the reason behind exclusion of the current user? I think filtering for one's own posts is a valid use case.

includes/author-filter/class-author-filter.php
```diff
@@ -0,0 +1,71 @@
...
+		$type_object = get_post_type_object( $post_type );
+
+		if ( ! $type_object instanceof WP_Post_Type ) {
+			return;
+		}
+
+		$capability = $type_object->cap->edit_posts;
+
+		$args = array(
+			'show_option_all'  => __( 'All Users', 'newspack' ),
```
adekbadek, line 55: Nit: since this select displays only users who can edit posts, WDYT about changing this to "All Authors"?

adekbadek (commented):
Works as described! I've left a couple of question, and here is one more – what about Co-Authors Plus (CAP) support? It's noted in the issue, but there's no mention of potential planned support in the PR description.

adekbadek:
>  I'm concerned that this can degrade the speed of the page for sites with many (hundreds) of users

To clarify – is this about all users, or users who can edit posts?

includes/author-filter/class-author-filter.php
leogermani: Good call

leogermani:
> Works as described! I've left a couple of question, and here is one more – what about Co-Authors Plus (CAP) support? It's noted in the issue, but there's no mention of potential planned support in the PR description.

Oops, forgot about that. I'm not familiar with that tbh, I'll look into it.

leogermani:
I gave it a go with jQuery's `suggest` that is included in WP code and it looks like it won't serve our purposes... I'll insist to see if I have a POC on Monday, otherwise we have to decide if we want to buid a custom solution or not.

Maybe we can add this behind a feature flag and enable it in some sites, test it in the real world and see if it impacts the performance or not.

One good thing about having the whole list of users there is that you can browse them even without knowing the name of all authors.

leogermani:
I've added support to Co Authors Plus

What do you think of releasing it behind a feature flag and testing it out in a few sites before making it enabled by default?

includes/author-filter/class-author-filter.php
leogermani: I'm not removing the current user from the list anymore and it plays nicely with the "Mine" tab. When you click "Mine", your user will be selected in that filter, so it is consistent.

leogermani: Only when Co Authors Plus is active that it will not play nicely and I removed the whole filter when you click "Mine". Otherwise, the filter on the Dropdown would override the "Mine" selection, but "Mine" would still be seen in bold as it it was selected.

(That's because with CAP we change the filter to look at author_name instead of author (ID). But the author filter is still present in the query string, making "Mine" look like it is selected, while the author_name filter will override the author one)

leogermani:
I've tested in my local site with 500 users and it looks fine, but it can behave differently in real world scenarios, especially with Co Authors Plus.

So instead of adding it behind a feature flag, I added a check for a disable flag we can add in case this doesn't work well in some sites: https://github.com/Automattic/newspack-plugin/pull/1985/commits/465c7a1fd5d976e657eef4541e73887584b2757d

adekbadek (dismissed):
Looks good and works well! There are only some [linting issues](https://app.circleci.com/pipelines/github/Automattic/newspack-plugin/5082/workflows/08039f73-d65f-4e43-bff1-a413d0ea33e2/jobs/23779/parallel-runs/0/steps/0-102) to be resolved.

includes/author-filter/class-author-filter.php
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Adds a filter by author to the Posts page in the Admin dashboard.
+ *
+ * @package Newspack
+ */
+
+/**
+ * Author Filter class
+ */
+class Author_Filter {
```
miguelpeixe, line 11: I saw the PR earlier but forgot to bring it up. We should either put this class behind the `Newspack` namespace or a name prefix.

leogermani: oh, you are totally right, I missed that! will create a new PR

adekbadek: Good idea! ➕

PR #1994, opened 2022-09-13
Adds meta pixel settings section

### LINKED ISSUE #1554: Add field for FB Pixel ID
If a site is AMP Plus, show field for Facebook Pixel ID in Engagement > Social. Using this ID, place the following tag just before the closing head tag. (Note there are **two** locations for this ID to be inserted within the tag)

```
<!-- Meta Pixel Code -->
<script>
!function(f,b,e,v,n,t,s)
{if(f.fbq)return;n=f.fbq=function(){n.callMethod?
n.callMethod.apply(n,arguments):n.queue.push(arguments)};
if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
n.queue=[];t=b.createElement(e);t.async=!0;
t.src=v;s=b.getElementsByTagName(e)[0];
s.parentNode.insertBefore(t,s)}(window, document,'script',
'https://connect.facebook.net/en_US/fbevents.js');
fbq('init', 'PIXELID');
fbq('track', 'PageView');
</script>
<noscript><img height="1" width="1" style="display:none"
src="https://www.facebook.com/tr?id=PIXELID&ev=PageView&noscript=1"
/></noscript>
<!-- End Meta Pixel Code -->
```

Background: After internal discussion of the difficulties for publishers trying to place an FB pixel, it was determined that the best solution would be to provide an ID field and place the pixel on their behalf rather than recommending a third party plugin to do so. Find full discussion in P2 here: pamTN9-3ZQ-p2

leogermani:
### All Submissions:

Closes #1554 .

Relevant discussion: https://wp.me/pamTN9-3ZQ

Adds an option to the Engagement > Social settings page to allow users to configure Meta (Facebook) pixel.

1. Go to Newspack > Engagement > Social
2. Activate "Meta Pixel"
3. Add a pixel ID
4. Save
5. Test invalid formats
6. Test enabling and disabling the option
7. Check the copy and the link
8. Visit the site and confirm the pixel was added to the page. You can use a browser extension to help you with that if you want: https://href.li/?https://chrome.google.com/webstore/detail/facebook-pixel-helper/fdgfkebogiimcoedlicjlajpkdmockpc?hl=en
9. Visit the site with AMP enabled and confirm the pixel is also there
10. I've only tested this with random IDs. It would be awesome if someone could test it with a real pixel ID in a real site

### Other information:

### DIFF
--- a/includes/class-meta-pixel.php
+++ b/includes/class-meta-pixel.php
@@ -154,7 +154,7 @@ class Meta_Pixel {
 	 * @return string
 	 */
 	public static function get_img_snippet() {
-		return '<img height="1" width="1" style="display: none;" src="https://www.facebook.com/tr?id=$PIXEL_ID$&ev=PageView&noscript=1&cd%5Bpage_title%5D=Homepage&cd%5Bpost_type%5D=page&cd%5Bpost_id%5D=125&cd%5Bplugin%5D=PixelYourSite&cd%5Buser_role%5D=guest&cd%5Bevent_url%5D=leogermani.jurassic.tube%2F" alt="">';
+		return '<img height="1" width="1" style="display: none;" src="https://www.facebook.com/tr?id=__PIXEL_ID__&ev=PageView&noscript=1&cd%5Bpage_title%5D=Homepage&cd%5Bpost_type%5D=page&cd%5Bpost_id%5D=125&cd%5Bplugin%5D=PixelYourSite&cd%5Buser_role%5D=guest&cd%5Bevent_url%5D=leogermani.jurassic.tube%2F" alt="">';
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


adekbadek (commented):
Please add a PR description here – it will help to debug any related issues later, as well as serve as a description of the feature to any non-developers.

leogermani:
> Please add a PR description here – it will help to debug any related issues later, as well as serve as a description of the feature to any non-developers.

It's still in progress :)

adekbadek:
Ah, all right! If you submit a PR, our GH team gets a ping to review it. Until the PR is in progress, please submit it as a draft PR :)

leogermani:
> 

ok, noted. I thought you would only be notified if I added the Needs Review label

adekbadek:
That's another level of notifications – our team is marked as [code owners](https://github.com/Automattic/newspack-plugin/blob/b6014167449c252c982b123e0d6e42ad167e4fde/.github/CODEOWNERS#L0-L1), so we all get a GH ping when a PR is submitted. Then when the label is added, that pings Slack. This is all to ensure PRs are reviewed as quickly as possible.

includes/class-meta-pixel.php
```diff
@@ -0,0 +1,213 @@
...
+		return function_exists( 'is_amp_endpoint' ) && is_amp_endpoint();
+	}
+
+	/**
+	 * Gets the template for the img tag snippet
+	 *
+	 * @return string
+	 */
+	public static function get_img_snippet() {
+		return '<img height="1" width="1" style="display: none;" src="https://www.facebook.com/tr?id=__PIXEL_ID__&ev=PageView&noscript=1&cd%5Bpage_title%5D=Homepage&cd%5Bpost_type%5D=page&cd%5Bpost_id%5D=125&cd%5Bplugin%5D=PixelYourSite&cd%5Buser_role%5D=guest&cd%5Bevent_url%5D=leogermani.jurassic.tube%2F" alt="">';
```
adekbadek, line 157: The parameters of the URL in the `src` should be variables, or omitted. There are a few which are dynamic – as e.g. `event_url` here with test site URL value ;)

includes/wizards/class-engagement-wizard.php
```diff
@@ -126,6 +126,34 @@ public function register_api_endpoints() {
...
+					'args'                => [
+						'active'   => [
+							'type'              => 'boolean',
+							'required'          => true,
+							'validate_callback' => [ 'Newspack\Meta_Pixel', 'validate_active' ],
+						],
+						'pixel_id' => [
+							'type'              => [ 'integer', 'string' ],
+							'required'          => true,
+							'validate_callback' => [ 'Newspack\Meta_Pixel', 'validate_pixel_id' ],
```
adekbadek, line 151: When the option is not yet saved, an API request will be sent on clicking the toggle. This request will fail, since the `pixel_id` is invalid (it's `""` by default). This makes the toggle un-toggleable. 

image

adekbadek (changes requested):

includes/class-meta-pixel.php
leogermani: oh sorry I missed that.

The JS script version of the snippet does not include any of these variables. I'm gonna remove them from here for now so we can validate if the whole thing works and we can improve it later

leogermani: actually, decided to add the variables there

includes/class-meta-pixel.php
```diff
@@ -0,0 +1,228 @@
...
+	 * @return string
+	 */
+	public static function get_img_snippet() {
+		global $wp;
+		$current_user = wp_get_current_user();
+		$event_params = [
+			'page_title' => get_the_title(),
+			'user_role'  => empty( $current_user->roles ) ? 'guest' : $current_user->roles[0],
+			'event_url'  => home_url( $wp->request ),
+		];
```
miguelpeixe, line 163: Are these values standard? Looks like the parameters being used for the img snippet are not being applied to the script. Won't this cause inconsistent tracking?

It would be nice to have them filterable in an isolated private method like `get_snippet_parameters()` to be used in both img and script snippets.

includes/class-meta-pixel.php
```diff
@@ -0,0 +1,228 @@
...
+		s.parentNode.insertBefore(t,s)}(window, document,'script',
+		'https://connect.facebook.net/en_US/fbevents.js');
+		fbq('init', '__PIXEL_ID__');
+		fbq('track', 'PageView');
+		</script>
+		";
+	}
+
+	/**
+	 * Prints sinppets in the header
```
miguelpeixe, line 198:
```suggestion
	 * Prints snippets in the header
```

includes/class-meta-pixel.php
```diff
@@ -0,0 +1,228 @@
...
+	 *
+	 * @return void
+	 */
+	public static function print_footer_snippet() {
+		$pixel_id = self::get_pixel_id();
+		if ( empty( $pixel_id ) ) {
+			return;
+		}
+		// If AMP plugin is enabled, it will convert the image into a <amp-pixel> tag.
+		echo '<noscript>' . str_replace( '__PIXEL_ID__', intval( $pixel_id ), self::get_img_snippet() ) . '</noscript>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
miguelpeixe, line 224: AMP enabled does not mean the browser doesn't support javascript. A `! self::is_amp()` check is more suitable for this.

miguelpeixe (changes requested):

includes/class-meta-pixel.php
leogermani: This is not a check for AMP enabled.

The `noscript` tag is there for non-AMP site. If visiting a non AMP site without JS, the IMG tag will be used for tracking.

If AMP is enabled, the AMP plugin will capture this and replace the IMG with a `<amp-pixel>` tag. So this covers both cases and we don't need to manually add the amp tag. We leave it to the AMP plugin

includes/class-meta-pixel.php
leogermani: All examples I've seen of the JS snippet do not include these vars...

leogermani: Added the custom properties to the JS snippet following what I see here:

https://developers.facebook.com/docs/meta-pixel/implementation/conversion-tracking#custom-properties

Or we could also remove them from both places... The request in the [original issue](https://github.com/Automattic/newspack-plugin/issues/1554) did not ask for these properties.

I think this is something I brought looking at one of the plugins that add the pixel.

miguelpeixe: I believe it's good practice to implement a default set of parameters and make it filterable! This is working well 👍

includes/class-meta-pixel.php
```diff
@@ -0,0 +1,244 @@
...
+	private static function get_event_params() {
+		global $wp;
+		$current_user = wp_get_current_user();
+		$event_params = [
+			'page_title' => get_the_title(),
+			'user_role'  => empty( $current_user->roles ) ? 'guest' : $current_user->roles[0],
+			'event_url'  => home_url( $wp->request ),
+		];
+
+		if ( is_single() ) {
```
miguelpeixe, line 165:
```suggestion
		if ( is_singular() ) {
```

Small tweak include these parameters to `page` types

miguelpeixe (changes requested):

miguelpeixe (approved):
🙌

PR #1999, opened 2022-09-14
test: update tests dependencies

leogermani:
This PR updates our phpunit tests dependencies to allow us to run them against the latest version of WordPress

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
 		"automattic/vipwpcs": "^2.0.0",
 		"wp-coding-standards/wpcs": "*",
 		"dealerdirect/phpcodesniffer-composer-installer": "*",
-		"phpcompatibility/phpcompatibility-wp": "*"
+		"phpcompatibility/phpcompatibility-wp": "*",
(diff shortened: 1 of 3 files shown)


dkoo (approved):

adekbadek:
@leogermani – with these changes, the [WP version used in the CI Orb](https://github.com/Automattic/newspack-scripts/blob/d6b2a1fa4a0d9468a2fef582be1acc9ea4cef393/src/jobs/test-php.yml#L28) will be possible to update to `latest`, right?

leogermani:
> @leogermani – with these changes, the [WP version used in the CI Orb](https://github.com/Automattic/newspack-scripts/blob/d6b2a1fa4a0d9468a2fef582be1acc9ea4cef393/src/jobs/test-php.yml#L28) will be possible to update to latest, right?

Yes, I was going to publish a p2 about it today before submitting these PRs for review. Now I understand that you get notified even if I don't add the "Needs Review" label..

I'll create the P2 anyway with the next steps

PR #2004, opened 2022-09-15
feat: stop auto-email on registration and rate limit unverified accounts

miguelpeixe:
### All Submissions:

This PR changes the original behavior of reader registration, which sends an account verification as soon as a reader is registered.

The default phrasing for the reader registration block success message was changed to not look for a verification email:

image

Now the account verification email must be requested on the "My Account" page, until then the account will remain unverified. Mind that magic links and password resets also verify an account, so even if that session is lost, the next session will authenticate **and** verify the account.

This PR also introduces some rate limiting for unverified accounts. Password resets and account verifications must be sent within an interval of at least 10 seconds. Should it be more?

image

Registrations coming from the Sign In modal will be redirected to the "My Account" page to suggest account verification.

1. Check out this branch and create a new Reader Registration block on a page or campaign prompt (if you have a previous one it will have the success message to "check your email")
2. Check out this branch and register a new account through the Registration Block
3. Confirm you didn't receive any email
4. Click on "My Account" and click to send a verification link
5. Quickly click again and confirm the error
6. Do the same thing for password reset and confirm the error
7. Click on the last received email and confirm you're verified
8. On a fresh session, click on the Sign In modal
9. Click "I don't have an account"
10. Register a new email and confirm you're redirected to the My Account page

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -100,6 +100,7 @@ final class Reader_Activation {
 			'auth_intention_cookie' => self::AUTH_INTENTION_COOKIE,
 			'cid_cookie'            => NEWSPACK_CLIENT_ID_COOKIE_NAME,
 			'authenticated_email'   => $authenticated_email,
+			'account_url'           => function_exists( 'wc_get_account_endpoint_url' ) ? \wc_get_account_endpoint_url( 'dashboard' ) : '',
 		];
 
 		if ( Recaptcha::can_use_captcha() ) {
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1457,7 +1457,7 @@ final class Reader_Activation {
 	 */
 	public static function rate_limit_lost_password( $errors, $user_data ) {
 		if ( $user_data && self::is_reader_email_rate_limited( $user_data ) ) {
-			$errors->add( 'newspack_password_reset_interval', __( 'Please wait before requesting another password reset email.', 'newspack' ) );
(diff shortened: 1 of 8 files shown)


aschweigert:
> Password resets and account verifications must be sent within an interval of at least 10 seconds. Should it be more?

We could also try a different (but also common) pattern where we limit it to some number (say five), warn people when they're down to their last one and then lock it for an hour or so.

miguelpeixe:
I'm not sure I agree this is a better approach. We don't want any consecutive requests, especially if they have a higher chance to bounce (unverified accounts).

aschweigert:
Mm, yeah good point. Especially if the limit is still 10 😅

includes/reader-activation/class-reader-activation.php
```diff
@@ -1421,5 +1445,23 @@ public static function set_password_reset_mail_content_type() {
...
+	 *
+	 * @param \WP_Error      $errors    A WP_Error object containing any errors generated
+	 *                                  by using invalid credentials.
+	 * @param \WP_User|false $user_data WP_User object if found, false if the user does not exist.
+	 *
+	 * @return \WP_Error
+	 */
+	public static function rate_limit_lost_password( $errors, $user_data ) {
+		if ( $user_data && self::is_reader_email_rate_limited( $user_data ) ) {
+			$errors->add( 'newspack_password_reset_interval', __( 'Please wait before requesting another password reset email.', 'newspack' ) );
```
dkoo, line 1460: Nit: this could be a bit more specific to help readers understand how long they need to wait.

```suggestion
			$errors->add( 'newspack_password_reset_interval', __( 'Please wait a moment before requesting another password reset email.', 'newspack' ) );
```

dkoo (approved):
Preapproving since the only nitpick is copy-related.

PR #2005, opened 2022-09-16
feat(plugins-screen): display a plugin review form, if supplied

### LINKED ISSUE #532: Support prompt when adding a plugin
There is a lot to consider when using plugins alongside Newspack. Compatibility and performance concerns are probably significant but there are many others, all considered by our research methodology. This is why we encourage publishers to come to us with feature requests first. However, that message doesn't always permeate at the time a publisher decides to try out a new plugin, or reinstall something used previously. This can cause issues with breaking features, performance degradation and in the worse case scenario, downtime.

Each time a publisher seeks to use the Add New Plugin functionality in WordPress they are reminded of the suggestion to consult the Newspack team first, via some sort of interrupting modal. They are given the opportunity to get in touch via a ticket to ask about the best way to implement the feature they're looking for. The friction of interrupting the installation process to show this message will help to ensure the publisher is fully aware of the options to lean on Newspack's experience and expertise.

An admin notice might provide the same reminder, but as there can be many of these they can be overlooked. It also doesn't introduce friction into the plugin installation process, which is important for encouraging publishers to get in touch.

adekbadek:
### All Submissions:

Adds a modal prompting the user to fill out a plugin review form before installing a new plugin.

image

Closes #532

1. Load the WP admin plugin install screen (`/wp-admin/plugin-install.php`) and observe nothing new
2. Define the `NEWSPACK_PLUGIN_REVIEW_FORM_URL` constant (e.g. `wp config set NEWSPACK_PLUGIN_REVIEW_FORM_URL 'https://forms.gle/ofFmGKqoaPWJm2bm8'`)
3. Reload the plugin install screen and observe the modal
4. Click the link in the modal, observe it links to the site under `NEWSPACK_PLUGIN_REVIEW_FORM_URL`
5. Click the "Close" button and observe the modal is closed

### Other information:

### DIFF
--- a/assets/plugins-screen/plugins-screen.scss
+++ b/assets/plugins-screen/plugins-screen.scss
@@ -26,29 +26,47 @@ table.plugins {
 	&--hidden {
 		display: none;
 	}
-	h1 {
-		margin-top: 0;
+	h1,
+	p,
+	> div > div {
+		padding-left: 32px;
+		padding-right: 32px;
 	}
-	button {
-		padding: 8px 12px;
-		border-radius: 3px;
-		border: none;
-		color: wp-colors.$white;
-		background-color: colors.$primary-500;
+	h1 {
+		margin-top: 21px;
+		font-size: 16px;
+		padding-bottom: 20px;
+		border-bottom: 1px solid #ddd;
 	}
 	> div {
+		position: relative;
 		display: flex;
+		padding-bottom: 32px;
 		flex-direction: column;
-		min-width: 30vw;
-		min-height: 20vw;
-		padding: 46px 50px;
 		background-color: wp-colors.$white;
-		> div:first-child {
+		box-shadow: 0 10px 10px rgb( 0 0 0 / 25% );
+		> button {
+			position: absolute;
+			top: 14px;
+			right: 27px;
+			padding: 0.5em;
+			background: none;
+			border: none;
+		}
+		> p {
 			flex: 1;
 		}
 		> div:last-child {
 			display: flex;
 			justify-content: flex-end;
+			padding-top: 20px;
+			button {
+				padding: 11px 19px;
+				border-radius: 2px;
+				border: none;
+				color: wp-colors.$white;
+				background-color: colors.$primary-500;
+			}
 		}
 	}
 }
--- a/assets/plugins-screen/plugins-screen.scss
+++ b/assets/plugins-screen/plugins-screen.scss
@@ -67,7 +67,7 @@ table.plugins {
 				color: wp-colors.$white;
 				background-color: colors.$primary-500;
 				text-decoration: none;
-				
+
 				&:hover,
 				&:active {
 					background: colors.$primary-600;
--- a/assets/plugins-screen/plugins-screen.scss
+++ b/assets/plugins-screen/plugins-screen.scss
@@ -1,6 +1,5 @@
-/**
- * Styles for the WP Admin Plugins screen.
- */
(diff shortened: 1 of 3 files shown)


laurelfulford:
@adekbadek This might be getting into the weeds a bit, so feel free to disregard: 

I know we've been moving to more standard Gutenberg components for the Newspack screens. Would following that and using the default WP Admin components here make sense? 

WP Admin mostly uses [Thickbox](https://codex.wordpress.org/Javascript_Reference/ThickBox) for popups that aren't totally custom -- that plus the default button classes would look something like:

image

The downside of course here is that it wouldn't be as clearly distinct and Newspack-UI-y as rolling our own. It's also not normally set up to fire on page load -- I faked out the above by adding this to a random WP admin page, but it's triggered on click:

```
<?php add_thickbox(); ?>

<a href="#TB_inline?&width=400&height=135&inlineId=my-content-id" class="thickbox">Open Popup</a>

<div id="my-content-id" style="display:none;">
   <p>Please submit plugins for review by the Newspack Team before installing them on your website.</p>
   <p><a href="#" class="button button-primary">Plugin Review Form</a></p>
</div>
```

Otherwise, I think my feedback would just be to try to get the scale closer to the popups in the Newspack wizard -- this is a straight edit of one to match the message, there's probably some kind of balance between nailing the details and not adding too much duplicate CSS. I'm happy to help with this if it seems like the way to go

image

I'm sure @thomasguillot has some good ideas, and more guidance around how the Newspack wizard elements should be used, if at all in this case! 

(Also, this popups works great for me!)

adekbadek:
>  [Thickbox] is also not normally set up to fire on page load

And while I'm sure it can be hacked to be triggered so, I don't think it's worth the effort here. 

b4bff6b3d9e173ee783645bd64a91dd77aea5f34 makes it look like Newspack (Gutenberg) modal, and updates the copy.

assets/plugins-screen/plugins-screen.scss
```diff
@@ -12,3 +11,63 @@ table.plugins {
...
+			display: flex;
+			justify-content: flex-end;
+			padding-top: 20px;
+			a {
+				padding: 11px 19px;
+				border-radius: 2px;
+				border: none;
+				color: wp-colors.$white;
+				background-color: colors.$primary-500;
+				text-decoration: none;
```
laurelfulford, line 69:
```suggestion
				text-decoration: none;
				
				&:hover,
				&:active {
					background: colors.$primary-600;
				}

				&:focus {
					box-shadow: inset 0 0 0 1px #fff, 0 0 0 2px colors.$primary-500;
				}
```

laurelfulford (approved):
> And while I'm sure it can be hacked to be triggered so, I don't think it's worth the effort here.

That's a great point 😄 

This looks bang on to me! A couple small things:

* I added a code suggestion to add hover and focus styles that match the other buttons - definitely not merge-critical. 
* I'm double-guessing the language I used a bit in the original mockup -- the heading is going to be a bit redundant either way, but I'm wondering if "Plugin review required" would make more sense/be more effective than just "Plugin review" -- what do you think?

I don't think either of these are blockers at all, so giving this a 👍

adekbadek:
> I'm wondering if "Plugin review required" would make more sense/be more effective than just "Plugin review" -- what do you think?

Good idea! "required" makes it a way stronger suggestion. Done in 3632ad4f6ffb6bd250bac64d98beaf0c8dcf51f8

PR #2021, opened 2022-09-19
Remove Stripe subs. on WP account deletion

adekbadek:
### All Submissions:

Adds Stripe and newsletter subscriptions cancellation on account deletion.

1. Set up a site with Stripe as the Reader Revenue platform and RAS enabled
3. Donate and subscribe to a newsletter as a new reader  
4. Verify in Stripe and ESP that the reader has a Stripe subscription and is subscribed to the newsletter
5. Complete the account deletion flow, starting on My Account page
6. After the account is deleted, verify that the Stripe subscription has been cancelled and the reader unsubscribed from the newsletters

### Other information:

### DIFF
--- a/includes/reader-revenue/class-stripe-connection.php
+++ b/includes/reader-revenue/class-stripe-connection.php
@@ -717,18 +717,22 @@ class Stripe_Connection {
 				}
 
 				if ( Reader_Activation::is_enabled() && method_exists( '\Newspack_Newsletters_Subscription', 'add_contact' ) ) {
-					$sub_end_date = gmdate( Newspack_Newsletters::METADATA_DATE_FORMAT, $payload['ended_at'] );
-					$contact      = [
-						'email'    => $customer['email'],
-						'metadata' => [
-							Newspack_Newsletters::$metadata_keys['sub_end_date']   => $sub_end_date,
-						],
-					];
-					if ( 0 === $active_subs && in_array( $payload['plan']['interval'], [ 'month', 'year' ] ) ) {
-						$membership_status = 'Ex-' . self::get_membership_status_field_value( $payload['plan']['interval'] );
-						$contact['metadata'][ Newspack_Newsletters::$metadata_keys['membership_status'] ] = $membership_status;
+					$contact_exists = ! \is_wp_error( \Newspack_Newsletters_Subscription::get_contact_data( $customer['email'] ) );
+					// Only handle subscription deletion of an existing contact.
+					if ( ! $contact_exists ) {
+						$sub_end_date = gmdate( Newspack_Newsletters::METADATA_DATE_FORMAT, $payload['ended_at'] );
+						$contact      = [
+							'email'    => $customer['email'],
+							'metadata' => [
+								Newspack_Newsletters::$metadata_keys['sub_end_date']   => $sub_end_date,
+							],
+						];
+						if ( 0 === $active_subs && in_array( $payload['plan']['interval'], [ 'month', 'year' ] ) ) {
+							$membership_status = 'Ex-' . self::get_membership_status_field_value( $payload['plan']['interval'] );
+							$contact['metadata'][ Newspack_Newsletters::$metadata_keys['membership_status'] ] = $membership_status;
+						}
+						\Newspack_Newsletters_Subscription::add_contact( $contact );
 					}
-					\Newspack_Newsletters_Subscription::add_contact( $contact );
 				}
 
 				// Update data in Campaigns plugin.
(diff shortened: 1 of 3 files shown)


includes/reader-revenue/class-stripe-connection.php
```diff
@@ -1267,6 +1303,24 @@ public static function create_wc_transaction_payload( $customer, $payment ) {
 			'subscribed'         => self::has_customer_opted_in_to_newsletters( $customer ),
 		];
 	}
+
+	/**
+	 * If a WP user is deleted, cancel their Stripe subscriptions.
+	 *
+	 * @param int $user_id User ID.
+	 */
+	public static function cancel_deleted_user_subscriptions( $user_id ) {
```
miguelpeixe, line 1312: The method itself does not take "user deletion" into account so it shouldn't state that.

miguelpeixe (changes requested):
I made a new recurring donation with my email to the Stripe test account, deleted the account, and noticed in the logs the canceling of a subscription (`sub_1LhgC9G3IJS4cjk80RnxUQKl`) that was not the one that had I just created (`sub_1LkXzsG3IJS4cjk8wreYDB4q`)

Turns out there are 2 stripe clients with the same email, which shouldn't happen and is a concern in itself that we should monitor why that would happen. My concern regarding this PR is that the donation was attached to a client and the subscription cancelation happened on another client.

includes/reader-revenue/class-stripe-connection.php
adekbadek: Fixed in f3c3c6cbd72fed407e81ad6d567e5c0acb3862e9

adekbadek:
> I made a new recurring donation with my email to the Stripe test account, deleted the account, […] the donation was attached to a client and the subscription cancelation happened on another client.

Hm, that should not have happened. The stripe customer ID is tied to a WP used ID either on RAS' [reader verification](https://github.com/Automattic/newspack-plugin/blob/cc26a6d4f41d5cacc777f99d4cafb1ea633a71ca/includes/reader-revenue/class-stripe-connection.php#L1239) or by [matching from user's WC orders](https://github.com/Automattic/newspack-plugin/blob/e1533b6a6d57deba98e986628fc86c991897f742/includes/reader-revenue/my-account/class-woocommerce-my-account.php#L366). I'm guessing the former happened in this case. Can you try to list reproduction steps for this issue to occur? How was the client with the `sub_1LhgC9G3IJS4cjk80RnxUQKl` subscription created?

> Canceling the subscription also triggers an action that recreates the deleted ESP contact with the "Ex-Monthly Donor" meta field, which shouldn't happen.

Should be fixed by b7da828ade68f04b79569b1c3d5648b647916742

miguelpeixe (changes requested):
> Can you try to list reproduction steps for this issue to occur? How was the client with the sub_1LhgC9G3IJS4cjk80RnxUQKl subscription created?

I wasn't aware I had 2 clients with the same email until I tested this PR, so I don't know the steps to reproduce this scenario.

What could be done to further test it would be manually registering 2 clients with the same email under Stripe and see how the subscription creation/cancelation is handled.

This is no longer blocking since I cleaned up the Stripe accounts and was not able to reproduce with a fresh subscription. I'm considering this to not work properly in any circumstance since email is used as the identifier throughout the codebase.

Before approving, I wonder if data protection laws compliance requires us to also delete the customer record. WDYT?

yogeshbeniwal:
> I wonder if data protection laws compliance requires us to also delete the customer record

Keeping financial transaction detail should be ok.

adekbadek:
(some internal discussion in p1663921713036149-slack-C0390J7K0KY)

includes/reader-revenue/class-stripe-connection.php
```diff
@@ -1267,6 +1311,24 @@ public static function create_wc_transaction_payload( $customer, $payment ) {
 			'subscribed'         => self::has_customer_opted_in_to_newsletters( $customer ),
 		];
 	}
+
+	/**
+	 * If a WP user is deleted, cancel their Stripe subscriptions.
```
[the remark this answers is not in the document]
adekbadek: Done in b8a1705dc3166dc10d1f968b799d9eadc2eb52e7

adekbadek:
> Before approving, I wonder if data protection laws compliance requires us to also delete the customer record. WDYT?

Internally, we've agreed that this should not block the subscriptions termination on account deletion. Removing PI data from Stripe can be added later. 

This branch was rebased on `master`, can you please re-review @miguelpeixe?

miguelpeixe (changes requested):
Canceling the subscription is working well 👌 

The latest layout changes to "My Account" made the Delete Account button look weird. I suggest a change to something like this:

| Current | Suggestion |
| --- | --- |
| image | image |

adekbadek:
Good idea, done in 8769b380dca829bc75d496add97b0b285e7eda82

miguelpeixe (approved):
Thank you for revising!

PR #2023, opened 2022-09-19
Update tests to php8

leogermani:
This upgrades the tests so they can run both on php 7.4 and 8.

To test it, load the repo in a php 8 environment and run `composer install`. the run `./vendor/bin/phpunit`

Do the same thing on a php 7.4 environment, but run composer `update` instead of `install`. (Discard the changes to composer.lock afterwards)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  newspack: adekbadek/newspack@dev:wp-version-tests-2
+  newspack: adekbadek/newspack@1.4.0
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  newspack: adekbadek/newspack@dev:wp-version-tests
+  newspack: adekbadek/newspack@dev:wp-version-tests-2
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  newspack: adekbadek/newspack@1.3.0
+  newspack: adekbadek/newspack@dev:wp-version-tests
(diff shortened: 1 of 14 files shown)


leogermani:
The CI failing here is expected, as we need to update them to either use php8 or run composer update

adekbadek (commented):
The tests run successfully in a PHP 8 environment, but I had to run `composer update`. `install` results in the following error:

```
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.
```

Also, when running tests, PHPUnit warns about the need to upgrade the config:

```
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!
```

It is resolved by running `./vendor/bin/phpunit --migrate-configuration`, should the updated `phpunit.xml` be commited here?

leogermani:
hmm the version dance can get quite confusing. I see the message now. Yes, we should fix it and commit. But let me confirm if it will still work with phpunit 7

adekbadek:
>  >  Also, when running tests, PHPUnit warns about the need to upgrade the config:

> I don't see this. How did you run the tests?

You can see it [on CI job here](https://app.circleci.com/pipelines/github/Automattic/newspack-plugin/5192/workflows/c1c00fc5-4de6-4c15-8107-3d2f888e3a54/jobs/24327?invite=true#step-104-214).

leogermani:
I thought we were using phpunit 7 when running php 7.4, but we are using phpunit 9.5 in all cases... At some point, I had phpunit 7 here running and that's why I was not seeing.

Since we unblocked tests with newer WP versions, we can actually drop phpunit 7 from our dependecies list.

adekbadek:
With the latest changes, `composer install` works smoothly with PHP 8, as `composer update` does with PHP 7.4, locally 👌

adekbadek (approved):

PR #2025, opened 2022-09-20
feat: more prominent already-have-account section

adekbadek:
### All Submissions:

Building on the changes from #1967 and feedback, this PR makes the "already have an account?" section more prominent.

On the auth modal:

image

Registration block, in a prompt:

image

Registration block, on a page:

image

1.
2.
3.

### Other information:


assets/reader-activation/auth.scss
```diff
@@ -196,6 +194,21 @@
...
+			width: 100%;
+			top: calc( -1em - 23px );
+			background-color: var( --newspack-cta-color );
+			padding: 4px 15px;
+			margin-bottom: 11px;
+			padding-top: 6px !important;
+			padding-bottom: 8px !important;
+			&,
+			a {
+				color: var( --newspack-cta-contrast-color );
```
dkoo, line 209: This element still has its hover color hard-coded to `#111`, which doesn't look great with the default red background:

Screen Shot 2022-09-21 at 2 21 16 PM

Recommend a light grey or even just keeping it white.

assets/blocks/reader-registration/edit.js
```diff
@@ -203,31 +203,29 @@ export default function ReaderRegistrationEdit( {
 				{ editedState === 'initial' && (
 					<div className={ `newspack-registration ${ className }` }>
 						<form onSubmit={ ev => ev.preventDefault() }>
+							<div className="newspack-registration__have-account">
```
dkoo, line 206: At smaller viewports, the close button overlaps the message, which looks unintentional:

Screen Shot 2022-09-21 at 2 26 02 PM

I'd expect the message to appear full-bleed on top of the modal, similar to how it looks in the auth modal:

Screen Shot 2022-09-21 at 2 27 38 PM

dkoo (changes requested):
I'll let @thomasguillot do a fuller design review, but there are a couple of issues I think should be addressed:

thomasguillot:
Here are some potential suggestions.

If we keep the banner red, it will conflict with the CTA so I suggest we switch to a very dark grey `#111` with white text.

The close button is a tricky one, especially on mobile. We can go 3 different ways. The 2nd one I like the least but it might be the easiest to implement.

adekbadek:
With changes in b675cecf6d71696f371af7165b938f82d0596717, here's the current look:

1. Auth modal, large screen:

image

2. Auth modal, small screen (here the close button can look nicer):

image

3. Registration block in an overlay prompt, large screen:

image

4. Registration block in an overlay prompt, small screen:

image

assets/reader-activation/auth.scss
adekbadek: Fixed in b675cecf6d71696f371af7165b938f82d0596717

assets/blocks/reader-registration/edit.js
adekbadek: Fixed in b675cecf6d71696f371af7165b938f82d0596717

dkoo (approved):
No more issues from my end—if it looks good to @thomasguillot then 👍

thomasguillot:
👍

PR #2026, opened 2022-09-20
fix(stripe): validate webhook existence on donation

adekbadek:
### All Submissions:

The Stripe integration relies on webhooks, which are currently validated on loading the Reader Revenue wizard. It might happen though that the integration was set up way back when the webhook creation was manual, and since then the user did not visit the RR wizard. This PR ensures that the webhook is partially* validated on every donation. 

\* only checked if the webhook-identifying WP option was created, we don't want to hit the Stripe API excessively

1. Set up a site with Stripe as the Reader Revenue platform
2. Remove the webhook-bearing option (`wp option delete newspack_stripe_webhook`)
3. Make a donation, observe the webhook has been recreated 

### Other information:

### DIFF
--- a/includes/reader-revenue/class-stripe-connection.php
+++ b/includes/reader-revenue/class-stripe-connection.php
@@ -695,17 +695,28 @@ class Stripe_Connection {
 
 	/**
 	 * Create Stripe webhooks if they are missing. Otherwise, validate the webhhooks.
+	 *
+	 * @param bool $validate_existence_only If true, only validate the existence of the webhooks.
 	 */
-	public static function validate_or_create_webhooks() {
+	public static function validate_or_create_webhooks( $validate_existence_only = false ) {
 		$is_valid        = true;
-		$webhook_events  = [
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/class-stripe-connection.php
```diff
@@ -695,17 +695,28 @@ private static function reset_webhooks() {
...
-		$webhook_events  = [
+		$created_webhook = get_option( self::STRIPE_WEBHOOK_OPTION_NAME );
+
+		if ( $validate_existence_only ) {
+			if ( false === $created_webhook ) {
+				// If the webhook does not exist, do the full validation & creation.
+				self::validate_or_create_webhooks();
+			}
+			return;
+		}
```
miguelpeixe, line 711:
```suggestion
		if ( true === $validate_existence_only ) {
			if ( false === $created_webhook ) {
				// If the webhook does not exist, do the full validation & creation.
				self::validate_or_create_webhooks( false );
			}
			return;
		}
```

NIT: to avoid confusion, it would be best to increase verbosity here

miguelpeixe (approved):

PR #2031, opened 2022-09-22
Experimental Stripe Connect test

claudiulodro:
**This requires https://github.com/Automattic/newspack-blocks/pull/1278 to work.**

This PR adds a minimum possible Stripe Connect integration to the Simplified Donate Block. This was primarily so that we have a simple way to create purchases under a Stripe Connected account, for migration testing, but it also identifies most of the areas we'll need to update when we do add proper Stripe Connect support. This PR is definitely not intended to ever be merged.

## To use
1. Set up Stripe Connect on a Stripe account, so that it can be a "parent" account to other Stripe accounts. 
2. Connect a "child" Stripe account to the "parent" account (I have one already connected, ping me in Slack if you want access)
3. At the Connect screen of the parent account, find the child account's account ID. This should start with `acct_`.
4. In your wp-config add the following constant: `define( 'NEWSPACK_STRIPE_CONNECTED_CHILD_ACCOUNT', 'acct_...' );`
5. In the Newspack Reader Revenue Stripe settings, use the API keys for the parent account, not the child account.
6. Use the Simplified Donate Block. It should work nicely, and in the child account's payments/subscriptions you should see a 5% application fee added to the purchases:
Screen Shot 2022-09-22 at 11 47 57 AM

### DIFF
--- a/includes/reader-revenue/class-stripe-connection.php
+++ b/includes/reader-revenue/class-stripe-connection.php
@@ -154,7 +154,7 @@ class Stripe_Connection {
 	private static function get_customer_by_id( $customer_id ) {
 		$stripe = self::get_stripe_client();
 		try {
-			return $stripe->customers->retrieve( $customer_id, [] );
+			return $stripe->customers->retrieve( $customer_id, [], [ 'stripe_account' => NEWSPACK_STRIPE_CONNECTED_CHILD_ACCOUNT ] );
 		} catch ( \Throwable $e ) {
 			return new \WP_Error( 'stripe_newspack', __( 'Could not fetch customer.', 'newspack' ), $e->getMessage() );
 		}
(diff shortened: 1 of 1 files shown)


yogeshbeniwal:
@claudiulodro Interesting, May I know what use case you are targeting by adding integration to Stripe Connect.

adekbadek:
Housekeeping as part of the monorepo cutover ([NPPM-2752](https://linear.app/a8c/issue/NPPM-2752)): this repository is now legacy and active development has moved to the [newspack-workspace monorepo](https://github.com/Automattic/newspack-workspace). Closing this stale draft (no activity in 6+ months) to keep the legacy backlog clean.

This isn't a rejection of the work. If the change is still relevant, please re-open it as a PR against the monorepo under the matching `plugins/` directory — code landed directly in this repo won't ship, since the monorepo is now the canonical source.

PR #2033, opened 2022-09-22
Add the Significant Revisions feature

### LINKED ISSUE #1732: As a publisher I want to be able to mark a Revision as "significant"
...and have this revision not automatically purged if we limit the number of revisions for performance reasons (see #1731)
...additionally the date/time for the last significant revision would be used as the "last updated date" on the post.

Discussion and further context: pamTN9-4Fd-p2#comment-6097

leogermani:
### All Submissions:

Closes #1732 . (But not covering the "additionally the date/time for the last significant revision would be used as the "last updated date" on the post " request)

See the test instructions on https://github.com/Automattic/newspack-plugin/pull/2030 for set up and base test instructions

* In the Revisions screens, Mark some revisions as significant
* Run the tests as the instructions above and make sure the significant revisions are never deleted
* Open the "Help" section on the top of the page and see the "Nespack" section
* read it and confirm it's clear and that it has no typos
* get the ID of one of the revisions marked as significant from the address bar
* Delete this revision directly from the database `DELETE FROM wp_posts WHERE ID = %d`
* Go back to the revisions page and confirm you see a Warning telling you about the deleted revision

Captura de tela de 2022-09-23 17-18-41

* Click to restore the backup and make sure the revision is restored

### Other information:

### DIFF
--- a/includes/revisions-control/class-significant-revision.php
+++ b/includes/revisions-control/class-significant-revision.php
@@ -60,7 +60,11 @@ class Significant_Revision {
 	 * @return array
 	 */
 	public function get_post_revisions() {
-		return get_post_meta( $this->post_id, self::RELEVANT_IDS_META_KEY );
+		$ids = get_post_meta( $this->post_id, self::RELEVANT_IDS_META_KEY );
+		if ( empty( $ids ) ) {
+			return [];
+		}
+		return $ids;
 	}
 
 	/**
--- a/includes/revisions-control/class-significant-revision.php
+++ b/includes/revisions-control/class-significant-revision.php
@@ -19,7 +19,7 @@ class Significant_Revision {
 	 *
 	 * @var string
 	 */
-	const RELEVANT_IDS_META_KEY = '_significant_revision';
+	const SIGNIFICANT_IDS_META_KEY = '_significant_revision';
 
 	/**
 	 * The meta key used to store the information of the ID of the revision a backup is of
@@ -60,7 +60,7 @@ class Significant_Revision {
 	 * @return array
 	 */
 	public function get_post_revisions() {
-		$ids = get_post_meta( $this->post_id, self::RELEVANT_IDS_META_KEY );
+		$ids = get_post_meta( $this->post_id, self::SIGNIFICANT_IDS_META_KEY );
 		if ( empty( $ids ) ) {
 			return [];
 		}
@@ -84,7 +84,7 @@ class Significant_Revision {
 	 */
 	public function mark_as_significant() {
 		if ( ! $this->is_significant() ) {
-			add_post_meta( $this->post_id, self::RELEVANT_IDS_META_KEY, $this->ID );
+			add_post_meta( $this->post_id, self::SIGNIFICANT_IDS_META_KEY, $this->ID );
 			$this->create_backup();
 		}
 	}
@@ -95,12 +95,12 @@ class Significant_Revision {
 	 * @return void
 	 */
 	public function unmark_as_significant() {
-		delete_post_meta( $this->post_id, self::RELEVANT_IDS_META_KEY, $this->ID );
+		delete_post_meta( $this->post_id, self::SIGNIFICANT_IDS_META_KEY, $this->ID );
 		$this->delete_backup();
 	}
 
 	/**
-	 * Toggles the revision relevance
+	 * Toggles the revision significance
 	 *
 	 * @return bool The resulting state
 	 */
--- a/includes/revisions-control/class-significant-revision.php
+++ b/includes/revisions-control/class-significant-revision.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * A class to handle relevant revisions checks and actions
+ * A class to handle significant revisions checks and actions
  *
  * @package Newspack
  */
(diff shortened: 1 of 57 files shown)


includes/revisions-control/class-significant-revision.php
```diff
@@ -0,0 +1,204 @@
...
+ * Significant Revision Object
+ */
+class Significant_Revision {
+
+	/**
+	 * The name of the post meta that stores the significant revisions IDs
+	 *
+	 * @var string
+	 */
+	const RELEVANT_IDS_META_KEY = '_significant_revision';
```
miguelpeixe, line 22: "Relevant" and "significant" are being used here for the same thing, but IMHO a more appropriate term for the whole thing is "kept". "Kept revisions" is a more direct term for what it does.

includes/revisions-control/class-significant-revisions.php
```diff
@@ -0,0 +1,260 @@
...
+	 */
+	public static function init() {
+		if ( ! Revisions_Control::is_active() ) {
+			return;
+		}
+		if ( ! self::$initiated ) {
+			add_action( 'load-revision.php', array( __CLASS__, 'admin_init' ) );
+			add_action( 'wp_ajax_newspack_toggle_revision_significant', array( __CLASS__, 'ajax_toggle_significant' ) );
+			add_filter( 'wp_prepare_revision_for_js', array( __CLASS__, 'filter_revision_for_js' ), 10, 3 );
+			add_action( 'init', array( __CLASS__, 'register_post_type' ) );
```
miguelpeixe, line 53: Really like the idea of a post type for backing up protected revisions 💯

miguelpeixe (changes requested):
This is working great!

A non-blocking suggestion, the `.revision-tickmarks` could be improved to more easily identify which revisions in the timeline have been marked as significant:

image

---

Marking the oldest revisions as significant results in the amount of kept revisions going over the determined limit of 30, is this expected?

After clicking to restore a deleted significant revision I'd expect to be redirected to its position (`/revision.php?revision={id}`). It took me some time to find it in the timeline.

leogermani:
@miguelpeixe 

Thanks for the review

> A non-blocking suggestion, the .revision-tickmarks could be improved to more easily identify which revisions in the timeline have been marked as significant:

Not trivial.

> Marking the oldest revisions as significant results in the amount of kept revisions going over the determined limit of 30, is this expected?

Yes

> After clicking to restore a deleted significant revision I'd expect to be redirected to its position (/revision.php?revision={id}). It took me some time to find it in the timeline.

There could be many deleted significant revisions, so there's not only one place to go... But we probably could go to one of them

includes/revisions-control/class-significant-revision.php
leogermani: I'd say `protected` or `locked` are even better. "Significant" is what has been proposed and used in the front end so that's what I used... should we ask around to see what's the best name for it?

miguelpeixe:
> Things get tricky here. I'm overwriting methods from some Backbone JS classes and I'm very limited.

I was thinking of going the other way around. Using the "significant revisions" API to access the DOM (`.revision-tickmarks > *:nth-child()`) and append an additional absolute positioned element to the ticks that match the index.

leogermani:
> Using the "significant revisions" API to access the DOM

in the current JS we don't have access to a list of all revisions and a flag saying whether they are significant or not. We only have access to the selected revision. we need to hook somewhere else, like the other model I used.

I think I found a good workaround, by adding an ID to each tickmark element so we can easily manipulate them later.

Check https://github.com/Automattic/newspack-plugin/pull/2033/commits/10f8aa172665a13d932636c7447383e8ef9d358a

btw. Any suggestion on the style? I dropped the idea of making the border 2px because I'm afraid this could break the layout when there are too many revisions

laurelfulford:
> Any suggestion on the style? I dropped the idea of making the border 2px because I'm afraid this could break the layout when there are too many revisions

I'm not sure how easy this would be, but the text information for each revision might be another spot to highlight the major revisions, without having to rely as much on the little space you get in the timeline bar (that thing can get very crowded!!). That way, they'd be noticeable as you dragged the little slider along, too -- this is clunky, but something like:

image

(I'm soooo looking forward to this feature!! 😄)

miguelpeixe:
Really like this suggestion by @laurelfulford, but I still miss a better reference on where the major revisions are positioned in the timeline. A small dot positioned above the ticks would help that:

```css
.revisions-tickmarks > div.newspack_major:after {
  content: "";
  width: 6px;
  height: 6px;
  background: #f00;
  top: 0;
  left: 0;
  display: block;
  border-radius: 100%;
  margin-left: -3px;
  margin-top: -10px;
}
```

leogermani:
@laurelfulford I was able to add the label to the revision title, but not to the tooltip. I won't say it's impossible, but this is getting way too complicated. Let's go without this for now.

@miguelpeixe the dots are looking good:

Captura de tela de 2022-10-05 15-27-52

Here is how it looks with 100 revisions:

Captura de tela de 2022-10-05 15-29-12

And here with 500:
Captura de tela de 2022-10-05 15-30-25

I was not happy with the colors, so I managed to get the colors from the selected admin color scheme. It's not perfect, as it will only work with the core color schemes, and not with custom added color schemes, but I think this is ok for now:

Captura de tela de 2022-10-05 16-35-59

Captura de tela de 2022-10-05 16-37-19

Captura de tela de 2022-10-05 17-02-16

Captura de tela de 2022-10-05 17-02-45

Captura de tela de 2022-10-05 17-03-12

miguelpeixe (approved):
🙌 

This is looking great! Thank you for the revisions @leogermani 🙇 

Only one small typo and we're good!

PR #2043, opened 2022-09-27
feat(reader-auth): improved otp ux

miguelpeixe:
### All Submissions:

This PR implements enhancements to the OTP input UX.

image

Each digit for the code is treated as a single input. Once a valid number is entered, the next input is focused. It also supports navigating through arrow keys and pasting from the clipboard from any position of the OTP input.

1. Check out this branch and with reader activation enabled, send an authorization code to a reader account
2. Confirm you see the new UI for entering the code
3. Attempt to enter a non-numeric value into any digit
4. Navigate between the OTP code using arrows and confirm it works as expected
5. Navigate between the OTP code using tab and shift+tab and confirm it works as expected
6. Enter numeric values and confirm the values are applied smoothly
7. Use backspace to clear the values
8. Copy the valid OTP code received, paste it into any digit, and confirm it's applied correctly
9. Sign in and confirm the pasted value is sent
10. Log out, send another authorization code
11. Attempt to copy and paste a random text and confirm it doesn't apply
12. Manually enter the valid code, submit, and confirm you're signed in

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -438,6 +438,8 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 							if ( prev ) {
 								prev.focus();
 							}
+							values[ i ] = '';
+							otpCodeInput.value = values.join( '' );
 							break;
 						case 'ArrowLeft':
 							ev.preventDefault();
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -414,6 +414,7 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 				return;
 			}
 			const inputContainer = originalInput.parentNode;
+			inputContainer.removeChild( originalInput );
 			const values = [];
 			const otpCodeInput = document.createElement( 'input' );
 			otpCodeInput.setAttribute( 'type', 'hidden' );
@@ -492,7 +493,6 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 				} );
 				inputContainer.appendChild( digit );
 			}
-			inputContainer.removeChild( originalInput );
 		} );
 
 		/**
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -454,7 +454,7 @@ const convertFormDataToObject = ( formData, includedFields = [] ) =>
 							}
 							break;
 						default:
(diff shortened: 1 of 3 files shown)


thomasguillot:
@miguelpeixe this is great thank you very much!

I pushed some small changes to the inputs (added a max-width, centered them). I also removed the transparent colour on the caret, I thought it was strange that the input was focused but I couldn't see if I could type in or not.

leogermani (changes requested):
Everything works exactly as described!

The paste feature that allows you to paste from any input is nice!

One problem though is that if you press any Function key (F1, F2, etc), it writes the key name to the input:

Captura de tela de 2022-10-06 17-59-26

Other than that, a minor non-blocking issue I saw, that is probably harmless but I'll point it out:

the values you see on the inputs are not always in sync with the value of the hidden input that is sent to the server.

For example here:

https://github.com/Automattic/newspack-plugin/pull/2043/files#diff-6acf95fe4c12b4b7a9e06e36f6eafd3dcf171efa68b93f2650c2c5adbbb28781R485

The pasted value is added to the value before any check is done. The same happens when you delete the value from the input, it is still there in the hidden field.

The only problem it creates is that you are able to submit the form even if the fields are not filled, because the hidden input is filled.

adekbadek (approved):
🙌

adekbadek (commented):
Whoops, didn't notice @leogermani 's review! Ignore my approval.

miguelpeixe:
Thank you for reviewing!

> One problem though is that if you press any Function key (F1, F2, etc), it writes the key name to the input:

Nice catch! 9a2baa42d09a85903f4ca4a15b454dfc42277c5b updates the regex and ensures only values of length 1 can be inserted

> the values you see on the inputs are not always in sync with the value of the hidden input that is sent to the server.

9a2baa42d09a85903f4ca4a15b454dfc42277c5b also ensures that the hidden value is updated with validated content

leogermani:
@miguelpeixe there was a small typo I fixed here: https://github.com/Automattic/newspack-plugin/pull/2043/commits/9a9511ab7f5fc00c1105bb631c6fdcec2959c426

And I also [suggested](https://github.com/Automattic/newspack-plugin/pull/2043/commits/13762775e383555205ca0d9c239d4bfd634fc551) to update the value when you delete one of the fields. If you think it's ok, we are good to go

leogermani (approved):

miguelpeixe:
Good idea! Thanks

PR #2065, opened 2022-10-05
fix: fire ini-load analytics events on DOMContentLoaded, not load

dkoo:
### All Submissions:

For non-AMP sites, `ini-load` AMP Analytics events are triggered on the [window `load` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event). This event only fires after all of the page's resources have been fully loaded, including HTML content, images, and JS (first and third-party). On live non-AMP pages, it seems that this event is sometimes never fired if any resources fail to fully load for whatever reason, which is more often than might be expected. This PR sidesteps the issue by triggering `ini-load` events on the [`DOMContentLoaded` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) instead, which fires as soon as the document's HTML is fully loaded. This makes these events much less likely to be blocked by external resources failing to load.

Closes https://github.com/Automattic/newspack-popups/issues/981.

1. Turn off AMP or turn on AMP Transitional mode on your test site.
2. In your browser's Dev Tools > Network tab, load a page that will display at least one prompt and filter requests by type: JS. Pick any JS file and temporarily block the request. This will prevent the window `load` event handler from ever firing.
3. On `master`, refresh the page and filter network requests by "Newspack Announcement." Observe that the "Load" events don't fire, but "Seen" events do if you scroll the prompt into the viewport.
4. Check out this branch and repeat step 3. Confirm that "Load" events do fire once the page is loaded.

### Other information:

### DIFF
--- a/includes/class-analytics.php
+++ b/includes/class-analytics.php
@@ -888,7 +888,7 @@ class Analytics {
 						observer.observe(elements[i], { attributes: true });
 					}
 				} else {
-					window.addEventListener('load', handleEvent)
+					window.addEventListener('DOMContentLoaded', handleEvent)
 				}
 			} )();
 		</script>


adekbadek (approved):
Great sleuthing!

PR #2075, opened 2022-10-11
fix: "Sign In" page styles

miguelpeixe:
### All Submissions:

#2025 introduced a few issues to the Sign In form styles while accessing through the "My Account" page. This PR fixes those issues:

 - Dark text color for "I don't have an account" link
 - Horizontal scrolling

| Master | This branch |
| --- | --- |
| image | image |

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/style.scss
+++ b/assets/blocks/reader-registration/style.scss
@@ -232,7 +232,8 @@
 		display: none;
 	}
 
-	.nphp {
+	.nphp,
+	input[type='email'].nphp {
 		@include mixins.visuallyHidden;
 	}
 }
--- a/assets/reader-activation/auth.scss
+++ b/assets/reader-activation/auth.scss
@@ -2,6 +2,7 @@
 @use '~@wordpress/base-styles/colors' as wp-colors;
 @use '../shared/scss/mixins';
 
+.entry-content .newspack-reader,
 .newspack-reader {
 	/**
 	 * Account Link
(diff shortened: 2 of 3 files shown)


dkoo (changes requested):
This works but also overrides the `active` and `hover` styles set [here](https://github.com/Automattic/newspack-plugin/pull/2075/files#diff-b97350f2168c727d03377092125570004f9fecf32bd45a8ac3433040ab230f4dR227). I'm curious about where the dark color is coming from (I'm not seeing it locally) and if we can fix it using a better selector instead of `!important`?

miguelpeixe:
I'm also not a fan of `!important`. My goal was to use the same solution already used for the registration block:

https://github.com/Automattic/newspack-plugin/blob/c89a4674809fe723b7400919ffe9c3e6400b99f4/assets/blocks/reader-registration/style.scss#L116-L133

The color applied varies according to the publisher-defined secondary color contrast resolution from `newspack-theme`:

https://github.com/Automattic/newspack-theme/blob/8f7796735ecb5f98e3481f3c8b92898f3d0224a9/newspack-theme/inc/color-patterns.php#L184-L191

This rule will be prioritized because it has the same amount of levels (`.entry-content a` and `.newspack-reader__auth-form__have-account a`) and is loaded afterward. To avoid `!important`, we'd have to add another selector or make the `.entry-content a` render before the scripts.

miguelpeixe:
639222f4477ee810d2b3e68480619f8d92acbe6f tweaks the selectors to reduce the need for the `!important` property

dkoo (approved):
Nice, this doesn't override the pseudoclasses. 👍

PR #2076, opened 2022-10-11
fix: check Woo plugins status before enabling RAS

dkoo:
### All Submissions:

RAS features will behave unpredictably if WooCommerce or WooCommerce Subscriptions plugins are not active. This PR adds a check for these required plugins before considering RAS enabled  Also adds a plugin installer component to the RAS wizard if either plugin is missing for ease of setup.

Also updates the description for WooCommerce Subscriptions with the correct plugin description from the WP.org repository.

1. On a site with RAS already enabled and configured, check out this branch and deactivate WooCommerce and/or WooCommerce Subscriptions.
2. Confirm on the front-end that no reader-facing RAS features are enabled (no My Account link, no registration block).
3. In the dashboard > Newspack > Engagement Reader Activation, confirm that you see a plugin installer prompt:

Screen Shot 2022-10-11 at 3 30 14 PM

4. Install/activate both plugins and confirm that the RAS wizard reappears after both plugins are confirmed active.
5. Revisit the front-end and confirm that reader-facing RAS features are available again.

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -259,6 +259,10 @@ final class Reader_Activation {
 	 * @return bool True if reader activation is enabled.
 	 */
 	public static function is_enabled( $strict = true ) {
+		if ( defined( 'IS_TEST_ENV' ) && IS_TEST_ENV ) {
+			return true;
+		}
+
 		$is_enabled = defined( 'NEWSPACK_EXPERIMENTAL_READER_ACTIVATION' ) && NEWSPACK_EXPERIMENTAL_READER_ACTIVATION;
 
 		if ( ! $strict ) {
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -175,6 +175,7 @@ final class Reader_Activation {
 			'sender_name'                 => Emails::get_from_name(),
 			'sender_email_address'        => Emails::get_from_email(),
 			'contact_email_address'       => Emails::get_reply_to_email(),
+			'plugins_configured'          => self::is_woocommerce_active(),
 		];
 
 		/**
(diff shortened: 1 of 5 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -240,6 +241,15 @@ public static function update_setting( $key, $value ) {
 		return \update_option( self::OPTIONS_PREFIX . $key, $value );
 	}
 
+	/**
+	 * Check if the required Woo plugins are active.
+	 *
+	 * @return boolean True if all required plugins are active, otherwise false.
+	 */
+	public static function is_woocommerce_active() {
```
adekbadek, line 249: There's already the `Donations::is_woocommerce_suite_active` method – it should be used instead.

adekbadek (changes requested):
Works fine, but there's one code duplication issue.

includes/reader-activation/class-reader-activation.php
leogermani: on top of that, that method also checks for the "Name your price" plugin. We should add this to the messaging as well.

adekbadek: Or maybe just abstract woo-checking plugin list, since it's [also used in the Reader Revenue wizard](https://github.com/Automattic/newspack-plugin/blob/39a247459b4151901799093e8bcd311d4f97abe7/assets/wizards/readerRevenue/views/platform/index.js#L55-L60).

dkoo: The checks don't necessarily overlap, though, because Reader Activation doesn't need Name Your Price or the Stripe Gateway extension. It only needs WC and WC Subscriptions if the Stripe platform is being used. So I think it's okay to keep these as separate methods and checks.

adekbadek (approved):

miguelpeixe:
`Reader_Activation:is_enabled()` can be called before the `WooCommerce` class is available. This is the case for the initialization of RAS:

https://github.com/Automattic/newspack-plugin/blob/3aedbd486ac1e3522256bb0a4a9765e01988d8c6/includes/reader-activation/class-reader-activation.php#L66-L68

This is currently breaking RAS functionality on the frontend and should be reverted or resolved asap.\

cc @dkoo

PR #2078, opened 2022-10-12
fix: use default terms text in reg block if no text defined

dkoo:
### All Submissions:

Uses the global "terms & conditions" text as set in the Reader Activation wizard UI as the default text in the registration block.

1. Check out this branch.
2. Visit the Newspack > Engagement > Reader Activation wizard and note or update the terms & conditions text here.
3. Add a Reader Registration block to a post or page. Confirm that the text in the RichText field at the bottom of the block matches the wizard text in step 2.
4. View the block on the front-end and confirm that the text matches here, too.
5. Update the text in the block and update. Confirm that the block now shows the updated text on the front-end.

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -59,8 +59,12 @@ export default function ReaderRegistrationEdit( {
 } ) {
 	const blockProps = useBlockProps();
 	const [ editedState, setEditedState ] = useState( editedStateOptions[ 0 ].value );
-	const { reader_activation_terms: defaultTermsText, reader_activation_url: defaultTermsUrl } =
+	let { reader_activation_terms: defaultTermsText, reader_activation_url: defaultTermsUrl } =
 		window.newspack_blocks;
+
+	if ( defaultTermsUrl ) {
+		defaultTermsText = `<a href="${ defaultTermsUrl }">` + defaultTermsText + '</a>';
+	}
 
 	const innerBlocksProps = useInnerBlocksProps(
 		{},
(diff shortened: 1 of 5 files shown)


assets/blocks/reader-registration/edit.js
```diff
@@ -313,8 +315,8 @@ export default function ReaderRegistrationEdit( {
 										<RichText
 											onChange={ value => setAttributes( { privacyLabel: value } ) }
 											placeholder={ __( 'Terms & Conditions statement…', 'newspack' ) }
-											value={ privacyLabel }
-											tagName="p"
+											value={ privacyLabel || defaultTermsText }
+											tagName={ defaultTermsUrl ? 'a' : 'p' }
```
miguelpeixe, line 319: The `tagName` must still be a `p` (with an `<a />` inside) in order for the link to be editable. This strategy also doesn't implement the correct markup:

```html
<p><a href="#">Text</a></p>
```

miguelpeixe (changes requested):

assets/blocks/reader-registration/edit.js
dkoo: Ah, thanks for highlighting. I wasn't sure what the correct nesting would be for the `RichText` tag.

dkoo: 333303b should render the correct markup in the editor.

miguelpeixe (changes requested):
The current approach generates the following markup when the URL is not available:

`<p><p>Text</p></p>`

Nesting the two paragraphs also breaks the styling with a tiny font, because of the double 0.5 relative font size.

Ideally, the `RichText` would always remain `p` and its `value` would have, or not, `<a />` or plain text.

dkoo:
> Ideally, the `RichText` would always remain `p` and its `value` would have, or not, `<a />` or plain text.

This isn't possible because the `value` for `RichText` cannot contain React components. If the main issue is showing that the Terms text will be linked in the editor if there's a URL in global settings, could we mock the styles with an editor-only class name on the parent element and make the value _look_ like a link in the editor?

On the front-end, it's much easier to simply render or not render an `a` wrapper based on whether the URL exists, so this seems like an editor preview thing to me.

miguelpeixe:
In that case, we should probably stick with `span` for the RichText element to not generate invalid markup.

dkoo:
I think de63f1d should do it.

miguelpeixe (changes requested):
The link is still hardcoded and not editable when applied with the default URL value.

I pushed 65c29745c72635ce18c940b194850402417e3db4 with the link as `<a />` inside the value. LMK if this works

dkoo:
Thank you for the suggestion—I was just not getting what you wanted me to do. 😆 The change seems like it should work, but it looks like the anchor tag is getting stripped when rendered as the `RichText` value.

miguelpeixe:
> it looks like the anchor tag is getting stripped when rendered as the `RichText` value.

That's strange. This is what I get on a freshly added block:

https://user-images.githubusercontent.com/820752/197850867-d0cfeb8b-06d3-4c42-8c8a-a579840e071c.mov

I have my TOS URL as `https://google.com`

dkoo:
Ah, it does work with a fresh block, but if you edit the text the link disappears.

miguelpeixe:
Couldn't reproduce it either... On my end, it behaves just like a regular link would:

https://user-images.githubusercontent.com/820752/197874137-5b6f4a1b-2b6a-45c5-be8f-64abc1041380.mov

The generated `RichText` value is exactly what this component would generate if a link was to be created through its UI.  I don't see a reason why it would be problematic.

dkoo:
> If you clear the contents, it's expected for the link to disappear as well. We want publishers to be able to tweak this text and link placement.

Ah, I understand what you mean now! Thanks for clarifying. I was confused by the default link not appearing on a block I had already edited, and also disappearing when I cleared the contents from a new block. But your solution makes sense and provides more flexibility than hardcoding the default URL. 👍 Should we have @leogermani do one last review since we both submitted code?

PR #2086, opened 2022-10-20
fix: apply My Account template changes only when RAS is active

dkoo:
### All Submissions:

Conditionally applies RAS template changes and logic to the WC My Account UI only when RAS is active. Non-RAS sites are reporting that customers are confused by UI changes that shouldn't apply to their account pages.

Also checks whether logged-in customers have any available downloads before hiding the "Downloads" item from the menu.

#### My Account UI

1. Disable RAS.
2. Log into My Account as a WC customer, or set your Reader Revenue platform to Newspack and make a recurring donation via WC.
3. Visit the My Account page. On `master`, observe that the RAS UI is applied (links to send a reset password email, delete account, etc. which will only work if RAS is active).
4. Apply this patch and refresh. Confirm that the My Account page now shows the standard WooCommere UI with all nav menu items, and RAS-specific UI not shown. Note that the custom theme styles still apply, e.g. the nav menu is still a sidebar, but these styles should be cosmetic only.

#### Downloads

1. In the Products dashboard, edit the product used to test the donation above and add a downloadable file. Hint: look for a panel like this:

Screen Shot 2022-10-20 at 5 08 15 PM

2. Enable RAS.
3. Log into My Account as a WC customer, or make a donation using this product via the simplified Donate block.\
4. On `master`, observe that there is no "Downloads" link in the My Account nav menu.
5. Apply the patch and refresh. Confirm that the Downloads link reappears.
6. In a new session, make a donation using a different product without any downloadable files.
7. In My Account, confirm that the Downloads link does not appear, since this customer doesn't have access to any downloadable files.

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -37,6 +37,10 @@ class WooCommerce_My_Account {
 	 * @codeCoverageIgnore
 	 */
 	public static function init() {
+		if ( ! Reader_Activation::is_enabled() ) {
+			return;
+		}
+
 		\add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
 		\add_filter( 'woocommerce_account_menu_items', [ __CLASS__, 'my_account_menu_items' ], 1000 );
 		\add_action( 'woocommerce_account_' . self::BILLING_ENDPOINT . '_endpoint', [ __CLASS__, 'render_billing_template' ] );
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
This change will also remove the Stripe billing link, which I don't think is related only to RAS. RAS-specific features are account verification, deletion, and password reset/creation, which is basically the "Account details" template replacement/forced redirection. 

> Confirm that the My Account page now shows the standard WooCommere UI 

The navigation design does not render the WooCommerce default, which I believe is also intended because it's a theme perk and an overall design improvement that fits well with any "My Account" page section:

image

dkoo:
> This change will also remove the Stripe billing link, which I don't think is related only to RAS.

b6e22cd restores the Stripe Billing portal nav menu item and link, with or without RAS.

> The navigation design does not render the WooCommerce default, which I believe is for the best because it's a theme perk and an overall design improvement that fits well with any "My Account" page section

Yes, that's expected—the design changes are implemented in the theme, are cosmetic only, and are not dependent on RAS.

miguelpeixe (approved):

PR #2093, opened 2022-10-25
feat(newsletter-contact): disallow overriding membership status field

adekbadek:
### All Submissions:

Prevents a situation where the `NP_Membership Status` field in ESP data is overridden in an undesirable way. This field should have the value denoting a one-time donor only if the donor has _never_ donated in a recurring manner.  

1. Set up a site with RAS enabled and ESP set to ActiveCampaign
2. Make a montly donation, observe the `NP_Membership Status` is set to `Monthly Donor`
3. Make a one-time donation, observe the `NP_Membership Status` field is not changed, but the values of `NP_Total Paid` and ``NP_Last Payment Amount` are updated accordingly

### Other information:

### DIFF
--- a/includes/plugins/class-newspack-newsletters.php
+++ b/includes/plugins/class-newspack-newsletters.php
@@ -146,7 +146,7 @@ class Newspack_Newsletters {
 					}
 				}
 
-				// If the membership status it to be switched from recurring to non-recurring, ignore this change.
+				// If the membership status is to be switched from recurring to non-recurring, ignore this change.
 				if ( $contact['existing_contact_data'] && isset( $contact['metadata'][ self::$metadata_keys['membership_status'] ], $existing_metadata[ self::$metadata_keys['membership_status'] ] ) ) {
 					$existing_metadata  = $contact['existing_contact_data']['metadata'];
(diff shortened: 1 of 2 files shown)


includes/plugins/class-newspack-newsletters.php
```diff
@@ -145,6 +146,22 @@ public static function contact_data( $contact, $selected_list_ids, $provider ) {
 					}
 				}
 
+				// If the membership status it to be switched from recurring to non-recurring, ignore this change.
```
miguelpeixe, line 149:
```suggestion
				// If the membership status is to be switched from recurring to non-recurring, ignore this change.
```

miguelpeixe (approved):
It's working as described, just a small nit below.

I'm getting a fatal error from Stripe's webhook request that is not affecting this PR:

```
PHP Fatal error:  Uncaught Error: Cannot use object of type WP_Error as array in newspack-plugin/includes/reader-revenue/class-stripe-connection.php:1369
```

Looks like the `Newspack\Stripe_Connection::get_invoice()` method is returning `WP_Error`, which is not being handled by `Newspack\Stripe_Connection::create_wc_transaction_payload()`

I filed #2094 for this.

PR #2102, opened 2022-11-01
Feat/author fields

### LINKED ISSUE #1723: Custom user profile fields
Some publishers (MTFP, Berkeleyside) use either [Trust Indicators](https://github.com/Automattic/newspack-theme/issues/1803) or [Yoast SEO Premium](https://yoast.com/help/implementing-schema-with-yoast-seo/) plugins to extend the fields available on user profiles. Some common fields are:

* Job title
* Role (staff, freelancer, etc.)
* Employer/Organization
* Short bio
* Phone number

Implementation of these fields is not standardized: currently, 36 sites have Yoast SEO Premium, and only 4 sites use Trust Indicators. However, the need for these fields is broader than that, with some publishers hacking solutions such as changing the display name to be the "Joe Example, Managing Editor". Storing that data as a single string prevents its discrete use elsewhere.

We should [develop our own standard](https://xkcd.com/927/) and build this functionality into Newspack Plugin. These fields can then be supported and displayed by `newspack-blocks`, such as [Author Profile & List](https://github.com/Automattic/newspack-blocks/issues/1166).

In a future state, we should also consider whether Co-Authors Plus and guest authors' profiles would be extended to support these fields. There is an [open issue for CAP to use Yoast's Schema API](https://github.com/Automattic/Co-Authors-Plus/issues/710). If Newspack Plugin employed a similar API, then perhaps CAP could effortlessly plug into either Newspack or Yoast. Such compatibility might also be beneficial to Newspack Blocks, which could be made aware of Yoast's fields, so that the 36 publishers wouldn't need to migrate their existing data from Yoast to Newspack Plugin.

adekbadek:
### All Submissions:

Closes #1723

1. Set Newspack Blocks to `feat/author-blocks-fields` branch 
2. Set Newspack Theme to `feat/author-custom-fields` branch
1. Ensure you have some regular and Co-Authors Plus (guest) authors on the site
1. Insert an Authors List block and two Author Profile blocks on a page
3. Set the list block to display both guest and regular authors
4. Set the profile blocks to display a regular and a guest author
5. View the editor and the front-end, verify nothing is broken
6. Edit the WP author, add some values in the "Additional user profile fields" section
7. Edit the guest author, add some values in new fields in the "About the guest author" section 
8. Again, view the page in the editor and on the front-end and see that field values are reflected there
9. View a post by the guest author and one by the WP author, see that the "Job Title" value is displayed in the author bio footer, next to the name

### Other information:

### DIFF
--- a/includes/authors/class-authors-custom-fields.php
+++ b/includes/authors/class-authors-custom-fields.php
@@ -13,12 +13,20 @@ defined( 'ABSPATH' ) || exit;
  * Newspack Authors Custom Fields class.
  */
 final class Authors_Custom_Fields {
+	const USER_META_NAMES = [
+		'job_title'    => 'newspack_job_title',
+		'role'         => 'newspack_role',
+		'employer'     => 'newspack_employer',
+		'phone_number' => 'newspack_phone_number',
+	];
+
 	/**
 	 * Initialize hooks.
 	 */
 	public static function init() {
 		\add_action( 'edit_user_profile', [ __CLASS__, 'edit_user_profile' ] );
 		\add_action( 'edit_user_profile_update', [ __CLASS__, 'edit_user_profile_update' ] );
+		\add_filter( 'newspack_author_bio_name', [ __CLASS__, 'newspack_author_bio_name' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


leogermani (changes requested):
It works as expected except from one issue

Editing guest author is fine, but the extra fields for regular users only show up if I'm editing another user other than myself.

To display and save the extra fields when editing your own profile, you must use `show_user_profile`  and `personal_options_update` hooks.

Also, not sure if this is related to this PR, but the link in the "More by __author__" in the blocks use the `?author_name=$name` format instead of the pretty permalink. Any reason for that? The same link looks fine in the Footer bio when you visit a post.

adekbadek:
> Editing guest author is fine, but the extra fields for regular users only show up if I'm editing another user other than myself.

Thanks for catching that! Fixed in c72631b2bb8eafab2c966093fe975a5018966887

> the link in the "More by author" in the blocks use the ?author_name=$name format instead of the pretty permalink. Any reason for that? 

This is not something this PR touches. It is like that [only for guest authors](https://github.com/Automattic/newspack-blocks/blob/7ccd995e580523b2b94117cb3cad50d5504cbb65/src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php#L256) – maybe they are not guaranteed to have a permalink, because there might be name collisions with a WP user.

leogermani (approved):

adekbadek:
Waiting with the merge for approval on the related [Blocks](https://github.com/Automattic/newspack-blocks/pull/1302) and [Theme](https://github.com/Automattic/newspack-theme/pull/1966) PRs, because all three should be released as a single feature-set.

PR #2108, opened 2022-11-03
feat: add setup cli command

leogermani:
### All Submissions:

Creates a CLI command to quickly set up a new site from scratch, running the onboarding wizard automatically

1. Checkout this branch on a fresh new site
2. Run `wp newspack setup`
3. Verify that the site is configures, with the required plugins and initial content. Then end result should be the same as if you had gone through the Onboarding wizard choosing all default values

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -153,6 +153,8 @@ final class Newspack {
 
 		// Filter by authors in the Posts page.
 		include_once NEWSPACK_ABSPATH . 'includes/author-filter/class-author-filter.php';
+
+		\Newspack\CLI\Initializer::init();
 	}
 
 	/**
--- a/includes/cli/class-setup.php
+++ b/includes/cli/class-setup.php
@@ -1,0 +1,195 @@
+<?php
+/**
+ * Newspack setup wizard. Required plugins, introduction, and data collection.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\CLI;
+
+use WP_CLI;
+use WP_REST_Request;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Setup Newspack CLI command.
+ */
+class Setup {
+
+	/**
+	 * Sets up a testing site
+	 *
+	 * This is a command used to set up a testing site from scratch, installing required plugins and running the onboard wizard automatically
+	 *
+	 * ## OPTIONS
+	 * [--skip-ras]
+	 * : If set, will not add the EXPERIMENTAL RAS flag to wp-config
+	 *
+	 * @param array $args Positional arguments.
+	 * @param array $assoc_args Assoc arguments.
+	 * @return void
+	 */
+	public function __invoke( $args, $assoc_args ) {
+		do_action( 'rest_api_init' );
+		wp_set_current_user( 1 );
+
+		$this->plugins();
+
+		$this->initial_content();
+
+		$this->campaigns_config();
+
+		$this->amp_plus();
+
+		if ( ! WP_CLI\Utils\get_flag_value( $assoc_args, 'skip-ras', false ) ) {
+			$this->ras_beta();
+		}
+
+		WP_CLI::success( 'Done!' );
+	}
+
+	/**
+	 * Do the initial-check request
+	 *
+	 * @return array The plugins array
+	 */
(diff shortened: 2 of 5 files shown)


includes/cli/class-setup.php
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Setup Newspack CLI command.
+ */
+class Setup {
+
+	/**
+	 * Sets up a testing site
+	 *
+	 * This is a command used to set up a testing site from scratch, installing required plugins and running the onboard wizard automatically
```
adekbadek, line 23: I don't think we should treat it as "testing site" setup. Rather a "site setup", simply. The starter content could be optional in the command, but anyway it can be deleted later.

includes/cli/class-setup.php
```diff
@@ -0,0 +1,195 @@
...
+	 * [--skip-ras]
+	 * : If set, will not add the EXPERIMENTAL RAS flag to wp-config
+	 *
+	 * @param array $args Positional arguments.
+	 * @param array $assoc_args Assoc arguments.
+	 * @return void
+	 */
+	public function __invoke( $args, $assoc_args ) {
+		do_action( 'rest_api_init' );
+		wp_set_current_user( 1 );
```
adekbadek, line 35: This assumes user of ID `1` is the admin. The command should check if that's true first, and perhaps create an admin user with an environment-defined username.

includes/cli/class-setup.php
```diff
@@ -0,0 +1,195 @@
...
+	 * Gets the initial check and loops though plugins. Configures plugins that are not yet active
+	 *
+	 * @return void
+	 */
+	private function plugins() {
+		WP_CLI::line( 'Checking for required plugins' );
+		$plugins = $this->get_initial_check();
+
+		foreach ( $plugins as $plugin ) {
+			if ( $plugin['Status'] === 'active' ) {
```
adekbadek, line 73: There are a few [linting issues](https://app.circleci.com/pipelines/github/Automattic/newspack-plugin/5492/workflows/569a5659-37d0-4036-afc5-2ec407816ec5/jobs/25747), starting here.

includes/class-newspack.php
```diff
@@ -153,6 +153,8 @@ private function includes() {
 
 		// Filter by authors in the Posts page.
 		include_once NEWSPACK_ABSPATH . 'includes/author-filter/class-author-filter.php';
+
+		\Newspack\CLI\Initializer::init();
```
adekbadek, line 157: When running the CLI command, I get `PHP Fatal error:  Uncaught Error: Class "Newspack\CLI\Initializer" not found` error. The `includes/cli/class-initializer.php` file is not required anywhere.

adekbadek (changes requested):

includes/cli/class-setup.php
leogermani: I think it can evolve into a multi purpose command. I just wanted to explicitly say that to keep the scope small for now, and allow the script to make certain choices

includes/cli/class-setup.php
leogermani: I updated it to find the first admin.

But as I said, I'm not too worried about this as the scope is only for testing sites. For example, we also assume wp-config.php is placed in the ABSPATH, which is not true for every site.

includes/cli/class-setup.php
leogermani: fixed

leogermani:
The onboard wizard install all the plugins but it doesn't activate most of them. Should we activate some of the plugins? Like Woocommerce for example?

adekbadek:
Which ones does it not activate? It looks like any call to [plugin configuration code would activate the plugin](https://github.com/Automattic/newspack-plugin/blob/04eb0be8fedbda216a07da9c2fe64a42bc9c0a73/includes/api/class-plugins-controller.php#L360-L368).

adekbadek (approved):

adekbadek:
It's not required to run Newspack, but is for RAS. The issue is that some of the RAS-required plugins are premium plugins, not installable from WPORG. I've a solution for that (developed for [newspack-e2e-tests](https://github.com/Automattic/newspack-e2e-tests), though. This can be tacked in a separate PR, not to block things here.

PR #2125, opened 2022-11-10
update newspack-scripts and CI orb

adekbadek:
Update `newspack-scripts` and CI orb. The latest version combo will handle building JS files on CI better:

- version `1.4.2` of the orb adds setting node version according to the project's `.nvmrc` file prior to installing npm modules and building the JS files
- version `4.3.8` of `newspack-scripts` removes a hacky partial fix for issues caused by node version mismatch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Builds okay locally—let's merge this and check the artifact from `master`

PR #2143, opened 2022-11-21
Stripe webhooks management

adekbadek:
### All Submissions:

Adds webhook management UI for Stripe Reader Revenue platform. 

Closes 1203451787793554-as-1203362361069776

1. Visit Stripe Settings tab in Reader Revenue wizard (set Stripe as the platform first)
2. If Stripe is connected, observe the webhooks list
3. Test that the webhooks can be disabled/enabled and deleted using this UI

### Other information:

### DIFF
--- a/assets/components/src/accordion/index.js
+++ b/assets/components/src/accordion/index.js
@@ -1,0 +1,33 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useState } from '@wordpress/element';
+import { Icon, chevronDown } from '@wordpress/icons';
+
+/**
+ * External dependencies
+ */
+import classNames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+const Accordion = ( { children } ) => {
+	const [ isOpen, setIsOpen ] = useState( false );
+	return (
+		<details
+			className={ classNames( 'newspack-accordion', { 'newspack-accordion--is-open': isOpen } ) }
+		>
+			<summary onClick={ () => setIsOpen( ! isOpen ) }>
+				{ __( 'Webhooks not connected to this site.', 'newspack' ) }
+				<Icon className="newspack-accordion__icon" icon={ chevronDown } size={ 24 } />
+			</summary>
+			{ children }
+		</details>
+	);
+};
+
+export default Accordion;
--- a/assets/components/src/accordion/index.js
+++ b/assets/components/src/accordion/index.js
@@ -15,14 +15,14 @@ import classNames from 'classnames';
  */
 import './style.scss';
 
-const Accordion = ( { children } ) => {
+const Accordion = ( { children, title } ) => {
 	const [ isOpen, setIsOpen ] = useState( false );
 	return (
 		<details
 			className={ classNames( 'newspack-accordion', { 'newspack-accordion--is-open': isOpen } ) }
 		>
 			<summary onClick={ () => setIsOpen( ! isOpen ) }>
-				{ __( 'Webhooks not connected to this site.', 'newspack' ) }
+				{ title }
 				<Icon className="newspack-accordion__icon" icon={ chevronDown } size={ 24 } />
 			</summary>
 			{ children }
--- a/assets/components/src/accordion/index.js
+++ b/assets/components/src/accordion/index.js
@@ -1,7 +1,6 @@
 /**
  * WordPress dependencies
  */
-import { __ } from '@wordpress/i18n';
 import { useState } from '@wordpress/element';
 import { Icon, chevronDown } from '@wordpress/icons';
 
--- a/assets/wizards/readerRevenue/views/stripe-setup/index.js
+++ b/assets/wizards/readerRevenue/views/stripe-setup/index.js
@@ -3,8 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { CheckboxControl, ExternalLink, ToggleControl } from '@wordpress/components';
-import { useDispatch, useSelect } from '@wordpress/data';
-import { trash } from '@wordpress/icons';
+import { useDispatch } from '@wordpress/data';
 
 /**
  * External dependencies
@@ -22,7 +21,7 @@ import {
 	SelectControl,
 	TextControl,
 	Wizard,
-	Accordion,
+	ActionCard,
 } from '../../../../components/src';
 import NewsletterSettings from './newsletter-settings';
 import { STRIPE, READER_REVENUE_WIZARD_SLUG } from '../../constants';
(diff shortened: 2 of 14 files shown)


assets/wizards/readerRevenue/views/stripe-setup/index.js
```diff
@@ -103,6 +105,115 @@ export const StripeKeysSettings = () => {
...
+					) }
+				/>
+			) }
+			{ activeThisSiteWebhooks.length === 0 && (
+				<Notice
+					isError
+					noticeText={ __( 'There are no active webhoooks for this site.', 'newspack' ) }
+				/>
+			) }
+			<WebhooksList webhooks={ thisSiteWebhooks } onUpdate={ handleWebhooksListUpdate } />
```
dkoo, line 207: This initial view was confusing to me for a bit:

Screen Shot 2022-11-30 at 4 55 58 PM

The "Webhooks not connected to this site." accordion heading at first looked to me like a message related to the webhook for my site because of the plain styling. It also wasn't immediately clear to me that I could click on that heading to display other webhooks.

Lastly, it's a little confusing to me that the toggle switches trigger an immediate change, with the "Save Settings" button being displayed right below them. It feels unclear at first whether I'll need to click that button to persist any changes I make.

Suggestions:

* Provide a more prominent heading for each list: maybe "Webhooks for this site" and "Other webhooks"
* I think the Accordion component will be really useful going forward, but the styling seems off to me. Can we increase the font-size and maybe apply some styling to the Accordion component's heading to make it look more like a heading inside a button? I can imagine it more visually resembling one of the rows in the expanded view, or perhaps wrap everything in a container panel like the [Material accordion](https://m1.material.io/components/expansion-panels.html).
* Perhaps move this whole section out into a page of its own? Since it's a relatively advanced feature, maybe it doesn't have to live alongside Stripe Settings in the main wizard nav, but can be accessed via a button or action card clicked from the Stripe Settings page, kind of like the GAM configuration settings in the Advertising wizard.
* Consider having a link out to the Stripe dashboard URL (https://dashboard.stripe.com/) to make it clearer that you're modifying data that lives in that external dashboard. If it's possible to intuit the Stripe dashboard page for each webhook from the webhook data, it might be useful to link out each webhook to that page in Stripe.

assets/wizards/readerRevenue/views/stripe-setup/index.js
```diff
@@ -103,6 +105,115 @@ export const StripeKeysSettings = () => {
...
+					noticeText={ __(
+						'There are too many webhoooks for this site. Please delete or disable the extra ones.',
+						'newspack'
+					) }
+				/>
+			) }
+			{ activeThisSiteWebhooks.length === 0 && (
+				<Notice
+					isError
+					noticeText={ __( 'There are no active webhoooks for this site.', 'newspack' ) }
```
dkoo, line 204: It's possible (in fact, very easy) to delete the active webhook for your site. Once that happens it seems like there's nothing to be done to bring it back, but if you refresh the page it's recreated automatically. Perhaps we could show a.) a warning before triggering the deletion, and b.) a "reset" button after deletion that will recreate the webhook and update the UI in this instance. This could avoid some panic that might occur if someone clicks the delete button by accident. In general I can also see a reset button being useful in case a webhook goes haywire on the Stripe end—you may want to recreate it to fix things.

dkoo (changes requested):
Thanks for the new UI—this works well on a technical level and solves the original issue in which there's currently no way to see or disable unwanted webhooks, or intentionally trigger a refresh of the current site's webhook if needed. I have some thoughts about the UI, though, which are open to discussion.

assets/wizards/readerRevenue/views/stripe-setup/index.js
dkoo: @thomasguillot I see you already gave a 👍 on the design, so I'd defer to you as to whether we should consider any of my above suggestions.

assets/wizards/readerRevenue/views/stripe-setup/index.js
dkoo: @thomasguillot could use your opinion on this, too. Maybe I'm overthinking things.

assets/components/src/accordion/index.js
```diff
@@ -0,0 +1,33 @@
...
+import './style.scss';
+
+const Accordion = ( { children } ) => {
+	const [ isOpen, setIsOpen ] = useState( false );
+	return (
+		<details
+			className={ classNames( 'newspack-accordion', { 'newspack-accordion--is-open': isOpen } ) }
+		>
+			<summary onClick={ () => setIsOpen( ! isOpen ) }>
+				{ __( 'Webhooks not connected to this site.', 'newspack' ) }
```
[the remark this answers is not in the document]
adekbadek: Fixed in 7e2aa95!

assets/wizards/readerRevenue/views/stripe-setup/index.js
thomasguillot: If we move this to its own page, I'm not sure we actually need an accordion. 

Screenshot 2022-12-06 at 12 29 51

Clicking on "Edit Webhooks" would then take the user to a new screen:

webhooks i3

assets/wizards/readerRevenue/views/stripe-setup/index.js
thomasguillot: I don't really understand the "Reset Webhook" thing but if you think it makes sense then we could add the button to the list of buttons at the bottom (see https://github.com/Automattic/newspack-plugin/pull/2143#discussion_r1040915445)

dkoo: Maybe I'm overcomplicating it. My main concern is that it seems too easy to accidentally delete the active webhook for the current site. Maybe the only change needed is to change the "trash" button on that one webhook to a "reset" button that will delete and recreate the webhook and update the UI accordingly. Or, more simply, to just remove the trash button entirely if we don't think that's a use case we need to support.

assets/wizards/readerRevenue/views/stripe-setup/index.js
adekbadek: >  it's a little confusing to me that the toggle switches trigger an immediate change, with the "Save Settings" button being displayed right below them.

I agree with this concern, and moving to a separate page sounds like a good solution. I don't like the proposed placement of the "Edit Webhooks" link, though. It feels out of place there, as the secondary action to saving the settings. An action card seems like the idiom we use in these cases:

image

thomasguillot: I'm not a fan of the `ActionCard` underneath the "Save Settings" `Button` but you're correct, it's better than a secondary `Button`.

adekbadek: >  it's a little confusing to me that the toggle switches trigger an immediate change, with the "Save Settings" button being displayed right below them.

I second this concern, and the suggestion to move it to a separate page. I don't like the proposed placement of the "Edit Webhooks" link, though. It feels out of place as a secondary action to settings saving. In other places we use the action card to link to other pieces of UI:

image

adekbadek: > I think the Accordion component will be really useful going forward, but the styling seems off to me. 

> If we move this to its own page, I'm not sure we actually need an accordion.

I think we do – the "other webhooks" section is not of interest 90% of the time. This UI is primarily for debugging the webhooks connected to the site. I agree with Derrick that the styling is a bit off – another solution would be to change the title content to indicate user action – e.g. "Click here to view webhooks not connected to the site". 

Also, the "Reset Webhooks" is an action that influences the connected webhook, not the other ones, so it should be placed accordingly – below the first list.

assets/wizards/readerRevenue/views/stripe-setup/index.js
adekbadek: I think a reset button, and a confirmation dialog before webhook deletion and reset will solve these issues.

assets/wizards/readerRevenue/views/stripe-setup/index.js
adekbadek: About the Accordion's styling – WDYT about something akin to the action card?

image

thomasguillot: webhooks i4

how about something like this then.

thomasguillot: I see. So we could move the reset button at the bottom, next to the "Back" button (similar to my previous screen)?

adekbadek: It should be clear from the UI that the reset is about the "connected" section – otherwise the user might think this button resets _all_ of their webhooks.

assets/wizards/readerRevenue/views/stripe-setup/index.js
dkoo: > I think a reset button, and a confirmation dialog before webhook deletion and reset will solve these issues.

I'm happy with this and moving the webhooks to a new sub-page as discussed above. Thanks for the discussion, @adekbadek and @thomasguillot!

assets/wizards/readerRevenue/views/stripe-setup/index.js
thomasguillot: Got it, thanks @adekbadek. Let's go ahead with this:

site

I based the "Reset" button on a small secondary button.

assets/wizards/readerRevenue/views/stripe-setup/index.js
adekbadek: Discussion continued in the [other thread](https://github.com/Automattic/newspack-plugin/pull/2143#discussion_r1036567641).

assets/wizards/readerRevenue/views/stripe-setup/index.js
adekbadek: a433814d61d38cd884c8466f64f2c3eefbabbf91 implements the latest changes. This PR is again ready for review, @dkoo.

assets/wizards/readerRevenue/views/stripe-setup/webhooks-settings.js
```diff
@@ -0,0 +1,193 @@
...
+				title={ () => (
+					<div className="flex justify-between items-center">
+						<h4 className="b f6 ma0">{ __( 'Webhooks connected to this site', 'newspack' ) }</h4>
+						<Button
+							variant="secondary"
+							isSmall
+							onClick={ () => {
+								if (
+									utils.confirmAction(
+										__( 'Are you sure you want to reset all webhooks?', 'newspack' )
```
dkoo, line 162:
```suggestion
										__( 'Are you sure you want to reset all webhooks connected to this site?', 'newspack' )
```

This only deletes and recreates the webhooks connected to the site, right?

assets/wizards/readerRevenue/views/stripe-setup/webhooks-settings.js
```diff
@@ -0,0 +1,193 @@
...
+						>
+							{ __( 'Reset', 'newspack' ) }
+						</Button>
+					</div>
+				) }
+				webhooks={ thisSiteWebhooks }
+				onUpdate={ handleWebhooksListUpdate }
+			/>
+			{ otherWebhooks.length ? (
+				<Accordion title={ __( 'Webhooks not connected to this site.', 'newspack' ) }>
```
dkoo, line 177:
```suggestion
				<Accordion title={ __( 'Webhooks not connected to this site', 'newspack' ) }>
```

OR

```suggestion
				<Accordion title={ __( 'Webhooks connected to other sites', 'newspack' ) }>
```

dkoo (approved):
This is looking great and works well! I have a few extremely minor nitpicks, below. I consider them non-blocking, so pre-approving.

assets/wizards/readerRevenue/views/stripe-setup/webhooks-settings.js
```diff
@@ -0,0 +1,193 @@
...
+					<WebhooksList webhooks={ otherWebhooks } onUpdate={ handleWebhooksListUpdate } />
+				</Accordion>
+			) : null }
+			<div className="newspack-buttons-card">
+				<Button href="#/stripe-setup" isPrimary>
+					{ __( 'Back to Stripe Settings', 'newspack' ) }
```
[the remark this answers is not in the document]
adekbadek: Done in 5ee8a3ebf6faf42f511904a3c92101661631c427! Since this is a small and innocuous change, I'll merge based on the pre-approval.

PR #2144, opened 2022-11-21
Feat/twitter pixel

adekbadek:
### All Submissions:

Adds Twitter pixel setting, building on refactored code from #1994 

1. Go to Newspack > Engagement > Social
2. Activate "Twitter Pixel"
3. Add a pixel ID
4. Save
6. Test enabling and disabling the option
7. Check the copy and the link
8. Visit the site and confirm the pixel was added to the page. You can use a browser extension to help you with that if you want: https://chrome.google.com/webstore/detail/twitter-pixel-helper/jepminnlebllinfmkhfbkpckogoiefpd/related
9. Visit the site with AMP enabled and confirm the pixel is also there
10. I've only tested this with random IDs. It would be awesome if someone could test it with a real pixel ID in a real site

### Other information:


leogermani (changes requested):
When this branch is active, the code for the twitter pixel, and also the code for the Facebook pixel, are added twice in the page. It doesn't happen on `master`.

Other than that, it seems to be working.

If you ask me, I don't think we need the the "Pixel" abstraction.. we could duplicate a little there. But no objections either.

leogermani (approved):
The last commit fixed it!

Nitpick/non-blocker: `print_codes` could have a name that better describes what the method does

PR #2145, opened 2022-11-21
fix(auth): prevent autocomplete for honeypot field

miguelpeixe:
### All Submissions:

In some cases, the honeypot field can be filled with autocomplete suggestions once selected for the actual email field.

This PR ensures that no autocomplete value should be used for this field.

I could not reproduce the issue locally, although it was confirmed with a publisher.

Check out this branch and ensure you can still use autocomplete for signing up with RAS.

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -729,7 +729,7 @@ final class Reader_Activation {
 			$placeholder = __( 'Enter your email address', 'newspack' );
 		}
 		?>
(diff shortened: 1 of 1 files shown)


dkoo (approved):
I can't replicate locally, either, but `autocomplete="off"` should be exactly what we need to prevent autocompletion in that field. https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

PR #2158, opened 2022-11-30
fix: override woo subscriptions login requirement on checkout

dkoo:
### All Submissions:

Fixes a UX conflict between RAS and WooCommerce Subscriptions. The latter requires that any subscription transaction be associated with a logged-in account, and will prompt the user to log in if it finds an existing account. But the former may have created an account upon newsletter signup, resulting in a lot of unnecessary friction on the user's part in order to complete the donation.

This fix checks if the inputted email address is an existing but not logged-in RAS reader, checks if the transaction normally require a login, and if so, temporarily logs in the reader account to allow the transaction to succeed without user input. Once the transaction is complete, the user is logged out.

#### Bonus fix

Also updates the default success message upon registration. On `master`, the default message instructs the reader to check their email inbox for a verification email, but no such email is sent upon registration—it's only sent if the user visits My Account and requests it. The new message instructs the reader to visit My Account to verify and manage their account.

1. On a site using RAS and with Newspack as the Reader Revenue platform, sign up for newsletters via the Subscription Form block, or register via the Register block.
2. In a new session, while still not logged-in, make a recurring donation to get to the checkout form, and fill it in using the same email address you used to register in step 1.
3. On `master`, observe an error message prompting you to log in:

Screen Shot 2022-11-30 at 12 49 55 PM

4. Check out this branch, repeat step 2, and confirm that the transaction succeeds without prompting you to log in.
5. Continue navigating around the site in the same session and confirm that you remain logged out.
6. Also confirm in the dashboard that the subscription is associated with the correct user account.

### Other information:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.97.1-hotfix.2](https://github.com/Automattic/newspack-plugin/compare/v1.97.1-hotfix.1...v1.97.1-hotfix.2) (2022-11-30)
+
+
+### Bug Fixes
+
+* only do temp login if Woo intends to create an account ([1fbe56c](https://github.com/Automattic/newspack-plugin/commit/1fbe56c33c08c3fae714de152a92aaab881e0f9d))
+
 ## [1.97.1-hotfix.1](https://github.com/Automattic/newspack-plugin/compare/v1.97.0...v1.97.1-hotfix.1) (2022-11-30)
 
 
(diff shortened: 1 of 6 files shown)


miguelpeixe:
We might want to revisit this approach sometime soon. There's business logic between `woocommerce_checkout_process` and `woocommerce_checkout_order_processed` and `wp_logout` might never get called.

In that case someone can remain authenticated without providing any credentials.

PR #2160, opened 2022-12-01
feat(reader-revenue): add "other" reader revenue platform

### LINKED ISSUE #2059: Reader Revenue Platform: Provide "Other" option
Currently, it is possible for sites to select "Newspack" as a Platform in Newspack Reader Revenue although it is not being used (due to choosing to go with a third-party option that is not available in the dropdown or not in use whatsoever). Because of this, incorrect data is pulled when reviewing overall Newspack sites that utilize "Newspack" as a reader revenue platform option.

Provide a fourth dropdown option such as "Other" or "Third Party" to clearly identify which sites are not using the other available options.

adekbadek:
### All Submissions:

Closes #2059.

1. Visit the Reader Revenue wizard, observe a new option in the platform selection – "Other" 
2. Select it, observe the only available tab is the "Platform" tab
3. Observe any Donate block added will no be rendered on the front-end

### Other information:

### DIFF
--- a/assets/wizards/readerRevenue/constants.js
+++ b/assets/wizards/readerRevenue/constants.js
@@ -5,5 +5,6 @@
 export const NRH = 'nrh';
 export const NEWSPACK = 'wc';
 export const STRIPE = 'stripe';
+export const OTHER = 'other';
 
 export const READER_REVENUE_WIZARD_SLUG = 'newspack-reader-revenue-wizard';
--- a/assets/wizards/readerRevenue/index.js
+++ b/assets/wizards/readerRevenue/index.js
@@ -16,7 +16,7 @@ import { __ } from '@wordpress/i18n';
  */
 import { Wizard, Notice } from '../../components/src';
 import * as Views from './views';
-import { READER_REVENUE_WIZARD_SLUG, NEWSPACK, NRH, STRIPE } from './constants';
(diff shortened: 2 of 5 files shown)


dkoo (changes requested):
This works in the dashboard, but it might have wider implications. Most notably, it breaks the Donate block, so we might want to have a check for this in the Blocks repo to not render that block on the front-end if "Other" is the active RR platform.

adekbadek:
I was thinking that any user of the "other" platform would maybe implement their own handling of the Donate block. But that's too ambitious of an assumption. 4af139953fb01988df6b36823420a6e2389bc649 updates the code so the Donate block renders blank if the platforms is set to 'other'.

dkoo (changes requested):
I think this not rendering the Donate block is enough to handle this, for now. The only other thing I'd request is some kind of message in the editor to indicate that the block isn't available/won't be shown on the front-end? My crude mockup (the last line is intended to be a link to the RR wizard):

Screen Shot 2022-12-02 at 11 03 06 AM

adekbadek:
Done in https://github.com/Automattic/newspack-blocks/pull/1323 – the error state already in use will be used:

image

dkoo (approved):
Works for me!

PR #2162, opened 2022-12-01
Improved existing Woo user subs handling

claudiulodro:
### All Submissions:

This PR proposes a different approach to the handling that allows existing users to donate without having to log in again. This approach doesn't require logging them in temporarily, so it should be more robust in theory. :)

1. Create a WP user.
2. In an incognito/logged-out window, purchase a donation using WooCommerce and that user's email address during checkout. It should work nicely before and after this PR.

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -85,7 +85,7 @@ final class Reader_Activation {
 			\add_action( 'lostpassword_post', [ __CLASS__, 'set_password_reset_mail_content_type' ] );
 			\add_filter( 'lostpassword_errors', [ __CLASS__, 'rate_limit_lost_password' ], 10, 2 );
 			\add_filter( 'woocommerce_checkout_customer_id', [ __CLASS__, 'associate_existing_woo_users_with_transactions_on_checkout' ] );
-			\add_filter( 'woocommerce_checkout_posted_data', [ __CLASS__, 'dont_force_registration_for_existing_woo_users' ] );
+			\add_filter( 'woocommerce_checkout_posted_data', [ __CLASS__, 'dont_force_registration_for_existing_woo_users' ], 11 );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -84,7 +84,8 @@ public static function init() {
 			\add_filter( 'retrieve_password_notification_email', [ __CLASS__, 'password_reset_configuration' ], 10, 4 );
 			\add_action( 'lostpassword_post', [ __CLASS__, 'set_password_reset_mail_content_type' ] );
 			\add_filter( 'lostpassword_errors', [ __CLASS__, 'rate_limit_lost_password' ], 10, 2 );
-			\add_action( 'woocommerce_checkout_process', [ __CLASS__, 'allow_subscription_purchase_without_login' ] );
+			\add_filter( 'woocommerce_checkout_customer_id', [ __CLASS__, 'associate_existing_woo_users_with_transactions_on_checkout' ] );
+			\add_filter( 'woocommerce_checkout_posted_data', [ __CLASS__, 'dont_force_registration_for_existing_woo_users' ] );
```
dkoo, line 88:
```suggestion
			\add_filter( 'woocommerce_checkout_posted_data', [ __CLASS__, 'dont_force_registration_for_existing_woo_users' ], 11 );
```

includes/reader-activation/class-reader-activation.php
```diff
@@ -1514,32 +1515,37 @@ public static function rate_limit_lost_password( $errors, $user_data ) {
...
+	 *
+	 * @param array $data Array of Woo checkout data.
+	 * @return array Modified $data.
+	 */
+	public static function dont_force_registration_for_existing_woo_users( $data ) {
+		$email    = $data['billing_email'];
+		$customer = \get_user_by( 'email', $email );
+		if ( $customer ) {
+			$data['createaccount'] = 0;
+			\add_filter( 'woocommerce_checkout_registration_required', '__return_false', 99 );
```
dkoo, line 1545: I'm not sure if `9999` is required, but it didn't work with `999`, so I went up another digit.

```suggestion
			\add_filter( 'woocommerce_checkout_registration_required', '__return_false', 9999 );
```

dkoo (changes requested):
I had to do this to get it to work locally:

dkoo (approved):

PR #2163, opened 2022-12-01
fix: fatal if no woo

dkoo:
### All Submissions:

Fixes a fatal error introduced by #2158.

1. On a site with RAS enabled, disable WooCommerce.
2. Visit a page containing a RAS Register block and observe fatal errors.
3. Check out this branch, confirm the fatals are avoided.

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -104,15 +104,18 @@ function render_block( $attrs, $content ) {
 		return '';
 	}
 
-	$registered       = false;
-	$my_account_url   = \wc_get_account_endpoint_url( 'dashboard' );
-	$message          = '';
-	$success_message  = __( 'Thank you for registering!', 'newspack' ) . '<br />';
-	$success_message .= sprintf(
-		// Translators: %s is a link to My Account.
-		__( 'Please visit %s to verify and manage your account.', 'newspack' ),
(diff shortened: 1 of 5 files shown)


assets/blocks/reader-registration/index.php
```diff
@@ -104,15 +104,18 @@ function render_block( $attrs, $content ) {
...
+	$registered      = false;
+	$my_account_url  = function_exists( 'wc_get_account_endpoint_url' ) ? \wc_get_account_endpoint_url( 'dashboard' ) : false;
+	$message         = '';
+	$success_message = __( 'Thank you for registering!', 'newspack' ) . '<br />';
+
+	if ( $my_account_url ) {
+		$success_message .= sprintf(
+			// Translators: %s is a link to My Account.
+			__( 'Please visit %s to verify and manage your account.', 'newspack' ),
+			'<a href="' . esc_url( $my_account_url ) . '">' . __( 'My Account', 'newspack' ) . '</a>'
```
claudiulodro, line 116: Is there ever a situation where a reader will see `Please visit  <a href="">My Account</a>` with no link in the case that `$my_account_url == false`?

claudiulodro (approved):
Good catch! I was able to reproduce the issue and confirm this solves it. Left a question, but it's not a blocker to merge/release.

assets/blocks/reader-registration/index.php
dkoo: There shouldn't be—the link is only appended if `$my_account_url` is not empty.

PR #2173, opened 2022-12-05
feat(data-events): agnostic data integration tool

miguelpeixe:
### All Submissions:

Implements a `Data_Events` class for non-blocking dispatch of data events to registered handlers.

The non-blocking strategy is inspired by [TechCrunch's `wp-async-task`](https://github.com/techcrunch/wp-async-task) and uses an HTTP request to trigger the handlers.

#### Registering an action

To dispatch an event an action must first be registered with the following:

```php
Newspack\Data_Events::register_action( 'action_name' );
```

#### Dispatching

Once registered, an array with arbitrary payload can be dispatched:

```php
$data = [ "test" => "data" ];
$use_client_id = true;
Newspack\Data_Events::dispatch( 'action_name', $data, $use_client_id );
```

The use of client ID, which is `true` by default, will send the `Reader_Activation::get_client_id();` as part of the data event payload. This client ID is pulled from the session's `newspack-cid` cookie. Most data events are expected to be user actions and third parties might make use of having an ID that groups anonymous sessions.

#### Registering handlers vs. WP hooks

There are 2 ways of attaching handlers to events. The primary one, which I'm currently favoring, is through the `::register_handler()` API:

```php
// Action handler
function my_action_handler( $timestamp, $data, $client_id ) {
	// Send data to a third-party.
}
Newspack::register_handler( 'my_action_handler', 'my_action' );

// Global handler, to be called on every data event
function my_global_handler( $action_name, $timestamp, $data, $client_id ) {
	// Do some global analytics.
}
Newspack::register_handler( 'my_global_handler' );
```

They don't have a `priority` argument because they are supposed to be independent. Each relates to its own data integration and is contained in a `try...catch` block so it never disrupts another handler.

A WP action hook can also be used and will be executed after the registered handlers:

```php
add_action( 'newspack_data_event_my_action', 'my_action_handler', 10, 3 );
add_action( 'newspack_data_event', 'my_global_handler', 10, 4 );
```

The advantage of using WP hooks is API familiarity. It's a common API that should be easy enough to use.

The disadvantage is that we might need to adapt how these hooks work in order to make them behave as a data event handler, which may grow in requirements as we start implementing it for our integrations. That's why I'm currently leaning toward our own API but would love some opinion and suggestions on it as well.

**UPDATE**

#### Listeners

After discussing with @leogermani, we decided to also add **listeners**. Listeners serve as a shortcut to dispatch an action once a WP hook is fired. Example:

```php
$action_name = 'wc_new_purchase';
$hook_name   = 'woocommerce_checkout_order_processed';
\Newspack\Data_Events::register_listener( $action_name, $hook_name );
```

With this listener, every registered handler will receive the `wc_new_purchase` data event.

You can also pass a `callable` as the 3rd argument to process/filter the data from the WP hook:

```php
\Newspack\Data_Events::register_listener(
	'wc_new_purchase',
	'woocommerce_checkout_order_processed',
	function ( $data ) {
		// Parse data
		return $data;
	}

);
```

All written tests must pass.

You can also use this small piece of code the test the entire non-blocking flow:

```php
add_action(
	'init',
	function () {
		$action_name = 'my_test_action';
		\Newspack\Data_Events::register_action( $action_name );
		$handler = function ( $timestamp ) {
			sleep( 10 );
			error_log( date( 'Y/m/d H:i:s', $timestamp ) );
		};
		\Newspack\Data_Events::register_handler( $handler, $action_name );
		\Newspack\Data_Events::dispatch( $action_name, [ 'foo' => 'bar' ] );
	}
);
```

The handler's `error_log()` should run 10 seconds after the `init` without blocking the page execution. 

### Other information:

### DIFF
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -95,6 +95,8 @@ final class Data_Events {
 	public static function register_action( $action_name ) {
 		if ( ! isset( self::$actions[ $action_name ] ) ) {
 			self::$actions[ $action_name ] = [];
+		} else {
+			return new WP_Error( 'action_already_registered', __( 'Action already registered.', 'newspack' ) );
 		}
 	}
 
@@ -104,11 +106,17 @@ final class Data_Events {
 	 * @param string   $action_name Action name.
 	 * @param callable $handler     Action handler.
 	 *
-	 * @return void|WP_Error Error if action not registered.
+	 * @return void|WP_Error Error if action not registered, handler already reigstered or is not callable.
 	 */
-	public static function register_action_handler( $action_name, $handler ) {
+	public static function register_handler( $action_name, $handler ) {
 		if ( ! isset( self::$actions[ $action_name ] ) ) {
 			return new WP_Error( 'action_not_registered', __( 'Action not registered.', 'newspack' ) );
+		}
+		if ( ! is_callable( $handler ) ) {
+			return new WP_Error( 'handler_not_callable', __( 'Handler is not callable.', 'newspack' ) );
+		}
+		if ( in_array( $handler, self::$actions[ $action_name ], true ) ) {
+			return new WP_Error( 'handler_already_registered', __( 'Handler already registered.', 'newspack' ) );
 		}
 		self::$actions[ $action_name ][] = $handler;
 	}
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -97,9 +97,27 @@ final class Data_Events {
 			}
 		}
 
-		// Execute WP hooks.
-		do_action( self::ACTION, $action_name, $timestamp, $data, $client_id );
-		do_action( self::ACTION . '_' . $action_name, $timestamp, $data, $client_id );
+		/**
+		 * Fires after all global and action-specific handlers have been executed.
+		 *
+		 * @param string $action_name Action name.
+		 * @param int    $timestamp   Timestamp.
+		 * @param mixed  $data        Data.
+		 * @param string $client_id   Client ID.
+		 */
+		do_action( 'newspack_data_event', $action_name, $timestamp, $data, $client_id );
+
+		/**
+		 * Fires after all global and action-specific handlers have been executed.
+		 *
+		 * The dynamic portion of the hook name, `$action_name`, refers to th name
+		 * of the action being fired.
+		 *
+		 * @param int    $timestamp   Timestamp.
+		 * @param mixed  $data        Data.
+		 * @param string $client_id   Client ID.
+		 */
+		do_action( 'newspack_data_event_' . $action_name, $timestamp, $data, $client_id );
 	}
 
 	/**
@@ -196,8 +214,29 @@ final class Data_Events {
 			$client_id = Reader_Activation::get_client_id();
 		}
 
+		/**
+		 * Fires when and action is dispatched. This occurs before any handlers are
+		 * executed.
+		 *
+		 * @param string $action_name Action name.
+		 * @param int    $timestamp   Timestamp.
+		 * @param mixed  $data        Data.
+		 * @param string $client_id   Client ID.
+		 */
 		do_action( 'newspack_data_event_dispatch', $action_name, $timestamp, $data, $client_id );
 
+		/**
+		 * Fires when and action is dispatched. This occurs before any handlers are
+		 * executed.
+		 *
+		 * The dynamic portion of the hook name, `$action_name`, refers to th name
+		 * of the action being fired.
+		 *
+		 * @param string $action_name Action name.
+		 * @param int    $timestamp   Timestamp.
+		 * @param mixed  $data        Data.
+		 * @param string $client_id   Client ID.
+		 */
 		do_action( "newspack_data_event_dispatch_{$action_name}", $timestamp, $data, $client_id );
 
 		$url = \add_query_arg(
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -19,6 +19,11 @@ final class Data_Events {
 	const ACTION = 'newspack_data_event';
 
 	/**
+	 * Header to be used while logging.
+	 */
+	const LOGGER_HEADER = 'NEWSPACK-DATA-EVENTS';
+
+	/**
 	 * Registered callable handlers, keyed by their action name.
 	 *
 	 * @var callable[]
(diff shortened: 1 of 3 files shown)


leogermani:
Nice! Good job. Things look clean and straight forward. I will run some tests today, but I'm already leaving a couple of minor comments in the code

includes/data-events/class-data-events.php
```diff
@@ -0,0 +1,230 @@
...
+			try {
+				call_user_func( $handler, $timestamp, $data, $client_id );
+			} catch ( \Throwable $e ) {
+				// Catch fatal errors so it doesn't disrupt other handlers.
+				Logger::error( $e->getMessage() );
+			}
+		}
+
+		// Execute WP hooks.
+		do_action( self::ACTION, $action_name, $timestamp, $data, $client_id );
```
leogermani, line 101: We should add a doc block for each action.

Also, I wouldn't make the action name dynamic like that. If we ever want to change the class constant for any reason this would break any consumer of this library using the hook previous names...

Also, it makes it slightly harder for a programmer to find where this action is being fired from (you could workaround it by adding the name to the comment to make it discoverable...) But I think in general it's not a good thing to make the whole action name dynamic.

includes/data-events/class-data-events.php
```diff
@@ -0,0 +1,230 @@
...
+			return new WP_Error( 'newspack_data_events_action_not_registered', __( 'Action not registered.', 'newspack' ) );
+		}
+
+		$timestamp = time();
+		$client_id = null;
+		if ( $use_client_id ) {
+			$client_id = Reader_Activation::get_client_id();
+		}
+
+		do_action( 'newspack_data_event_dispatch', $action_name, $timestamp, $data, $client_id );
```
leogermani, line 199: Do we have any immediate use for these hooks? If not, I'd say drop them... this is precisely the function we want to execute fast and adding these hooks here gives room for developers to defeat this purpose.

If we do have a use for these hooks, document them and add this warning that any heavy load should be done in the action handler and not here

includes/data-events/class-data-events.php
```diff
@@ -0,0 +1,230 @@
...
+		];
+
+		return \wp_remote_post(
+			$url,
+			[
+				'timeout'   => 0.01,
+				'blocking'  => false,
+				'body'      => $body,
+				'cookies'   => $_COOKIE, // phpcs:ignore
+				'sslverify' => false,
```
leogermani, line 225: WP core offers a filter for people to change it and we could do the same... this could probably be

`'sslverify' => apply_filters( 'https_local_ssl_verify', false ),`

includes/data-events/class-data-events.php
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	/**
+	 * Maybe handle an event.
+	 */
+	public static function maybe_handle() {
+		// Don't lock up other requests while processing.
+		session_write_close(); // phpcs:ignore
+
+		if ( ! isset( $_REQUEST['nonce'] ) || ! \wp_verify_nonce( \sanitize_text_field( $_REQUEST['nonce'] ), self::ACTION ) ) {
```
leogermani, line 50: You can use `check_ajax_referrer` here. a bit simpler -> https://developer.wordpress.org/reference/functions/check_ajax_referer/

includes/data-events/class-data-events.php
miguelpeixe: Good call! 61ae87cc7de97b807a3547dbf204270b00d1c95d improves readability and docs for all WP hooks.

includes/data-events/class-data-events.php
miguelpeixe: This is meant for the extensibility of the tool, not as an actual third-party action handler. There are 2 use cases off the top of my head, but in terms of "extensibility" _anything_ is possible 😅 

1. The webhooks functionality will have its own async handling, because of its particular error-handling flow. It will very likely hook into the dispatch so it creates webhook requests accordingly.
2. If we ever decide to print front-end js for pushing analytics events, this would have to be in the dispatch execution. Example of a very hypothetical code:

```php
add_action(
	'newspack_data_event_dispatch',
	function( $action_name, $timestamp, $data, $client_id ) {
		add_action(
			'wp_footer',
			function() {
				echo '<script>dataLayer.push({"event": "' . esc_attr( $action_name ) . '"});</script>';
			}
		)
	}
	10,
	4
);
```

includes/data-events/class-data-events.php
miguelpeixe: Done in 831f51be3841cad0be6ecaf1a901d48e5789db5f

includes/data-events/class-data-events.php
miguelpeixe: I opted for a more verbose approach for readability. That function feels like an unnecessary shortcut.

includes/data-events/class-data-events.php
leogermani: For number 1, is there any reason why webhooks couldn't do their thing in the handler? It would trigger its async call from an async call...

miguelpeixe: This is something yet to experiment with actual code, but since a request is going to be a `post`, it's more suitable to use its scheduling functionality for this. Each registered webhook endpoint will create a request, which will be rescheduled to send the data until it returns `2xx`.

It feels unnecessary to hook the creation of the request as an async handler if the handling is not happening at this point.

miguelpeixe:
Updated the PR description to include listeners:

Listeners serve as a shortcut to dispatch an action once a WP hook is fired. Example:

```php
$action_name = 'wc_new_purchase';
$hook_name   = 'woocommerce_checkout_order_processed';
\Newspack\Data_Events::register_listener( $action_name, $hook_name );
```

With this listener, every registered handler will receive the `wc_new_purchase` data event.

You can also pass a `callable` as the 3rd argument to process/filter the data from the WP hook:

```php
\Newspack\Data_Events::register_listener(
	'wc_new_purchase',
	'woocommerce_checkout_order_processed',
	function ( $data ) {
		// Parse data
		return $data;
	}

);
```

cc @leogermani

leogermani (approved):
Working as expected!

Let's start to draft a README page for this in following PRs.

Also, I suggest to add a bunch more of `Logger` calls to help us debug this. And I think we can add a new method or option to the logger to prefix the message with the process id (`getmypid()`). This allows us to look at the log and be able to identify which messages are being sent from the same requests or new requests.

miguelpeixe:
> Let's start to draft a README page for this in following PRs.

Good idea, where do you suggest this to live?

> Also, I suggest to add a bunch more of Logger calls to help us debug this.

Perhaps inside `dispatch()`? Not sure where else it's actually useful.

leogermani:
> > Let's start to draft a README page for this in following PRs.
> 
> Good idea, where do you suggest this to live?

The `includes/data-events` folder is a good start`. But we can do this in follow up PRs.

> > Also, I suggest to add a bunch more of Logger calls to help us debug this.
> 
> Perhaps inside `dispatch()`? Not sure where else it's actually useful.

Don't worry, add them as needed when you start debugging and feel the need

miguelpeixe:
1c8eedf7be6e704d1da67481a3cd995be691ed60 improved the logging a bit. Should be enough for now. Logging registration of actions or listeners really floods the log.

Tweaked the listener a bit on d151224d2c85a95ab5306b050b81a5557cf507c5. The `dispatch()` should happen last so that any transformed data along the way is captured.

> The includes/data-events folder is a good start`. But we can do this in follow up PRs.

Would we be deploying `.md` files or our scripts clean this up before packaging?

leogermani:
> Would we be deploying .md files or our scripts clean this up before packaging?

Not aware of any cleaning script. I don't think it's a big deal, but we can also create a docs folder in the plugin root and make sure not to ship it. But I like the README being alongside the code so it's easy to find

miguelpeixe:
> I like the README being alongside the code so it's easy to find

I also think it's a good home for documentation.

PR #2178, opened 2022-12-07
Hotfix(stripe): force sync'd subscriptions to manual renewal

adekbadek:
### All Submissions:

Fixes an issue with Stripe to WC Subscriptions sychronization. 

1. On `master`
1. Enable RAS, set the Reader Revenue platform to Stripe
2. Make a recurring donation as user A
3. Switch the Reader Revenue platform to Newspack 
4. Log in to `/my-account` as user A
5. Go to "Payment Methods" and add a credit card
6. Check "Update the payment method used for all of my active subscriptions"
7. Now WC Subs will update the payment method on the WC Subscription to the new payment method, because there were no payment methods for this customer before
8. Switch to this branch and repeat the test steps – observe the payment method is not updated on the WC Subscription

### Other information:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.97.3-hotfix.1](https://github.com/Automattic/newspack-plugin/compare/v1.97.2...v1.97.3-hotfix.1) (2022-12-07)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


miguelpeixe (changes requested):
How do I see that the subscription is going to use the newly registered payment method?

This is the result of my subscription on `master`, after following the described steps:

image

Doesn't look like it picked up the credit card added.

adekbadek:
I've forgotten to list this crucial step (when adding the payment method):

image

I'll update the testing steps.

miguelpeixe (approved):
Working as described!

PR #2181, opened 2022-12-09
chore(admin): bundle tachyons with WP assets instead of fetching from…

adekbadek:
Small tweak – the tachyons CSS library is fetched from unpkg.com. This service might take longer to respond, or be unavailable – it's faster and more reliable to fetch the library from NPM and serve it locally. 

## To test

Go to Analytics wizard -> Custom dimensions, and observe that `flex` and `justify-end` classes are effective and "defined" in the CSS:

image

### DIFF
--- a/assets/admin/index.js
+++ b/assets/admin/index.js
@@ -1,0 +1,1 @@
+import './style.scss';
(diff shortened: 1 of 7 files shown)


miguelpeixe (approved):
Works as described

PR #2182, opened 2022-12-09
feat(data-events): webhooks

miguelpeixe:
### All Submissions:

As part of the core Data Events system (#2173), webhooks are responsible for integrating the data events to any API that is capable of interpreting the payload.

### Requests

The webhook requests are registered as a custom post type. This allows flexibility to log, query, and more efficiently manage the relationship with registered endpoints.

Once a data event is dispatched, webhook requests may be created if there are endpoints registered to receive that type of data.

Requests make use of WP's post scheduling in order to send a request. Once a request is created, its send is scheduled for 1 minute from now. Every webhook request is expected to have a slight delay but the original timestamp is preserved in the payload.

### Endpoints

A custom taxonomy for registering the webhook URLs and their configuration.

In the next PR, the publisher will be able to register new webhook URLs through the “Connections” wizard. A section at the bottom of the wizard will allow the entire user-facing management:

- List existing endpoints
- Display the latest request response code (endpoint health)
- Enable/disable endpoints
- Add and edit endpoints with their URL, whether it's global or select the data events actions it's supposed to listen to

If we wish to, it’s also possible to display the log of the most recent requests per endpoint.

### Handling request response codes

The registered endpoints are prone to unexpected errors and we should allow the client to fix them while preserving data integrity.

Until the server receives a 2xx response from the client, it attempts using an exponential backoff (2<sup>n</sup> minutes intervals) maxed to 12 hours. On the 15th attempt, an error will kill the request and deactivate the endpoint.

### Cron events

- Twice a day every finished request older than 7 days is permanently deleted
- Every hour ensure to send any request that should've been sent according to its scheduling but it was not

### Tweaks to the data events API

The 3rd argument of `register_listener()` can now be a `callable` **or an `array`**. The array should be a list of strings to be mapped with the hook's arguments, e.g.:

https://github.com/Automattic/newspack-plugin/blob/87b628be7463ddaff2e15d873a779bfdc8ade462/includes/data-events/listeners.php#L79-L86

A listener also no longer dispatches an action if the resulting data is empty.

### Inaugurating listeners

This PR introduces common listeners to be registered at [`includes/data-events/listeners.php`](https://github.com/Automattic/newspack-plugin/blob/feat/data-events-webhooks/includes/data-events/listeners.php). They create the following actions:

 - reader_registered
 - reader_logged_in
 - newsletter_subscribed
 - newsletter_updated

Confirm all written tests are passing.

1. Visit this site to mock a webhook receiver: https://webhook.site/ and leave this page open
2. Open your site shell with `$ wp shell`:
   1. Copy the endpoint URL and place it in a variable: `$url = 'https://webhook.site/{some-id-here}';`
   2. Create a global endpoint with: `Newspack\Data_Events\Webhooks::create_endpoint( $url, [], true );`
   3. Confirm the endpoint is created by the return of an `int`, which is the endpoint ID
3. Confirm you have RAS active, AMP Plus and newspack log enabled:
```php
```
4. Visit your site in incognito and go through the reader registration flow
5. Confirm the `NEWSPACK-DATA-EVENTS` and `NEWSPACK-WEBHOOKS` logs:
```
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "reader_logged_in".
[NEWSPACK-WEBHOOKS][Newspack\Data_Events\Webhooks::schedule_request]: Scheduling request 346 for 2022-12-14 13:24:23.
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::handle]: Executing global action handlers for "reader_logged_in".
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::handle]: Executing action handlers for "reader_logged_in".
```
6. Wait one minute until the scheduled request is sent and confirm you than see the following log:
```
[NEWSPACK-WEBHOOKS][Newspack\Data_Events\Webhooks::send_request]: Sending request 346
[NEWSPACK-WEBHOOKS][Newspack\Data_Events\Webhooks::finish_request]: Finishing request 346.
```
7. The https://webhook.site page should be updated with this:
Captura de Tela 2022-12-14 às 10 26 30

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -79,6 +79,7 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/class-admin-plugins-screen.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-data-events.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-webhooks.php';
+		include_once NEWSPACK_ABSPATH . 'includes/data-events/listeners.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-api.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-profile.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-analytics.php';
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -49,6 +49,7 @@ final class Newspack {
 		add_action( 'network_admin_notices', [ $this, 'remove_notifications' ], -9999 );
 		add_action( 'all_admin_notices', [ $this, 'remove_notifications' ], -9999 );
 		register_activation_hook( NEWSPACK_PLUGIN_FILE, [ $this, 'activation_hook' ] );
+		register_deactivation_hook( NEWSPACK_PLUGIN_FILE, [ $this, 'deactivation_hook' ] );
 	}
 
 	/**
@@ -235,6 +236,14 @@ final class Newspack {
 	 */
 	public function activation_hook() {
 		set_transient( NEWSPACK_ACTIVATION_TRANSIENT, 1, 30 );
+		do_action( 'newspack_activation' );
+	}
+
+	/**
+	 * Deactivation Hook
+	 */
+	public function deactivation_hook() {
+		do_action( 'newspack_deactivation' );
 	}
 
 	/**
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -246,6 +246,9 @@ final class Newspack {
 	 * Deactivation Hook
 	 */
 	public function deactivation_hook() {
+		/**
+		 * Fires on the newspack plugin deactivation hook
+		 */
 		do_action( 'newspack_deactivation' );
 	}
 
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -78,6 +78,7 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/class-theme-manager.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-admin-plugins-screen.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-data-events.php';
+		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-webhooks.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-api.php';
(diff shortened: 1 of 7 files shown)


includes/data-events/class-webhooks.php
```diff
@@ -0,0 +1,615 @@
...
+			'clear_finished'     => 'twicedaily',
+			'send_late_requests' => 'hourly',
+		];
+	}
+
+	/**
+	 * Register webhook cron events.
+	 */
+	public static function register_cron_events() {
+		\register_deactivation_hook( __FILE__, [ __CLASS__, 'clear_cron_events' ] );
```
leogermani, line 121: non blocking: is this the best place for this? I think for deactivation hook it's better something that's always ran when you include the main plugin file, without conditionals. Even if there's nothing to cleanup, I think that's fine

leogermani (approved):
It tests well and looks good. I have some considerations on how we are relying on wp cron here but let's discuss it elsewhere

includes/data-events/class-webhooks.php
miguelpeixe: Great point! I don't think we can register multiple activation and deactivation hooks like that.

72188207f86c29fc33f786c7a652e544388030f5 creates a centralized hook on the main file. You can test it with:

```php
add_action( 'newspack_activation', function() { error_log( 'activating newspack' ); } );
add_action( 'newspack_deactivation', function() { error_log( 'deactivating newspack' ); } );
```

Deactivate, activate and confirm the logs.

includes/class-newspack.php
```diff
@@ -233,6 +236,14 @@ public function remove_notifications() {
 	 */
 	public function activation_hook() {
 		set_transient( NEWSPACK_ACTIVATION_TRANSIENT, 1, 30 );
+		do_action( 'newspack_activation' );
```
leogermani, line 239:
```suggestion
		/**
		 * Fires on the newspack plugin activation hook
		 */
		do_action( 'newspack_activation' );
```

includes/class-newspack.php
```diff
@@ -233,6 +236,14 @@ public function remove_notifications() {
...
+	}
+
+	/**
+	 * Deactivation Hook
+	 */
+	public function deactivation_hook() {
+		do_action( 'newspack_deactivation' );
```
leogermani, line 246:
```suggestion
		/**
		 * Fires on the newspack plugin deactivation hook
		 */
		do_action( 'newspack_deactivation' );
```

PR #2209, opened 2022-12-21
fix(stripe): webhooks list edge-cases

adekbadek:
### All Submissions:

Fixes a webhook configuration edge case and updates the UI text to reflect the fact that a webhook might be connected to the site, but is used for a different integration. 

1. Create a new webhook via Stripe dashboard – same as the existing Newspack webhook, but with different protocol (`https` -> `http` or the other way around)
2. On `master`, observe that the duplicated webhook is not listed in `/wp-admin/admin.php?page=newspack-reader-revenue-wizard#/stripe-webhooks` as connected to this site
3. Switch to this branch, observe the webhook is listed and the warning about duplicate webhook is displayed

### Other information:

### DIFF
--- a/assets/wizards/readerRevenue/views/stripe-setup/webhooks-settings.js
+++ b/assets/wizards/readerRevenue/views/stripe-setup/webhooks-settings.js
@@ -152,7 +152,9 @@ const WebhooksSettings = () => {
 				className="mb4"
 				title={ () => (
 					<div className="flex justify-between items-center">
-						<h4 className="b f6 ma0">{ __( 'Webhooks connected to this site', 'newspack' ) }</h4>
+						<h4 className="b f6 ma0">
+							{ __( 'Webhooks connected to Newspack on this site', 'newspack' ) }
+						</h4>
 						<Button
 							variant="secondary"
 							isSmall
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Non-blocking: I don't think we need to change the message from `Webhooks connected to this site`. I think the new message adds another layer of potential confusion—what does "to Newspack" mean?

Approving because it works well and warns about the extra webhooks.

adekbadek:
> what does "to Newspack" mean?

I've observed on a live site that there was another webhook connected to the site present – it was for some other integration, not Newspack. Such a webhook would not be listed in the "Webhooks connected to this site" even though it's _connected to this site_. Hence the change. But maybe that's being overly exact? WDYT?

dkoo:
> I've observed on a live site that there was another webhook connected to the site present – it was for some other integration, not Newspack. Such a webhook would not be listed in the "Webhooks connected to this site" even though it's connected to this site. Hence the change. But maybe that's being overly exact? WDYT?

Hmm, that's an interesting edge case. I'd argue that the webhooks we're showing in the Newspack UI are implicitly connected to Newspack, especially if we're not showing other webhooks _not_ connected to Newspack. So the "to Newspack" would be unnecessary, IMO.

adekbadek:
> webhooks we're showing in the Newspack UI are implicitly connected to Newspack

That's a good point! Reverted the copy change in 5e31c0be14d89a5d32bd79f4fe805171a08cd88e

PR #2215, opened 2023-01-04
docs: data events

miguelpeixe:
### All Submissions:

Adds technical documentation to `\Newspack\Data_Events`.

Navigate to [this branch's `data-events` directory](https://github.com/Automattic/newspack-plugin/tree/docs/data-events/includes/data-events) and review the contents and formatting of the new readme file.

### Other information:


includes/data-events/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+## A tool for non-blocking dispatch of data events to registered handlers.
+
+The purpose of this tool is to unify the strategy for sending reader activity data to third-party services, focused primarily for analytics, ESPs and CRMs.
+
+The non-blocking strategy is inspired by [TechCrunch's `wp-async-task`](https://github.com/techcrunch/wp-async-task) and uses an HTTP request to trigger the handlers.
+
+---
+
+## Core Actions
```
leogermani, line 11: Maybe call it "Newspack core" action so we avoid confusion with WP core's actions

includes/data-events/README.md
```diff
@@ -0,0 +1,162 @@
+# Newspack Data Events
+
+## A tool for non-blocking dispatch of data events to registered handlers.
+
+The purpose of this tool is to unify the strategy for sending reader activity data to third-party services, focused primarily for analytics, ESPs and CRMs.
+
+The non-blocking strategy is inspired by [TechCrunch's `wp-async-task`](https://github.com/techcrunch/wp-async-task) and uses an HTTP request to trigger the handlers.
+
+---
```
leogermani, line 9: I miss an "overview" section here, presenting the basic concepts and stakeholders...

events are triggered somewhere in the application
we capture them, dispatch and handlers handle them... we can have multiple handlers like Mailchimp or Salesforce, etc.

I started to draft this but then I realized we might have to adjust something for clarity. I see we are using both "action" and "event" to refer to the same thing. Or maybe I got this wrong, and "action" is what happens in the application and "event" is what we dispatch based on an action? in any case, it does get a bit confusing without the overview (and extra confusing when we introduce Listeners that listen to wp core's action hooks)

What do you think?

includes/data-events/README.md
leogermani: Add: These events are registered by the Newspack Plugin and present by default

miguelpeixe: 428e0663305a143bc88ef92f46933035ae852ff7 updates this section

includes/data-events/README.md
miguelpeixe: I've added a chart to illustrate the flow and stakeholders.

>  I see we are using both "action" and "event" to refer to the same thing.

I hope the image clarifies that. The action is how we name the group of events about the same thing. "reader_registered", for example, is an action that occurs on site. Each time this action happens, a dispatch with the action arguments (payload) forms an event. Sort of like:

```
action -> dispatch(arguments) -> event
```

leogermani (approved):

PR #2224, opened 2023-01-11
Remove donation products from cart if WC is not the donations platform

adekbadek:
### All Submissions:

If the Reader Revenue platform is not set to Newspack (WooCommerce), but the WC suite is active, "buying" the donation products using WC UI should not be possible. 

See 1203536997639429-as-1203626633832534

1. Set the Reader Revenue platform to Newspack 
2. Complete the donation, via the Donate block or by directly buying using a WC product link (e.g. `/product/donate-one-time/`
3. Switch the Reader Revenue platform to a different one
4. Visit the WC product page, add it to cart
5. Navigate to the cart, observe it is empty

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -69,6 +69,7 @@ class Donations {
 			add_action( 'woocommerce_checkout_update_order_meta', [ __CLASS__, 'woocommerce_checkout_update_order_meta' ] );
 			add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'woocommerce_billing_fields' ] );
 			add_filter( 'pre_option_woocommerce_enable_guest_checkout', [ __CLASS__, 'disable_guest_checkout' ] );
+			add_action( 'template_redirect', [ __CLASS__, 'handle_cart' ] );
 			add_filter( 'amp_skip_post', [ __CLASS__, 'should_skip_amp' ], 10, 2 );
 		}
 	}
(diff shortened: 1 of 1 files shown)


includes/class-donations.php
```diff
@@ -848,5 +849,26 @@ public static function can_use_streamlined_donate_block() {
...
+	 * If WC is not the donations platform, the donation products should not be buyable.
+	 */
+	public static function handle_cart() {
+		if ( ! self::is_woocommerce_suite_active() ) {
+			return;
+		}
+		if ( self::is_platform_wc() ) {
+			return;
+		}
+		if ( \is_cart() || \is_checkout() ) {
```
leogermani, line 864: Since this is running in every request, I was wondering what would be the way to impact the less on performance.

I guess if we first checked for `is_cart() || is_checkout` and do the other 2 checks after that we will save some processing cycles, wouldn't we? we would only check those things if we are on the right page... WDYT?

leogermani (approved):
Works as described.

I left a comment to ask you about the order of the "ifs" and if we could make it in a way that will have the less impact possible

includes/class-donations.php
adekbadek: I'm not sure, actually. `is_cart` and `is_checkout` are the "heaviest" checks – both read an option and check for shortcodes in page, among other things. On the other hand, if we assume that most sites have WC active, the first check would be passed in most cases. Overall, I don't have a strong opinion on this kind of micro-optimisation.

leogermani: yeah it's hard to know. Best case scenario would be to use a hook that is fired only on checkout and cart pages. Have you considered using `woocommerce_before_checkout_form` and `woocommerce_before_cart`?

I just like to avoid having thins running in each and every request, but feel free to go ahead and merge it if you don't think this is a problem

adekbadek: Great suggestion! Done in 27e8095fc87aadb59fa2e46ba655fa27d1fc0262

leogermani:
It works but with a small glitch. When I go to my cart, I see an empty cart:

Captura de tela de 2023-01-12 08-47-53

Which is kind of a broken cart. If a refresh the page, I see the proper empty cart message:

Captura de tela de 2023-01-12 08-48-02

That's because we already loaded the `cart` template at this point, and not the `cart-empty` one.

There's one hook that looks better and precisely designed for this -> https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/shortcodes/class-wc-shortcode-cart.php#L91

`woocommerce_check_cart_items`

I've just tested it and it works fine.

adekbadek:
Thanks for the tip! Updated in a2d38d898e4760828e8fcc70cea0d9a8730c1c52

leogermani (approved):

PR #2225, opened 2023-01-11
feat(data-events): donation listeners

miguelpeixe:
### All Submissions:

This PR tweaks the hook into `woocommerce_checkout_order_processed` so it's more easily accessible for other parts of the code that might want to reach donation processing. There shouldn't be any change into how the `sync_reader_from_order()` behaves.

The following donation-related listeners are added:

### `donation_new`

When there's a new donation, either through Stripe or Newspack (WooCommerce) platforms.

### Data

| Name            | Type     |
| --------------- | -------- |
| `user_id`       | `int`    |
| `email`         | `string` |
| `amount`        | `float`  |
| `currency`      | `string` |
| `recurrence`    | `string` |
| `platform`      | `string` |
| `platform_data` | `array`  |

### `donation_subscription_new`

When there's a new WooCommerce Subscription. This action does not replace the `donation_new` that create the subscription.

| Name            | Type     |
| --------------- | -------- |
| `user_id`       | `int`    |
| `email`         | `string` |
| `amount`        | `float`  |
| `currency`      | `string` |
| `recurrence`    | `string` |
| `platform`      | `string` |
| `platform_data` | `array`  |

### `donation_subscription_cancelled`

When a WooCommerce Subscription is cancelled.

| Name              | Type     |
| ----------------- | -------- |
| `subscription_id` | `int`    |
| `user_id`         | `int`    |
| `email`           | `string` |
| `amount`          | `float`  |
| `currency`        | `string` |
| `recurrence`      | `string` |
| `platform`        | `string` |

### `donation_subscription_changed`

When a WooCommerce Subscription status changes.

| Name              | Type     |
| ----------------- | -------- |
| `subscription_id` | `int`    |
| `user_id`         | `int`    |
| `email`           | `string` |
| `status_before`   | `string` |
| `status_after`    | `string` |
| `amount`          | `float`  |
| `currency`        | `string` |
| `recurrence`      | `string` |
| `platform`        | `string` |

1. Make sure you have the following constants:
```
```
2. Make sure you have the Reader Revenue Newspack (WooCommerce) platform configured
3. Make a new one-time donation using the donation block
4. Confirm you see the logs with `[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "donation_new" via Action Scheduler.`
5. Visit WooCommerce's Status -> Scheduled Actions
6. Confirm you see the `newspack_data_events_as_dispatch ` hook with `donation_new` as well
7. Now make a recurring donation (subscription) and confirm you see the same with: `donation_subscription_new`, `donation_subscription_changed` (switching from `pending` to `active`)
8. WIth Reader Activation enabled, make sure you have ActiveCampaign configured with a master list
9. Make a donation with a new email and confirm the contact data with donation details is being updated to ActiveCampaign as expected
10. Switch the platform to Stripe and repeats steps from 3 to 7

### Other information:

### DIFF
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -109,6 +109,10 @@ When a WooCommerce Subscription is cancelled.
 | `subscription_id` | `int`    |
 | `user_id`         | `int`    |
 | `email`           | `string` |
+| `amount`          | `float`  |
+| `currency`        | `string` |
+| `recurrence`      | `string` |
+| `platform`        | `string` |
 
 ### `donation_subscription_changed`
 
@@ -121,6 +125,10 @@ When a WooCommerce Subscription status changes.
 | `email`           | `string` |
 | `status_before`   | `string` |
 | `status_after`    | `string` |
+| `amount`          | `float`  |
+| `currency`        | `string` |
+| `recurrence`      | `string` |
+| `platform`        | `string` |
 
 ## Registering a new action
 
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -80,6 +80,58 @@ When a reader updates their lists subscription from Newspack Newsletters.
 | `lists_added`   | `string[]` |
 | `lists_removed` | `string[]` |
 
+### `donation_new`
+
+When there's a new donation, either through Stripe or Newspack (WooCommerce) platforms.
+
+### Data
+
+| Name            | Type     |
+| --------------- | -------- |
+| `user_id`       | `int`    |
+| `email`         | `string` |
+| `amount`        | `float`  |
+| `currency`      | `string` |
+| `recurrence`    | `string` |
+| `platform`      | `string` |
+| `platform_data` | `array`  |
+
+### `donation_subscription_new`
+
+When there's a new WooCommerce Subscription. This action does not replace the `donation_new` that create the subscription.
+
+| Name            | Type     |
+| --------------- | -------- |
+| `user_id`       | `int`    |
+| `email`         | `string` |
+| `amount`        | `float`  |
+| `currency`      | `string` |
+| `recurrence`    | `string` |
+| `platform`      | `string` |
+| `platform_data` | `array`  |
+
+### `donation_subscription_cancelled`
+
+When a WooCommerce Subscription is cancelled.
+
+| Name              | Type     |
+| ----------------- | -------- |
+| `subscription_id` | `int`    |
+| `user_id`         | `int`    |
+| `email`           | `string` |
+
+### `donation_subscription_changed`
+
+When a WooCommerce Subscription status changes.
+
+| Name              | Type     |
+| ----------------- | -------- |
+| `subscription_id` | `int`    |
+| `user_id`         | `int`    |
+| `email`           | `string` |
+| `status_before`   | `string` |
+| `status_after`    | `string` |
+
 ## Registering a new action
 
 To dispatch an event, an action must first be registered with the following:
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -74,8 +74,6 @@ When a reader updates their lists subscription from Newspack Newsletters.
 
 When there's a new donation, either through Stripe or Newspack (WooCommerce) platforms.
 
-### Data
-
 | Name            | Type     |
 | --------------- | -------- |
 | `user_id`       | `int`    |
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -26,8 +26,6 @@ These events are registered by the Newspack Plugin and available by default.
 
 When a reader registers.
 
-#### Data
-
 | Name       | Type      |
 | ---------- | --------- |
 | `user_id`  | `integer` |
@@ -38,8 +36,6 @@ When a reader registers.
 
 When a reader authenticates.
 
-#### Data
-
 | Name      | Type      |
 | --------- | --------- |
 | `user_id` | `integer` |
@@ -49,8 +45,6 @@ When a reader authenticates.
 
 When a reader verifies their email address.
 
-#### Data
-
 | Name      | Type      |
 | --------- | --------- |
 | `user_id` | `integer` |
@@ -58,8 +52,6 @@ When a reader verifies their email address.
 ### `newsletter_subscribed`
 
 When a reader subscribes to newsletter lists from Newspack Newsletters subscription.
-
-#### Data
 
 | Name       | Type       |
 | ---------- | ---------- |
@@ -70,8 +62,6 @@ When a reader subscribes to newsletter lists from Newspack Newsletters subscript
 ### `newsletter_updated`
 
 When a reader updates their lists subscription from Newspack Newsletters.
-
-#### Data
 
 | Name            | Type       |
 | --------------- | ---------- |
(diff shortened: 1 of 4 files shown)


includes/data-events/listeners.php
```diff
@@ -100,3 +100,104 @@ function( $provider, $email, $lists_added, $lists_removed, $result ) {
...
+);
+
+/**
+ * For when there's a new donation subscription.
+ *
+ * This will be fetched from a new donation, so we're hooking into the 'donation_new' dispatch.
+ */
+Data_Events::register_listener(
+	'newspack_data_event_dispatch_donation_new',
+	'donation_subscription_new',
```
dkoo, line 161: Testing these steps:

> 3. Make a new one-time donation using the donation block
> 4. Confirm you see the logs with [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "donation_new" via Action Scheduler.

Making a one-time donation, I see both `donation_subscription_new` and `donation_new` actions being dispatched.

```
[12-Jan-2023 23:31:07 UTC] [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "donation_subscription_new" via Action Scheduler.
[12-Jan-2023 23:31:07 UTC] [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "donation_new" via Action Scheduler.
```

I also see this in the Scheduled Actions log:

Screen Shot 2023-01-12 at 5 17 26 PM

Note also the empty timestamps of the scheduled actions—not sure if that's a bug or not?

includes/data-events/listeners.php
```diff
@@ -100,3 +100,104 @@ function( $provider, $email, $lists_added, $lists_removed, $result ) {
...
+ */
+Data_Events::register_listener(
+	'woocommerce_subscription_status_updated',
+	'donation_subscription_changed',
+	function( $subscription, $status_from, $status_to ) {
+		return [
+			'subscription_id' => $subscription->get_id(),
+			'user_id'         => $subscription->get_customer_id(),
+			'email'           => $subscription->get_billing_email(),
+			'status_before'   => $status_from,
```
dkoo, line 199: Testing this step:

> 7. Now make a recurring donation (subscription) and confirm you see the same with: donation_subscription_new, donation_subscription_changed (switching from pending to active)

The `donation_subscription_changed` action is dispatched going from `active` to `pending` instead of `pending` to `active` as described above:

Screen Shot 2023-01-12 at 5 22 45 PM

Again, not sure if it's a bug or not, but it does differ from the testing steps.

dkoo: It looks like the order of arguments here is off. According to [WC's docs](https://woocommerce.com/document/subscriptions/develop/action-reference/#section-2) for the `woocommerce_subscription_status_updated` action hook, the second argument is the "new" status (`status_to`) and the third is the old (`status_from`).

dkoo (changes requested):
Ran into one issue where the `donation_subscription_new` action is getting dispatched upon a new single donation order, and a couple of questions which may or may not be issues.

includes/data-events/listeners.php
```diff
@@ -100,3 +100,104 @@ function( $provider, $email, $lists_added, $lists_removed, $result ) {
...
+	}
+);
+
+/**
+ * For when a WooCommerce Subscription is cancelled.
+ */
+Data_Events::register_listener(
+	'woocommerce_subscription_status_updated',
+	'donation_subscription_cancelled',
+	function( $subscription, $status_from, $status_to ) {
```
leogermani, line 176: Thinking from an analytics POV, it would make sense to also have the subscription details here: amount, currency, recurrence and platform.

In case we are triggering non identifiable events and analyzing them, we could then compare to see what kind of subscriptions are being cancelled.

leogermani: same for status_update.

makes sense?

includes/data-events/listeners.php
```diff
@@ -100,3 +100,104 @@ function( $provider, $email, $lists_added, $lists_removed, $result ) {
...
+ * For when there's a new donation subscription.
+ *
+ * This will be fetched from a new donation, so we're hooking into the 'donation_new' dispatch.
+ */
+Data_Events::register_listener(
+	'newspack_data_event_dispatch_donation_new',
+	'donation_subscription_new',
+	function( $timestamp, $data ) {
+		if ( 'once' === $data['recurrence'] ) {
+			return;
```
leogermani, line 164: Nice! That's a great use of listeners I haven't thought about... same hook but dispatching different actions depending on the params

includes/data-events/README.md
```diff
@@ -80,6 +80,58 @@ When a reader updates their lists subscription from Newspack Newsletters.
 | `lists_added`   | `string[]` |
 | `lists_removed` | `string[]` |
 
+### `donation_new`
+
+When there's a new donation, either through Stripe or Newspack (WooCommerce) platforms.
+
+### Data
```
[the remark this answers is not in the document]
miguelpeixe: Good call! bb409c13105b650e49d8917bfaf1a74f3e9ad051 removes this heading from all instances

includes/data-events/listeners.php
miguelpeixe: Indeed! Fixed with 8671e7b6b5654da06a17bc4bb5f24c6bdf3ee0db

includes/data-events/listeners.php
miguelpeixe: Good call! 47e503963a39cd19212588cdee5a6810cb9318e6 expands the data for both dispatches

miguelpeixe:
@dkoo, were you able to confirm that the AC integration works as expected while using WooCommerce? It worked without issues for me:

image

includes/data-events/listeners.php
dkoo: Can confirm that the `donation_subscription_new` is only fired on new subscriptions now, not one-time donations.

The timestamps still look like `0000-00-00`, but I'm not sure if this is a bug.

includes/data-events/listeners.php
dkoo: Confirmed!

dkoo (approved):
The updates look good to me, and I note the additional data on the events based on @leogermani's suggestion. I'd wait until @leogermani gives approval too before merging, though.

I did still have one question about the timestamps of scheduled actions looking like `0000-00-00 00:00:00`, but I'm not sure if this is an issue because the proper timestamp is logged in the "Log" column. So, I'll approve this at your and @leogermani's discretion.

miguelpeixe:
Thank you, @dkoo! I noticed the `0000-00-00 00:00:00` timestamp as well. It looks like the default behavior when enqueueing a new task without scheduling, so it executes as soon as possible.

Taking a look at the AS repo, I noticed [this recently merged PR](https://github.com/woocommerce/action-scheduler/pull/898). Looks like it changes this default behavior also to improve the scheduling so overdue tasks are processed before the enqueued ones.

leogermani (approved):
This looks good to me.

In the future I'll check if it's possible to add some more data to the events to be used in the analytics, like where the donation came from? (from a prompt, sdb, etc...)

PR #2233, opened 2023-01-17
feat(data-events): mailchimp connector

miguelpeixe:
### All Submissions:

Implements the Mailchimp connector using Data Events, producing a result similar to what is achieved for ActiveCampaign with RAS.

Just like the AC implementation, this also requires a "master list", or a single audience, to send the data to. It will automatically fetch from the "Mailchimp for WooCommerce" plugin setting if it's not configured in the RAS setting `mailchimp_audience_id`.

The UI for configuring Mailchimp's master list should be built next, not in this PR.

The data transformation borrows the logic from `WooCommerce_Connection`, by moving the contact generation portion to a `get_contact_from_order( $order )` method so it can be reused.

The same logic used for ActiveCampaign in `Newspack\Newspack_Newsletters` is not reproduced, meaning:
 - No UTM integration
 - `NP_Product Name` is removed. This value will change with every different purchase, it's not clear to me the purpose of treating it as a singular contact data. Perhaps we need a better strategy for this?

The contact is upserted as soon as registered, with `status` set to `transactional` if it's new (not yet subscribed), through a handler attached to the `reader_registered` action.

Purchase updates are sent through a handler at the `donation_new` action. The order data is transformed with `WooCommerce_Connection::get_contact_from_order( $order );`.

1. Make sure you have the following constants:
```php
```
2. Make sure you have Mailchimp connected and reader revenue configured with either Stripe or Newspack platform
3. Install the `Mailchimp for WooCommerce` plugin configure with the same account and set it to an audience (this is required only until we have our own UI for setting the master list)
4. Configure a Reader Registration prompt
5. In a fresh session, register a reader through the prompt **without** subscribing to newsletters
6. Wait until `Executing action handlers for "reader_registered"` is logged
7. Confirm the contact is created in Mailchimp with the status `transactional` and the following merge fields filled:
   1. `NP_Account`
   2. `NP_Registration Page`
   3. `NP_Registration Method`
8. Make a purchase and confirm the data from `get_contact_from_order()` is updated to the same contact
9. Through the "My Account" page subscribe to the newsletter and confirm the you're subscribed without any merge field data being altered/removed

### Other information:

### DIFF
--- a/includes/data-events/connectors/class-mailchimp.php
+++ b/includes/data-events/connectors/class-mailchimp.php
@@ -92,9 +92,9 @@ class Mailchimp {
 					$fields_ids[ $field['merge_id'] ] = $field_name;
 				}
 				// If field name is found, add it to the payload.
-				if ( ! empty( $field_name ) ) {
+				if ( ! empty( $field_name ) && isset( $data[ $field_name ] ) ) {
 					$merge_fields[ $field['tag'] ] = $data[ $field_name ];
-					unset( $data[ $field['name'] ] );
+					unset( $data[ $field_name ] );
 				}
 			}
 			// Create remaining fields.
--- a/includes/data-events/connectors/class-mailchimp.php
+++ b/includes/data-events/connectors/class-mailchimp.php
@@ -23,7 +23,7 @@ class Mailchimp {
 	 * Constructor.
 	 */
 	public function __construct() {
-		if ( true === Reader_Activation::get_setting( 'sync_esp' ) ) {
+		if ( Reader_Activation::is_enabled() && true === Reader_Activation::get_setting( 'sync_esp' ) ) {
 			Data_Events::register_handler( [ __CLASS__, 'reader_registered' ], 'reader_registered' );
 			Data_Events::register_handler( [ __CLASS__, 'donation_new' ], 'donation_new' );
 		}
@@ -36,9 +36,7 @@ class Mailchimp {
 	 */
 	private static function get_audience_id() {
 		/** TODO: UI for handling Mailchimp's master list in RAS. */
-		if ( Reader_Activation::is_enabled() ) {
-			$audience_id = Reader_Activation::get_setting( 'mailchimp_audience_id' );
-		}
+		$audience_id = Reader_Activation::get_setting( 'mailchimp_audience_id' );
 		/** Attempt to use list ID from "Mailchimp for WooCommerce" */
 		if ( ! $audience_id && function_exists( 'mailchimp_get_list_id' ) ) {
 			$audience_id = mailchimp_get_list_id();
--- a/includes/data-events/connectors/class-mailchimp.php
+++ b/includes/data-events/connectors/class-mailchimp.php
@@ -159,7 +159,16 @@ class Mailchimp {
 	 * @param int   $client_id ID of the client that triggered the event.
 	 */
 	public static function reader_registered( $timestamp, $data, $client_id ) {
-		self::put( $data['email'], $data['metadata'] );
+		$metadata = [
+			'NP_Account' => $data['user_id'],
+		];
+		if ( isset( $data['metadata']['current_page_url'] ) ) {
+			$metadata['NP_Registration Page'] = $data['metadata']['current_page_url'];
+		}
+		if ( isset( $data['metadata']['registration_method'] ) ) {
+			$metadata['NP_Registration Method'] = $data['metadata']['registration_method'];
+		}
+		self::put( $data['email'], $metadata );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


miguelpeixe:
Requesting a review from @adekbadek because it uses a lot of his code on a different backend strategy to produce a similar result.

leogermani:
I've tested everything just from a functional point of view, following the instructions, and everything worked just fine.

I also tested making a donation before registering and that did not sync (not sure it was expected to)

miguelpeixe:
> I also tested making a donation before registering and that did not sync (not sure it was expected to)

It will bail if a contact (registered user) is not available, but if a reader is created during (after) the donation is made we might need to tweak it on the WC integration side so that it becomes available at the hook.

includes/oauth/class-mailchimp-api.php
```diff
@@ -143,37 +132,98 @@ private static function get_api_endpoint_from_key( $api_key ) {
...
+
+	/**
+	 * Perform a POST request to Mailchimp's API
+	 *
+	 * @param string $path API path.
+	 * @param array  $data Data to send.
+	 *
+	 * @return array|WP_Error API response or error.
+	 */
+	public static function post( $path = '', $data ) {
```
adekbadek, line 225: In PHP 8, required parameters must be listed first.

includes/oauth/class-mailchimp-api.php
```diff
@@ -143,37 +132,98 @@ private static function get_api_endpoint_from_key( $api_key ) {
...
-		return \rest_ensure_response( [ 'username' => $response_body['username'] ] );
+	/**
+	 * Perform a PUT request to Mailchimp's API
+	 *
+	 * @param string $path API path.
+	 * @param array  $data Data to send.
+	 *
+	 * @return array|WP_Error API response or error.
+	 */
+	public static function put( $path = '', $data ) {
```
adekbadek, line 213: Same as below, in PHP 8, required parameters must be listed first.

includes/data-events/connectors/class-mailchimp.php
```diff
@@ -0,0 +1,205 @@
...
+ * Main Class.
+ */
+class Mailchimp {
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		if ( Reader_Activation::is_enabled() && true === Reader_Activation::get_setting( 'sync_esp' ) ) {
+			Data_Events::register_handler( [ __CLASS__, 'reader_registered' ], 'reader_registered' );
+			Data_Events::register_handler( [ __CLASS__, 'donation_new' ], 'donation_new' );
```
adekbadek, line 28: The methods other than these two seem to be strictly related to the API, I was surprised to find them here and not in `class-mailchimp-api.php`.

adekbadek (approved):
Since @leogermani did functional testing, I've only looked at the code. The changes to `class-woocommerce-connection.php` look fine. Left one note about argument order – after this is fixed we're good to merge.

adekbadek:
> NP_Product Name is removed. This value will change with every different purchase, it's not clear to me the purpose of treating it as a singular contact data. Perhaps we need a better strategy for this?

This field (and a few others) are based on the assumption that a reader would buy a single product, which aligns with a membership-driven model. Newspack supports more models, but this specific integration was requested by a membership-focused third party.

includes/oauth/class-mailchimp-api.php
miguelpeixe: In order to preserve the argument order like in the order methods, 7c715b361520426ca75a4ed6711d26c8768f4fd0 makes `$data` also optional

includes/oauth/class-mailchimp-api.php
miguelpeixe: In order to preserve the argument order like in the order methods, 7c715b361520426ca75a4ed6711d26c8768f4fd0 makes `$data` also optional

includes/data-events/connectors/class-mailchimp.php
miguelpeixe: I don't find the connector methods to be abstract enough to be used within the generic `Mailchimp_API` class. I personally like that the API class is light and provides simple CRUD methods.

miguelpeixe:
> This field (and a few others) are based on the assumption that a reader would buy a single product, which aligns with a membership-driven model.

In the interest of feature parity, 89e956688f82df13c91b38dcb0102cd43d523d0a implements the `NP_Product Name` support through the `donation_subscription_new` action, instead of every donation. In this case, every new recurring donation will trigger an update in MC's contact. WDYT?

It can be extended to also handle `donation_subscription_cancelled`, but I wonder about its use.

miguelpeixe:
> I also tested making a donation before registering and that did not sync (not sure it was expected to)

@leogermani I tried on both Stripe and Newspack platforms and an anonymous donation (fresh session) did sync without issues. Which platform were you using, can you provide more details?

adekbadek:
> NP_Product Name support through the donation_subscription_new action

Sounds good!

PR #2239, opened 2023-01-19
feat(performance): minify all inline CSS

adekbadek:
### All Submissions:

Adds minification of all inline CSS. The minification will remove whitespace and comments. 

1. On `master`, load a page and inspect the source
2. Observe that there are `<style>` tags with CSS, which contains unnecessary whitespace and comments
3. Switch to this branch, reload the page
4. Observe the inline CSS is minified
5. Smoke-test how the page renders, to ensure the minification did not break any CSS

### Other information:

### DIFF
--- a/tests/unit-tests/performance.php
+++ b/tests/unit-tests/performance.php
@@ -1,0 +1,37 @@
+<?php
+/**
+ * Tests the Performance optmisations.
+ *
+ * @package Newspack\Tests
+ */
+
+use Newspack\Performance;
+
+/**
+ * Tests the performance optmisations.
+ */
+class Newspack_Test_Newspack_Performance extends WP_UnitTestCase {
+	/**
+	 * Test CSS minification.
+	 */
+	public function test_performance_css_minification() {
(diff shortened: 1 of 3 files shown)


tests/unit-tests/performance.php
```diff
@@ -0,0 +1,37 @@
...
+				 * And this is a multi-line comment, with a link: https://example.com/hello-world
+				 */
+				body {
+					color: red;
+				}
+			</style>
+		</html>';
+		self::assertEquals(
+			'<html>
+			<style>body{color: red;}</style>
```
dkoo, line 31: Definitely a nit, but maybe more savings to be had: couldn't we strip any whitespace space after the colon in `color: red;`?

dkoo (approved):
Didn't see any CSS issues on my local test site, and the inline CSS is minified as expected.

tests/unit-tests/performance.php
adekbadek: That would be possible, but I'd prefer to keep the regular expression simple and less error-prone.

PR #2241, opened 2023-01-20
feat: bootstrap GA4 connector

leogermani:
### All Submissions:

This pull request bootstraps the GA4 connector for our Data Events api and registers one first handler.

```
```
Now, let's test:

```
```
3. Brwose the site in different/browser devices and watch the pageviews and user appear on the GA Real time dashboard
4. Login with a reader (either via password or OTP)
5. Watch the `reader_login` event being fired in the logs
```
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "reader_logged_in" via Action Scheduler.
```
6. Watch the handler being triggered and sending the `reader_login` GA event:
```
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::handle]: Executing action handlers for "reader_logged_in".
[NEWSPACK-DATA-EVENTS-GA4][Newspack\Data_Events\Connectors\GA4::log]: Event sent - reader_login
```

7. Go to the Google Analytics Realtime dashboard and watch the events coming in

Captura de tela de 2023-01-19 17-56-27

8. Do this a couple of times and confirm all the events are coming in and have the `logged_in` and `ga_session_id` params.
Captura de tela de 2023-01-23 17-12-40

(Note, `logged_in` is a global param added to all events, it may not make much sense in this particular event, but this is fine)

9. Compare the `ga_session_id` param of the `reader_login` event with the others events like `page_view` or `session_start`. They should match. There should be one session id for each device/browser you use.

10. Make sure the number of user and sessions did not increase after the `reader_login` events were fired.

11. Disable woocommerce and test this being triggered without Action Scheduler

### Other information:

### DIFF
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -103,7 +103,6 @@ final class Data_Events {
 			sprintf( 'Executing global action handlers for "%s".', $action_name ),
 			self::LOGGER_HEADER
 		);
-
 		foreach ( self::$global_handlers as $handler ) {
 			try {
 				call_user_func( $handler, $action_name, $timestamp, $data, $client_id );
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -276,6 +276,11 @@ final class Data_Events {
 			return new WP_Error( 'newspack_data_events_action_not_registered', __( 'Action not registered.', 'newspack' ) );
 		}
 
+		// Append User Agent to the data so we can identify it in the async requests.
+		if ( isset( $_SERVER['HTTP_USER_AGENT'] ) ) {
+			$data['user_agent'] = sanitize_text_field( $_SERVER['HTTP_USER_AGENT'] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
+		}
+
 		$timestamp = time();
 		$client_id = null;
 		if ( $use_client_id ) {
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -43,7 +43,7 @@ final class Data_Events {
 	public static function init() {
 		\add_action( 'wp_ajax_' . self::ACTION, [ __CLASS__, 'maybe_handle' ] );
 		\add_action( 'wp_ajax_nopriv_' . self::ACTION, [ __CLASS__, 'maybe_handle' ] );
-		\add_action( 'newspack_data_events_as_dispatch', [ __CLASS__, 'handle' ], 10, 5 );
+		\add_action( 'newspack_data_events_as_dispatch', [ __CLASS__, 'handle' ], 10, 4 );
 	}
 
 	/**
@@ -80,12 +80,11 @@ final class Data_Events {
 			\wp_die();
 		}
 
-		$timestamp  = isset( $_POST['timestamp'] ) ? \sanitize_text_field( \wp_unslash( $_POST['timestamp'] ) ) : null;
-		$data       = isset( $_POST['data'] ) ? $_POST['data'] : null; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-		$client_id  = isset( $_POST['client_id'] ) ? \sanitize_text_field( \wp_unslash( $_POST['client_id'] ) ) : null;
-		$user_agent = isset( $_POST['user_agent'] ) ? \sanitize_text_field( \wp_unslash( $_POST['user_agent'] ) ) : null;
+		$timestamp = isset( $_POST['timestamp'] ) ? \sanitize_text_field( \wp_unslash( $_POST['timestamp'] ) ) : null;
+		$data      = isset( $_POST['data'] ) ? $_POST['data'] : null; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$client_id = isset( $_POST['client_id'] ) ? \sanitize_text_field( \wp_unslash( $_POST['client_id'] ) ) : null;
 
-		self::handle( $action_name, $timestamp, $data, $client_id, $user_agent );
+		self::handle( $action_name, $timestamp, $data, $client_id );
 
 		\wp_die();
 	}
@@ -97,19 +96,13 @@ final class Data_Events {
 	 * @param int    $timestamp   Timestamp.
 	 * @param array  $data        Data.
 	 * @param string $client_id   Client ID.
-	 * @param string $user_agent  The User agent of the original request.
 	 */
-	public static function handle( $action_name, $timestamp, $data, $client_id, $user_agent = null ) {
+	public static function handle( $action_name, $timestamp, $data, $client_id ) {
 		// Execute global handlers.
 		Logger::log(
 			sprintf( 'Executing global action handlers for "%s".', $action_name ),
 			self::LOGGER_HEADER
 		);
-
-		// Override user agent with the user agent of the request that triggered the event.
-		if ( $user_agent ) {
-			$_SERVER['HTTP_USER_AGENT'] = $user_agent; // phpcs:ignore
-		}
 
 		foreach ( self::$global_handlers as $handler ) {
 			try {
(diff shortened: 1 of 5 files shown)


includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	/**
+	 * Get the default parameters that are added to all events
+	 *
+	 * @return array
+	 */
+	private static function get_default_params() {
+		$params = [
+			'logged_in' => is_user_logged_in() ? 'yes' : 'no',
```
miguelpeixe, line 61: The handler executes outside of the user's original request. `is_user_logged_in()` or any other session related information is not to be trusted unless explicitly passed as raw data.

If the handler/connection requires additional information, the action payload should be revised, and make that data available for every handler. I don't think it's the case here because "logged in" is the actual information being sent.

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	/**
+	 * Extracts the Client ID from the _ga cookie
+	 *
+	 * If the cookie is not found, it will be created
+	 *
+	 * @return string
+	 */
+	private static function extract_cid_from_cookies() {
```
miguelpeixe, line 151: Right before the `Event sent - reader_login` log I'm getting this warning:

```
PHP Warning:  Cannot modify header information - headers already sent in newspack-plugin/includes/data-events/connectors/ga4/class-ga4.php on line 166
```

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -0,0 +1,185 @@
...
+	private static function get_default_params() {
+		$params = [
+			'logged_in' => is_user_logged_in() ? 'yes' : 'no',
+		];
+
+		if ( class_exists( 'Automattic\Jetpack\Device_Detection' ) ) {
+			self::log( 'Detecting device using user agent: ' . $_SERVER['HTTP_USER_AGENT'] ?? '(empty)' ); // phpcs:ignore
+			$device                = Device_Detection::is_phone() ? 'phone' : ( Device_Detection::is_tablet() ? 'tablet' : 'desktop' );
+			$params['device_type'] = $device;
+		}
```
miguelpeixe, line 68: How does GA treat the event if it's attached to a `_ga` cookie (GA session)? I was hoping all the standard GA tracking would provide all the browser/client data without having to duplicate it ourselves.

Ideally, we would keep events more agnostic and only contain the information related to the action.

includes/data-events/connectors/ga4/class-ga4.php
miguelpeixe: I'm guessing it's not happy with the `setcookie()` part

miguelpeixe (changes requested):

includes/data-events/connectors/ga4/class-ga4.php
miguelpeixe: These can get quite extensive, I'm not sure we should attach too much session information unrelated to the actual action/event. As I commented below, doesn't GA have the ability to integrate the session information through the `_ga` cookie and allow you to contextualize the event?

leogermani: I'm looking into it. Currently it's not, but I'll dig a bit more.

But this will never include WP specific information, such as logged_in, is_subscribed, is_donor and other information

includes/data-events/connectors/ga4/class-ga4.php
leogermani: Even though I'm passing the `client_id` based on the cookie, as the documenation suggests, I don't see anything in the GA dashboard. When I see the event details, they don't have any of these standard dimensions... 

I'll keep looking. I think this is a critical piece of the implementation

includes/data-events/connectors/ga4/class-ga4.php
miguelpeixe: True, for WP-specific data we can think of a generic and filterable `user_metadata` sent on every event. WDYT?

includes/data-events/connectors/ga4/class-ga4.php
leogermani: I've done a long exploration. See pamTN9-5QY-p2#comment-7460

It looks like these params will never be automatically added, but I managed to add a session_id to our events that will allow us to tie them to other front-end events when we process the raw data in BigQuery.

I've updated the testing intructions.

includes/data-events/connectors/ga4/class-ga4.php
leogermani: I've removed it. If the cookie is not set, we use a anonymous user, just as we are already doing [here](https://github.com/Automattic/newspack-plugin/blob/add%2Fga4-connector/includes/oauth/class-google-services-connection.php#L109)

includes/data-events/connectors/ga4/class-ga4.php
leogermani: Will work on this default data on the next PR

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -0,0 +1,218 @@
...
+
+		if ( isset( $_COOKIE['_ga'] ) ) {
+			$cookie_pieces = explode( '.', $_COOKIE['_ga'], 3 ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			if ( 1 === count( $cookie_pieces ) ) {
+				$cid = reset( $cookie_pieces );
+			} else {
+				list( $version, $domain_depth, $cid ) = $cookie_pieces;
+			}
+		} else {
+			$cid = 555;
```
miguelpeixe, line 182:
```suggestion
			$cid = '555';
```

If we are to keep the "anonymous client", the `cid` should be a `string` or GA's API is unable to parse the JSON payload.

miguelpeixe (changes requested):
Working well! Only a feel tweaks/requests below

includes/data-events/class-data-events.php
```diff
@@ -103,6 +103,7 @@ public static function handle( $action_name, $timestamp, $data, $client_id ) {
 			sprintf( 'Executing global action handlers for "%s".', $action_name ),
 			self::LOGGER_HEADER
 		);
+
```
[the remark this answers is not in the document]
leogermani: accident

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -0,0 +1,218 @@
...
+	 * @return void
+	 */
+	public static function handle_reader_logged_in( $timestamp, $data, $user_id ) {
+
+		$params    = $data['ga_params'];
+		$client_id = $data['ga_client_id'];
+
+		if ( empty( $client_id ) ) {
+			throw new \Exception( 'Missing client ID' );
+		}
```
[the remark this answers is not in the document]
leogermani: yes, let's throw. And let's add a note somewhere to talk about some day -> p1674664784606179-slack-C015W6BES8J.

includes/data-events/connectors/ga4/class-ga4.php
leogermani: let's throw

includes/data-events/class-data-events.php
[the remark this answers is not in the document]
leogermani: fixed

includes/data-events/connectors/ga4/class-ga4.php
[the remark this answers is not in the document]
leogermani: done

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -0,0 +1,218 @@
...
+
+	/**
+	 * Sends an event to GA4.
+	 *
+	 * @param Event  $event The event object.
+	 * @param string $client_id The GA client ID obtained from the cookie.
+	 * @param int    $timestamp The timestamp of the event.
+	 * @param string $user_id User identifier. Use Reader_Activation::get_client_id().
+	 * @throws \Exception If the credentials are missing.
+	 * @return void
```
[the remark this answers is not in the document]
leogermani: fixed

includes/data-events/connectors/ga4/class-ga4.php
leogermani: done

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -179,7 +180,7 @@ private static function extract_cid_from_cookies() {
 				list( $version, $domain_depth, $cid ) = $cookie_pieces;
 			}
 		} else {
-			$cid = 555;
+			throw new \Exception( 'Missing _ga Cookie' );
```
miguelpeixe, line 183: We shouldn't `throw` here. This is not running on the handler execution. It's behind the dispatch's body filter, which is synchronous and will throw a fatal to the user.

miguelpeixe (changes requested):

includes/data-events/connectors/ga4/class-ga4.php
leogermani: Thanks!

miguelpeixe (approved):
🙌

PR #2249, opened 2023-01-24
feat(ras): enable custom contact metadata prefixes

dkoo:
### All Submissions:

Adds a field to the RAS settings page to set the prefix for contact metadata field keys. Defaults to `NP_` which ensures continuity on sites already using RAS.

Screen Shot 2023-01-24 at 4 38 43 PM

Closes 1200550061930446/1203749513702559.

1. On a site with RAS enabled and connected to ActiveCampaign, visit the **Newspack > Engagement > Reader Activation** dashboard. Confirm there's a new "Metadata Field Prefix" field if the the "Synchronize reader to ESP" is enabled, and that it's set to `NP_`.
2. Without making changes, smoke test the field syncing to ensure that it still works with the default behavior: make a donation and ensure that the contact is created in ActiveCampaign with all the Newspack fields prefixed with `NP_`.
3. Update the "Metadata Field Prefix" field to a new string and save.
4. Make a new donation using the same email address as in step 2. Confirm in AC that the contact now has a new set of fields with the prefix you set in step 3 which contain the most updated data, in addition to the `NP_` fields from before. (The old fields will remain to prevent data loss, but will no longer be synced or updated unless you reset the prefix.)
5. Make a new donation in a new session with a different email address. Confirm in AC that the data for this contact is only synced to the fields with the new prefix.

* Note: The new fields will appear in the General Details field group in ActiveCampaign. They can be moved to a different group in the AC dashboard.

### Other information:

### DIFF
--- a/includes/plugins/class-newspack-newsletters.php
+++ b/includes/plugins/class-newspack-newsletters.php
@@ -13,35 +13,9 @@ defined( 'ABSPATH' ) || exit;
  * Main class.
  */
 class Newspack_Newsletters {
-	const METADATA_DATE_FORMAT = 'Y-m-d';
-	const METADATA_PREFIX      = 'NP_';
-
-	/**
-	 * Metadata keys map for Reader Activation.
-	 *
-	 * @var array
-	 */
-	public static $metadata_keys = [
-		'account'              => self::METADATA_PREFIX . 'Account',
-		'registration_date'    => self::METADATA_PREFIX . 'Registration Date',
-		'connected_account'    => self::METADATA_PREFIX . 'Connected Account',
-		'signup_page'          => self::METADATA_PREFIX . 'Signup Page',
-		'signup_page_utm'      => self::METADATA_PREFIX . 'Signup UTM: ',
-		'newsletter_selection' => self::METADATA_PREFIX . 'Newsletter Selection',
-		// Payment-related.
-		'membership_status'    => self::METADATA_PREFIX . 'Membership Status',
-		'payment_page'         => self::METADATA_PREFIX . 'Payment Page',
-		'payment_page_utm'     => self::METADATA_PREFIX . 'Payment UTM: ',
-		'sub_start_date'       => self::METADATA_PREFIX . 'Current Subscription Start Date',
-		'sub_end_date'         => self::METADATA_PREFIX . 'Current Subscription End Date',
-		'billing_cycle'        => self::METADATA_PREFIX . 'Billing Cycle',
-		'recurring_payment'    => self::METADATA_PREFIX . 'Recurring Payment',
-		'last_payment_date'    => self::METADATA_PREFIX . 'Last Payment Date',
-		'last_payment_amount'  => self::METADATA_PREFIX . 'Last Payment Amount',
-		'product_name'         => self::METADATA_PREFIX . 'Product Name',
-		'next_payment_date'    => self::METADATA_PREFIX . 'Next Payment Date',
-		'total_paid'           => self::METADATA_PREFIX . 'Total Paid',
-	];
+	const METADATA_DATE_FORMAT   = 'Y-m-d';
+	const METADATA_PREFIX        = 'NP_';
+	const METADATA_PREFIX_OPTION = '_newspack_metadata_prefix';
 
 	/**
 	 * Initialize hooks and filters.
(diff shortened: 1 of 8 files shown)


includes/plugins/class-newspack-newsletters.php
```diff
@@ -54,6 +28,68 @@ public static function init() {
...
+	 */
+	public static function get_metadata_prefix() {
+		$prefix = \get_option( self::METADATA_PREFIX_OPTION, self::METADATA_PREFIX );
+
+		// Guard against empty strings and falsy values.
+		if ( empty( $prefix ) ) {
+			return self::METADATA_PREFIX;
+		}
+
+		return $prefix;
```
miguelpeixe, line 45: It can be useful to have this filterable.

miguelpeixe (changes requested):

includes/plugins/class-newspack-newsletters.php
```diff
@@ -54,6 +28,68 @@ public static function init() {
...
+			'sub_start_date'       => $metadata_prefix . 'Current Subscription Start Date',
+			'sub_end_date'         => $metadata_prefix . 'Current Subscription End Date',
+			'billing_cycle'        => $metadata_prefix . 'Billing Cycle',
+			'recurring_payment'    => $metadata_prefix . 'Recurring Payment',
+			'last_payment_date'    => $metadata_prefix . 'Last Payment Date',
+			'last_payment_amount'  => $metadata_prefix . 'Last Payment Amount',
+			'product_name'         => $metadata_prefix . 'Product Name',
+			'next_payment_date'    => $metadata_prefix . 'Next Payment Date',
+			'total_paid'           => $metadata_prefix . 'Total Paid',
+		];
```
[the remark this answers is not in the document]
dkoo: That sounds like a good idea—does 3659d6e do what you suggest? Or should the whole field name be filterable too, in your opinion?

includes/plugins/class-newspack-newsletters.php
dkoo: Good idea. Done in 3659d6e!

includes/plugins/class-newspack-newsletters.php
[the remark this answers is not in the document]
miguelpeixe: Not exactly what I had in mind, I was thinking of something like this:

```php
$keys = [
	'account' => 'Account',
];

function get_key( $key ) {
	if ( ! isset( $keys[ $key ] ) ) {
		return false;
	}
	// Prefix could be either here or as an added filter.
	$prefix = get_prefix();
	$key    = $prefix . $keys[ $key ];
	return apply_filters( 'get_key', $key, $key );
}

function function_that_uses_key() {
	$key = get_key( 'account' );
	// Do something with the key.
}
```

This allows a cleaner map of keys and the processing of the key value only occurs when used, which saves some time.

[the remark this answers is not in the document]
dkoo: Thanks for the hint. [7843fa9](https://github.com/Automattic/newspack-plugin/pull/2249/commits/7843fa95e3149f31d71e934916e63db942c551cf) implements your suggestion (with some merge conflict resolution afterward).

includes/plugins/class-newspack-newsletters.php
```diff
@@ -54,6 +56,69 @@ public static function init() {
...
+	 * Update the prefix for synced metadata fields.
+	 *
+	 * @param string $prefix Value to set.
+	 *
+	 * @return boolean True if updated, false otherwise.
+	 */
+	public static function update_metadata_prefix( $prefix ) {
+		if ( empty( $prefix ) ) {
+			$prefix = self::METADATA_PREFIX;
+		}
```
miguelpeixe, line 91: Is it intentional to not allow this value to be an empty string? If I clear the input value and save I get `NP_`.

miguelpeixe (approved):
I pushed in 50fe6fc658a6b31cd0f83087cab8f55a2bd87d53 the necessary changes for this new system to work with the Mailchimp connector as well.

There's only a non-blocking question below.

includes/plugins/class-newspack-newsletters.php
dkoo: Thanks for the changes on the Mailchimp connector!

> Is it intentional to not allow this value to be an empty string? If I clear the input value and save I get `NP_`.

That's intended to provide continuity for all the sites already using RAS. They won't have any option set for the prefix, but the prior code assumed `NP_` is the prefix, so we want to ensure those sites continue to send data to fields with an `NP_` prefix.

That said, it might be overly cautious since the `NP_` prefix is the default value of the `get_option` call. What do you think—should we support an empty prefix here?

miguelpeixe: > That's intended to provide continuity for all the sites already using RAS.

I think a default value can be determined without necessarily disallowing an empty value.

> What do you think—should we support an empty prefix here?

I don't see why not. Although `NP_` is our default, I can imagine a publisher not wanting any prefix at all.

dkoo: Thanks for the discussion—will amend to allow an empty string.

dkoo: Update: We have some [security behavior in place](https://github.com/Automattic/newspack-plugin/blob/master/includes/plugins/class-newspack-newsletters.php#L192) to ensure that a prefix exists in order to avoid accidentally overwriting other fields that might have the same name. For example, some of our unprefixed fields have pretty generic names that are likely to collide, e.g. `Account`. c18f910 retains the non-empty requirement, but adds some help text to the input field explaining that a prefix is required and why.

miguelpeixe (approved):
🙌

dkoo:
Awaiting resolution of https://github.com/Automattic/newspack-newsletters/pull/1070 before merging.

PR #2251, opened 2023-01-26
Stripe: create native WC Subscriptions (if a flag is set)

adekbadek:
### All Submissions:

Changes the way Stripe donations are handled, so that it's possible that instead of a Stripe Subscription, a WC Subscriptions' Subscription is created for recurring payments. 

There is an underlying change in how Stripe API is interacted with. The way to charge a customer is either through [Sources](https://stripe.com/docs/api/sources), [Tokens](https://stripe.com/docs/api/tokens), or [Payment Methods](https://stripe.com/docs/api/payment_methods) – the latter [replaces the former two](https://stripe.com/docs/payments/payment-methods/transitioning). This integration has used Payment Methods, but since [WC Stripe Gateway uses a Source to process WC Subscriptions renewals](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/compat/trait-wc-stripe-subscriptions.php#L245-L247), the integration was updated to use Sources instead of Payment Methods.  

There are two ways to pay, and two modes of payment. When testing, ideally test all four scenarios (let's name them with letters for easier communication):

|         | Once | Recurring (e.g. monthly) 
|--------------|-----------|--|
| Card form | A | B  | 
| PaymentRequest (e.g. Apple Pay)      | C | D

1. Set up Stripe, ensure the site is reachable by Stripe webhooks
2. Ensure the WC suite is active, and "Stripe – Credit Card (Stripe)" set as the sole payment method in WC
1. Test the payment scenarios for regressions – the recurring payments should result in Stripe Subscriptions shadowed by sync'd WC Subscriptions (which can't be edited, etc.)
3. Set the `NEWSPACK_USE_WC_SUBSCRIPTIONS_WITH_STRIPE_PLATFORM` environment flag to `true`
4. Test the payment scenarios with this new configuration – recurring payments should result in WC Subscriptions and one-time charges in Stripe (_not_ Stripe Subscriptions)
5. Open the WC Subscriptions (scenarios B & D) in WP admin renew them manually (via the "Subscription actions" meta box)
7. Observe that the two subscriptions were renewed:
    1. Each got a new Stripe charge (successful)
    2. Each got a renewal order in WC
8. Disable the Stripe gateway in WC, observe that a recurring donation results in a Stripe Subscription

### Other information:

### DIFF
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -382,6 +382,26 @@ class WooCommerce_Connection {
 	}
 
 	/**
+	 * Update an order.
+	 *
+	 * @param int   $order_id Order ID.
+	 * @param array $update Update data.
+	 */
+	public static function update_order( $order_id, $update ) {
+		$order = \wc_get_order( $order_id );
+		if ( false === $order ) {
+			Logger::error( 'Could not find WC order with id: ' . $order_id );
+		} else {
+			if ( isset( $update['status'] ) ) {
+				$order->set_status( $update['status'] );
+			}
+			self::add_wc_stripe_gateway_metadata( $order, $update );
+			$order->save();
+			Logger::log( 'Updated WC order with id: ' . $order->get_id() );
+		}
+	}
+
+	/**
 	 * Update WC Stripe Gateway related metadata to an order or subscription.
 	 * The order has to be saved afterwards.
 	 *
@@ -577,6 +597,14 @@ class WooCommerce_Connection {
 					Logger::error( 'Error creating WC subscription: ' . $subscription->get_error_message() );
 				} else {
 					$subscription_id = $subscription->get_id();
+
+					$wc_subscription_payload = [
+						'stripe_customer_id' => $order_data['stripe_customer_id'],
+					];
+					if ( isset( $order_data['stripe_source_id'] ) ) {
+						$wc_subscription_payload['stripe_source_id'] = $order_data['stripe_source_id'];
+					}
+					self::add_wc_stripe_gateway_metadata( $subscription, $wc_subscription_payload );
 					self::add_universal_order_data( $subscription, $order_data );
 					if ( false === $stripe_subscription_id ) {
 						$subscription->add_order_note( __( 'This subscription was created via Newspack.', 'newspack' ) );
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -108,6 +108,7 @@ class WooCommerce_Connection {
 		$item->set_product( \wc_get_product( $product_id ) );
 		$item->set_total( $amount );
 		$item->set_subtotal( $amount );
+		$item->save();
 		return $item;
 	}
 
@@ -623,6 +624,8 @@ class WooCommerce_Connection {
 						$subscription->add_order_note( sprintf( __( 'Newspack subscription with frequency: %s. The recurring payment and the subscription will be handled in Stripe, so you\'ll see "Manual renewal" as the payment method in WooCommerce.', 'newspack' ), $frequency ) );
 						$subscription->update_status( 'active' ); // Settings status via method (not in wcs_create_subscription), to make WCS recalculate dates.
 					}
+					// Mint a new item – subscription is a new WC order.
+					$item = self::get_donation_order_item( $frequency, $order_data['amount'] );
 					$subscription->add_item( $item );
 					$subscription->calculate_totals();
 
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -610,7 +610,7 @@ class WooCommerce_Connection {
 			 * Handle WooCommerce Subscriptions - new subscription.
 			 */
 			if ( 'created' === $subscription_status ) {
-				// A subscription needs a valid user. If not user ID is provided, attribute the
+				// A subscription needs a valid user. If no user ID is provided, attribute the
 				// subscription to the user with the supplied email address.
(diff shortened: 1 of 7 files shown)


includes/reader-revenue/stripe/class-stripe-connection.php
```diff
@@ -786,38 +785,112 @@ public static function handle_donation( $config ) {
...
+			$is_recurring = 'once' != $frequency;
+
+			// In this mode, WC will create a subscription for the customer, instead of using
+			// native Stripe subscriptions.
+			$is_wc_first         = false;
+			$is_wc_first_enabled = defined( 'NEWSPACK_USE_WC_SUBSCRIPTIONS_WITH_STRIPE_PLATFORM' ) && NEWSPACK_USE_WC_SUBSCRIPTIONS_WITH_STRIPE_PLATFORM;
+			if ( $is_wc_first_enabled && Donations::is_woocommerce_suite_active() ) {
+				$gateways = WooCommerce_Configuration_Manager::get_payment_gateways( true );
+				// Check if Stripe (Credit Card) is the only payment gateway enabled.
+				if ( 1 === count( $gateways ) && isset( $gateways['stripe'] ) ) {
```
dkoo, line 797: Why limit this behavior to sites that _only_ have the Stripe payment gateway enabled? Why not also create Woo subscriptions as long as it's one of the configured payment gateways?

dkoo (commented):
I wasn't able to get through all of the testing steps. Perhaps some of these are due to user error or incorrect config on my test site, and maybe others are due to changes from other branches. Here's what I saw:

1. Manually renewing a Woo subscription:

> Open the WC Subscriptions (scenarios B & D) in WP admin renew them manually (via the "Subscription actions" meta box)

When I did this, it resulted in a pending order being added to the subscription. The order never seems to complete. Because the order hangs as pending, the subscription status is changed to "On Hold", and a transaction never occurs on Stripe.

2. Custom metadata fields don't seem to be fully syncing. On `master`, I see all the `NP_` fields being synced, but on this branch only a few are:

Screen Shot 2023-01-30 at 4 11 47 PM

I wonder if this is due to these changes ([1](https://github.com/Automattic/newspack-plugin/pull/2249), [2](https://github.com/Automattic/newspack-newsletters/pull/1070)) to the custom fields? I tried rebasing against `master` but that didn't change anything for me.

3. I wasn't able to test the C & D scenarios (PaymentIntent). Do we have a way to test these payment methods without requiring a real credit card transaction?

adekbadek:
> When I did [manually renew a Woo subscription] it resulted in a pending order being added to the subscription. The order never seems to complete. 

The order is created synchronously (as pending), but it's completed on webhook callback. Can you make sure the site is reachable for Stripe's webhook requests? Alternatively, maybe there's a PHP error logged during the webhook request?

> Custom metadata fields don't seem to be fully syncing.

Good catch! This led me to fix an unidiomatic use of WC's order items (2621c96497ce7c2d83e380c9b2be48d93cf52406 – an order item has to be unique). But for the most part, the ESP data syncing is fixed in be22c5f0293148bb26a9f54b686f5ab013156983. 

> I wasn't able to test the C & D scenarios (PaymentIntent). Do we have a way to test these payment methods without requiring a real credit card transaction?

As long as Stripe is configured to use test mode, you can test with real payment details and you won't be charged. See https://support.stripe.com/questions/testing-apple-pay-with-stripe

includes/reader-revenue/stripe/class-stripe-connection.php
adekbadek: Good question! This was premature on my part – in the future there should be just one on-site RR platform. In this feature, SDB would render only if `Stripe (Credit Card)` is the only gateway enabled. Otherwise the other gateways would be unavailable. Reverted this change in aa1bc0991e958e87543bbdcca82860faa9365bba.

includes/reader-revenue/class-woocommerce-connection.php
```diff
@@ -186,7 +188,11 @@ public static function get_contact_from_order( $order ) {
 
 		$metadata[ Newspack_Newsletters::get_metadata_key( 'account' ) ]           = $order->get_customer_id();
 		$metadata[ Newspack_Newsletters::get_metadata_key( 'registration_date' ) ] = $customer->get_date_created()->date( Newspack_Newsletters::METADATA_DATE_FORMAT );
-		$metadata[ Newspack_Newsletters::get_metadata_key( 'payment_page' ) ]      = \wc_get_checkout_url();
```
adekbadek, line 189: Note that `payment_page` meta field value [is computed](https://github.com/Automattic/newspack-plugin/blob/173121e41d23bfe6fb1044d3b84bdfe5860ccac9/includes/plugins/class-newspack-newsletters.php#L235-L244) in `Newspack_Newsletters_Subscription::add_contact`, it's not meant to be passed directly. So it's removed here.

dkoo:
> The order is created synchronously (as pending), but it's completed on webhook callback. Can you make sure the site is reachable for Stripe's webhook requests? Alternatively, maybe there's a PHP error logged during the webhook request?

I'm not seeing any PHP error resulting from the renewal, and in fact I don't see any webhook request in the Stripe dashboard when I process a renewal from the WooCommerce Subscription page. Is there another way I should try manually renewing the subscription?

dkoo:
> Good catch! This led me to fix an unidiomatic use of WC's order items (https://github.com/Automattic/newspack-plugin/commit/2621c96497ce7c2d83e380c9b2be48d93cf52406 – an order item has to be unique). But for the most part, the ESP data syncing is fixed in https://github.com/Automattic/newspack-plugin/commit/be22c5f0293148bb26a9f54b686f5ab013156983.

Can confirm that the data syncing is fixed, now!

dkoo (approved):
I'm going to approve this since the only outstanding issues seem to be related to my testing environment. Thanks for the fixes!

PR #757, opened 2020-12-18
fix: show campaigns with pending and future status in UI

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Current campaign UI only shows campaign posts that have `publish` and `draft` statuses. This would cause posts with the `pending` or `future` status to be dropped before being rendered in the UI.

This PR, along with https://github.com/Automattic/newspack-popups/pull/356, shows pending drafts under the "Draft" header and scheduled campaigns under the "Inactive" header, with a corresponding "Pending review" or "Scheduled" label for each.

Screen Shot 2020-12-18 at 3 36 46 PM

Closes https://github.com/Automattic/newspack-popups/issues/355.

### How to test the changes in this Pull Request:

1. On `master` for both this plugin and in Campaigns, create a new overlay or inline campaign and save it as a draft with the "Pending review" option enabled:

Screen Shot 2020-12-18 at 3 37 56 PM

2. ~Go to **Newspack > Campaigns** and observe that the campaign is not rendered anywhere.~ Because of https://github.com/Automattic/newspack-popups/pull/399, you'll actually see these campaigns in the list, but they'll appear as drafts.
3. Check out this branch and run `npm run build`.
4. In **Newspack > Campaigns** under the Campaigns tab confirm that you now see the pending campaign with the "Draft" color with a "Pending review" label.
5. Edit the campaign again. Change the publish date to some date in the future and click "Schedule" button to schedule it for future publishing (the pending option is automatically disabled with the `future` and `publish` statuses).
6. Refresh the **Newspack > Campaigns** page. Confirm that the campaign still appears with the "Draft" color with a "Scheduled" label.
7. ~Turn on Test mode for the campaign and refresh the **Newspack > Campaigns** page. Confirm that the campaign still appears with the "Draft" color with a "Scheduled" label.~ There's no longer a test mode, so you can skip this step.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
[...]

Issue #756, opened 2020-12-18
"View as" UI design


dkoo:
From @thomasguillot via: https://github.com/Automattic/newspack-plugin/pull/741#issuecomment-748185372

> For the Groups, I suggest we display them all (with a max-height and scrollable to avoid a massive list) and can be selected via a checkbox. I don't find the autocomplete extremely useful. It can be hard sometimes to remember all the Groups someone might have.

> Also, I'd move the "Preview" tab before "Settings".

dkoo:
Oh, haha. This is a dupe of #754.

PR #2254, opened 2023-01-27
WCS & Stripe: add payment metadata and update renewal description

adekbadek:
### All Submissions:

WC Stripe Gateway plugin adds little metadata to any Stripe payment. It's not possible to learn that a payment was related to a specific subscription. WCSG only adds the `order_id` metadata and indicates a WC Subscription with a `payment_type='recurring'` metadata.

To make any third-party integrations more aware of what WCS is doing in Stripe:

- `subscription_id` will be added to any payments which are WC Subscriptions renewals
- `subscription_status` will be added, either `created` or `renewed`
- additionally, for Newspack-originating subscriptions, the payment description will be updated to `Newspack <frequency> Donation (Order <id>, Subscription <id>)` (e.g. `Newspack Monthly Donation (Order #756, Subscription #757)`)

1. Set up Stripe, ensure the site is reachable by Stripe webhooks
1. Ensure the WC suite is active, and "Stripe – Credit Card (Stripe)" set as the only payment method in WC
1. Set the environment variable `NEWSPACK_USE_WC_SUBSCRIPTIONS_WITH_STRIPE_PLATFORM` to `true` (see https://github.com/Automattic/newspack-plugin/pull/2251)
2. Make a recurring donation using SDB 
3. Open the resulting WC Subscription in WP Admin an renew it manually (via the "Subscription actions" meta box) 
4. Go to Stripe dashboard, inspect the initial payment and the renewal payment – both should have `subscription_id` meta set (to the applicable value), the first one should have the `subscription_status` meta set to `created`, and the second one to `renewed`.  Both should have `origin` meta set to `newspack`.

### Other information:

### DIFF
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -990,7 +990,8 @@ class WooCommerce_Connection {
 			}
 			// Add the description only for Newspack-created subscriptions.
 			if ( self::CREATED_VIA_NAME === $subscription->get_created_via() ) {
-				$post_data['description'] = self::create_payment_description(
+				$post_data['metadata']['origin'] = 'newspack';
+				$post_data['description']        = self::create_payment_description(
 					[
 						'order_id'        => $order->get_id(),
 						'subscription_id' => $subscription_id,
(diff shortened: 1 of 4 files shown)


leogermani:
That's awesome, thanks!

Would it be possible/easy to add what payment of the subscription this is? (if it's the 1st payment, second, etc... just a number `1`, `2`, `3`...?)

leogermani:
ah..

another thing we identified is that we could add a `origin=>newspack` metadata.

I know we have referer... but this would be more generic. See pamTN9-5II-p2#comment-7375

adekbadek:
> The primary goal of that would be to easily identify new subscriptions from ongoing ones.

There will be `subscription_status` meta with `created` or `renewed` value, would that suffice?

> another thing we identified is that we could add a origin=>newspack metadata.

Added in 5e0598fcdf439b5079b914ca23b1804f5493a73c.

leogermani:
> There will be subscription_status meta with created or renewed value, would that suffice?

Yes, that's good

adekbadek:
> For a regular payment during a subscription what would this look like?

That would be `renewed`.

leogermani (approved):

Issue #1531, opened 2022-02-25
PluginSettings: Disable save button until a setting is changed


thomasguillot:
The "Save Settings" button is very prominent now in the header.

We should disable it until one of the settings in its card is changed.

See https://github.com/Automattic/newspack-plugin/pull/1525

PR #2259, opened 2023-01-31
Feat: Add disable functionality to save settings button (#1531)

hbhalodia:
### All Submissions:

- By adding this PR, we can ensure that save settings button is disabled until user changes any settings in a particular section. Before this commit save settings button was always enabled. This would only enable the save settings button for the section whose inner settings are changed, that would ensure all the sections on the page are async.
- Also on this PR in file change `index.js`, we need to look upon the comment added on `handleSectionUpdate` function, because we have kept all sections save settings button async, hence user can change all settings at once and can try to update one by one which would override his settings because on update we are fetching the saved settings and adding it again on the response. We can remove this behavior by removing the setState again for saved settings instead use prev state only.

1. Checkout to the current branch.
2. Head to Advertising --> Settings page.
3. On first load you can see the button is disabled, Change any section settings you can see that section button is enabled, On save again button gets disabled.
4. You can check with multiple section and each section works async.

### Screencast for the current change
[Newspack Disable Settings Button](https://user-images.githubusercontent.com/58802366/212673315-bfd6e0b7-209b-444b-a137-b54f2363e7b2.webm)

### Other information:


assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -100,7 +110,7 @@ const SettingsSection = props => {
 					'buttons',
 					<Button
 						variant="primary"
-						disabled={ disabled }
+						disabled={ sectionSettingsButtonDisabled }
```
leogermani, line 113: I think this should be `disabled || sectionSettingsButtonDisabled`. What happens now is that when you click save the button is still active while processing the request

leogermani (changes requested):
Thank you for this!

It works as described but I left one comment about the button not being disabled while saving.

Also, it doesn't seem to apply to all SettingsSection components. If you go to Engagement > Social, you'll see that this will not work with the sections there. Apparently Ads settings are created in a very different way. cc @miguelpeixe

assets/components/src/plugin-settings/SettingsSection.js
hbhalodia: Hi @leogermani, The above feedback is being addressed. Thanks.

hbhalodia:
Hi @leogermani, You need me to update for engagement > social settings also or we can keep this just for ads, as you said ads settings are created differently? Let me know on this. Thanks.

hbhalodia:
Hi @leogermani, I have added for engagement settings also. We need to apply this change where `PluginSettings` component is being used. Thanks.

assets/components/src/plugin-settings/SettingsSection.js
```diff
@@ -52,8 +52,18 @@ const getControlType = setting => {
...
+		error,
+		sectionKey,
+		active,
+		title,
+		description,
+		fields,
+		disabled,
+		onChange,
+		onUpdate,
+		sectionSettingsButtonDisabled,
```
miguelpeixe, line 65: We're already in the context of the `SettingsSection` component, for readability this could be called `saveDisabled`.

miguelpeixe (changes requested):
Thank you for your contribution, @hbhalodia!

May I suggest a different approach? The `SettingsSection` component is aware of the value changes and saves so we don't need to rely on another component to set this:

https://github.com/Automattic/newspack-plugin/blob/6904356f15d10c49b81e1eb1ccfb40452d3ac871/assets/components/src/plugin-settings/SettingsSection.js#L71-L73

https://github.com/Automattic/newspack-plugin/blob/6904356f15d10c49b81e1eb1ccfb40452d3ac871/assets/components/src/plugin-settings/SettingsSection.js#L104-L106

With the current approach, every component that uses this will have to reimplement the same (or very similar) logic for a consistent experience. Case in point, the Ad Refresh Control is missing it, so that's how my Ads settings render currently:

image

assets/components/src/plugin-settings/index.js
```diff
@@ -90,8 +90,6 @@ class PluginSettings extends Component {
 			},
 		} )
 			.then( settings => {
-				this.setState( { settings, error: null } );
-
```
hbhalodia, line 94: @miguelpeixe This line is removed because, When user updates more than one settings and did not hit the `save settings` button one after other, then when they change all the settings and then hit any one of the `save settings` button then all the changes he made to settings would get lost because this line fetches the saved settings. Hence, we need to take care of that thing, so instead adding fetched settings we just show settings what are saved in state, which overcomes this issue.

hbhalodia:
Thanks! @miguelpeixe for the approach. I have worked on this approach and achieved the results for all settings. No need to reimplement the same logic again on each component. I have updated the PR with the new approach. You can review now. 

Thanks!

assets/components/src/plugin-settings/SettingsSection.js
hbhalodia: I have added the suggestion. Thanks!

assets/components/src/plugin-settings/index.js
miguelpeixe: Nice catch!

miguelpeixe (approved):
Thank you for revising, @hbhalodia!

It's working well 👌

PR #2260, opened 2023-01-31
feat: require all woo plugins for RAS

dkoo:
### All Submissions:

Separates out the plugin requirements check from options-based RAS settings, and requires all WooCommerce plugins and extensions for RAS.

Closes 1203695565716937/1203706216921098.

1. On a site with Reader Activation enabled, visit **Newspack > Engagement > Reader Activation**.
2. Disable the following plugins one-by-one, and confirm that the Reader Activation settings page is replaced by a plugin installer component if any of them is missing:

Screen Shot 2023-01-31 at 3 35 21 PM

3. Confirm that the settings page reappears automatically after installing/activating all required plugins.

### Other information:

### DIFF
--- a/assets/wizards/engagement/views/reader-activation/index.js
+++ b/assets/wizards/engagement/views/reader-activation/index.js
@@ -37,7 +37,10 @@ export default withWizardScreen( () => {
 		apiFetch( {
 			path: '/newspack/v1/wizard/newspack-engagement-wizard/reader-activation',
 		} )
-			.then( setConfig )
+			.then( ( { config: fetchedConfig, pluginsStatus } ) => {
+				setHasPlugins( pluginsStatus );
+				setConfig( fetchedConfig );
+			} )
 			.catch( setError )
 			.finally( () => setInFlight( false ) );
 	};
@@ -53,18 +56,7 @@ export default withWizardScreen( () => {
 			.catch( setError )
 			.finally( () => setInFlight( false ) );
 	};
-	const fetchPluginStatus = () => {
-		setError( false );
-		setInFlight( true );
-		apiFetch( {
-			path: '/newspack/v1/wizard/newspack-engagement-wizard/reader-activation-plugins',
-		} )
-			.then( setHasPlugins )
-			.catch( setError )
-			.finally( () => setInFlight( false ) );
-	};
 	useEffect( fetchConfig, [] );
-	useEffect( fetchPluginStatus, [] );
 	useEffect( () => {
 		apiFetch( {
 			path: '/newspack/v1/wizard/newspack-engagement-wizard/newsletters',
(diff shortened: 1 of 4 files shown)


assets/wizards/engagement/views/reader-activation/index.js
```diff
@@ -27,6 +27,7 @@ export default withWizardScreen( () => {
 	const [ config, setConfig ] = useState( {} );
 	const [ error, setError ] = useState( false );
 	const [ isActiveCampaign, setIsActiveCampaign ] = useState( false );
+	const [ hasPlugins, setHasPlugins ] = useState( false );
```
adekbadek, line 30: Since `hasPlugins` is `false` by default, the UI starts in missing-plugins state and then transitions to the intended initial state only after the request completes. The default value should rather be `null`/`undefined`. 

```suggestion
	const [ hasPlugins, setHasPlugins ] = useState( null );
```

assets/wizards/engagement/views/reader-activation/index.js
```diff
@@ -52,7 +53,18 @@ export default withWizardScreen( () => {
 			.catch( setError )
 			.finally( () => setInFlight( false ) );
 	};
+	const fetchPluginStatus = () => {
+		setError( false );
+		setInFlight( true );
+		apiFetch( {
+			path: '/newspack/v1/wizard/newspack-engagement-wizard/reader-activation-plugins',
```
adekbadek, line 60: Why another request – the plugin status could be a part of the response from `GET /newspack-engagement-wizard/reader-activation` request.

adekbadek (changes requested):

assets/wizards/engagement/views/reader-activation/index.js
dkoo: It was that way originally, but doing so tied the plugins check to config that's fetched from and saved to the options table. So what was happening was that the check only occurred once and then the success state was saved to `wp_options` the first time all required plugins were available.

I tried separating the check from the other options in the same request, but it resulted in a lot of unnecessary re-renders of the component, so I thought separating the plugin status check from the options table settings would ensure that we were always getting the current state of the plugins without causing a re-render if not needed.

adekbadek: > I tried separating the check from the other options in the same request, but it resulted in a lot of unnecessary re-renders of the component

I don't understand. Have you tried it like so – https://github.com/Automattic/newspack-plugin/commit/6745d23d2eefc8fe52f002b4e928fda5c1594201 ?

dkoo: No, no I didn't. 😆 I was trying to fit the plugin status into the existing `config` object, but your solution is much better.

dkoo: cca1393 implements the single fetch and gets rid of the extra REST route. No extra re-renders!

yogeshbeniwal:
Couple of related queries:

- Is Woocommerce and additional plugins not required with Stripe as selected platform?
- What is recommended platform selection, Is it newspack or stripe?

dkoo:
> Is Woocommerce and additional plugins not required with Stripe as selected platform?

WooCommerce + additional plugins are required if using Reader Activation features. This is because WooCommerce's user management is a core dependency of Reader Activation registration and user account features.

> What is recommended platform selection, Is it newspack or stripe?

We are moving toward merging the two strategies and making WooCommerce a more central part of the Reader Revenue experience, but no change is required at this time if you're using Stripe.

adekbadek (approved):

PR #2156, opened 2022-11-29
Feat/perfmatters defaults

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds default setup for the Perfmatters plugin, behind a URL-param feature flag. 

Closes 1202855962089497-as-1203286553017184

### How to test the changes in this Pull Request:

1. Install Perfmatters plugin
2. Visit the plugin settings, observe some things configured by default
3. Visit the front-end, observe no optimisations are at play (e.g. all JS & CSS is loaded on page load)
4. Visit the front-end with `?newspack-perfmatters-defaults` param, observe the optimisations visible in the plugin settings are applied

And a small bonus:

1. On `master`, visit homepage, observe the `my-account.css` file is being loaded 
2. Switch to this branch and with RAS enabled, observe the file is not loaded on the homepage (or anywhere besides `/my-account` page)
3. Visit the `/my-account` page – observe the aforementioned CSS file is loaded 

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2271, opened 2023-02-08
Fix/perfmatters defaults adjustments

adekbadek:
### All Submissions:

- Removes feature flag on Perfmatters defaults (https://github.com/Automattic/newspack-plugin/pull/2156)
- Updates the defaults
- Removes superflous use of [`default_option_perfmatters_options` filter](https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-includes/option.php#L188) – `option_perfmatters_options` filter is [ran anyway later](https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-includes/option.php#L253).

1. Install the Perfmatters plugin
2. Observe that the defaults are applied (see https://github.com/Automattic/newspack-plugin/pull/2156)

### Other information:

### DIFF
--- a/includes/plugins/class-perfmatters.php
+++ b/includes/plugins/class-perfmatters.php
@@ -85,6 +85,15 @@ class Perfmatters {
 	}
 
 	/**
+	 * Selectors to exclude from the "Unused CSS" feature.
+	 */
+	private static function unused_css_excluded_selectors() {
+		return [
+			'body',
+		];
+	}
+
+	/**
 	 * URLs to preconnect to.
 	 *
 	 * @param array $existing_urls Existing URLs to filter out.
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Works for me. The only comment I have is that we are not actually setting defaults, we are overwriting the plugin settings.

So when we activate Perfmatters, our defaults are applied, but we can't change it. I wonder if we should explain this better. For example we could add a Notice to the Perfmatters admin page saying that Newspack is overwritting some of those params?

adekbadek:
Good idea, added a notice in 604be7e17644ad44f76c478b5f7904239c394b9b

image

earcos:
@leogermani adding all the Newspack CSS files to the exclude CSS list beats part of the purpose of Perfmatters and LCP values on Google Search Console are beyond what Google considers acceptable (2.5+ seconds).

What's the best way to add the `NEWSPACK_IGNORE_PERFMATTERS_DEFAULTS` flag?

Thanks a lot

adekbadek:
That would be in your `wp-config.php` file. If you're a Newspack customer, a TAM or developer can do that for you.

PR #2276, opened 2023-02-13
fix(webhooks): check if endpoint exists

adekbadek:
### All Submissions:

Fixes a small webhooks issue. 

1. Create a webhook in the Engagement wizard and then delete it 
2. Trigger an action
3. On `master`, PHP will warn that `$endpoint` is null or empty
4. On this branch, there should be no such warnings

### Other information:

### DIFF
--- a/includes/data-events/class-webhooks.php
+++ b/includes/data-events/class-webhooks.php
@@ -580,10 +580,12 @@ final class Webhooks {
 		if ( ! $endpoint ) {
 			self::add_request_error( $request_id, __( 'Endpoint not found.', 'newspack' ) );
 			self::kill_request( $request_id );
+			return;
 		}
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
Thank you for catching that!

On a further look, looks like we need to return early here instead of checking twice if `$endpoint` exists. The previous conditional block should actually `return;` after calling `kill_request()`. The same goes for the `disabled` block you're tweaking.

Does that make sense to you?

adekbadek:
It does, like so? 14e94625058d11cb8b633cae32b8dc6a6e8a6e1a

miguelpeixe:
Yes! The first change becomes superfluous then because on `! $endpoint` it'll return.

miguelpeixe (approved):
Thank you!

PR #2278, opened 2023-02-13
feat(my-account): display stripe billing link regardless of RR platform

adekbadek:
### All Submissions:

Current behavior: The Stripe billing portal link is only shown if Stripe is the current active Reader Revenue platform. This means if the platform was Stripe at one point but subsequently changed, readers who donated via Stripe can no longer self-serve their own subscription management.

Desired behavior: Rather than tie the portal link to the active RR platform, the logic should check whether the currently logged-in reader has any Stripe subscriptions associated with their account.

Closes 1203869877018965-as-1203870488073009

1. On `master`,
1. Set RR platform to Stripe and make a recurring donation. Log to "My Account" page 
2. Observe the "Billing" link is among the menu items
3. Switch RR platform to Newspack
4. As the logged-in donor, observe the "Billing" item is missing
5. Switch to this branch, reload the page – observe the "Billing" item is there

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -78,8 +78,12 @@ class WooCommerce_My_Account {
 	 * @param array $items Items.
 	 */
 	public static function my_account_menu_items( $items ) {
-		if ( ! Donations::is_platform_stripe() ) {
-			return $items;
+		// If the user has a Stripe customer ID, they're a Stripe customer (regardless of the donations platform).
+		// Add a nav item for Stripe's billing portal.
+		$stripe_customer_id = self::get_current_user_stripe_id();
+		if ( false !== $stripe_customer_id ) {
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
Using the new logic the customer with a Stripe subscription but under the "Newspack" platform will also have their WooCommerce links removed (Addresses and Downloads).

Perhaps it should be decoupled so there's a method to handle each context (Stripe subscription exists vs. Stripe is the defined platform).

adekbadek:
> Using the new logic the customer with a Stripe subscription but under the "Newspack" platform will also have their WooCommerce links removed (Addresses and Downloads).

This is independent of the platform – hiding these fields is a RAS feature. Before, it was assumed that RAS & Stripe RR platform go together, but we've since made RAS compatible with WooCommerce.

miguelpeixe (approved):

miguelpeixe:
We might want to revisit this now that WC is supported on RAS. "Addresses" is a standard page for editing your billing information. Non-blocking for this PR, though

PR #2279, opened 2023-02-13
feat: allow external links in dashboard via a filter

adekbadek:
### All Submissions:

Allows the dashboard links to be filterable, so a plugin may add something.

Ping me for specific instructions. 

### Other information:

### DIFF
--- a/assets/wizards/dashboard/views/dashboardCard.js
+++ b/assets/wizards/dashboard/views/dashboardCard.js
@@ -30,7 +30,7 @@ class DashboardCard extends Component {
 	 * Render.
 	 */
 	render() {
-		const { name, description, slug, url } = this.props;
+		const { name, description, slug, url, is_external } = this.props;
 		const iconMap = {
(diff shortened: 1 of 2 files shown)


leogermani (changes requested):
I'm able to add a new item using the filter. I can't choose the icon though. Passing `is_external` correctly opens it in a new window.

I think we just need to add a doc block to the filter and make sure to document this new accepted property

adekbadek:
Added a doc comment in aeb1d7631cc94040813bbd5d44ebc99fa949e140. Setting the icon is not implemented ATM, the icon will fall back to the question mark.

leogermani (approved):
> icon will fall back to the question mark.

In my case the fallback was the "plugin" icon

adekbadek:
> In my case the fallback was the "plugin" icon

Ah that's correct! I've been testing with an item with `support` slug, which is already mapped to the question mark icon.

PR #2288, opened 2023-02-17
Alpha release Feb 17

adekbadek:
A new alpha release.


adekbadek:
Already done in https://github.com/Automattic/newspack-plugin/pull/2287, `1.104.0-alpha.1` is out.

PR #2298, opened 2023-02-20
Stripe Subscriptions to WC subscriptions migrator

adekbadek:
### All Submissions:

A migration script that turns Stripe Subscriptions into WooCommerce Subscriptions.

Closes 1204005426582602-as-1203870488073017

1. Set up the site to use Stripe as the Reader Revenue platform and ensure the test site is reachable for webhooks
2. Ensure the WooCommerce suite is active, so the Stripe data will be synchronised to WC
3. Make a recurring donation twice, observe two WC Subscriptions were created to shadow (sync) the Stripe subscriptions
4. Delete one of the WC Subscriptions (ideally via CLI – `$ wp post delete <sub-id>`)
5. Run the migration in dry-run mode – `$ wp newspack stripe sync-stripe-subscriptions-to-wc --dry-run`
6. Observe the output and note that neither Stripe nor WooCommerce subscriptions were updated
7. Now run the migration for real – `$ wp newspack stripe sync-stripe-subscriptions-to-wc`
8. Observe the output, it should successfully migration both and cancel the Stripe subscriptions
9. Verify that the Stripe subscriptions have been cancelled in the [Stripe dashboard](https://dashboard.stripe.com)
10. Observe that the WC Subscriptions have received helpful order notes about the process ("This subscription has been migrated…") 
11. Observe that both WC Subscriptions have `cancelled-newspack-stripe-subscription-id` metadata, with the Stripe subscription ID as value

### Other information:

### DIFF
--- a/includes/reader-revenue/stripe/class-stripe-sync.php
+++ b/includes/reader-revenue/stripe/class-stripe-sync.php
@@ -297,7 +297,17 @@ class Stripe_Sync {
 			'sync-to-esp' => false, // Sync data to the ESP.
 			'dry-run'     => false,
 		];
-		$passed_args  = array_merge( $default_args, $assoc_args );
+
+		\WP_CLI::log(
+			'
+
+Running data backfill from Stripe...
+
+		'
+		);
+
+
+		$passed_args = array_merge( $default_args, $assoc_args );
 		if ( false !== $passed_args['dry-run'] ) {
 			\WP_CLI::warning( __( 'This is a dry run, no changes will be made.', 'newspack' ) );
 		}
@@ -363,7 +373,15 @@ class Stripe_Sync {
 		$force_override = ! empty( $assoc_args['force'] );
 		$batch_size     = ! empty( $assoc_args['batch-size'] ) ? intval( $assoc_args['batch-size'] ) : 10;
 
-		$customers = self::get_batch_of_customers_for_stripe_connect_to_stripe( $batch_size );
+		\WP_CLI::log(
+			'
+
+Running Stripe Connect to Stripe Subscriptions Migration...
+
+		'
+		);
+
+		$customers = self::get_batch_of_customers_with_subscriptions( $batch_size );
 		while ( $customers ) {
 			$customer = array_shift( $customers );
 
@@ -481,7 +499,7 @@ class Stripe_Sync {
 				);
 				\WP_CLI::log( sprintf( '    * Created subscription: %s with next renewal at %s', $subscription->id, gmdate( 'Y-m-d', $existing_subscription->current_period_end ) ) );
 			} catch ( \Throwable $e ) {
-				\WP_CLI::error( sprintf( 'Failed to create subscription: %s', $e->getMessage() ) );
+				\WP_CLI::warning( sprintf( 'Failed to create subscription: %s', $e->getMessage() ) );
 			}
 
 			try {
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Working really well as described on my end. Seamlessly migrates Stripe subscriptions back to Woo so that it's like all those customers subscribed via Woo in the first place. This will be great for sites that had to switch from Stripe to Woo and are currently stuck managing data in both dashboards!

adekbadek:
Thanks @dkoo ! I have to ask you for another look, as I've uncovered an issue and fixed it in 043e44f30c4dbc1f18a291be139a4a51275e5ee6.

### Here's the gist of it: 

- To charge customers, Stripe had `Sources` API and then later introduced `PaymentMethods` API, which is more encouraged
- Since `woocommerce-gateway-stripe` uses the old `Sources` API, #2251 – which introduces a tighter integration with it – switched from `PaymentMethods` to `Sources`
- This means that any Stripe donors created before #2251 (non on live sites, since the changes are awaiting release) will not have any sources, and thus would be not migrateable
- 043e44f30c4dbc1f18a291be139a4a51275e5ee6 fixes that, so that the pre-#2251 donors will be handled

### What this means for this PR

To test the full scope of changes, travel back in git history to 887aab6d6d882e15f9de42b48a541b2f5c28142f on this repository and to [02cd767a40956aad0f1f99f558a56ea5d58b1f82](https://github.com/Automattic/newspack-blocks/commit/02cd767a40956aad0f1f99f558a56ea5d58b1f82) on the `newspack-blocks` repository, make a donation, then travel back to the present state and follow the test the migration. 

### What this means for #2251 

Switching from `Sources` to `PaymentMethods` API was not necessary, as it turns out. I haven't dived deep enough into `woocommerce-gateway-stripe` to see that what's stored as `_stripe_source_id` post meta [may be a Stripe `Source`, `PaymentMethod`, or a `Card`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/compat/trait-wc-stripe-subscriptions.php#L506-L512). The meta name and the fact `woocommerce-gateway-stripe` uses `Sources` misled me.

includes/reader-revenue/stripe/class-stripe-sync.php
```diff
@@ -454,36 +520,139 @@ protected static function process_customer_for_stripe_connect_to_stripe( $custom
...
+							$subscription_id              = $wc_transaction_creation_data['subscription_id'];
+							$subscription                 = \wcs_get_subscription( $subscription_id );
+							\WP_CLI::success( sprintf( 'Created subscription: %s with next renewal at %s', $subscription_id, gmdate( 'Y-m-d', $existing_subscription->current_period_end ) ) );
+						}
+					}
+
+					if ( $subscription ) {
+						// Add the cancelled Stripe subscription ID to the meta data, so it can be found later.
+						$subscription->add_meta_data( 'cancelled-' . WooCommerce_Connection::SUBSCRIPTION_STRIPE_ID_META_KEY, $existing_subscription->id );
+						$subscription->add_order_note(
```
dkoo, line 600: This note is added to orders on dry runs, too. We should probably only add the note if it's not a dry run.

dkoo (approved):
Tested the migration with a subscription made with the `Sources` API and it worked well, so I think we're good on that. I noticed one more behavior which I think we should change, but it's a minor one, so pre-approving.

includes/reader-revenue/stripe/class-stripe-sync.php
adekbadek: Thanks for catching that! Fixed in cdfed16e218b91491ce98f093152ec665ceacc03

PR #2299, opened 2023-02-21
Update/normalize donation events

leogermani:
### All Submissions:

Normalizes the `donation_new` event so it works the same way in both platforms.

Before the change in this PR, donations processed through Woocommerce would trigger a `donation_new` event before they were confirmed, while in Stripe it would only be triggered after confirmation.

Now the event is only triggered after confirmation in both platforms.

We keep the current WooCommerce listener as a new event called `donation_order_processed` as suggested in p1676562193774819-slack-1676536201.225739 (even though I'm not sure we need it)

1. Add the following snippet to your site:

```PHP
Data_Events::register_handler( 'debug_donation_new_events', 'donation_new' );

function debug_donation_new_events( $timestamp, $data, $user_id ) {
	error_log( print_r( $data, true ) );
}

Data_Events::register_handler( 'debug_donation_subscription_new_events', 'donation_subscription_new' );

function debug_donation_subscription_new_events( $timestamp, $data, $user_id ) {
	error_log( print_r( $data, true ) );
}
```
2. Set up Stripe as the Reader Revenue platform
3. Make donations
4. Watch the logs and make sure you see the correct events being triggered at the correct times, with the correct data
5. Do the same for the Newspack platform and confirm they are consistent
6. Make sure to test with cards that will return failed payments to make sure events are not triggered in those situations

### Other information:

### DIFF
--- a/includes/data-events/listeners.php
+++ b/includes/data-events/listeners.php
@@ -108,6 +108,32 @@ Data_Events::register_listener(
 	}
 );
 
+/**
+ * For when there's a new donation processed through WooCommerce.
+ */
+Data_Events::register_listener(
+	'newspack_donation_order_processed',
+	'donation_order_processed',
+	function( $order_id, $product_id ) {
+		$order = \wc_get_order( $order_id );
+		if ( ! $order ) {
+			return;
+		}
+		return [
+			'user_id'       => $order->get_customer_id(),
+			'email'         => $order->get_billing_email(),
+			'amount'        => (float) $order->get_total(),
+			'currency'      => $order->get_currency(),
+			'recurrence'    => get_post_meta( $product_id, '_subscription_period', true ),
+			'platform'      => 'wc',
+			'platform_data' => [
+				'order_id'   => $order_id,
+				'product_id' => $product_id,
+			],
+		];
+	}
+);
+
 
 /**
  * For when a Subscription is confirmed.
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Approving because it works, but I noticed something that may be outside the scope of this PR—when using the Stripe RR platform, the `user_id` value always seems to be `0`. Tested on both an existing and a new browser session.

leogermani:
> Approving because it works, but I noticed something that may be outside the scope of this PR—when using the Stripe RR platform, the user_id value always seems to be 0. Tested on both an existing and a new browser session.

I could not reproduce this. Did you have RAS enabled when you tested? This will happen if RAS is not enabled because a user won't be created.

I really want @miguelpeixe to have a look at this, but I'll merge it if I get to the next task that depends on it so I'm not blocked. We can always revisit/revert it if Miguel finds a problem.

dkoo:
> I could not reproduce this. Did you have RAS enabled when you tested?

RAS is enabled on my site, and the user ID is populated when Newspack is the Reader Revenue platform. It's only `0` for Stripe.

miguelpeixe (approved):
Working as described

> when using the Stripe RR platform, the user_id value always seems to be 0. Tested on both an existing and a new browser session.

@dkoo Perhaps you are unauthenticated and donating with an already registered email? The donation will not be tied to the user ID if that's the case.

leogermani:
> The donation will not be tied to the user ID if that's the case.

Why not? Shouldn't it?

dkoo:
> Why not? Shouldn't it?

That was my assumption as well—I was assuming if using an email address that is known to be associated with a user ID, that user ID should be sent with the data.

dkoo:
Thanks for the additional info—let's leave as-is for now

PR #2301, opened 2023-02-24
feat: Remove the campaign rendered event

leogermani:
### All Submissions:

We should not trigger an event when a campaign is rendered.

1. This happens all the time, and since we are using Action Scheduler to trigger events, this could overflow our database with events
2. This event does not tell us anything useful. The campaigns are rendered all the time, even when the user doesn't match the segment and won't see the popup. So we can't use this to build funnels and calculating conversion rates.
3. The page output will be cached. So the counts of renders will not match the times the campaign was rendered for users. Again, this count won't be useful in our analytics to build funnels.

1. Nothing really. Just make sure you don't see `[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "campaign_interaction" via Action Scheduler.` in the logs when just loading a page.

### Other information:

### DIFF
--- a/includes/data-events/class-popups.php
+++ b/includes/data-events/class-popups.php
@@ -23,12 +23,6 @@ final class Popups {
 	 */
 	public static function init() {
 		Data_Events::register_listener(
-			'newspack_campaigns_after_campaign_render',
-			'campaign_interaction',
-			[ __CLASS__, 'campaign_rendered' ]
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Agreed, I don't think this event is really useful in any real way.

PR #2310, opened 2023-03-02
Add/popups donation events

leogermani:
### All Submissions:

Adds the popup events for Donations

Setup
1. Make sure you have the latest version (`master`) of the popups and blocks plugins
2. Not required, but I suggest checking out this fix on the newsletters plugin -> https://github.com/Automattic/newspack-newsletters/pull/1101
3. add this to your site:
```PHP
add_action(
	'plugins_loaded',
	function() {
		Data_Events::register_handler( 'debug_popups_events', 'campaign_interaction' );

	}
);

function debug_popups_events( $timestamp, $data, $user_id ) {
	error_log( print_r( $data, true ) );
}
```
4. Setup a Donation block inside a popup
5. Setup a Donation block in a page, not inside a popup

Now let's test

Now we are going to make donations in the following ways:
* With Stripe and Newspack as Reader Revenue platform
* For each platform, one recurring donation and one one-time donation
* For each type of donation, one with a valid card (4242424242424242) and one with a failing card (4000000000000002)
* Also, for each platform, make one donation from inside the block that is not in a popup

Inspect the logs while you do the donation. For each donation you should see 2 `campaign_interaction` events being fired
```
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "campaign_interaction" via Action Scheduler.
```
(for Stripe, the second one might take a little while, as it happens when the Stripe server reaches your site via webhook)

When the events handlers are fired, you will see one event with `action = form_submission` and one with `form_submission_sucess`or `form_submission_failure` depending on the case.

Here's an example:
```
(
    [campaign_id] => 143
    [campaign_title] => Inline Donation (secondary)
    [campaign_frequency] => always
    [campaign_placement] => inline
    [campaign_has_registration_block] => 
    [campaign_has_donation_block] => 1
    [campaign_has_newsletter_block] => 
    [action] => form_submission_success
    [donation_amount] => 15.76
    [donation_currency] => BRL
    [donation_recurrence] => month
    [donation_platform] => stripe
    [referer] => /2023/02/07/egestas-praesent-curabitur-aptent-nunc-vel-laoreet-neque-tristique-pretium-condimentum/
)
```
Make sure all events have the right information.

Make sure donations done from the block on the page does not trigger any `campaign_interaction` event

Known issue: when using woocommerce and the transaction fails, yo are are able to try again. If the transaction fails again a second time, the form submission failure will not be triggered. I could not find a hook for that. Since woocommerce uses the same order and its status is already failed, the statuc change action is not fired

### Other information:

### DIFF
--- a/includes/class-salesforce.php
+++ b/includes/class-salesforce.php
@@ -539,7 +539,7 @@ class Salesforce {
 		$order_id = isset( $args['id'] ) ? $args['id'] : 0;
 		$order    = \wc_get_order( $order_id );
 
-		if ( is_wp_error( $response ) ) {
+		if ( is_wp_error( $response ) && is_a( $order, 'WC_Order' ) ) {
 			$order->add_order_note(
 				sprintf(
 					// Translators: Note added to order when sync is unsuccessful.
@@ -547,7 +547,7 @@ class Salesforce {
 					$response->get_error_message()
 				)
 			);
-		} else {
+		} elseif ( is_a( $order, 'WC_Order' ) ) {
 			$order->add_order_note( __( 'Order successfully synced to Salesforce.', 'newspack' ) );
 		}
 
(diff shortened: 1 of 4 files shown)


adekbadek (changes requested):
With both platforms, the `form_submission_success` for one-time donations has empty `donation_recurrence`. On Newspack platform, `form_submission_failure` for one-time donations also is missing the value.

leogermani:
@adekbadek Thanks for catching that. It's fixed!

miguelpeixe:
@leogermani let's try to keep the [documentation](https://github.com/Automattic/newspack-plugin/blob/master/includes/data-events/README.md) up to date so we keep track of what we're building

leogermani:
> @leogermani let's try to keep the [documentation](https://github.com/Automattic/newspack-plugin/blob/master/includes/data-events/README.md) up to date so we keep track of what we're building

I will. I promise

PR #2333, opened 2023-03-08
fix: show RAS My Account button in Customizer preview

dkoo:
### All Submissions:

While testing https://github.com/Automattic/newspack-theme/pull/2051, I realized that the "Sign In"/"My Account" button from RAS isn't showing up in the Customizer preview. I figure it probably should since it likely has a largeish impact on how the header appears.

1. On `master`, enable RAS and the "Enable Sign In/Account link" option.
2. Observe that the button appears in the header on the front-end in a new session, but not in the Customizer while logged in as an admin.
3. Check out this branch and confirm that now it does appear in the Customizer.

### Other information:

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1308,16 +1308,16 @@
         },
         {
             "name": "myclabs/deep-copy",
-            "version": "1.11.0",
+            "version": "1.11.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/myclabs/DeepCopy.git",
-                "reference": "14daed4296fae74d9e3201d2c4925d1acb7aa614"
+                "reference": "7284c22080590fb39f2ffa3e9057f10a4ddd0e0c"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
Hm, I _do_ see the link on `master` in Customizer:

image

Sidenote: there are `composer.lock` changes in this PR, which is probably not intended.

dkoo:
Haven't had a chance to look more deeply into this, but since it only seems to be an issue for me I'm going to close this for now

laurelfulford:
Gah, just catching up on this super-late! 

@dkoo I think it depends on the breakpoint: the mobile button is visible for me in the Customizer when the window's smaller (like in @adekbadek's screenshot), but when the window is wider, it's not: 

image

image

The login button is using the 'desktop' breakpoint, which is pretty much 1168px -- part of the issue is that the Customizer sidebar can eat up a decent amount of space, so you're pretty likely to see some smaller screen styles -- especially on a laptop -- even if your viewing the full screen of the site. 

If you wanted to give this PR another run, I'm happy to take a look!

PR #2337, opened 2023-03-09
feat: Add popup events to GA4

leogermani:
### All Submissions:

Connects the Popup events to GA4.

Note that the Popup events have already been tested. This only connects them to GA4. It also adds a little more detail to the newsletter subscription events.

This PR also fixes a bug in the `donation_new` event and the `form_submission_success` event for donations. Both when using Stripe as platform.

Since the new donation in Stripe is triggered when Stripe server hits our webhook, the GA Client and Session IDs were not bein fetched by the cookie. So I had to add these information in the payment metadata and retrieve them for the order when they are marked as completed.

Since we changed the structure of the data and also fixed a bug, we should do an extensive tests of as many events as possible.

```
```
Now, let's test:

```
```
4. Checkout the newsletters plugin on https://github.com/Automattic/newspack-newsletters/pull/1111
5. Setup a popup with a Newsletter subscription form
6. Subscribe to a newsletter via this popup
8. Visit the Google Analytics Realtime dashboard
9. See the `campaign_interaction` events coming in

Captura de tela de 2023-03-09 13-15-59

11. Inspect them and see check its attributes to see if the values are correct.

Captura de tela de 2023-03-09 13-51-25

### Other information:

### DIFF
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -186,7 +186,7 @@ When a user interacts with a Newspack Popup's campaign prompt.
 | `campaign_blocks`    | `array`  | Array containing the blocks that are inside the campaign. Only 3 blocks are tracked: `donation`, `registration` and `newsletters_subscription` |
 | `action`             | `string` | `form_submission`, `form_submission_success` or `form_submission_failure`                                                                      |
 | `action_type`        | `string` | `donation`, `registration` or `newsletters_subscription`                                                                                       |
-| `interaction_data`   | `array`  | Depending on the actino type, it will return different information about the interaction.                                                      |
+| `interaction_data`   | `array`  | Depending on the action type, it will contain different information about the interaction.                                                      |
 
 #### Possible values for `interaction_data`
 
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -71,37 +71,75 @@ When a reader updates their lists subscription from Newspack Newsletters.
 | `user_id`       | `integer`  |
 | `email`         | `string`   |
 | `provider`      | `string`   |
-| `email`         | `string`   |
 | `lists_added`   | `string[]` |
 | `lists_removed` | `string[]` |
+
+### `woocommerce_donation_order_processed`
+
+For when there's a new donation processed through WooCommerce.
+
+| Name            | Type     | Obs                                                    |
+| --------------- | -------- | ------------------------------------------------------ |
+| `user_id`       | `int`    |                                                        |
+| `email`         | `string` |                                                        |
+| `amount`        | `float`  |                                                        |
+| `currency`      | `string` |                                                        |
+| `recurrence`    | `string` |                                                        |
+| `platform`      | `string` | Always `wc` in this case                               |
+| `referer`       | `string` |                                                        |
+| `popup_id`      | `string` | If the donation was triggered by a popup, the popup ID |
+| `platform_data` | `array`  |                                                        |
+
+### `woocommerce_order_failed`
+
+For when there's a new donation payment failed through WooCommerce.
+Known issue: If the user tries to pay again after a failed payment, and the payment fails for a second time,
+the order is already marked as failed so this hook will not trigger.
+
+| Name            | Type     | Obs                                                    |
+| --------------- | -------- | ------------------------------------------------------ |
+| `user_id`       | `int`    |                                                        |
+| `email`         | `string` |                                                        |
+| `amount`        | `float`  |                                                        |
+| `currency`      | `string` |                                                        |
+| `recurrence`    | `string` |                                                        |
+| `platform`      | `string` | Always `wc` in this case                               |
+| `referer`       | `string` |                                                        |
+| `popup_id`      | `string` | If the donation was triggered by a popup, the popup ID |
+| `platform_data` | `array`  |                                                        |
+
 
 ### `donation_new`
 
 When there's a new donation, either through Stripe or Newspack (WooCommerce) platforms.
 
-| Name            | Type     |
-| --------------- | -------- |
-| `user_id`       | `int`    |
-| `email`         | `string` |
-| `amount`        | `float`  |
-| `currency`      | `string` |
-| `recurrence`    | `string` |
-| `platform`      | `string` |
-| `platform_data` | `array`  |
+| Name            | Type     | Obs                                                    |
+| --------------- | -------- | ------------------------------------------------------ |
+| `user_id`       | `int`    |                                                        |
+| `email`         | `string` |                                                        |
+| `amount`        | `float`  |                                                        |
+| `currency`      | `string` |                                                        |
+| `recurrence`    | `string` |                                                        |
+| `platform`      | `string` |                                                        |
+| `referer`       | `string` |                                                        |
+| `popup_id`      | `string` | If the donation was triggered by a popup, the popup ID |
+| `platform_data` | `array`  |                                                        |
 
 ### `donation_subscription_new`
 
 When there's a new WooCommerce Subscription. This action does not replace the `donation_new` that create the subscription.
 
-| Name            | Type     |
-| --------------- | -------- |
-| `user_id`       | `int`    |
-| `email`         | `string` |
-| `amount`        | `float`  |
-| `currency`      | `string` |
-| `recurrence`    | `string` |
-| `platform`      | `string` |
-| `platform_data` | `array`  |
+| Name              | Type     | Obs                                                    |
+| ----------------- | -------- | ------------------------------------------------------ |
+| `user_id`         | `int`    |                                                        |
+| `email`           | `string` |                                                        |
+| `subscription_id` | `int`    |                                                        |
+| `amount`          | `float`  |                                                        |
+| `currency`        | `string` |                                                        |
+| `recurrence`      | `string` |                                                        |
+| `platform`        | `string` |                                                        |
+| `referer`         | `string` |                                                        |
+| `popup_id`        | `string` | If the donation was triggered by a popup, the popup ID |
 
 ### `donation_subscription_cancelled`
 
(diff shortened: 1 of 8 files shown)


includes/data-events/class-popups.php
```diff
@@ -180,18 +180,21 @@ public static function registration_submission_with_status( $email, $user_id, $p
...
 		if ( ! $popup_id ) {
 			return;
 		}
 		$popup_data = self::get_popup_metadata( $popup_id );
 		return array_merge(
 			$popup_data,
 			[
-				'action' => self::FORM_SUBMISSION,
+				'action'                         => self::FORM_SUBMISSION,
+				'newsletter_subscription_method' => $metadata['newsletters_subscription_method'],
```
miguelpeixe, line 196: For the sake of preserving the root object of `campaign_interaction` more agnostic and with permanent keys, I'd place this type of metadata nested in an isolated key since it's regarding a block inside the campaign. Something like `campaign_data` or `blocks_data`?

Something similar is done here, for donations:

https://github.com/Automattic/newspack-plugin/blob/aa5543a7c507b673650b0bb9ce9af0fd4a3965d3/includes/data-events/listeners.php#L219-L233

The platform may vary and have specific metadata associated with it. That shouldn't mess with the root object.

miguelpeixe: Also, documenting in `README.md` also helps with perspective on how we're building these.

leogermani: > Something similar is done here, for donations:

IN donations all info about the donation is block specific as well. `amount`, `currency`, etc... It's the same case. If I change here I'd have to change there as well.

I get the point, can I just ask to do this refactor on another PR? because this will also touch the gha-4 connector and there's another PR open and they'll all conflict if I start this now.

> Also, documenting in README.md also helps with perspective on how we're building these.

I'm doing this! :) Since I'm touching this in several PRs and they are still changing a bit I'll do it all at once as one PR at the end.

GA4 connector will have its own README as well.

miguelpeixe: > IN donations all info about the donation is block specific as well. amount, currency, etc... It's the same case. If I change here I'd have to change there as well.

I don't understand the relation. Every `donation_new` must have `amount`, `currency`, etc... It's a requirement of the action event regardless of the platform.

The `campaign_interaction` have some clear requirements, like `action` and `referer`. Other arguments are specific to the contents of the campaign, therefore not always there. If the event was called `campaign_interaction_donation` or `campaign_interaction_newsletter_subscription` it would be a different case.

We're not ensuring a good schema here.

miguelpeixe: These listeners are not just to serve GA4, they are general purpose.

leogermani: ah, the other PR got merged. I can do this refactor now.

I can add `registration_data`, `donation_data` and `newsletter_subscription_data` keys in the root and place the values there. WDYT?

miguelpeixe: `interaction_data` sounds like a good candidate as well for holding these

miguelpeixe: why not an abstract one, so APIs don't have to struggle with it? if the action tells what it is, why would they key also have to?

leogermani: > interaction_data sounds like a good candidate as well for holding these

To ensure we always have the same schema. And to be consistent with `has_registration`, `has_newsletter` and `has_donation`.

I know it's unlikely, but in theory you can have more than one of these blocks inside a popup.

Also, we can know beforehand the schema inside each of these keys as well.

>  if the action tells what it is, why would the key also have to?

The action does not tell us that. At least not yet. that's being done by the `has_*` params

miguelpeixe:
This also goes for the `campgaign_has_{block-name}_block`. If we manage to take that information out of the schema and in a `has_blocks` array containing the names of the block, it's also a more scalable and generic solution for campaign prompts.

leogermani:
> This also goes for the campgaign_has_{block-name}_block. If we manage to take that information out of the schema and in a has_blocks array containing the names of the block, it's also a more scalable and generic solution for campaign prompts.

agreed

> My suggestion is we use the action value to inform what type of interaction it is, and an abstract key like interaction_data to hold every interaction metadata.

This keeps the schema simpler, on one hand. But on the other hand I have no idea what's inside `interaction_data`. If we have one for each type, we can also describe what they look like.

It's not like we are going to have dozens of interaction types there.

miguelpeixe:
> But on the other hand I have no idea what's inside interaction_data. If we have one for each type, we can also describe what they look like.

Looks to me like an easier documentation problem to tackle, not a schema one.

leogermani:
> Looks to me like an easier documentation problem to tackle, not a schema one.

Captura de tela de 2023-03-10 10-47-58
(you win)

miguelpeixe:
That must be a first! 😆

leogermani:
Just found a bug that I don't know how we didn't catch before. Will address it

leogermani:
Still working on this big gap I've found on tracking Stripe donation events in GA4. will need more time on Monday. I'll mark the PR as needs review again once it's done

leogermani:
@miguelpeixe 

This is ready for review again. I also fixed a bug (described it in the PR description).

I also added the documentation.

Since we touched the structure of all events, I think we should do an extensive tests of as many events as possible, as well as donations (one time and recurring) with both platforms.

leogermani:
@miguelpeixe :wave:

miguelpeixe:
Sorry for the delay, I plan on reviewing this today.

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -259,6 +305,31 @@ public static function handle_newsletter_subscribed( $params, $data ) {
...
+	public static function handle_campaign_interaction( $params, $data ) {
+		$transformed_data = $data;
+		// remove data added in the body filter.
+		unset( $transformed_data['ga_params'] );
+		unset( $transformed_data['ga_client_id'] );
+
+		unset( $transformed_data['interaction_data'] );
+		$transformed_data = array_merge( $transformed_data, $data['interaction_data'] );
+
+		unset( $transformed_data['campaign_blocks'] );
```
miguelpeixe, line 325: When the `reader_registered` and `newsletter_subscribed` events are sent, they are populated with these campaign metadata but not sanitized like here for `campaign_interaction`.

Should some handling happen for such cases or just let the generic sanitizer handle it? This is how it's currently handled:

```
[NEWSPACK-DATA-EVENTS-GA4][Newspack\Data_Events\Connectors\GA4::log]: Parameter campaign_blocks has an invalid value. It was removed from the np_newsletter_subscribed event.
[NEWSPACK-DATA-EVENTS-GA4][Newspack\Data_Events\Connectors\GA4::log]: Parameter interaction_data has an invalid value. It was removed from the np_newsletter_subscribed event.
```

includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -139,6 +147,44 @@ public static function filter_event_body( $body, $event_name ) {
 
 	}
 
+	/**
+	 * When the donation_new event is dispatched from the Stripe webhook, it can't catch GA client ID from Cookies.
```
miguelpeixe, line 151: This is also an issue with other logic inside `GA4::filter_event_body()`. The event triggered by Stripe webhook is being dispatched with `logged_in` set to `no` and without `is_reader`, even though I'm a logged-in reader.

includes/data-events/README.md
```diff
@@ -139,16 +177,43 @@ When a WooCommerce Subscription status changes.
...
+| `campaign_frequency` | `string` |                                                                                                                                                |
+| `campaign_placement` | `string` |                                                                                                                                                |
+| `campaign_blocks`    | `array`  | Array containing the blocks that are inside the campaign. Only 3 blocks are tracked: `donation`, `registration` and `newsletters_subscription` |
+| `action`             | `string` | `form_submission`, `form_submission_success` or `form_submission_failure`                                                                      |
+| `action_type`        | `string` | `donation`, `registration` or `newsletters_subscription`                                                                                       |
+| `interaction_data`   | `array`  | Depending on the actino type, it will return different information about the interaction.                                                      |
+
+#### structure of interaction_data:
```
miguelpeixe, line 191:
```suggestion
#### Possible values for `interaction_data`
```

miguelpeixe (changes requested):
Thank you for the documentation!

includes/data-events/connectors/ga4/class-ga4.php
leogermani: weird, I didn't see this happening. I'm removing these 2 keys from the data and they should never be sent in tha GA event...

I'll double check

`campaign_blocks` is transformed into the `has_XX_block` params

`interaction_data` is merged and moved to the same level as root params...

miguelpeixe: They are but only in the context of the `campaign_interaction` event, not the other ones that also end up getting populated with the same values.

leogermani: https://github.com/Automattic/newspack-plugin/pull/2337/commits/4fa17fa9cd685c9cd3042e36664a0c609c9e9c24

includes/data-events/connectors/ga4/class-ga4.php
leogermani: I was tired when I noticed this the other day.

Looking at this again today I was happy to see it was an easy fix after all I did.

I also figured we could default the value of `is_reader` to `no` if the user is not logged in

https://github.com/Automattic/newspack-plugin/pull/2337/commits/df620d276bb0a0417399b94fb4b997f1e6c3f576

kmwilkerson:
@yogeshbeniwal That's correct. Due to GA4's [automatically collected events](https://support.google.com/analytics/answer/9234069?hl=en), NCI (who produces the News Tagging Guide) will be updating the guide to use those auto events so that users no longer have to manually create events.

miguelpeixe (approved):

PR #2348, opened 2023-03-15
fix: prevent page auto-scroll when RAS is enabled

adekbadek:
### All Submissions:

Fixes an issue that was reported on one live site, but may appear on more. 

The RAS JS was focusing on an input on page load. If the input was hidden with CSS, it would not be focused on, but if for some reason the CSS was delayed or stripped, the input would be focused on and the page would scroll to where the input was – in this case the bottom of the page. 

1. On `master`, ensure RAS is enabled and the header sign-in link is rendered
2. Visit the page, and block the `/wp-content/plugins/newspack-plugin/dist/reader-auth.css` file from loading in the devtools network tab, reload and observe the page is scrolled to bottom on load
3. Switch to this branch, observe the issue cannot be reproduced
4. Open the sign-in modal by clicking the header link and ensure that the input is focused when switching sign-in methods ("sign in using a…" text link)

### Other information:

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -236,7 +236,7 @@ window.newspackRAS.push( function ( readerActivation ) {
 			/**
 			 * Handle auth form action selection.
 			 */
-			function setFormAction( action ) {
+			function setFormAction( action, shouldFocus = false ) {
 				if ( 'otp' === action ) {
 					if ( ! readerActivation.getOTPHash() ) {
 						return;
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Amazing catch! 🎉

miguelpeixe:
Should we also add some inline CSS to prevent the `newspack-reader-auth` element to be visible at all in cases of delayed CSS?

adekbadek:
Good idea, let's tackle this separately.

PR #2363, opened 2023-03-21
fix: bail silently when recaptcha key is not set

miguelpeixe:
### All Submissions:

When the reCAPTCHA library is loaded, RAS' library assumes it should be used and renders an error when submitting a form if the connection is not configured.

This PR tweaks the token fetch to bail silently when a site key is not set.

1. While on the master branch, make sure you have RAS configured and reCAPTCHA **not** configured
2. Install the [reCAPTCHA for WooCommerce](https://wordpress.org/plugins/recaptcha-woo/) plugin
3. Place a "Reader Registration" block
4. Submit the form and confirm you see a reCAPTCHA error
5. Check out this branch, repeat step 4, and confirm the error is gone
6. Enable and configure reCAPTCHA in the Connections wizard
7. Repeat step 4 and confirm it also works with reCAPTCHA

### Other information:

### DIFF
--- a/assets/reader-activation/index.js
+++ b/assets/reader-activation/index.js
@@ -272,7 +272,13 @@ export function getCaptchaToken( action = 'submit' ) {
 		}
 
 		const { captcha_site_key: captchaSiteKey } = newspack_reader_activation_data;
-		if ( ! grecaptcha?.ready || ! captchaSiteKey ) {
+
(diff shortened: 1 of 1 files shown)


dkoo (approved):
I can't replicate the original bug, but seems okay to me.

PR #2377, opened 2023-03-27
feat(memberships): overlay style for content gate

miguelpeixe:
### All Submissions:

Implements overlay style options for the content gate page:

image

When selected, **the overlay appears once the beginning of the post content reaches half the viewport**. The settings for the number of visible paragraphs/more block do not change:

https://user-images.githubusercontent.com/820752/228651035-9f865350-5839-4b59-b681-99dbd79bbd65.mov

---

Overlay options are:

 - **Placement**: center or bottom
 - **Size (width)**: small (400px) , medium (600px), large (800px)

On mobile, the overlay will always be placed at the bottom and full-width. 

---

For later discussions: some ads are rendered behind this overlay and we should probably have a way around this. One possibility, for a separate PR, is to allow the publisher to tweak visible ad placements in gated content, just like it's done for regular pages/posts:

image

1. Check out this branch and confirm you have the latest `newspack-blocks` (for [this](https://github.com/Automattic/newspack-blocks/pull/1411))
2. Edit the gate and change to "Overlay" and use the presented default settings
3. Place a donate block in the gate
4. Visit any restricted content, scroll down and confirm the gate appears as described above, in the center and 800px wide
5. Visit on mobile, scroll down, and confirm the gate is positioned at the bottom and full-width
6. Change the placement to "bottom", visit the page, and confirm the change on desktop
7. Tweak the size and confirm the editor changes as well,save, and confirm the change on the frontend
8. Click to donate and confirm the checkout takes over the overlay
9. Close the checkout and confirm the gate is back
10. Complete the membership flow and confirm the page is refreshed, the content is unlocked and the gate no longer appears

### Other information:


assets/memberships-gate/editor.js
```diff
@@ -2,22 +2,34 @@
...
+
+const overlayPositionsLabels = {
+	center: __( 'center', 'newspack' ),
+	bottom: __( 'bottom', 'newspack' ),
+};
+
+const overlaySizes = [
+	{ value: 'small', label: __( 'Small', 'newspack' ) },
+	{ value: 'medium', label: __( 'Medium', 'newspack' ) },
+	{ value: 'large', label: __( 'Large', 'newspack' ) },
```
dkoo, line 32: What about a `full-width` option here, too? On a large screen the "large" size still seems quite small.

Screen Shot 2023-03-31 at 12 22 54 PM

Maybe within the full-width container we can still restrict the content to the same bounds as the site content itself, so the blocks aren't flush-left with the viewport.

assets/memberships-gate/overlay.scss
```diff
@@ -0,0 +1,81 @@
...
+			right: 0;
+			bottom: 0;
+			background: #000;
+			opacity: 0;
+			transition: opacity 200ms ease;
+		}
+		&[data-visible='true'] {
+			pointer-events: auto;
+			&::before {
+				opacity: 0.75;
```
dkoo, line 25: What about making this configurable like we do with overlay prompts?

includes/plugins/class-wc-memberships.php
```diff
@@ -335,6 +390,40 @@ public static function excerpt( $excerpt, $post, $message_code ) {
...
+			return;
+		}
+		$gate_post_id = self::get_gate_post_id();
+		$style        = \get_post_meta( $gate_post_id, 'style', true );
+		if ( 'overlay' !== $style ) {
+			return;
+		}
+		global $post;
+		$_post = $post;
+		$post  = \get_post( $gate_post_id ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
```
dkoo, line 411: Overwriting the global seems risky—any particular reason why we need this?

dkoo (changes requested):
This is working great! I ran into one JS error in the editor, and have a few questions below. I also wanted to mention that on a larger screen it's very easy to read the content under the overlay. Wonder if there's anything we can do to mitigate that beyond adding a "full-width" option for the gate and a configurable overlay opacity (both suggested below)?

Could we somehow set the CSS height of `html` so that the page won't scroll past a certain point beyond where the gate is triggered? Or truncate the content under the overlay like we do with the inline gate?

includes/plugins/class-wc-memberships.php
miguelpeixe: Some `the_content` filter is forcefully replacing the gate `post_content` with the global post's content. This was the only way I could make the gate content be rendered properly. The global variable is immediately restored to its original value after the gate is rendered, so I don't think we are doing anything bad.

You can test it locally by commenting-out lines 409-414 and 423-424 and checking if the `post_content` is working.

If your local instance manages to render the correct content, it's probably something here that I should figure out.

assets/memberships-gate/overlay.scss
miguelpeixe: I'd rather use a standard that follows the same applied to the modal checkout for a seamless transition when rendering the checkout.

miguelpeixe:
> Or truncate the content under the overlay like we do with the inline gate?

The content should be truncated with the same settings as applied to the inline gate. Did it not on your instance?

assets/memberships-gate/editor.js
miguelpeixe: 01e1ff73f46190dd076b371c072e1f74fb6660aa tweaks the sizes so it reflects the available set in Newspack Campaigns and full width.

Since large is rather large, I set the default to medium now.

assets/memberships-gate/editor.js
```diff
@@ -63,6 +83,27 @@ const GateEdit = ( { editPost, createNotice, meta } ) => {
...
+						<SelectControl
+							label={ __( 'Size', 'newspack' ) }
+							value={ meta.overlay_size }
+							options={ overlaySizes }
+							onChange={ value => editPost( { meta: { overlay_size: value } } ) }
+						/>
+						<PositionControl
+							label={ __( 'Position', 'newspack' ) }
+							value={ meta.overlay_position }
+							allowedPositions={ [ 'bottom', 'center' ] }
```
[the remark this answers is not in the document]
miguelpeixe: Perhaps because I was spreading it from `props` down to the `ButtonGroup` by accident? Couldn't reproduce locally but I'm no longer spreading it down. Please let me know if the error is gone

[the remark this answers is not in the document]
dkoo: Yes, no longer seeing it!

assets/memberships-gate/editor.js
```diff
@@ -2,22 +2,36 @@
...
+	center: __( 'center', 'newspack' ),
+	bottom: __( 'bottom', 'newspack' ),
+};
+
+const overlaySizes = [
+	{ value: 'x-small', label: __( 'Extra Small', 'newspack' ) },
+	{ value: 'small', label: __( 'Small', 'newspack' ) },
+	{ value: 'medium', label: __( 'Medium', 'newspack' ) },
+	{ value: 'large', label: __( 'Large', 'newspack' ) },
+	{ value: 'full', label: __( 'Full Width', 'newspack' ) },
```
dkoo, line 34:
```suggestion
	{ value: 'full-width', label: __( 'Full Width', 'newspack' ) },
```

If you set this value to `full-width`, you get this nice change in the style of the `PositionControl`:

Screen Shot 2023-03-31 at 2 27 56 PM

assets/memberships-gate/editor.js
```diff
@@ -63,6 +87,27 @@ const GateEdit = ( { editPost, createNotice, meta } ) => {
...
+					<Fragment>
+						<SelectControl
+							label={ __( 'Size', 'newspack' ) }
+							value={ meta.overlay_size }
+							options={ overlaySizes }
+							onChange={ value => editPost( { meta: { overlay_size: value } } ) }
+						/>
+						<PositionControl
+							label={ __( 'Position', 'newspack' ) }
+							value={ meta.overlay_position }
```
dkoo, line 100:
```suggestion
							value={ meta.overlay_position }
							size={ meta.overlay_size }
```

To enable the alternative `PositionControl` styles for full-width

assets/memberships-gate/editor.scss
```diff
@@ -1,9 +1,33 @@
...
+		max-width: gate.$size__x-small;
+	}
+	.editor-styles-wrapper[data-overlay-size='small'] & {
+		max-width: gate.$size__small;
+	}
+	.editor-styles-wrapper[data-overlay-size='medium'] & {
+		max-width: gate.$size__medium;
+	}
+	.editor-styles-wrapper[data-overlay-size='large'] &,
+	.editor-styles-wrapper[data-overlay-size='full'] & {
```
dkoo, line 26:
```suggestion
	.editor-styles-wrapper[data-overlay-size='full-width'] & {
```

dkoo (changes requested):
> The content should be truncated with the same settings as applied to the inline gate. Did it not on your instance?

Oh, you're right, it is! I just didn't notice because my article content is all lorem ipsum. 😆 Nit: Could we append an ellipsis `[...]` to the end of the truncated content to indicate that it's truncated?

I also left some small suggestions which would apply a feature from the Campaigns plugin for `PositionControl` selector when the overlay is full-width. After that, this looks good to go!

miguelpeixe:
Nice! Thank you for the suggestion!

Did you manage to confirm the `the_content` filter issue on global post?

miguelpeixe:
61382e04b981f30a1409268dada8e332705953e0 adds ellipsis when in overlay and fixes a minor bug. The last paragraph of the truncated content was missing a closing `</p>`

includes/plugins/class-wc-memberships.php
dkoo: I'm seeing the same thing, and I think it's the `apply_filters( 'the_content' )` that's messing with the post returning the content, but it's also necessary to transform the block markup correctly, so let's leave it as-is.

dkoo (approved):
Thanks for the revisions!

PR #2381, opened 2023-03-29
feat: start new RAS settings page

dkoo:
### All Submissions:

Scaffolds a new page layout for the RAS settings page according to the latest i2 designs. The list of prerequisites and all copy/links are TBD, so this PR contains a list of a few known requirements and a boilerplate config object to serve as a template to add more.

Because discussions are ongoing about what to do about the existing RAS settings on this page, these are simply hidden behind an "advanced settings" flag to show/hide those settings. This advanced settings section will need to be redesigned in a future PR.

The `ActionCard` component to kick off the Campaigns default prompts/segments setup is a placeholder, and is non-interactive for now.

1. Visit the **Newspack > Engagement > Reader Activation** admin page and ensure that the prerequisites are shown according to the config array in `includes/reader-activation/class-reader-activation.php` and that all the links and buttons work. Feel free to copy the existing config objects and create your own to test different scenarios. This is the state of the page before all RAS prerequisites are completed:

Screen Shot 2023-03-29 at 3 10 49 PM

2. Open the Advanced Settings section and ensure that all of these settings options still work as in `master`, for now. Note that some fields have been removed based on discussions that they should always be set one way or another if RAS is enabled. This section will need to be redesigned in a future PR.

### Other information:

### DIFF
--- a/assets/wizards/engagement/views/reader-activation/index.js
+++ b/assets/wizards/engagement/views/reader-activation/index.js
@@ -112,6 +112,7 @@ export default withWizardScreen( () => {
 							'Newspack’s Reader Activation system is a set of features that aim to increase reader loyalty, promote engagement, and drive revenue. ',
 							'newspack'
 						) }
+						{ /** TODO: Update this URL with the real one once the docs are ready. */ }
 						<ExternalLink href={ 'https://help.newspack.com' }>
 							{ __( 'Learn more', 'newspack-plugin' ) }
 						</ExternalLink>
--- a/assets/wizards/engagement/views/reader-activation/index.js
+++ b/assets/wizards/engagement/views/reader-activation/index.js
@@ -1,3 +1,4 @@
+/* global newspack_engagement_wizard */
 /**
  * WordPress dependencies
  */
@@ -25,6 +26,7 @@ import ActiveCampaign from './active-campaign';
 export default withWizardScreen( () => {
 	const [ inFlight, setInFlight ] = useState( false );
 	const [ config, setConfig ] = useState( {} );
+	const [ membershipsConfig, setMembershipsConfig ] = useState( {} );
 	const [ error, setError ] = useState( false );
 	const [ allReady, setAllReady ] = useState( false );
 	const [ isActiveCampaign, setIsActiveCampaign ] = useState( false );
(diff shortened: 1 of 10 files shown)


assets/wizards/engagement/views/reader-activation/index.js
```diff
@@ -87,159 +102,180 @@ export default withWizardScreen( () => {
...
 		<>
+			<SectionHeader
+				title={ __( 'Reader Activation', 'newspack' ) }
+				description={ () => (
+					<>
+						{ __(
+							'Newspack’s Reader Activation system is a set of features that aim to increase reader loyalty, promote engagement, and drive revenue. ',
+							'newspack'
+						) }
+						<ExternalLink href={ 'https://help.newspack.com' }>
```
laurelfulford, line 115: We may want to sneak a `TODO` in here so we don't forget to change the link to the actual Reader Activation docs 🙂

laurelfulford (approved):
This is looking nice, @dkoo! 

The advanced settings all look good, as do the new panels. I also successfully added a new one by editing the array. 

The panel status doesn't seem correct -- I tried unsetting the ESP settings and it still had the checkmark -- but I think this may be expected since some of them, like Terms & Conditions, aren't "hooked up" yet. 

I added a code comment for a spot where we could add a `TODO`, but that's only because it's the kind of thing I would forget, it might not be necessary.
 
The only thing I noticed that seemed a bit odd was the alignment on the last panel, but that's probably best fixed in the final copy (making the text on the left shorter, or a shorter "waiting for settings" message would help!):

image

Marking as approved as I don't think anything I noted is a blocker based on the testing steps, but let me know if you have any questions at all!

assets/wizards/engagement/views/reader-activation/index.js
dkoo: Good call—added in 50c2b70!

dkoo:
Good catch—I was using the wrong method to check whether the ESP is setup (the method I was using just checks if the plugin itself is active). 7f82975 should fix this!

dkoo:
> The only thing I noticed that seemed a bit odd was the alignment on the last panel, but that's probably best fixed in the final copy (making the text on the left shorter, or a shorter "waiting for settings" message would help!):

I think the latest designs address this by hiding the full description until you "expand" the expandable cards, but we'll tackle that in the next PR.

dkoo:
@laurelfulford thanks for the review—could use a quick re-review!

assets/wizards/engagement/views/reader-activation/index.js
laurelfulford: Thanks Derrick!

laurelfulford (approved):
Thanks Derrick! 

I re-checked the ESP settings, and then double-checked reCAPTCHA and Reader Revenue, and all work as expected (disabling reCAPTCHA and uninstalling Woo, respectively made those two panels say that they needed to be set up! :shipit:

PR #2385, opened 2023-03-31
Improved handling of trial subscriptions in ESP sync

claudiulodro:
### All Submissions:

Closes [this issue](https://app.asana.com/0/1200133283755575/1204276644179086)

After running [the command](https://github.com/Automattic/newspack-plugin/blob/fix/trial-sync/includes/reader-revenue/stripe/class-stripe-sync.php#L712) to cancel Stripe subscriptions and re-create them as Stripe Donate Block subscriptions, the re-created subscriptions are in 'trialing' status until their next renewal date in Stripe (so we don't charge the customers until they would normally have been charged). 

Then, when we run [the ESP backfill command](https://github.com/Automattic/newspack-plugin/blob/fix/trial-sync/includes/reader-revenue/stripe/class-stripe-sync.php#L684), these customers incorrectly show up in the ESP with the 'Ex-Monthly Donor' or 'Ex-Yearly Donor' status because the backfill checks the subscription associated with the last payment, but that subscription was cancelled by the first command. 

This PR adjusts the logic a little to prevent this situation.

1. Set up Reader Revenue with the legacy Stripe platform and connect it to ActiveCampaign.
2. Before applying this PR, make a monthly donation. Run `wp newspack stripe sync-stripe-connect-to-stripe`. In Stripe you should now see that the previous Stripe subscription was cancelled and there is a new active trialing subscription in its place:

Screenshot 2023-03-31 at 1 42 52 PM
Screenshot 2023-03-31 at 1 42 45 PM

4. Run `wp newspack stripe sync-customers --sync-to-wc=0 --sync-to-esp=1`. In ActiveCampaign you should see the donor and the status should say 'Ex-Monthly Donor' (though the customer has an active subscription in trial):

Screenshot 2023-03-31 at 1 20 58 PM

5. Apply this patch. Run `wp newspack stripe sync-customers --sync-to-wc=0 --sync-to-esp=1`. In ActiveCampaign you should now see the correct status as 'Monthly Donor':

Screenshot 2023-03-31 at 1 33 14 PM

### Other information:

### DIFF
--- a/includes/reader-revenue/stripe/class-stripe-sync.php
+++ b/includes/reader-revenue/stripe/class-stripe-sync.php
@@ -229,7 +229,7 @@ class Stripe_Sync {
 					$subscription['start_date']
 				);
 				$metadata                   = array_merge( $recurring_related_metadata, $metadata );
-				if ( ! in_array( $subscription['status'], [ 'active', 'trialing' ] ) ) {
+				if ( ! in_array( $subscription['status'], [ 'active', 'trialing' ], true ) ) {
 					$metadata[ Newspack_Newsletters::get_metadata_key( 'membership_status' ) ] = 'Ex-' . $membership_status;
 				}
 				if ( $subscription['ended_at'] ) {
--- a/includes/reader-revenue/stripe/class-stripe-sync.php
+++ b/includes/reader-revenue/stripe/class-stripe-sync.php
@@ -212,7 +212,15 @@ class Stripe_Sync {
 				}
 			}
 
-			$subscription = Stripe_Connection::get_subscription_from_payment( $payment );
+			// Check if there are any active subscriptions.
+			$subscriptions = Stripe_Connection::get_subscriptions_of_customer( $customer->id );
+			if ( ! empty( $subscriptions['data'] ) ) {
+				$subscription = reset( $subscriptions['data'] );
+			} else {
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/stripe/class-stripe-sync.php
```diff
@@ -221,7 +229,7 @@ private static function sync_to_esp( $customer, $args ) {
 					$subscription['start_date']
 				);
 				$metadata                   = array_merge( $recurring_related_metadata, $metadata );
-				if ( 'active' !== $subscription['status'] ) {
+				if ( ! in_array( $subscription['status'], [ 'active', 'trialing' ] ) ) {
```
leogermani, line 232: not sure why PHPCS didn't get this, but I we enforce the usage of strict comparison in `in_array`. So you need to add a third `true` argument there

claudiulodro: Thanks for flagging. Weird that neither CI nor local PHPCS caught it! I've fixed in https://github.com/Automattic/newspack-plugin/pull/2385/commits/07e465107c04a95ecdc8731fabec9e7103f0c3ee

dkoo:
Just triggering a re-review since it seems like it's ready

dkoo (approved):

PR #2400, opened 2023-04-07
fix: don't show gate unless attached to a specific post

dkoo:
### All Submissions:

An alpha hotfix. Fixes some weird content gate behavior in the current alpha version, so needs to be merged before the alphas are deployed.

When a WC Membership plan is created and associated with a taxonomy term, a notice is displayed when a query for that term shows its posts. This causes the content gate to be displayed in, for example, Homepage Posts blocks that query that term.

This PR simply hides the gate entirely, but an alternative solution would be to retain the WC Memberships plugin's default behavior and show the default message instead of replacing with our own gate.

1. Create a membership plan related to a specific category or tag.
2. On a post, add a Homepage Posts block targeting that term.
3. On the front-end, on `alpha` or `master`, observe that the content gate is rendered at the top of the Homepage Posts block.
4. Check out this branch and repeat step 3. Confirm that the gate is no longer shown.
5. Confirm that the gate is shown on singular posts associated with the term.

### Other information:

### DIFF
--- a/includes/plugins/class-wc-memberships.php
+++ b/includes/plugins/class-wc-memberships.php
@@ -328,13 +328,13 @@ class WC_Memberships {
 	 * @param array  $message_args associative array of message arguments.
 	 */
 	public static function notice_html( $notice, $message_body, $message_code, $message_args ) {
-		// Don't show gate unless attached to a specific post.
-		if ( empty( $message_args['post'] ) ) {
-			return '';
-		}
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Nice catch!

I moved the check to after the `has_gate()` check. This way we are not altering any of the original behavior unless we are taking over the gate.

dkoo:
Good change, thanks for reviewing!

PR #2407, opened 2023-04-12
fix(newsletters): only attempt token verification for Constant Contact

miguelpeixe:
### All Submissions:

The token verification is meant for validating short and long-lived tokens from OAuth-based ESPs. Currently, this is only the case for Constant Contact and we are sending unnecessary requests that fail whenever the settings are saved with a different provider.

This is just to ensure that the request is only sent when Constant Contact is the provider.

To improve the UX, 6abd1302c194adcd7978041b22b032bf7eef1498 adds an in-flight state that disables form fields while saving. Although this doesn't mean a success message, it indicates that something happened.

1. While on the master branch, visit the Engagement -> Newsletters wizard
2. Save the settings with a valid Mailchimp API with your DevTools Network tab open
3. Confirm you see the successful `newspack/v1/wizard/newspack-engagement-wizard/newsletters` request
4. Confirm you see right after this request, a failed request to `newspack-newsletters/v1/mailchimp/verify_token`, which is not a valid endpoint
5. Check out this branch, repeat step 2, and confirm the second request is not fired
6. Do the same with Constant Contact and confirm the request is sent

### Other information:

### DIFF
--- a/assets/wizards/engagement/views/newsletters/index.js
+++ b/assets/wizards/engagement/views/newsletters/index.js
@@ -285,6 +285,10 @@ const Newsletters = () => {
 		if ( ! provider ) {
 			return;
 		}
+		// Constant Contact is the only provider using an OAuth strategy.
+		if ( 'constant_contact' !== provider ) {
+			return;
+		}
 		apiFetch( { path: `/newspack-newsletters/v1/${ provider }/verify_token` } )
 			.then( response => {
 				if ( ! response.valid && response.auth_url ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Since we're on this component, I also pushed 1b09313 which fixes a pet peeve on mine (no visual indication that we're fetching subscription lists, and no disabled state for the "Save" buttons while the wizard is in flight). If it looks good to you, the fix for the `verify_token` requests looks good to me.

miguelpeixe:
Thank you for that tweak! Looks good to me

PR #2410, opened 2023-04-14
fix(perfmatters): exclude Jetpack and Newsletters css delay

miguelpeixe:
### All Submissions:

Remove CSS delay from Jetpack's sharing buttons and Newspack Newsletters (Subscription Form block)

1. Make sure you have Perfmatters installed and active
2. Configure JP's sharing buttons on your instance and confirm the CSS is rendered immediately
3. Add a Newsletters Subscription Form block to page, visit and confirm the `subscribeBlock.css` is loaded immediately

### Other information:

### DIFF
--- a/includes/plugins/class-perfmatters.php
+++ b/includes/plugins/class-perfmatters.php
@@ -80,6 +80,8 @@ class Perfmatters {
 	private static function unused_css_excluded_stylesheets() {
 		return [
 			'plugins/newspack-blocks', // Newspack Blocks.
+			'plugins/newspack-newsletters', // Newspack Newsletters.
+			'plugins/jetpack/modules/sharedaddy', // Jetpack's share buttons.
 			'/themes/newspack-', // Any Newspack theme stylesheet.
 			'wp-includes',
 		];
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described. There is still a slight delay in the loading of the Jetpack share icons, but that seems to be something separate—perhaps a custom font?

miguelpeixe:
Which aspect of the button is delayed? The icon rendering? I'm getting all synchronously using the "Icon only" style.

miguelpeixe:
Just noticed that JP's "social logos" CSS is also being delayed. It's not used at all in my local instance for rendering the sharing buttons, but should probably not be delayed.

Excluded as well in 028df2a72bdbba432b7c6cee39eb316d978f4cf9

dkoo:
> Just noticed that JP's "social logos" CSS is also being delayed. It's not used at all in my local instance for rendering the sharing buttons, but should probably not be delayed.

The various options for displaying the icons may enqueue different assets. I was using the "icon only" option:

Screen Shot 2023-04-18 at 11 42 18 AM

dkoo:
The share buttons now load with no delay when using the "Icon & Text", "Icon Only", and "Text Only" options. There's still a slight delay when using the "Official Buttons" option, but I'm not sure we can fix that as it seems to pull styles from the services themselves.

PR #2423, opened 2023-04-27
feat(content-gate): metering

miguelpeixe:
### All Submissions:

Implements metering functionality to the content gate.

image

If anonymous metering is set, it will occur on the front end using localStorage in order to work with caching. Once signed in, metering is processed in the backend if the content is restricted for that reader.

Along with implementing metering functionality, this also proposes a few changes to the existing code:
 - Switch the gate editor javascript to use `useSelect` and `useDispatch` hooks instead of `withSelect` and `withDispatch` HOCs for a cleaner looking code
 - Change the integration class and namespace from `WC_Memberships` to `Memberships` so its methods can be used without being confused with Woo's original class
 - Use a configuration array for the editor meta to avoid code repetition
 - Added a static method that retrieves the inline gate content so the metering class can use it

1. Make sure you have RAS enabled and WC Memberships active with configured membership plans
2. Visit the Engagement wizard and click to configure the Content Gate
3. Confirm you see a new "Metering" panel in the editor
4. Toggle Metering on and confirm the parameters set are: 0 anonymous, 0 registered, weekly reset
5. Change the "Available views for registered readers" to 2 and save
6. Visit a restricted post anonymously and confirm the gate is rendered
7. Sign In to an account that is not under the restricted post plan
8. Visit the post and confirm the full content is rendered
9. Visit another post and confirm it's also not restricted
10. Visit a 3rd restricted post and confirm the gate renders
11. Change the gate style to overlay (or inline if it was already set as overlay)
12. Visit the restricted post and confirm the gate renders as expected
13. Edit the gate and change the "Available views for anonymous readers" to 1
14. Visit a restricted post anonymously and confirm it renders
15. Open DevTools and navigate to the localStorage
16. Confirm you see an item named `newspack_metering` with this schema: `{ content: [ 123 ], expiration: 123457890 }`
17. Visit another post and confirm the gate is rendered

### Other information:

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -6,6 +6,7 @@
 *.config.js
 .git
 .gitlab-ci.yml
+.hooks
 .travis.yml
 .DS_Store
 Thumbs.db
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# [1.113.0](https://github.com/Automattic/newspack-plugin/compare/v1.112.1...v1.113.0) (2023-04-24)
+
+
+### Features
+
+* **reader-registration:** add recaptcha panel in the editor ([#2397](https://github.com/Automattic/newspack-plugin/issues/2397)) ([336ee94](https://github.com/Automattic/newspack-plugin/commit/336ee942f31cba9613f2c3065c7475165ee5fa9d))
+
+# [1.113.0-alpha.1](https://github.com/Automattic/newspack-plugin/compare/v1.112.1...v1.113.0-alpha.1) (2023-04-13)
+
+
+### Features
+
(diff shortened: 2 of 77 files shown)


leogermani (approved):
Works beautifully!

One thing I noticed, out of the scope of this PR, but that might be a problem for publishers setting up a Reg-wall.

If a set up a reg-wall, all my existing registered users will not be associated to the plan. So if they log in they will still be locked out.

I guess publishers will have to manually add all existing users to this membership plan once they create it. Is that what you're thinking? Maybe it's worth adding that to the docs (if not there already?)

miguelpeixe:
Thanks, Leo!

> If a set up a reg-wall, all my existing registered users will not be associated to the plan.

That's actually covered by the memberships plugin! Thanks for the reminder because that needs to be part of our docs 🙌 

Captura de Tela 2023-05-17 às 17 47 35

PR #2425, opened 2023-04-27
feat: update NRH settings page and redirect behavior

dkoo:
### All Submissions:

Updates the News Revenue Hub default redirect URL to the latest schema. Also revamps the NRH settings page slightly and adds a new field for custom domain, which NRH will soon start supporting.

Screen Shot 2023-04-27 at 4 09 28 PM

Note that the redirect will 404 if the org ID and/or custom domains are not also set up on the NRH side; this is expected.

Closes `1200133283801628/1204490226165899`.

1. Visit **Newspack > Reader Revenue** and set your platform to "News Revenue Hub".
2. Visit the News Revenue Hub tab and confirm you see the new layout (with any preexisting values retained) and the new "Custom Domain" field. Ensure that the Organization ID and Salesforce Campaign ID fields are filled in, but leave the Custom Domain field blank.
3. Visit the site front-end and make donations via the Donate block, testing different combinations of frequency and amounts.
4. Confirm that the site redirects you to the default URL, which looks like this:

```
https://<ORGANIZATION ID>.fundjournalism.org/?amount=<DONATION AMOUNT>&installmentPeriod=<DONATION FREQUENCY>&org_id=<DONATION ID>&campaign=<SALESFORCE CAMPAIGN ID>
```

5. In settings, update the Custom Domain field with any domain value (minus protocol).
6. Make more donations on the front-end and confirm that the site redirects to a URL with your custom base domain, but the same URL parameters attached.

### Other information:

### DIFF
--- a/includes/class-nrh.php
+++ b/includes/class-nrh.php
@@ -93,7 +93,7 @@ class NRH {
 
 		$domain_parts = explode( '//', $domain );
 
-		return end( $domain_parts );
+		return \wp_unslash( end( $domain_parts ) );
 	}
 
 	/**
--- a/includes/class-nrh.php
+++ b/includes/class-nrh.php
@@ -10,16 +10,90 @@ namespace Newspack;
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Handles Salesforce functionality.
+ * Handles News Revenue Hub functionality.
  */
 class NRH {
+	/**
+	 * Allowed config keys for NRH settings.
+	 *
+	 * @var array
+	 */
+	protected static $allowed_keys = [
+		'nrh_organization_id',
+		'nrh_custom_domain',
+		'nrh_salesforce_campaign_id',
+	];
+
 	/**
 	 * Add hooks.
 	 */
 	public static function init() {
-		add_filter( 'newspack_donation_checkout_url', [ __CLASS__, 'redirect_to_nrh_checkout' ], 10, 3 );
-		add_filter( 'newspack_blocks_donate_block_html', [ __CLASS__, 'handle_custom_campaign_id' ], 10, 2 );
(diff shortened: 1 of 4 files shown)


leogermani (approved):
Works well.

Nit pick: You can set an invalid custom domain, with spaces for example, in which case you'll break the redirect

includes/class-nrh.php
```diff
@@ -10,16 +10,90 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Handles Salesforce functionality.
+ * Handles News Revenue Hub functionality.
```
claudiulodro, line 13: lol that must have been there for years!

claudiulodro (changes requested):
This looks solid. Thanks for the speedy implementation!

The one thing I do notice in testing (and you totally did everything 100% to spec) is that the `installmentPeriod` query param doesn't do anything in my testing (e.g. you can use orgid `sahanjournal` and salesforce ID `701EX00000093YPYAY` to see how it redirects to a real checkout). It's always the "Once" frequency regardless of what you select in the Donate block. 

Let me follow up around whether that param has changed in RevEngine!

claudiulodro:
@dkoo it's now `frequency` instead of `installmentPeriod`. I can push up that change assuming we don't need to find a new PR reviewer since I collaborated on this :)

dkoo:
@claudiulodro if you have the change working locally already, go ahead and push it up!

claudiulodro:
Updated in https://github.com/Automattic/newspack-plugin/pull/2425/commits/7e30540b37b314955f0a1edd8a058b623654b3d0 LGTM now!

claudiulodro (approved):

PR #2434, opened 2023-05-02
fix: ensure migrated Stripe subs have next_payment scheduled

dkoo:
### All Submissions:

Plugs a hole in the migration script for migrating Stripe subscriptions to WooCommerce. When programmatically creating subscriptions via code, Woo does not automatically schedule next payment dates—this must be done explicitly. This PR ensures this is done, and also implements a script to retroactively look up any previously migrated subscriptions and schedule next payment dates for any that are missing them. 

~Note that the scheduled next payment date won't always be in line with the start date of the subscription (e.g. if started April 3, the next payment would normally be May 3), but it should fall within 2 weeks of that date. This is because the calculate_date method is based on the current timetamp rather than the subscription's start date.~ We were able to get around this in d4978df by explicitly setting the scheduled date to the next payment date from Stripe.

Also adds a `--force` flag to the `sync-stripe-subscriptions-to-wc` script which will allow us to migrate Stripe subscriptions that were created via a migration from Pico.

1. While on `master` or `release`, set the Reader Revenue platform for your test site to "Stripe".
2. Make a monthly or annual donation via the SDB and confirm that the subscription gets created in your Stripe account.
3. Run `wp newspack stripe sync-stripe-subscriptions-to-wc` via CLI to migrate the subscription to WooCommerce.
4. In the **WooCommerce > Subscriptions** dashboard, observe that the subscription was migrated, but that it lacks a scheduled "next payment" date.
5. Check out this branch. Run `wp newspack stripe set-next-payment-dates-for-migrated-subscriptions`.
6. Confirm that the script outputs a message stating that your previous subscription was updated with a next payment date, e.g.

> No next payment date set. Setting to YYYY-MM-DD.

7. Make another new monthly or annual donation (still while on Stripe).
8. Run `wp newspack stripe sync-stripe-subscriptions-to-wc` again and this time confirm that the migrated subscription is created with a scheduled next payment date.

### Other information:

### DIFF
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -701,6 +701,14 @@ class WooCommerce_Connection {
 						update_post_meta( $subscription_id, self::SUBSCRIPTION_STRIPE_ID_META_KEY, $stripe_subscription_id );
 					}
 
+					// Ensure the next payment is scheduled.
+					$next_payment_date = $subscription->calculate_date( 'next_payment' );
+					$subscription->update_dates(
+						[
+							'next_payment' => $next_payment_date,
+						]
+					);
+
 					$subscription->save();
 
 					Logger::log( 'Created WC subscription with id: ' . $subscription_id );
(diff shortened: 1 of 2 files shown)


leogermani:
> In the WooCommerce > Subscriptions dashboard, observe that the subscription was migrated, but that it lacks a scheduled "next payment" date.

I could not confirm this. Running on `master` the subscription has a Next payment date after the migration

dkoo:
> I could not confirm this. Running on `master` the subscription has a Next payment date after the migration

Thanks for testing—that's interesting. In all cases when we ran the script without this change, every migrated subscription came in without a next payment date.

Can you confirm that migrated subs still come in with a next payment date with this change? If so, I think we can safely merge.

leogermani:
Yes, everything else worked fine

leogermani (approved):

PR #2441, opened 2023-05-05
fix: remove unneeded blocks from patterns

laurelfulford:
### All Submissions:

This PR matches https://github.com/Automattic/newspack-blocks/pull/1377 and removes unnecessary group and columns blocks from the Homepage patterns in the site setup wizard. 

* The group blocks that have been removed don't have background colours.
* The columns that have been removed are cases where they could be replaced with the Homepage Posts block set in 'grid' layout mode.

See 1200550061930446-as-1204083986338479

1. Apply the PR.
2. Run through the set-up wizard 12 times (sorry!); and pick a different homepage starter pattern each time.
3. Confirm there are no errors in the editor.
4. Review the code changes and confirm nothing unexpected is getting removed.

### Other information:


dkoo (approved):
This is looking good on a new site, and the patterns all work without errors. I love the simplification of the markup, too.

The only thing I'd note is that this will remove the pattern wrapper class names like `.newspack-pattern homepage-posts__style-24`, which might break styles for sites targeting those class names in custom CSS. I'm not sure if any sites are doing this, but we should be ready to roll this change back and reintroduce those class names if needed.

laurelfulford:
Thanks @dkoo!

> The only thing I'd note is that this will remove the pattern wrapper class names like .newspack-pattern homepage-posts__style-24, which might break styles for sites targeting those class names in custom CSS. I'm not sure if any sites are doing this, but we should be ready to roll this change back and reintroduce those class names if needed.

The upside is that this won't affect patterns already in use, it will only be on sites that are newly created since these are coming from the wizard! 🙂 We did do something similar [with the block patterns you can add through the editor](https://github.com/Automattic/newspack-blocks/pull/1377) and there may have been some risk there (if a publisher had styled a specific class and continued to add that pattern to content and it stopped getting the class on new additions), but AFAIK no issues cropped up (thankfully!).

PR #2446, opened 2023-05-08
feat: filter existing webhooks endpoints

leogermani:
### All Submissions:

Allows other plugins to programatically register endpoints.

1. Add this snippet to your site:

```PHP
add_action( 'init', function() {
    \Newspack\Data_Events\Webhooks::register_system_endpoint( 'my-id', 'https://my-url.com', [], true );
} );
```

3. Perform some actions that will dispatch events (register to a newsletters for example)
4. Confirm you see the new endopint on the Connection webhooks -> UI
5. Confirm you can't disable, edit or delete this endpoint
6. Confirm the requests are created and sent
7. Bonus point for testing the filter on the event body

### Other information:

### DIFF
--- a/includes/data-events/class-webhooks.php
+++ b/includes/data-events/class-webhooks.php
@@ -217,7 +217,7 @@ final class Webhooks {
 		}
 		\update_term_meta( $endpoint['term_id'], 'actions', $actions );
 		\update_term_meta( $endpoint['term_id'], 'global', $global );
-		return self::get_endpoint( $endpoint['term_id'] );
+		return self::get_endpoint_by_term( $endpoint['term_id'] );
 	}
 
 	/**
@@ -247,7 +247,7 @@ final class Webhooks {
 		\update_term_meta( $endpoint['term_id'], 'actions', $actions );
 		\update_term_meta( $endpoint['term_id'], 'global', $global );
 		\update_term_meta( $endpoint['term_id'], 'disabled', $disabled );
-		return self::get_endpoint( $endpoint['term_id'] );
+		return self::get_endpoint_by_term( $endpoint['term_id'] );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/data-events/class-webhooks.php
```diff
@@ -217,7 +224,33 @@ public static function create_endpoint( $url, $actions = [], $global = false ) {
...
+	 *
+	 * @param string $id      An unique identifier for the endpoint.
+	 * @param string $url     Endpoint URL.
+	 * @param array  $actions Array of action names.
+	 * @param bool   $global  Whether the endpoint should be triggered for all actions.
+	 *
+	 * @return void
+	 */
+	public static function register_system_endpoint( $id, $url, $actions = [], $global = false ) {
+
```
miguelpeixe, line 241: We should add some validation here and return an error if the ID is already being used by an existing endpoint (system or term).

miguelpeixe (changes requested):
Tests well! Only one comment below:

includes/data-events/class-webhooks.php
leogermani: The function will override the previous endpoint if you register a new one with the same ID, but this will only look at system endpoints...

I didn't want to have to query for all endpoints to do this check because system endpoints will be registered on ìnit` in every request (maybe we could be smarter about this and only but that's another topic)

What do you think of making sure that the ID is a string - and not a number that could conflict with endpoints stored in the DB?

As for the possibility of one plugin overriding the endpoint of another, that's the nature of plugins

miguelpeixe: System endpoints could have their ID generated inside the registration method with a hardcoded prefix so they're at least conflict-free with term endpoints. Something like:

```php
public static function register_system_endpoint( $slug, $url, $actions = [], $global = false ) {
  $id = 'system-' . sanitize_title( $slug );
  ...
}
```

leogermani: > hardcoded prefix so they're at least conflict-free with term endpoints.

Don't like that because when you register an endpoint you'll want to use the ID to identify it later...

I added some validation in https://github.com/Automattic/newspack-plugin/pull/2446/commits/a6fe32b44c58001b20d3804adf346af560e2cd7f, see what you think

miguelpeixe: Good call! I liked your approach better

miguelpeixe (approved):

PR #2447, opened 2023-05-08
feat: hook reCAPTCHA v3 to WooCommerce checkout

miguelpeixe:
### All Submissions:

Implements reCAPTCHA v3 to WooCommerce checkout.

1. Checkout this branch and change the [reCAPTCHA threshold](https://github.com/Automattic/newspack-plugin/blob/2b0af0b41addac3e47b67c3833af2daa2c9410f6/includes/class-recaptcha.php#L17) to `1.0`
2. Visit the Newspack -> Connections page and make sure you have reCAPTCHA v3 enabled and configured
3. Deactivate any other reCAPTCHA related plugin you may have
4. With the Reader Revenue set to "Newspack" attempt to donate
5. Confirm you get a captcha error
6. Change the threshold back to `0.5`
7. Donate again and confirm it works as expected

### Other information:

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -317,13 +317,13 @@ final class Recaptcha {
 		if ( ! self::can_use_captcha() ) {
 			return;
 		}
-		$settings = self::get_settings();
+		$site_key = self::get_setting( 'site_key' );
 		?>
 		<script src="<?php echo \esc_url( self::get_script_url() ); ?>"></script>
 		<script>
 			grecaptcha.ready( function() {
 				grecaptcha.execute(
-					'<?php echo \esc_attr( $settings['site_key'] ); ?>',
+					'<?php echo \esc_attr( $site_key ); ?>',
 					{ action: 'checkout' }
 				).then( function( token ) {
 					var field   = document.createElement('input');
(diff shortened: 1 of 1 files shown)


includes/class-recaptcha.php
```diff
@@ -286,6 +298,58 @@ public static function verify_captcha( $captcha_token ) {
...
+	}
+
+	/**
+	 * Add reCAPTCHA v3 to checkout.
+	 */
+	public static function add_recaptcha_to_checkout() {
+		if ( ! self::can_use_captcha() ) {
+			return;
+		}
+		$settings = self::get_settings();
```
dkoo, line 320: Nit: we can use `get_setting( 'site_key' )` here as a shortcut (since the options exist as individual rows and there's no need to fetch the other settings).

includes/class-recaptcha.php
```diff
@@ -286,6 +298,58 @@ public static function verify_captcha( $captcha_token ) {
...
+		?>
+		<script src="<?php echo \esc_url( self::get_script_url() ); ?>"></script>
+		<script>
```
dkoo, line 323: Part of me wants to ask if we should enqueue this as a JS file because of how we were talking the other day about not liking mixing PHP + JS, but this works and I don't see any security concerns about it. :)

dkoo (approved):
Approved with non-blocking nitpicks!

includes/class-recaptcha.php
miguelpeixe: Good call!

includes/class-recaptcha.php
miguelpeixe: Being such a small snippet, I'd rather have this approach. The enqueue handling adds more lines of code and I don't think is warranted for such a case.

In this case, we're also adding the script synchronously so we don't have to handle `grecaptcha` not being available when we use it.

PR #2460, opened 2023-05-17
fix: remove 'username' from Twitter menu label

laurelfulford:
### All Submissions:

As part of an accessibility review done for one of our publishers, we received feedback that the menu label "Twitter username" is sub-optimal for links pointing to a site's Twitter feed -- it's not really accurate since it's doesn't reflect what the link truly does. 

This PR removes the 'username' part of the Twitter label; because the way the plugin is set up, this also involves removing it from the visual label above the text field in the setup wizard and on the SEO screen. In my testing, this seems okay -- it's less clear that you should only enter your twitter username, but the validation on the field removes `https://twitter.com` if it's added by accident so it seems okay. 

See 1200550061930446-as-1203785172630073

2. Run through the Newspack Wizard and add a full Twitter URL (https://twitter.com/newspackpub) instead of just the username.
3. At the end, view the front-end of the site, and use the element inspector to confirm that the Twitter label is just "Twitter" in the social links menu. 

image

4. Navigate to WP Admin > Newspack > SEO and confirm only the Twitter username was saved in the settings (this is just to make sure removing the "username" part of the label doesn't mess things up!). 

Unfortunately this won't fix sites that are already live, but the menu label can be manually edited in the Customizer to remove "username" if it's an issue for them.

### Other information:

### DIFF
--- a/assets/wizards/seo/views/settings/index.js
+++ b/assets/wizards/seo/views/settings/index.js
@@ -78,7 +78,7 @@ class Settings extends Component {
 							placeholder={ __( 'https://facebook.com/page', 'newspack' ) }
 						/>
 						<TextControl
-							label={ __( 'Twitter Username', 'newspack' ) }
+							label={ __( 'Twitter', 'newspack' ) }
 							onChange={ value => onChange( { urls: { ...urls, twitter: value } } ) }
 							value={ twitter }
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):

laurelfulford:
Thanks @miguelpeixe! 🙌

PR #2462, opened 2023-05-17
feat(ras): mailchimp audience config

miguelpeixe:
### All Submissions:

Allow the Mailchimp audience to be configured for RAS integration. Also removes the `NEWSPACK_DATA_EVENTS_MAILCHIMP` feature flag.

image

1. Check out this branch and make sure you have Mailchimp as your ESP and RAS properly configured
2. Navigate to the "Newspack -> Engagement" page and while in the "Reader Activation" tab, click on "Show Advanced Settings"
3. Scroll down and confirm you see the Mailchimp integration section like in the image above
4. Select an audience and save
5. Refresh the page and confirm the saved audience persists
6. In a new session, register a new reader and make a donation
7. Wait until the event dispatches execute the handlers (with `define( 'NEWSPACK_LOG_LEVEL', 2 );` you should see:
```
[NEWSPACK-DATA-EVENTS][Newspack\Data_Events::handle]: Executing global action handlers for "donation_new".
```
8. Go to your Mailchimp audience and confirm the contact is updated with `NP_` prefixed metadata

### Other information:

### DIFF
--- a/assets/wizards/engagement/components/mailchimp.js
+++ b/assets/wizards/engagement/components/mailchimp.js
@@ -1,0 +1,54 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
+import { useEffect, useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { Notice, SectionHeader, SelectControl } from '../../../components/src';
+
+export default function ActiveCampaign( { value, onChange } ) {
+	const [ inFlight, setInFlight ] = useState( false );
+	const [ lists, setLists ] = useState( [] );
+	const [ error, setError ] = useState( false );
+	const fetchLists = () => {
+		setError( false );
+		setInFlight( true );
+		apiFetch( {
+			path: '/newspack-newsletters/v1/lists',
+		} )
+			.then( res => setLists( res.filter( list => ! list.type ) ) )
+			.catch( setError )
+			.finally( () => setInFlight( false ) );
+	};
+	useEffect( fetchLists, [] );
+	const handleChange = key => val => onChange && onChange( key, val );
+	return (
+		<>
+			{ error && (
+				<Notice
+					noticeText={ error?.message || __( 'Something went wrong.', 'newspack' ) }
+					isError
+				/>
+			) }
+			<SectionHeader
+				title={ __( 'Mailchimp', 'newspack' ) }
+				description={ __( 'Settings for the Mailchimp integration.', 'newspack' ) }
+			/>
+			<SelectControl
+				label={ __( 'Audience ID', 'newspack' ) }
+				help={ __( 'Choose an audience to receive reader activity data.', 'newspack' ) }
+				disabled={ inFlight }
+				value={ value.audienceId }
+				onChange={ handleChange( 'audienceId' ) }
+				options={ [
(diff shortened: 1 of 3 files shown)


assets/wizards/engagement/components/mailchimp.js
```diff
@@ -0,0 +1,54 @@
...
+import { __ } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
+import { useEffect, useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { Notice, SectionHeader, SelectControl } from '../../../components/src';
+
+export default function ActiveCampaign( { value, onChange } ) {
```
dkoo, line 13: paste error?
```suggestion
export default function Mailchimp( { value, onChange } ) {
```

dkoo (changes requested):
Just a minor typo, otherwise it works great! Another thing to note: this is probably expected because Mailchimp doesn't have the concept of a "master list" like AC (which contacts are added to but don't receive newsletter content from). If you donate in a new session without registering or signing up for a newsletter first, the contact will not be synced to Mailchimp. If you subsequently sign up for a newsletter, your donation data will be synced after the fact.

miguelpeixe:
Thanks, Derrick!

> If you donate in a new session without registering or signing up for a newsletter first, the contact will not be synced to Mailchimp.

That's not the expected behavior. Mailchimp supports transactional contacts, which is how we handle this API request, with an upsert strategy with `put()` and `status_if_new` set to `transactional`:

https://github.com/Automattic/newspack-plugin/blob/412a241562a70c40bb070858124095325fb669df/includes/data-events/connectors/class-mailchimp.php#L131-L158

Looking into it.

miguelpeixe:
I'm not sure if that was the case for you, but I caught an error with my local instance while generating the merge fields. My MC account has reached the limit of merge fields and creating a new field was returning an API error.

4596d7db20f3f29f39d266354fc6cade05110a31 skips fields that failed to create in order to allow the data push

dkoo (changes requested):
It seems to be working now, thanks! I didn't see any error before, but I'm seeing non-registered donors being synced immediately now.

I did notice some bugginess in the data syncing, though:

- The `NP_Total Paid` seems to be double what I'd expect after making a single one-time or recurring donation. e.g. if I donate $90 one time, this gets synced as $180. If I sign up for a $15 monthly subscription, it gets synced as $30.
- If I make subsequent donations using the same account, the `NP_Total Paid` amount updates incorrectly as well. With the recurring test donor, I made a $15 monthly donations, and `NP_Total Paid` got synced at first as $30. I then made a $9 one-time donation, and it got synced to $33.
- Based on this behavior I think we're adding the last payment amount to the total spend, since I would have expected the total to be $24 after the second donation.

miguelpeixe:
Can you confirm that this is only happening in the Mailchimp connector? This looks like a problem in the logic that generates the data, which is a separate code that also feeds ActiveCampaign's integration.

Trying to reproduce now

miguelpeixe:
Yep, reproduced with a new account donating $90 one-time, results in a $180 as "total paid". After paying an additional $20, the "total paid" changes to $130 😕

dkoo (approved):
I see similar but slightly different behavior with ActiveCampaign. The total after the first donation is doubled, just like in Mailchimp. After subsequent donations, the total comes in correctly. But since this seems unrelated to Mailchimp specifically, let's handle that in a separate hotfix PR. This one looks good!

PR #2464, opened 2023-05-17
fix(ras): total paid sum race condition

dkoo:
### All Submissions:

Fixes a possible race condition when calculating the total amount paid per donor when syncing the donor's info to the connected ESP. At the `woocommerce_checkout_order_processed` hook, the order status may be `pending` or `complete`, so the `$customer->get_total_spent()` method may or may not already include the order being processed in the total amount. This fix checks whether the status is `pending` and only adds the processing order amount to the total if it's still pending.

1. On `release`, with your reader revenue platform set to Newspack, ensure that RAS is setup and that the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` and `NEWSPACK_DATA_EVENTS_MAILCHIMP` flags are set in wp-config. Test the following with ActiveCampaign.
2. On the front-end, make a one-time donation as a new reader.
3. In the connected ESP, find the contact and observe that the `NP_Total Paid` amount is double what you'd expect. On subsequent donations the total would reflect the true total plus the amount from the latest order (because the latest order is added twice).
4. Check out this branch and repeat with a one-time donation, then a subscription donation. Confirm that the totals are synced correctly to the ESP for the first and subsequent donations.
5. Change your ESP connection to Mailchimp and repeat step 4.

### Other information:

### DIFF
--- a/includes/reader-revenue/class-woocommerce-connection.php
+++ b/includes/reader-revenue/class-woocommerce-connection.php
@@ -58,7 +58,7 @@ class WooCommerce_Connection {
 		\add_action( 'wc_memberships_user_membership_created', [ __CLASS__, 'wc_membership_created' ], 10, 2 );
 
 		// WC Subscriptions hooks in and creates subscription at priority 100, so use priority 101.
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
🙌

PR #2470, opened 2023-05-24
Only display author filter for posts

leogermani:
### All Submissions:

When first introduced in https://github.com/Automattic/newspack-plugin/pull/1985, it seemed a good idea to enable this filter for custom post types, but now it looks (at least to me) that we are interfering in other plugins behavior.

Posts page is where this filter is most useful anyway. Plugins that register custom post types might not want to have this filter on their CPT page.

1. Confirm that the filter still shows for posts and work
2. Confirm it does not show up for Pages or other custom post types

### Other information:

### DIFF
--- a/includes/author-filter/class-author-filter.php
+++ b/includes/author-filter/class-author-filter.php
@@ -61,14 +61,19 @@ class Author_Filter {
 	 */
 	public static function author_filters_admin( $post_type ) {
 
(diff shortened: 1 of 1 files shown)


leogermani:
It's disabled if you have CAP enabled.

We did that because there was a performance issue with CAP that would break the page. But I believe we have addressed that issue and we could re-enable it. But that's something for another PR and testing.

PR #2482, opened 2023-05-31
feat(plugin-installer): handle premium plugins in PluginInstaller

dkoo:
### All Submissions:

A small tweak to the PluginInstaller component that disables the "Install" action button if the plugin can't be installed programmatically (e.g. if the plugin is a premium plugin without a public download link). Previously the component would attempt to install but fail with an error message due to lacking a URL to install from, which is a subpar user experience.

Also tweaks the message shown while activating a plugin to "Activating...". Previously this said "Installing..." whether installing or activating a plugin.

Closes `1204498218048567/1204685119775505`.

This applies to any instance of the PluginInstaller and any plugin that lacks a `Download` property in [`class-plugin-manager.php`](https://github.com/Automattic/newspack-plugin/blob/master/includes/class-plugin-manager.php#L39), but the easiest place to test is in the new Reader Activation wizard:

1. Delete `woocommerce` (publicly downloadable) and the premium `woocommerce-subscriptions` and `woocommerce-name-your-price` plugins.
2. Visit **Newspack > Engagement > Reader Activation** and confirm that WooCommerce can be installed from here, but the other two plugins cannot, and show a message to contact Newspack support for those plugins.

Screen Shot 2023-05-31 at 4 46 37 PM

3. Manually install the plugins but do not activate.
4. Refresh the wizard, and confirm that the plugins can now be activated (since they're already installed, just inactive) and that they show an "Activating..." message while activating.

### Other information:

### DIFF
--- a/assets/components/src/action-card/index.js
+++ b/assets/components/src/action-card/index.js
@@ -173,6 +173,7 @@ class ActionCard extends Component {
 									</Handoff>
 								) : onClick || hasInternalLink ? (
 									<Button
+										disabled={ disabled }
 										isLink
 										href={ href }
 										onClick={ onClick }
--- a/assets/components/src/plugin-installer/index.js
+++ b/assets/components/src/plugin-installer/index.js
@@ -189,12 +189,21 @@ class PluginInstaller extends Component {
 					slugs.length > 0 &&
 					slugs.map( slug => {
 						const plugin = pluginInfo[ slug ];
-						const { Name, Description, Status, installationStatus, notification } = plugin;
+						const { Name, Description, Download, Status, installationStatus, notification } =
+							plugin;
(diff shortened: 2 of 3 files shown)


miguelpeixe:
Ideally, check for the `ATOMIC_SITE_ID` constant so a properly targeted message is sent.

dkoo:
Oh, great idea! I'll revisit this.

dkoo:
@leogermani @miguelpeixe 3fc8b26 adds a check for the `ATOMIC_SITE_ID` constant and shows the "Contact Newspack support" message if it exists, otherwise it shows a more generic "Plugin must be installed manually" message.

miguelpeixe (approved):
🎉 🙌

PR #2453, opened 2023-05-11
feat(ras-defaults): warn if navigating away from wizard with unsaved changes

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

If you make changes to the prompt settings in the RAS setup prompt wizard and try to navigate away from the wizard without saving, this will pop up a browser warning.

Screen Shot 2023-05-11 at 5 19 16 PM

**Note:** For some reason, unlike the same feature in the **Campaigns > Segmentation** dashboard, React Router will not reset the route if you block the navigation by selecting "cancel" when prompted. I couldn't figure out why, but if someone can help me figure it out I'm happy to update the PR.

### How to test the changes in this Pull Request:

1. Delete the `newspack_reader_activation_enabled` option on your test site.
2. Navigate to `<domain>/wp-admin/admin.php?page=newspack-engagement-wizard#/reader-activation/campaign` and make some changes to prompts without saving. Confirm that the prompt's status changes to indicate that there are "Unsaved changes", and if the prompt was previously saved and ready, the ready status is removed.
3. Try to navigate to another route in the wizard, or refresh the page/close the browser tab. Confirm you see a warning asking you to confirm as in the above screenshot. (Note that if trying to refresh or close the browser tab, you will see a generic warning instead of the custom message—this is due to browser limitations in setting the message on the `beforeunload` event.)
4. Repeat step 2, but this time save the unsaved prompt(s) before trying to navigate away, and confirm that you can without being warned.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2483, opened 2023-06-02
fix(ras): check that unblock exists before calling it

dkoo:
### All Submissions:

Fixes an intermittent error introduced by #2453. Sometimes the `hooks.usePrompt` method doesn't return a function before the `savePrompt` method is called and attempts to call `unblock`. This avoids a JS error in that case.

1. Follow testing instructions in #2453. Observe that if you make changes and save a prompt ASAP quickly, you'll sometimes see a JS error: `Uncaught (in promise) TypeError: unblock is not a function`
2. Check out this branch and repeat. Confirm that you don't see the error, and that the prompt still works when you try to navigate away from the page with unsaved changes.

### Other information:

### DIFF
--- a/assets/wizards/engagement/components/prompt.tsx
+++ b/assets/wizards/engagement/components/prompt.tsx
@@ -124,7 +124,9 @@ export default function Prompt( { inFlight, prompt, setInFlight, setPrompts }: P
 
 	const savePrompt = ( slug: string, data: InputValues ) => {
 		return new Promise< void >( ( res, rej ) => {
-			unblock();
+			if ( unblock ) {
+				unblock();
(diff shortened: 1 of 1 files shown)


leogermani (approved):
fixes it for me. I don't see errors and the notice is still there

dkoo:
To record our discussion, my theory is that the error was happening when trying to save without making any changes (the `isDirty` state is `false`). When `hooks.usePrompt` was called with a falsy value, it would return nothing instead of the `unblock` function. The fix here works because we don't use the hook if there are no unsaved changes (no need to prompt the user before leaving, so nothing to unblock).

PR #2503, opened 2023-06-13
Update/ga4 backend destinations

leogermani:
### All Submissions:

Removes the feature flag for GA4 back-end events.

Add the ability to send the events to more than one property via a filter.

1. Have your site connected with two different properties in GA
2. Make sure you are running Newspack Manager on https://github.com/Automattic/newspack-manager/pull/128
3. Add the options that will set the "local" GA property:

```
update_option( 'ga4_measurement_protocol_secret', 'xxx' );
update_option( 'ga4_measurement_id', 'G-XXX' );
```

5. Add the option that will set the property that is handled by Newspack Manager:

```
update_option( 'newspack_ga4_info', [ 'measurement_id'=>'G-ZZZ, 'measurement_protocol_secret'=>'zzz'  ] );
```

6. Visit your site and open the two GA Real time dashboards
7. Interact with your site and perform actions that will trigger back-end events, like submitting forms from a prompt
8. Check that the back end events, such as `np_reader_registered` and `np_prompt_interaction` with a `form_submission_received` as the `action` param.

### Other information:

### DIFF
--- a/includes/data-events/connectors/ga4/class-ga4.php
+++ b/includes/data-events/connectors/ga4/class-ga4.php
@@ -406,22 +406,22 @@ class GA4 {
 			$payload['user_id'] = $user_id;
 		}
 
-		$destinations = [
+		$properties = [
 			self::get_credentials(),
 		];
 
 		/**
-		 * Filters the destinations of the GA4 events in the GA4 Data Events connector.
+		 * Filters the properties of the GA4 events in the GA4 Data Events connector.
 		 *
-		 * Each destination is an array with two keys: `measurement_id` and `measurement_protocol_secret`.
+		 * Each property is an array with two keys: `measurement_id` and `measurement_protocol_secret`.
 		 *
-		 * @param array $destinations The destinations.
+		 * @param array $properties The properties.
 		 */
-		$destinations = apply_filters( 'newspack_data_events_ga4_destinations', $destinations );
+		$properties = apply_filters( 'newspack_data_events_ga4_properties', $properties );
 
-		foreach ( $destinations as $destination ) {
+		foreach ( $properties as $property ) {
 
-			$url = self::get_api_url( $destination['measurement_id'] ?? '', $destination['measurement_protocol_secret'] ?? '' );
+			$url = self::get_api_url( $property['measurement_id'] ?? '', $property['measurement_protocol_secret'] ?? '' );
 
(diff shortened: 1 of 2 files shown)


includes/data-events/connectors/ga4/class-ga4.php
```diff
@@ -412,14 +406,37 @@ private static function send_event( Event $event, $client_id, $timestamp, $user_
...
 
-		wp_remote_post(
-			$url,
-			[
-				'body' => wp_json_encode( $payload ),
-			]
-		);
+		$destinations = [
+			self::get_credentials(),
+		];
```
miguelpeixe, line 411: Shouldn't this array start empty and only include the local credentials if they are filled?

dkoo (commented):
Wanted to note that I ran through the testing steps and all is working as described. I'll let @miguelpeixe approve based on his change requests.

includes/data-events/connectors/ga4/class-ga4.php
leogermani: > Shouldn't this array start empty and only include the local credentials if they are filled?

This is basically what happens. `get_credentials` will return an array with empty values.

> Also, what do you think of calling it $properties?

Agree. Changed in https://github.com/Automattic/newspack-plugin/pull/2503/commits/7cdf13a892053973f80dc1868a98ad2e92733ec9

Also updated the README and changed the filter name on the manager plugin -> https://github.com/Automattic/newspack-manager/pull/128/commits/83d82123e5a491e7b9b540d0c39f24a3d1950ffd

miguelpeixe (approved):
Thanks for revising, it tests well!

With the current strategy of an array with an empty value, I'm always getting the `Error sending event - np_prompt_interaction - Error: Missing GA4 API credentials.` because it tries to fetch the API URL when GA4 wasn't yet configured. If this is expected, it's ok to merge.

leogermani:
> With the current strategy of an array with an empty value, I'm always getting the Error sending event - np_prompt_interaction - Error: Missing GA4 API credentials. because it tries to fetch the API URL when GA4 wasn't yet configured. If this is expected, it's ok to merge.

Yes, that's expected for now. Notice that this only gets outputted to the error log with debugging mode enabled.

I'm considering all this temporary as we still need to figure out how we'll handle the "local" credentials

PR #2507, opened 2023-06-14
fix: allow high-res youtube thumbs via constant

dkoo:
### All Submissions:

This is a quick fix to satisfy a publisher request to force Perfmatters to use high-resolution YouTube thumbnails in YT embeds. The reason we're doing this via a constant is that it's a tradeoff—it will result in a slight decrease in CWV scores, so it's not clear that it should be a default behavior. And it's also unclear whether this is something most sites would want the ability to set, so there's no need to implement UI for now.

1. On a site with Perfmatters enabled, embed a YouTube video into a post. Try this one, which is easy to see the difference in resolution because of the high-contrast text: https://www.youtube.com/watch?v=XZxWRaxqxGI
2. View on the front-end in a large desktop browser window. Observe that the thumbnail looks a bit pixelated and blurry.
3. Check out this branch and add `define( 'NEWSPACK_PERFMATTERS_USE_HIGH_RES_YOUTUBE_IMAGES', true );` to your wp-config.php.
4. Refresh on the front-end and confirm that the thumbnail is now high-resolution and crisp.
5. View the same post on a mobile device and confirm that the thumbnail is still lower-resolution on that device (because speed is paramount on mobile devices).

### Other information:

### DIFF
--- a/includes/plugins/class-perfmatters.php
+++ b/includes/plugins/class-perfmatters.php
@@ -19,6 +19,7 @@ class Perfmatters {
 	public static function init() {
 		add_filter( 'option_perfmatters_options', [ __CLASS__, 'set_defaults' ] );
 		add_action( 'admin_notices', [ __CLASS__, 'admin_notice' ] );
+		add_filter( 'perfmatters_lazyload_youtube_thumbnail_resolution', [ __CLASS__, 'maybe_serve_high_res_youtube_thumbs' ] );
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Works as described!

leogermani:
Do we have  Newspack - Perfmatters page somewhere we could document this?

PR #2520, opened 2023-06-23
feat: add 'article_view' reader activity data

miguelpeixe:
### All Submissions:

Implements `article_view` reader activity to the reader data library integrated with memberships and metering.

Content locked from memberships and front-end or back-end metering strategies should not push an `article_view` activity.

This PR also implements the `getUniqueActivitiesBy( name, iteratee )` helper function to allow fetching unique activity based on its data, e.g.:

```js
const uniquePostsViews = ras.getUniqueActivitiesBy( 'article_view', 'post_id' );
```

1. Make sure you have WC Memberships configured with content gating and metering (anonymous and registered readers)
2. On a fresh anonymous session visit an article
3. Open DevTools and navigate to the localStorage
4. Confirm there's a new item in the `newspack_reader_1_activity` array similar to this:
```js
{
  "action": "article_view",
  "data": {
    "post_id": 120,
    "permalink": "https://example.com/montes-egestas-a-neque-morbi-cras-maecenas-senectus-molestie-in-congue-consequat-sed/",
    "categories": [ 4, 3 ],
    "tags": [],
    "author": "johndoe"
  },
  "timestamp": 1687530871161
}
```
5. Navigate until content is locked and confirm the anonymous metering locked content does not push a new activity
6. Register and do the same with registered metering, navigate confirming activities are pushed until it locks and no longer pushes
7. Become a member and confirm activity pushes as a member
8. Deactivate WC Memberships, navigate and confirm activity continues to push correctly

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-metering.php
+++ b/includes/plugins/wc-memberships/class-metering.php
@@ -18,6 +18,11 @@ class Metering {
 	const METERING_META_KEY = 'np_memberships_metering';
 
 	/**
+	 * Article view activity to be handled by frontend metering.
+	 */
+	private static $article_view = null;
+
+	/**
 	 * Cache of the user's metering status for posts.
 	 *
 	 * @var boolean[] Map of post IDs to booleans.
@@ -30,7 +35,8 @@ class Metering {
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
 		add_action( 'wp', [ __CLASS__, 'handle_restriction' ], 11 );
-		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
+		add_action( 'wp_footer', [ __CLASS__, 'enqueue_scripts' ] );
+		add_filter( 'newspack_reader_activity_article_view', [ __CLASS__, 'get_article_view' ], 20 );
 	}
 
 	/**
@@ -102,6 +108,7 @@ class Metering {
 				'count'              => \get_post_meta( $gate_post_id, 'metering_anonymous_count', true ),
 				'period'             => \get_post_meta( $gate_post_id, 'metering_period', true ),
 				'post_id'            => get_the_ID(),
+				'article_view'       => self::$article_view,
 			]
 		);
 	}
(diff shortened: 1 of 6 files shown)


leogermani (approved):
Works well after your last fix!

miguelpeixe:
@leogermani @dkoo could use some help figuring out why the PHP linter is erroring here...

leogermani:
I'm able to reproduce the lint error on my local... probably some conflict between package and php versions.. I'm looking into it

miguelpeixe:
Thanks! I tried rebasing with master a few times without luck. It's definitely triggered by something in this code 😕

includes/plugins/wc-memberships/class-metering.php
```diff
@@ -270,5 +279,18 @@ public static function is_logged_in_metering_allowed( $post_id = null ) {
...
+
+	/**
+	 * Store the article view activity push for use in the frontend metering
+	 * strategy.
+	 *
+	 * @param array $activity Activity data.
+	 *
+	 * @return array
+	 */
+	public static function get_article_view( $activity )  {
```
leogermani, line 291: @miguelpeixe the linting error is here. two spaces after the parenthesis. Now I'm trying to understand why phpcs is breaking

miguelpeixe: Ah ha! 🙇

miguelpeixe:
While investigating a bug I found with metering having some inconsistent dispatches I ended up tweaking some code placement as well.

de00bd9b50e9208b009f1b48cd2322bb8442f49f moves the code to the frontend by localizing the activity push setup. It also corrects a bug with metering dispatch that I couldn't reproduce anywhere else. For some reason when it uses the array push strategy it doesn't run the dispatch, only when it uses the replaced push handler, it becomes a coin toss. I guess it might have to do with the method already running inside a push... Moving to calling the dispatch directly from RAS fixes the issue and is also a better practice since the code is already in a push.

miguelpeixe:
If you could take another look @leogermani . Sorry for the back and forth 😬

leogermani:
I'll have a look. 

In the meantime, this is the reason for the WPCS to crash: https://github.com/WordPress/WordPress-Coding-Standards/commit/7cd46bed1e6a7a2af3fe24c7f4a044da3076d8f4

TL;DR is:
* If you run phpcs on PHP 8, it will crash on some sniffs.
* recommendation is to run phpcs on php 7.4
* There is a [tag](https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/2.3.0-fix-vsprintf) with the fixes we could use, but those also include other unstable changes...

I guess we should look to see if it's possible to configure circleCI to use php 7.4 for this particular job. If not, use the [unstable tag](https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/2.3.0-fix-vsprintf).

miguelpeixe:
Thank you!

Nice finding on the linter issue, should we open a gh issue for dedicated discussion?

leogermani (approved):
I like this approach much better!

PR #2525, opened 2023-06-26
fix: check next_payment_dates for past dates and update those, too

dkoo:
### All Submissions:

https://github.com/Automattic/newspack-plugin/pull/2434 introduced a fix to ensure that Stripe subscriptions migrated to Woo have a proper start date and next payment date set, matching the original subscription. However, on at least one site I saw an edge case where the next payment date got migrated as a date in the past, resulting in confusion on the part of WooCommerce. This updates the `wp newspack stripe set-next-payment-dates-for-migrated-subscriptions` script to check for next payment dates in the past (as well as empty ones) and updates those, and it also checks next payment dates upon migrating Stripe subs to Woo (so there should be no reason to run the `wp newspack stripe set-next-payment-dates-for-migrated-subscriptions` script separately in the future—it would only be to catch subscriptions migrated with incorrect next payment dates before this fix).

This one might be tough to create the conditions for testing, so you may have to set the next payment date manually to a date in the past.

1. Create a Woo subscription by making a recurring donation.
2. Using WP CLI or WP shell, set the next payment date to a date in the past (I had to do the latter because the date string kept messing up the CLI command, for some reason):

```
$ wp shell
$ update_post_meta( <postID>, '_schedule_next_payment', '2023-06-22 18:41:43' );
```

3. This should result in the subscription being in an awkward state:

Screen Shot 2023-06-26 at 5 39 57 PM

4. Run `wp newspack stripe set-next-payment-dates-for-migrated-subscriptions` and confirm that the next payment date is updated to a future date, based on the original start date of the subscription.

### Other information:

### DIFF
--- a/includes/reader-revenue/stripe/class-stripe-sync.php
+++ b/includes/reader-revenue/stripe/class-stripe-sync.php
@@ -492,37 +492,10 @@ Running script to set next payment dates on migrated subscriptions...
 			$subscription_id = array_shift( $migrated_subscriptions );
 
 			// Set next payment date.
-			$subscription = \wcs_get_subscription( $subscription_id );
-			if ( $subscription ) {
-				\WP_CLI::log(
-					sprintf(
-						'Found subscription with ID %d and start date %s.',
-						$subscription_id,
-						$subscription->get_date( 'start' )
-					)
-				);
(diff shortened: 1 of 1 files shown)


dkoo:
Closing—these scripts have been moved to a separate repository.

PR #2526, opened 2023-06-28
feat(data-events): broader coverage for 'reader_registered'

miguelpeixe:
### All Submissions:

Changes the `reader_registered` action to listen to WP's `user_register` action hook instead of the hook coming from `Reader_Activation:register_reader()`.

This allows for broader coverage of reader registrations, like from WC product purchases.

The original metadata coming from `Reader_Activation:register_reader()` is preserved by storing its entire value as a `np_registration_metadata` user meta.

Note: tests were initially failing because the `user_register` will trigger on mock users created during unit testing. This triggers a filter inside the GA4 connector that calls for `wp_get_current_user()`, which fails the `is_reader()` check. 47469878b56ab3416dac27d85e5c05dc4cf4aebd (with a fix at 999622040a36f72df7f3f1c0e339f3d3b1e51986) fixes that behavior by preventing GA4 connector logic to run when not configured. If we don't want to touch the GA4 logic I can look at how to prevent the error on the tests. cc @leogermani 

1. Make sure you have `define( 'NEWSPACK_EXPERIMENTAL_WEBHOOKS', true );`
2. Setup a webhook using https://webhook.site/ to get all incoming data events 
3. Purchase a product anonymously and confirm that the created account triggers a `reader_registered` event and it reaches the webhook
4. On a fresh session, register through a RAS prompt with a reader registration block
5. Confirm the event reaches the webhook with no regression on the metadata contents

### Other information:

### DIFF
--- a/includes/data-events/connectors/ga4/class-ga4.php
+++ b/includes/data-events/connectors/ga4/class-ga4.php
@@ -59,7 +59,7 @@ class GA4 {
 	 */
 	public static function can_use_ga4() {
 		$properties = self::get_ga4_properties();
-		return ! empty( $propeties ) && ! empty( $properties[0]['measurement_id'] );
+		return ! empty( $properties ) && ! empty( $properties[0]['measurement_id'] );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


leogermani (approved):
Looks good and tests well

PR #2527, opened 2023-06-29
feat: trigger reader_registered for Woo checkout

leogermani:
### All Submissions:

https://github.com/Automattic/newspack-plugin/pull/2526 didn't actually work well. The metadata was being added after the user was registered, so the event was never sending any metadata.

This PR reverts the event to listen to the `newspack_registered_reader` hook and creates a new hook specifically for the readers registered during a Woocommerce checkout.

1. Setup a webhook using https://webhook.site/ to get all incoming data events 
2. Purchase a product anonymously and confirm that the created account triggers a `reader_registered` event and it reaches the webhook
3. On a fresh session, register through a RAS prompt with a reader registration block
4. Confirm the event reaches the webhook with no regression on the metadata contents
5. Test purchasing products and making donations via different entry points and check the metadata includes what's expected, see list below. (Important: be sure to not be logged in before making the donations/purchases)

A registration triggered by a donation in a campaign prompt should include:
* registration_method = woocommerce
* referer
* newspack_popup_id

A registration triggered by a donation outside a campaign prompt should include:
* registration_method = woocommerce
* referer

A registration triggered by any other Woocommerce checkout should include
* registration_method = woocommerce

### Other information:

### DIFF
--- a/includes/data-events/class-woo-user-registration.php
+++ b/includes/data-events/class-woo-user-registration.php
@@ -1,0 +1,99 @@
+<?php
+/**
+ * Newspack Data Events Woo Registration hooks.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Data_Events;
+
+/**
+ * Class that triggers a registration event with Newspack metadata for users registered during the Woocommerce checkout process.
+ */
+final class Woo_User_Registration {
+
+	/**
+	 * Whether the current request is processing a checkout.
+	 *
+	 * @var boolean
+	 */
+	private static $processing_checkout = false;
+
+	/**
+	 * Metadata to send with the registration event.
+	 *
+	 * @var array
+	 */
+	private static $metadata = [];
+
+	/**
+	 * Initialize the class.
+	 *
+	 * @return void
+	 */
+	public static function init() {
+		// is processing checkout?
+		add_action( 'woocommerce_checkout_process', [ __CLASS__, 'checkout_process' ] );
+
+		// created a user?
+		add_action( 'woocommerce_created_customer', [ __CLASS__, 'created_customer' ], 1 );
+	}
+
+	/**
+	 * During the checkout process, set the processing_checkout flag and store metadata.
+	 *
+	 * @return void
+	 */
+	public static function checkout_process() {
+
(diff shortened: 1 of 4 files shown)


includes/data-events/class-woo-user-registration.php
```diff
@@ -0,0 +1,99 @@
...
+		 * Here, we are going to read the same information from the cart and use it to send the metadata to Newspack on registration events.
+		 */
+		foreach ( \WC()->cart->get_cart() as $cart_item_key => $values ) {
+			if ( ! empty( $values['newspack_popup_id'] ) ) {
+				self::$metadata['newspack_popup_id'] = $values['newspack_popup_id'];
+			}
+			if ( ! empty( $values['referer'] ) ) {
+				self::$metadata['referer'] = $values['referer'];
+			}
+		}
```
leogermani, line 66: This data will only exist for Newspack donations.. maybe we could then set registration_method to "newspack_donation" instead of a generic "woocommerce" ?

miguelpeixe: Would the registration_method still be `woocommerce` if it's a premium subscription purchase (not donation)?

miguelpeixe (approved):
Tests well!

includes/data-events/class-woo-user-registration.php
leogermani: yes. wdyt?

leogermani: I could add `\WC()->cart->get_cart()` as a metadata to the event.

But this could also be fetched by the integration directly from Woocommerce if needed. No need to attach it to the event...

miguelpeixe: I think having the `registration_method` consolidated as `woocommerce` whenever WC is registering makes more sense to me, but I don't mind if you'd rather have it as `newspack_donation` for this case and `woocommerce` for every other.

leogermani: I agreed to keep it woocommerce in every case. I was just pointing that we don't need to add additional metadata. If we want to look for the products in the cart in a particular integration we can look directly into the cart itself :)

PR #2529, opened 2023-06-29
fix(content-gate): close excerpt tags before inline gate

miguelpeixe:
### All Submissions:

Implements [`force_balance_tags()`](https://developer.wordpress.org/reference/functions/force_balance_tags/) to the generated excerpt so the inline gate is never caught inside a blockquote or other parent tags the final `<p />` may be in. Also adds a clearfix to clear from any floating elements it may encounter.

1. While on master setup memberships and a gate without metering (so the backend strategy is used for locking content)
2. Set the gate number of paragraphs to 2
3. Create an article that will be gated and add a paragraph block and a blockquote block second with sample content, e.g.:
```html
<!-- wp:paragraph -->
<p>First Paragraph</p>
<!-- /wp:paragraph -->
<!-- wp:quote {"style":{"typography":{"fontStyle":"normal","fontWeight":"700"}},"className":"is-style-default","fontSize":"large"} -->
<blockquote class="wp-block-quote is-style-default has-large-font-size" style="font-style:normal;font-weight:700"><!-- wp:paragraph -->
<p>Testing Quote</p>
<!-- /wp:paragraph --><cite>Mary</cite></blockquote>
<!-- /wp:quote -->
```
4. Publish, visit the post anonymously, and confirm the gate is inside the blockquote and inheriting its CSS
5. Check out this branch, build, and refresh
6. Confirm the gate is outside the blockquote and formatted as expected

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -317,6 +317,9 @@ class Memberships {
 		}
 		$gate = \apply_filters( 'the_content', \get_the_content( null, null, $gate_post_id ) );
 
+		// Add clearfix to the gate.
+		$gate = '<div style=\'content:"";clear:both;display:table;\'></div>' . $gate;
+
 		// Apply inline fade.
 		if ( \get_post_meta( $gate_post_id, 'inline_fade', true ) ) {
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Looks good!

PR #2537, opened 2023-07-03
feat: segments criteria ui

miguelpeixe:
### All Submissions:

1204189396255901-as-1204857763418497

Changes the Segments UI to use the new criteria system and updated schema.

Confirm updated tests are passing and have appropriate coverage.

1. Checkout this branch and https://github.com/Automattic/newspack-popups/tree/epic/rearchitecture
2. Create, edit and delete segments and confirm everything works as expected
3. Create a segment using all available criteria
4. Visit the "Segments" home and confirm the description renders correctly for every item
5. Visit the "Campaigns" home and confirm the same description also renders correctly

### Other information:

### DIFF
--- a/__mocks__/@wordpress/api-fetch.js
+++ b/__mocks__/@wordpress/api-fetch.js
@@ -1,8 +1,5 @@
 const apiFetch = ( { path } ) =>
 	new Promise( resolve => {
-		if ( path.indexOf( '/newspack/v1/wizard/newspack-popups-wizard/segmentation-reach' ) === 0 ) {
-			resolve( { total: 10, in_segment: 10 } );
-		}
 		resolve();
 	} );
 
--- a/assets/components/src/settings/MinMaxSetting.js
+++ b/assets/components/src/settings/MinMaxSetting.js
@@ -16,9 +16,10 @@ const MinMaxSetting = ( {
 	onChangeMax,
 	minPlaceholder,
 	maxPlaceholder,
+	...props
 } ) => {
 	return (
-		<>
+		<div { ...props }>
 			<div className="newspack-settings__min-max">
 				<CheckboxControl
 					checked={ min > 0 }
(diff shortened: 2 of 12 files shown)


leogermani:
@miguelpeixe 

https://github.com/Automattic/newspack-plugin/pull/2537/commits/14761601f8cb63a393857549f3343eab055688d9 fixes an issue that criteria was being deleted if you toggled the status from the segments list. -> Note that this will only work after the segment was migrated. But I think that's fine, we will run the migration for all segments soon.

https://github.com/Automattic/newspack-plugin/pull/2537/commits/e293da105dd69c3375332a6558652d55e03b1f85 fixes an issue that the Save button was not being enabled if you toggled the status from the segment page

and https://github.com/Automattic/newspack-plugin/pull/2537/commits/bbd5d9fe37566329c0a36591879cbf7da767d23e gets rid of some warnings

Is there anything else you want to address here? I think it's good to go

miguelpeixe:
Thanks for the fixes!

> Is there anything else you want to address here? I think it's good to go

I'm still working on the draft to make sure all wizard screens understand the new criteria system. It's still not working well when rendering the segment description on the `SegmentGroup` component (Campaigns tab)

I'll place it for code review when it's ready, it should go through regular code review before merging to the epic branch.

miguelpeixe:
@leogermani can you show me the warnings that warranted bbd5d9fe37566329c0a36591879cbf7da767d23e? I'm not getting any in my instance and it should be handled already by returning `null` with `getCriteriaValue()`.

leogermani:
> @leogermani can you show me the warnings that warranted [bbd5d9f](https://github.com/Automattic/newspack-plugin/commit/bbd5d9fe37566329c0a36591879cbf7da767d23e)? I'm not getting any in my instance and it should be handled already by returning `null` with `getCriteriaValue()`.

Example:

```
Warning: `value` prop on `select` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

miguelpeixe:
Thanks! For some reason, I'm not getting that warning. I changed the handling so the function doesn't return the unhandled value 8e0193c92751b774cb26ecd76cc77332ca724ccd.

miguelpeixe:
@dkoo [commented](https://github.com/Automattic/newspack-popups/pull/1159#pullrequestreview-1538145518):

> I ran into an issue when editing an existing segment. Updating the select dropdown values doesn't always update the isDirty state, so the "Save" button remains disabled until you edit another field. Editing other types of fields seems to work as expected, and the issue is intermittent with dropdowns. I'm approving this PR because I think the issue is coming from https://github.com/Automattic/newspack-plugin/pull/2537, not this one.

dkoo (changes requested):
Looks good, and mostly works! The only thing is [this](https://github.com/Automattic/newspack-plugin/pull/2537#issuecomment-1642862965) is still happening for me. Replication steps:

1. Create a new segment and set all the of the select dropdowns to something other than the default. Save.
2. Edit the segment again. Change the values of the dropdowns to another value that is not the default. Observe that the "Save" button doesn't become active.
3. If you add `console.log( segmentCriteriaInitially, segmentCriteria );` to https://github.com/Automattic/newspack-plugin/pull/2537/files#diff-a975ee831b86af1483494488d751b64b384c16269ad9450f03c7c697606548d1 right where `isDirty` is being defined, you'll see that the values of `segmentCriteriaInitially` are somehow being updated on change, so `isDirty` is set to `false`.

miguelpeixe:
Good catch! e293da105dd69c3375332a6558652d55e03b1f85 changed a strategy that was using `JSON.stringify()` instead of  `[ ...array ]` cloning for checking criteria changes.

Fixed with 76421c70667793fda7e662b9f5f519e4671510d2

dkoo (approved):
That did it!

PR #2539, opened 2023-07-04
feat(reader-data): add newsletter subscription and donation

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Implements the following common reader data:
 - `is_newsletter_subscriber`
 - `is_donor`
 - `is_former_donor`

### How to test the changes in this Pull Request:

1. Check out this branch and subscribe to a newsletter in a fresh session (either through Newsletter Subscription Form block or Reader Registration block)
2. Confirm the next page render have the `is_newsletter_subscriber` store item set as true either by running `newspackReaderActivation.store.get('is_newsletter_subscriber')` on the frontend or checking the localStorage items in devtools
3. Do a one-time donation through the Donate block
4. Confirm the next page render has the `is_donor` set as true and `is_former_donor` set as false
5. Subscribe to a recurring donation
6. Cancel the subscription through "My Account" (it'll be set as pending and not canceled just yet)
7. Expedite the cancellation on the dashboard (WooCommerce -> Subscriptions -> Click Cancel Now on the order item)
8. Refresh the page on the reader session and confirm `is_donor` is false and `is_former_donor` is true

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2541, opened 2023-07-04
feat(reader-data): set referrer

miguelpeixe:
### All Submissions:

Similar to #2520 and #2539, set the referrer store data item if coming from an external source.

1. Create a link to your instance from a different domain (use Google Drive, Gmail, or any code playground)
2. Click on the link to visit the site and confirm the referrer is stored (`newspackReaderActivation.store.get( 'referrer' )`)
3. Navigate on the site and confirm the referrer persists (not replaced by referring itself)

### Other information:

### DIFF
--- a/assets/reader-activation/index.js
+++ b/assets/reader-activation/index.js
@@ -237,6 +237,16 @@ function fixClientID() {
 }
 
 /**
+ * Store the referrer.
+ */
+function setReferrer() {
+	const referrer = document.referrer ? new URL( document.referrer ).hostname : '';
+	if ( referrer && referrer !== window.location.hostname ) {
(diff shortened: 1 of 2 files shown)


leogermani (commented):
Looks good but couldn't this be done directly on the front-end?

leogermani (approved):
Looks good.

What about what we've discussed about keeping the original external referrer when the user navigates to internal pages on the site? Want to address this later?

miguelpeixe:
It should be working like that. It doesn't update if the referrer is empty or it matches the window's location hostname

leogermani:
> It should be working like that. It doesn't update if the referrer is empty or it matches the window's location hostname

ah, you're right!

PR #2553, opened 2023-07-11
Feat: add an image size control for the RSS images

laurelfulford:
### All Submissions:

We recently switched the RSS feed to use the large image rather than the medium image size. Unfortunately, this has caused email layout issues on sites using Active Campaign -- the email layouts look bad, and there isn't an option in Active Campaign itself to scale the image down. As a compromise between the two, this PR adds an RSS image size control to the Media Settings page. It still defaults to the standard WordPress large size (1024), but can be changed from there.

Note: this will be a breaking change for any publications with non-standard large image sizes; we may at least want to make a more noticeable announcement when this goes out.

See: 1200550061930446-as-1204959665169392

1. Apply the PR.
2. Create and publish a new post with a featured image that's newly uploaded to your Media Library (alternatively, you can regenerate your site's thumbnails, but new images can be nice for comparing). 
3. View your RSS feed to confirm your new post's image uses maximum image width and height of 1024 -- for vertical images, this means the maximum width is 1024, and the height will reflect the image's aspect ratio. You can test this by subscribing to the RSS feed in an RSS reader, or you can _kind of_ see it in action by visiting [yourtestsite.com]/feed in Chrome and looking at the image size in the code. 
4. Next we'll change the image settings -- navigate to Settings > Media Settings.
5. Confirm you have an option to set an RSS Image size at the bottom, and that it's set to a maximum width of 1024, and a maximum height of 1024.
6. Change the height or width values, or both. I recommend changing to a smaller size since that matches our actual use case, being able to make these images smaller for Active Campaign.
7. Repeat step two (creating a new post with a newly uploaded image).
8. Revisit your feed in an RSS reader, and confirm that the new post is using your new maximum height or width. 

Note: the fields for the RSS Image aren't 1:1 with the fields for the other images -- I used the standard `do_settings_sections` output, which doesn't group them the same way, and places them in separate table cells with are spaced pretty far apart. I "worked around" this with some inline styles because it seemed like the most straight-forward approach, but I'm definitely open to other suggestions!

### Other information:

### DIFF
--- a/includes/class-rss-add-image.php
+++ b/includes/class-rss-add-image.php
@@ -11,12 +11,156 @@ namespace Newspack;
  * RSS_Add_Image class.
  */
 class RSS_Add_Image {
+
+	/**
+	 * Option key for the RSS image width setting.
+	 *
+	 * @var string
+	 */
+	const OPTION_RSS_IMAGE_WIDTH = 'newspack_rss_image_size_width';
+
+	/**
+	 * Option key for the RSS image height setting.
+	 *
+	 * @var string
+	 */
+	const OPTION_RSS_IMAGE_HEIGHT = 'newspack_rss_image_size_height';
+
 	/**
 	 * Hook actions and filters.
 	 */
 	public static function init() {
+		add_action( 'admin_init', [ __CLASS__, 'add_rss_image_size_section' ] );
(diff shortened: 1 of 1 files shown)


dkoo (approved):

laurelfulford:
Thanks @dkoo! 🎉

PR #2585, opened 2023-07-28
feat: add anonymized email to ga4 events

leogermani:
### All Submissions:

Adds an anonymized email para to GA4 events.

1. Set up a site with GA4 back end events (Newspack > Analytics > Newspack Custom Events)
2. Fire a back-end event by registering via the registration block, making a donation or subscribing to a newsletter
3. In GA, confirm you see the event coming through and that it has the new `user_hash` param

### Other information:

### DIFF
--- a/includes/data-events/connectors/ga4/README.md
+++ b/includes/data-events/connectors/ga4/README.md
@@ -26,7 +26,7 @@ These parameters are added to all events:
 * `ga_session_id`: The GA Session ID, retrieved from the cookie
 * `logged_in`: Whether the user is logged in when the event got fired
 * `is_reader`: Whether the user is a RAS reader
-* `user_hash`: The anonymized user email, if the user is logged in
+* `email_hash`: The anonymized user email, if the user is logged in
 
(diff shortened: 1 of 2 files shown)


miguelpeixe:
Maybe `email_hash` is more accurate? I'd assume a `user_hash` wouldn't change, but in theory, a user could change their email (even if we don't support email changes now)

miguelpeixe:
@leogermani ping ⬆️

leogermani:
> Maybe `email_hash` is more accurate? I'd assume a `user_hash` wouldn't change, but in theory, a user could change their email (even if we don't support email changes now)

yep, at first I was thinking of using the email in conjunction with the ID, or the site name, or something tat would tell this email appart from the same email used in another site. But then I figured only the email looks ok, wdyt?

miguelpeixe (approved):
Works well!

> email in conjunction with the ID, or the site name, or something tat would tell this email appart from the same email used in another site

Isn't a different site a different GA property?

PR #2595, opened 2023-08-11
fix(campaigns): move try/catch while fetching fav cat names

miguelpeixe:
### All Submissions:

Moves the try/catch so other favorite categories that might be part of the segment render their names if one fails.

1. Checkout https://github.com/Automattic/newspack-plugin/tree/epic/campaigns-rearchitecture, https://github.com/Automattic/newspack-popups/tree/epic/rearchitecture and https://github.com/Automattic/newspack-blocks/tree/epic/campaigns-rearchitecture
2. Create a segment that matches 2 favorite categories
3. Delete one of the categories
4. Visit the "Segments" wizard and confirm the "Favorite Categories: " description shows empty
5. Check out this branch, refresh, and confirm it shows the remaining category

### Other information:

### DIFF
--- a/assets/wizards/popups/utils.js
+++ b/assets/wizards/popups/utils.js
@@ -11,6 +11,7 @@ import { useEffect, useState, Fragment } from '@wordpress/element';
  * External dependencies.
  */
 import memoize from 'lodash/memoize';
+import compact from 'lodash/compact';
 import { format, parse } from 'date-fns';
 
 const allCriteria = window.newspack_popups_wizard_data?.criteria || [];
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Good.

But we should change this and avoid having to do one request per category. Couldn't we just pass all the we need via localize_script? or make the endpoint that returns the segment already include the name?

miguelpeixe:
> Couldn't we just pass all the we need via localize_script? or make the endpoint that returns the segment already include the name?

We can. I tried not to refactor too much of this bit, just the necessary changes to work with the new API. We got a pretty big changeset already 😅

PR #2623, opened 2023-08-31
feat: new option to require all membership plans for restricted content

dkoo:
### All Submissions:

Some publishers want to gate most articles behind a metered reg wall, but gate some smaller subset of those articles (e.g. those with a "Premium" category) behind a hard paywall. Free registrants should get access to most articles, but not to paywalled articles until they purchase the necessary membership. This is a common content gating strategy used by many large-scale publishing sites but which we currently can't support without the proposed change.

We do have the ability to show different paywalls for these two "tiers" of restricted content. This is done by creating two membership plans: Plan 1 restricts all posts with a metered reg wall. Plan 2 restricts Premium articles with a paywall.

However, the way WooCommerce Memberships works, if multiple different membership plans have different rules restricting the same content (as in the above example—Premium articles are a subset of all articles) it will grant access to the restricted content as long as the reader is a member of ANY applicable membership plan. So this defeats the purpose of the paywall in the above scenario: as soon as you register a free account, you get access to all articles, which includes the Premium articles.

This PR introduces a new option in the Reader Activation wizard where we already show other Content Gate settings. This option lets us implement a "stricter" evaluation of membership rules when deciding whether or not to grant access to restricted content. 

Screenshot 2023-08-31 at 2 03 31 PM

The default (disabled) behavior introduces no changes to the current behavior, as some publishers have content gating strategies that require this behavior. When enabled, the behavior changes such that if multiple membership plans are restricting access to the same content, a reader must be a member of ALL plans in order to get access to the content, not just one of them.

Closes `1200550061930446/1205368936341187`.

1. On `master`, set up two membership plans.
  - Plan 1: Should grant access upon "user account registration" and should restrict access to all posts of post type "Post".
  - Plan 2: Should grant access upon product "product(s) purchase", tied to a paid subscription product, and should restrict access to a specific category or categories of posts.

2. Go to **Newspack > Engagement > Reader Activation** and open the advanced settings panel. Edit and publish your primary content gate and insert a Registration wall pattern. Turn on metering and set it to an easily testable value like `2`.
3. From the WooCommerce Memberships sidebar panel in this editor, create or edit the content gate tied to the subscription-based Plan 2 above. Add a paywall pattern tied to the subscription product with no metering and publish.
4. On the front-end, start a new reader session and immediately visit a post in the premium category. Confirm that the post is blocked by the paywall gate right away.
5. Visit 3 other non-premium posts. Confirm that you're able to see the first two, but that the third is blocked with the reg wall gate.
6. Register for a free account from the reg wall and confirm that you gain access as expected.
7. Visit any premium post and observe that you also now have access to these posts, despite not owning the required subscription/membership.
8. Check out this branch and go back to **Newspack > Engagement > Reader Activation**. Confirm you see the new option right under the content gate and that it's off by default (note that you will only see the new option if your site has more than one active membership plan):

Screenshot 2023-08-31 at 2 03 31 PM

9. Leave it off and do a smoke test by repeating steps 4–7. The behavior should be exactly the same as before.
10. Toggle on the new option and save. Refresh the admin page and confirm your setting persists.
11. Repeat steps 4–7 again, but this time confirm that at step 7 after registering for a free account, you have access to non-premium posts while premium posts are still blocked by the paywall gate.
12. Purchase a subscription product via the paywall gate and confirm that you now have access to all premium posts as well.

### Other information:

### DIFF
--- a/assets/wizards/engagement/index.js
+++ b/assets/wizards/engagement/index.js
@@ -76,7 +76,7 @@ class EngagementWizard extends Component {
 	 * Render
 	 */
 	render() {
-		const { pluginRequirements } = this.props;
+		const { pluginRequirements, wizardApiFetch } = this.props;
 		const { relatedPostsEnabled, relatedPostsError, relatedPostsMaxAge, relatedPostsUpdated } =
 			this.state;
 
@@ -114,6 +114,7 @@ class EngagementWizard extends Component {
 		const props = {
 			headerText: __( 'Engagement', 'newspack' ),
 			tabbedNavigation: tabbed_navigation,
+			wizardApiFetch,
 		};
 		return (
 			<Fragment>
--- a/assets/wizards/engagement/views/reader-activation/index.js
+++ b/assets/wizards/engagement/views/reader-activation/index.js
@@ -26,7 +26,7 @@ import ActiveCampaign from '../../components/active-campaign';
 import Mailchimp from '../../components/mailchimp';
 import { HANDOFF_KEY } from '../../../../components/src/consts';
 
-export default withWizardScreen( () => {
+export default withWizardScreen( ( { wizardApiFetch } ) => {
 	const [ inFlight, setInFlight ] = useState( false );
 	const [ config, setConfig ] = useState( {} );
 	const [ membershipsConfig, setMembershipsConfig ] = useState( {} );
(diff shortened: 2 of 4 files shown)


miguelpeixe (approved):
Working great! 🙌

PR #2632, opened 2023-09-08
fix: sync scripts for Stripe-to-Woo migrations

dkoo:
### All Submissions:

Fixes some lingering issues with the Stripe-to-Woo migration script. Also adds a new script that allows us to trigger a resync of contact data to the connected ESP, in case the ESP's synced data has diverged from the source of truth (WooCommerce).

#### Stripe-to-Woo migration script fix

1. On `master`, set your Reader Revenue platform to Stripe and make sure your wp-config.php doesn't have the `NEWSPACK_USE_WC_SUBSCRIPTIONS_WITH_STRIPE_PLATFORM` constant. Also make sure that RAS is enabled and that your connected ESP is either ActiveCampaign or Mailchimp (the ESPs to which we sync RAS data).
2. Make a recurring donation as a new reader. Confirm that the contact is synced to the connected ESP with our usual Newspack metadata. Take special note of the `NP_Membership Status` value: it should be something like "Monthly Donor", indicating your donor status and the recurrence period of the donation.
3. Run the `wp newspack stripe sync-stripe-subscriptions-to-wc` script. This will migrate the Stripe customer and their subscription to WooCommerce and cancel the Stripe subscription.
4. In the connected ESP, observe that the Stripe subscription cancellation has triggered a contact sync: the `NP_Membership Status` field will now say something like "Ex-Monthly donor", which isn't accurate because the subscription wasn't actually cancelled, it was just migrated to WooCommerce. Make note of this contact for testing the resync script, below.
5. Check out this branch and repeat steps 2-3 with a new customer email and subscription. This time, confirm that in the connected ESP, the contact has not been updated because their migrated subscription is still active in WooCommerce.
6. In WooCommerce > Subscriptions, find and edit the migrated subscription. In the Subscription Actions panel, process a manual renewal to trigger another charge on the subscription. In the connected ESP, confirm that the contact's metadata has been updated (e.g. the `NP_Total Paid` value has increased by the amount of the manual renewal), indicating that the WooCommerce subscription is now syncing to this contact in the ESP.

Note: in this test, the manual renewal will trigger a one-time donation transaction for the customer, which will change the `NP_Membership Status` of the contact in the ESP to "Donor" instead of e.g. "Monthly Donor". This won't happen for real automated subscription renewals.

#### Resync script

1. Run the `wp newspack woo resync` CLI script. You can run this script without flags, which will resync all customers in the site, or with the following options:
  - `--migrated-subscriptions`: will only resync customers attached to subscriptions that were migrated from Stripe. This takes precedence over any other args passed.
  - `--subscription-ids`: will resync customers attached to the given subscription IDs (comma-delimited list).
  - `--user-ids`: will resync customers with the given user IDs (comma-delimited list).
  - `--order-ids`: will resync customers attached to the given order IDs (comma-delimited list).

2. Look up the Stripe customer that was accidentally flagged as an "ex-" donor in step 3 above, and confirm that they were resynced with the proper active status (no "ex-" in their `NP_Membership Status` value).
3. Test the script with each flag (and no flags) and confirm that in every case, only the contacts you'd expect are resynced.
4. Confirm that when resyncing, only the contact data and metadata fields are updated, not the ESP list subscription status (so that if a contact was subscribed to any lists, they won't be unsubscribed by the resync, and vice versa).

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-sync.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-sync.php
@@ -118,7 +118,8 @@ Running WooCommerce-to-ESP contact resync...
 
 		if ( ! empty( $subscription_ids ) ) {
 			\WP_CLI::log( __( 'Migrating by subscription ID...', 'newspack' ) );
-			while ( $subscription_ids ) {
+
+			while ( ! empty( $subscription_ids ) ) {
 				$subscription_id = array_shift( $subscription_ids );
 				$subscription    = new \WC_Subscription( $subscription_id );
 
@@ -142,8 +143,6 @@ Running WooCommerce-to-ESP contact resync...
 					$subscription_ids = Stripe_Sync::get_migrated_subscriptions( $batch_size, $batches * $batch_size );
 				}
 			}
-
-			return;
 		}
 
 		// If user-ids flag is passed, resync those users.
@@ -177,7 +176,7 @@ Running WooCommerce-to-ESP contact resync...
 		}
 
 		// Default behavior: resync all customers and subscribers.
-		if ( empty( $user_ids ) && empty( $order_ids ) && empty( $subscription_ids ) ) {
+		if ( empty( $user_ids ) && empty( $order_ids ) && empty( $subscription_ids ) && ! $migrated_only ) {
 			\WP_CLI::log( __( 'Migrating all customers...', 'newspack' ) );
 			$user_ids = self::get_batch_of_customers( $batch_size );
 			$batches  = 0;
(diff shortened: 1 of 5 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-sync.php
```diff
@@ -0,0 +1,306 @@
...
+	 */
+	private static $results = [
+		'processed' => 0,
+	];
+
+		/**
+		 * Initialize.
+		 *
+		 * @codeCoverageIgnore
+		 */
```
claudiulodro, line 33: Minor nit: docblock spacing gets weird here :)

claudiulodro (approved):
Thanks for the speedy turnaround on this! It looks solid and works well in my testing. 

Long-term I do think we should move these CLI commands into Custom Content Migrator, so we don't have to wait so long for them to become available, but that's far outside the scope of this PR. :)

includes/reader-revenue/woocommerce/class-woocommerce-sync.php
dkoo: Whoops, fixed in 44605bc9!

PR #2646, opened 2023-09-15
fix: partial text domain update for the newspack plugin

laurelfulford:
### All Submissions:

The text domain in the Newspack Plugin is incorrect -- it should be `newspack-plugin`, matching the plugin's slug, but it is `newspack`. Because of this, translations don't work. 

This flew under the radar for a long time because the plugin just contained backend strings, but now that it has front-end facing text -- like RAS -- it's a lot more noticeable. 

This PR fixes a handful of publisher-specific strings, to help with a launch and make this PR more reasonable to test. The plan is to handle the rest of the strings in small batches so we don't end up with overwhelming test scenarios. 

See 1200550061930446-1205509524123559

1. Visually review the PR and make sure the only changes that have been made are changing `newspack` to `newspack-plugin` in the translation strings, and greatly reducing the POT file (only strings with the updated text-domain should show up there).
2. Test to make sure translations actually load -- for this, can make your own .PO file and translate some of the strings and make sure they show up on the front-end. I also have a fake PO and MO German translation you can use - you can drop these two files in the /newspack-plugin/languages directory, change your site's language to German, and confirm some of the front-end RAS strings -- like the Sign In button, and Log Out link on My Account -- have 'translated' appended to them.

[fake-german-translation.zip](https://github.com/Automattic/newspack-plugin/files/12621781/fake-german-translation.zip)

### Other information:


dkoo (approved):
Looks good!

laurelfulford:
Thanks @dkoo!

PR #2658, opened 2023-09-22
feat(campaigns): newsletter subscription list criteria

miguelpeixe:
### All Submissions:

Changes required for https://github.com/Automattic/newspack-popups/pull/1218:

 - Implements a new `newsletter_subscribed_lists` store data item
 - Adds the necessary changes to the campaigns' wizard to support the new criteria
 - Creates a `SubscriptionListsControl` component to select newsletter subscription lists from the API using the `AutocompleteTokenField` component

1. Follow instructions at https://github.com/Automattic/newspack-popups/pull/1218
2. Make sure the `newsletter_subscribed_lists` data item updates on different occasions:
    1. subscribe a fresh account and confirm the selected lists are stored
    2. unsubscribe to a list from "My Account -> Newsletters" and confirm it updates the item
    3. subscribe to another list from "My Account -> Newsletters" and confirm it updates the item

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-data.php
+++ b/includes/reader-activation/class-reader-data.php
@@ -277,6 +277,7 @@ final class Reader_Data {
 	 */
 	public static function update_newsletter_subscribed_lists( $timestamp, $data ) {
 		if ( ! empty( $data['lists'] ) ) {
+			self::update_item( $data['user_id'], 'is_newsletter_subscriber', true );
 			self::update_item( $data['user_id'], 'newsletter_subscribed_lists', wp_json_encode( $data['lists'] ) );
 		}
 		if ( ! empty( $data['lists_added'] ) ) {
@@ -289,6 +290,7 @@ final class Reader_Data {
 			} else {
 				$newsletter_subscribed_lists = $data['lists_added'];
 			}
+			self::update_item( $data['user_id'], 'is_newsletter_subscriber', true );
 			self::update_item( $data['user_id'], 'newsletter_subscribed_lists', wp_json_encode( $newsletter_subscribed_lists ) );
 		}
 		if ( ! empty( $data['lists_removed'] ) ) {
@@ -299,10 +301,9 @@ final class Reader_Data {
 			if ( ! empty( $newsletter_subscribed_lists ) && is_array( $newsletter_subscribed_lists ) ) {
 				$newsletter_subscribed_lists = array_diff( $newsletter_subscribed_lists, $data['lists_removed'] );
 			}
+			self::update_item( $data['user_id'], 'is_newsletter_subscriber', ! empty( $newsletter_subscribed_lists ) );
 			self::update_item( $data['user_id'], 'newsletter_subscribed_lists', wp_json_encode( $newsletter_subscribed_lists ) );
 		}
-		// Deprecate 'is_newsletter_subscriber' item.
-		self::delete_item( $data['user_id'], 'is_newsletter_subscriber' );
 	}
 
 	/**
@@ -386,6 +387,7 @@ final class Reader_Data {
 		}
 		$subscribed_lists = \Newspack_Newsletters_Subscription::get_contact_lists( $data['email'] );
 		if ( ! is_wp_error( $subscribed_lists ) && is_array( $subscribed_lists ) ) {
+			self::update_item( $data['user_id'], 'is_newsletter_subscriber', ! empty( $subscribed_lists ) );
 			self::update_item( $data['user_id'], 'newsletter_subscribed_lists', wp_json_encode( $subscribed_lists ) );
 		}
 	}
--- a/includes/reader-activation/class-reader-data.php
+++ b/includes/reader-activation/class-reader-data.php
@@ -25,13 +25,6 @@ final class Reader_Data {
 	private static $reader_activity = [];
 
 	/**
-	 * List of deleted items to send to the front-end.
-	 *
-	 * @var string[]
-	 */
-	private static $deleted_items = [];
-
-	/**
 	 * Initialize hooks.
 	 */
 	public static function init() {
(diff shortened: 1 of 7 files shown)


includes/reader-activation/class-reader-data.php
```diff
@@ -348,17 +373,14 @@ public static function check_newsletter_subscription( $timestamp, $data ) {
 		if ( ! empty( $is_newsletter_subscriber ) && gettype( $is_newsletter_subscriber ) === 'string' ) {
 			$is_newsletter_subscriber = json_decode( $is_newsletter_subscriber );
 		}
-		// Bail if reader is already a newsletter subscriber.
-		if ( $is_newsletter_subscriber ) {
-			return;
-		}
 		// Check if the user is subscribed to a newsletter.
```
adekbadek, line 376: This comment should be removed or updated :)

includes/reader-activation/class-reader-data.php
```diff
@@ -348,17 +373,14 @@ public static function check_newsletter_subscription( $timestamp, $data ) {
 		if ( ! empty( $is_newsletter_subscriber ) && gettype( $is_newsletter_subscriber ) === 'string' ) {
 			$is_newsletter_subscriber = json_decode( $is_newsletter_subscriber );
 		}
-		// Bail if reader is already a newsletter subscriber.
-		if ( $is_newsletter_subscriber ) {
```
adekbadek, line 352: If `$is_newsletter_subscriber` is not used anymore, lines above defining & mutating it can also be removed.

includes/reader-activation/class-reader-data.php
```diff
@@ -348,17 +373,14 @@ public static function check_newsletter_subscription( $timestamp, $data ) {
...
-			return;
-		}
 		// Check if the user is subscribed to a newsletter.
 		if ( ! class_exists( '\Newspack_Newsletters' ) || ! class_exists( '\Newspack_Newsletters_Subscription' ) ) {
 			return;
 		}
 		$subscribed_lists = \Newspack_Newsletters_Subscription::get_contact_lists( $data['email'] );
 		if ( ! is_wp_error( $subscribed_lists ) && ! empty( $subscribed_lists ) && is_array( $subscribed_lists ) ) {
```
adekbadek, line 381: Shouldn't there be an else here, where `is_newsletter_subscriber` would be set to  `false`? After I've unsubscribed from all lists, this value (as seen in `localStorage`) was not updated: `newsletter_subscribed_lists` was `[]`, but `is_newsletter_subscriber` was `true`.

adekbadek (changes requested):

includes/reader-activation/class-reader-data.php
```diff
@@ -265,7 +265,32 @@ public static function api_delete_item( $request ) {
 	 * @param array $data      Data.
 	 */
 	public static function set_is_newsletter_subscriber( $timestamp, $data ) {
-		self::update_item( $data['user_id'] ?? \get_current_user_id(), 'is_newsletter_subscriber', true );
+		self::update_item( $data['user_id'], 'is_newsletter_subscriber', true );
```
adekbadek, line 268: I still encounter the issue described here: https://github.com/Automattic/newspack-plugin/pull/2658#discussion_r1338714577 – after unsubscribing from all lists on the My Account page, the value of `*_is_newsletter_subscriber` in `localStorage` is `"true"`. This is because the value is set to true here, regardless of the list update outcome. If the user ends up with no lists they are subscribed to, `'is_newsletter_subscriber'` should be updated. 

A sidenote – `is_newsletter_subscriber` and `newsletter_subscribed_lists` seem redundant. The former can be surmised from the latter.

adekbadek (changes requested):

includes/reader-activation/class-reader-data.php
miguelpeixe: I like the idea of deprecating the bool in favor of checking against the list. Implemented in 2fdde856087dd8e15a2d03f683e5ced81ea5beb6 and https://github.com/Automattic/newspack-popups/pull/1218/commits/ef849734e9436c1143d5aa2d8ff5832ffc435856

miguelpeixe: Even though the `delete_item` deletes it from the database, the localStorage item will remain as we don't have an isomorphic deprecation strategy. I wonder if we should 🤔

miguelpeixe: Decided to add support for deletion from server -> local in eefdbb99c948852579bd2a65600f815f0eef45cd

Any change in the subscription list selection should trigger the deletion of the bool item from the server, which also deletes from localStorage.

The deprecation will not occur on the `reader_logged_in` handler because that's async. It won't send the deleted item key back to the browser.

includes/reader-activation/class-reader-data.php
```diff
@@ -264,8 +275,34 @@ public static function api_delete_item( $request ) {
 	 * @param int   $timestamp Timestamp.
 	 * @param array $data      Data.
 	 */
-	public static function set_is_newsletter_subscriber( $timestamp, $data ) {
```
adekbadek, line 267: After this method is removed, [this line](https://github.com/Automattic/newspack-popups/blob/ada4246e7677ecb03213b40a32a4da8942f6a9be/includes/class-newspack-popups-segmentation.php#L75) should be also removed (or updated to check for different method).

includes/reader-activation/class-reader-data.php
adekbadek: I still see references to `is_newsletter_subscriber` in a few places ([1](https://github.com/Automattic/newspack-popups/blob/5b794adc26f808fca59ddc568cebda71b4440aa0/includes/class-newspack-segments-migration.php#L395), [2](https://github.com/Automattic/newspack-popups/blob/5b794adc26f808fca59ddc568cebda71b4440aa0/includes/class-newspack-segments-migration.php#L401), [3](https://github.com/Automattic/newspack-popups/blob/ada4246e7677ecb03213b40a32a4da8942f6a9be/src/view/README.md#L17)). Shouldn't these be updated?

adekbadek (commented):
Works as described, but the changes related to item deletion are giving me a pause because it's not clear to me how to test them. How can I verify that this code works?

miguelpeixe:
Since they were not tested in the previous review and they are not going to be used now, it's better to just not have them now. Removed in e1f2b4d9c87c4e8104d87518a3884b82f20cbcd4

The idea is to remove a localStorage item when it was removed in the backend.

adekbadek (approved):

PR #2663, opened 2023-09-25
feat(data-events): activecampaign connector

miguelpeixe:
### All Submissions:

Closes # .

1.
2.
3.

### Other information:


includes/plugins/class-newspack-newsletters.php
```diff
@@ -119,179 +117,6 @@ public static function get_metadata_key( $key ) {
...
-								$lists_names = [];
-								foreach ( $selected_list_ids as $selected_list_id ) {
-									foreach ( $lists as $list ) {
-										if ( $list['id'] === $selected_list_id ) {
-											$lists_names[] = $list['name'];
-										}
-									}
-								}
-								// Note: this field will be overwritten every time it's updated.
-								$metadata[ self::get_metadata_key( 'newsletter_selection' ) ] = implode( ', ', $lists_names );
```
adekbadek, line 181: I assume this metadata is deprecated? It makes sense because it duplicates what's already available in AC, but I wanted to make sure.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -342,53 +322,6 @@ public static function sync_reader_from_order( $order, $verify_created_via = tru
...
-	/**
-	 * If the site is using woocommerce-memberships, create a new user and a
-	 * membership, if the donation type calls for it.
-	 *
-	 * @param string $email_address Email address.
-	 * @param string $full_name Full name.
-	 * @param string $frequency Donation frequency.
-	 * @param array  $metadata Donor metadata.
-	 */
-	public static function set_up_membership( $email_address, $full_name, $frequency, $metadata = [] ) {
```
adekbadek, line 354: This method [is used in Blocks](https://github.com/Automattic/newspack-blocks/blob/31da96848de290f1db08f4a2940b60b4453fd25d/src/blocks/donate/class-wp-rest-newspack-donate-controller.php#L163-L170), why is it removed here?

adekbadek (changes requested):
One blocking question about code removed. Otherwise works as described, nice to see the improved simplicity!

includes/plugins/class-newspack-newsletters.php
miguelpeixe: Thank you for catching that. Although I agree it's duplicated data, I don't think we should deprecate it without a broader conversation.

Implemented support for the key in 42ea81210b9f891bd2fc1349a09e9e3645682a10

You should be able to see the `Newsletter Selection` key updating in the contact on the `newsletter_updated` data event. This happens when subscribing or changing the newsletter selection in "My Account -> Newsletters".

adekbadek: > This happens when subscribing or changing the newsletter selection in "My Account -> Newsletters".

It should also happen whenever someone subscribes. When testing, I subscribe via the Newsletter Subscribe block, and the metadata is not updated then.

adekbadek (changes requested):
(see [this comment](https://github.com/Automattic/newspack-plugin/pull/2663#discussion_r1361845886))

includes/plugins/class-newspack-newsletters.php
miguelpeixe: Added in 26f93d2f83b04168680a6b75645a996e34e2520e

PR #2673, opened 2023-09-29
fix(woocommerce): check method exists before syncing reader

miguelpeixe:
### All Submissions:

Check if the Newspack Newsletters method is available before syncing the reader on WooCommerce orders.

1. While on the master branch, make sure you have RAS configured and deactivate the Newspack Newsletters plugin
2. Donate using the donate block and confirm you get a fatal error
3. Checkout this branch, donate again, and confirm it works without errors
4. Activate Newspack Newsletters and donate again
5. Confirm the contact syncs to the ESP

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -336,7 +336,9 @@ class WooCommerce_Connection {
 		if ( ! $contact ) {
 			return;
 		}
-
+		if ( ! method_exists( 'Newspack_Newsletters_Subscription', 'add_contact' ) ) {
+			return;
+		}
 		return \Newspack_Newsletters_Subscription::add_contact( $contact );
 	}
 


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -336,7 +336,9 @@ public static function sync_reader_from_order( $order, $verify_created_via = tru
 		if ( ! $contact ) {
 			return;
 		}
-
+		if ( ! method_exists( 'Newspack_Newsletters_Subscription', 'add_contact' ) ) {
```
dkoo, line 339: Does this need a `\` prefix since the `Newspack_Newsletters_Subscription` class is in a different namespace than this class?

dkoo (changes requested):
Question below!

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
miguelpeixe: Not really, when passing it as a string argument it doesn't inherit the file's namespace. It'll always require the fully-qualified class name, which in the root doesn't call for `\`.

dkoo: TIL! Thanks for explaining.

dkoo (approved):

PR #2695, opened 2023-10-12
Modal checkout: allow covering Stripe's fees 

adekbadek:
### All Submissions:

Adds the option for the donor to cover Stripe's transaction fees:

image

See 1200133283036252-as-1205668937699532

The feature is on by default. To opt-out, a change the "Allow donors to cover transaction fees" setting in Reader Revenue wizard, Stripe Gateway tab.

There's one non-covered edge case – use of coupons. The logic to handle covering fees with coupons applied would make this code overly complex, and since we're talking about donations here, usage of coupons seems like an extreme edge case. If coupons are applied to the order, the feature will disappear gracefully. 

1. Make sure you have "Newspack" set as the Reader Revenue platform and `Stripe (Credit Card)` enabled as a payment gateway
2. Visit a page with a donation prompt, and initiate the flow to trigger the modal checkout
3. Observe a "Cover transaction fees?" checkbox above the card details inputs in the "Payment info" section of the checkout
4. Complete transactions with and without opting to cover the fees, observe the WC orders in the latter case:
    - have the total order amount increased to offset the Stripe fee
    - have a `newspack_donor_covers_fees` meta field with value of `1`
    - have an order note attached commenting about the fact of covering fees
5. In WC settings, turn on coupons and create a coupon. Use the coupon in the checkout – observe that the feature is not active when the coupon is applied
6. Visit Newspack's Reader Revenue wizard, Stripe Gateway tab. Update the fee calculation inputs and observe these are reflected on the front-end.

### Other information:


adekbadek:
I don't think that's fair to the donor – they start the flow with an amount and then are offered to increase it slightly. Increasing it by default does not seem user-friendly, especially if someone misses the checkbox when looking at the UI.

dkoo:
Agree with @adekbadek—this is an upsell, and it feels like a dark pattern to not make it opt-in-only. I'm not a legal expert, but I also suspect it might run afoul of consumer regulations in places that have them.

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -0,0 +1,233 @@
...
+		}
+		if ( 0 < count( WC()->cart->get_coupon_discount_totals() ) ) {
+			// If the checkout has coupons applied, bail. This can be develped in the future,
+			// but at this point handling coupons + covering fees is an edge case.
+			return false;
+		}
+		if ( defined( 'NEWSPACK_DISABLE_ALLOW_COVERING_FEES' ) && NEWSPACK_DISABLE_ALLOW_COVERING_FEES ) {
+			return false;
+		}
+		if ( isset( $_REQUEST['modal_checkout'] ) && 1 === intval( $_REQUEST['modal_checkout'] ) ) {  // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
miguelpeixe, line 98: Non-blocking: this can change to the new `Newspack_Blocks\Modal_Checkout::is_modal_checkout()` method from https://github.com/Automattic/newspack-blocks/pull/1573

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	/**
+	 * Should this feature be active?
+	 *
+	 * Some of this code is taken from Newspack Blocks, but can't be reused directly because
+	 * WC renders the checkout form before Blocks' code is available.
+	 */
+	private static function should_allow_covering_fees() {
```
miguelpeixe, line 86: The PR describes covering Stripe fees for **donors**, but the option is added to any modal checkout. I believe this should be restricted to donation products, right?

I don't think it's desirable to have this option available when purchasing a membership subscription.

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -0,0 +1,233 @@
...
+	 */
+	private static function get_stripe_fee_static_value() {
+		return floatval( get_option( 'newspack_blocks_donate_fee_static', '0.3' ) );
+	}
+
+	/**
+	 * Get the fee human-redable value.
+	 */
+	private static function get_fee_human_readable_value() {
+		$fee_static = sprintf( get_woocommerce_price_format(), get_woocommerce_currency_symbol(), self::get_stripe_fee_static_value() );
```
miguelpeixe, line 215: How about using `wc_price()` so it uses all WC logic for price rendering? Using it will render $0.50 instead of $0.5 and theoretically handle any other exceptions a different currency may have.

miguelpeixe (changes requested):
Toggling the checkbox is working well but if I leave the option on and switch payment methods I still get the Stripe fee as part of the total:

https://github.com/Automattic/newspack-plugin/assets/820752/3e7490f3-1df7-4776-ba60-6d44f8b601d8

miguelpeixe:
Also, why not make the cover fee optional through the Reader Revenue -> Stripe Gateway form instead of a feature flag?

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
adekbadek: Good idea, this is gonna necessitate [a small change in Blocks](https://github.com/Automattic/newspack-blocks/pull/1577).

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
adekbadek: Right, fixed in 8c2f42972780c78ce97c31de2732b7412290710e

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
adekbadek: Good idea, done in 8c2f42972780c78ce97c31de2732b7412290710e

adekbadek:
> Toggling the checkbox is working well but if I leave the option on and switch payment methods I still get the Stripe fee as part of the total:

Done in 8c2f42972780c78ce97c31de2732b7412290710e

> Also, why not make the cover fee optional through the Reader Revenue -> Stripe Gateway form instead of a feature flag?

Good call, done in 867afb71b917f89b819daf1ed33bc6925a91918b

includes/configuration_managers/class-woocommerce-configuration-manager.php
```diff
@@ -114,12 +114,13 @@ public function stripe_data() {
...
-			'secretKey'          => $stripe->get_option( 'secret_key', '' ),
-			'testPublishableKey' => $stripe->get_option( 'test_publishable_key', '' ),
-			'testSecretKey'      => $stripe->get_option( 'test_secret_key', '' ),
+			'enabled'             => 'yes' === $stripe->get_option( 'enabled', false ) ? true : false,
+			'testMode'            => 'yes' === $stripe->get_option( 'testmode', false ) ? true : false,
+			'publishableKey'      => $stripe->get_option( 'publishable_key', '' ),
+			'secretKey'           => $stripe->get_option( 'secret_key', '' ),
+			'testPublishableKey'  => $stripe->get_option( 'test_publishable_key', '' ),
+			'testSecretKey'       => $stripe->get_option( 'test_secret_key', '' ),
+			'allow_covering_fees' => get_option( 'newspack_donations_allow_covering_fees', true ),
```
miguelpeixe, line 123: Non-blocking: should we care about case consistency here (lower camel vs. snake)?

miguelpeixe (approved):
Thank you for the revisions! 🙌

includes/configuration_managers/class-woocommerce-configuration-manager.php
adekbadek: I think we should, but I'd opt for using snake case rather than camel case. Something for a later PR I think.

PR #2696, opened 2023-10-13
feat: segmentation criteria for subscriptions and memberships

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-popups/pull/1230, adds support and UI for two new sets of criteria, based on subscriptions and memberships.

Also splits the UI for the single segment screen into smaller subsections, as the current set of sections was looking crowded as we add more criteria. (cc @thomasguillot)

Note that I moved some files around and renamed some others for consistency. 

- The `ListsControl` component is only used in the context of the Campaigns wizard, so I moved it out of the Newspack Components directory and into the components for the wizard. Also made this more abstract so it could be reused for all of the `list__in` and `list__not_in` criteria.
- Renamed some of the components files in the Campaigns wizard folder from `CamelCase` to `kebab-case`, which matches most of our file naming across repos. Didn't try to be exhaustive about this.
- Made some minor style and markup updates to the `AutocompleteTokenfield` component, to make the placement of the spinner component during loading more consistent and predictable.

Closes `1205451924709229/1205603536115944`.

1. On `master`, register two new user accounts in separate sessions. On Account A, purchase a non-donation subscription product. On Account B, join any membership plan by satisfying the plan's requirements for granting access.
2. Check out this branch and https://github.com/Automattic/newspack-popups/pull/1230. Create two new segments at the highest priority, one for Subscriptions with "Has active subscription(s)" targeting the subscription product purchased by Account A, and one for Memberships with "Has active membership(s)" targeting the membership plan joined by Account B. Also note the new organization of the single segment screen here, which splits the "Reader Activity" section into three separate sections:

Screenshot 2023-10-13 at 4 51 46 PM

3. Test support for segmenting user accounts who already have active subscriptions/memberships upon login. In separate sessions, log in as Account A and Account B and confirm that after logging in, each session gets segmented to the corresponding subscription/membership segment.
4. Test support for new subscriptions/memberships: In two new sessions, purchase the required subscription product and join the required membership plan. In each case confirm that you're segmented after completing the purchase/membership action.
5. Test support for changes subscription/membership statuses. As an admin, cancel the subscription/membership for the test users. Log in as those test users and confirm that you no longer match the segments.
6. Also test the new "Does not have active subscription/membership" criteria: the test users who have active subscriptions and memberships matching these criteria should not match these segments, while everyone else should.

### Other information:


assets/wizards/popups/components/lists-control/index.js
```diff
@@ -0,0 +1,45 @@
...
+				const lists = await apiFetch( {
+					path,
+				} );
+				const values = Array.isArray( lists ) ? lists : Object.values( lists );
+				return values
+					.filter( item => ids.includes( item.id ) )
+					.map( item => ( {
+						value: isNaN( parseInt( item.id ) ) ? item.id.toString() : parseInt( item.id ),
+						label: item.title || item.name,
+					} ) );
```
adekbadek, line 40: Nit: the function passed to `.map` could be named and reused here and in the other prop. This would be more readable.

includes/data-events/listeners.php
```diff
@@ -301,3 +302,123 @@ function( $subscription, $status_to, $status_from ) {
...
+
+/**
+ * When a non-donation subscription is activated.
+ * The subscription will be added to the user's list of active subscriptions.
+ */
+Data_Events::register_listener(
+	'woocommerce_subscription_status_active',
+	'product_subscription_active',
+	function( $subscription ) {
+		// We only want to fire this for non-donation products.
```
adekbadek, line 314: If so, shouldn't the function return early if the subscription is for a donation product? I see it's handled later in `includes/reader-activation/class-reader-data.php`, but the comment gave me a pause.

includes/data-events/listeners.php
```diff
@@ -301,3 +302,123 @@ function( $subscription, $status_to, $status_from ) {
...
+Data_Events::register_listener(
+	'woocommerce_subscription_status_updated',
+	'product_subscription_inactive',
+	function( $subscription, $status_to, $status_from ) {
+		// We only want to fire this for non-active statuses.
+		if ( 'active' === $status_to ) {
+			return;
+		}
+
+		// We only want to fire this for non-donation products.
```
adekbadek, line 349: Same as in the `woocommerce_subscription_status_active` listener – shouldn't the function return early if `$product_ids` is empty?

adekbadek (approved):
Left some non-blocking comments.

includes/data-events/listeners.php
dkoo: Good call, 6b26f574 returns early without firing the data event if there are no non-subscription products associated with the subscription.

assets/wizards/popups/components/lists-control/index.js
dkoo: Good idea, added in 6b26f574

includes/data-events/listeners.php
dkoo: Good call, 6b26f574 returns early without firing the data event if there are no non-subscription products associated with the subscription.

dkoo:
Also in 6b26f57, I fixed (hopefully for the last time) a style error in the positioning of the loading spinner for the `AutocompleteTokenfield` component. It should now always be overlaid on the autocomplete input field, whether or not there's a label or help description, and whether or not suggestions are being displayed.

adekbadek (approved):

adekbadek (approved):

PR #2725, opened 2023-11-01
feat(stripe-migration): support mapping of products by SKU

dkoo:
### All Submissions:

Adds a method to migrate a subscription from Stripe to Woo using a specific Woo product by SKU.

1. Check out this branch.
2. In WP admin, go to Products and create a new subscription-type product. Set the SKU to some string, e.g. `TEST-PRODUCT-SKU`.
3. SSH into your test site and run `wp shell`.
4. Create some fake order data. You can use this to test:

```php
$order_data = [ 'status' => 'completed', 'subscription_status' => 'created', 'wc_subscription_status' => 'active', 'email' => 'test@email.com', 'name' => 'Test Customer Name', 'date' => 1698796940, 'amount' => 100, 'frequency' => 'year', 'currency' => 'USD', 'client_id' => '', 'user_id' => '' ];
```

5. Create a transaction using the SKU you created in step 2:

```php
\Newspack\WooCommerce_Connection::create_transaction( $order_data, 'TEST-PRODUCT-SKU' );
```

6. Confirm that a new subscription and parent order were created, mapped to the product with matching SKU instead of to a Newspack donation product.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -99,7 +99,7 @@ class WooCommerce_Connection {
 	}
 
 	/**
-	 * Get WC's Order Item related to a given SKU.
+	 * Get a WC_Order_Item related to the product with the given SKU, or a Newspack donation product based on frequency.
 	 * 
 	 * @param string      $frequency Frequency of the order's recurrence.
 	 * @param number      $amount Donation amount.
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -119,6 +119,28 @@ class WooCommerce_Connection {
 	}
 
 	/**
+	 * Get WC's Order Item related to a given SKU.
+	 * 
+	 * @param string $product_sku Product's SKU string.
+	 * @param number $amount Donation amount.
+	 * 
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -118,6 +118,28 @@ private static function get_donation_order_item( $frequency, $amount = 0 ) {
...
 
+	/**
+	 * Get WC's Order Item related to a given SKU.
+	 * 
+	 * @param string $product_sku Product's SKU string.
+	 * @param number $amount Donation amount.
+	 * 
+	 * @return boolean|WC_Order_Item_Product Order item product, or false if there's no product with matching SKU.
+	 */
+	public static function get_order_item_by_sku( $product_sku, $amount = 0 ) {
```
adekbadek, line 129: 1. A method should be private by default, since it's not used outside of this class. This helps in future refactoring.
2. `get_order_item_by_sku` and `get_donation_order_item` are almost identical – could be refactored to a function that takes a product ID and amount and creates the item.

adekbadek (approved):
Works as described, left some non-blocking suggestions.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
dkoo: Good call! The method is public because I have a migration script (not in this repo) that needs to call it directly. But I'll look into refactoring these two methods into a single, more abstract one.

claudiulodro (approved):
This works well in my testing

PR #2730, opened 2023-11-02
Filter for Stripe connection

claudiulodro:
### All Submissions:

This PR adds a small filter for changing the active Stripe key. This is very useful for subscription migrations:
- It allows us to run scripts from our local environment that use a publisher's Stripe key without having to connect our local environment to a publisher's Stripe.
- It allows us to run scripts using a live Stripe key while a staging site has Stripe test mode on.

1. Create a new script:
```
add_filter( 'newspack_stripe_connection_secret_key', function( $secret_key ) {
    return 'sk_live...'; // Use some different Stripe key than what you have "officially" set up on the site.
}

$stripe = Stripe_Connection::get_stripe_client();
$response = $stripe->customers->all( [ 'limit' => 5 ] );
var_dump( $response );
```
2. Run script with `wp eval-file`. Observe that the customers are from the Stripe account in the filter, not the connected Stripe account.

### Other information:

### DIFF
--- a/includes/reader-revenue/stripe/class-stripe-connection.php
+++ b/includes/reader-revenue/stripe/class-stripe-connection.php
@@ -554,7 +554,7 @@ class Stripe_Connection {
 	 */
 	private static function get_stripe_secret_key() {
 		$stripe_data = self::get_saved_stripe_data();
-		return $stripe_data['usedSecretKey'];
+		return apply_filters( 'newspack_stripe_connection_secret_key', $stripe_data['usedSecretKey'] );
 	}
 
 	/**


miguelpeixe (changes requested):
Functionally works, but I worry that this filter can be misleading. The Stripe connection is [fetched](https://github.com/Automattic/newspack-plugin/blob/08f95691e7facda5555f8003b19dbeb66fb2e276/includes/configuration_managers/class-woocommerce-configuration-manager.php#L104-L125) from the [gateway plugin](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/de88153dbd36350ec50b28fc14b1805cf615b168/includes/class-wc-stripe-api.php#L35-L51) and modifying the result of `get_stripe_secret_key()` does not impact Woo's connection. Having 2 different stored values for the connection sounds dangerous, WDYT?

Alternatively, the same could be achieved by filtering the gateway's option field:

```php
add_filter(
	'option_woocommerce_stripe_settings',
	function( $value ) {
		$value['secret_key'] = 'my-secret-key';
		return $value;
	}
);

claudiulodro:
That's definitely a better solution, @miguelpeixe! And it doesn't require us to maintain a filter forever.

PR #2742, opened 2023-11-07
fix(password-reset): redirect non-readers to WP login

miguelpeixe:
### All Submissions:

The default password reset flow with WooCommerce redirects the user to the `my-account` page. Because we replace the authentication form on that page with a flow that is restricted to reader accounts only, non-readers get stuck in a form they can't use after a password reset.

This PR redirects non-readers to the regular `wp-login.php` page after a password reset.

1. Check out this branch and go through password reset for an admin account
2. Reset the password and confirm you're redirected to `wp-login.php`
3. Do the same but with a reader account
4. Reset the password and confirm you're redirect to `my-account`

### Other information:


dkoo (approved):
Works as described!

PR #2745, opened 2023-11-10
feat: new options for Stripe fees

dkoo:
### All Submissions:

Adds two new configuration options for the "Stripe fees" option introduced in https://github.com/Automattic/newspack-plugin/pull/2695:

- Make the checkbox checked by default (opt-out instead of opt-in): This is configurable by publisher because some sites want it to be opt-out, but others must follow regulations to make such options opt-in.
- Custom fee message. Also tweaks the default message to the slightly more reader-friendly:

> I’d like to cover the [fee percentage] transaction fee to ensure my full donation goes towards [site name]'s mission.

Also integrates the flat percentage fee display from #2746 for better readability in the default messaging.

Closes `1205451924709234/1205912365838338`.

1. Check out this branch.
2. Visit Newspack > Reader Revenue > Stripe Gateway (make sure platform is "Newspack")
3. Confirm the new options here (both should be empty/unchecked by default):

Screenshot 2023-11-10 at 4 30 00 PM

4. Without changing any settings, smoke test a donation and confirm that the checkout modal shows defaults:

Screenshot 2023-11-10 at 4 51 53 PM

5. Update the settings and confirm with a new donation that they take effect (the checkbox should be checked by default, and the custom message should appear below the label):

Screenshot 2023-11-10 at 4 53 07 PM

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
@@ -138,21 +138,38 @@ class WooCommerce_Cover_Fees {
 					<?php endif; ?>
 				>
 				<label for=<?php echo esc_attr( self::CUSTOM_FIELD_NAME ); ?> style="display:inline;">
-					<b><?php echo esc_html( __( 'Cover transaction fees?', 'newspack-plugin' ) ); ?></b><br/>
+					<b><?php echo esc_html( __( 'Cover transaction fee?', 'newspack-plugin' ) ); ?></b><br/>
 					<?php
+					$custom_message = get_option( 'newspack_donations_allow_covering_fees_label', '' );
+					if ( ! empty( $custom_message ) ) {
+						echo esc_html( $custom_message );
+					} else {
 						printf(
-							// Translators: %s is the transaction fee, as percentage with static portion (e.g. 2% + $0.3), %s is the site title.
-							esc_html__( 'Cover Stripe’s %1$s transaction fee, so that %2$s receives 100%% of your payment.', 'newspack-plugin' ),
-							self::get_fee_display_value(), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-							esc_html( get_option( 'blogname' ) )
+							// Translators: %s is the possessive form of the site name.
+							esc_html__(
+								'I’d like to cover the %1$s transaction fee to ensure my full donation goes towards %2$s mission.',
+								'newspack-plugin' 
+							),
+							esc_html( self::get_fee_display_value() ),
+							esc_html( self::get_possessive( get_option( 'blogname' ) ) )
 						);
+					}
 					?>
-
 				</label>
 			</p>
 		<?php
 		$desc .= ob_get_clean();
 		return $desc;
+	}
+
+	/**
+	 * Get possessive form of the given string. Proper nouns ending in S should not have a trailing S.
+	 * 
+	 * @param string $string String to modify.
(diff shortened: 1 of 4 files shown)


includes/wizards/class-reader-revenue-wizard.php
```diff
@@ -204,7 +210,7 @@ public function register_api_endpoints() {
 		// Update Donations settings.
 		register_rest_route(
 			NEWSPACK_API_NAMESPACE,
-			'/wizard/' . $this->slug . '/donations/',
+			' / wizard / ' . $this->slug . ' / donations / ',
```
adekbadek, line 213: This does not seem correct 🤔

dkoo: Weird, not sure where that came from—probably a PHPCS misconfiguration in my IDE?

dkoo: Fixed in 773c6d9

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -132,25 +133,44 @@ public static function add_input_to_stripe_gateway_description( $desc ) {
...
-							esc_html__( 'Cover Stripe’s %1$s transaction fee, so that %2$s receives 100%% of your payment.', 'newspack-plugin' ),
-							self::get_fee_human_readable_value(), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-							esc_html( get_option( 'blogname' ) )
+							// Translators: %s is the possessive form of the site name.
+							esc_html__(
+								'I’d like to cover the %1$s transaction fee to ensure my full donation goes towards %2$s mission.',
+								'newspack-plugin' 
+							),
+							esc_html( self::get_fee_display_value() ),
+							esc_html( self::get_possessive( get_option( 'blogname' ) ) )
```
miguelpeixe, line 153: Non-blocking: we're hardcoding English grammar here. We should change this eventually.

miguelpeixe (approved):
Working as described! 🙌

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
dkoo: Yes, I wasn't sure the best way to handle this. Wouldn't the placement of this as a placeholder in the `esc_html__` function make it possible for sites in other languages to use their own solution?

miguelpeixe: It's added as an interpolated variable to the string, so it would output "Newspack's" regardless of the localized string.

dkoo: How should we solve this? I can't think of a solution to get around the silly English grammar rule.

miguelpeixe: [This person](https://mattermost.com/blog/avoiding-common-internationalization-mistakes/) says `’s` is still correct with names ended in s, "according to other popular grammar guidelines". Looks weird to me, though.

It's not the prettiest, but we could make the transformation conditioned to `get_locale()`

dkoo: It is technically correct in English either way, but since it's related to the site's brand name I can see it being an issue if we allow `'s` with a site name ending in S. Let's leave it for now since the message itself is customizable.

PR #2753, opened 2023-11-15
Feat/add media partners

adekbadek:
### All Submissions:

Adds the https://github.com/Automattic/newspack-media-partners plugin as an optional module. Similar to the RSS Enhancements plugin story (https://github.com/Automattic/newspack-plugin/pull/1688), the module will replace the plugin if the plugin is active. 

This also adds few new features to the Media Partners taxonomy (see 1200133283036252-as-1205658718219681):

1.  Ability to remove the media partner logo
2. Ability to modify the "This story also appeared in" text
3. Ability to skip the media partner markup in RSS feeds (a global setting) 

Note: https://github.com/Automattic/newspack-media-partners should be archived after this is released

1. Install and activate the [`newspack-media-partners` plugin](https://github.com/Automattic/newspack-media-partners) 
5. Load the admin page, go to Posts -> Media Partners 
6. Observe that the plugin has been deactivated
7. Create a media partner with an image and assign it to a post (it's a taxonomy)
8. Visit the post – observe the attribution message close to the beginning of the post 
9. Edit the media partner, remove the image and change the attribution message 
10. Verify this is updated in the post
11. Visit your site's `/feed` and observe the media partner markup is there
12. Toggle the "Skip in RSS feeds" setting, observe that now the markup does not appear in the feed

### Other information:

### DIFF
--- a/includes/optional-modules/class-media-partners.php
+++ b/includes/optional-modules/class-media-partners.php
@@ -22,16 +22,11 @@ class Media_Partners {
 			include_once ABSPATH . 'wp-admin/includes/plugin.php';
 		}
 
-		// If the standalone plugin is active, deactivate it and activate as a module.
-		if ( is_plugin_active( 'newspack-media-partners/newspack-media-partners.php' ) ) {
-			deactivate_plugins( 'newspack-media-partners/newspack-media-partners.php' );
-			Settings::activate_optional_module( 'media-partners' );
-		}
-
 		if ( ! Settings::is_optional_module_active( 'media-partners' ) ) {
 			return;
 		}
 
+		add_filter( 'admin_init', [ __CLASS__, 'switch_from_standalone_plugin' ] );
 		add_action( 'init', [ __CLASS__, 'register_taxonomies' ] );
 
 		add_action( 'partner_add_form_fields', [ __CLASS__, 'add_partner_meta_fields' ] );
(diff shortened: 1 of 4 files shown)


includes/optional-modules/class-media-partners.php
```diff
@@ -0,0 +1,522 @@
...
+	 * Initialize everything.
+	 */
+	public static function init() {
+		if ( ! function_exists( 'is_plugin_active' ) ) {
+			include_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+
+		// If the standalone plugin is active, deactivate it and activate as a module.
+		if ( is_plugin_active( 'newspack-media-partners/newspack-media-partners.php' ) ) {
+			deactivate_plugins( 'newspack-media-partners/newspack-media-partners.php' );
```
leogermani, line 27: Apparently this is being called too early and  is causing https://github.com/Automattic/newspack-manager/blob/2993ce9f04612c581a08ac7131f6036a953560a1/includes/class-notifier.php#L61 to fail because the function `wp_get_current_user` does not exit. I suppose this coud cause other conlficts with other plugins as well...

Maybe delay it or add a try catch?

adekbadek: Good catch! Should be fixed in 500cf1bb68d152388bb402741fc580842e8f3561

includes/optional-modules/class-media-partners.php
```diff
@@ -47,6 +47,7 @@ public static function switch_from_standalone_plugin() {
 			deactivate_plugins( 'newspack-media-partners/newspack-media-partners.php' );
 			Settings::activate_optional_module( 'media-partners' );
 		}
+		return true;
```
leogermani, line 50: just curious. Why we need this?

adekbadek: We don't, I've accidentally used filters instead of actions and then mindlessly followed linter's advice :D Good that you asked! Fixed in ed47d6f808645a22d6636e61ac0297af0918258f

leogermani: ah, I didn't even notice! perfect

PR #2754, opened 2023-11-16
fix(ras): contact metadata sync to mailchimp

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-newsletters/pull/1355, fixes some issues which prevent RAS contact metadata from being fully synced to Mailchimp. The problems are twofold:

1. RAS metadata is currently only synced if the data event contains a list ID to sync to. This means that metadata is not synced for data events such as login, donations/purchases, or registration without signing up for a list. It should be synced to the audience selected in RAS > Advanced Settings, if any.
2. When syncing metadata fields to Mailchimp merge fields, we hit the [`merge-fields` endpoint](https://mailchimp.com/developer/marketing/api/list-merges/list-merge-fields/). By default, this endpoint only returns a max of 10 results. So if there are more than 10 merge fields in the audience (likely there will be), we won't see all of the available fields and will attempt to create duplicates of fields that have already been created.

This PR also introduces the concept of different membership statuses to distinguish between donors and paid members. This is increasingly becoming a requirement for sites that collect both donations and paid purchases. However, there's a weakness in the current logic that we may want to address in the future. The data sync will currently update based on the reader's most recent transaction, so it's possible that if a reader has both a recurring donation and a paid non-donation subscription, their status will ping pong back and forth between `Donor` and `Subscriber`. We may want to address this in another PR to establish a hierarchy of statuses based on conversion funnel depth, but I think that goes beyond the scope of this PR.

This will self-correct for readers who log in, but readers who don't log in may have outdated or incomplete metadata in Mailchimp. We can correct this in bulk by running a Woo resync script for sites if necessary.

1. On `master` in this plugin and Newsletters, set up RAS, connect Mailchimp as the ESP and set an audience in RAS advanced settings.
4. As a new reader, purchase a donation or non-donation product without signing up for any newsletters and observe that in Mailchimp, only a small amount of reader data is synced to the contact in Mailchimp, and few or none of the [metadata fields](https://github.com/Automattic/newspack-plugin/blob/master/includes/plugins/class-newspack-newsletters.php#L26).
5. (optional) If your Mailchimp account has 30 merge fields, observe that there may be duplicate merge fields in the audience, and that not all of the metadata fields have been created as merge fields.
6. Check out this branch and https://github.com/Automattic/newspack-newsletters/pull/1355.
7. In a new session, log into the account that you created in step 3. Confirm that full metadata is now synced for the contact in Mailchimp.
8. Also confirm that the `NP_Membership Status` field value now distinguishes between `Donor` (has made a one-time or has an active recurring donation), `Buyer` (has made a one-time non-donation purchase), `Subscriber` (has an active non-donation subscription). Or `Ex-Donor` or `Ex-Subscriber` if the recurring subscription has been cancelled or expired due to non-payment.
9. Test other scenarios that would trigger a metadata sync to the ESP, such as new account registration via the Registration block, newsletter signup via the Newsletter Subscription Form block, one-time and recurring donations, and purchases of non-donation one-time and subscription products, as well as logging in as those accounts in each case.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -31,6 +31,12 @@ class WooCommerce_Connection {
 		'switched',
 	];
 
+	const MEMBERSHIP_STATUSES = [
+		'donor'      => 'Donor', // Has made at least one one-time or recurring donation.
+		'buyer'      => 'Buyer', // Has purchased a non-recurring non-donation product.
+		'subscriber' => 'Subscriber', // Has purchased a recurring non-donation product.
+	];
+
 	private static $created_membership_id; // phpcs:ignore Squiz.Commenting.VariableComment.Missing
 
 	/**
@@ -169,7 +175,7 @@ class WooCommerce_Connection {
 			return;
 		}
 
-		self::sync_reader_from_order( $customer->get_last_order() );
+		self::sync_reader_from_order( $customer->get_last_order( 'all' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


leogermani:
Here's what I could test:

* I confirm that some metadata is missing without this PR and the PR on the newsletter plugin
* For donations made through a Checkout button (not a donation block), the donation metadata is not filled
* The HAS_DONATED metadata is always empty (is it some legacy on my MC testing account?)
* newsletters_subscription_method is not populated from the post-checkout newsletter subscription. We need to add it there

dkoo:
After some discussion, we've decided to remove the membership status updates from this PR and handle it in a separate PR. This PR (and its [counterpart](https://github.com/Automattic/newspack-newsletters/pull/1355)) now only encompass fixing the contact metadata sync to Mailchimp, and ensuring that the sync only occurs for donation-type transactions. Non-donation-type transactions will be handled with the new membership status updates in the upcoming separate PR.

dkoo:
Closing this as we've decided to rethink the approach as a wider, non-urgent change.

PR #2755, opened 2023-11-16
fix(data-events): no longer use ActionScheduler for dispatches

miguelpeixe:
### All Submissions:

Removes the use of ActionScheduler for dispatching data events. It should use the fallback strategy in place for when AS is not available, which is to send an unblocking HTTP request.

The reason for this change is that the amount of data events we dispatch drastically increased, especially with Newspack Newsletters' tracking functionality. This can be too much for AS' queue to handle.

This change shouldn't impact the processing or reliability of data events, since this is already the strategy in place for sites without AS available (without WooCommerce installed).

Update: 30ff7642fb84d141b43efb53b9fd6147eee28fdd adds a check for a WC function on the GA4 connector so it doesn't fatal if WC is not installed.

1. Checkout this branch, make sure you have GA4 connected and `define( 'NEWSPACK_LOG_LEVEL', 2 );`
2. Make sure you also have a global webhook configured in Newspack -> Connections. Make sure you have `define( 'NEWSPACK_EXPERIMENTAL_WEBHOOKS', true );`. You can use https://webhook.site/ to register a test endpoint
3. Go through the RAS registration flow and confirm the dispatches are logged, e.g.: `Dispatching action "reader_logged_in".`
4. Confirm you see the events also reaching GA4 through GA's real-time report page.
5. Confirm the request also reached the webhook endpoint

### Other information:

### DIFF
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -43,26 +43,8 @@ final class Data_Events {
 	public static function init() {
 		\add_action( 'wp_ajax_' . self::ACTION, [ __CLASS__, 'maybe_handle' ] );
 		\add_action( 'wp_ajax_nopriv_' . self::ACTION, [ __CLASS__, 'maybe_handle' ] );
-		\add_action( 'newspack_data_events_as_dispatch', [ __CLASS__, 'handle' ], 10, 4 );
 	}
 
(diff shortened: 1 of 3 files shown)


leogermani (approved):
Works well!

PR #2760, opened 2023-11-17
fix: sync non-donation data to ESP, but not donor status

dkoo:
### All Submissions:

Testing #2758, I have an alternative proposal. When a reader completes a non-donation order, we should still sync relevant reader and transaction details such as the reader's billing details, total spend, last order amount and product, etc. These transactions should factor into total spend. We just don't need to update the "Donor" or "Ex-Donor" membership status fields, because these transactions shouldn't affect the reader's donor status at all.

This proposal shouldn't change anything about the other next steps we discussed in the other PR:

> 2. Unify the syncing strategy to use the existing data events connectors and no longer the `WooCommerce_Connection::sync_reader_from_order` for better maintainability and to avoid confusion.
> 3. Discuss/decide how we'd like the schema and implement support for non-donation product data sync.

Same instructions as in #2758, but in step 3, confirm that only the following fields are synced, not the Membership Status field:

Screenshot 2023-11-16 at 5 25 12 PM

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -171,11 +171,6 @@ class WooCommerce_Connection {
 
 		$last_order = $customer->get_last_order();
 
-		// If last order is not a donation, don't sync.
-		if ( ! Donations::is_donation_order( $last_order ) ) {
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Good approach! Working well!

PR #2769, opened 2023-11-23
fix(recaptcha): refresh token on checkout error

miguelpeixe:
### All Submissions:

When going through checkout and the submission fails, the next attempt might send the same token, which is not allowed by reCAPTCHA.

This change ensures the captcha token is regenerated on checkout updates and errors to prevent duplicate tokens while placing the order.

1. While on the master branch, add an item to cart and visit the regular WC checkout page (/checkout/)
2. Place order without entering the required billing fields and confirm you get a regular WC error
3. Place the order again and confirm you get a `reCaptcha error: timeout-or-duplicate` along with the other validation errors
4. Check out this branch, repeat steps 2 and 3 and confirm you only get the validation errors
5. Without refreshing, fill the required fields and confirm you are able to place the order

### Other information:

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -323,7 +323,7 @@ final class Recaptcha {
 		<script>
 			grecaptcha.ready( function() {
 				var field;
-				setInterval( function() {
+				function update() {
 					grecaptcha.execute(
 						'<?php echo \esc_attr( $site_key ); ?>',
 						{ action: 'checkout' }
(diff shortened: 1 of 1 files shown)


leogermani:
I faced this issue many times in the last couple of days. But now that I wanted to test this PR I can no longer reproduce it. Will try again Monday

miguelpeixe:
Since it also refreshes the token every 30 seconds, you might not be able to reproduce if the interval hits the refresh between the thrown error and the next attempt.

leogermani (approved):

PR #2780, opened 2023-11-28
fix: more fixes for Mailchimp RAS data sync

dkoo:
### All Submissions:

Adds `Registration Method` and `Registration Page` metadata to `reader_registered` data events fired by the auth form. Currently new registrations via this method aren't capturing this info.

1. On `master`, visit any page in a new session and click the "Sign In" button in the header, then "I don't have an account". Register a new account via this form.
2. In the connected ESP, observe that the synced contact doesn't have any `NP_Registration Page` or `NP_Registration Method` values.
3. Check out this branch and https://github.com/Automattic/newspack-newsletters/pull/1370, repeat with a different email address, and confirm that the synced contact has a `NP_Registration Page` with the URL of the page you were viewing when you clicked "Sign In", and a `NP_Registration Method` of `auth-form`.

This PR now also contains an alternative solution for the issue described in https://github.com/Automattic/newspack-newsletters/pull/1365. Please follow testing instructions there to test this solution.

### Other information:

### DIFF
--- a/includes/plugins/class-newspack-newsletters.php
+++ b/includes/plugins/class-newspack-newsletters.php
@@ -30,6 +30,11 @@ class Newspack_Newsletters {
 		'signup_page'          => 'Signup Page',
 		'signup_page_utm'      => 'Signup UTM: ',
 		'newsletter_selection' => 'Newsletter Selection',
+		'referer'              => 'Referrer Path',
+		'registration_page'    => 'Registration Page',
+		'current_page_url'     => 'Registration Page',
+		'registration_method'  => 'Registration Method',
+
 		// Payment-related.
 		'membership_status'    => 'Membership Status',
 		'payment_page'         => 'Payment Page',
@@ -50,6 +55,7 @@ class Newspack_Newsletters {
 	 */
 	public static function init() {
 		if ( Reader_Activation::is_enabled() && Reader_Activation::get_setting( 'sync_esp' ) ) {
+			\add_filter( 'newspack_newsletters_contact_data', [ __CLASS__, 'normalize_contact_data' ] );
 			\add_filter( 'newspack_newsletters_contact_lists', [ __CLASS__, 'add_activecampaign_master_list' ], 10, 3 );
 		}
 	}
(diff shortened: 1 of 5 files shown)


miguelpeixe (changes requested):
I'm not sure this is producing the desired data.

The `referer` is stored as the URL of the page I'm at when registering, I expected it to be the page before.

The `current_page_url` is stored with the full URL as a path, which is what `home_url( $path )` does:

```
https://example.com/https://example.com/path
```

The usage of `add_query_arg()` is also not clear to me. If no value is being added in the first argument, what's the purpose of this function?

dkoo:
I was mainly copying what the Register block is doing, but let me revisit in case the functions here are producing different results.

miguelpeixe:
The same code works well for the registration block 🤔 

```
  "data": {
    "user_id": 175,
    "email": "contact@example.com",
    "metadata": {
      "lists": [
        "newspack-1128"
      ],
      "referer": "/register/",
      "current_page_url": "https://site.com/register/",
      "registration_method": "registration-block",
      "newspack_popup_id": false
    }
  }
```

dkoo:
The difference between the Register block and the auth form is that the latter submits registration request via REST API, where as the Register block more simply POSTs the form data. So 64317803 accounts for this by passing the referrer to the auth form as a hidden form field, and assuming that the result of `wp_get_raw_referer()` is actually the page the reader is on when registering, because in the context of the REST API the "referrer" is the page that submitted the request.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -220,9 +220,6 @@ public static function get_contact_from_order( $order, $payment_page_url = false
 		}
 
 		$customer = new \WC_Customer( $user_id );
-		$metadata = [
```
dkoo, line 223: Removing this because we only want the registration method to be set if the order results in a new account registration. And it turns out we're already handling that [here](https://github.com/Automattic/newspack-plugin/blob/master/includes/data-events/class-woo-user-registration.php#L50).

includes/plugins/class-newspack-newsletters.php
```diff
@@ -117,6 +123,45 @@ public static function get_metadata_key( $key ) {
...
+	 * Normalizes contact metadata keys before syncing to ESP. If RAS is enabled, we should favor RAS metadata keys.
+	 * 
+	 * @param array $contact Contact data.
+	 * @return array Normalized contact data.
+	 */
+	public static function normalize_contact_data( $contact ) {
+		// Parse full name into first + last for MC, which stores these as separate merge fields.
+		if ( method_exists( 'Newspack_Newsletters', 'service_provider' ) && 'mailchimp' === \Newspack_Newsletters::service_provider() ) {
+			if ( isset( $contact['name'] ) ) {
+				$name_fragments                    = explode( ' ', $contact['name'], 2 );
```
dkoo, line 136: Hmm, do we need to normalize first + last name for Mailchimp for non-RAS sites, too?

dkoo: 2a301c4c allows normalization for just first/last name to occur for Mailchimp, regardless of whether the site is using RAS.

miguelpeixe (approved):
🎉 

The `normalize_contact_data()` method is working as intended and the sync is going through without issues through the `Mailchimp::put()` method.

The `NP_Registration Page` and `NP_Referrer Path` fields are also being populated with the expected values.

dkoo:
Thanks for the review! Holding for deployment until Monday.

PR #2799, opened 2023-12-08
feat(donation): additional receipt email template variables

miguelpeixe:
### All Submissions:

Add support for the following new template variables to the RR receipt email:

- First name
- Last name
- Product name
- Billing frequency

image

This change should only impact the "Newspack" (WC) platform. Customers using Stripe should not support the new variables.

Along with the new values, the `*AMOUNT*` has changed so it uses a Woo function to determine the formatted value.

This PR also corrects the textdomain for the modified files so it uses `newspack-plugin` instead of `newspack`.

1. Checkout this branch and make sure you have Newspack as your RR platform
2. Edit the Receipt email (Newspack -> Reader Revenue -> Emails)
3. Confirm you see the new variables as above
4. Edit your content to include these variables (see example above)
5. Go through the donation flow for each type (one-time, monthly, and yearly)
6. Confirm all 3 emails include the new values
7. Edit your RR platform and set it to Stripe
8. Edit the receipt email and confirm the instructions do not include the new variables

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -360,7 +360,7 @@ class WooCommerce_Connection {
 
 	/**
 	 * Upsert reader data to the ESP connected via Newspack Newsletters.
-	 * 
+	 *
 	 * @param array $contact Reader data to sync.
 	 * @return bool|WP_Error Contact data if it was added, WP_Error otherwise.
 	 */
@@ -480,7 +480,7 @@ class WooCommerce_Connection {
 		$order->set_billing_first_name( $order_data['name'] );
 
 		if ( isset( $order_data['subscribed'] ) && $order_data['subscribed'] ) {
-			$order->add_order_note( __( 'Donor has opted-in to your newsletter.', 'newspack' ) );
+			$order->add_order_note( __( 'Donor has opted-in to your newsletter.', 'newspack-plugin' ) );
 		}
 
 		if ( ! empty( $order_data['client_id'] ) ) {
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -1182,32 +1182,57 @@ public static function send_customizable_receipt_email( $enable, $order, $class
...
+		];
+		$product_map = [];
+		foreach ( $frequencies as $frequency => $label ) {
+			$product_id = Donations::get_donation_product( $frequency );
+			if ( $product_id ) {
+				$product_map[ $product_id ] = $label;
+			}
+		}
+
+		$item = array_values( $order->get_items() )[0];
```
leogermani, line 1197: tip: `array_shift` is also an alternative here

leogermani (changes requested):
First and last name worked fine, but frequency and product name did not.

image

I also noticed something that doesn't seem to be related with this PR, as I was able to reproduce on master.

When I do a new donation, with an unregistered user in a new session, I don't get the receipt. If I do a second donation in the same session, then I get it. As if the email is only sent after I'm logged in

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
miguelpeixe: Updated in e2cb1073e3886930806c53d18e53c7817a7d8d6a

miguelpeixe:
> First and last name worked fine, but frequency and product name did not.

I was not able to reproduce the issue. Can you confirm it was a donation product and under the "Newspack " RR platform?

> When I do a new donation, with an unregistered user in a new session, I don't get the receipt. If I do a second donation in the same session, then I get it. As if the email is only sent after I'm logged in

That's strange. The email is hooked to `woocommerce_email_enabled_customer_completed_order`, which is triggered by Woo. I don't think we modify the trigger behavior.

leogermani:
Found the issue:

image

The `code` tag was added because I copied and pasted the string from the sidebar, and there they are in a code tag. 

When you double click the string on the sidebar to copy them, it selects the string without the astersks, which I then manually added in the editor.

It's an edge case, but not maybe more people can do the same and it's hard to know things are wrong. If you think there's anything easy to avoid that, we could add it now, otherwise we can leave it open for another PR.

I'm trying to think of a way to avoid this without doing too much work in the UI, If I come up with anything I let you know...

leogermani:
> That's strange. The email is hooked to woocommerce_email_enabled_customer_completed_order, which is triggered by Woo. I don't think we modify the trigger behavior.

This no longer happened to me today

leogermani (approved):

miguelpeixe:
Thanks! I think we should handle this edge case separately.

PR #2800, opened 2023-12-09
feat(ras-acc): support shipping, custom fields, and order notes in modal checkout

dkoo:
### All Submissions:

Adds support for a new option to show different billing fields for donations vs. non-donation transactions, and to show order notes in either case. Also adds support for shipping methods and fields as implemented in https://github.com/Automattic/newspack-blocks/pull/1626.

Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1626.

#### Order notes

1. Visit **Newspack > Reader Revenue > Donations**. Scroll down to the "Donation Billing Fields" section and confirm there's slightly different copy and styles, as well as a `*` indicator next to fields that are required for checkout when shown.

Screenshot 2023-12-08 at 6 53 58 PM

2. Confirm there's a new checkbox option for "Order notes", disabled by default.
3. Check it and save.
4. As a reader, start a new donation and confirm you see the "Order notes" field on the first modal checkout screen:

Screenshot 2023-12-08 at 6 41 34 PM

5. Fill out the Order Notes field and complete the transaction. In the admin, find the order and confirm that your note is attached to the order:

Screenshot 2023-12-08 at 6 42 29 PM

#### Customized billing fields

1. Install and activate the the [Checkout Field Editor plugin](https://woo.com/products/woocommerce-checkout-field-editor/) (let me know if you need the package as it's a premium plugin).
2. In WP admin, visit **WooCommerce > Checkout Fields**. Make some customizations to the core billing fields, such as changing the labels and placeholders, as well as the "required" validation status for some fields (it will only let you make certain fields optional, such as Phone). Also add some of your own custom fields with various options. Save.

Screenshot 2023-12-08 at 6 51 35 PM

3. Visit **Newspack > Reader Revenue > Donations**. Confirm that the available fields for both the Donation and Non-donation Billing Fields sections now reflect your customizations from step 2: core field labels will be as configured in the Checkout Fields plugin, and your custom fields will appear as additional fields that can be enabled or disabled like the other fields.
4. Enable the customized core fields and the new custom fields for both Donation and Non-donation Billing Fields and save.
5. As a reader, complete both a donation and a non-donation transaction and confirm that your customizations and custom fields are presented during checkout, and that you can complete both transactions as expected.

Screenshot 2023-12-08 at 6 52 26 PM

### Other information:

---
  - https://app.asana.com/0/0/1205855958320407

### DIFF
--- a/assets/wizards/readerRevenue/views/donation/index.tsx
+++ b/assets/wizards/readerRevenue/views/donation/index.tsx
@@ -2,7 +2,7 @@
  * WordPress dependencies.
  */
 import { useDispatch } from '@wordpress/data';
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import { ToggleControl, CheckboxControl } from '@wordpress/components';
 
 /**
@@ -41,6 +41,16 @@ const FREQUENCIES: {
 };
 const FREQUENCY_SLUGS: FrequencySlug[] = Object.keys( FREQUENCIES ) as FrequencySlug[];
 
+type FieldConfig = {
+	autocomplete: string;
+	class: string[];
+	label: string;
+	priority: number;
+	required: boolean;
+	type: string;
+	validate: string[];
+};
+
 type WizardData = {
 	donation_data:
 		| { errors: { [ key: string ]: string[] } }
@@ -61,16 +71,9 @@ type WizardData = {
 		status: string;
 	};
 	available_billing_fields: {
-		[ key: string ]: {
-			autocomplete: string;
-			class: string[];
-			label: string;
-			priority: number;
-			required: boolean;
-			type: string;
-			validate: string[];
-		};
+		[ key: string ]: FieldConfig;
 	};
+	order_notes_field: FieldConfig;
 };
 
 export const DonationAmounts = () => {
@@ -258,6 +261,8 @@ const BillingFields = () => {
 		} );
 
 	const availableFields = wizardData.available_billing_fields;
+	const orderNotesField = wizardData.order_notes_field;
+	console.log( wizardData );
 	if ( ! availableFields || ! Object.keys( availableFields ).length ) {
 		return null;
 	}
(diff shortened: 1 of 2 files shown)


leogermani (approved):
Everything works as expected. The custom fields validation also work. The only thing I noticed is that the validation message for the custom fields, created via the plugin:

* Show up at the top of the modal, not by the field as all others
* Are not translated, as all others

These should all be related to the plugin, so it's fine.

## Design review needed

I also want to flag that we need to rethink the Reader Revenue > Donations admin page. (And excuse me if this is already covered).

Not sure if in this PR, but certainly during this project, I think we should refactor how we present the options there a bit (before we work on the whole rearchitecture), because it's very confusing.

The page is about donations, but then all of a sudden there are "non-donation fields there".

Also, in the "Donation fields" we say "Configure which billing fields should be shown by the donation checkout form.". And then, just a sentence later: "Note that for shippable products, address fields will always be shown." It doesn't make a lot of sense.

We need to be explicit about where those billing fields apply -> The checkout button and the donation blocks.

Maybe have a "Billing fields" section of its own, separated from "Donations".

Is this flagged somewhere already?

miguelpeixe:
> The page is about donations, but then all of a sudden there are "non-donation fields there".

I agree. We should be more intentional about how we handle this field customization. "Non-donation fields" is too abstract of a feature and competes with products like the one mentioned (Checkout Field Editor).

As this is meant for the modal checkout, the place where we temporarily placed the "Donation fields" as the field selection for the checkout suggests a path forward:

https://github.com/Automattic/newspack-blocks/blob/115e9fb95c78a13f1d87f4d086e767311dc7007d/includes/class-modal-checkout.php#L603-L609

Having the field customization as a Checkout Button block feature is a more flexible approach. It adds complexity, as the validation must be stored separately, but makes more sense to me as a feature.

dkoo:
I hear you both, and I like the idea of approaching the field customization as a feature of the Checkout Button block. How about I remove that feature from this PR and we tackle that separately? I had mainly embarked on it here because it felt like an easy next step (since I was already in that part of the code for the Order Notes field).

Is the order notes field something that works well enough to keep here, in your opinion?

miguelpeixe:
Yep! I agree we can tackle it as a separate feature and the order notes field is a good addition to the existing "donate fields" custom selection.

dkoo:
@leogermani @miguelpeixe I've removed the feature to allow different billing fields for donations vs. non-donation transactions and updated testing instructions. Could use one more look, please!

leogermani:
Looking at this again now. There's a conflict there if you can have a look at it

leogermani:
@dkoo It seems you forgot to push your latest changes. The last commit is still from before my last comment

dkoo:
> @dkoo It seems you forgot to push your latest changes. The last commit is still from before my last comment

Whoops, you're right. I force-pushed to revert the non-donation billing commits.

leogermani (approved):
:+1:

PR #2805, opened 2023-12-12
feat: modal checkout redesign tweaks

miguelpeixe:
### All Submissions:

Combines #2794 and #2771 for https://github.com/Automattic/newspack-blocks/pull/1602.

2aaa84ade9633864be5e2a65540ff80d191cd875 skips reCAPTCHA when only performing form validation. There's no need for the extra API request if it's already performed while placing the order.

29ca89219c1fdeec12aef57142e6c40c63e9b8a0 moves the custom text for the order button to the main plugin's donation feature. It's best if the modal checkout, where it currently is, doesn't handle donation-specific tweaks.

Also changes how the initializer method registers hooks. I noticed all the hooks are registered inside a ! is_admin() condition, which doesn't seem like good practice. The origin of this is here https://github.com/Automattic/newspack-plugin/pull/225 and the intention is to prevent the process_donation_form, hooked to wp_loaded from running in the admin. This check was moved to inside the process_donation_form, which is more appropriate for that purpose.

Instructions at https://github.com/Automattic/newspack-blocks/pull/1602

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -74,6 +74,7 @@ class Donations {
 		add_filter( 'amp_skip_post', [ __CLASS__, 'should_skip_amp' ], 10, 2 );
 		add_filter( 'newspack_blocks_donate_billing_fields_keys', [ __CLASS__, 'get_billing_fields' ] );
 		add_action( 'woocommerce_checkout_create_order_line_item', [ __CLASS__, 'checkout_create_order_line_item' ], 10, 4 );
+		add_filter( 'woocommerce_coupons_enabled', [ __CLASS__, 'disable_coupons' ] );
 		add_filter( 'wcs_place_subscription_order_text', [ __CLASS__, 'order_button_text' ], 9 );
 		add_filter( 'woocommerce_order_button_text', [ __CLASS__, 'order_button_text' ], 9 );
 		add_filter( 'option_woocommerce_subscriptions_order_button_text', [ __CLASS__, 'order_button_text' ], 9 );
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -1069,7 +1069,7 @@ class Donations {
 	 */
 	public static function order_button_text( $text ) {
 		if ( self::is_donation_cart() ) {
-			return __( 'Donate now', 'newspack-blocks' );
+			return __( 'Donate now', 'newspack-plugin' );
 		}
 		return $text;
 	}
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -65,17 +65,18 @@ class Donations {
 	 */
 	public static function init() {
 		self::$donation_product_name = __( 'Donate', 'newspack' );
-		if ( ! is_admin() ) {
-			add_action( 'wp_loaded', [ __CLASS__, 'process_donation_form' ], 99 );
-			add_action( 'woocommerce_checkout_update_order_meta', [ __CLASS__, 'woocommerce_checkout_update_order_meta' ] );
-			add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'woocommerce_billing_fields' ] );
-			add_filter( 'pre_option_woocommerce_enable_guest_checkout', [ __CLASS__, 'disable_guest_checkout' ] );
-			add_action( 'woocommerce_check_cart_items', [ __CLASS__, 'handle_cart' ] );
-			add_filter( 'amp_skip_post', [ __CLASS__, 'should_skip_amp' ], 10, 2 );
-			add_filter( 'newspack_blocks_donate_billing_fields_keys', [ __CLASS__, 'get_billing_fields' ] );
-			add_action( 'woocommerce_checkout_create_order_line_item', [ __CLASS__, 'checkout_create_order_line_item' ], 10, 4 );
-			add_action( 'woocommerce_coupons_enabled', [ __CLASS__, 'disable_coupons' ] );
-		}
+
+		add_action( 'wp_loaded', [ __CLASS__, 'process_donation_form' ], 99 );
+		add_action( 'woocommerce_checkout_update_order_meta', [ __CLASS__, 'woocommerce_checkout_update_order_meta' ] );
+		add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'woocommerce_billing_fields' ] );
+		add_filter( 'pre_option_woocommerce_enable_guest_checkout', [ __CLASS__, 'disable_guest_checkout' ] );
+		add_action( 'woocommerce_check_cart_items', [ __CLASS__, 'handle_cart' ] );
+		add_filter( 'amp_skip_post', [ __CLASS__, 'should_skip_amp' ], 10, 2 );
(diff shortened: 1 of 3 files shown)


includes/class-donations.php
```diff
@@ -1055,5 +1061,17 @@ public static function disable_coupons( $enabled ) {
...
 	}
+
+	/**
+	 * Set the "Place order" button text.
+	 *
+	 * @param string $text The button text.
+	 */
+	public static function order_button_text( $text ) {
+		if ( self::is_donation_cart() ) {
+			return __( 'Donate now', 'newspack-blocks' );
```
dkoo, line 1072:
```suggestion
			return __( 'Donate now', 'newspack-plugin' );
```

includes/class-recaptcha.php
```diff
@@ -366,6 +366,10 @@ public static function verify_recaptcha_on_checkout() {
 		if ( ! $should_verify_captcha ) {
 			return;
 		}
+		// Bail if validating form.
+		if ( isset( $_POST['woocommerce_checkout_update_totals'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
dkoo, line 370: Non-blocking suggestion: would it be useful to create a utility method in Blocks' `class-modal-checkout.php` to determine whether a form submission request is a modal checkout form validation request, in case we want to check other things in addition to this specific `POST` value?

dkoo (changes requested):
Looking good! Very minor feedback below.

includes/class-recaptcha.php
miguelpeixe: Not sure... We use `woocommerce_checkout_update_totals` specifically because that's what Woo provides. We don't have other ways to validate the checkout form.

miguelpeixe:
Update: 6ff7dd04fba6ab130fba0bdf443bd03bf31e4219 tweaks the "cover fee" so it updates more instances in which the price is rendered. Fixes an issue when rendering the order review table: https://github.com/Automattic/newspack-blocks/pull/1602#pullrequestreview-1787910743

miguelpeixe:
The solution proposed in 6ff7dd04fba6ab130fba0bdf443bd03bf31e4219 doesn't work well. Every price is being contained in the `newspack-wc-price` class, so the result is this:

image

To properly implement the "cover fee" integrated with Woo's taxes or other features that may impact the final order price, the strategy for this might have to change. It's currently not working well in master either, so I think we should tackle it separately so it's no tied to this epic.

miguelpeixe:
More on the "cover fee". The design proposes adding a row to the table, which seems like a better approach which also seems to better integrate with how Woo does things:

image

cc @adekbadek

miguelpeixe:
I started [a draft here](https://github.com/Automattic/newspack-plugin/pull/2820) to refactor the cover fee approach to use Woo's `WC()->cart->add_fee()`. This should tackled outside of the RAS-ACC project since it's affecting live sites.

includes/class-recaptcha.php
dkoo: And that's fine, so right now such a method could simply be

```php
public static function is_validation_request() {
	return isset( $_POST['woocommerce_checkout_update_totals'] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
}
```

But it gives us the option to add other conditions like checking our own custom params if we ever want to in the future. Definitely not a requirement at this point.

miguelpeixe: At this point, the code is about Woo's `update_totals`. By calling a Block's method we'd be making it not about Woo, but our code. It may not benefit the codebase unless we identify other use cases.

dkoo (approved):
> I started [a draft here](https://github.com/Automattic/newspack-plugin/pull/2820) to refactor the cover fee approach to use Woo's `WC()->cart->add_fee()`. This should be tackled outside of the RAS-ACC project since it's affecting live sites.

Sounds good! In that case this looks good.

includes/class-donations.php
```diff
@@ -65,17 +65,19 @@ class Donations {
...
-			add_action( 'woocommerce_checkout_update_order_meta', [ __CLASS__, 'woocommerce_checkout_update_order_meta' ] );
-			add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'woocommerce_billing_fields' ] );
-			add_filter( 'pre_option_woocommerce_enable_guest_checkout', [ __CLASS__, 'disable_guest_checkout' ] );
-			add_action( 'woocommerce_check_cart_items', [ __CLASS__, 'handle_cart' ] );
-			add_filter( 'amp_skip_post', [ __CLASS__, 'should_skip_amp' ], 10, 2 );
-			add_filter( 'newspack_blocks_donate_billing_fields_keys', [ __CLASS__, 'get_billing_fields' ] );
-			add_action( 'woocommerce_checkout_create_order_line_item', [ __CLASS__, 'checkout_create_order_line_item' ], 10, 4 );
-			add_action( 'woocommerce_coupons_enabled', [ __CLASS__, 'disable_coupons' ] );
-		}
+
```
dkoo, line 68: I'm not sure if this is related to removing this block, but the warning is coming from this file. On this branch I see a lot of this warning in debug.log:

```
wc_get_product was called incorrectly. wc_get_product should not be called before the woocommerce_init, woocommerce_after_register_taxonomy and woocommerce_after_register_post_type actions have finished. Backtrace: require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, Automattic\WooCommerce\Admin\Features\Features::load_features, Automattic\WooCommerce\Internal\Admin\Coupons->__construct, Automattic\WooCommerce\Internal\Admin\Notes\CouponPageMoved->init, wc_coupons_enabled, apply_filters('woocommerce_coupons_enabled'), WP_Hook->apply_filters, Newspack\Donations::disable_coupons, Newspack\Donations::is_donation_cart, Newspack\Donations::get_donation_product_child_products_ids, Newspack\Donations::get_parent_donation_product, wc_get_product, wc_doing_it_wrong. This message was added in version 3.9.
```

miguelpeixe: From your stack, looks like it's coming from here:

https://github.com/Automattic/newspack-plugin/blob/1dcfd82d04755c5f781bc79195dfcd095f526ac9/includes/class-donations.php#L1047-L1063

I'll take a closer look tomorrow

miguelpeixe: Found it! Here's a PR with the fix: https://github.com/Automattic/newspack-plugin/pull/2822

Pushing the fix separately as it's unrelated to RAS-ACC and should be out sooner.

dkoo: Aha, I see! Thanks for the quick turnaround. However, I don't think we need to push #2822 separately to `master` as we're only removing the [`if ( ! is_admin() )`](https://github.com/Automattic/newspack-plugin/blob/master/includes/class-donations.php#L68) condition in this PR.

miguelpeixe: Of course! Updated it here with b0c4828c2869456a01d0a9f37468573a2c778ceb

dkoo (approved):

PR #2806, opened 2023-12-13
feat: give editors permission for restricted content

leogermani:
### All Submissions:

Allow editors to see content locked behind a membership

1. Create a user with the `editor` role and make sure they don't have any memberships
2. Set up a content gate
3. Visit the site anonymously and confirm the content is restricted
4. on `master`, log in as the editor and confirm the content is still restricted
5. Checkout this branch and confirm the content is allowed for the Editor, but still restricted for other logged in users in other roles (except admins)

Dev notes:

I just want to clarify something that may show up during review.

As you'll see, I changed the `if` statement with a `select`, so I could catch many other capabilities all at once.

The original code had a `break;` at the end of the `if`, meaning that the `foreach` was not continuing in case `$caps[0]` was the cap we are looking for. (in many examples in the internet for the `user_has_cap` filter you'll see that people only check for `$caps[0]`, which is not totally correct. So this `foreach` approach here is better)

But what I wanted to say that is now, all the `break`s we have are breaking out of the `switch` clause and not out of the `foreach` - and that's fine, it's not an oversight. It doens't really matter. If it happens that we receive more than one cap in the checks, the code will still work the same. And we usually only get more than one cap there when checking for meta caps, which is not the case.

### Other information:

---
  - https://app.asana.com/0/0/1205902823686548

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -778,8 +778,8 @@ class Memberships {
 							break;
 						}
 
-						// Allow editors.
-						if ( isset( $all_caps['editor'] ) && true === $all_caps['editor'] ) {
+						// Allow ueser who can edit posts (by default: editors, authors, contributors).
+						if ( isset( $all_caps['edit_posts'] ) && true === $all_caps['edit_posts'] ) {
 							$all_caps[ $cap ] = true;
 							break;
 						}
@@ -810,8 +810,8 @@ class Memberships {
 					case 'wc_memberships_view_delayed_taxonomy_term':
 					case 'wc_memberships_view_delayed_post_content':
 					case 'wc_memberships_view_delayed_product':
-						// Allow editors.
-						if ( isset( $all_caps['editor'] ) && true === $all_caps['editor'] ) {
+						// Allow ueser who can edit posts (by default: editors, authors, contributors).
+						if ( isset( $all_caps['edit_posts'] ) && true === $all_caps['edit_posts'] ) {
 							$all_caps[ $cap ] = true;
 							break;
 						}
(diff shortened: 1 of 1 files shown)


includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -770,24 +770,52 @@ public static function user_has_cap( $all_caps, $caps, $args ) {
...
+						if ( self::can_manage_woocommerce( $all_caps ) ) {
+							$all_caps[ $cap ] = true;
+							break;
+						}
 
-					if ( wc_memberships()->get_restrictions_instance()->is_post_public( $post_id ) ) {
-						$all_caps[ $cap ] = true;
-						break;
-					}
+						// Allow ueser who can edit posts (by default: editors, authors, contributors).
```
miguelpeixe, line 781:
```suggestion
						// Allow user who can edit posts (by default: editors, authors, contributors).
```

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -770,24 +770,52 @@ public static function user_has_cap( $all_caps, $caps, $args ) {
...
+					case 'wc_memberships_view_delayed_product_taxonomy_term':
+					case 'wc_memberships_view_restricted_taxonomy_term':
+					case 'wc_memberships_view_restricted_taxonomy':
+					case 'wc_memberships_view_restricted_post_type':
+					case 'wc_memberships_view_delayed_post_type':
+					case 'wc_memberships_view_delayed_taxonomy':
+					case 'wc_memberships_view_delayed_taxonomy_term':
+					case 'wc_memberships_view_delayed_post_content':
+					case 'wc_memberships_view_delayed_product':
+						// Allow ueser who can edit posts (by default: editors, authors, contributors).
```
miguelpeixe, line 813:
```suggestion
						// Allow user who can edit posts (by default: editors, authors, contributors).
```

miguelpeixe (approved):

PR #2808, opened 2023-12-14
feat: force option to enble retries of failed payments

dkoo:
### All Submissions:

Forces the WooCommerce Subscriptions option to "Retry Failed Payments" to ON unless the `NEWSPACK_PREVENT_WC_ALLOW_FAILED_PAYMENT_RETRIES_OVERRIDE` constant is enabled in wp-config.php.

1. Check out this branch.
2. Similarly to #2784, go to **WP admin > WooCommerce > Settings > Subscriptions** and test that "Retry Failed Payments" is enabled by default and can't be turned off.
3. Optionally, find an existing subscription ID on your site and update or delete the `_stripe_source_id` meta value to force the next renewal payment to fail (`wp post meta update <subscription_id> _stripe_source_id fake_token`) then trigger a manual renewal for that subscription in WP admin and confirm that it enters renewal retry mode (note the "**Renewal Payment Retry:** In 12 hours" in the Schedule panel):

Screenshot 2023-12-13 at 7 27 53 PM

4. Add the `NEWSPACK_PREVENT_WC_ALLOW_FAILED_PAYMENT_RETRIES_OVERRIDE` constant to wp-config.php and confirm you can update the option as usual.

### Other information:

---
  - https://app.asana.com/0/0/1206145382718791

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -1175,12 +1175,11 @@ class WooCommerce_Connection {
 	 * 
 	 * See: https://woo.com/document/subscriptions/failed-payment-retry/
 	 * 
-	 * @param bool   $should_retry Whether WooCommerce should automatically retry failed payments.
-	 * @param string $option_name The name of the option.
+	 * @param bool $should_retry Whether WooCommerce should automatically retry failed payments.
 	 *
 	 * @return string Option value.
 	 */
-	public static function force_allow_failed_payment_retry( $should_retry, $option_name ) {
+	public static function force_allow_failed_payment_retry( $should_retry ) {
 		if ( defined( 'NEWSPACK_PREVENT_WC_ALLOW_FAILED_PAYMENT_RETRIES_OVERRIDE' ) && NEWSPACK_PREVENT_WC_ALLOW_FAILED_PAYMENT_RETRIES_OVERRIDE ) {
 			return $can_switch;
 		}
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -45,10 +45,10 @@ class WooCommerce_Connection {
 		\add_action( 'admin_init', [ __CLASS__, 'disable_woocommerce_setup' ] );
 		\add_filter( 'option_woocommerce_subscriptions_allow_switching', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
 		\add_filter( 'option_woocommerce_subscriptions_allow_switching_nyp_price', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
-		\add_filter( 'option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
+		\add_filter( 'option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_failed_payment_retry' ], 10, 2 );
 		\add_filter( 'default_option_woocommerce_subscriptions_allow_switching', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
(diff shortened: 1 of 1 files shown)


leogermani (changes requested):
I believe you want to update the comments on the `force_allow_subscription_switching` callback method, where it describes exactly what that callback is affecting.

Also, is it something that will always be tied along those other options? My question is, does it make sense to use the same constant or should it be a constant of its own? Feels like something different

dkoo:
> Also, is it something that will always be tied along those other options? My question is, does it make sense to use the same constant or should it be a constant of its own? Feels like something different

Good call. I've separated the functions in in b8adb3e0 and updated the testing instructions with the new constant name.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -1167,6 +1169,25 @@ public static function force_allow_subscription_switching( $can_switch, $option_
...
+	 * See: https://woo.com/document/subscriptions/failed-payment-retry/
+	 * 
+	 * @param bool   $should_retry Whether WooCommerce should automatically retry failed payments.
+	 * @param string $option_name The name of the option.
+	 *
+	 * @return string Option value.
+	 */
+	public static function force_allow_failed_payment_retry( $should_retry, $option_name ) {
+		if ( defined( 'NEWSPACK_PREVENT_WC_ALLOW_FAILED_PAYMENT_RETRIES_OVERRIDE' ) && NEWSPACK_PREVENT_WC_ALLOW_FAILED_PAYMENT_RETRIES_OVERRIDE ) {
+			return $can_switch;
```
leogermani, line 1184:
```suggestion
			return $should_retry;
```

leogermani: nit. if we don't need `option_name` here, we can remove it from the signature and the `2` from `add_filter`

dkoo: Yes, I removed the unused arg in 4d99f80 but forgot to remove it from the action callback definition. Fixed that in 36b10f1a

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -45,8 +45,10 @@ public static function init() {
 		\add_action( 'admin_init', [ __CLASS__, 'disable_woocommerce_setup' ] );
 		\add_filter( 'option_woocommerce_subscriptions_allow_switching', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
 		\add_filter( 'option_woocommerce_subscriptions_allow_switching_nyp_price', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
+		\add_filter( 'option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_failed_payment_retry' ], 10 );
```
leogermani, line 48:
```suggestion
		\add_filter( 'option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_failed_payment_retry' ] );
```

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -45,8 +45,10 @@ public static function init() {
...
 		\add_filter( 'default_option_woocommerce_subscriptions_allow_switching', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
 		\add_filter( 'default_option_woocommerce_subscriptions_allow_switching_nyp_price', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
+		\add_filter( 'default_option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_failed_payment_retry' ], 10 );
```
leogermani, line 51:
```suggestion
		\add_filter( 'default_option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_failed_payment_retry' ] );
```

leogermani (approved):
Working well. Left two minor suggestions

dkoo:
Thanks for the incremental line-by-line updates, @leogermani. I must be trying to do too many things at once

PR #2810, opened 2023-12-15
feat: Newspack UI styles

laurelfulford:
### All Submissions:

This PR starts the process of adding more general Newspack UI styles to the Newspack Plugin.

1. Apply the PR to the epic/ras-acc branch. 
2. Confirm that the styles don't really do anything; they shouldn't be applied to any elements currently in the branch or plugin because they require them to be wrapped in the `newspack-ui` CSS class.
3. Confirm that you can load testing components by appending ?ui-demo to the URL, but only when logged in as an administrator.

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -2,12 +2,13 @@
 	&__modal {
 		background: var( --newspack-ui-color-body-bg );
 		border-radius: var( --newspack-ui-border-radius );
+		max-height: 90vh;
 		max-width: var( --newspack-ui-modal-width-medium );
 		overflow: hidden;
 
 		// Modal header & footer
 		&__header,
-		&__footer, {
+		&__footer {
 			padding: var( --newspack-ui-spacer-4 ) var( --newspack-ui-spacer-5 );
 		}
 
@@ -43,6 +44,7 @@
 
 		&__content {
 			backface-visibility: visible;
+			overflow: auto;
 			padding: var( --newspack-ui-spacer-5 );
 		}
 
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -2,6 +2,8 @@
 	&__modal {
 		background: var( --newspack-ui-color-body-bg );
 		border-radius: var( --newspack-ui-border-radius );
+		display: flex;
+		flex-direction: column;
 		max-height: 90vh;
 		max-width: var( --newspack-ui-modal-width-medium );
 		overflow: hidden;
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -6,7 +6,7 @@
 		flex-direction: column;
 		max-height: 90vh;
 		max-width: var( --newspack-ui-modal-width-medium );
-		overflow: hidden;
+		overflow: auto;
 
 		// Modal header & footer
 		&__header,
@@ -16,9 +16,12 @@
 
 		&__header {
 			align-items: center;
+			background: var( --newspack-ui-color-body-bg );
 			border-bottom: 1px solid var( --newspack-ui-color-border );
 			display: flex;
 			justify-content: space-between;
+			position: sticky;
+			top: 0;
 
 			h2 {
 				font-size: var( --newspack-ui-font-size-medium );
@@ -46,7 +49,6 @@
 
 		&__content {
 			backface-visibility: visible;
-			overflow: auto;
 			padding: var( --newspack-ui-spacer-5 );
 		}
 
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -1,0 +1,77 @@
+.newspack-ui {
+	&__modal {
+		background: var( --newspack-ui-color-body-bg );
+		border-radius: var( --newspack-ui-border-radius );
+		max-width: var( --newspack-ui-modal-width-medium );
+		overflow: hidden;
+
+		// Modal header & footer
+		&__header,
+		&__footer, {
+			padding: var( --newspack-ui-spacer-4 ) var( --newspack-ui-spacer-5 );
+		}
+
+		&__header {
+			align-items: center;
+			border-bottom: 1px solid var( --newspack-ui-color-border );
+			display: flex;
+			justify-content: space-between;
+
+			h2 {
+				font-size: var( --newspack-ui-font-size-medium );
+				margin: 0;
+			}
+
+			button { // Close button
+				background: transparent;
+				border: 0;
+				color: var( --newspack-ui-color-text-high-contrast );
+				cursor: pointer;
+				margin: -8px;
+				padding: 8px;
+
+				&:focus,
+				&:hover {
+					background: transparent;
+					color: var( --newspack-ui-color-text-med-contrast );
+				}
+				svg {
+					display: block;
+				}
+			}
+		}
+
+		&__content {
+			backface-visibility: visible;
+			padding: var( --newspack-ui-spacer-5 );
+		}
+
+		&__footer {
+			background: var( --newspack-ui-color-gray-100 );
+			color: var( --newspack-ui-color-text-med-contrast );
+			font-size: var( --newspack-ui-font-size-small );
+			line-height: var( --newspack-ui-line-height-small );
+		}
+
+		// Narrow modal
+		&__small {
+			max-width: var( --newspack-ui-modal-width-small );
+		}
+
+		// Contents
+		h3 {
+			font-size: var( --newspack-ui-font-size-medium );
+		}
+
+		&__footer,
+		&__content {
+			> *:first-child {
+				margin-top: 0;
+			}
+
+			> *:last-child {
+				margin-bottom: 0;
+			}
+		}
+	}
+}
--- a/assets/newspack-ui/scss/elements/_index.scss
+++ b/assets/newspack-ui/scss/elements/_index.scss
@@ -1,0 +1,25 @@
+@use 'boxes';
+@use 'forms';
+@use 'icons';
+@use 'tables';
+@use 'typography';
+
+// TODO: Find a better home for this:
+.newspack-ui {
+	&__word-divider {
+		align-items: center;
+		display: flex;
+		font-size: var( --newspack-ui-font-size-small );
+		gap: var( --newspack-ui-spacer-2 );
+		justify-content: space-evenly;
+		margin: var( --newspack-ui-spacer-5 ) 0;
+
+		&::before,
+		&::after {
+			border-top: 1px solid var( --newspack-ui-color-border );
+			content: '';
+			display: block;
+			width: 100%;
+		}
+	}
+}
--- a/assets/newspack-ui/scss/elements/_index.scss
+++ b/assets/newspack-ui/scss/elements/_index.scss
@@ -1,25 +1,7 @@
 @use 'boxes';
 @use 'forms';
 @use 'icons';
+@use 'misc';
 @use 'tables';
 @use 'typography';
-
-// TODO: Find a better home for this:
-.newspack-ui {
-	&__word-divider {
-		align-items: center;
-		display: flex;
-		font-size: var( --newspack-ui-font-size-small );
-		gap: var( --newspack-ui-spacer-2 );
-		justify-content: space-evenly;
-		margin: var( --newspack-ui-spacer-5 ) 0;
-
-		&::before,
-		&::after {
-			border-top: 1px solid var( --newspack-ui-color-border );
-			content: '';
-			display: block;
-			width: 100%;
-		}
-	}
-}
+@use 'woocommerce';
--- a/assets/newspack-ui/scss/variables/_colors.scss
+++ b/assets/newspack-ui/scss/variables/_colors.scss
@@ -1,0 +1,48 @@
+:root {
+	// Newspack Blues
+	--newspack-ui-color-primary-000: #f5fdff;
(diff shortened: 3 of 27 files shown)


assets/newspack-ui/scss/elements/_index.scss
```diff
@@ -0,0 +1,25 @@
+@use 'boxes';
+@use 'forms';
+@use 'icons';
+@use 'tables';
+@use 'typography';
+
+// TODO: Find a better home for this:
```
miguelpeixe, line 7: I don't mind having misc in the index file. Alternatively, we could go with a `misc.scss`. WDYT?

miguelpeixe: Or even a `misc/` directory with each misc feature as its own `word-divider.scss` file. It helps organizing and forces us to think twice before adding more "misc" to the mix 😅

assets/newspack-ui/scss/variables/_colors.scss
```diff
@@ -0,0 +1,48 @@
...
+	--newspack-ui-color-primary-100: #b4e4ff;
+	--newspack-ui-color-primary-200: #93ccfd;
+	--newspack-ui-color-primary-300: #72affb;
+	--newspack-ui-color-primary-400: #528dfc;
+	--newspack-ui-color-primary-500: #36f;
+	--newspack-ui-color-primary-600: #2240d5;
+	--newspack-ui-color-primary-700: #1522af;
+	--newspack-ui-color-primary-800: #0b0b8d;
+	--newspack-ui-color-primary-900: #0d046e;
+	--newspack-ui-color-primary-950: #0e0052;
```
miguelpeixe, line 14: Not for this PR, but we should tweak this to incorporate a dynamic primary color generated in the backend. The backend logic should use HSL to determine if the publisher's primary color hue is between 50 and 300 (not red). If it's red-ish, it should fall back to a dark grey.

I wonder how to merge that with the proposed grades. If we are to implement different gradings for the publisher's color, I'd propose we do it with opacity rather than dynamically calculating lightness. Still, I'm not a fan of messing with the publisher's brand color in that way...

Correct me if I'm wrong, but I think we only implement the different grades in the checkbox/radio input lists. I _think_ these should always be blue, but we should double-check. If that's the case, we don't need to worry about it.

assets/newspack-ui/scss/variables/_colors.scss
```diff
@@ -0,0 +1,48 @@
...
+	--newspack-ui-color-text-high-contrast: var( --newspack-ui-color-gray-900 );
+	--newspack-ui-color-text-med-contrast: var( --newspack-ui-color-gray-700 );
+	--newspack-ui-color-border: var( --newspack-ui-color-gray-300 );
+	--newspack-ui-color-body-bg: var( --newspack-ui-color-white );
+
+	// Specific assignments - form elements
+	--newspack-ui-color-button-bg: var( --newspack-ui-color-gray-900 );
+	--newspack-ui-color-button-bg-hover: var( --newspack-ui-color-gray-800 ); // TODO: Replace w/correct value.
+	--newspack-ui-color-button-text: var( --newspack-ui-color-white );
+	--newspack-ui-color-radio-bg: var( --newspack-ui-color-gray-900 );
```
miguelpeixe, line 47: Shouldn't this be `--newspack-ui-color-primary-500`?

Ref:

image

assets/newspack-ui/scss/variables/_spacing.scss
```diff
@@ -0,0 +1,12 @@
+:root {
+	// Spacing - not ems because the font size of the element itself caused a cascade
+	--newspack-ui-spacer-1: 8px;
```
miguelpeixe, line 3: Suggestion: change this to `--newspack-ui-spacer-base` for better readability when tweaking per context (in a modal or media queries)

miguelpeixe: Something like this:

```css
--newspack-ui-spacer-base: 8px;
--newspack-ui-spacer-1: var( --newspack-ui-spacer-base );
--newspack-ui-spacer-2: calc( var( --newspack-ui-spacer-base ) * 1.5 ); // 12px
...etc
```

assets/newspack-ui/style.scss
```diff
@@ -0,0 +1,35 @@
...
+	h3 {
+		font-size: var( --newspack-ui-font-size-medium );
+	}
+
+	a,
+	a:hover {
+		color: inherit !important;
+	}
+
+	// TODO: This specifically is a WooCommerce HTML structure; have been using it for other elements so far, but do we want to?
```
miguelpeixe, line 25: We might realize at some point that the newspack-ui should have a `woocommerce/` directory with all the Woo-specific required tweaks.

miguelpeixe (changes requested):
Thank you, @laurelfulford! This is looking great!

Some comments inline:

assets/newspack-ui/scss/variables/_colors.scss
laurelfulford: > we should tweak this to incorporate a dynamic primary color generated in the backend

Yes! 

> I wonder how to merge that with the proposed grades.

I copied these colours over as-is [from the _colors.scss file used for the back-end components](https://github.com/Automattic/newspack-plugin/blob/master/assets/shared/scss/_colors.scss) when I was under the impression a lot more of the back-end would be using Newspack blue. So I think we can probably pare this down to whatever we're actually using. 

> Correct me if I'm wrong, but I think we only implement the different grades in the checkbox/radio input lists. I think these should always be blue, but we should double-check. If that's the case, we don't need to worry about it.

This is my impression now, too! I'll follow up and confirm just in case 🙂

assets/newspack-ui/scss/variables/_colors.scss
laurelfulford: This stemmed from my confusion around how we're handling the colours! Followed up about that here (p1704831597129219-slack-newspack-reader-activation) for this comment and the previous one -- thanks for pointing these out, I kind of forgot about them! 😅

assets/newspack-ui/style.scss
laurelfulford: That's a really good point, and is kind of on trend for what we've had to do for things like the theme. I spun up a seperate /woocommerce directory, and added these styles to an `_overrides.scss` file inside of there, since they're just undoing styles WC is adding so far.

assets/newspack-ui/scss/variables/_colors.scss
laurelfulford: @miguelpeixe Regarding the above, we'll be going forward with the brand colour for primary actions and black/grey as a fallback -- does it sound okay to update everything to use the black/greys for now?

assets/newspack-ui/scss/_modals.scss
```diff
@@ -0,0 +1,77 @@
+.newspack-ui {
+	&__modal {
+		background: var( --newspack-ui-color-body-bg );
+		border-radius: var( --newspack-ui-border-radius );
+		max-width: var( --newspack-ui-modal-width-medium );
```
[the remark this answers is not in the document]
laurelfulford: Good call! This has been added in [5089ffe](https://github.com/Automattic/newspack-plugin/pull/2810/commits/5089ffe0df9b4e09d1bc381792e4ab9e2e234a4b).

assets/newspack-ui/scss/_modals.scss
```diff
@@ -0,0 +1,77 @@
...
+		overflow: hidden;
+
+		// Modal header & footer
+		&__header,
+		&__footer, {
```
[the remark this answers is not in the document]
laurelfulford: 🤦‍♀️ Thanks for catching this! Also fixed in [5089ffe](https://github.com/Automattic/newspack-plugin/pull/2810/commits/5089ffe0df9b4e09d1bc381792e4ab9e2e234a4b).

assets/newspack-ui/scss/_modals.scss
```diff
@@ -0,0 +1,77 @@
...
+				svg {
+					display: block;
+				}
+			}
+		}
+
+		&__content {
+			backface-visibility: visible;
+			padding: var( --newspack-ui-spacer-5 );
+		}
```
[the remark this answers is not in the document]
laurelfulford: Yes! Added in [5089ffe](https://github.com/Automattic/newspack-plugin/pull/2810/commits/5089ffe0df9b4e09d1bc381792e4ab9e2e234a4b).

assets/newspack-ui/scss/elements/_index.scss
laurelfulford: `/misc` directory added in [d81a167](https://github.com/Automattic/newspack-plugin/pull/2810/commits/d81a16709a64af9cae5fb89564ebdbc5acfd8c8b) 🙂

assets/newspack-ui/scss/variables/_colors.scss
laurelfulford: I ended up removing the blue variables in [3bb8809](https://github.com/Automattic/newspack-plugin/pull/2810/commits/3bb8809a9c40527e54dcf1cb709812bef3a9f051) -- should I make a separate task for the custom colour stuff?

assets/newspack-ui/scss/variables/_colors.scss
laurelfulford: The above has been done in [3bb8809](https://github.com/Automattic/newspack-plugin/pull/2810/commits/3bb8809a9c40527e54dcf1cb709812bef3a9f051) kind of as a temporary state 🙂

assets/newspack-ui/style.scss
laurelfulford: Updated in [d81a167](https://github.com/Automattic/newspack-plugin/pull/2810/commits/d81a16709a64af9cae5fb89564ebdbc5acfd8c8b)!

assets/newspack-ui/scss/_modals.scss
```diff
@@ -0,0 +1,77 @@
+.newspack-ui {
+	&__modal {
```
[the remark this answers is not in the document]
laurelfulford: I think that would be good! It would help avoid duplication for anything else that needs a similar modal, too! 🙂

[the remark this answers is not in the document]
miguelpeixe: Cool! We can address that in the next PR so it doesn't block the foundational changes this PR introduces.

assets/newspack-ui/scss/_modals.scss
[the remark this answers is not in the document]
miguelpeixe: Thanks! Following the max-height change, shouldn't the overflow change from hidden to auto?

assets/newspack-ui/scss/variables/_colors.scss
miguelpeixe: Awesome! Thank you for looking further into this. I like the approach with the temporary state, which might turn out to be the permanent state 🤷

assets/newspack-ui/scss/_modals.scss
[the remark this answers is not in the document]
laurelfulford: Awesome, thanks Miguel! I'll spin up a separate task to keep track of this one.

assets/newspack-ui/scss/_modals.scss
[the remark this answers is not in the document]
laurelfulford: @miguelpeixe I _think_ this should be handled by the [overflow added to the `.newspack-ui__modal__content` class](https://github.com/Automattic/newspack-plugin/pull/2810#discussion_r1430170027), right? That means the header bit is always visible, though, and only the content bit works -- that's definitely changeable if it'd be better to scroll everything! 

(Just let me know if I'm missing something here!)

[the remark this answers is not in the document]
miguelpeixe: Currently the `.newspack-ui__modal__content` is not inheriting the max height:

image

This should be fixed through flexboxes, adding `display: flex; flex-direction: column;` to `.newspack-ui__modal`.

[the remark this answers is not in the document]
laurelfulford: OHHHH thank you @miguelpeixe! I was testing the styles with the modal checkout, which was adding an inline height and tricking me into thinking the scrolling was working 😅 I see it now. 

I went the flexbox route in [210c7ba](https://github.com/Automattic/newspack-plugin/pull/2810/commits/210c7ba86fbd7d68a80c23a33fc31b8cb13c2060) and works well -- the only bit I'm waffling about is the modal footer. I'm wondering if it should also scroll instead of remaining in place (like it does with the current markup).

I'm thinking this may be something we can test/verify in action and see if it works or not. Based on the mockups, the only case a modal with a footer _may_ need to scroll would be on mobile, and maybe not even then. (I kind of faked the issue in the screenshot below, but the modal with a footer in the demo content isn't actually tall enough to cause this issue):

image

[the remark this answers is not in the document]
miguelpeixe: Hmmmm, I believe the footer should scroll along.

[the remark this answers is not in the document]
laurelfulford: Yeah, agreed -- it seems weird to treat it with the same importance as the header when it's more of a disclaimer space 🙂 

I've made an update in [2331596](https://github.com/Automattic/newspack-plugin/pull/2810/commits/2331596702837a5854dc47f593dd91f1f5881fdb) that makes the header sticky, and moves the overflow so everything else scrolls but the header stays put!

miguelpeixe (approved):
Thank you for all the discussions and revisions, @laurelfulford!

laurelfulford:
Yay! Thanks for the great review, @miguelpeixe!

PR #2820, opened 2023-12-19
fix: use Woo's cart fee for covering transaction fees

miguelpeixe:
### All Submissions:

Refactors the "cover fee" strategy to use  `WC()->cart->add_fee()` instead of modifying the order total directly.

This PR also changes the UI to show the fee through the order review table and and no longer modify the top summary value. This consolidates how we handle additional values added to the order (coupons, fees, and taxes excluded from the total).

image

image

1. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1633
2. Through the Donate block, start a one-time donation flow and confirm you can toggle the "I’d like to cover the {percentage}% transaction fee to ensure my full donation goes towards {site name}'s mission." checkbox
3. Confirm toggling the checkbox updates the fee as shown in the image above
4. Confirm you can place the order
5. Visit the order details in the dashboard and confirm you can see the fee as shown in the image above
6. Confirm the note "The donor opted to cover Stripe’s transaction fee." is also added to the order
7. Back to the Donate block, start a recurring donation flow, and also check the fee checkbox
8. Confirm the order review table also shows the fee added to the "Recurring totals"
9. Place the order and confirm in the dashboard that the fee has been added
10. Go to the subscription dashboard in WooCommerce -> Subscriptions
11. In the "Subscription actions" metabox, select "Process renewal"
12. Confirm the renewal order also includes the fee

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
@@ -20,8 +20,9 @@ class WooCommerce_Cover_Fees {
 	 */
 	public static function init() {
 		\add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'add_checkout_fields' ] );
-		\add_action( 'woocommerce_checkout_order_processed', [ __CLASS__, 'add_order_note' ], 1, 3 );
+		\add_action( 'woocommerce_checkout_update_order_review', [ __CLASS__, 'persist_fee_selection' ] );
 		\add_action( 'woocommerce_cart_calculate_fees', [ __CLASS__, 'add_transaction_fee' ] );
+		\add_action( 'woocommerce_checkout_order_processed', [ __CLASS__, 'add_order_note' ], 1, 3 );
 		\add_action( 'wc_stripe_payment_fields_stripe', [ __CLASS__, 'render_stripe_input' ] );
 		\add_action( 'wp_enqueue_scripts', [ __CLASS__, 'print_checkout_helper_script' ] );
 	}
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -112,6 +132,26 @@ private static function should_allow_covering_fees() {
 		return false;
```
dkoo, line 132: Non-blocking suggestion: I actually tested changing this to `true` with Woo's standard non-modal checkout flow and it worked flawlessly in that flow, as well. I think we should consider turning it on by default for non-modal donation checkouts too since it works. It would only have any effect for publishers who intentionally use standard checkouts for donations, which could be nobody, but why not? 🤷

dkoo (approved):
Works exactly as described, and I love this change because it clarifies the addition of the fee on top of the subtotal to both the customer and the admin. I also verified that the fee is included in the total populated by our custom receipt emails, so I see no reason to get this change deployed ASAP.

Screenshot 2023-12-22 at 11 55 37 AM

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
miguelpeixe: I guess it would be fine, but the messaging around the feature is all about "donors" and "donations". Let's discover whether it makes sense to ask a customer to pay for the transaction fee when purchasing a product or subscription.

dkoo: It would still only apply to donation transactions because of [this check](https://github.com/Automattic/newspack-plugin/pull/2820/files/9157eba4d68e436c6ce8244f9c1378671734ac68#diff-2f1147d3b8e30d61c3f9c27067874236a35d0aaadee05a80893cda118580d4e5R111). But it's an extreme edge case because donation products are not revealed in the standard Woo shop pages unless the publisher explicitly edits those products to allow it.

miguelpeixe: Whoops, sorry about that. I misinterpreted your inline comment. I agree we can make it a more generic feature

miguelpeixe: Updated in 67bf801d7d629fb094eb5d4209d67601b8d38b03 to support any checkout

dkoo (approved):
LGTM!

PR #2821, opened 2023-12-19
fix(ras): consolidate RAS <> ESP data syncs in data event connectors

dkoo:
### All Submissions:

#2804 fixed syncing of contact metadata for UTM data via purchases, but did not fix for registration via non-transactions. This PR fixes syncing of UTM fields for all registration and purchase-related events, and also does a bit of refactoring to consolidate data syncs in general into the ActiveCampaign and Mailchimp Data Events connectors. Currently, the data syncs are happening in a few different places using different triggers (a mix of Woo hooks, regular WP hooks, Newspack Newsletter hooks, and data events throughout different files). Consolidating them in data events should make it easier to see and control what triggers an ESP sync going forward. As part of this we've added some new data event listeners and tweaked some existing ones to make them a bit more standardized.

In new incognito sessions, perform the following actions from a URL with UTM parameters appended (e.g. `?utm_source=marketinglist&utm_medium=email&utm_campaign=2023-Newspack-Campaign`):

- Register for a new account via the Registration block without signing up for any newsletter lists
- Sign up for newsletter lists via the Subscription Form block or the Registration block
- Complete a one-time donation
- Complete a subscription donation
- Complete a non-donation non-subscription purchase
- Complete a non-donation subscription purchase

Make sure to repeat the above for both Mailchimp and ActiveCampaign (so the end result should be 12 new contacts). In each case, find the contact synced to the ESP and confirm that UTM metadata fields are synced. For registration-based actions, the field keys should start with `Signup UTM:`. For payment-based actions, the field keys should start with `Payment UTM:`.

Please also smoke test that logging into an existing account and making a renewal order on an active subscription also continue to trigger an ESP data sync as before.

### Other information:

---
  - https://app.asana.com/0/0/1206082237141820

### DIFF
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -140,7 +140,6 @@ When any WooCommerce order is completed with a successful payment.
 | `email`           | `string` |                                                        |
 | `amount`          | `float`  |                                                        |
 | `currency`        | `string` |                                                        |
-| `recurrence`      | `string` |                                                        |
 | `platform`        | `string` |                                                        |
 | `referer`         | `string` |                                                        |
 | `popup_id`        | `string` | If the order was triggered by a popup, the popup ID    |
--- a/includes/data-events/README.md
+++ b/includes/data-events/README.md
@@ -130,6 +130,24 @@ When there's a new donation, either through Stripe or Newspack (WooCommerce) pla
 | `subscription_id` | `int`    | The related subscription id (if any)                   |
 | `platform_data`   | `array`  |                                                        |
 
+### `order_completed`
+
+When any WooCommerce order is completed with a successful payment.
+
+| Name              | Type     | Obs                                                    |
+| ----------------- | -------- | ------------------------------------------------------ |
+| `user_id`         | `int`    |                                                        |
+| `email`           | `string` |                                                        |
+| `amount`          | `float`  |                                                        |
+| `currency`        | `string` |                                                        |
+| `recurrence`      | `string` |                                                        |
+| `platform`        | `string` |                                                        |
+| `referer`         | `string` |                                                        |
+| `popup_id`        | `string` | If the order was triggered by a popup, the popup ID    |
+| `is_renewal`      | `bool`   | If this is a subscription renewal (recurring payment)  |
+| `subscription_id` | `int`    | The related subscription id (if any)                   |
+| `platform_data`   | `array`  |                                                        |
+
 ### `donation_subscription_new`
 
 When there's a new WooCommerce Subscription. This action does not replace the `donation_new` that create the subscription.
@@ -176,22 +194,9 @@ When a WooCommerce Subscription status changes.
 | `recurrence`      | `string` |
 | `platform`        | `string` |
 
-### `product_subscription_active`
+### `product_subscription_changed`
 
-When a non-donation subscription is activated.
-
-| Name              | Type     |
-| ----------------- | -------- |
-| `user_id`         | `int`    |
-| `email`           | `string` |
-| `subscription_id` | `int`    |
-| `amount`          | `float`  |
-| `currency`        | `string` |
-| `recurrence`      | `string` |
-
-### `product_subscription_inactive`
-
-When a non-donation subscription is changed to any non-active status.
+When a non-donation subscription status changes.
 
 | Name              | Type     |
 | ----------------- | -------- |
--- a/includes/data-events/connectors/class-mailchimp.php
+++ b/includes/data-events/connectors/class-mailchimp.php
@@ -181,13 +181,15 @@ class Mailchimp {
 			'status_if_new' => 'transactional',
 		];
 
-		$merge_fields = self::get_merge_fields( $audience_id, $data );
+		// Normalize contact metadata.
+		$contact      = Newspack_Newsletters::normalize_contact_data( [ 'metadata' => $data ] );
+		$merge_fields = self::get_merge_fields( $audience_id, $contact['metadata'] );
 		if ( ! empty( $merge_fields ) ) {
 			$payload['merge_fields'] = $merge_fields;
 		}
 
 		Logger::log(
-			'Syncing contact with metadata key(s): ' . implode( ', ', array_keys( $data ) ) . '.',
+			'Syncing contact with metadata key(s): ' . implode( ', ', array_keys( $contact['metadata'] ) ) . '.',
 			Data_Events::LOGGER_HEADER
 		);
 
--- a/includes/data-events/connectors/class-mailchimp.php
+++ b/includes/data-events/connectors/class-mailchimp.php
@@ -32,10 +32,18 @@ class Mailchimp {
 	 * Register handlers.
 	 */
 	public static function register_handlers() {
-		if ( Reader_Activation::is_enabled() && true === Reader_Activation::get_setting( 'sync_esp' ) ) {
+		if ( ! method_exists( 'Newspack_Newsletters', 'service_provider' ) ) {
+			return;
+		}
+		if (
+			Reader_Activation::is_enabled() &&
+			true === Reader_Activation::get_setting( 'sync_esp' ) &&
+			'mailchimp' === \Newspack_Newsletters::service_provider()
+		) {
 			Data_Events::register_handler( [ __CLASS__, 'reader_registered' ], 'reader_registered' );
-			Data_Events::register_handler( [ __CLASS__, 'donation_new' ], 'donation_new' );
-			Data_Events::register_handler( [ __CLASS__, 'donation_subscription_new' ], 'donation_subscription_new' );
+			Data_Events::register_handler( [ __CLASS__, 'reader_logged_in' ], 'reader_logged_in' );
+			Data_Events::register_handler( [ __CLASS__, 'order_new' ], 'donation_new' );
+			Data_Events::register_handler( [ __CLASS__, 'order_new' ], 'product_order_new' );
 		}
 	}
 
@@ -165,27 +173,28 @@ class Mailchimp {
 	/**
 	 * Update a Mailchimp contact
 	 *
-	 * @param string $email Email address.
-	 * @param array  $data  Data to update.
+	 * @param array $contact Contact info to sync to ESP without lists.
 	 *
 	 * @return array|WP_Error response body or error.
 	 */
-	public static function put( $email, $data = [] ) {
+	public static function put( $contact ) {
 		$audience_id = self::get_audience_id();
 		if ( ! $audience_id ) {
 			return;
 		}
-		$hash    = md5( strtolower( $email ) );
+		$hash    = md5( strtolower( $contact['email'] ) );
 		$payload = [
-			'email_address' => $email,
+			'email_address' => $contact['email'],
 			'status_if_new' => 'transactional',
 		];
 
 		// Normalize contact metadata.
-		$contact      = Newspack_Newsletters::normalize_contact_data( [ 'metadata' => $data ] );
-		$merge_fields = self::get_merge_fields( $audience_id, $contact['metadata'] );
-		if ( ! empty( $merge_fields ) ) {
-			$payload['merge_fields'] = $merge_fields;
+		$contact = Newspack_Newsletters::normalize_contact_data( $contact );
+		if ( ! empty( $contact['metadata'] ) ) {
+			$merge_fields = self::get_merge_fields( $audience_id, $contact['metadata'] );
+			if ( ! empty( $merge_fields ) ) {
+				$payload['merge_fields'] = $merge_fields;
+			}
 		}
 
 		Logger::log(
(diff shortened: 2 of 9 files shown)


includes/data-events/connectors/class-mailchimp.php
```diff
@@ -181,13 +181,15 @@ public static function put( $email, $data = [] ) {
 			'status_if_new' => 'transactional',
 		];
 
-		$merge_fields = self::get_merge_fields( $audience_id, $data );
+		// Normalize contact metadata.
```
leogermani, line 184: Don't we need to check if the class exists here? What if Newsletters plugin is not present?

dkoo: In this case `Newspack_Newsletters` refers to the namespaced class in this plugin: https://github.com/Automattic/newspack-plugin/blob/master/includes/data-events/connectors/class-mailchimp.php#L13

leogermani:
@dkoo Two things I couldn't confirm.

1. On Mailchimp, I can't see the fields for donations and purchases, but that's because I've reached the limits. I tried deleting 3 fields from my Audience, but still with no luck. All fields look legit, so maybe some are recreated by us?

2. On AC, everything worked fine, except the purchase of a non-donation product. It's a subscription. I don't see any errors in the log, but I also don't see those `Normalizing contact data for reader ESP sync` entries. The contact was not added to AC at all

adekbadek (commented):
In both AC and Mailchimp: in all cases but the last one (buying a non-donation product) the expected metadata was added.

dkoo:
@leogermani @adekbadek This led me down a deep rabbit hole. It turns out, we had a hole in our Data Events listeners where we haven't been sending any data events upon a non-subscription, non-donation-type purchase. Thus, we haven't been syncing any data from those purchases. Luckily, it's a pretty rare edge case for our publishers, so the impact should be minimal.

Beyond that hole, I had a hard time tracking where the ESP data syncs were happening because they were being triggered in a few different places. Now that we're deprecating the Stripe RR platform, I've consolidated all of the syncs that are triggered on Woo-based hooks in `includes/reader-revenue/woocommerce/class-woocommerce-connection.php` into each ESP's Data Events connector class instead. There's a lot of duplicate code between them, so there's future opportunity to optimize this better so that the ESP-specific connectors extend a more generic interface, but no need to do that now.

This consolidation resulted in a bunch of cleanup of the data being synced upon new orders, subscriptions, etc. We now properly distinguish between Registration Page and Payment Page meta fields (the former should only be populated and synced upon registration, not on recurring orders or login). And I tested all of the above scenarios and confirmed that UTM fields are syncing for both MC and AC. 

Most of these changes are in e08a654, in case you want to review those separately from the prior changes.

adekbadek:
Again this did not work for me for non-donation WC checkout flow. I'm testing by buying a WC product starting from the product page (`/product/<slug>/?utm_source=marketinglist&utm_medium=email&utm_campaign=2023-Newspack-Campaign`), then the site is reloaded with the "added to cart" message – at this point the UTM params are stripped. Maybe that's the issue?

includes/data-events/listeners.php
```diff
@@ -281,6 +279,54 @@ function ( $order_id, $order ) {
 	}
 );
 
+/**
+ * For non-donation purchases.
+ */
+Data_Events::register_listener(
+	'woocommerce_order_status_completed',
+	'product_order_new',
```
leogermani, line 287: suggestion: call the event `product_order_completed`

leogermani: I know we used `donation_new` but that actually created a problem, because when we wanted to differentiate new donations from renewal donation payments, the name was taken (donation_new fires for new donations and renewals).

Keeping it complete makes it more tied to what exactly this is and leaves room for other events we might want to add in the future

leogermani:
> It turns out, we had a hole in our Data Events listeners where we haven't been sending any data events upon a non-subscription, non-donation-type purchase

Hmmm sorry about that. @miguelpeixe and I knew about it. It was intentional. We took an approach of just adding listeners we were actively using for something.

It definitely makes sense to have the one you created. Not 100% sure if we should remove Donations from it though... I know there will be some overlap with `donation_new` but that may be a good thing. You will only need to listen to one event for data sync.

And in the cases we want to filter and listen only to Donations, there is an event specific for that.

In the future, we might create other events with subsets of purchases, that will have specific uses (e.g. the discussion about "Memberships/Subscriptions" being another category of special products like Donations are)

> Lastly, I'm not sure if we need to update anything in the Newspack Network plugin after this, but I suspect we will?

We might. Network currently registers its own listeners because they were not present in the main plugin. Maybe we should remove them from there once we have everything we need here.

https://github.com/Automattic/newspack-network/blob/9e370127caef28382ef9811ea5b26fe241719037/includes/class-data-listeners.php#L36-L37

The listeners there are even more generic, because we want to capture every status change of both orders and subscriptions. Not sure if we want that here or not.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -256,7 +227,7 @@ public static function get_contact_from_order( $order, $payment_page_url = false
 			}
 			$metadata[ Newspack_Newsletters::get_metadata_key( 'total_paid' ) ] = (float) $customer->get_total_spent();
 
-			if ( 'pending' === $order->get_status() ) {
+			if ( $is_new && 'pending' === $order->get_status() ) {
 				$metadata[ Newspack_Newsletters::get_metadata_key( 'total_paid' ) ] += (float) $order->get_total();
 			}
```
miguelpeixe, line 232: Could we change this to query the customer orders and always calculate the total from the database? It helps making this method context-agnostic and prevents that a single misuse of the method breaks the "Total Paid" data forever because it'll just add to a wrong number.

miguelpeixe:
> > It turns out, we had a hole in our Data Events listeners where we haven't been sending any data events upon a non-subscription, non-donation-type purchase
>
> Hmmm sorry about that. @miguelpeixe and I knew about it. It was intentional. We took an approach of just adding listeners we were actively using for something.

Indeed, I believe one of the actions has a misleading name: `woocommerce_order_failed`. This is only triggered for donations; although the doc block specifies this, the action name should also reflect that to avoid confusion.

Other action names are prefixed with `donation_` because we are intentionally only supporting donations.

We do have listeners for non-donation products for the following:

- product_subscription_active
- product_subscription_inactive

These are also documented in the [README.md](https://github.com/Automattic/newspack-plugin/blob/master/includes/data-events/README.md) file.

dkoo:
> Again this did not work for me for non-donation WC checkout flow. I'm testing by buying a WC product starting from the product page (`/product/<slug>/?utm_source=marketinglist&utm_medium=email&utm_campaign=2023-Newspack-Campaign`), then the site is reloaded with the "added to cart" message – at this point the UTM params are stripped. Maybe that's the issue?

@adekbadek that explains it—the non-modal checkout flow (via Cart and separate Checkout page) doesn't support those values. Since the use of the non-modal checkout flow is an edge case for our sites, let's bookmark that and handle it in a separate PR. We'll need to make sure to handle the referer URL differently in that scenario.

> It definitely makes sense to have the one you created. Not 100% sure if we should remove Donations from it though... I know there will be some overlap with `donation_new` but that may be a good thing. You will only need to listen to one event for data sync.
> 
> And in the cases we want to filter and listen only to Donations, there is an event specific for that.
> 
> In the future, we might create other events with subsets of purchases, that will have specific uses (e.g. the discussion about "Memberships/Subscriptions" being another category of special products like Donations are)

@leogermani this is a good suggestion—I'll amend the event to handle new orders of any type, donation or not. That way, `donation_new` is only used for analytics purposes and not for syncing.

dkoo:
> We do have listeners for non-donation products for the following:
> 
> - product_subscription_active
> - product_subscription_inactive

@miguelpeixe that's true, but those only cover subscription-type products, not one-off purchases. And the data payload for those events is different, so I think there's value in a more generic "order completed" event for all completed orders that we use to trigger syncs, as Leo suggests.

dkoo:
@leogermani @adekbadek @miguelpeixe 8afd1007 simplifies the data event listeners as discussed above. It also tweaks some other listeners to make them a bit more consistent:

- `product_order_new` is now `order_completed` and executes after a successfully completed order of ANY type (one-time, recurring, donation, and non-donation). It fires with the same data as `donation_new` so it's a one-to-one replacement and is now the only event that triggers an ESP contact sync when a purchase is completed.
- The `product_subscription_active` and `product_subscription_inactive` event listeners have been consolidated into a single `product_subscription_changed` listener. It still only fires for non-donation products, but it now mirrors the `donation_subscription_changed` listener. In the future, we might consider consolidating these two listener like we did with the `order_completed` listener.
- The handlers for `donation_subscription_changed` and `product_subscription_changed` in both MC and AC have been updated so that they don't attempt to sync to the ESP if the subscription is changing status from "pending" or "on hold" to "active". In those cases, the status change indicates that the subscription has completed a successful order, which will trigger a sync via the new `order_completed` event, so there's no need to sync again based on this subscription status change. If the subscription status is changing to any inactive status or from a `pending-cancel`, `cancelled`, or `expired` status, we still want to trigger a sync.
- Speaking of MC, it was missing handlers to sync on subscription status update, so I added those.

includes/data-events/listeners.php
```diff
@@ -253,20 +253,18 @@ function ( $order_id, $order ) {
...
 			return;
 		}
-		$recurrence      = get_post_meta( $product_id, '_subscription_period', true );
-		$is_renewal      = function_exists( 'wcs_order_contains_renewal' ) && wcs_order_contains_renewal( $order );
-		$subscription_id = null;
-		if ( function_exists( 'wcs_get_subscriptions_for_renewal_order' ) ) {
-			$subscriptions   = array_values( wcs_get_subscriptions_for_renewal_order( $order ) );
-			$subscription_id = is_array( $subscriptions ) && ! empty( $subscriptions ) && is_a( $subscriptions[0], 'WC_Subscription' ) ? $subscriptions[0]->get_id() : null;
-		}
+		$recurrence       = \get_post_meta( $product_id, '_subscription_period', 'once' );
```
miguelpeixe, line 256: The 3rd argument of post meta is not the default value. It's a bool for whether to return a single value or not.

includes/data-events/README.md
```diff
@@ -130,6 +130,24 @@ When there's a new donation, either through Stripe or Newspack (WooCommerce) pla
...
+
+When any WooCommerce order is completed with a successful payment.
+
+| Name              | Type     | Obs                                                    |
+| ----------------- | -------- | ------------------------------------------------------ |
+| `user_id`         | `int`    |                                                        |
+| `email`           | `string` |                                                        |
+| `amount`          | `float`  |                                                        |
+| `currency`        | `string` |                                                        |
+| `recurrence`      | `string` |                                                        |
```
miguelpeixe, line 143: Given that the `recurrence` data is only available for donation orders, I suggest we don't add it to this action to avoid confusion. We can leave `donation_new` and the other existing donation-related actions to provide this data.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
dkoo: It is actually calculated each time from the customer's "total spent" value, on [this line](https://github.com/Automattic/newspack-plugin/pull/2821/files/13ae6b9afd2bb4857eb946e7d668545cb2eac5c6#diff-b1dbb7bb086f4afe9581e4edecca68795028891e7d380656342dc169973ad5a5L257). However, this value won't include the amount from the current pending order, so we still have to add that manually. In this PR I've [added a new condition](https://github.com/Automattic/newspack-plugin/pull/2821/files/13ae6b9afd2bb4857eb946e7d668545cb2eac5c6#diff-b1dbb7bb086f4afe9581e4edecca68795028891e7d380656342dc169973ad5a5R230) so that this only happens for brand-new subscriptions or one-time orders.

miguelpeixe: Aah, thank you for explaining. I missed that previous line that creates the value.

includes/data-events/README.md
dkoo: Sounds good—removed in 74b8fb80

includes/data-events/listeners.php
dkoo: Incidentally fixed by removing `recurrence` from the payload in 74b8fb80 (see https://github.com/Automattic/newspack-plugin/pull/2821#discussion_r1450851895)

miguelpeixe: It should also be reverted for the `donation_new` listener.

dkoo: Ugh, sorry 6570674e

dkoo:
> We might. Network currently registers its own listeners because they were not present in the main plugin. Maybe we should remove them from there once we have everything we need here.
> 
> https://github.com/Automattic/newspack-network/blob/9e370127caef28382ef9811ea5b26fe241719037/includes/class-data-listeners.php#L36-L37
> 
> The listeners there are even more generic, because we want to capture every status change of both orders and subscriptions. Not sure if we want that here or not.

If the more generic `item_changed` listeners do the job in Newspack Network because it needs to capture any status change for anything, then I don't see a need to refactor based on the new listeners in this PR. The flexibility of data events should allow for creating very specific listeners for very specific needs, right?

adekbadek (approved):
> For now, can you test purchasing non-donation products via a Checkout Button block?

With this interaction, it works as described 👍

includes/data-events/listeners.php
dkoo: Updated ([details](https://github.com/Automattic/newspack-plugin/pull/2821#issuecomment-1888116911))

PR #2837, opened 2024-01-05
fix: prevent switch order synchronisation

adekbadek:
### All Submissions:

When a subscription is switched to a different amount, WC creates a "switch order", which is an order of 0 value. This order should not be synchronised to third parties, as it's not a real order.  

1. On `master`,
1. Allow subscription switching in WC Settings -> Subscriptions
2. Ensure logging is enabled (`NEWSPACK_LOG_LEVEL` set to `2`)
3. Run through the process of starting a donation subscription and updating its amount*
5. Observe in the logs that events are being dispatched, most importantly and ESP contact update
6. Switch to this branch, repeat, observe that the events are not being dispatched**
7. Make another recurring donation, observe events being dispatched as before

\* It's a confusing process – go to /my-account page, Subscriptions, then click "Change price", update the price, proceed to cart to "buy" the 0-value item
\** With the Newspack Network plugin active, `newspack_node_order_changed` event will be the only one dispatched

### Other information:

---
  - https://app.asana.com/0/0/1206209020023382

### DIFF
--- a/includes/data-events/class-memberships.php
+++ b/includes/data-events/class-memberships.php
@@ -229,7 +229,10 @@ final class Memberships {
 	 */
 	public static function woocommerce_checkout_order_processed( $order_id ) {
 		$order = \wc_get_order( $order_id );
-		$data  = self::get_order_data( $order_id, $order );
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Works as described!

PR #2845, opened 2024-01-10
feat(ras): better reader display/nice names

dkoo:
### All Submissions:

Tweaks and standardizes the logic that generates default user display names (updatable by the user), nicenames (the slug used for author archive pages, which shouldn't affect reader accounts unless they're upgraded to an author/editor/admin role), and login strings (never updatable) when registering a new RAS reader account. This only affects newly registered accounts going forward, not preexisting accounts unless they make a purchase and provide a billing name to reset the display name.

Also adds support for a new environment constant, mainly for dev convenience, but could also be used by publishers if the behavior is desired. If `NEWSPACK_ALLOW_MY_ACCOUNT_ACCESS_WITHOUT_VERIFICATION` is defined and true, My Account will allow logged-in users to access and update account settings without requiring verification via email.

Another new environment constant `NEWSPACK_ALLOW_GENERIC_READER_DISPLAY_NAMES` allows generated display names to be shown in My Account, if the publisher desires.

Two new filters allow for further customization of behavior on a per-site basis:

- `newspack_register_reader_user_data` filters the user data array right before the reader account is created by [`wp_insert_user`](https://developer.wordpress.org/reference/functions/wp_insert_user/)
- `newspack_myaccount_required_fields` filters the fields required for saving account details in My Account. Currently we only show and require Display Name and Email, but this could theoretically be changed via a custom plugin using this filter plus a custom `form-edit-account.php` template.

1. Check out this branch.
2. In a new session, register a reader via the Registration block or auth modal (which accepts an email address only). Note the user ID of the generated account for the next step.
3. Run `wp shell` and then `get_userdata( $user_id )`. Confirm that the user's data looks like the following:
  - `user_login`: Matches the user's email address.
  - `user_nicename`: A URL-sanitized version of the user's email address, minus the domain (everything up to the `@`). Unfortunately this can't be updated once the user has been created.
  - `display_name`: Matches the `user_nicename`.

4. Log into this user's account and confirm that the "Display Name" field here renders blank (the generated display name isn't shown). Confirm that if you try to click "save changes" without filling out Display Name, you get the "Display Name is a required field" error (no reason not to require it for saving since it's the only field we show—this might even encourage people to enter a true value). 

Screenshot 2024-01-15 at 3 22 11 PM

5. In the same session, complete a new purchase with billing names (to test updating an existing user account upon purchase). Then go to My Account, verify if needed, and confirm that the display name is shown with the billing name value, and that you can update it to something else, as long as it's not empty or the same as the reader's email address—if the latter, you should see an error message like this:

Screenshot 2024-01-15 at 3 12 57 PM

6. Still in the same session while in My Account, enter a "generic" display name (such as the email address minus the `@domain` part) and save, and confirm that it's now shown for this user. If the the reader intentionally saves a display name that matches our generated ones, we add a user meta value that tells us this so we can treat this value as the reader's preferred display name. 
7. In a new session, complete a purchase including billing first name + last name (to test new user registration upon purchase).
8.  Run `wp shell` and then `get_userdata( $user_id )`. Confirm that the user's data looks like the following:
  - `display_name`: Matches the user's full name (billing first + last names).
  - `user_login`: Matches the user's email address.
  - `user_nicename`: A URL-sanitized version of the user's display name (e.g. `firstname-lastname`).
9. Go to My Account, verify as needed, and confirm that you can see and update the display name.

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -26,11 +26,11 @@ final class Reader_Activation {
 	/**
 	 * Reader user meta keys.
 	 */
-	const READER                          = 'np_reader';
-	const EMAIL_VERIFIED                  = 'np_reader_email_verified';
-	const WITHOUT_PASSWORD                = 'np_reader_without_password';
-	const REGISTRATION_METHOD             = 'np_reader_registration_method';
-	const READER_HAS_GENERIC_DISPLAY_NAME = 'np_reader_has_generic_display_name';
+	const READER                            = 'np_reader';
+	const EMAIL_VERIFIED                    = 'np_reader_email_verified';
+	const WITHOUT_PASSWORD                  = 'np_reader_without_password';
+	const REGISTRATION_METHOD               = 'np_reader_registration_method';
+	const READER_SAVED_GENERIC_DISPLAY_NAME = 'np_reader_saved_generic_display_name';
 
 	/**
 	 * Unverified email rate limiting
@@ -1756,6 +1756,7 @@ final class Reader_Activation {
 				if ( empty( $userdata['display_name'] ) ) {
 					// If the reader lacks a display name, generate one.
 					$data['display_name'] = self::generate_user_nicename( $userdata['user_email'] );
+					\delete_user_meta( $user_id, self::READER_SAVED_GENERIC_DISPLAY_NAME );
 				} else {
 					// Otherwise, don't update it.
 					$data['display_name'] = $userdata['display_name'];
@@ -1767,7 +1768,7 @@ final class Reader_Activation {
 				self::generate_user_nicename( $userdata['user_email'] ) === $data['display_name'] || // New generated construction (URL-sanitized version of the email address minus domain).
 				self::strip_email_domain( $userdata['user_email'] ) === $data['display_name'] // Legacy generated construction (just the email address minus domain).
 			) {
-				\update_user_meta( $user_id, self::READER_HAS_GENERIC_DISPLAY_NAME, 1 );
+				\update_user_meta( $user_id, self::READER_SAVED_GENERIC_DISPLAY_NAME, 1 );
 			}
 		}
 
@@ -1830,7 +1831,7 @@ final class Reader_Activation {
 		}
 
 		// If the reader has intentionally saved a display name we consider generic, treat it as not generic.
-		if ( \get_user_meta( $user_id, self::READER_HAS_GENERIC_DISPLAY_NAME, true ) ) {
+		if ( \get_user_meta( $user_id, self::READER_SAVED_GENERIC_DISPLAY_NAME, true ) ) {
 			return false;
 		}
 
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -81,7 +81,7 @@ final class Reader_Activation {
 			\add_action( 'init', [ __CLASS__, 'setup_nav_menu' ] );
 			\add_action( 'wc_get_template', [ __CLASS__, 'replace_woocommerce_auth_form' ], 10, 2 );
 			\add_action( 'template_redirect', [ __CLASS__, 'process_auth_form' ] );
-			\add_filter( 'woocommerce_new_customer_data', [ __CLASS__, 'sanitize_user_data' ] );
+			\add_filter( 'woocommerce_new_customer_data', [ __CLASS__, 'canonize_user_data' ] );
 			\add_filter( 'amp_native_post_form_allowed', '__return_true' );
 			\add_filter( 'woocommerce_email_actions', [ __CLASS__, 'disable_woocommerce_new_user_email' ] );
 			\add_filter( 'retrieve_password_notification_email', [ __CLASS__, 'password_reset_configuration' ], 10, 4 );
@@ -1620,7 +1620,7 @@ final class Reader_Activation {
 			/**
 			 * Create new reader.
 			 */
-			$user_data = self::sanitize_user_data(
+			$user_data = self::canonize_user_data(
 				[
 					'display_name' => $display_name,
 					'user_email'   => $email,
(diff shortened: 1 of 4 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -1699,6 +1690,41 @@ public static function register_reader( $email, $display_name = '', $authenticat
...
+	 *
+	 * @param array $user_data          Default args for the new user.
+	 *              $user_data['email] Email address for the new user (required).
+	 */
+	public static function get_user_data( $user_data = [] ) {
+		if ( empty( $user_data['user_email'] ) ) {
+			return $user_data;
+		}
+
+		// If we don't have a display name, make it match the email address.
```
adekbadek, line 1705: When making a donation via the checkout modal, the `$user_data' passed in the call to this filter does not include `display_name`, which results in email address being used as the base for `display_name`. For this reason, the second testing scenario cannot be fulfilled.

adekbadek (changes requested):

yogeshbeniwal:
> * `display_name`: Matches the user's email address

Probably not a good idea to use email address for display name, may be displayed to other readers, e.g. when user comment. Can we make it as `user_nicename` A URL-sanitized version of the user's email address, minus the domain (everything up to the @)

includes/reader-activation/class-reader-activation.php
dkoo: It's true that the [`woocommerce_new_customer_data` filter](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-user-functions.php#L96-L107) doesn't include `display_name`. However, if a new customer account is created during checkout, the checkout method to do so [adds the display name based on billing first/last name](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-checkout.php#L1161) if available. So the effect is the same.

This condition is actually here for the benefit of [our own `register_reader` method](https://github.com/Automattic/newspack-plugin/pull/2845/files#diff-7f238c1a510d794797749876c3ac59626a1672f70f67e0dc41873ac9337f9793R1623-R1628), which calls `wc_create_new_customer` outside of the checkout. Our method is called to register readers via the Register and Newsletter Subscription Form blocks, and the latter includes optional name fields which might be passed and used to build the display name if available.

dkoo:
Thanks, @yogeshbeniwal—we discussed internally and agree. d9470fb changes things so that if we don't have a name to use for display name, it uses the same URL-sanitized string as `user_nicename`, generated from the email address. It also adds two additional features for user-friendliness:

- If the reader's display name is a generated string based on the email address, we no longer show or require it in the My Account dashboard (most sites have no use for display name at this time).
- If the reader's display name was generated and they subsequently complete a purchase in which they provide an actual billing name, we automatically upgrade the display name to match the provided billing name.

Testing instructions updated.

includes/reader-activation/class-reader-activation.php
```diff
@@ -1699,6 +1690,96 @@ public static function register_reader( $email, $display_name = '', $authenticat
...
 	}
 
+	/**
+	 * Get user data args for creating a new WP user.
+	 * See https://developer.wordpress.org/reference/functions/wp_insert_user/ for supported args.
+	 *
+	 * @param array $user_data          Default args for the new user.
+	 *              $user_data['email] Email address for the new user (required).
+	 */
+	public static function get_user_data( $user_data = [] ) {
```
miguelpeixe, line 1700: NIT: Can we rename this to `sanitize_user_data()`?

adekbadek: NIT: NIT:  I'd rather opt for ["canonize"](https://en.wikipedia.org/wiki/Canonicalization), since data is added, not only sanitized

adekbadek (commented):
Works as described in principle, apart from this step:

> Log into this user's account and confirm that you no longer see the "Display Name" field here.
 
At this point, the "Display Name" field is present and required, as expected looking at the code (since the name is not generic). Is a step missing before this one?

Also, this feature feels odd to me. If a user registers using email only, they can't set a display name late. Why is that?

dkoo:
> At this point, the "Display Name" field is present and required, as expected looking at the code (since the name is not generic). Is a step missing before this one?

Whoops, I put those steps in the wrong order. Updated!

> Also, this feature feels odd to me. If a user registers using email only, they can't set a display name late. Why is that?

I agree with this. It's the result of a very long internal discussion about how currently, very few sites actually use reader Display Names for anything, and because of that, in most cases it's preferable to not show readers their display names at all if we can't show something meaningful. Perhaps that means we should implement a constant to override this behavior for sites that do want to allow readers to see and update their display names as usual.

I'm thinking of this as a temporary improvement for reader experience, that will require revisiting during the next project that reconsiders the overall My Account dashboard design.

includes/reader-activation/class-reader-activation.php
dkoo: Renamed to `canonize_user_data` in 6e3f10c8

dkoo:
> Perhaps that means we should implement a constant to override this behavior for sites that do want to allow readers to see and update their display names as usual.

Added support for a `NEWSPACK_ALLOW_GENERIC_READER_DISPLAY_NAMES` constant in 341931bc. Setting this constant to `true` will restore the legacy (current, pre-PR) behavior and allow all display names to be seen and edited by readers.

dkoo:
> If a user registers using email only, they can't set a display name later

Hmm, another "easy" solution to this would be to show the Display Name field in My Account, but as optional and with a blank/empty value if their display name is calculated to be generic/generated. 🤔

dkoo:
Ok, 9cbd601 updates the behavior so that I think this improves the reader experience in a way that both avoids showing them a generated display name, and also lets all readers update their display name if desired. Testing instructions updated.

adekbadek (changes requested):
In my testing, Display Name was never set from billing fields (First Name, Last Name)  I've tested these flows: 

- register -> donate -> verify
- register -> verify -> donate
- donate 

In every case, the Display Name in the end is generated from the email address.

dkoo:
> In every case, the Display Name in the end is generated from the email address.

Aha, I think I figured out why. I was setting the `np_reader_saved_generic_display_name` user meta too liberally, so it was getting set upon initial registration if the account was created without billing name first. 63fa6de should fix this for all of these scenarios.

adekbadek (approved):
I've re-tested the three scenarios from my previous review and it's all working as expected now 👌

PR #2848, opened 2024-01-12
feat: add cli command to fix stripe covered fees

leogermani:
### All Submissions:

Since https://github.com/Automattic/newspack-plugin/pull/2820 we use an actual woocommerce Fee to add the Stripe fee to the order. Before that, we would manually add the fee to the order total if a metadata was present.

We didn't take into account the need to update existing subscriptions so that renewal orders will also get the fee added. (we assumed they would copy from the Sub's parent order, but they don't).

This PR adds a CLI command to migrate the subscriptions that had fees added the old way to the new way.

It's a CLI command because it should affect few subscriptions in a few sites.

1. Checkout the plugin in the version prior to #2820 `git checkout v2.12.0`
2. Make sure the option for the donor to cover fees is enabled
3. Make a recurring donation covering Stripe fees
4. Checkout master or this branch
5. Force the subscription to renewal and verify the total of the renewal order does not include the fee
6. Run the commands in this PR to and confirm the subscription is listed to be fixed, and that the fix is applied
7. `wp newspack-woocommerce list_subscriptions_missing_fee` should list the subscription
8. `wp newspack-woocommerce fix_subscriptions_missing_fee --dry-run` should output what the fix will look like
9.  `wp newspack-woocommerce fix_subscriptions_missing_fee` should fix it
10. Go back to the admin, see the subscription, verify that the fee is there and the total is updated
11. Force a renewal and confirm the renewal order has the fee
12. Also test a new recurring subscription on this branch with fees covered, and its renewal orders, to make sure we didn't break anything

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
@@ -1,0 +1,218 @@
+<?php
+/**
+ * CLI tools for the woocommerce support
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+use WP_CLI;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WooCommerce Order UTM class.
+ */
+class WooCommerce_Cli {
+
+	/**
+	 * Lists the subscriptions that needs to be fixed by the fix_subscriptions_missing_fee command.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--format=<format>]
+	 * : Accepted values: table, csv, json, count, yaml. Default: table
+	 *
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
Working flawlessly.

NIT suggestion: maybe add a meta to the subscription so it's easier to track which of them got migrated? Just in case we need to revisit them. The order note works too, but it's more of a UI feature.

leogermani:
Thank you! Added post mea in https://github.com/Automattic/newspack-plugin/pull/2848/commits/8a4bd001cedc6feceed3d3246ee28b0896246c01 - tested again and it's all fine

PR #2849, opened 2024-01-15
hotfix(optional-modules): ensure the switch to media-partners can happen

adekbadek:
### All Submissions:

https://github.com/Automattic/newspack-plugin/pull/2753 introduced a new optional module. The module should be switched on if the corresponding plugin is active, and then the plugin should be deactivated. Because of a logical error, though, that would never happen, because the is-module-active check happens before the potential switch. This PR fixes that. 

Also, an unneeded check for `is_plugin_active` function existence is removed. Since the method is called on `admin_init` hook, this won't be needed. 

1. Follow instructions in https://github.com/Automattic/newspack-plugin/pull/2753

### Other information:

### DIFF
--- a/includes/optional-modules/class-media-partners.php
+++ b/includes/optional-modules/class-media-partners.php
@@ -18,17 +18,13 @@ class Media_Partners {
 	 * Initialize everything.
 	 */
 	public static function init() {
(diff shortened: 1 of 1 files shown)


adekbadek:
Closing, this code will be removed.

PR #2859, opened 2024-01-16
feat(newspack-ui): modal functionality

miguelpeixe:
### All Submissions:

Implements modal functionality to the Newspack UI.

https://github.com/Automattic/newspack-plugin/assets/820752/50b8bff1-408d-49cf-9796-9d1448db2fc7

The functionality is applied to a `.newspack-ui__modal-container`, which will condition the inner `.newspack-ui__modal`. The open state is determined by the `data-state="open"` attribute. We could go with something like a `newspack-ui__modal-container--open` class as well. Not sure which is best, it's a matter of style.

This proposal does not implement shared js logic, and I don't think we should, at least for now. When using the modal, each functionality should apply its own JS to change the attribute for toggling the modal.

1. Checkout this branch and visit your site with the `?ui-demo` query param
2. Scroll down to the "Open Modal" button and click
3. Confirm it appears as in the video above
4. Click to close and confirm it closes

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -22,7 +22,8 @@
 			z-index: 2;
 			width: 100%;
 			transform: translateY( 50px );
-			transition: transform 0.1s linear;
+			opacity: 0;
+			transition: transform 0.1s linear, opacity 0.1s linear;
 		}
 		&[data-state='open'] {
 			z-index: 99999;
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
I really like how this UI elements are coming together <3

laurelfulford (approved):
@miguelpeixe this is just delightful! 😄 I like the approach you took, and it's fun to play with! I think it makes sense to stick with just moving the styles for now.

PR #2862, opened 2024-01-17
feat: button styles tweaks; use BEM consistently

adekbadek:
### All Submissions:

1. Fixes a modal close button styling issue:

| Before | After | 
|---|---|
| image | image |

2. Updates CSS class names to follow [BEM naming convention](https://css-tricks.com/bem-101/). The code _looked like_ BEM, but was not. 
3. Updates button styles to apply only to elements with the `newspack-ui__button` class names, not all buttons. The latter approach was partially a cause of the modal close button issue, and may have created more similar issues down the road. If we control the elements which will be rendered as part of this UI, styling using class selectors only is a more future-proof approach. 

1. Open the UI demo (`<site-url>?ui-demo`)
4. Observe no changes vs. `epic/rass-acc` branch apart from the fixed modal close button

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/elements/forms/_buttons.scss
+++ b/assets/newspack-ui/scss/elements/forms/_buttons.scss
@@ -1,16 +1,12 @@
 .newspack-ui {
 	// TODO: define some kind of size hierachy for buttons eg. small, medium, large;
 	// not needed for RAS, but likely for My Account.
-	.button,
-	button,
-	input[type='button'],
-	input[type='reset'],
-	input[type='submit'] {
+	&__button {
 		align-items: center;
 		border: 0;
 		border-radius: var( --newspack-ui-border-radius-small );
 		display: inline-flex;
-		font-family: var( --newspack-ui-font-family);
+		font-family: var( --newspack-ui-font-family );
 		font-size: var( --newspack-ui-font-size-small );
 		font-weight: 600;
 		gap: var( --newspack-ui-spacer-2 );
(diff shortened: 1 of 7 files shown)


assets/newspack-ui/scss/elements/forms/_buttons.scss
```diff
@@ -1,16 +1,12 @@
 .newspack-ui {
 	// TODO: define some kind of size hierachy for buttons eg. small, medium, large;
 	// not needed for RAS, but likely for My Account.
-	.button,
-	button,
-	input[type='button'],
-	input[type='reset'],
-	input[type='submit'] {
+	&__button {
```
miguelpeixe, line 4: I think it'd be good to have a second opinion from @laurelfulford about this change. I assume these selectors help enforce the button style on different scenarios, especially Woo-related. Maybe we move that to the `woocommerce` directory when necessary?

Personally, I like this change.

miguelpeixe (approved):
Thank you for enforcing good BEM practices! Styles are being applied without issues.

Left one non-blocking comment that can be readjusted later if necessary. Let's merge this so the UI integration can happen with the updated class names.

assets/newspack-ui/scss/elements/forms/_buttons.scss
adekbadek: To minimise surprise, we should be selective with styles. If the button styles should apply to elements not rendered by Newspack (e.g. Woo), I'd argue for a whitelist approach – explicitly apply the styles to Woo elements.

PR #2865, opened 2024-01-19
fix(ras): handle invalid order sync request

dkoo:
### All Submissions:

Fixes a fatal that can happen if `\Newspack\WooCommerce_Connection::sync_reader_from_order()` is called without a valid `$order` arg.

1. Run `wp shell` and then `\Newspack\WooCommerce_Connection::sync_reader_from_order( false )`. Observe a fatal error
2. Check out this branch, repeat, and observe no error.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -368,7 +368,7 @@ class WooCommerce_Connection {
 	 * @param bool|string $payment_page_url Payment page URL. If not provided, checkout URL will be used.
 	 */
 	public static function sync_reader_from_order( $order, $verify_created_via = true, $payment_page_url = false ) {
-		if ( ! self::can_sync_customers() ) {
+		if ( ! $order || ! self::can_sync_customers() ) {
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -368,7 +368,7 @@ public static function should_sync_order( $order ) {
 	 * @param bool|string $payment_page_url Payment page URL. If not provided, checkout URL will be used.
 	 */
 	public static function sync_reader_from_order( $order, $verify_created_via = true, $payment_page_url = false ) {
-		if ( ! self::can_sync_customers() ) {
+		if ( ! $order || ! self::can_sync_customers() ) {
```
leogermani, line 371: isn't it safer to check if $order is an instance of WC_Order?

dkoo: Yes, you're right. ee65f29c updates this and also moves the check to the existing `should_sync_order()` validator method.

leogermani (approved):

PR #2866, opened 2024-01-19
feat(reader-revenue): make NYP and Stripe Gateway optional

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-blocks/pull/1645, removes requirements for both WooCommerce Name Your Price and WooCommerce Gateway Stripe plugins. The Newspack Reader Revenue platform and Donation/Checkout Button blocks can now be used without these plugins, as long as all products have a set price and at least one [payment gateway valid for subscriptions](https://woo.com/document/subscriptions/payment-gateways/) is enabled.

Closes `1206098184615535/1206328415142056`.

1. Before checking out this branch, make sure you have Reader Revenue fully set up with the Newspack platform and a page containing several Donate blocks:
  - Frequency-based Donate block, not manually configured
  - Frequency-based Donate block, manually configured and set to show tiered suggestions
  - Tiers-based Donate block
2. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1645. Before doing anything else, smoke test Reader Revenue wizard pages and the Donate blocks in both the editor and on the front-end—all should function as usual.
3. Deactivate the WooCommerce Name Your Price and WooCommerce Gateway Stripe plugins.
4. In WP admin, visit **Newspack > Reader Revenue**. Confirm that in the Platform, Stripe Gateway, and Donation platforms, you no longer see a plugin requirement message for these plugins. Make sure Stripe remains disabled in Stripe Gateway. **Note:** you may want to enable another non-Stripe payment gateway in WooCommerce > Settings > Payments so that you can still complete test transactions. "Check" payments will work in a pinch.
5. In The Donations page, confirm that the settings page no longer allows you to choose between Tiered and Untiered mode, and that the settings are locked in Untiered mode (no NYP = no tiers because basic Woo can't modify the product pricing on the fly like with NYP).
6. Update the suggested values here and hit "Save Settings". Visit the **Products** admin page and confirm that the set prices for each Donate product are updated according to your settings (this isn't the case in `master`, but is important when not using NYP as all products must have set prices in order to be purchasable without NYP).
7. Edit the page containing your Donate blocks from step 1. Confirm that:
  - The Tiers-based block has reverted to a frequency-based layout
  - All blocks no longer have the "Configure Manually" option or sub-options (manual settings are not applicable when NYP isn't available)
  - The manually-configured block has reverted to the same settings and prices as the non-manual block
  - The inputs are no longer editable but simply display as static/flat prices, e.g.:

Screenshot 2024-01-19 at 12 33 13 PM

8. View the page on the front-end and confirm that the front-end blocks look the same as in the editor.
9. Test all three block styles (Default, Alternate, and Minimal) and confirm they all work in both the editor and on the front-end.
10. On the front-end, complete a donation using one of the blocks. Note that you won't be able to purchase subscriptions unless you enable one of the [supported payment gateways](https://woo.com/document/subscriptions/payment-gateways/), but you should be able to complete a one-time purchase with a Check payment.

### Other information:


miguelpeixe (commented):
Tested well on my local instance and on a fresh JN site.

Non-blocking, it would be nice to have a message in the "Stripe settings" tab if the Stripe gateway is not enabled. Possibly even disable the form, since it's not an available feature without the gateway plugin.

There's also a failing test that appeared after rebasing.

chickenn00dle (approved):
Tested and looks good @dkoo 

Just one observation that is not a blocker. Now that stripe is no longer a required plugin, I'm wondering why we need to keep the stripe settings menu in the RR wizard?

Screenshot 2024-01-29 at 15 17 57

dkoo:
Thanks, @chickenn00dle!

> Now that stripe is no longer a required plugin, I'm wondering why we need to keep the stripe settings menu in the RR wizard?

Great question! Stripe in this context actually means the [Stripe payment gateway for WooCommerce](https://wordpress.org/plugins/woocommerce-gateway-stripe/). Since it's still the overwhelming dominant payment processor for Woo sites, we offer some important settings for it here. This might be relabelled and moved for clarity in a future project to reorganize the admin screens.

dkoo:
> Just one observation that is not a blocker. Now that stripe is no longer a required plugin, I'm wondering why we need to keep the stripe settings menu in the RR wizard?

It's still the most widely used payment processor, and given that we have an IA project going to reconsider the organization of settings pages, I think we should leave this page as-is for now.

PR #2877, opened 2024-01-23
RAS UI: add primary colour to modal buttons

laurelfulford:
### All Submissions:

This PR does a couple things: 

1. It fixes a couple small issues in the demo content CSS classes (just minor stuff I noticed on the buttons to do with font sizing/spacing/hover states).
2. It adds the CSS variables `--newspack-ui-primary-color` and `--newspack-ui-color-against-primary`
3. It sets the default values for those variables to the ones in the classic Newspack theme. 
4. It also checks for a block theme, and if one exists, tries to set those values to match the element/button colours (background and text); by default this should be the equivalent of the primary colour, but if not, will at least be a contrasting background/text combo we can use on the site. 
5. It then uses the `--newspack-ui-primary-color` and `--newspack-ui-color-against-primary` CSS variables for the modal buttons.

See 1205234045751551-as-1206301643351944

1. To start, make sure the Newspack Theme is up to date on your test site; [it recently had a PR merged](https://github.com/Automattic/newspack-theme/pull/2109) that's required for this to work. 
2. Apply this PR, and view the /?ui-demo page of your test site. 
3. Review the button styles, focusing on the colours used by the "modals"; the buttons should use the default dark grey outside of the modals, and use the theme's primary colour inside of the modals.

image

4. Navigate to Customizer > Colors, and change the primary colour; confirm it still works:

image

This next bit is a little out of scope for what we actually need right now, but I wanted to make sure we could actually do something like this with a block theme, too. Also, this approach may not be the way we want to go for getting the custom colours from this theme!

5. Switch the theme [to this theme](https://github.com/automattic/newspack-block-theme). 
6. Repeat the testing steps 2 and 3 and confirm that the regular buttons are using the dark grey, and the buttons inside of the modal are using the primary colour (it should be dark blue by default). Note: there are some display issues around `<button>` and `<input>`, but they can be sorted out later. 
7. Navigate to Edit Site > Styles (if you drop into the Templates or Front Page section, you may need to hit the back arrow next to the heading to get to the 'Design' header, where the Styles live).

Screen Shot 2024-01-23 at 9 48 08 AM

8. Under Styles, click 'Colors', then change the button colour: 

image

11. Repeat steps 2 and 3, and confirm that this new colour is being picked up. This is specifically the colour set for the `button` element, which is picked up by the button blocks (at the moment we don't define a separate colour for those blocks, but we could). While the button may not always be the "primary" colour, this will make sure the modal buttons are using a colour from the site, with text with adequate contrast.

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -39,6 +39,11 @@
 		}
 	}
 	&__modal {
+		// CSS Variables - update button background, with fallbacks to the Newspack UI defaults
+		--newspack-ui-color-button-bg: var( --newspack-ui-color-primary, var( --newspack-ui-color-gray-900 ) );
+		--newspack-ui-color-button-text: var( --newspack-ui-color-against-primary, var( --newspack-ui-color-white ) );
+
+		// Modal styles
 		background: var( --newspack-ui-color-body-bg );
 		border-radius: var( --newspack-ui-border-radius );
 		display: flex;
--- a/assets/newspack-ui/scss/variables/_colors.scss
+++ b/assets/newspack-ui/scss/variables/_colors.scss
@@ -21,8 +21,8 @@
 	--newpsack-ui-color-alert-red-bg: #fcf0f1;
 
 	// Theme variables - from the classic theme; overridden in class-newspack-ui.php for block theme.
-	--newspack-ui-color-primary: var(--newspack-theme-color-primary);
-	--newspack-ui-color-against-primary: var(--newspack-theme-color-against-primary);
+	--newspack-ui-color-primary: var( --newspack-theme-color-primary );
+	--newspack-ui-color-against-primary: var( --newspack-theme-color-against-primary );
 
 	// Specific assignments - general:
 	--newspack-ui-color-text-high-contrast: var( --newspack-ui-color-gray-900 );
--- a/assets/newspack-ui/scss/variables/_colors.scss
+++ b/assets/newspack-ui/scss/variables/_colors.scss
@@ -20,6 +20,10 @@
 	--newspack-ui-color-alert-green-bg: #edfaef;
 	--newpsack-ui-color-alert-red-bg: #fcf0f1;
 
+	// Theme variables - from the classic theme; overridden in class-newspack-ui.php for block theme.
+	--newspack-ui-color-primary: var(--newspack-theme-color-primary);
+	--newspack-ui-color-against-primary: var(--newspack-theme-color-against-primary);
+
 	// Specific assignments - general:
 	--newspack-ui-color-text-high-contrast: var( --newspack-ui-color-gray-900 );
 	--newspack-ui-color-text-med-contrast: var( --newspack-ui-color-gray-700 );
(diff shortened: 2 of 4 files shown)


assets/newspack-ui/scss/_modals.scss
```diff
@@ -39,6 +39,11 @@
 		}
 	}
 	&__modal {
+		// CSS Variables - update button background, with fallbacks to the Newspack UI defaults
+		--newspack-ui-color-button-bg: var( --newspack-ui-color-primary, var( --newspack-ui-color-gray-900 ) );
+		--newspack-ui-color-button-text: var( --newspack-ui-color-against-primary, var( --newspack-ui-color-white ) );
```
miguelpeixe, line 44: Nice!

assets/newspack-ui/scss/variables/_colors.scss
```diff
@@ -20,6 +20,10 @@
 	--newspack-ui-color-alert-green-bg: #edfaef;
 	--newpsack-ui-color-alert-red-bg: #fcf0f1;
 
+	// Theme variables - from the classic theme; overridden in class-newspack-ui.php for block theme.
+	--newspack-ui-color-primary: var(--newspack-theme-color-primary);
+	--newspack-ui-color-against-primary: var(--newspack-theme-color-against-primary);
```
miguelpeixe, line 25: NIT

```suggestion
	--newspack-ui-color-primary: var( --newspack-theme-color-primary );
	--newspack-ui-color-against-primary: var( --newspack-theme-color-against-primary );
```

includes/class-newspack-ui.php
```diff
@@ -33,6 +37,24 @@ public static function enqueue_styles() {
...
 	}
 
+	/**
+	 * Adds inline styles CSS for the element/button colors from the theme.json.
+	 * See: https://developer.wordpress.org/reference/functions/wp_get_global_styles/
+	 */
+	public static function colors_css_wrap() {
+		$global_styles = wp_get_global_styles();
+
+		$custom_css = 'body {';
```
miguelpeixe, line 47: Other variables are under `:root`. Any specific reason why these are under `body`?

miguelpeixe (approved):
Working great! 🎉 

Couple of non-blocking nits below

includes/class-newspack-ui.php
laurelfulford: It's the only way I could get it to work for some reason 😄 `:root` alone (or `:root` plus `!important`) for these styles didn't do the trick; I tried inserting them in a bunch of different locations in the header, but it didn't seem to make a difference. 

My guess is it works because `body` is a hair more specific than `:root`, so it's overriding the defaults; when `:root` was used, I was still getting the value set in the `_colors.scss` file.

This might have to do with how I'm inserting these styles to begin with! I tried a few different things in `<style>` tags, but maybe the cascade would be more logical if they were also enqueued in a separate file. We can definitely experiment! 🙂

assets/newspack-ui/scss/variables/_colors.scss
laurelfulford: Good catch -- thanks!! 🤦‍♀️ Fixed in [d2a61cf](https://github.com/Automattic/newspack-plugin/pull/2877/commits/d2a61cf99d642c06727dd0b1aad20b1a4d8a8a76).

laurelfulford:
Thanks @miguelpeixe! I've fixed a nit, and am going to go ahead and merge this but we can definitely refine/experiment with the approach to the block theme colours 🙂 I mostly wanted to make sure we could get the custom colours and apply them somehow.

includes/class-newspack-ui.php
miguelpeixe: Aah, I see! Thank you for explaining 🙇

PR #2881, opened 2024-01-25
fix: ensure WC session is present if triggering WC notice

adekbadek:
### All Submissions:

Adds an additional check when triggering a WC notice. 

When testing, an error with the following stack trace was encountered:

```
PHP Fatal error:  Uncaught Error: Call to a member function get() on null in /srv/htdocs/wp-content/plugins/woocommerce/includes/wc-notice-functions.php:80
Stack trace:
#0 /srv/htdocs/wp-content/plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php(728): wc_add_notice('Thank you for v...', 'success')
#1 /srv/htdocs/wp-content/plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php(1684): Newspack\Reader_Activation::set_reader_verified(3)
```

The `get()` mentioned is called on `WC()->session`. In order to prevent this issue, all calls to `wc_add_notice` are now wrapped in a `WooCommerce_Connection::add_wc_notice` method, which will check for validity of `WC()->session`. 

I could not reproduce the issue locally, but it appeared on a JN test site. 

1. Use Google to register with a RAS-configured site, observe that the reader registration works as expected
2. In a different session, create a new account (without SSO, simply with an email) and then verify the account
3. After following the verification link, observe the WC notice displayed:

image

### Other information:

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -369,7 +369,7 @@ final class Recaptcha {
 		$token = isset( $_POST['g-recaptcha-response'] ) ? sanitize_text_field( wp_unslash( $_POST['g-recaptcha-response'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
 		$check = self::verify_captcha( $token );
 		if ( \is_wp_error( $check ) ) {
-			\wc_add_notice( $check->get_error_message(), 'error' );
+			WooCommerce_Connection::add_wc_notice( $check->get_error_message(), 'error' );
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


laurelfulford (commented):
Thanks @adekbadek! Reviewing from a functional perspective, everything looks good! 

* I tried signing up though Google on a new JN site and no longer see the original issue.
* I was able to successfully sign up normally with an email address
* I got the success message after verifying my second account
* I didn't run into any issues deleting both accounts

Leaving as a comment for now just in case you'd like some sharper eyes on the code before merging, but it's a 👍 from me!

PR #2905, opened 2024-02-05
fix: fixes for prod release feb 5

dkoo:
### All Submissions:

Closes # .

**For the Organic Profile issue:**

Installing this version of the plugin on a JN site caused a fatal error:

```
PHP Fatal error: Uncaught Error: Call to undefined function Newspack\is_plugin_active() in /srv/htdocs/wp-content/plugins/newspack-plugin/includes/plugins/class-organic-profile-block.php:30
```

To fix, we replaced the `is_plugin_active` check for one that looks for a function from the plugin. To test:

1. Set up a JN site with this PR; confirm you don't get any fatal errors/errors on the front-end. 
2. Install and activate copy of the [Profile Block](https://wordpress.org/plugins/organic-profile-block/).
3. Add a profile to a post or page and publish.
4. Deactivate the plugin.
5. Confirm that your profile block is still styled (for example, the fonts match how they looked when the plugin was active, and the profile image is aligned left). 

**For the reader registration issue:**

1. On `alpha`, register a new account via the Registration block or auth modal.
2. Observe a PHP warning upon account registration: `PHP Warning:  Undefined array key "random_password" in /newspack-repos/newspack-plugin/includes/reader-activation/class-reader-activation.php on line 1641`
3. Check out this branch, repeat, and confirm the warning doesn't happen.

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1638,7 +1638,7 @@ final class Reader_Activation {
 				 * Create WooCommerce Customer if possible.
 				 * Email notification for WooCommerce is handled by the plugin.
 				 */
-				$user_id = \wc_create_new_customer( $email, $user_data['user_login'], $user_data['random_password'], $user_data );
+				$user_id = \wc_create_new_customer( $email, $user_data['user_login'], $user_data['user_pass'], $user_data );
 			} else {
 				$user_id = \wp_insert_user( $user_data );
 				\wp_new_user_notification( $user_id, null, 'user' );
--- a/includes/plugins/class-organic-profile-block.php
+++ b/includes/plugins/class-organic-profile-block.php
@@ -17,6 +17,10 @@ class Organic_Profile_Block {
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
+		if ( ! function_exists( 'is_plugin_active' ) ) {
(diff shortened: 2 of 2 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -1638,7 +1638,7 @@ public static function register_reader( $email, $display_name = '', $authenticat
 				 * Create WooCommerce Customer if possible.
 				 * Email notification for WooCommerce is handled by the plugin.
 				 */
-				$user_id = \wc_create_new_customer( $email, $user_data['user_login'], $user_data['random_password'], $user_data );
+				$user_id = \wc_create_new_customer( $email, $user_data['user_login'], $user_data['user_pass'], $user_data );
```
adekbadek, line 1641: Is this change intended? I can't see how it relates to PR's description.

dkoo: It was an initial attempt at addressing the fatal described in the description, but turned out to be a red herring. However, I think it's still not a bad practice to use the `\` namespace for WP globals since the file itself is also namespaced. Just for code readability/clarity.

dkoo: Whoops, looked at the wrong line. This change is intended, yes. I'll update the testing instructions

leogermani (approved):

PR #2913, opened 2024-02-08
fix(categories): fix pager urls

naxoc:
### All Submissions:

* [x] Have you followed the [[Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [[WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [[VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [[Pull Requests](https://github.com/Automattic/newspack-plugin/pulls)](../../pulls) for the same update/change?

Because WP allows a url that shouldn't really exist when a site has the `/%category%/%postname%`/ permalink structure, the pager links break. See https://core.trac.wordpress.org/ticket/8905 and https://core.trac.wordpress.org/ticket/21209

This PR tries to help out the pager only on sites with that specific url structure.

**Before applying the patch:**

1. On a test site, set the permalink structure to `/%category%/%postname%`
2. Create a category (or use an existing one). It needs to have enough posts that it gets a pager. Let's say it's called `kitten`.
3. Go to https://your-site.local/kitten This page shouldn't really exist, but it does. Now hover the pager or click "2". You will be taken to `/kitten/page/2` and the page 404s.

**Now apply the patch**

1. Go to `/kitten`
2. Hover over or navigate to an item in the pager.
3. Verify that you were taken to a page that does not 404 and is correct :)
4. Try changing the category base and verify that the pager still works.


includes/class-category-pager.php
```diff
@@ -0,0 +1,68 @@
...
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'on_init' ] );
+	}
+
+	/**
+	 * Action callback for init.
+	 */
+	public static function on_init() {
+		if ( '/%category%/%postname%/' === get_option( 'permalink_structure' ) ) {
```
leogermani: I've ran some tests and the issue also occurs with other settings, for example `/%category%/%post_id%/`. 

It does not happen if you do `/%post_id%//%category%` though.

Didn't go too far to get at the bottom of why this happens, but I think we can assume that it happens if the permalink structure starts with `category`?

includes/class-category-pager.php
```diff
@@ -0,0 +1,68 @@
...
+	 * Initialize Hooks.
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'on_init' ] );
+	}
+
+	/**
+	 * Action callback for init.
+	 */
+	public static function on_init() {
```
leogermani: WDYT of hooking directly into `paginate_links` and checking the `permalink_structure` option there?

I think it's better to have that check done only when really needed than on every single request.

leogermani (changes requested):
It works well and fixes the issue! I left two comments

includes/class-category-pager.php
naxoc: Good catch! I changed it to sniff if the permalink structure starts with `/%category%/`.

includes/class-category-pager.php
naxoc: Makes much better sense. I changed it. Thanks

naxoc:
This is ready for re-review. I also caught that the `category_base` can be empty (and `category` is then assumed by WP but not returned in the option, so I'm setting a default now.

adekbadek:
(merged in `trunk` so that CI jobs pass)

leogermani:
I think this is good to go. @naxoc do you want to turn this into a hotfix so we can release it immediately? Or can it wait the regular release flow?

If you want to make it a hotfix, we need to rebase it against the release branch, and prefix the branch name with `hotfix/`. (Sorry! I should probably have told you this earlier)

leogermani (approved):

PR #2917, opened 2024-02-09
fix(GA): update spelling of referrer param

chickenn00dle:
### All Submissions:

Closes `0/1206331646877116/1206412046494468/f`.

1. Make sure your site is [connected to GA](https://github.com/Automattic/newspack-plugin/tree/trunk/includes/data-events/connectors/ga4#setting-up)
2. Trigger a user registration event by registering for a new user by donating as a new unsubscribed user
3. In GA reports > realtime tab, look for the `event count by event name` card
4. Look for the `np_reader_registered` and `np_donation_new` events
5. Clicking each of these, confirm you see a `referrer` field. For example:
Screenshot 2024-02-09 at 14 18 23
6. There are two other events which pass along a referrer param to GA: `np_newsletter_subscribed` and `np_gate_interaction`. Verify these as well by triggering their respective actions as a subscriber
7. Have a look through the code changes to verify nothing important was missed. Note that some instances of `referer` will remain in the code as some wordpress functions use this spelling.

Note: I'm not sure if it will be a problem for publishers if we completely remove the `referer` field in favor of `referrer`. If it will be, I can re-add `referer` alongside `referrer`.

### Other information:

---
  - https://app.asana.com/0/0/1206412046494468


leogermani:
It's ok to search and replace all internal usages, but we need to be mindful about external apis and integrations.

For example, GA4 events should keep both spellings imo, so we don't break reports people might have created looking at that field.

Also, we need to check other plugins in our suite that could be affected by this change when they interact with this plugin's hooks, and the other way around as well

chickenn00dle:
Thanks for looking at this @leogermani 🙇 

> For example, GA4 events should keep both spellings imo, so we don't break reports people might have created looking at that field.

Makes sense. I've re-added the old spelling to GA4 events in https://github.com/Automattic/newspack-plugin/pull/2917/commits/59d4fc2e51554a6446ac764d3421fde2c676717f

> but we need to be mindful about external apis and integrations.

Makes sense. Looking through the changes, I'm not sure how exposed they are to external integrations. That said, I did go ahead and add a utility function to normalize order referral meta whenever its fetched from the order in https://github.com/Automattic/newspack-plugin/pull/2917/commits/1b26974ba470c9a34634419034604e3cee76368e just in case this ends up being a problem. Note that I didn't do this in the data event listeners since these seem to only be associated with new events, but I could be wrong.

>  we need to check other plugins in our suite that could be affected by this change when they interact with this plugin's hooks, and the other way around as well

Will do! I'm wondering if you have any thoughts on which of our extensions I should be extra careful about while doing this?

chickenn00dle:
> I'm wondering if you have any thoughts on which of our extensions I should be extra careful about while doing this?

I had a look through what I believe are some of our core extensions for cases of GA4 events, and am only seeing an instance of this in popups which doesn't seem to rely on the main plugin for analytics. 

I may have missed some though, so let me know if there are any extensions you had in mind.

leogermani:
I'm reading through the changes and I don't feel confident to move forward. So many moving parts.

For example, where the `DONATION_ORDER_META_KEYS` are used? this stores additional data in the orders, where is this data consumed? And so many other things it's very hard to parse and test.

When I said we needed to "check other plugins" I was not referring to them using the GA4 analytics, but the Data events api as a whole, and maybe some of these other data we write here and there. (example, is [this](https://github.com/Automattic/newspack-blocks/blob/trunk/includes/class-modal-checkout.php#L154) related?)

I would feel more confidence if we kept this change to address only what the issue talks about - the data that goes to GA4.

So no need to touch anything other than `Newspack\Data_Events\Connectors\GA4`. There, see all the events that are processed and add an additional param with the correct spelling, while keep the wrong spelling as well. This is 100% safe and addresses the issue

chickenn00dle:
Just noting I've addressed Leo's feedback and this is now awaiting review again!

adekbadek (approved):

PR #2925, opened 2024-02-13
feat(rss): enable control over feed update frequency and guid

adekbadek:
### All Submissions:

Adds two controls in a custom RSS feed's settings:

- update frequency
- whether to use post ID as the `guid`

1. Enable the RSS Feeds optional module if not active *
2. Create or edit a feed, observe two new settings, "Update frequency" and "Use post ID as the guid instead of post URL": 

image

4. Change values of the first setting, open the feed in a browser, observe the `sy:updatePeriod` and `sy:updateFrequency` values are changed accordingly
5. Change value of the latter setting, observe the `guid` value on a single feed item is updated from post URL to post ID

\* `\Newspack\Settings::activate_optional_module('rss')`

### Other information:


includes/optional-modules/class-rss.php
```diff
@@ -505,6 +539,34 @@ public static function modify_feed_query( $query ) {
...
+			add_filter(
+				'rss_update_period',
+				function() use ( $settings ) {
+					return $settings['update_frequency'][0];
+				}
+			);
+			add_filter(
+				'rss_update_frequency',
+				function() use ( $settings ) {
+					return $settings['update_frequency'][1];
```
chickenn00dle, line 555: For the case of 'daily', explode will return an array with only one item.

chickenn00dle (changes requested):

includes/optional-modules/class-rss.php
adekbadek: Good catch! Fixed in 25156bb3406b34849e46ee39aa0db003291edfed

chickenn00dle (approved):
Looks good!

PR #2930, opened 2024-02-14
fix: add frequency tab options for donations, even when tiers are disabled

laurelfulford:
### All Submissions:

This change comes out of feedback for the alpha release for this change: https://github.com/Automattic/newspack-plugin/pull/2866 

If you don't install Name Your Price, or you turn off the Tiered view of the donate block's site-wide settings under Newspack > Reader Revenue > Donations, you can no longer change the frequency amount site-wide. Instead, it's sticks to whatever the last saved selection was when the Tiered view was visible. 

This is not a new issue, but is more noticeable with us supporting not using Name Your Price: with that plugin, you can still switch back to the Tiered view and change the default frequencies. Without it, you can't.

I'm not entirely sold on the UI decisions I made here, but this PR adds the frequency toggles to the untiered view. I'm definitely open to other suggestions about fixing this both for the short and long term! 

See pamTN9-8My-p2-#comment-10416

2. Navigate to Newspack > Reader Revenue > Donations.
3. Toggle the view to 'Untiered', and confirm that you now see frequency toggles above each field:

image

4. Try toggling different options on and off. Confirm that the associated field is hidden when the frequency is 'off'. 
5. Try toggling between the Tiered and Untiered view, and make changes to the frequency settings on each. Confirmed they're carried over, even when the page is saved and refreshed. 
6. With the Untiered view selected and one or more frequencies is disabled, add a Donate block to a page, and confirm the initial appearance matches your selection (for example, if you've disabled the One-Time donation option, it should not appear in the block by default): 

image

7. Turn back on the Tiered view under Reader Revenue > Donations, and change the selection again and save it.
8. Try disabling the Name Your Price plugin, and confirm your active frequencies remain the same.
9. Change the frequencies again and save. 
10. Try adding another Donation block to a page, and confirm the visible tabs match your frequency choices. 

### Other information:

### DIFF
--- a/assets/wizards/readerRevenue/views/donation/index.tsx
+++ b/assets/wizards/readerRevenue/views/donation/index.tsx
@@ -208,24 +208,44 @@ export const DonationAmounts = () => {
 			) : (
 				<Card isMedium>
 					<Grid columns={ 3 } rowGap={ 16 }>
-						{ availableFrequencies.map( section => (
-							<MoneyInput
-								currencySymbol={ currencySymbol }
-								label={ section.staticLabel }
-								value={ amounts[ section.key ][ 3 ] }
-								min={ minimumDonationFloat }
-								error={
-									amounts[ section.key ][ 3 ] < minimumDonationFloat
-										? __(
-												'Warning: suggested donations should be at least the minimum donation amount.',
-												'newspack'
-										  )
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
Looks good!

PR #2935, opened 2024-02-16
Feat/tweaks for network update

adekbadek:
### All Submissions:

A few tweaks for the upcoming changes to `newspack-network`, and one fix:

1. Extricates a `get_reader_roles` method in Reader Activation
2. Adds metadata to webhook request posts, so they can be found later (and avoid duplication)
3. Fixes an issue where `subscription_id` was set only on renewal orders, not on the initial order of a subscription

_Follow instructions in https://github.com/Automattic/newspack-network/pull/51_

### Other information:

### DIFF
--- a/includes/data-events/class-utils.php
+++ b/includes/data-events/class-utils.php
@@ -14,9 +14,10 @@ final class Utils {
 	/**
 	 * Get order data.
 	 *
-	 * @param int $order_id Order ID.
+	 * @param int  $order_id Order ID.
+	 * @param bool $process_donations_only Whether to process only donation orders.
 	 */
-	public static function get_order_data( $order_id ) {
+	public static function get_order_data( $order_id, $process_donations_only = false ) {
 		$order = \wc_get_order( $order_id );
 		if ( ! $order ) {
 			return;
@@ -24,8 +25,12 @@ final class Utils {
 		if ( ! \Newspack\WooCommerce_Connection::should_sync_order( $order ) ) {
 			return;
 		}
-		// Donation orders always have just a single product, but other orders can have more than one.
-		$product_id = \Newspack\Donations::get_order_donation_product_id( $order_id ) ?? array_values( \Newspack\WooCommerce_Connection::get_products_for_order( $order_id ) );
+		if ( $process_donations_only ) {
+			$product_id = \Newspack\Donations::get_order_donation_product_id( $order_id );
+		} else {
+			// Donation orders always have just a single product, but other orders can have more than one.
+			$product_id = array_values( \Newspack\WooCommerce_Connection::get_products_for_order( $order_id ) );
+		}
 		if ( ! $product_id ) {
 			return;
 		}
@@ -60,11 +65,11 @@ final class Utils {
 	}
 
 	/**
-	 * Get subscription data.
+	 * Get recurring donation data.
 	 *
-	 * @param WC_Subscription $subscription Subscription.
+	 * @param WC_Subscription $subscription Subscription which is a recurring donation.
 	 */
-	public static function get_subscription_data( $subscription ) {
+	public static function get_recurring_donation_data( $subscription ) {
 		$product_id = \Newspack\Donations::get_order_donation_product_id( $subscription->get_id() );
 		if ( ! $product_id ) {
 			return;
--- a/includes/data-events/class-utils.php
+++ b/includes/data-events/class-utils.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * Newspack Data Events Utils.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Data_Events;
+
+/**
+ * Main Class.
+ */
+final class Utils {
+	/**
+	 * Get order data.
+	 *
+	 * @param int $order_id Order ID.
+	 */
+	public static function get_order_data( $order_id ) {
+		$order = \wc_get_order( $order_id );
+		if ( ! $order ) {
+			return;
+		}
+		if ( ! \Newspack\WooCommerce_Connection::should_sync_order( $order ) ) {
+			return;
+		}
+		// Donation orders always have just a single product, but other orders can have more than one.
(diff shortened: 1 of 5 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -679,7 +686,7 @@ public static function is_user_reader( $user, $strict = false ) {
 			 *
```
leogermani: Move the filter doc along with the filter

includes/data-events/class-utils.php
```diff
@@ -0,0 +1,83 @@
...
+		];
+	}
+
+	/**
+	 * Get subscription data.
+	 *
+	 * @param WC_Subscription $subscription Subscription.
+	 */
+	public static function get_subscription_data( $subscription ) {
+		$product_id = \Newspack\Donations::get_order_donation_product_id( $subscription->get_id() );
```
leogermani, line 73: we are using this for any kind of subscription, so we should not check and return nothing if it's not a donation subscription. I think this check should still be kept in the listerers

Note that we still need to fetch the product ID to be able to get the recurrence

includes/data-events/class-utils.php
```diff
@@ -0,0 +1,83 @@
...
+		$order = \wc_get_order( $order_id );
+		if ( ! $order ) {
+			return;
+		}
+		if ( ! \Newspack\WooCommerce_Connection::should_sync_order( $order ) ) {
+			return;
+		}
+		// Donation orders always have just a single product, but other orders can have more than one.
+		$product_id = \Newspack\Donations::get_order_donation_product_id( $order_id ) ?? array_values( \Newspack\WooCommerce_Connection::get_products_for_order( $order_id ) );
+		if ( ! $product_id ) {
```
leogermani, line 34: Make sure it works for any order and not only for donations

includes/data-events/listeners.php
```diff
@@ -134,37 +134,7 @@ function( $provider, $email, $lists_added, $lists_removed, $result ) {
...
-			'referer'         => $order->get_meta( '_newspack_referer' ),
-			'popup_id'        => $order->get_meta( '_newspack_popup_id' ),
-			'is_renewal'      => $is_renewal,
-			'subscription_id' => $subscription_id,
-			'platform_data'   => [
-				'order_id'   => $order_id,
-				'product_id' => $product_id,
-			],
-		];
+		return \Newspack\Data_Events\Utils::get_order_data( $order_id );
```
leogermani: here is where we want to check if this is a donation related order or not

includes/data-events/listeners.php
```diff
@@ -178,35 +148,7 @@ function( $order_id, $product_id ) {
...
-			'popup_id'        => $order->get_meta( '_newspack_popup_id' ),
-			'is_renewal'      => $is_renewal,
-			'subscription_id' => $subscription_id,
-			'platform_data'   => [
-				'order_id'   => $order_id,
-				'product_id' => $product_id,
-				'client_id'  => $order->get_meta( NEWSPACK_CLIENT_ID_COOKIE_NAME ),
-			],
-		];
+		return \Newspack\Data_Events\Utils::get_order_data( $order_id );
```
leogermani: same as above. Here is where we should check if it's a donation.

The event name is misleading btw

includes/data-events/listeners.php
```diff
@@ -246,36 +188,7 @@ function( $subscription, $status_to, $status_from ) {
...
```
leogermani: check donation

includes/data-events/listeners.php
```diff
@@ -325,20 +213,7 @@ function( $subscription, $status_to, $status_from ) {
...
-		return [
-			'user_id'         => $subscription->get_customer_id(),
-			'email'           => $subscription->get_billing_email(),
-			'subscription_id' => $subscription->get_id(),
-			'amount'          => (float) $subscription->get_total(),
-			'currency'        => $subscription->get_currency(),
-			'recurrence'      => empty( $recurrence ) ? 'once' : $recurrence,
-			'platform'        => Donations::get_platform_slug(),
-		];
+		return \Newspack\Data_Events\Utils::get_subscription_data( $subscription );
```
leogermani: again a misleading event name. But the check to see if it's a donation should happen here, not in the Utils method

includes/data-events/listeners.php
```diff
@@ -349,22 +224,17 @@ function( $subscription, $status_to, $status_from ) {
 	'woocommerce_subscription_status_updated',
 	'donation_subscription_changed',
 	function( $subscription, $status_to, $status_from ) {
-		$product_id = Donations::get_order_donation_product_id( $subscription->get_id() );
-		if ( ! $product_id ) {
+		$data = \Newspack\Data_Events\Utils::get_subscription_data( $subscription );
```
leogermani: same

leogermani (changes requested):

includes/data-events/class-utils.php
adekbadek: That's what happens here – if the donation product ID is not found, the fallback is `\Newspack\WooCommerce_Connection::get_products_for_order`.

includes/data-events/class-utils.php
adekbadek: > we are using this for any kind of subscription

No, this is only for recurring donations (`donation_subscription_changed`, `donation_subscription_cancelled`) listeners. `product_subscription_changed` listener has its own data processing below.

adekbadek:
Addressed the is-donation-order/subscription comments in 85a624b042e0960878e4b683b7d22f3d524e30e2.

includes/reader-activation/class-reader-activation.php
adekbadek: Done in 2eb469d

includes/data-events/class-utils.php
leogermani: Let's rename these methods then? `...donation_subscriptions...` ? or even `get_recurring_donations_data` ?

leogermani: oops, I commented before looking at the changes :)

leogermani (approved):

PR #2936, opened 2024-02-16
feat(blocks): add meta to core/image blocks

jaredrethman:
### All Submissions:

- Added hooks to allow `show_in_rest` post-meta to be added as attributes to `core/image` blocks.
- Added styling for visual feedback during media requests:
   - Individual `image/core` blocks have a gradient animation and `<Spinner />` overlay.
   - "Update" post button is locked, pending media request resolution
- Added fallback logic to credit html to use block attributes if empty and `meta` attribute exists.
- Registered attachment credit post-meta.

These change are necessary for user who use plugins like [Distributor](https://github.com/10up/distributor) and are unable to display credit meta on pulling sites. This PR aims to solve this by adding it the core/image blocks attributes

Asana: 1205909204837811-as-1206457195610270/f

1. Ensure you have [Distributor](https://github.com/10up/distributor) installed and active
2. You have two sites connected using Distributors External Connections
5. Ensure you have an image (Media Library) where credit post-meta has been added

Screenshot 2024-02-16 at 16 42 10

6. Create a post, or edit existing one, and add above image to the block editor (`post_content`, then click Save/Update. Once saved, click "Distribute" post.

Screenshot 2024-02-16 at 16 57 42

8. Confirm credit is on both the source & destination sites for in-content attachments.

Screenshot 2024-02-16 at 17 00 50

### Other information:


assets/blocks/core-image/index.js
```diff
@@ -0,0 +1,83 @@
...
+				},
+			},
+		};
+	}
+);
+
+/**
+ * Display spinner and lock post saving until meta attributes are added to block
+ */
+const AttributesLoader = props => {
```
adekbadek: With `woocomerce` plugin active, this component causes an error to be thrown:

image

assets/blocks/core-image/index.js
```diff
@@ -0,0 +1,83 @@
...
+	const { meta } = useSelect( select => select( 'core' ).getMedia( imageId ), [ imageId ] ) ?? {
+		meta: {},
+	};
+	// Meta added, proceed
+	if ( Object.keys( meta ).length ) {
+		props.setAttributes( { meta } );
+		unlockPostSaving( 'attachment-meta-empty' );
+	} else {
+		lockPostSaving( 'attachment-meta-empty' );
+	}
```
adekbadek: These calls to `unlockPostSaving` and `lockPostSaving` happen on each render, which is not optimal. That code block should happen in a `useEffect` hook, only called if the value of `meta` has changed.

assets/blocks/core-image/index.js
```diff
@@ -0,0 +1,83 @@
...
+addFilter(
+	'editor.BlockEdit',
+	'newspack-plugin/block-edit-hook/core-image',
+	createHigherOrderComponent( BlockEdit => {
+		const blockEditComponent = props => {
+			if ( props.name === 'core/image' ) {
+				return (
+					<div className="newspack-block__core-image">
+						<BlockEdit { ...props } />
+						<AttributesLoader { ...props } />
```
adekbadek: Suggestion: render the credits below here, so it's a more WYSIWYG experience.

adekbadek (changes requested):

assets/blocks/core-image/index.js
jaredrethman: Nice catch, fixed.

assets/blocks/core-image/index.js
jaredrethman: This was automatically fixed with https://github.com/Automattic/newspack-plugin/pull/2936/commits/3142c27b72185fa0e54677b27abba52bd86791c2

assets/blocks/core-image/index.js
jaredrethman: Nice suggestion, this did cross my mind, I decided against adding any net-new functionality to try stay within scope. However, I've come up with a simple way of doing this without significant code additions.

See - [feat(blocks): make image credit display in editor](https://github.com/Automattic/newspack-plugin/pull/2936/commits/d4b9138c27875f62c936e3d2e940c27a47b7bf71)

Screenshot 2024-02-19 at 10 21 23

Look forward to your feedback.

assets/blocks/core-image/index.js
```diff
@@ -0,0 +1,113 @@
...
+	'editor.BlockEdit',
+	'newspack-plugin/block-edit-hook/core-image',
+	createHigherOrderComponent( BlockEdit => {
+		const blockEditComponent = props => {
+			if ( props.name === 'core/image' ) {
+				return (
+					<div
+						className="newspack-block__core-image"
+						style={ {
+							'--wp-block-image-caption-credit': `" ${ parseCreditToText(
```
adekbadek: This will only work if the "Caption" block attribute is toggled on: 

image

Feel free to drop this since as you've mentioned it's not part of requirements. Otherwise, the "Caption" attribute should be programmatically toggled on as part of these changes.

assets/blocks/core-image/index.js
```diff
@@ -0,0 +1,113 @@
...
+			},
+		};
+	}
+);
+
+/**
+ * Display spinner and lock post saving until meta attributes are added to block
+ */
+const AttributesLoader = props => {
+	const imageId = props.attributes.id;
```
adekbadek: If lacking `imageId`, this component should return `null`. Otherwise, the spinner will obscure the image choosing UI:

image

assets/blocks/core-image/index.js
```diff
@@ -0,0 +1,113 @@
...
+				</div>
+			) }
+		</>
+	);
+};
+
+/**
+ * Helper to parse credit meta
+ *
+ * @param {{}} credit Credit meta to parse
```
adekbadek: Sidenote: using TypeScript is preferred to JSDoc comments. But no harm in adding those.

adekbadek (changes requested):

assets/blocks/core-image/index.js
jaredrethman: @adekbadek I have recreated the caption mechanism/components to mimic current behavior + include credits.
Look forward to your feedback

assets/blocks/core-image/index.js
jaredrethman: Nice catch, and rookie mistake not testing this on my part. This has been fixed 527e83b4b0169ab3cf8c4a9cac1f2c55e56a9b59

assets/blocks/core-image/index.js
jaredrethman: Thanks for pointing this out. I've gone ahead and migrated all the code to TS (I'm a huge advocate). I hope thats okay.

adekbadek: That's great!

assets/blocks/core-image/loader.tsx
```diff
@@ -0,0 +1,50 @@
...
+	const { lockPostSaving, unlockPostSaving } = useDispatch( 'core/editor' );
+	const { meta }: { meta: AttributesMeta } = useSelect(
+		select => select( 'core' ).getMedia( imageId ),
+		[ imageId ]
+	) ?? {
+		meta: {},
+	};
+	useEffect( () => {
+		if ( Object.keys( meta ).length ) {
+			setAttributes( { meta } );
```
adekbadek: Sorry, this slipped my previous review: not _all_ meta values should be saved. We're only interested in the credit-related metadata. Otherwise, the block attributes will be polluted with unneeded meta:

image

assets/blocks/core-image/index.js
adekbadek: I appreciate the effort, but recreating the whole toolbar is overcomplicating the code and not solving the issue, unfortunately 😬  With these changes, the credit will be visible on the front-end regardless of caption, but in the editor it will be visible only if the caption is toggled on. What I meant in my suggestion was that the  block attribute should be uneditable if the credit is set – this way the presentation is the same in the editor and on the front-end.

adekbadek (changes requested):
Found one new small issue, and then there's this – https://github.com/Automattic/newspack-plugin/pull/2936#discussion_r1497346572 – to be resolved.

assets/blocks/core-image/index.js
jaredrethman: @adekbadek thanks for the feedback. The issue (the premise of this PR) was to ensure credits display on distributed (using 10up/distributor) posts. This is done and works 100%, if you feel differently please share your feedback concerning this specifically.

Since the issue was resolved we've moved onto your recommendation i.e. "WYSIWYG". Let me try and explain what was necessary to achieve this.

Current functionality:
* Frontend: If credit exists, it will always display
* Frontend: If caption exists and has been enabled on block, it will always display
* Editor: `image/core` displays a caption if it exists and is toggled on
* Editor: No credit is displayed

The challenge:
`core/image` is a static and dynamic block. Captions are static and since we're unable to provide a `deprecation` mechanism the only way to manipulate how image captions display/behave in the editor is by recreating them and applying customizations to that. I'm unaware of another approach?

The last piece which I discovered this morning was to clear the caption when toggled-off. 1646f7c0bdf51be674e613d19e11f1ca0bd7a11f fixes that.

assets/blocks/core-image/loader.tsx
jaredrethman: @adekbadek This was by design in an effort to eliminate future needs for future attachment meta in distributed posts. 

I have filtered out all meta except for `_media_credit`, `_media_credit_url`, `_navis_media_credit_org` (Required to recreate the credit on distributed posts), in a4f0f53274e74cc46f341dd90f70cc7c342c0bdb

assets/blocks/core-image/index.js
adekbadek: > The issue (the premise of this PR) was to ensure credits display on distributed (using 10up/distributor) posts. This is done and works 100%

Yep, that's all good 👌 

> The challenge

I'm aware that this is challenging and that's why I've proposed to drop the the editor part of this PR. As it currently is, it's not WYSIWYG – the display of the credit is tied to the display of the caption, which is not true for the front-end. This is confusing for the user, less so than the credit appearing on the front-end while missing in the editor.  

The changes to the front-end are enough to satisfy requirements, so if the editor UI part is removed then we're good to go here.

jaredrethman:
> Found one new small issue, and then there's this – [#2936 (comment)](https://github.com/Automattic/newspack-plugin/pull/2936#discussion_r1497346572) – to be resolved.

Can you share a description of the "new small issue", thanks.

assets/blocks/core-image/index.js
jaredrethman: >the display of the credit is tied to the display of the caption, which is not true for the front-end.

With the latest code _this_ is current behavior.

**Has caption and credit:**

(Frontend)
Screenshot 2024-02-21 at 10 19 23

(Backend)
Screenshot 2024-02-21 at 10 29 15


**No caption, has credit**

(Frontend)
Screenshot 2024-02-21 at 10 21 33

(Backend)
Screenshot 2024-02-21 at 10 20 49

Let me know if you would still like me to remove "editor" parts and what those are.

adekbadek:
I was referring to the content of the code review, which was this comment -> https://github.com/Automattic/newspack-plugin/pull/2936#discussion_r1497329552

assets/blocks/core-image/index.js
adekbadek: > With the latest code this is current behavior.

Confirmed! 1646f7c0bdf51be674e613d19e11f1ca0bd7a11f fixed the issue I was describing.

adekbadek (approved):
Re-tested with various combinations of caption/credit, works flawlessly 🙌 Thanks!

PR #2944, opened 2024-02-22
Reverts for epic/network release

adekbadek:
These PRs were meant for an `epic/` release, but were merged to `trunk`:

- https://github.com/Automattic/newspack-plugin/pull/2935
- https://github.com/Automattic/newspack-plugin/pull/2936
- https://github.com/Automattic/newspack-plugin/pull/2940

This PR reverts all three, so they are not included in the current release cycle.


laurelfulford (approved):
Thanks for this, @adekbadek! 🙌 

As a (possibly unnecessary) precaution, I checked the changes against each diff to make sure everything was accounted for. I also did some quick functional testing around donations, sign ups and network syncing with the trunk versions of this and the Newspack Network plugin, and a bit around just using image credits. It all looks good to me!

PR #2951, opened 2024-02-26
fix: set wc default address option when donation country optional but state is required

chickenn00dle:
### All Submissions:

Closes [0/1206331646877121/1205955387558360/f
](https://app.asana.com/0/1206331646877121/1205955387558360/f)

This addresses a bug where, when WC is platform, and donation country is optional, but state is required, the donation checkout field errors with a missing state warning. We fix this by ensuring the `woocommerce_default_customer_address` setting is `base` for this specific configuration.

Screenshot 2024-02-27 at 12 17 13

1. In WooCommerce Settings set `Default customer location` to `no location by default`: WooCommerce > Settings
2. On a test site with donations setup via WooCommerce, go to the donations settings: Newspack > Reader Revenue > Donations
3. In the billing fields section of donations settings, unset country as a required field and set state as a required field (pictured above). This configuration causes a checkout bug where new customers can't get past the billing fields portion since new customers have no default country set and billing states won't populate as expected.
4. Confirm an information notice appears letting you know default store country will be used for new customers
5. Check the WC settings page again and confirm the `Default customer location` is now set to `Shop country/region`

### Other information:

### DIFF
--- a/assets/wizards/engagement/components/types.ts
+++ b/assets/wizards/engagement/components/types.ts
@@ -44,6 +44,7 @@ declare global {
 		};
 		newspack_reader_revenue: {
 			can_use_name_your_price: boolean;
+			is_platform_wc: boolean;
 		};
 	}
 }
--- a/assets/wizards/readerRevenue/views/donation/index.tsx
+++ b/assets/wizards/readerRevenue/views/donation/index.tsx
@@ -295,6 +295,11 @@ const BillingFields = () => {
 		? wizardData.donation_data.billingFields
 		: Object.keys( availableFields );
 
+	const isDefaultBillingCountryForced =
+		window.newspack_reader_revenue?.is_platform_wc &&
+		! billingFields.includes( 'billing_country' ) &&
+		billingFields.includes( 'billing_state' );
+
 	return (
 		<>
 			<Card noBorder headerActions>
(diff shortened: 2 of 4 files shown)


dkoo (approved):
@chickenn00dle this is working as described, so approving. BUT I have a suggestion for a different approach, as the one you've taken (by updating the option value when donation settings are saved) will not fix the user-facing "Billing State is a required field" error during checkout until the publisher re-saves donation settings.

Instead of triggering `maybe_update_woocommerce_default_customer_address` on saving donation settings, what if we filtered the option value (and default option value) of `woocommerce_default_customer_address` and then return the `base` value only if the actual DB value is empty and the donation billing fields have state but no country? Here's an example of where we take this approach to force certain WooCommerce option values in certain scenarios: https://github.com/Automattic/newspack-plugin/blob/trunk/includes/reader-revenue/woocommerce/class-woocommerce-connection.php#L29-L34

Let me know if you think this won't work for addressing this particular bug.

chickenn00dle:
Nice! Thanks for the suggestion @dkoo! My only concern about this approach is it can potentially lead to confusion for any site admins poking around Woo settings and wondering why the default location is not always behaving as it should. 

That said, if this is the general approach we take for this sort of issue, I think we should follow it here. So I've created a new PR for this here: https://github.com/Automattic/newspack-plugin/pull/2960. Mind giving that one a review as well?

I'll go ahead and close this one out.

dkoo:
> My only concern about this approach is it can potentially lead to confusion for any site admins poking around Woo settings and wondering why the default location is not always behaving as it should.

That's a very valid concern and something that we should always be mindful of. One thing we could do is to isolate the filter so that we only force the values when we absolutely need to, i.e. inside the modal checkout window. Otherwise, for standard checkouts where this isn't a concern (because the state + country billing fields will always be present), we should adhere to the standard behavior.

PR #2954, opened 2024-02-28
fix: donate block layout for NRH

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-blocks/pull/1682, fixes a regression introduced by #2866.

1. On `release`, set your Reader Revenue platform to NRH and disable WooCommerce.
2. View a Donate block in the editor and on the front-end.
3. Observe that the Frequency-based Donate block has only a single static option per frequency (no tiers) and no "other" number input. Similar to how it appears with Woo minus NYP:

Screenshot 2024-01-19 at 12 33 13 PM

4. Observe that the Tiers-based Donate block shows "0" in all tiers in the editor.
5. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1682, and confirm that the blocks appear more as expected:

Screenshot 2024-02-27 at 6 22 13 PM

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -127,6 +127,11 @@ class Donations {
 	 * @return bool True if available, false if not.
 	 */
 	public static function can_use_name_your_price() {
+		// If the donation platform is NRH, the Donate block should behave as if Name Your Price is available.
+		if ( self::is_platform_nrh() ) {
+			return true;
+		}
+
 		return class_exists( 'WC_Name_Your_Price_Helpers' );
 	}
 


laurelfulford (changes requested):
I took this for a test run, but not on a site that has NRH connected, just on a site set to use NRH as a donation platform. 

Overall it looks good, but I ran into one issue: in the editor, the tiered layout of the block shows a $0 amount, even though it displays a proper dollar amount on the front end. The frequency view was fine and showed dollar amounts: 

image

dkoo:
> Overall it looks good, but I ran into one issue: in the editor, the tiered layout of the block shows a $0 amount, even though it displays a proper dollar amount on the front end.

@laurelfulford I pushed a fix for this in the other repo: [abf4dd9](https://github.com/Automattic/newspack-blocks/pull/1682/commits/abf4dd97da958d93bbff9de00b74762dd36416c2)

Also applies some CSS so that the look matches the Woo-based block more closely:

Screenshot 2024-02-28 at 3 08 13 PM

dkoo:
Whoops, looks like that last change broke things if the platform is set to NRH, but WooCommerce _is_ active. Fixed that in [7fa0e66](https://github.com/Automattic/newspack-blocks/pull/1682/commits/7fa0e66be31f24d9aa0e628f33aceb8f09cca673).

laurelfulford (changes requested):
Thanks @dkoo! 

I'm just going to keep the feedback on this PR since it's easier to have it all in one place, though I think this is all still related to the block part of the changes. 

Testing, I'm still getting a bit of weirdness in the editor:

* Both on localhost and on the JN site, I'm getting the currency locale (US) in the editor (not on the front-end).
* The buttons are also misaligned in the editor (sorry I missed that earlier!) 

image

I also compared the block to how it used to look, and it's still not quite 1:1: the price should use the heading font as well (not the body) and be a bit bigger/bold; it should also be followed by the frequency amount:

image

When I switch back to Woo/NYP on this branch, some of these issues persist:
* The font is wrong on the prices on the front-end and in the editor
* The currency symbol is missing
* In the editor, the button text is also left-aligned:

Woo/NYP front-end: 
image

Just let me know if you have any questions, or if I can help with any fixes!

laurelfulford (approved):
I re-tested these this PR and the block PR with the following combinations, attempting to check out with each:

* NRH as RR platform; Woo, NYP enabled
* NRH as RR platform; Woo enabled
* NRH as RR platform
* Newspack as RR platform; Woo, NYP enabled
* Newspack as RR platform; Woo enabled

... and compared them to a JN sites running a previous version. Outside of [one small display issue I fixed](https://github.com/Automattic/newspack-blocks/pull/1682/commits/9494ef40cbd52067585b8d127bf5b79e422fd03e) (which existed before any of the NYP changes), things look good! 

I also tested changing the button colours, and the theme's fonts -- both were picked up in expected ways with this PR.

The only thing I noticed is that if you don't have NYP enabled, you lose the ability to configure a block manually. Without tiers or custom pricing, this would only be the ability to enable/disable tabs on a specific block -- so if you have 'Once' turned off under Reader Revenue, you can't turn it on one an individual block. I think this could be changed as an enhancement to the non-NYP, non-NRH version of the block down the road if we wanted to -- it's not a regression.

Going to mark this and the related block PR as approved!

PR #2959, opened 2024-02-29
feat: jetpack default modules

jaredrethman:
### All Submissions:

Adds `jetpack_get_default_modules` hook for delegating Jetpack default active modules on install. 

1. Check out this branch.
2. Run `wp jetpack reset modules` and confirm with "yes".
3. Navigate to Jetpack modules admin page i.e. `/wp-admin/admin.php?page=jetpack_modules&activated=true`
4. Cross reference list of active modules matches the list on https://github.com/Automattic/newspack-plugin/blob/feat/jetpack-default-modules/includes/plugins/class-jetpack.php#L33

### Other information:

### DIFF
--- a/includes/plugins/class-jetpack.php
+++ b/includes/plugins/class-jetpack.php
@@ -155,7 +155,7 @@ class Jetpack {
 		add_filter( 'jetpack_active_modules', [ __CLASS__, 'remove_google_analytics_from_active' ], 10, 2 );
 		add_filter( 'jetpack_get_available_modules', [ __CLASS__, 'remove_google_analytics_from_available' ] );
 
-		add_filter( 'jetpack_get_default_modules', [ __CLASS__, 'default_modules' ] );
+		add_filter( 'jetpack_get_default_modules', [ __CLASS__, 'get_default_modules' ] );
 	}
 
 	/**
@@ -306,7 +306,7 @@ class Jetpack {
 	 * @uses https://developer.jetpack.com/hooks/jetpack_get_default_modules/
 	 * @return string[] Default active modules
 	 */
-	public static function default_modules() {
+	public static function get_default_modules() {
 		return static::$default_active_modules;
 	}
 }
--- a/includes/plugins/class-jetpack.php
+++ b/includes/plugins/class-jetpack.php
@@ -92,11 +92,11 @@ class Jetpack {
 		 */
 		'copy-post',
 		/**
-				 * Extra Sidebar Widgets
-				 *
-				 * @link https://jetpack.com/support/extra-sidebar-widgets/
-				 */
-			'widgets',
+		 * Extra Sidebar Widgets
+		 *
+		 * @link https://jetpack.com/support/extra-sidebar-widgets/
+		 */
+		'widgets',
 		/**
 		 * Gravatar Hovercards
 		 *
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
@jaredrethman this is a good start, and works as described in the testing instructions. But I'm not sure it goes quite far enough, since you have to run the `wp jetpack reset modules` CLI command for it to take effect. The way Atomic sites are configured, they come with Jetpack preinstalled, so the default modules hook may never get called for Newspack sites. At the very least, we should probably enable the default modules when a Newspack site is finished getting setup, right? 

To that end, there's an [option](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/wizards/class-setup-wizard.php#L14) that gets updated when a Newspack site first gets setup. Maybe we can hook into this option value getting set to `true` and set the modules then. I would check whether this is enough to satisfy the Support team's original request. 

Beyond that, I'm getting a warning when I run the `wp jetpack reset modules` CLI command:

```
Warning: Trying to access array offset on value of type bool in /srv/htdocs/wp-content/plugins/jetpack/jetpack_vendor/automattic/jetpack-waf/src/class-brute-force-protection.php on line 1146
```

jaredrethman:
> @jaredrethman this is a good start, and works as described in the testing instructions. But I'm not sure it goes quite far enough...

I misunderstood the request in the ticket, was using the filter suggested there.

> The way Atomic sites are configured, they come with Jetpack preinstalled, so the default modules hook may never get called for Newspack sites. At the very least, we should probably enable the default modules when a Newspack site is finished getting setup, right?

I'm trying to find information on this, are you able to confirm that Jetpack is installed + active at the start/end of setup?

> To that end, there's an [option](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/wizards/class-setup-wizard.php#L14) that gets updated when a Newspack site first gets setup. Maybe we can hook into this option value getting set to `true` and set the modules then. I would check whether this is enough to satisfy the Support team's original request.

I'll proceed using the `update_option_{$option_name}` hook when `newspack_service_enabled` is set to `true`, does that sounds correct?

> Beyond that, I'm getting a warning when I run the `wp jetpack reset modules` CLI command:
> 
> ```
> Warning: Trying to access array offset on value of type bool in /srv/htdocs/wp-content/plugins/jetpack/jetpack_vendor/automattic/jetpack-waf/src/class-brute-force-protection.php on line 1146
> ```

I think this relates to your local setup, specifically your `host` see here - https://github.com/Automattic/jetpack-production/blob/trunk/jetpack_vendor/automattic/jetpack-waf/src/class-brute-force-protection.php#L1146, will confirm.

jaredrethman:
@dkoo have adjusted (https://github.com/Automattic/newspack-plugin/pull/2959/commits/ce86bb3f96f5f42de07a0556cc45bc11025f8fbe) Jetpack default modules activation when:
* `newspack_setup_complete` option is added and value is set to `1`
* `newspack_setup_complete` option is set to `1` https://github.com/Automattic/newspack-plugin/blob/ce86bb3f96f5f42de07a0556cc45bc11025f8fbe/includes/wizards/class-setup-wizard.php#L227

dkoo (approved):
@jaredrethman this is working how I expect, now—when I install the Newspack plugin and complete setup, the default Jetpack modules are set automatically. Thanks for the changes!

gamebits:
Thank you for completing this request! pamTN9-8n2-p2

PR #2960, opened 2024-03-01
fix(donations): force base customer location when country field is optional but state required

chickenn00dle:
### All Submissions:

Closes 0/1206331646877121/1205955387558360/f

This PR replaces the fix in https://github.com/Automattic/newspack-plugin/pull/2951

This addresses a bug where, when WC is platform, and donation country is optional, but state is required, the donation checkout field errors with a missing state warning. We fix this by ensuring the woocommerce_default_customer_address setting is base for this specific configuration.

Screenshot 2024-03-01 at 11 29 19

1. On a test site with donations setup via WooCommerce, go to the donations settings: Newspack > Reader Revenue > Donations
2. In the billing fields section of donations settings, unset country as a required field and set state as a required field (pictured above). This configuration causes a checkout bug where new customers can't get past the billing fields portion since new customers have no default country set and billing states won't populate as expected.
3. On `trunk` attempt to checkout as a new customer and confirm a checkout error appears requiring a billing state
4. On this branch confirm that error doesn't appear

### Other information:


dkoo:
@chickenn00dle this is pretty much exactly what I had in mind! But if we implement the callback in the Blocks plugin, we can isolate it so that it will only affect modal checkout instances instead of all Woo checkouts. What do you think? https://github.com/Automattic/newspack-blocks/pull/1683

chickenn00dle:
> But if we implement the callback in the Blocks plugin, we can isolate it so that it will only affect modal checkout instances instead of all Woo checkouts. What do you think? 

Great idea. I'll go ahead and close this one then in favor of yours. Thanks for the great feedback.

PR #2968, opened 2024-03-04
fix: prevent post lock when fetching metadata

adekbadek:
### All Submissions:

#2936 added a feature to add credit-related metadata to images. If the media post request fails for any reason, the loading animation and post locking will be on indefinitely. 

Ideally, the request failure should be handled somehow, but as a quick solution (just not to block workflow), this PR removes the post locking.

1. Insert an image with a credit metadata in a post, observe the "Publish" button is not disabled while the metadata is loading, but the image features a loader

### Other information:

### DIFF
--- a/assets/blocks/core-image/index.tsx
+++ b/assets/blocks/core-image/index.tsx
@@ -9,6 +9,7 @@
 // External
 import classnames from 'classnames';
 // WordPress
+import { useSelect } from '@wordpress/data';
 import { addFilter } from '@wordpress/hooks';
 import { useState, useEffect } from '@wordpress/element';
 import { createHigherOrderComponent } from '@wordpress/compose';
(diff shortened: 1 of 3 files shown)


jaredrethman:
@adekbadek unfortunately these changes will only enable "Update" being clickable. 

The underlying issue is that the media cannot be fetched from the DB as it is from a distributed post i.e. different origins. I would recommend preventing the `<Loader />` component from loading if the image is not hosted on the current site. 

In `index.tsx` after:
```ts
const [ isCaptionVisible, setIsCaptionVisible ] = useState< boolean >( '' !== caption );
```
add:
```ts
// fetch current site url
const { url: siteUrl }: { url: string } = useSelect(
	select => select( 'core' ).getSite(),
	[ props.attributes.url ]
) ?? {
	url: '',
};
```
Update:
```ts
<Loader attributes={ props.attributes } setAttributes={ props.setAttributes } />
```
to:
```ts
{ siteUrl && props.attributes.url.includes( siteUrl ) && (
	<Loader attributes={ props.attributes } setAttributes={ props.setAttributes } />
) }
```
Import `useSelect`:
```ts
import { useSelect } from '@wordpress/data';
```

For convenience I have added these changes to a separate branch - https://github.com/Automattic/newspack-plugin/compare/fix/image-meta-post-lock...fix/image-meta-post-lock_2

jaredrethman (changes requested):
See https://github.com/Automattic/newspack-plugin/pull/2968#issuecomment-1977161306

adekbadek:
Thanks @jaredrethman ! I've incorporated the proposed changes to this PR. I'll consider this approved.

PR #2971, opened 2024-03-04
fix(ras): reset otp inputs when switching login actions

chickenn00dle:
### All Submissions:

Closes [0/1206709674365921/1205129530451099/f](https://app.asana.com/0/1206709674365921/1205129530451099/f)

This addresses an issue where logging in with OTP does not reset OTP inputs when a reader attempts to either have another code sent OR sign in with a different email address

Screenshot 2024-03-04 at 15 05 30

1. Access a test site while logged out
2. Click the Sign In button to trigger the RAS login flow
3. Choose to login via email (OTP flow)
4. Enter a valid email address and click the `Send authorization code` button
5. Once the OTP code modal renders, input any values for the OTP code, but DO NOT submit
6. Click either the `Try a different email` OR `Send another code` link at the bottom of the modal

Screenshot 2024-03-04 at 15 07 52

7. Re-enter a valid email address to retrigger the OTP flow
8. On `trunk` once the OTP code inputs reappear, the previous input will still be present despite no longer being valid. On this branch, they should be empty once again.

### Other information:

---
  - https://app.asana.com/0/0/1205129530451099

### DIFF
--- a/assets/reader-activation/auth.js
+++ b/assets/reader-activation/auth.js
@@ -250,9 +250,14 @@ window.newspackRAS.push( function ( readerActivation ) {
 					if ( ! readerActivation.getOTPHash() ) {
 						return;
 					}
+					createOTPInputs();
 				}
 				if ( [ 'link', 'pwd' ].includes( action ) ) {
 					readerActivation.setAuthStrategy( action );
+					if ( 'link' === action ) {
+						// Destroy any existing OTP input.
+						destroyOTPInputs();
+					}
 				}
 				actionInput.value = action;
 				container.removeAttribute( 'data-form-status' );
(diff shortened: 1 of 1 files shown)


assets/reader-activation/auth.js
```diff
@@ -293,11 +298,6 @@ window.newspackRAS.push( function ( readerActivation ) {
 					);
 				}
 			} );
-			readerActivation.on( 'reader', () => {
-				if ( readerActivation.getOTPHash() ) {
-					setFormAction( 'otp' );
-				}
-			} );
```
chickenn00dle, line 300: It looks like this was causing setFormAction to fire twice during the OTP flow in conjunction with this call here: https://github.com/Automattic/newspack-plugin/blob/c1a68538ac81c242e1103dcaecc707eb1df03005/assets/reader-activation/auth.js#L433

Removing this seems to do the trick to prevent this from being called twice and doesn't appear to break the flow, but I'm not really sure what other contexts this might come into play so would love it if someone could confirm why this was added initially.

dkoo: This change was intentional in #2053. However, with the move in RAS-ACC towards detecting whether or not a reader has a password and showing the password or OTP input automatically, I don't think it matters.

dkoo (approved):
Nice, works as described!

PR #2979, opened 2024-03-07
Add setting for controlling MailChimp reader default status

claudiulodro:
### All Submissions:

Closes [an issue we've been encountering on a couple client sites](https://app.asana.com/0/1206477424334739/1206714953071858/f).

The original reasoning behind making the default status 'transactional' in MailChimp when people register on a site is correct at a strictly technical level because they technically haven't signed up to receive mailings from the site, but have merely registered on it. However **in practice** it is not the behavior publishers expect and it prevents them from doing e.g. a welcome series of emails when someone registers on the site.

This PR adds a new setting for controlling the default status of these readers in MailChimp. 

1. In the Engagement wizard, make sure you have this reader activity data setting set with an audience ID as well as some newsletters available in your Registration modal:

Screenshot 2024-03-07 at 2 26 52 PM

2. You should now see a new setting **if an audience is set** for controlling the default status. Leave it in "Transactional/Non-Subscribed" for now.

Screenshot 2024-03-15 at 10 14 00 AM

4. In an incognito window, register on the site but **do not** select any newsletters:

Screenshot 2024-03-15 at 10 13 02 AM

5. In MailChimp, observe the reader is in "Non-Subscribed" status:

Screenshot 2024-03-15 at 10 13 32 AM

6. Set the setting to "Subscribed"

Screenshot 2024-03-15 at 10 14 08 AM

7. In an incognito window, register on the site again and **do not** select any newsletters:

Screenshot 2024-03-15 at 10 19 02 AM

8. In MailChimp, observe the reader is in "Subscribed" status:

Screenshot 2024-03-15 at 10 19 24 AM

9. Just to be safe, set the setting back to "Transactional/Non-Subscribed" and register on the site but **do** select a newsletter. In MailChimp the user should be in "Subscribed" status:

Screenshot 2024-03-15 at 10 20 11 AM
Screenshot 2024-03-15 at 10 20 35 AM

### Other information:

---
  - https://app.asana.com/0/0/1206740961392205


dkoo:
(Updated title so that when merging it will trigger a hotfix release)

dkoo (approved):
If almost all publishers are sending newsletter content by group/tag/segment or some other means of sub-grouping their audience contacts, then this is the correct behavior. I think we decided on `transactional` originally because we can't guarantee that a publisher won't have their MC account set up to send content to the entire audience, and we can't enforce it either way. But I'm okay with this as long as it's the behavior publishers expect.

claudiulodro:
Would it be "safer" if I made it a checkbox/toggle in that reader activity data section? Or would that muddle the UI? @Automattic/newspack-design what do you think?

claudiulodro:
I've updated this PR with a setting, changed it from a hotfix (since this is now a new feature), and revised the testing instructions. Can I please get another review? Thanks!

includes/data-events/connectors/class-mailchimp.php
```diff
@@ -62,6 +62,20 @@ private static function get_audience_id() {
...
+	 *
+	 * @return string MailChimp status slug, 'transactional' or 'subscriber'. (Default: 'transactional').
+	 */
+	private static function get_default_reader_status() {
+		$allowed_statuses = [
+			'transactional',
+			'subscribed',
+		];
+		$default_status = Reader_Activation::get_setting( 'mailchimp_reader_default_status' );
+		return in_array( $default_status, $allowed_statuses ) ? $default_status : 'transactional';
```
dkoo, line 76:
```suggestion
		return in_array( $default_status, $allowed_statuses, true ) ? $default_status : 'transactional';
```

CI might throw a warning if we don't use `in_array` with strict type checking

dkoo (approved):
Working as described! left one suggestion, but since it's so minor we can just merge this as soon as that's accepted.

PR #2986, opened 2024-03-11
fix: update support link and hide gam setup for manager connected sites

chickenn00dle:
### All Submissions:

See [this thread in slack](https://a8c.slack.com/archives/C015W6BES8J/p1710190649198039?thread_ts=1710190498.816149&cid=C015W6BES8J).

This PR makes some small adjustments to the GAM setup in Ads wizard:

1. Renders the correct onboarding modal depending on connection to manager. When not connected:
Screenshot 2024-03-14 at 16 49 57

OR when connected:
Screenshot 2024-03-14 at 16 50 24

2. Updates the support link for service account setup
Screenshot 2024-03-11 at 18 12 46

1. Go to the Newspack Ads Wizard on a site connected to a network manager site (Newspack > Advertising)
2. Initiate the setup by selecting the `Click here to connect your account` link (Your site will need to be disconnected from GAM for this to appear)
3. On `trunk` you should see the service account upload onboarding flow. On this branch you should see the OAuth connection flow.
4. Now disconnect your site from both GAM and network manager by going to Newspack > Connections and disconnecting Google, then deactivating the manager plugin.
5. Repeat steps 1-3, and this time confirm you see the service account upload flow on this branch (should also still be present on trunk)
6. In the Service account upload onboarding modal, look for the `How to get a service account user for API access` support link and click it. On `trunk` it should take you here: https://support.google.com/admanager/answer/6078734. On this branch it should take you here: https://developers.google.com/ad-manager/api/authentication

### Other information:

### DIFF
--- a/assets/wizards/advertising/views/providers/index.js
+++ b/assets/wizards/advertising/views/providers/index.js
@@ -1,5 +1,3 @@
-/* globals newspack_ads_wizard */
-
 /**
  * Ad Providers view.
  */
@@ -48,14 +46,6 @@ const Providers = ( { services, fetchAdvertisingData, toggleService } ) => {
 		} );
 	};
 
-	const handleGAMOnboardingClick = () => {
-		if ( newspack_ads_wizard.show_gam_sa_setup ) {
-			return setIsOnboarding( true );
-		}
-
-		window.location = newspack_ads_wizard.gam_connection_url;
-	};
-
 	let notifications = [];
 
 	if ( google_ad_manager.enabled && google_ad_manager.status.error ) {
@@ -80,7 +70,7 @@ const Providers = ( { services, fetchAdvertisingData, toggleService } ) => {
 		! google_ad_manager.status.connected
 	) {
 		notifications.push(
-			<Button key="gam-connect-account" isLink onClick={ handleGAMOnboardingClick }>
+			<Button key="gam-connect-account" isLink onClick={ () => setIsOnboarding( true ) }>
 				{ __( 'Click here to connect your account.', 'newspack-plugin' ) }
 			</Button>
 		);
--- a/assets/wizards/advertising/views/providers/index.js
+++ b/assets/wizards/advertising/views/providers/index.js
@@ -1,3 +1,5 @@
+/* globals newspack_ads_wizard */
+
 /**
  * Ad Providers view.
  */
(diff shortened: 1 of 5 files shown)


assets/wizards/advertising/views/providers/index.js
```diff
@@ -78,33 +80,37 @@ const Providers = ( { services, fetchAdvertisingData, toggleService } ) => {
...
-							setIsOnboarding( true );
-						}
-					} );
-				} }
-				titleLink={ google_ad_manager?.enabled ? '#/google_ad_manager' : null }
-				href={ google_ad_manager?.enabled && '#/google_ad_manager' }
-				notification={ notifications.length ? notifications : null }
-				notificationLevel={ google_ad_manager.created_targeting_keys?.length ? 'success' : 'error' }
-			/>
+			{ newspack_ads_wizard.show_gam_sa_setup && (
```
dkoo: This condition overreaches. If I force a truthy `show_gam_sa_setup` value to simulate a Newspack-hosted site, it ends up hiding the entire GAM card, which makes it impossible to connect a site that's not already connected to GAM.

Screenshot 2024-03-14 at 11 07 04 AM

Instead of hiding the entire card, we should change the behavior of the "Click here to connect your account" link to redirect to the Connections page for Newspack-hosted sites.

dkoo (changes requested):
The _logic_ for handling different behavior between Newspack-hosted and non-Newspack-hosted sites is sound, but the behavior for Newspack-hosted sites isn't quite right. See below for details!

Also, this isn't strictly related to this PR since it's already occurring on `trunk`. But it's a pet peeve of mine, so if you're able to resolve it I'd be much obliged. But it's also not very important and not a blocker, so if it seems like more than ~10 minutes of effort then please feel free to ignore. When clicking "Remove service account credentials" on a site connected via service account creds, it throws a fatal error that forces you to redirect back to the main Newspack dashboard. I would love it if we could handle that more gracefully and just kick you back out to the main Advertising wizard page without thowing an error.

Screenshot 2024-03-14 at 11 05 03 AM

assets/wizards/advertising/views/providers/index.js
chickenn00dle: Thanks for catching this! I thought if the site was connected to manager, we wanted to use the Connections wizard to onboard, but after looking more deeply at the code, I realized we had already accounted for the oauth connection flow in the Advertising wizard, but the problem was the oauth flow was always failing and falling back to the SA upload flow. This was happening because the wizard was always attempting to connect via oauth on render, even when the account hadn't been onboarded yet.

I've made some changes in https://github.com/Automattic/newspack-plugin/pull/2986/commits/8c052b671342db4b80bc6b4398d8a6cf62edd23b to correctly display the correct onboarding step in advertising wizard depending on connection to manager. I've also updated the testing steps to reflect the new changes.

chickenn00dle:
> if you're able to resolve it I'd be much obliged. But it's also not very important and not a blocker, so if it seems like more than ~10 minutes of effort then please feel free to ignore. 

I am having a helluva time trying to get this SA upload flow to work in the first place to reproduce removing it on my end 😞 . So I think we should get a new task created and I can tackle this in a separate PR once we've worked through our higher priority newsletter tasks. Let me know if that works for you.

dkoo:
No worries! It was just something I noticed while testing and thought it might be a quick fix. If it isn't, let's leave it for now.

dkoo (approved):
Working well now! Thanks for the updates.

PR #2989, opened 2024-03-14
feat: add option to choose which RAS meta fields to sync

dkoo:
### All Submissions:

Adds a new option under **Reader Activation > Advanced Settings** that allows admins to choose whether contact data is synced to the connected ESP, and which metadata fields should be synced. Also tweaks some UI component styles and some specific styles within the Reader Activation settings page.

Supersedes https://github.com/Automattic/newspack-network/pull/61.

1. Visit **Reader Activation > Advanced Settings** and confirm there's a new section grouping all settings related to ESP syncing behavior:

Screenshot 2024-03-15 at 12 09 34 PM

2. Confirm that the the "Sync contacts to ESP" toggle is ON by default, and all metadata fields are enabled by default.
3. Smoke test contact syncing via a subscription purchase without changing any settings and confirm behavior is unchanged (all fields in [Newspack_Newsletters class](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/plugins/class-newspack-newsletters.php#L26-L51) are synced).
4. De-select some fields and save. Confirm that the settings are persisted after refreshing the page.
5. Re-test a subscription purchase in a new session and confirm that only the fields kept selected in step 4 are synced.
6. Toggle OFF the "Sync contacts to ESP" option and re-test another subscription purchase. This time confirm that no contact data is synced to the ESP.

### Other information:

---
  - https://app.asana.com/0/0/1206151442802191


dkoo:
Tagging @kevinzweerink for a design review.

assets/components/src/accordion/index.js
```diff
@@ -24,7 +24,7 @@ const Accordion = ( { children, title } ) => {
 				{ title }
 				<Icon className="newspack-accordion__icon" icon={ chevronRight } size={ 24 } />
 			</summary>
-			{ children }
+			<div className="newspack-accordion__content">{ children }</div>
```
dkoo, line 27: Note: for this change and the one in `assets/components/src/accordion/style.scss`, we didn't end up needing to us the Accordion component in this PR, but I'm retaining the changes because it does fix some display issues in the (currently unused) component.

includes/plugins/class-newspack-newsletters.php
```diff
@@ -210,7 +282,7 @@ function( $key ) {
 		}
 
 		// Parse full name into first + last for MC, which stores these as separate merge fields.
-		if ( method_exists( 'Newspack_Newsletters', 'service_provider' ) && 'mailchimp' === \Newspack_Newsletters::service_provider() ) {
+		if ( 'mailchimp' === \get_option( 'newspack_newsletters_service_provider', false ) ) {
```
dkoo, line 294: Had to change this to use the raw option in order to get unit tests to play nicely.

laurelfulford (approved):
This work as described! 🙌 The original fields are syncing with the default, and disabling specific fields leaves them unpopulated in MailChimp. Disabling syncing all together stops information from subscriptions from being synced at all.

Edited to add: I also agree with the comment from the demo, this section is looking really nice now!

PR #2990, opened 2024-03-15
feat(reg-block): add loading message while registration is processing

chickenn00dle:
### All Submissions:

Closes [0/1200550061930446/1206650276976874](https://app.asana.com/0/1200550061930446/1206650276976874/f).

This PR adds a "loading" message while the registration block is processing. This is because the block does a lot behind the scenes and can take a good amount of time to complete registration. It must:

- Send a captcha request
- Create a new user account
- If the registration block is set to add ESP contacts, add a new contact

This is similar to https://github.com/Automattic/newspack-newsletters/pull/1456, however we can't optimistically succeed here since in the case of content gates, we need to verify a user has been created before we move on.

So this PR adds a simple loading message while the form processes to reassure users that the gears are still turning.

Screenshot 2024-03-15 at 17 18 18

1. On a test site, set up a registration wall content gate: https://help.newspack.com/engagement/reader-activation-system/content-gating/#getting-started
2. Activate the memberships plugin and create a membership plan for all new registrations, restricting all posts to members of this plan
3. Visit any post on your site as a logged out user and scroll until the registration wall triggers
4. Enter a valid email address and verify a message appears at the bottom indicating the process may take a few seconds

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/index.php
+++ b/assets/blocks/reader-registration/index.php
@@ -74,7 +74,7 @@ function enqueue_scripts() {
 	\wp_enqueue_script(
 		$handle,
 		\Newspack\Newspack::plugin_url() . '/dist/reader-registration-block.js',
-		[ 'wp-polyfill', 'newspack-reader-activation' ],
+		[ 'wp-polyfill', 'newspack-reader-activation','wp-i18n' ],
 		NEWSPACK_PLUGIN_VERSION,
 		true
 	);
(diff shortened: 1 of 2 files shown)


adekbadek (changes requested):
Works as described, but the new message is almost imperceptible, it's so small and pushed to the bottom of the UI. @Automattic/newspack-design could you please weigh in here?

image

chickenn00dle:
> but the new message is almost imperceptible, it's so small and pushed to the bottom of the UI. 

Good point. Do we have a universal "loading" component to fall back to for cases like this? That might be a good alternative.

In any case, I'll wait to hear what design might have to say about this one!

thomasguillot:
Can we change the "Sign up" button to include a spinner, while maintaining the original width?

(Spinner e.g. https://codepen.io/thomasa8c/pen/yLjKqvW)

thomasguillot:
FWIW this will become a recurring pattern in My Account

Screenshot 2024-04-05 at 22 21 31

chickenn00dle:
Closing in favor of https://github.com/Automattic/newspack-plugin/pull/3180

PR #2992, opened 2024-03-18
feat(ras): updates to metadata fields synced to ESP

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Updates the behavior of the `Membership Status` field when syncing reader data to the connected ESP. Current behavior in `trunk` only syncs this status field for donation-type transactions. This PR also syncs this field for non-donation transactions as follows:

- For one-time purchases: `customer`
- For subscription purchases: the status of the most recent subscription ([see list of possible values](https://woo.com/document/subscriptions/statuses/))

### How to test the changes in this Pull Request:

1. On a site with RAS enabled and connected to an ESP, add the following to a post or page: 
  - A Donate block
  - A checkout button block connected to a non-donation, one-time (non-subscription) product
  - A checkout button block connected to a non-donation subscription product
2. As a new reader, purchase:
  - A one-time donation
  - A subscription donation
  - The one-time non-donation product
  - The subscription non-donation product
3. After each purchase, check the contact data synced to the connected ESP and confirm that the `NP_Membership Status` field value matches the expected behavior:
  - For one-time donations: `Donor`
  - For subscription donations: `Monthly Donor` or `Yearly Donor`
  - For one-time non-donation: `customer`
  - For subscription non-donation: subscription status (generally `pending` or `active` for new subscriptions)
4. Change the statuses of the subscription purchases and confirm that the status is updated as expected:
  - For donations: `Ex-Monthy`/`Ex-Yearly Donor` (`Ex-` is appended if the subscription enters a non-active status)
  - For non-donations: subscription status ([see list of possible values](https://woo.com/document/subscriptions/statuses/))

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2993, opened 2024-03-18
feat(ras): updates to metadata fields synced to ESP

dkoo:
Hotfix version of #2992. Needed because some sites currently mid-migration need this before their migration is completed.

### All Submissions:

Updates the behavior of the `Membership Status` field when syncing reader data to the connected ESP. Current behavior in `trunk` only syncs this status field for donation-type transactions. This PR also syncs this field for non-donation transactions as follows:

- For one-time purchases: `customer`
- For subscription purchases: the status of the most recent subscription ([see list of possible values](https://woo.com/document/subscriptions/statuses/))

1. On a site with RAS enabled and connected to an ESP, add the following to a post or page: 
  - A Donate block
  - A checkout button block connected to a non-donation, one-time (non-subscription) product
  - A checkout button block connected to a non-donation subscription product
2. As a new reader, purchase:
  - A one-time donation
  - A subscription donation
  - The one-time non-donation product
  - The subscription non-donation product
3. After each purchase, check the contact data synced to the connected ESP and confirm that the `NP_Membership Status` field value matches the expected behavior:
  - For one-time donations: `Donor`
  - For subscription donations: `Monthly Donor` or `Yearly Donor`
  - For one-time non-donation: `customer`
  - For subscription non-donation: subscription status (generally `pending` or `active` for new subscriptions)
4. Change the statuses of the subscription purchases and confirm that the status is updated as expected:
  - For donations: `Ex-Monthy`/`Ex-Yearly Donor` (`Ex-` is appended if the subscription enters a non-active status)
  - For non-donations: subscription status ([see list of possible values](https://woo.com/document/subscriptions/statuses/))

### Other information:

### DIFF
--- a/includes/data-events/connectors/class-activecampaign.php
+++ b/includes/data-events/connectors/class-activecampaign.php
@@ -146,16 +146,6 @@ class ActiveCampaign {
 			return;
 		}
 
-		/*
-		 * If the subscription is being activated after a successful first or renewal payment,
-		 * the contact will be synced when that order is completed, so no need to sync again.
-		 */
-		if (
-			( 'pending' === $data['status_before'] || 'on-hold' === $data['status_before'] ) &&
-			'active' === $data['status_after'] ) {
-			return;
-		}
-
 		$contact = WooCommerce_Connection::get_contact_from_order( $data['subscription_id'] );
 
 		if ( ! $contact ) {
--- a/includes/data-events/connectors/class-activecampaign.php
+++ b/includes/data-events/connectors/class-activecampaign.php
@@ -142,7 +142,7 @@ class ActiveCampaign {
 	 * @param int   $client_id ID of the client that triggered the event.
 	 */
 	public static function subscription_updated( $timestamp, $data, $client_id ) {
-		if ( empty( $data['status_before'] ) || empty( $data['status_after'] ) || empty( $data['user_id'] ) ) {
+		if ( empty( $data['status_before'] ) || empty( $data['status_after'] ) || empty( $data['subscription_id'] ) ) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


dkoo:
15b8f4a tweaks the method that builds contact data from an order so that it can now handle an "empty" `WC_Subscription` object in addition to pure `WC_Order` objects. This is needed for subscription migration tools which recreate subscriptions without a parent order.

edd4818 makes a slight tweak to the data event handlers for `donation_subscription_changed` and `product_subscription_changed` listeners so that contact data is built directly from the subscription passed in the data event, instead of fetching the customer from the subscription and then the subscription again from the customer.

laurelfulford (approved):
Thanks @dkoo! 

Overall this worked well -- thanks for confirming in Slack what the outcomes should be! I ran the suggested tests plus a couple more with faking out failing credit cards, and got the expected statuses: 

As a new reader, I ran the following transactions back to back, and got these statuses in MC:
* One-time donation set the status to 'Donor'
* Monthly donation set the status to 'Monthly Donor'
* Non-donation, non-subscription purchase made the status 'customer'
* Non-donation subscription purchase made the status 'active'

With a new reader I then tried: 
* A regular annual donation, which made the status 'Yearly Donor'
* As the reader, navigated to My Account and cancelled the order, which set it to pending and didn't change the status in MC.
* As an admin, I cancelled the pending subscription, and that changed the status to Ex-Yearly Donor.

With a new reader I also tried: 
* Buying a non-donation subscription, just to check if it'd be pending or active -- I did see it go through in the logs as pending first, but then it switched to active and that's what got captured in MC. 
* As that reader, I changed the credit card to one that would fail (`4000000000000341`).
* As an admin I forced a renewal that didn't go through, and the status was changed to on-hold in MailChimp and in WooCommerce.

Lastly, with a new reader:
* Starting a new monthly donation and confirmed the 'Monthly Donor'
* As that reader, I changed my payment information to a failing credit card.
* As an admin I forced a renewal that didn't go through.
* In MailChimp the status stayed 'Monthly Donor', and in WooCommerce the subscription was put On Hold.

This all looks good to me 🚢 Just let me know if you have any questions at all about the above!

PR #2995, opened 2024-03-18
fix(data-events): fallback logic in get_order_data() util method

dkoo:
### All Submissions:

Fixes a logical error introduced in #2935. The `get_order_data()` method correctly checks whether it should return data for only donation orders or for any order, but it did so in a way that changed the behavior of `donation_new` vs. `order_completed` data event listeners.

What we want: the `order_completed` listener should return data for _any_ transaction, whether or not it's a donation.

What's happening: `order_completed` is _only_ firing for non-donation orders.

Also refactors one piece of code in the `normalize_contact_data` method which currently sets the `status` field on every iteration of a `foreach` loop, whereas it only needs to be set once after the loop has completed.

1. On `trunk`, on a RAS-enabled site, make a new donation and observe that not all metadata is synced to the ESP (it will be missing details about the actual transaction).
2. Confirm that a non-donation order via the Checkout Button block does sync all metadata to the ESP.
3. Check out this branch, and confirm that both donation and non-donation-type orders sync all expected metadata to the ESP.
4. Smoke-test the instructions for #2924 to make sure it still works after moving the code to set `status` in the `normalize_contact_data` method.

### Other information:

### DIFF
--- a/includes/data-events/class-utils.php
+++ b/includes/data-events/class-utils.php
@@ -16,6 +16,8 @@ final class Utils {
 	 *
 	 * @param int  $order_id Order ID.
 	 * @param bool $process_donations_only Whether to process only donation orders.
+	 *
+	 * @return array|null
 	 */
 	public static function get_order_data( $order_id, $process_donations_only = false ) {
 		$order = \wc_get_order( $order_id );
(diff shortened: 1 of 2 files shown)


adekbadek (approved):

adekbadek:
Cherry-picked the commits from this PR onto the `alpha` branch, as an "alpha-hotfix".

PR #3001, opened 2024-03-21
fix:  update template for woo 8.7.0

laurelfulford:
### All Submissions:

Replaces #3000; I switched it to a hotfix since these alerts can cause worries for publishers/lot of communication for support.

In WooCommerce 8.7.0, the `woocommerce/templates/myaccount/form-edit-account.php` template received a couple updates, causing our `includes/reader-revenue/templates/myaccount-edit-account.php` template to be considered out of date. This is causing a warning to be thrown:

image

I reviewed [the changes](https://github.com/Automattic/woocommerce/commit/d858d2293044e844c1091350695f8e675eca41c4#diff-fa1b452a90f972190216ad5efd0fbd6acdd409e7575fa9fadd1c963a235afb5f), and it looks like they involve:
* Adding a hook
* Adding a comment to describe an existing hook

In the Newspack plugin, the markup is slightly different than WooCommerce, so the existing hook and the new one would be back-to-back. We've also replaced all of the WooCommerce hooks with ones prefaced with `newspack-`, so they're don't have 1:1 names.

Due to the above, to get the template up-to-date I opted to add a code comment before the existing hook (to line up with the one in Woo), and skip adding an equivalent to the `'woocommerce_edit_account_form_fields'` hook, since it seems kind of redundant.

See: 1200550061930446-as-1206883834710822

1. Apply this PR.
2. Spot check the changes against the changes to [the same section in WooCommerce](https://github.com/Automattic/woocommerce/commit/d858d2293044e844c1091350695f8e675eca41c4#diff-fa1b452a90f972190216ad5efd0fbd6acdd409e7575fa9fadd1c963a235afb5f), to make sure they make sense.
3. Chime in on whether this makes sense, or whether you think we should also add an equivalent to the `'woocommerce_edit_account_form_fields'` hook added in 8.7.0. Thanks!

### Other information:


dkoo (changes requested):
This all looks good to me, but I do think we should add the new `woocommerce_edit_account_form_fields` hook even though it'd appear right next to the `woocommerce_edit_account_form` hook in our part of the code.

The reason we [prefix](https://github.com/Automattic/newspack-plugin/pull/1963) the hook names with `newspack_` is so we can selectively support features that use those hooks. We want to make sure we still have the option to support features that use the new hook, and if we redesign the My Account page we might move it around anyway.

laurelfulford:
Thanks @dkoo! That makes sense to me 🙂 both hooks should now be represented with their `newspack_` equivalents!

dkoo (approved):

PR #3011, opened 2024-03-25
fix: cherry-pick new features for Network sites

dkoo:
### All Submissions:

Cherry-picks new features from #2989 and #2979 for earlier release, as these features are needed urgently for some publishers.

1. Follow testing instructions in #2989 and #2979
2. Ensure no other changes from `trunk` are included here

### Other information:


chickenn00dle (approved):
I spent some time testing this and ran into two issues that aren't necessarily immediate blockers:

1. When I update the Metadata field prefix and create a new contact in MC, I see that the previous NP_ prefixed fields are still being created:

Screenshot 2024-03-26 at 13 07 16

I would expect these to not be populated at all for new contacts, and if they already existed for existing contacts, to be removed and updated with the correct prefix on the next sync for the contact.

2. When setting a default reader status of subscribed, it feels strange to still offer the option to subscribe to a newsletter on registration. As a reader, I would be confused if I did not check this box, but was still subscribed to a newsletter after registration:

Screenshot 2024-03-26 at 13 17 33

Since the primary functionality works as expected though, I'll go ahead and approve this and leave it up to you if you feel these issues should be addressed here or not.

dkoo:
> When I update the Metadata field prefix and create a new contact in MC, I see that the previous NP_ prefixed fields are still being created:
> I would expect these to not be populated at all for new contacts, and if they already existed for existing contacts, to be removed and updated with the correct prefix on the next sync for the contact.

This isn't the case. Maybe a future enhancement, but we want to be careful about destroying contact data in publisher ESPs. Currently, if any merge fields exist in your Mailchimp account already, then they'll continue to appear for all contacts (but the ones that aren't selected won't be populated in new data syncs). The feature will mainly benefit publishers who will be turning on reader activation features for the first time, although publishers can always delete the merge fields they don't need in Mailchimp.

To test this, you can start fresh:

- Using WP CLI, delete the merge fields option `wp option delete _newspack_metadata_fields`
- Using `wp shell`, delete all non-standard merge fields in your connected Mailchimp account (just a shortcut for the delete functionality you can use in Mailchimp, which forces you to do it one field at a time. **Make sure you're connected to your test account and don't mind destroying this data**):

```
wp> $fields = Newspack\Mailchimp_API::get( 'lists/<list ID>/merge-fields?count=1000' )['merge_fields'];
wp> foreach ( $fields as $field ) : if ( ! in_array( $field['name'], [ 'Address', 'First Name', 'Last Name', 'Phone' ] ) ) : Newspack\Mailchimp_API::delete( 'lists/<list ID>/merge-fields/' . $field['merge_id'] ); endif; endforeach;
```

(note: you can get the list ID from Mailchimp > **Audience dashboard > Manage Audience > Settings > Audience name and campaign defaults**)

- Deselect some meta fields in Reader Activation > Advanced Settings and save
- Make a purchase as a reader and confirm that any deselected fields are not created in MC and don't appear on any contacts

> When setting a default reader status of subscribed, it feels strange to still offer the option to subscribe to a newsletter on registration. As a reader, I would be confused if I did not check this box, but was still subscribed to a newsletter after registration:

Yes, you're right about this. But from the [original PR](https://github.com/Automattic/newspack-plugin/pull/2979):

> The original reasoning behind making the default status 'transactional' in MailChimp when people register on a site is correct at a strictly technical level because they technically haven't signed up to receive mailings from the site, but have merely registered on it. However in practice it is not the behavior publishers expect and it prevents them from doing e.g. a welcome series of emails when someone registers on the site.

In practice, most publishers don't use the audience (or master list, in AC) to send actual editorial content. They use it for more transactional-type emails, and manage editorial lists using things like groups or segments. But contacts still need to have a "subscribed" status in order to receive ANY emails (even the transactional ones). So in this case it's up to the publisher to make sure they abide by the user's preferences and not send editorial content to people who haven't signed up to specific lists.

dkoo:
@chickenn00dle I don't think any changes are required to address your feedback at the moment, and since you've approved I'm going to merge this. But let me know if you have any questions about the above!

PR #3015, opened 2024-03-26
refactor(data-events): add abstract Connector class for shared methods

dkoo:
### All Submissions:

Just a small refactor of Data Events to get rid of some duplicated code in two very similar classes. Creates an abstract `Connector` class with shared methods that can be extended for ESP-specific functionality.

No changes to functionality—test that RAS contact syncing still works as before with both Mailchimp and ActiveCampaign.

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -86,6 +86,7 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-popups.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-memberships.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/connectors/ga4/class-ga4.php';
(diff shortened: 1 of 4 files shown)


adekbadek (commented):
Besides the refactor, there's one additive change: the `newsletter_updated` method of the new `Connector` class. Is that intentional?

dkoo:
> Besides the refactor, there's one additive change: the `newsletter_updated` method of the new `Connector` class. Is that intentional?

@adekbadek Whoops, you're right! The `newsletter_updated` method is [specific to ActiveCampaign](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/data-events/connectors/class-activecampaign.php#L173) and already exists there, so it doesn't need to exist in the Connector class. Removed and also resolved conflicts with `trunk` in 712881d8.

adekbadek (approved):

PR #3021, opened 2024-03-28
feat(ia): dashboard site status

jaredrethman:
### All Submissions:

Adds site status component. View `then` properties for determining _successful_ status.

https://github.com/Automattic/newspack-plugin/blob/4503fce97886a1e94289ef85e6321edfa30bd1d8/assets/wizards/dashboard/components/site-statuses/index.tsx#L18-L48

1. Checkout this branch and build assets i.e. `git checkout feat/ia-dash-sitestatus && npm run build`
2. Navigate to - /wp-admin/admin.php?page=newspack#/
3. Observe you see a variation (depending on how your site is setup) of the below.

Screenshot 2024-03-28 at 14 36 53

### Other information:

### DIFF
--- a/assets/wizards/dashboard/components/site-statuses/index.tsx
+++ b/assets/wizards/dashboard/components/site-statuses/index.tsx
@@ -24,14 +24,14 @@ const actions: Statuses = {
 	},
 	googleAnalytics: {
 		...siteStatuses.googleAnalytics,
-		then( payload ) {
-			return payload === 'true';
+		then( { propertyID = '' }: { propertyID: string } ) {
+			return propertyID !== '';
 		},
 	},
 	googleAdManager: {
 		...siteStatuses.googleAdManager,
-		then( { user_basic_info } ) {
-			return Boolean( user_basic_info && user_basic_info.email );
+		then( { services: { google_ad_manager } } ) {
+			return google_ad_manager.available && google_ad_manager.enabled === '1';
 		},
 	},
 } as const;
--- a/assets/wizards/dashboard/components/site-statuses/index.tsx
+++ b/assets/wizards/dashboard/components/site-statuses/index.tsx
@@ -15,32 +15,21 @@ const {
 	newspack_dashboard: { siteStatuses },
 } = window;
 
-const actions: Actions = {
+const actions: Statuses = {
 	readerActivation: {
-		label: __( 'Reader Activation', 'newspack-plugin' ),
-		statusLabels: {
-			success: __( 'Enabled', 'newspack-plugin' ),
-			error: __( 'Disabled', 'newspack-plugin' ),
-		},
-		endpoint: '/newspack/v1/wizard/newspack-engagement-wizard/reader-activation',
-		dependencies: siteStatuses.readerActivation.dependencies,
+		...siteStatuses.readerActivation,
 		then( { prerequisites_status }: { prerequisites_status: PrerequisitesStatus } ) {
 			return Object.values( prerequisites_status ).every( status => status.active );
 		},
 	},
 	googleAnalytics: {
-		label: __( 'Google Analytics', 'newspack-plugin' ),
-		endpoint: '/google-site-kit/v1/core/site/data/connection-check',
-		dependencies: siteStatuses.googleAnalytics.dependencies,
+		...siteStatuses.googleAnalytics,
 		then( payload ) {
 			return payload === 'true';
 		},
 	},
 	googleAdManager: {
-		label: __( 'Google Ad Manager', 'newspack-plugin' ),
-		endpoint: '/newspack/v1/oauth/google',
-		canConnect: siteStatuses.googleAdManager.isAvailable,
-		dependencies: siteStatuses.googleAdManager.dependencies,
+		...siteStatuses.googleAdManager,
 		then( { user_basic_info } ) {
 			return Boolean( user_basic_info && user_basic_info.email );
 		},
--- a/assets/wizards/dashboard/components/site-statuses/index.tsx
+++ b/assets/wizards/dashboard/components/site-statuses/index.tsx
@@ -18,8 +18,8 @@ const {
 const actions: Statuses = {
 	readerActivation: {
 		...siteStatuses.readerActivation,
-		then( { prerequisites_status }: { prerequisites_status: PrerequisitesStatus } ) {
-			return Object.values( prerequisites_status ).every( status => status.active );
+		then( { config: { enabled = false } }: { config: { enabled: boolean } } ) {
+			return enabled;
 		},
 	},
 	googleAnalytics: {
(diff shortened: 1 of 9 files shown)


assets/wizards/dashboard/components/site-statuses/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+const actions: Actions = {
+	readerActivation: {
+		label: __( 'Reader Activation', 'newspack-plugin' ),
+		statusLabels: {
+			success: __( 'Enabled', 'newspack-plugin' ),
+			error: __( 'Disabled', 'newspack-plugin' ),
+		},
+		endpoint: '/newspack/v1/wizard/newspack-engagement-wizard/reader-activation',
+		dependencies: siteStatuses.readerActivation.dependencies,
+		then( { prerequisites_status }: { prerequisites_status: PrerequisitesStatus } ) {
```
dkoo: This is *technically* correct, but in practice, some sites choose to turn off reCAPTCHA and other prerequisites after the fact (against our advice). So there's actually a single boolean option that gets turned on after completing the onboarding flow that you can check for Reader Activation status instead. The option is `newspack_reader_activation_enabled` and its value should be readable in response for the same endpoint you're using already as `response.config.enabled`

assets/wizards/dashboard/components/site-statuses/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+		label: __( 'Google Analytics', 'newspack-plugin' ),
+		endpoint: '/google-site-kit/v1/core/site/data/connection-check',
+		dependencies: siteStatuses.googleAnalytics.dependencies,
+		then( payload ) {
+			return payload === 'true';
+		},
+	},
+	googleAdManager: {
+		label: __( 'Google Ad Manager', 'newspack-plugin' ),
+		endpoint: '/newspack/v1/oauth/google',
```
dkoo: This will only work for production Newspack sites connected to the Newspack manager (which have access to the proxied Google OAuth connection). A more foolproof way to check this is using [this util method](https://github.com/Automattic/newspack-plugin/blob/feat/ia-dash-sitestatus/includes/configuration_managers/class-newspack-ads-configuration-manager.php#L139).

assets/wizards/dashboard/components/site-statuses/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+		then( { prerequisites_status }: { prerequisites_status: PrerequisitesStatus } ) {
+			return Object.values( prerequisites_status ).every( status => status.active );
+		},
+	},
+	googleAnalytics: {
+		label: __( 'Google Analytics', 'newspack-plugin' ),
+		endpoint: '/google-site-kit/v1/core/site/data/connection-check',
+		dependencies: siteStatuses.googleAnalytics.dependencies,
+		then( payload ) {
+			return payload === 'true';
```
dkoo: I'm not familiar with how this Site Kit endpoint works, but after I reset my Site Kit connection (in Site Kit > Settings > Admin Settings) and disconnect all services, this component is still showing a "connected" status.

assets/wizards/dashboard/components/site-statuses/site-status.tsx
```diff
@@ -0,0 +1,147 @@
...
+	const statusLabel =
+		dependencies && 'error-dependency' === requestStatus ? (
+			<Tooltip
+				text={ sprintf(
+					// translators: %s is a comma separated list of needed dependencies.
+					__( '%s must be installed & activated!' ),
+					failedDependencies.map( dep => dependencies[ dep ].label ).join( ', ' )
+				) }
+			>
+				<button
```
dkoo: This is a non-blocking suggestion (cc @kevinzweerink for a UX perspective). I love how if you're missing dependencies, clicking on the status component will pop up the modal to install the missing dependencies. But if all dependencies are installed and the status is still showing "Disconnected", it's not immediately clear how to resolve the status to "Connected". In that case, could we make clicking on the disconnected status ~redirect~ link to the relevant WP admin page to complete setup?

- Reader Activation: `/wp-admin/admin.php?page=newspack-engagement-wizard#/reader-activation`
- Google Analytics: `/wp-admin/admin.php?page=googlesitekit-splash` if Site Kit isn't connected, or `/wp-admin/admin.php?page=googlesitekit-settings#/connected-services/analytics-4` if Site Kit is connected but Analytics isn't
- GAM: `/wp-admin/admin.php?page=newspack-advertising-wizard#/`

dkoo (changes requested):
@jaredrethman This looks great so far. There are a few things that need to be changed in terms of accurately checking each status. Details below:

assets/wizards/dashboard/components/site-statuses/index.tsx
jaredrethman: Thanks for the suggestion, added!

assets/wizards/dashboard/components/site-statuses/index.tsx
jaredrethman: I have added your suggested method into the preflight-check mechanism (i.e before any XMLHttpRequests are made) here - https://github.com/Automattic/newspack-plugin/blob/feat/ia-dash-sitestatus/includes/wizards/class-dashboard.php#L310

Edit:
In addition I have replaced the endpoint with the one used on /admin.php?page=newspack-advertising-wizard to determine connectivity i.e. `/newspack/v1/wizard/billboard`.

assets/wizards/dashboard/components/site-statuses/index.tsx
jaredrethman: Updated to use same endpoint used on Google Site Kit / Analytics Module i.e. `/google-site-kit/v1/modules/analytics-4/data/settings` which will give use a more accurate representation of connectivity.

Changes also include providing relevant config link (like you suggested):
https://github.com/Automattic/newspack-plugin/blob/7eb836665901f11179ac2666b69a4aef86cdcf03/includes/wizards/class-dashboard.php#L319-L329

assets/wizards/dashboard/components/site-statuses/site-status.tsx
jaredrethman: Thanks for this feedback, have updated to include config links when status fails.

assets/wizards/dashboard/components/site-statuses/index.tsx
dkoo: Much simpler and looks good!

assets/wizards/dashboard/components/site-statuses/index.tsx
dkoo: Confirmed it's working for GAM accounts connected via service accounts, too!

assets/wizards/dashboard/components/site-statuses/index.tsx
dkoo: Showing the correct status now!

dkoo (approved):
Looking good now. Thanks for the changes, @jaredrethman!

PR #3032, opened 2024-04-03
feat(ras-acc): update otp email template

chickenn00dle:
### All Submissions:

**PR 1 of 4**

This PR updates the OTP signin email with:

- A new footer per RAS-ACC email designs
- Use of primary and secondary color backgrounds
- Some alignment updates

https://github.com/Automattic/newspack-plugin/assets/17905991/1eb55e9a-802b-4a6a-a6fa-e6859682e9d4

Note: Dynamic text color and FAQ sections added in https://github.com/Automattic/newspack-plugin/pull/3033 and https://github.com/Automattic/newspack-plugin/pull/3034

Note: I am only updating the OTP email for now. Once all of the kinks are worked out it should be pretty simple to update the others.

1. Go to the WP Customizer and set a custom primary and secondary color
2. With these changes checked out, go to Newspack > Engagment > Show Advanced Settings and select the one-time password email under Transactional Email Content section:
Screenshot 2024-04-04 at 15 05 26
3. Trash the post (so that we can force our updated template to be used on next otp sign in attempt)
4. As a logged out user, select the sign in link and attempt to login via email to trigger an otp email
5. Check the otp email and confirm the above points are implemented in the email design

### Other information:

---
  - https://app.asana.com/0/0/1205668937699542


chickenn00dle:
I'm gonna mark this one ready for review despite more work needing to be done to match the designs to hopefully help make review a bit simpler.

I should also note that there is a bug in the newsletters plugin (which it seems the main plugin emails rely on to generate email html) which causes the editor to force buttons to be centered. So our default OTP template will have a correctly left aligned button, but if you go into the post editor for the email template, make any changes, and save then the button will become centered. Similarly, if you send yourself a test email via the post editor, the button will also be centered.

I spent a bunch of time trying to figure out why this is happening, but had no luck as of now. I will tackle this, as well as some of the other email designs in another PR.

assets/other-scripts/emails/style.scss
```diff
@@ -1,3 +1,8 @@
 .editor-post-title {
 	display: none;
 }
+
+.editor-styles-wrapper .block-editor-block-list__layout .wp-block-button {
+	margin-left: 0 !important;
+	margin-right: 0 !important;
+}
```
chickenn00dle, line 8: Added these styles to resolve an issue where the button block was forced to be centered in the editor. This overrides the behavior in the editor, but actual emails will still see centered buttons for some reason. (Aside from our updated OTP default template)

adekbadek (changes requested):
> Trash the post (so that we can force our updated template to be used on next otp sign in attempt)

1. How will migration work for the sites out in the wild?

2. How will the template be updated in the event of site colors update?

chickenn00dle:
Great questions @adekbadek 

> How will migration work for the sites out in the wild?

I think we want to avoid forcing sites to migrate since many publishers have likely put in some work on updating these emails on their end. My thought was to add an option to reset templates for those that would like to use our new designs. We might also want to add a similar reset link to each email template next to the edit link in the Transactional Email Content section:
Screenshot 2024-04-04 at 15 05 26

That said, I was hoping to focus this PR strictly on the design portion of this and tackle these issues in a follow up PR. Happy to add this in here as well if you prefer though!

> How will the template be updated in the event of site colors update?

This one is a bit tricky for publishers that have spent time customizing their email templates. Again, I'd avoid forcing these publishers to fall back to our designs on color update, and leave it up to them to "reset" via the links mentioned above, OR manually update their custom emails on their end.

I'm definitely open to other suggestions for both of these questions too if you have any thoughts!

adekbadek:
> I think we want to avoid forcing sites to migrate since many publishers have likely put in some work on updating these emails on their end

I agree. But publishers who had not touched the templates should have them updated automatically. 

> > How will the template be updated in the event of site colors update?

> I'd avoid forcing these publishers to fall back to our designs on color update, and leave it up to them to "reset" via the links mentioned above, OR manually update their custom emails on their end.

As a publisher, I'd expect brand colors to be updated in the emails if they are updated for the site, even if email copy has changes.

chickenn00dle:
> I agree. But publishers who had not touched the templates should have them updated automatically.

> As a publisher, I'd expect brand colors to be updated in the emails if they are updated for the site, even if email copy has changes.

Sure! I'm happy to address this in a follow up PR, unless you feel it needs to be implemented here. My thought was to keep this PR and some follow-ups focused on the email redesign and handle the reset/update functionality separately.

includes/util.php
```diff
@@ -492,3 +494,50 @@ function newspack_get_countries() {
...
+}
+
+/**
+ * Get theme primary and secondary colors or defaults if none present.
+ *
+ * @return array An array containing primary and secondary colors.
+ */
+function newspack_get_theme_colors() {
+	$primary_color   = get_theme_mod( 'primary_color_hex', '#3366ff' );
+	$secondary_color = get_theme_mod( 'secondary_color_hex', '#f0f0f0' );
```
adekbadek: Newspack Theme functions should be used ([like here](https://github.com/Automattic/newspack-plugin/blob/36238e96a863723720979520400a9afc65dd0cfa/includes/wizards/class-setup-wizard.php#L332-L333)).

adekbadek (approved):
Approving pending one small change requested and keeping in mind that the migration strategy should be addressed (as discussed above) before merging the epic branch.

includes/util.php
chickenn00dle: Nice! I've fixed this in https://github.com/Automattic/newspack-plugin/pull/3032/commits/cc2f29806f1a07870d51adea496646c0d4e247e8

chickenn00dle:
> keeping in mind that the migration strategy should be addressed

As a heads up @adekbadek, I've got some additional PRs up to tackle:

- Migrating unmodified transactional emails: https://github.com/Automattic/newspack-plugin/pull/3054
- Triggering transactional email updates on theme color change: https://github.com/Automattic/newspack-plugin/pull/3052
- Adding a link to manually reset transactional email templates: https://github.com/Automattic/newspack-plugin/pull/3049

These three PRs should collectively address migration concerns.

PR #3033, opened 2024-04-04
feat(ras-acc): dynamically set otp email text color

chickenn00dle:
### All Submissions:

**PR 2 of 4**

This PR adds onto the work in https://github.com/Automattic/newspack-plugin/pull/3032 by adjusting email text, button, and social icon color for the default email template to either white or black depending on background color:

Screenshot 2024-04-04 at 17 35 31

Note: The changes only apply to the one time password email for the time being. Once that email is fully updated, I will get a PR up to update the remaining templates.

1. Go to the WP Customizer and set a dark custom primary and secondary color
2. With these changes checked out, go to Newspack > Engagment > Show Advanced Settings and select the one-time password email under Transactional Email Content section
3. Trash the post (so that we can force our updated template to be used on next otp sign in attempt)
4. As a logged out user, select the sign in link and attempt to login via email to trigger an otp email
5. Check the otp email and confirm all of the text is properly contrasting
6. Repeat the above steps with light custom colors in customizer

### Other information:


includes/util.php
```diff
@@ -495,16 +495,78 @@ function newspack_get_countries() {
...
+ * Pick either white or black, whatever has sufficient contrast with the color being passed to it.
+ * (Copied from the Newspack theme: https://github.com/Automattic/newspack-theme/blob/6dc4e89a65c465abdd207d990e313921f2972a9a/newspack-theme/inc/template-functions.php#L547)
+ *
+ * @param  string $hex Hexidecimal value of the color to adjust.
+ *
+ * @return string Either black or white hexidecimal values.
+ *
+ * @ref https://stackoverflow.com/questions/1331591/given-a-background-color-black-or-white-text
+ */
+function newspack_get_color_contrast( $hex ) {
```
chickenn00dle, line 508: This utility function lives in newspack theme, but given not every publisher will be using this theme, and given how useful this function is, I've copied it over to the main newspack plugin. 

I'm happy to move it back if someone feels it shouldn't be copied over though!

includes/templates/reader-activation-emails/otp.php
```diff
@@ -9,49 +9,54 @@
...
-	<div class="wp-block-button"><a class="wp-block-button__link has-background has-text-align-center wp-element-button" href="*MAGIC_LINK_URL*" style="border-radius:4px;background-color:' . esc_attr( $primary_color ) . '">' . sprintf( /* Translators: s: site title. */ __( 'Continue to %s', 'newspack' ), '*SITE_TITLE*' ) . '</a></div>
+	<div class="wp-block-buttons has-custom-font-size has-medium-font-size" style="font-style:normal;font-weight:400"><!-- wp:button {"textAlign":"center","textColor":"' . esc_attr( $primary_text_color ) . '","style":{"color":{"background":"' . esc_attr( $primary_color ) . '"},"border":{"radius":"4px"},"elements":{"link":{"color":{"text":"' . esc_attr( $primary_text_color ) . '"}}}}} -->
+	<div class="wp-block-button"><a class="wp-block-button__link has-' . esc_attr( $primary_text_color ) . '-color has-text-color has-background has-link-color has-text-align-center wp-element-button" href="*MAGIC_LINK_URL*" style="border-radius:4px;background-color:' . esc_attr( $primary_color ) . '">Continue to *SITE_TITLE*</a></div>
```
adekbadek: Is the removal of translation function (`__()`) intentional here?

includes/templates/reader-activation-emails/otp.php
```diff
@@ -9,49 +9,54 @@
...
+	<!-- wp:social-links {"iconColor":"' . esc_attr( $primary_text_color ) . '","iconColorValue":"' . ( $primary_text_color === 'black' ? '#000000' : '#ffffff' ) . '","iconBackgroundColor":"' . esc_attr( $primary_color ) . '","iconBackgroundColorValue":"' . esc_attr( $primary_color ) . '","className":"is-style-filled-white","layout":{"type":"flex","flexWrap":"nowrap"}} -->
+	<ul class="wp-block-social-links has-icon-color has-icon-background-color is-style-filled-white">' . $social_links . '</ul>
 	<!-- /wp:social-links -->
-
 	<!-- wp:paragraph {"fontSize":"small"} -->
 	<p class="has-small-font-size">' . sprintf( /* Translators: 1: site title 2: site base address. */ __( '%1$s - %2$s', 'newspack' ), '<strong>*SITE_TITLE*</strong>', '*SITE_ADDRESS*' ) . '<br>' . sprintf( /* Translators: 1: link to site url. */ __( 'You received this email because you requested to sign in to %s', 'newspack' ), '<a href="*SITE_URL*">*SITE_URL*</a>' ) . '</p>
```
adekbadek: I've just noticed – the text domain in the translation functions is wrong. Should be `'newspack-plugin'`.

adekbadek (changes requested):
Two issues with the template file:
1. the `$email_html` is not updated along with the `$post_content` changes
2. there's a hard-coded OTP code with `895671` value (instead of a placeholder)

includes/templates/reader-activation-emails/otp.php
chickenn00dle: This was not intentional. Fixed in https://github.com/Automattic/newspack-plugin/pull/3033/commits/6510cc350130095d99547fc49098875bdd880ede

includes/templates/reader-activation-emails/otp.php
chickenn00dle: Good catch! Fixed in https://github.com/Automattic/newspack-plugin/pull/3033/commits/6510cc350130095d99547fc49098875bdd880ede

chickenn00dle:
Thanks for the review @adekbadek!

> the $email_html is not updated along with the $post_content changes

I've updated the email html and did some formatting to hopefully make parsing the email a bit easier.

> there's a hard-coded OPT code with 895671 value (instead of a placeholder)

Good catch. Fixed in https://github.com/Automattic/newspack-plugin/pull/3033/commits/6510cc350130095d99547fc49098875bdd880ede

I also made some additional changes to how we set colors in the default block in https://github.com/Automattic/newspack-plugin/pull/3033/commits/e574d6c8d2ba63ebc42fcd62f87e40188d855c6e and https://github.com/Automattic/newspack-plugin/pull/3033/commits/e4c84731219bd1c3a170fcbb016e93acd061f427. I realized the previous method would not allow the email template to update when theme colors were changed since the block colors were being explicitly set as hex values. 

Unfortunately, it looks like we will not be able to use a dynamic approach for social icons without some changes in the newsletter plugin first since newsletters pretty strictly decides on a colorsheme for social icon images when generating the email html markup. (at least I wasn't able to find a straightforward way without having to make changes in newsletters)

So I've reverted the icons to use the default layout and colorscheme for the time being. I will try to address them in a follow-up PR once I've worked out all of the other design bits for emails:

Screenshot 2024-04-16 at 18 30 13

includes/util.php
```diff
@@ -495,49 +495,116 @@ function newspack_get_countries() {
...
 	$cm          = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'wordpress_seo' );
 	$social_urls = [];
+	$markup      = [
+		'block_markup' => '',
+		'html_markup'  => '',
+	];
+
 	if ( ! is_wp_error( $cm ) ) {
 		$social_urls = [
+			// TODO: Add more social media services here.
```
adekbadek: Why not add them now? From what I see [here](https://github.com/Automattic/newspack-plugin/blob/8b5d5c4c8b7372dd33603988142097435997bf2a/includes/class-profile.php#L36-L69), only LinkedIn and Pinterest are missing.

includes/templates/reader-activation-emails/otp.php
```diff
@@ -9,217 +9,278 @@
...
 	<!-- /wp:paragraph -->
-
 	<!-- wp:buttons {"layout":{"type":"flex","orientation":"horizontal","flexWrap":"nowrap","justifyContent":"left"},"style":{"typography":{"fontStyle":"normal","fontWeight":"400"}},"fontSize":"medium"} -->
-	<div class="wp-block-buttons has-custom-font-size has-medium-font-size" style="font-style:normal;font-weight:400"><!-- wp:button {"textAlign":"center","style":{"color":{"background":"' . esc_attr( $primary_color ) . '"},"border":{"radius":"4px"}}} -->
-	<div class="wp-block-button"><a class="wp-block-button__link has-background has-text-align-center wp-element-button" href="*MAGIC_LINK_URL*" style="border-radius:4px;background-color:' . esc_attr( $primary_color ) . '">' . sprintf( /* Translators: s: site title. */ __( 'Continue to %s', 'newspack' ), '*SITE_TITLE*' ) . '</a></div>
+	<div class="wp-block-buttons has-custom-font-size has-medium-font-size" style="font-style:normal;font-weight:400"><!-- wp:button {"textAlign":"center","backgroundColor":"primary","textColor":"secondary","style":{"border":{"radius":"4px"},"elements":{"link":{"color":{"text":"secondary"}}}}} -->
```
adekbadek, line 42: Is setting the text color to the secondary color intentional? On the screenshot in PR description it's the contrasting text color, not theme secondary color:

image

adekbadek (changes requested):

includes/util.php
chickenn00dle: Fair point! I've added linkedin in https://github.com/Automattic/newspack-plugin/pull/3033/commits/ff88d4b851c20019c59edacd517f9237a95f0f53

It looks like newsletters doesn't support pinterest though so I've omitted that one for the time being:
https://github.com/Automattic/newspack-newsletters/blob/6711238e09a20dd215833207810e9e341cd54a3f/includes/class-newspack-newsletters-renderer.php#L164-L173

I've added this to the list of small changes needed in newsletters and will tackle this altogether in a follow-up PR: 0-as-1205668937699542/1207101310249278/f

includes/templates/reader-activation-emails/otp.php
chickenn00dle: > Is setting the text color to the secondary color intentional? On the screenshot in PR description it's the contrasting text color, not theme secondary color

Sorry about that. It is. I had another look at the designs (c6T1goPQpkchdh7XBAM6aH-fi-1_10676) and it seems like the button text is meant to be the secondary color.

adekbadek (approved):

PR #3050, opened 2024-04-08
fix: ensure active status for memberships tied to active subs

dkoo:
### All Submissions:

Fixes what seems to be a bug in the WC Memberships plugin. When the same user has both a previously expired/cancelled subscription and a newer active one, the "Expires" date from the expired/cancelled subscription forces the user's membership to a `expired` status, even though the newer subscription is still active and should grant membership.

1. On a site with WC Memberships + Subscriptions, create a Membership Plan and tie it to "Product Purchase" + "Subscription length". Set it to restrict some content such as a post category.

Screenshot 2024-04-08 at 4 47 10 PM

2. As a reader, purchase a subscription to the product selected in step 1. Observe that a membership is created in **WooCommerce > Memberships** with "active" status.
3. In WP admin, manually cancel the subscription. Observe that the membership is updated to "expired" status with an "Expires" date matching when you changed the subscription status.
4. As the same reader in step 2, purchase another subscription of the same product.
5. While on `relase`, in WP admin, confim that the membership gets reassociated with the new active subscription. But observe that it gets set to "active" status and then immediately to "expired" status again due to the persistence of the "Expires" date from the first cancelled subscription. 

Screenshot 2024-04-08 at 4 50 53 PM

6. Also observe that as a reader, you're not able to access any content restricted by the Membership Plan. 
7. Check out this branch and refresh the membership in WP admin. Confirm that the "Expires" date is cleared out and that the status has been reset to "active".
8. As the reader, confirm that you once again have access to restricted content due to the active subscription.

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -34,6 +34,14 @@ class Memberships {
 	public static $active_statuses = [ 'active', 'complimentary', 'free-trial', 'pending' ];
 
 	/**
+	 * WooCommerce Subscription statuses that should be considered active.
+	 * See: https://woo.com/document/subscriptions/statuses/
+	 *
+	 * @var array
+	 */
+	public static $active_subscription_statuses = [ 'active', 'pending', 'pending-cancel' ];
+
+	/**
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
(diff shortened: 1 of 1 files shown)


adekbadek (commented):
> Observe that the membership is updated to "expired" status with an "Expires" date matching when you changed the subscription status.

I can't reproduce the issue on the `release` branch. After cancelling the subscription, the membership enters "Pending Cancellation" state (my test product is a monthly subscription, so the membership is active for another month still), not "Expired".

dkoo:
> I can't reproduce the issue on the release branch. After cancelling the subscription, the membership enters "Pending Cancellation" state (my test product is a monthly subscription, so the membership is active for another month still), not "Expired".

What about if you force it into a `cancelled` status immediately via WP admin?

claudiulodro (approved):
Works well in my testing. I like how defensively this is coded too!

PR #3052, opened 2024-04-09
feat(ras-acc): automatically update transactional email templates on theme color update

chickenn00dle:
### All Submissions:

This PR automatically updates the transactional email templates whenever the theme color is updated in the customizer

1. Go to Newspack > Engagement > Show Advanced Settings and select the edit link next to any transactional email template
2. Modify some of the text color to the themes primary and secondary colors:
Screenshot 2024-04-09 at 13 53 45
3. Go to the customizer and make changes to the primary and secondary colors
4. Go back to the transactional email template you edited previously and observe the colors have changed in the editor.
5. Repeat the above steps, but use the Newspack > Site Design UI to change theme colors instead of the customizer

### Other information:

### DIFF
--- a/includes/emails/class-emails.php
+++ b/includes/emails/class-emails.php
@@ -28,7 +28,7 @@ class Emails {
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
 		add_filter( 'newspack_newsletters_email_editor_cpts', [ __CLASS__, 'register_email_cpt_with_email_editor' ] );
 		add_filter( 'newspack_newsletters_allowed_editor_actions', [ __CLASS__, 'register_scripts_enqueue_with_email_editor' ] );
-		add_action( 'customize_save_after', [ __CLASS__, 'customize_save_after' ] );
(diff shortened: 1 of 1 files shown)


includes/emails/class-emails.php
```diff
@@ -587,5 +588,30 @@ public static function get_password_reset_url( $user, $key ) {
 
 		return wp_lostpassword_url();
 	}
+
+	/**
+	 * Trigger an update to all email template posts when theme color is updated in customizer.
```
adekbadek, line 623: The colors can also be changed in the Site Design wizard:

image

Theme mods are stored in options, so an option value update could be a better event to listen for.

adekbadek (changes requested):

includes/emails/class-emails.php
chickenn00dle: Good catch! I've updated to use the `update_option_theme_mods_{theme}` hook in https://github.com/Automattic/newspack-plugin/pull/3052/commits/bb074670a08299814fc0cc365fdc46195de9276d

adekbadek (approved):

PR #3053, opened 2024-04-09
Fallback to subscription for sync data when customer has no orders

claudiulodro:
### All Submissions:

**Note: this is a hotfix.**

This PR adds fallback handling to the ESP sync for the case where a user has an active subscription but hasn't made any payments yet, so doesn't have a latest order. This is super common after a subscription migration, where there are a ton of subscriptions that haven't renewed yet on a site. In this scenario we can get a lot of the info we need from the subscription since it's technically an Order object.

1. Manually create a Subscription for a user in WP Admin. This should be a user that has not made any transactions yet.
2. Before applying this patch: `wp newspack woo resync --user-ids={user_id_from_above_user}`. Observe in MailChimp or ActiveCampaign that none of the product/subscription-status info synchronizes to the ESP.
3. Apply this patch. `wp newspack woo resync --user-ids={user_id_from_above_user}`. Observe that the product/subscription-status info synchronizes to the ESP.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -283,6 +283,14 @@ class WooCommerce_Connection {
 
 		if ( ! $order ) {
 			$order = self::get_last_successful_order( $customer );
+
+			// If customer has no order, they might still have a Subscription.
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works well!

PR #3054, opened 2024-04-09
feat(ras-acc): migrate transactional emails to updated templates

chickenn00dle:
### All Submissions:

This PR migrates any unmodified transactional email templates by trashing email template posts that have not been updated since the publish date.

There will be no visual changes to the emails after a migration since the transactional email redesign lives in another PR, but we can test this by checking the post IDs:

1. First, make sure you have unmodified transactional emails. If you have already modified all of these templates, you can "reset" them by deleting them all using something like the following wp cli command: `wp post delete $(wp post list --post_type='newspack_rr_email' --format=ids) --force`
2. Go to Newspack > Engagement > Show Advanced Settings
3. Take note of the post ids for the transactional emails. You can get these by hovering over the edit link of each email and finding the post ID in the link that appears at the bottom of the browser:
Screenshot 2024-04-09 at 15 50 58
4. Pick one or two email templates and make some changes then update them (You should leave one or more untouched).
5. On the frontend, logged in as a reader, trigger any transactional email. You can do this by creating a new reader account for example.
6. Back as an admin, go to the engagement transactional email settings page and recheck the post IDs for the transactional emails.
7. Confirm the edited templates have retained their post IDs, while the untouched templates now have new post IDs.
8. Again as a reader, trigger another transactional email.
9. Back as admin again, recheck email template post IDs and confirm none have changed (the migration should only occur once)

### Other information:

### DIFF
--- a/includes/emails/class-emails.php
+++ b/includes/emails/class-emails.php
@@ -240,7 +240,7 @@ class Emails {
 		}
 
 		// Migrate to RAS-ACC email templates if migration option is not set AND there have been no manual updates to the templates.
-		if ( ! get_option( 'newspack_email_templates_migrated', false ) ) {
+		if ( get_option( 'newspack_email_templates_migrated', '' ) !== 'v1' ) {
 			$migrated  = true;
 			$templates = get_posts(
 				[
@@ -263,7 +263,9 @@ class Emails {
 				}
 			}
 
-			update_option( 'newspack_email_templates_migrated', $migrated );
+			if ( $migrated ) {
+				update_option( 'newspack_email_templates_migrated', 'v1' );
+			}
 		}
 
 
--- a/includes/emails/class-emails.php
+++ b/includes/emails/class-emails.php
@@ -239,6 +239,34 @@ class Emails {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/emails/class-emails.php
```diff
@@ -239,6 +239,34 @@ public static function send_email( $config_name, $to, $placeholders = [] ) {
 			return false;
 		}
 
+		// Migrate to RAS-ACC email templates if migration option is not set AND there have been no manual updates to the templates.
+		if ( ! get_option( 'newspack_email_templates_migrated', false ) ) {
```
adekbadek: This assumes there will be only one migration needed ever. Suggestion: store the current version ID in the option (e.g. `'v1'`) and compare to a hard-coded string. If we need another migration, we'll just update the hardcoded string. 

With the code as is now, a new migration will necessitate option name update, which will also work but will be less legible.

adekbadek (approved):
Works as described. Left a suggestion about the versioning strategy.

includes/emails/class-emails.php
chickenn00dle: Wonderful idea! Updated to use versioned migrations in https://github.com/Automattic/newspack-plugin/pull/3054/commits/c8d938765732163e011a4d93debf70e6aafbeab8

PR #3056, opened 2024-04-10
feat(ras-acc): add donation cancellation email template

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1206810965479792/1205743341940640/f

This PR adds a new donation cancellation email:

Screenshot 2024-04-10 at 15 40 37

Screenshot 2024-04-10 at 15 39 30

Note: ~~I've populated the cancellation email with super basic copy based mostly off of the email template for receipt. Since we are revisiting email design in ras-acc and this is going into epic/ras-acc, I think this is fine for now. If it's alright, I'd like to address updating the email content if necessary in https://app.asana.com/0/1205234045751551/1205668937699542/f where we will be updating all transactional emails.~~ Scratch that, I found the copy for this here https://newspackdesignp2.wordpress.com/2023/12/05/cancellation-emails/. I will still be addressing the design updates in another pr, so this PR only gets the email setup with the old design.
 
1. Go to Newspack > Reader Revenue > Emails
2. Confirm a new Cancellation email is present as pictured above
3. Select the edit link
4. Confirm the email template post is populated with content relevant to donation cancellation
5. Make some changes to the post and update
6. Go back to the reader revenue emails menu
7. Select the reset link for the Cancellation email
8. Select the edit link once again and confirm your changes have been reset to default
9. As a reader, trigger a donation cancellation email by purchasing a recurring donation then cancelling via my account
10. Confirm you receive a cancellation email and its contents appear as expected (pictured above)
11. As a site admin, disable the cancellation email and repeat step 9
12. Confirm no cancellation email is sent

### Other information:

### DIFF
--- a/assets/wizards/readerRevenue/views/emails/index.js
+++ b/assets/wizards/readerRevenue/views/emails/index.js
@@ -87,6 +87,13 @@ const Emails = () => {
 		<>
 			{ emails.map( email => {
 				const isActive = email.status === 'publish';
+				const notification =
+					email.type === 'cancellation'
+						? __( 'This email is not active.', 'newspack-plugin' )
+						: __(
+								'This email is not active. The default receipt will be used.',
+								'newspack-plugin'
+						  );
 				return (
 					<ActionCard
 						key={ email.post_id }
@@ -115,10 +122,7 @@ const Emails = () => {
 						{ ...( isActive
 							? {}
 							: {
-									notification: __(
-										'This email is not active. The default receipt will be used.',
-										'newspack'
-									),
+									notification,
 									notificationLevel: 'info',
 							  } ) }
 					>
--- a/assets/wizards/readerRevenue/views/emails/index.js
+++ b/assets/wizards/readerRevenue/views/emails/index.js
@@ -88,12 +88,9 @@ const Emails = () => {
 			{ emails.map( email => {
 				const isActive = email.status === 'publish';
 				const notification =
-					email.type === 'cancellation'
-						? __( 'This email is not active.', 'newspack-plugin' )
-						: __(
-								'This email is not active. The default receipt will be used.',
-								'newspack-plugin'
-						  );
+					email.type === 'receipt'
+						? __( 'This email is not active. The default receipt will be used.', 'newspack-plugin' )
+						: __( 'This email is not active.', 'newspack-plugin' );
 				return (
 					<ActionCard
 						key={ email.post_id }
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -33,6 +33,7 @@ class WooCommerce_Connection {
 		\add_filter( 'default_option_woocommerce_subscriptions_allow_switching_nyp_price', [ __CLASS__, 'force_allow_subscription_switching' ], 10, 2 );
 		\add_filter( 'default_option_woocommerce_subscriptions_enable_retry', [ __CLASS__, 'force_allow_failed_payment_retry' ] );
 		\add_filter( 'woocommerce_email_enabled_customer_completed_order', [ __CLASS__, 'send_customizable_receipt_email' ], 10, 3 );
+		\add_filter( 'woocommerce_email_enabled_cancelled_subscription', [ __CLASS__, 'send_customizable_cancellation_email' ], 10, 3 );
 		\add_action( 'woocommerce_order_status_completed', [ __CLASS__, 'maybe_update_reader_display_name' ], 10, 2 );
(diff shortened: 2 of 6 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -522,6 +523,95 @@ public static function send_customizable_receipt_email( $enable, $order, $class
 		return false;
 	}
 
+	/**
+	 * Send the customizable cancellation email in addition to WooCommerce Subscription's default.
+	 * We still want to allow WCS to send its cancellation email since this targets the store admin.
+	 *
+	 * @param bool            $enable        Whether to send the default receipt email.
```
adekbadek: It seems that there is no default _cancellation_ email. This comment and the notice in the UI should be updated:

image

adekbadek (changes requested):

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
chickenn00dle: Good catch! Fixed in https://github.com/Automattic/newspack-plugin/pull/3056/commits/e5b442295492885b5d95f3994a60c2e15a9ea22c

assets/wizards/readerRevenue/views/emails/index.js
```diff
@@ -87,6 +87,13 @@ const Emails = () => {
 		<>
 			{ emails.map( email => {
 				const isActive = email.status === 'publish';
+				const notification =
+					email.type === 'cancellation'
+						? __( 'This email is not active.', 'newspack-plugin' )
+						: __(
+								'This email is not active. The default receipt will be used.',
+								'newspack-plugin'
+						  );
```
adekbadek: This should rather be the other way around – if the type is receipt, use the message about the default receipt. Otherwise, use the generic message.

adekbadek (approved):
Pre-approving pending a small tweak.

assets/wizards/readerRevenue/views/emails/index.js
chickenn00dle: > This should rather be the other way around – if the type is receipt, use the message about the default receipt. Otherwise, use the generic message.

Do you mean this should be changed to:

```
const notification =
    email.type === 'receipt'
        ? __(
            'This email is not active. The default receipt will be used.',
	    'newspack-plugin'
            )
        : __( 'This email is not active.', 'newspack-plugin' );
```

If so, this will have the same result as the current changes. On my end, it looks like the messaging is correct on the frontend as well:

Screenshot 2024-04-17 at 09 35 47

Or did you mean something else?

adekbadek: That's what I meant. The result is the same, but the code is more logical.

chickenn00dle: I see. Happy to make the change if you feel that way! Done in https://github.com/Automattic/newspack-plugin/pull/3056/commits/052255fd74e40b7ef2183d8c0fa4be2c29d1919f

adekbadek: It's not about feelings, it's logic & readability:

```
type === 'receipt' ? notification_about_receipt : generic_notification
```

instead of 

```
type === 'cancellation' ?  generic_notification : notification_about_receipt
```

As long as type can only be `'receipt'` or `'cancellation'` the former version seems fine, but it's not a good assumption to build on (there will be other types).

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -522,6 +523,95 @@ public static function send_customizable_receipt_email( $enable, $order, $class
...
+		foreach ( $frequencies as $frequency => $label ) {
+			$product_id = Donations::get_donation_product( $frequency );
+			if ( $product_id ) {
+				$product_map[ $product_id ] = $label;
+			}
+		}
+
+		$items = $subscription->get_items();
+		$item  = array_shift( $items );
+		$is_donation = Donations::is_donation_product( $item->get_product_id() );
```
adekbadek, line 556: When deleting and account which has active subscriptions, this method will throw with 

```
PHP Fatal error:  Uncaught Error: Call to a member function get_product_id() on null
```

because the `$item` is `null` at this point.

adekbadek (commented):
Since this is an epic branch, I did not submit an issue for this. But let me know if there's a better way.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
chickenn00dle: Thanks for catching this! I'll get a fix up for this today 👍

PR #3058, opened 2024-04-10
fix: add option to display subscriptionless memberships

laurelfulford:
### All Submissions:

This PR is to work with an edge case where readers can have WooCommerce Memberships with expiration dates but without subscriptions. Because Newspack hides the Memberships tab in My Account, there's no way for readers to get information about these memberships.

In this specific case, we don't want to create associated subscriptions for these memberships because they'll send out a bunch of renewal emails. Instead, we just want a way for readers to be able to see when their memberships are expiring, so they can proactively renew them. 

See: 1200550061930446-as-1206901401819708

_I wrote these test steps as though you'd be using different browser sessions to act as the admin and reader, but you can also use the User Switching plugin._

1. Set up a test site with RAS set up, plus WooCommerce Memberships installed.
2. Create two membership plans:
     *  Make a membership plan that is not associated with a subscription. Set it to grant access upon "manual assignment only", and set the membership length to have a "specific length" (this can be any value, though one week, month, year is probably more realistic to test).
     * Make a membership plan that is associated with a subscription. Set up a Checkout Button block that allows you to purchase the associated subscription.
3. Create a test reader.
4. Manually assign the test reader the non-suscription membership plan.
5. In an incognito window, log in as this reader in an incognito window/different browser, and go to My Account. Note that you have no way to view your Memberships -- there's no Membership tab, and if you go to Subscriptions, it says you don't have any.

image

6. Apply this PR and run `npm run build`.
7. As your WordPress Admin user, navigate to WP Admin > WooCommerce > Settings > Memberships, and scroll to the bottom; confirm you have a Reader Activation section, and a checkmark to show active Memberships that don't have associated Subscriptions. The checkmark should be unchecked.

image

9. Check this option, and click Save.
10. Back in your reader browser session, refresh the Subscriptions tab.
11. Confirm that the 'No Subscriptions' message is gone, and you also get an extra table listing your non-subscription Memberships product with its expiry date:

image

12. Still logged in as this reader, navigate to where you set up your Checkout Button block for your subscription membership product, and purchase it.
13. Navigate back to My Account - confirm that 'Subscriptions' still says just 'Subscriptions' (and not 'My Subscription' -- normally the plugin will override that and link directly to your subscription if you have just one). 
14. Click on Subscriptions, and confirm you see your subscription listed, as well as the Memberships table:

image

15. From the Subscription tab, click the 'View' tab next to your Membership, then click 'Manage' in the left column.
16. Click 'Cancel' to cancel your Membership.

image

17. Confirm that the Subscriptions tab now just says 'My Subscription' (because you don't have any active unsubscription-related memberships, it should be back at the default behaviour).
18. As your Admin user, manually re-assign the non-sub Membership to your reader. 
19. As your reader, refresh My Account; if you're still viewing the single Subscription detail, you might need to click out of that section and back to get to the main Subscriptions list view.
20. As the Admin user, navigate to WP Admin > WooCommerce > Settings > Memberships.
21. Uncheck "Display memberships that don't have active subscriptions on the My Account Subscriptions tab. " and click Save.
22. As your reader, refresh My Account and confirm the membership is gone again (to confirm it's still assigned, you can go to /my-account/members-area).

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -46,6 +46,8 @@ class WooCommerce_My_Account {
 			\add_action( 'template_redirect', [ __CLASS__, 'verify_saved_account_details' ] );
 			\add_action( 'logout_redirect', [ __CLASS__, 'add_param_after_logout' ] );
 			\add_action( 'template_redirect', [ __CLASS__, 'show_message_after_logout' ] );
+			\add_action( 'woocommerce_account_subscriptions_endpoint', [ __CLASS__, 'append_membership_table' ], 11 );
+			\add_filter( 'wc_memberships_general_settings', [ __CLASS__, 'option_display_memberships_without_subs' ] );
 		}
 	}
 
@@ -129,6 +131,8 @@ class WooCommerce_My_Account {
 					$default_disabled_items[] = 'downloads';
 				}
 			}
+
+
 
 			$disabled_wc_menu_items = \apply_filters( 'newspack_my_account_disabled_pages', $default_disabled_items );
 			foreach ( $disabled_wc_menu_items as $key ) {
@@ -513,6 +517,74 @@ class WooCommerce_My_Account {
 		if ( isset( $_GET['logged_out'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 			WooCommerce_Connection::add_wc_notice( __( 'You have successfully logged out.', 'newspack-plugin' ), 'success' );
 		}
+	}
+
+	/**
+	 * Optionally append a table of active memberships without subscriptions on the My Account Subscriptions tab.
+	 */
+	public static function append_membership_table() {
+		if ( function_exists( 'wc_memberships_get_user_active_memberships' ) ) {
+			$customer_id            = \get_current_user_id();
+			$memberships_info       = wc_memberships_get_user_active_memberships( $customer_id );
+			$no_sub_memberships     = [];
+
+			// If this option is not enabled, don't return the table.
+			if ( 'no' === get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
+				return;
+			}
+
+			// Create an array of active memberships without active subscriptions.
+			if ( function_exists( 'wc_memberships_has_subscription_product_granted_access' ) ) {
+				foreach ( $memberships_info as $membership ) {
+					if ( ! wc_memberships_has_subscription_product_granted_access( $membership ) ) {
+						$no_sub_memberships[] = $membership;
+					}
+				}
+			}
+
+			// If there are active memberships without subscriptions, present them in a table.
+			if ( $no_sub_memberships ) {
+				echo '<div class="woocommerce-memberships-without-subs">';
+				echo '<h4>' . esc_html__( 'Active Memberships', 'newspack-plugin' ) . '</h4>';
+				echo '<p>' . esc_html__( 'These memberships are active, but don\'t have an associated subscription. They will need to be manually renewed when they expire.', 'newspack-plugin' ) . '</p>';
+				wc_get_template(
+					'myaccount/my-memberships.php',
+					array(
+						'customer_memberships' => $no_sub_memberships,
+						'user_id'              => get_current_user_id(),
+					)
+				);
+				echo '</div>';
+			}
+		}
+	}
+
+	/**
+	 * Adds option to display memberships without subscriptions on the Subscriptions tab of My Account..
+	 *
+	 * @param array $settings WooCommerce Memberships settings.
+	 * @return array
+	 */
+	public static function option_display_memberships_without_subs( $settings ) {
+		$settings[] = array(
+			'name' => __( 'Newspack Reader Activation', 'newspack-plugin' ),
+			'type' => 'title',
+			'id'   => 'wc_memberships_show_unsubbed_memberships',
+		);
+
+		$settings[] = array(
+			'type'    => 'checkbox',
+			'id'      => 'wc_memberships_show_unsubbed_memberships',
+			'name'    => __( 'Memberships without subscriptions', 'newspack-plugin' ),
+			'desc'    => __( 'Display memberships that don\'t have active subscriptions on the My Account Subscriptions tab.', 'newspack-plugin' ),
+			'default' => 'no',
+		);
+
+		$settings[] = array(
+			'type' => 'sectionend',
+		);
+
+		return $settings;
 	}
 }
 
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -524,34 +524,6 @@ class WooCommerce_My_Account {
 	}
 
 	/**
-	 * Adds option to display memberships without subscriptions on the Subscriptions tab of My Account..
-	 *
-	 * @param array $settings WooCommerce Memberships settings.
-	 * @return array
-	 */
-	public static function option_display_memberships_without_subs( $settings ) {
-		$settings[] = [
-			'name' => __( 'Newspack Reader Activation', 'newspack-plugin' ),
-			'type' => 'title',
-			'id'   => 'wc_memberships_show_unsubbed_memberships',
-		];
-
-		$settings[] = [
-			'type'    => 'checkbox',
-			'id'      => 'wc_memberships_show_unsubbed_memberships',
-			'name'    => __( 'Memberships without subscriptions', 'newspack-plugin' ),
-			'desc'    => __( 'Display memberships that don\'t have active subscriptions on the My Account Subscriptions tab.', 'newspack-plugin' ),
-			'default' => 'no',
-		];
-
-		$settings[] = [
-			'type' => 'sectionend',
-		];
-
-		return $settings;
-	}
-
-	/**
 	 * Check if a reader has memberships that aren't associated with subscriptions.
 	 *
 	 * @return array
@@ -580,7 +552,7 @@ class WooCommerce_My_Account {
 	 */
 	public static function append_membership_table() {
 		// If this option is not enabled, stop.
-		if ( 'no' === \get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
+		if ( ! Memberships::get_show_on_subscription_tab_setting() ) {
 			return;
 		}
 
@@ -609,7 +581,7 @@ class WooCommerce_My_Account {
 	 */
 	public static function redirect_to_single_subscription() {
 		// If this option is not enabled, stop.
-		if ( 'no' === \get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
+		if ( ! Memberships::get_show_on_subscription_tab_setting() ) {
 			return true;
 		}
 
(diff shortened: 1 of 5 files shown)


laurelfulford:
@thomasguillot @kevinzweerink This one's kind of "it is what it is" visually, but I'd appreciate your thoughts on hiding the 'Cancel' button from the main Memberships table. Normally it shows in the list, but it seemed like extra clutter, and it wasn't consistent with how the Subscriptions list looked, when both were displaying. 

(I'd also appreciate any feedback on little things we can improve!).

thomasguillot:
Can we hide the "next bill on" column since it's not applicable?

laurelfulford:
> Can we hide the "next bill on" column since it's not applicable?

@thomasguillot Oooo, good call! I just pushed an update -- here's what it looks like without: 

image

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+
+	/**
+	 * Check if a reader has memberships that aren't associated with subscriptions.
+	 *
+	 * @return array
+	 */
+	public static function get_memberships_without_subs() {
+		if ( function_exists( 'wc_memberships_get_user_active_memberships' ) ) {
+			$customer_id              = \get_current_user_id();
+			$memberships_info         = wc_memberships_get_user_active_memberships( $customer_id );
```
dkoo: Nit: this is not currently throwing any errors for me, but it's something we've run into occasionally, so this is a good practice to get into the habit of:

```suggestion
			$memberships_info         = \wc_memberships_get_user_active_memberships( $customer_id );
```

If you're working in a PHP file that has a namespace (as [this one does](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/reader-revenue/my-account/class-woocommerce-my-account.php#L8)), when calling any global function, it's a good idea to prefix it with `\`. This tells PHP to look for this function in the global namespace. If you don't include the `\` prefix, then PHP will first look for the function in the current namespace and then fall back to global if it doesn't exist in the current namespace. I've seen instances where this can cause fatal errors, so adding the `\` sidesteps that whole issue and also makes the code a little clearer in a namespaced file (even if it's not required to avoid errors like in this case).

The exception to this rule is [internal PHP functions](https://www.php.net/manual/en/functions.internal.php), so if you can find the function in the PHP manual, you don't need to specify a namespace (global or otherwise) for it. In my code editor internal PHP functions are helpfully highlighted in a different color than other functions:

Screenshot 2024-04-16 at 1 03 47 PM

You also don't need to prefix the function name when checking whether it exists (like `function_exists`) as this will check for the function in every namespace including the global one, unless you pass a specific namespace.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+			$memberships_without_subs = [];
+
+			// Create an array of active memberships without active subscriptions.
+			if ( function_exists( 'wc_memberships_has_subscription_product_granted_access' ) ) {
+				foreach ( $memberships_info as $membership ) {
+					if ( ! wc_memberships_has_subscription_product_granted_access( $membership ) ) {
```
dkoo:
```suggestion
					if ( ! \wc_memberships_has_subscription_product_granted_access( $membership ) ) {
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+		// If there are active memberships without subscriptions, present them in a table.
+		if ( $memberships_without_subs ) {
+			echo '<div class="woocommerce-memberships-without-subs">';
+			echo '<h2>' . esc_html__( 'Active Memberships', 'newspack-plugin' ) . '</h2>';
+			echo '<p>' . esc_html__( 'These memberships are active, but don\'t have an associated subscription. They will need to be manually renewed when they expire.', 'newspack-plugin' ) . '</p>';
+			wc_get_template(
+				'myaccount/my-memberships.php',
+				array(
+					'customer_memberships' => $memberships_without_subs,
+					'user_id'              => get_current_user_id(),
```
dkoo:
```suggestion
					'user_id'              => \get_current_user_id(),
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
 	}
+
+	/**
+	 * Adds option to display memberships without subscriptions on the Subscriptions tab of My Account..
+	 *
+	 * @param array $settings WooCommerce Memberships settings.
+	 * @return array
+	 */
+	public static function option_display_memberships_without_subs( $settings ) {
+		$settings[] = array(
```
dkoo: Very very minor nitpick, but as a team we tend to prefer the shorthand array syntax instead of the "classic" `array()` function for creating new arrays. So this would look like this instead:

```php
		$settings[] = [
			'name' => __( 'Newspack Reader Activation', 'newspack-plugin' ),
			'type' => 'title',
			'id'   => 'wc_memberships_show_unsubbed_memberships',
		];
```

Neither is better than the other, but I think we just decided that `[]` is more slightly more readable at a glance than `array()`.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+			return $memberships_without_subs;
+		}
+	}
+
+	/**
+	 * Optionally append a table of active memberships without subscriptions on the My Account Subscriptions tab.
+	 */
+	public static function append_membership_table() {
+		// If this option is not enabled, stop.
+		if ( 'no' === get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
```
dkoo:
```suggestion
		if ( 'no' === \get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+	}
+
+	/**
+	 * Returns whether or not to redirect the Subscriptions link to a single subscription, or to the main Subscriptions screen.
+	 *
+	 * @return bool
+	 */
+	public static function redirect_to_single_subscription() {
+		// If this option is not enabled, stop.
+		if ( 'no' === get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
```
dkoo:
```suggestion
		if ( 'no' === \get_option( 'wc_memberships_show_unsubbed_memberships', 'no' ) ) {
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+	}
+
+	/**
+	 * Removes the 'Next Bill On' column in the main Memberships table to tidy it up.
+	 *
+	 * @param array $columns WooCommerce Memberships table columns.
+	 * @return array
+	 */
+	public static function remove_next_bill_on( $columns ) {
+		unset( $columns['membership-next-bill-on'] );
```
dkoo: When reading a property of an array, it's always a good idea to check first whether the property exists in the array before reading it, otherwise it can and will throw a fatal error if the property doesn't exist. That's why you'll see this pattern everywhere in our code:

```suggestion
		if ( ! empty( $columns['membership-next-bill-on'] ) ) {
			unset( $columns['membership-next-bill-on'] );
		}	
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -514,6 +522,128 @@ public static function show_message_after_logout() {
...
+	}
+
+	/**
+	 * Hides 'Cancel' button on main Memberships table to tidy it up.
+	 *
+	 * @param array $actions WooCommerce Memberships available actions.
+	 * @return array
+	 */
+	public static function hide_cancel_button_from_memberships_table( $actions ) {
+		unset( $actions['cancel'] );
```
dkoo:
```suggestion
		if ( ! empty( $actions['cancel'] ) ) {
			unset( $actions['cancel'] );
		}
```

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -333,11 +339,13 @@ public static function is_user_verified() {
 	 * Redirect to "Account details" if accessing "My Account" directly.
 	 * Do not redirect if the request is a resubscribe request, as resubscribe
 	 * requests do their own redirect to the cart/checkout page.
+	 * Do not redirect if this request is a membership cancellation.
 	 */
 	public static function redirect_to_account_details() {
-		$resubscribe_request = isset( $_REQUEST['resubscribe'] ) ? 'shop_subscription' === get_post_type( absint( $_REQUEST['resubscribe'] ) ) : false; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$resubscribe_request       = isset( $_REQUEST['resubscribe'] ) ? 'shop_subscription' === get_post_type( absint( $_REQUEST['resubscribe'] ) ) : false; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
dkoo:
```suggestion
		$resubscribe_request       = isset( $_REQUEST['resubscribe'] ) ? 'shop_subscription' === \get_post_type( absint( $_REQUEST['resubscribe'] ) ) : false; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

This already wasn't globally namespaced in `release`, but since we're changing this line anyway we may as well add it!

dkoo (changes requested):
@laurelfulford works well and with very clean code! Just a few minor suggestions below.

Adding the extra option in WooCommerce settings makes sense, but in practice we usually try not to add our own options into the WooCommerce settings UI. I think we should try to reimplement this new setting in the Newspack > Reader Activation > Advanced Settings wizard page. I'm okay with doing this as a separate follow-up PR so it doesn't block this one, though.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
laurelfulford: TIL -- nice! I didn't realize the purpose of the `\`, I figured it had something to do with namespaced/not namespaced, but didn't realize what it actually did under the hood.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
laurelfulford: Ah nice! Thanks @dkoo -- consistency is always good 🙂 I updated this, but then ended up deleting it for the new React option.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
laurelfulford: Thanks for catching this!

laurelfulford:
Thanks for this, @dkoo! This was all very helpful! 

I made the code corrections in [44698b9](https://github.com/Automattic/newspack-plugin/pull/3058/commits/44698b90fd6e558b2ce02a6ab48c1c0095a936eb), and then took a swing at moving the option into the Reader Activation advanced settings in [d130fae](https://github.com/Automattic/newspack-plugin/pull/3058/commits/d130fae07b3121f15ab6b58563a6769eb01c0568).

It should now appear with the other Membership settings there, and, when the Memberships plugin isn't enabled, it shouldn't appear:

image

(It's also off by default like the original, which my screenshot doesn't really convey!)

Just let me know if it'd be better to roll back this bit and focus on the original set of changes, or if there's anything else that looks amiss -- thanks!

dkoo (approved):
Thanks for the changes! Looks good and works well!

PR #3060, opened 2024-04-11
fix: hourly cron job to check for prematurely expired memberships

dkoo:
### All Submissions:

Addresses the same issue as in #3050, but with a more performant approach.

Rather than filtering data at the moment any membership/subscription is fetched, this implements an hour cron job that iterates through all active subscriptions on the site, checks any memberships associated with those subscriptions, and reactivates memberships that aren't in an active state. This should ensure that anyone with an active subscription should have access to the memberships their subscription(s) grant them.

Follow the same testing instructions as in #3050, but also note that if you have `define( 'NEWSPACK_LOG_LEVEL', 2 );` defined in wp-config.php, you'll see some additional logging showing that the cron job is working:

```
[11-Apr-2024 16:51:00 UTC] [325]  [NEWSPACK][Newspack\Memberships::fix_expired_memberships_for_active_subscriptions]: Checking for expired memberships linked to active subscriptions...
[11-Apr-2024 16:51:02 UTC] [325]  [NEWSPACK][Newspack\Memberships::fix_expired_memberships_for_active_subscriptions]: Reactivated 5 memberships linked to active subscriptions.
```

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -16,7 +16,8 @@ defined( 'ABSPATH' ) || exit;
  */
 class Memberships {
 
-	const GATE_CPT = 'np_memberships_gate';
+	const GATE_CPT  = 'np_memberships_gate';
+	const CRON_HOOK = 'np_memberships_fix_expired_memberships';
 
 	/**
 	 * Whether the gate has been rendered in this execution.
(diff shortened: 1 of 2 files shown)


claudiulodro (approved):
Looks solid. I think we may run into scalability issues if it's processing thousands of subscriptions, but since it's cron it won't really affect the site functionality and there are a few different ways we can adjust it to compensate. I'd like to see if it does run into issues first though :)

PR #3072, opened 2024-04-18
feat: add product option to autocomplete orders

dkoo:
### All Submissions:

When a new order is paid for in WooCommerce, the default behavior is to move the order into `processing` status and require a store admin to manually move it to `completed` status, unless the order contains only products that are both virtual and downloadable. If an order never moves to `completed` status, any data events connected to the `order_completed` hook (including RAS contact data syncing) will never happen. This is counterintuitive to most Newspack publishers, who simply want the orders to complete automatically without manual intervention. 

This PR adds a new "Auto-complete orders" option to all products and product variations, ON by default, which lets orders with those products autocomplete regardless of whether they're virtual and/or downloadable. This essentially changes the default behavior so that orders will be autocompleted unless the "Auto-complete orders" option is intentionally unchecked for a product or variation.

Note that this should be considered a **breaking change** because it will alter the default order processing behavior for existing products across all Newspack sites.

Screenshot 2024-04-18 at 12 49 58 PM

1. On `trunk`, create a new simple product (one-time or subscription) and a variable product (one-time or subscription). Make sure the products and all variations are NOT downloadable.
2. As a reader, purchase the simple + variable products.
3. In WP admin > WooCommerce > Orders, observe that the orders are stuck in `processing` status unless you manually update them to `complete`.
4. Check out this branch and edit the products you created in step 1. Confirm there's a new "Auto-complete orders"  option for the simple product and each variation, checked by default. Don't change anything yet.
5. Purchase the products as a reader again, but this time confirm that these orders do automatically finish in `completed` status without requiring manual intervention.
6. Edit the products again and uncheck the "Auto-complete orders"  option.
7. Purchase the products as a reader again, and confirm that these orders again get stuck in `processing` status unless you manually update them to `complete` (this lets you restore Woo's default order processing behavior on a per-product basis, if needed).

### Other information:

---
  - https://app.asana.com/0/0/1206970859525478

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-products.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-products.php
@@ -1,0 +1,191 @@
+<?php
+/**
+ * Extensions for WooCommerce products.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+use WP_Error;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Connection with WooCommerce's features.
+ */
+class WooCommerce_Products {
+	/**
+	 * Initialize.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function init() {
+		\add_filter( 'product_type_options', [ __CLASS__, 'show_custom_product_options' ] );
+		\add_action( 'woocommerce_variation_options', [ __CLASS__, 'show_custom_variation_options' ], 10, 3 );
+		\add_action( 'woocommerce_process_product_meta', [ __CLASS__, 'save_custom_product_options' ] );
+		\add_action( 'woocommerce_admin_process_variation_object', [ __CLASS__, 'save_custom_variation_options' ], 30, 2 );
+		\add_filter( 'woocommerce_order_item_needs_processing', [ __CLASS__, 'require_order_processing' ], 10, 2 );
+	}
+
+	/**
+	 * Get custom product options for the given product type.
+	 * Product option values are always booleans but represented by string.
+	 * See: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-formatting-functions.php#L37
+	 *
+	 * @return array Keyed array of custom product options.
+	 */
+	public static function get_custom_options() {
+		return [
+			'newspack_autocomplete_orders' => [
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
Is there a use case where an order for a virtual product should **not** be completed immediately? If not, then I don't think there's need for UI or a custom product option here. Instead, the "downloadable & virtual" condition could be changed to just "virtual" (`'woocommerce_order_item_needs_processing'` filter).

dkoo:
> Is there a use case where an order for a virtual product should **not** be completed immediately?

I don't think we can confidently answer this for all Newspack sites, hence offering the option as an escape hatch.

claudiulodro:
> Is there a use case where an order for a virtual product should not be completed immediately?

There are a couple I can think of, like selling ads or (virtual) event tickets, where there are further logistics needed after someone gives money.

includes/reader-revenue/woocommerce/class-woocommerce-products.php
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		$product = is_a( $product, 'WC_Product' ) ? $product : \wc_get_product( $product );
+		if ( ! $product ) {
+			return null;
+		}
+
+		$custom_option = $custom_options[ $option_name ];
+		$meta_key      = $custom_option['id'];
+		$option_value  = $product->meta_exists( $meta_key ) ? $product->get_meta( $meta_key ) : $custom_option['default'];
```
adekbadek, line 71: Shouldn't the true-by-default be conditional on the product being a Newspack Donation product?

adekbadek (approved):
Pre-approving with one question. If the answer is no then we're good to go here. 

> Note that this should be considered a breaking change because it will alter the default order processing behavior for existing products across all Newspack sites.

Don't forget a "BREAKING CHANGE" string in the squashed commit message.

includes/reader-revenue/woocommerce/class-woocommerce-products.php
dkoo: The impetus for this feature was that several publishers implemented custom subscription-based products but didn't make them virtual + downloadable, and therefore weren't syncing contact data upon purchase (because the orders were getting stuck in "processing" status). So we want the defaults to apply across the board to cover this case.

PR #3083, opened 2024-04-23
feat(ras-acc): update newsletters palette on theme primary color updates

chickenn00dle:
### All Submissions:

Part of https://app.asana.com/0/1206943664367847/1205668937699542/f

This PR is meant to be reviewed alongside these newsletters changes: https://github.com/Automattic/newspack-newsletters/pull/1487

These changes ensure we are updating the newsletter color palette option with the latest primary and secondary theme colors before creating transactional emails from templates OR before triggering an update to email HTML when updating a sites primary or secondary newspack theme colors.

1. With both the changes in the newsletters pr as well as this one checked out, go to Newspack > Engagement > Show advanced settings and reset the one time password email.
2. Go to the post editor for the OTP email by selecting the edit link
3. Confirm the social icons are filled either white or black depending on the color of the email footer (should be contrasting)
4. Send a test email to yourself
5. Confirm the social icons are also filled with the same color in the email
6. Go to Newspack > Site Design and change the sites primary color to something of the opposite darkness/lightness than it previously was and save
7. Go back to the OTP email edit page and confirm the social icons are now the opposite color than they were before (black or white)
8. Send another test email to yourself
9. Confirm the social icons are also filled with the same color in the email

### Other information:

### DIFF
--- a/includes/emails/class-emails.php
+++ b/includes/emails/class-emails.php
@@ -493,10 +493,8 @@ class Emails {
 				$theme_colors = newspack_get_theme_colors();
 				\Newspack_Newsletters::update_color_palette(
 					[
-						'primary'        => $theme_colors['primary_color'],
-						'secondary'      => $theme_colors['secondary_color'],
-						'primary-text'   => $theme_colors['primary_text_color'],
-						'secondary-text' => $theme_colors['secondary_text_color'],
+						'primary'      => $theme_colors['primary_color'],
+						'primary-text' => $theme_colors['primary_text_color'],
 					]
 				);
 			}
@@ -649,18 +647,16 @@ class Emails {
 		}
 
 		// Check for theme mod color settings in case a non-newspack theme is installed.
-		if ( ! isset( $previous_value['primary_color_hex'], $previous_value['secondary_color_hex'], $updated_value['primary_color_hex'], $updated_value['secondary_color_hex'] ) ) {
+		if ( ! isset( $previous_value['primary_color_hex'], $updated_value['primary_color_hex'] ) ) {
 			return;
 		}
 
-		if ( $previous_value['primary_color_hex'] !== $updated_value['primary_color_hex'] || $previous_value['secondary_color_hex'] !== $updated_value['secondary_color_hex'] ) {
+		if ( $previous_value['primary_color_hex'] !== $updated_value['primary_color_hex'] ) {
 			// Update the newsletters color palette.
 			$updated = \Newspack_Newsletters::update_color_palette(
 				[
-					'primary'        => $updated_value['primary_color_hex'],
-					'secondary'      => $updated_value['secondary_color_hex'],
-					'primary-text'   => newspack_get_color_contrast( $updated_value['primary_color_hex'] ),
-					'secondary-text' => newspack_get_color_contrast( $updated_value['secondary_color_hex'] ),
+					'primary'      => $updated_value['primary_color_hex'],
+					'primary-text' => newspack_get_color_contrast( $updated_value['primary_color_hex'] ),
 				]
 			);
 
@@ -683,15 +679,11 @@ class Emails {
 				$email_html = str_replace(
 					[
 						$previous_value['primary_color_hex'],
-						$previous_value['secondary_color_hex'],
 						newspack_get_color_contrast( $previous_value['primary_color_hex'] ),
-						newspack_get_color_contrast( $previous_value['secondary_color_hex'] ),
 					],
 					[
 						$updated_value['primary_color_hex'],
-						$updated_value['secondary_color_hex'],
 						newspack_get_color_contrast( $updated_value['primary_color_hex'] ),
-						newspack_get_color_contrast( $updated_value['secondary_color_hex'] ),
 					],
 					$email_html
 				);
--- a/includes/emails/class-emails.php
+++ b/includes/emails/class-emails.php
@@ -488,6 +488,17 @@ class Emails {
 			/** Only attempt to create the email post if wp-includes/pluggable.php is loaded. */
 			return false;
 		} else {
+			// Make sure newsletters color palette is updated with latest theme colors.
+			$theme_colors = newspack_get_theme_colors();
+			\Newspack_Newsletters::update_color_palette(
+				[
+					'primary'        => $theme_colors['primary_color'],
+					'secondary'      => $theme_colors['secondary_color'],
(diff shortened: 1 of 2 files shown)


includes/emails/class-emails.php
```diff
@@ -488,6 +488,19 @@ public static function get_email_config_by_type( $type ) {
 			/** Only attempt to create the email post if wp-includes/pluggable.php is loaded. */
 			return false;
 		} else {
+			// Make sure newsletters color palette is updated with latest theme colors.
+			if ( self::supports_emails() ) {
```
adekbadek: Despite this check, the [test are failing](https://app.circleci.com/pipelines/github/Automattic/newspack-plugin/9667/workflows/2352b9aa-477b-420f-82fc-410773375045/jobs/46291).

adekbadek (changes requested):
After updating the theme color, the primary email color is visible when editing the email, but not in the sent email HTML.

includes/emails/class-emails.php
chickenn00dle: Ah. I think this is because the `Plugin_Manager` downloads the latest release for newsletters, which does not yet have the required method: https://github.com/Automattic/newspack-plugin/blob/725786b70e43ff8d8b9e3789c43cea13f70d4eb1/includes/class-plugin-manager.php#L144

I guess we can hold off on moving forward with this one until the next newsletters release.

chickenn00dle:
Thanks for testing this @adekbadek!

> After updating the theme color, the primary email color is visible when editing the email, but not in the sent email HTML.

That's strange. I am seeing the colors change in both the editor as well as the email:

https://github.com/Automattic/newspack-plugin/assets/17905991/05bba3d3-97f3-4636-a75d-7c7e1dd1c786

Are you resetting the email template once again after changing the primary/secondary colors by any chance?

adekbadek:
@chickenn00dle –

> Are you resetting the email template once again after changing the primary/secondary colors by any chance?

No. I tested by requesting the OTP as a reader, but when testing with the "Testing" sidebar panel (as on the video), the effect is the same. I've inspected the `newspack_email_html` post meta, and even after saving the post in the editor it still has the old color. 

FWIW, I've found that on updating the theme color in the Site Design wizard, the [first condition](https://github.com/Automattic/newspack-plugin/blob/bfc02fce8705e446271c0539d456743e510896dc/includes/emails/class-emails.php#L484-L485) in the `get_email_config_by_type` is always satisfied, so `Newspack_Newsletters::update_color_palette` is never called.

chickenn00dle:
> FWIW, I've found that on updating the theme color in the Site Design wizard, the [first condition](https://github.com/Automattic/newspack-plugin/blob/bfc02fce8705e446271c0539d456743e510896dc/includes/emails/class-emails.php#L484-L485) in the get_email_config_by_type is always satisfied, so Newspack_Newsletters::update_color_palette is never called.

Ah. We actually update the newsletters color palette in two places in this PR. The first is where you pointed out here, where we need to be sure the palette is updated with the latest before we initially create our email templates (after a template is reset for example).

The second is here: https://github.com/Automattic/newspack-plugin/blob/99320355069832c097161673f4bc8f3efaa14d8d/includes/emails/class-emails.php#L656-L665

This triggers whenever the sites primary or secondary color are updated. 

That said, I spoke with @dkoo about this and he pointed out the issue is actually related to a limitation with the mjml library in which newsletter HTML meta won't actually update unless the editor is open and the send/test button is clicked. Luckily, he was able to get a workaround for this added in https://github.com/Automattic/newspack-plugin/pull/3083/commits/7463b3cbe9b5fc7265e1a5e483685eb5c6356c84

So this should be working as expected now, regardless of how you trigger the email after updating site color.

adekbadek (changes requested):
Just noticed that `maybe_update_email_templates` will only be called if no child theme is used (due to `get_template()` use in the hook). The hook won't be called if a child theme is set, and so the email templates won't get updated.

chickenn00dle:
Hmm. I don't think this is the case. The docs for `get_template` say the parent theme is returned for child themes: https://developer.wordpress.org/reference/functions/get_template/

> This function retrieves the directory name of the current theme, without the trailing slash. In the case a child theme is being used, the directory name of the parent theme will be returned.

chickenn00dle:
> Just noticed that maybe_update_email_templates will only be called if no child theme is used (due to get_template() use in the hook). The hook won't be called if a child theme is set, and so the email templates won't get updated.

Good catch @adekbadek! I've fixed in https://github.com/Automattic/newspack-plugin/pull/3083/commits/96aaf98fe6d3627b5743f82db75a009b8ce5707e

includes/emails/class-emails.php
```diff
@@ -488,6 +488,19 @@ public static function get_email_config_by_type( $type ) {
...
 		} else {
+			// Make sure newsletters color palette is updated with latest theme colors.
+			if ( self::supports_emails() && method_exists( '\Newspack_Newsletters', 'update_color_palette' ) ) {
+				$theme_colors = newspack_get_theme_colors();
+				\Newspack_Newsletters::update_color_palette(
+					[
+						'primary'        => $theme_colors['primary_color'],
+						'secondary'      => $theme_colors['secondary_color'],
+						'primary-text'   => $theme_colors['primary_text_color'],
+						'secondary-text' => $theme_colors['secondary_text_color'],
```
adekbadek: `newspack_get_theme_colors` does not return `primary_text_color` and `secondary_text_color` keys:

https://github.com/Automattic/newspack-plugin/blob/725786b70e43ff8d8b9e3789c43cea13f70d4eb1/includes/util.php#L549-L552

So it should be:

```
'primary-text'   => newspack_get_color_contrast( $theme_colors['primary_color'] ),
'secondary-text' => newspack_get_color_contrast( $theme_colors['secondary_color'] ),
```

adekbadek (changes requested):
Child theme is now handled, but noticed a different issue 😬

includes/emails/class-emails.php
chickenn00dle: Ah! That's right. We removed the secondary color from the email templates, so now this should only reference `primary_color` and `primary_text_color`. I've updated in https://github.com/Automattic/newspack-plugin/pull/3083/commits/e89850054f454305a5b9c41e9944414a63daee16. 

Thanks for the keen eye!

adekbadek (approved):

PR #3086, opened 2024-04-24
fix(ras): sync purchase data only for most recent order/subscription

dkoo:
### All Submissions:

Tweaks the reader data sync behavior so that any purchase-related fields only relate to the reader's most recent active order/subscription, if any. It's possible for a single reader to have more than one subscription in different states, so in this case an active one should take precedence when it comes to contact data in the connected ESP.

1. Publish three different non-donation subscription-type products with different attributes (name, price, billing period, etc.) so they're easily distinguished in the connected ESP data.
2. As a reader, purchase subscriptions for two of the products.
3. As an admin, force immediate cancellation of the older subscription (in WooCommerce > Subscriptions > Edit Subscription, change the status to "Cancelled" or "Expired").
4. On `release`, log in as the reader in a new session.
5. In the connected ESP, observe that the contact's metadata reflects a mix of data from both subscriptions: e.g. the `NP_Current Subscription End Date` will reflect the time you changed the status of the cancelled subscription in step 3, but the `NP_Membership Status`, `NP_Billing Cycle`, `* Payment Date` fields, etc. will all reflect the more recent active one.
6. Check out this branch and purchase a third subscription with the third product (so you should have one cancelled/expired sub and 2x active ones). Confirm that the contact data in the ESP now reflects only this most recent subscription, and that `NP_Current Subscription End Date` has been emptied out (because there's an active subscription, so there shouldn't be an end date).
7. As an admin, cancel the older active subscription. Confirm that the contact data in the ESP still reflects only the remaining active subscription.

### Other information:

---
  - https://app.asana.com/0/0/1207153288488122

### DIFF
--- a/includes/plugins/class-newspack-newsletters.php
+++ b/includes/plugins/class-newspack-newsletters.php
@@ -29,32 +29,7 @@ class Newspack_Newsletters {
 	 *
 	 * @var array
 	 */
-	public static $metadata_keys = [
-		'account'              => 'Account',
-		'registration_date'    => 'Registration Date',
-		'connected_account'    => 'Connected Account',
-		'signup_page'          => 'Signup Page',
-		'signup_page_utm'      => 'Signup UTM: ',
-		'newsletter_selection' => 'Newsletter Selection',
-		'referer'              => 'Referrer Path',
-		'registration_page'    => 'Registration Page',
-		'current_page_url'     => 'Registration Page',
-		'registration_method'  => 'Registration Method',
-
-		// Payment-related.
-		'membership_status'    => 'Membership Status',
-		'payment_page'         => 'Payment Page',
-		'payment_page_utm'     => 'Payment UTM: ',
-		'sub_start_date'       => 'Current Subscription Start Date',
-		'sub_end_date'         => 'Current Subscription End Date',
-		'billing_cycle'        => 'Billing Cycle',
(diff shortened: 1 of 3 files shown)


adekbadek (changes requested):
In my testing, only the "total paid" field gets updated*. I've two products with different prices:
1. Add subscription w/ product A -> ESP updated as expected
2. Add subscription w/ product B -> only "total paid" field updated
3. Cancel subscription for product A -> only "end date" is updated

In other words, fields don't get overwritten, only previously empty fields are updated.


\* by which I mean both what I see in ESP (AC) and the value of `$metadata` here: https://github.com/Automattic/newspack-plugin/blob/4df766b29d49552a218ada376a010dfad6d86e2e/includes/reader-revenue/woocommerce/class-woocommerce-connection.php#L367

dkoo:
> In other words, fields don't get overwritten, only previously empty fields are updated.

@adekbadek hmm, I can't replicate this. Are your subscription products set to both "virtual" and "downloadable"? If not, their orders might be stuck in "processing" status, and they won't be found by the method that looks for the last completed order (`get_last_successful_order`). #3072 should resolve this for most sites.

adekbadek (approved):
> Are your subscription products set to both "virtual" and "downloadable"? If not, their orders might be stuck in "processing" status, and they won't be found

That was it! One of my test products was not "downloadable". Works fine now!

PR #3091, opened 2024-04-25
feat(ras-acc): additional component updates

laurelfulford:
### All Submissions:

This PR makes some more updates to the component styles to match the latest mockups. Specifically, it:

#### Adds an example of an error state, and applies those styles to the checkout and sign up modals. 

I approached this like we would add a CSS class around the field that has an error (`.newspack-ui__error`), and then used the existing behaviour (the `400` data-status on the sign-in/up modal and the classes WooCommerce adds) to apply those styles there rather than trying to work in the classes. I'd love to hear if this sounds okay, or terrible!

image

image

image

----

#### Reworked the checkbox/radio code to allow for helper text, status labels

Overall, this includes switching the checkbox labels to display as `grid`, and include helper/error text. 

image

I also added an example of a Badge, applied to a input card. This one is set up with the assumption that the badge text will be injected into the `label` when applicable + some `:has` CSS to adjust the grid if the status is there. That might be overkill though -- it can also be reworked to add a CSS class when a label should be applied. This component isn't used in the RAS-ACC styles, but I wanted to make sure it would work with the other checkbox changes:

image

As part of this, I also tweaked the Woo payment option styles to remove some extra space the above added - I basically used the assumption that if there's only one payment option we're hiding the radio button, but just let me know if that's not correct! 

#### Edited to add: updated the newsletter "coda" styles

image

----

#### And some small things:
* I set a base font colour and label colour to the styles.
* I renamed the `newspack-ui__input-list` to `newspack-ui__input-card`, which is closer to what it's referred to in Figma.
* I started beefing out the demo content a bit more, adding anchors and some more examples. 
* Reordered the Stripe styles a bit in the Woo overrides SCSS, to keep 'em together.
* Added a style for the 'required' asterisk, following how Woo approaches the markup (`abbr`).

See 1205234045751551-as-1206899964006422

1. Apply PR and https://github.com/Automattic/newspack-blocks/pull/1719, and run `npm run build`.
2. Turn on the newsletter 'coda' in the checkout by adding `define( 'NEWSPACK_ENABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP', true );` to the wp-config.php file.
3. Log into your test site, and append `?ui-demo` 
4. Review the appearance of the components, and their structure (primarily the changes around checkbox/radio buttons -- does this approach for errors & adding a badge make sense?). 
5. Set up a checkout button block or donate block, and run through the checkout. At each prompt, submit the form without doing anything, or by doing stuff wrong, to see the error styles:

image

image

image

6. Confirm that the newsletter coda is picking up the input 'card' styles -- it's not split into two screens yet (success and newsletter):

image

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -149,6 +149,11 @@
 		&__content {
 			backface-visibility: visible;
 			padding: var( --newspack-ui-spacer-5 );
+
+			// Make sure there's enough space above the first button in modals
+			.newspack-ui__button:not( :first-child ):first-of-type {
+				margin-top: var( --newspack-ui-spacer-5 );
+			}
 		}
 
 		&__footer {
--- a/assets/newspack-ui/scss/elements/_typography.scss
+++ b/assets/newspack-ui/scss/elements/_typography.scss
@@ -1,4 +1,5 @@
 .newspack-ui {
+	color: var( --newspack-ui-color-neutral-90 );
 	font-family: var( --newspack-ui-font-family );
 	font-size: var( --newspack-ui-font-size-s );
 	line-height: var( --newspack-ui-line-height-s );
--- a/assets/newspack-ui/scss/elements/forms/_buttons.scss
+++ b/assets/newspack-ui/scss/elements/forms/_buttons.scss
@@ -1,6 +1,6 @@
 .newspack-ui {
 	// TODO: define some kind of size hierachy for buttons eg. small, medium, large.
-	&__button {
+	.newspack-ui__button {
 		align-items: center;
 		background: var( --newspack-ui-color-neutral-10 );
 		border: 0;
--- a/assets/newspack-ui/scss/elements/forms/_buttons.scss
+++ b/assets/newspack-ui/scss/elements/forms/_buttons.scss
@@ -2,8 +2,10 @@
 	// TODO: define some kind of size hierachy for buttons eg. small, medium, large.
 	&__button {
 		align-items: center;
+		background: var( --newspack-ui-color-neutral-10 );
 		border: 0;
 		border-radius: var( --newspack-ui-border-radius-m );
+		color: var( --newspack-ui-color-neutral-90 );
 		display: inline-flex;
 		font-family: var( --newspack-ui-font-family );
 		font-size: var( --newspack-ui-font-size-s );
@@ -18,9 +20,14 @@
 			margin-bottom: var( --newspack-ui-spacer-2 );
 		}
 
+		&:hover {
+			background: var( --newspack-ui-color-neutral-30 );
+			color: var( --newspack-ui-color-neutral-90 );
+		}
+
 		&:disabled {
-			background: var( --newspack-ui-color-neutral-10 ) !important;
-			color: var( --newspack-ui-color-neutral-70 ) !important;
+			background: var( --newspack-ui-color-neutral-5 ) !important;
+			color: var( --newspack-ui-color-neutral-40 ) !important;
 			cursor: default;
 			pointer-events: none;
 		}
--- a/assets/newspack-ui/scss/elements/forms/_index.scss
+++ b/assets/newspack-ui/scss/elements/forms/_index.scss
@@ -57,7 +57,7 @@
 	}
 
 	// Error class to apply to specific fields.
-	&__error,
+	&__field-error,
 	[data-form-status="400"] { // Attribute added to sign-up modal form on failure.
 		--newspack-ui-color-input-border: var( --newspack-ui-color-error-50 );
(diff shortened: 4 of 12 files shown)


thomasguillot:
> I approached this like we would add a CSS class around the field that has an error (.newspack-ui__error), and then used the existing behaviour (the 400 data-status on the sign-in/up modal and the classes WooCommerce adds) to apply those styles there rather than trying to work in the classes. I'd love to hear if this sounds okay, or terrible!

I must admit I was slightly confused at first because I was expecting an `error` class directly applied to the `input` (and to the `label`). I think adding the class to the wrapper, in the case of the demo `p`, make sense but I'm slightly concerned if we end up having a case where the wrapper has somehow 2 inputs. 

I must admit, I was initially confused because I anticipated an "error" class to be directly applied to the `input` (and the `label`).
While it makes sense to add the class to the wrapper, in this case of the demo, a `p`, I'm a little bit concerned about potentially encountering a scenario where the wrapper contains 2 inputs.

thomasguillot:
>I approached this like we would add a CSS class around the field that has an error (.newspack-ui__error), and then used the existing behaviour (the 400 data-status on the sign-in/up modal and the classes WooCommerce adds) to apply those styles there rather than trying to work in the classes. I'd love to hear if this sounds okay, or terrible!


I must admit, I was initially confused because I anticipated an "error" class to be directly applied to the input (and the label).
While it makes sense to add the class to the wrapper, in this case of the demo, a p, I'm a little bit concerned about potentially encountering a scenario where the wrapper contains 2 inputs.

thomasguillot:
>I also added an example of a Badge, applied to a input card. This one is set up with the assumption that the badge text will be injected into the label when applicable + some :has CSS to adjust the grid if the status is there. That might be overkill though -- it can also be reworked to add a CSS class when a label should be applied. This component isn't used in the RAS-ACC styles, but I wanted to make sure it would work with the other checkbox changes:

I think this works well!

thomasguillot:
Screenshot 2024-04-26 at 13 48 48@2x

The margin here is wrong, it should be `var(--newspack-ui-spacer-base)` but the input is in a separate `div` and has `var(--newspack-ui-spacer-2)`

thomasguillot:
Screenshot 2024-04-26 at 14 03 53@2x

I can't remember if it was there before but there's a big gap between the "Donate now" button and the "Back" button.

The "Pay with your credit card..." helptext should be in gray-60 (same if we have multiple payment method and they are unselected)

Is it possible to move the "Your personal data will..." to the modal footer (not sure if this should be part of this PR tbh)

laurelfulford:
Thanks @thomasguillot! 

> I must admit, I was initially confused because I anticipated an "error" class to be directly applied to the input (and the label).

I think this is a fair point! I've updated the PR to add the CSS class to the input and label instead.

> The margin here is wrong, it should be var(--newspack-ui-spacer-base) but the input is in a separate div and has var(--newspack-ui-spacer-2)

I think I have the spacing sorted, but let me know if anything still doesn't seem right! (I think the gap in the checkout screen was a regression from when I was trying to fix other spacing issues). 

> Is it possible to move the "Your personal data will..." to the modal footer (not sure if this should be part of this PR tbh)

Unfortunately it doesn't look like it, at least not easily: the contents of the mobile checkout are all coming from an iframe -- it's probably technically possible to move, but I think it might be pretty hacky, and at least best to leave for now. 

Just let me know if there's anything else!

thomasguillot:
> I've updated the PR to add the CSS class to the input and label instead.

Awesome, thank you @laurelfulford, I think this is better.

---

2 small nitpicks otherwise we're good to go 👍

> I think I have the spacing sorted, but let me know if anything still doesn't seem right!

The error message when the account is already registered and the "Continue" button, the gap should be 24px.

> Unfortunately it doesn't look like it, at least not easily: the contents of the mobile checkout are all coming from an iframe -- it's probably technically possible to move, but I think it might be pretty hacky, and at least best to leave for now.

Ok, no worries then! Can we just change the colour to `neutral-60`?

laurelfulford:
> The error message when the account is already registered and the "Continue" button, the gap should be 24px.

I double-checked, and this should be fixed in the PR though the screenshot's a little out of date! This is what it looks like right now:

image

> Ok, no worries then! Can we just change the colour to `neutral-60`?

Yep, updated!

image

thomasguillot:
✅ Design

Thank you so much, @laurelfulford!

chickenn00dle (approved):
Changes look good 👍

laurelfulford:
Thanks @chickenn00dle!! 🙌 🎉  I'll work on getting [the related blocks PR](https://github.com/Automattic/newspack-blocks/pull/1719) fixed before merging this, since they should both go at the same time.

PR #3107, opened 2024-05-10
feat: update variation styles

laurelfulford:
### All Submissions:

This PR, along with https://github.com/Automattic/newspack-blocks/pull/1729, makes some tweaks to the variation styles:

* It reorders the SCSS a bit (moving the variations to the bottom of the modal CSS)
* It adjusts the styles to account for mixed height variations, for products that include fees, and products that are on sale.

See: 1205234045751551-as-1206482127342706

1. Apply this PR and https://github.com/Automattic/newspack-blocks/pull/1729, and run `npm run build` for both plugins.
2. Create two products with variations: one with a regular variable product, and one with variable subscriptions.
3. Of the variations, try adding some different price configurations: for either, add a sales price to at least one product, and for the subscription product, add an initial fee and a trial.
4. Assign each of those products to a checkout button block.
5. Do a visual review of the front-end and confirm things look okay -- primarily the additional info (sale price, fees)

image

image

6. As an edge case, you can try to make one variation taller (usually by having a really long variation name, or two + variations applied) and confirm that the visual treatment of the taller variation makes sense: 

image

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/_mixins.scss
+++ b/assets/newspack-ui/scss/_mixins.scss
@@ -1,3 +1,5 @@
+@use 'variables/breakpoints';
+
 @mixin newspack-ui-elevation($size: 1) {
     @if $size == 1 {
         box-shadow: 0 1px 10px rgba( 0, 0, 0, 0.07 );
@@ -6,4 +8,24 @@
     }  @else if $size == 3 {
         box-shadow: 0 3px 30px rgba( 0, 0, 0, 0.21 );
     }
+}
+
+@mixin media( $res ) {
+	@if mobile == $res {
+		@media only screen and ( min-width: breakpoints.$mobile_width ) {
+			@content;
+		}
+	}
+
+	@if tablet == $res {
+		@media only screen and ( min-width: breakpoints.$tablet_width ) {
+			@content;
+		}
+	}
+
+	@if desktop == $res {
+		@media only screen and ( min-width: breakpoints.$desktop_width ) {
+			@content;
+		}
+	}
 }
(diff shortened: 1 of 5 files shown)


laurelfulford:
Copying over some good feedback from @thomasguillot:

> At first glance, it looks okay. My only suggestion would be to align the prices horizontally.

image

image

laurelfulford:
@thomasguillot do you think it would work to add more space around the price always to create space for the sale price? Or is this too much of a change for something that won't be on most sites?

image

dkoo (approved):
"...and a $80 sign-up fee" sounds funny, but looks like that's not from our template: https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wc-subscriptions-product.php#L400

Otherwise looks great!

Screenshot 2024-05-15 at 4 43 01 PM

laurelfulford:
Thanks @dkoo! 

> "...and a $80 sign-up fee" sounds funny...

It should be **an** $80 fee 😆 Seriously though, it is a little strange in this context -- possibly something we can experiment with filtering down the road?

dkoo:
Seems like it'd be a rare edge case for us. I'm not even sure there's been a Newspack publisher who charges signup fees? Probably best fixed in the source repo, anyway...

thomasguillot:
@laurelfulford I pushed some minor changes: https://github.com/Automattic/newspack-plugin/pull/3107/commits/66e3548814bc7555786b81b9983d656fcfb7e583
(I messed up the commit message. It's not just a fix for the line-height)
One thing I noticed is we were using the wrong colour (`#111111` vs `#1e1e1e`) so this should be fixed now.
I also tweaked the spacings a little bit.

laurelfulford:
Thanks @thomasguillot! 

The changes look good -- the only thing I tweaked was removing the `color: red` [here](https://github.com/Automattic/newspack-plugin/commit/66e3548814bc7555786b81b9983d656fcfb7e583#diff-c7bef86c40e15929c368d329daf3c59dcedd05ef4ec7bc5b69e73465b45bf3f9R129) -- that looks like a test line to make sure that everything else was inheriting correctly.

I'm going to go ahead and merge this one!

PR #3109, opened 2024-05-13
feat: update donation landing page

thomasguillot:
### All Submissions:

This PR adds cover and columns blocks to the automatically generated donation page. Hopefully, it will enable publishers to be a bit more creative with this page, rather than the boring white page we were previously generating.

eg

1. Switch to this branch
2. Install newspack on a fresh site
3. Run through the onboarding and go to the reader revenue wizard to save the donation page

### Other information:


includes/class-donations.php
```diff
@@ -801,35 +801,59 @@ public static function checkout_create_order_line_item( $item, $cart_item_key, $
...
-
+		$title   = esc_html__( 'Support our publication', 'newspack' );
+		$intro   = esc_html__( 'With the support of readers like you, we provide thoughtfully researched articles for a more informed and connected community. This is your chance to support credible, community-based, public-service journalism. Please join us!', 'newspack' );
+		$content = esc_html__( "Edit and add to this content to tell your publication's story and explain the benefits of becoming a member. This is a good place to mention any special member privileges, let people know that donations are tax-deductible, or provide any legal information.", 'newspack' );
+
+		$container_block       = '
+			<!-- wp:cover {"url":"https://newspack.com/wp-content/uploads/2024/05/pexels-brotin-biswas-158640-518543-scaled.jpg","dimRatio":60,"overlayColor":"black","isUserOverlayColor":true,"minHeight":80,"minHeightUnit":"vh","align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|80","bottom":"var:preset|spacing|80","left":"var:preset|spacing|80","right":"var:preset|spacing|80"}}},"layout":{"type":"constrained"}} -->
```
adekbadek: This assumes that the `https://newspack.com/wp-content/uploads/2024/05/pexels-brotin-biswas-158640-518543-scaled.jpg` file will always be available, which is not a given.

adekbadek (changes requested):

includes/class-donations.php
thomasguillot: Should I use `newspack.newspackstaging.com` instead?

adekbadek: I don't think this block pattern should rely on any externally hosted image. The user may be happy with the image (so they don't change it), and then later it's removed from the external host for any reason. The page would suddenly be broken. The image should be downloaded first, and used from the site's media library.

thomasguillot: Frankly, I sincerely hope the publisher will change this image (I might change it to something more abstract to make sure it’s changed…). Also, we already have patterns using external images hosted on `newspack.newspackstaging.com`. I’ve also noticed Core's patterns uses external images (e.g. https://s.w.org/images/core/5.8/architecture-04.jpg).

thomasguillot: Ok I moved the image to `/includes/`. I also changed it to make sure a publisher will replace it!

https://github.com/Automattic/newspack-plugin/pull/3109/commits/d662d423e5aeca41204db579a99bdb450370c29e

adekbadek: That's a better solution! WDYT about renaming the image file to something like "placeholder-image", so its purpose is clear?

> Also, we already have patterns using external images hosted on newspack.newspackstaging.com.

That's not good. The problem is not in external image hosting, but in reliable hosting. `newspack.newspackstaging.com` is a staging site, subject to random changes. It can't be relied upon to host images used by live sites. These instances should be updated to use the placeholder image introduced in this PR. 

If a pattern was inserted, and then updated in this repository, will it be updated on a post that used the pattern?

adekbadek (approved):
Suggested a file name change, other than that all good here.

includes/class-donations.php
thomasguillot: Awesome, thank you @adekbadek.

 > WDYT about renaming the image file to something like "placeholder-image"

Done in https://github.com/Automattic/newspack-plugin/pull/3109/commits/8c1236d95615f4e99b48c06ef285a1e9ceb7af99

> These instances should be updated to use the placeholder image introduced in this PR

I'll create a new issue for the `newspack-blocks` patterns 👍

> If a pattern was inserted, and then updated in this repository, will it be updated on a post that used the pattern.

No, the pattern will be the old one.

PR #3110, opened 2024-05-13
fix: revert #3072

dkoo:
Reverts #3072 so we can have more time to craft a communication plan about this breaking change.


dkoo:
Superseded by #3111

PR #3114, opened 2024-05-14
feat(ras-acc): add segment control & build out button styles

laurelfulford:
### All Submissions:

_Note: This PR depends on the changes in https://github.com/Automattic/newspack-plugin/pull/3091 to be merged into it (like some class name changes), so I'm leaving it as a draft 'til then._ 

This PR adds the following block styles plus examples of each:

* It adds the small and x-small button styles for the regular buttons (medium is the default).
* It adds the small and medium button styles for icon buttons (x-small is the default). 
* It adds examples of buttons with icons, and tweaks the spacing.
* It adds the 'segment component' which kind of uses button styles combined with tabbed behaviour. 
* It adds a loading style for the button.

2. Review the X-Small and Small button examples, which include icons -- these appear after the 'wide' buttons. Because it's the most commonly used size, the 'Medium' size is the default for the regular buttons.

image

3. Review the Loading button examples -- these appear next. The animation is added through CSS, but the markup relies on a `<span>` being added around the button's text -- open to feedback on this specific point!

image

4. Review the Segment Controls -- there should be two examples, one with actual tab content, and one that just acts like free-standing buttons:

image

At the very bottom, there's also an example of what this component looks like inside of a modal (which better matches how it's used in the My Account mockups):

image

5. Lastly, review the icon buttons -- a new Small and Medium size have been added. The regular buttons use 'Medium' as a default, but the icon button uses X-Small -- for both, this is because they're the most-used size, but it may be unnecessarily confusing. I'd appreciate some any feedback around how best to balance the defaults, to make sure things make sense and are easy to use:

image

### Other information:


dkoo (approved):
Looks great! Tested in both desktop and mobile browsers.

assets/newspack-ui/js/segmented-control/index.js
```diff
@@ -0,0 +1,34 @@
+import { domReady } from '../utils'; // Global utils.
+
+domReady( function () {
+	function segmented_control( element ) {
+		const header = element.querySelector( '.newspack-ui__segmented-control__tabs' );
```
dkoo, line 5: Non-blocking nit: can we always assume these child elements will exist inside a segmented control element? (e.g. header, tabs, etc.)

laurelfulford: @dkoo We should be able to assume the header and tabs will be there, but I'm happy to add checks if it'd be better! 

I discussed this component a bit with @thomasguillot and it won't always have the content/panels, it can just be tabs acting more like a standalone radio component. 

This needs a yay/nay from design still, so I'm happy to make tweaks in the meantime!

thomasguillot:
@laurelfulford I've just push a small change to the loading button. I also noticed we have `.newspack-ui__spinner`. I'm not sure I understand what this component is (like why do we need a border-radius and an opacity for example).

thomasguillot:
@laurelfulford I pushed more changes. It's looking good. And my only question mark is about the `.newspack-ui__spinner` but it's not really part of this PR anyway.

thomasguillot:
✅ Design

laurelfulford:
Thanks @thomasguillot and @dkoo! I'm going to go ahead and merge this as is, but @dkoo let me know if there's any checks you'd recommend!

> I also noticed we have .newspack-ui__spinner...

I know this is used for when the modal is loading the checkout, but I'm not super familiar with why the styles are as they are. If we do end up swapping it with a skeleton loader in the checkout it may not be needed, but in the meantime we could also pare it down and get rid of anything that's not needed in a separate PR 🙂

PR #3118, opened 2024-05-15
feat(ras-acc): add basic SVG handling?

laurelfulford:
### All Submissions:

I'm not sure we actually need this, but this PR takes a swing at consolidating all the SVGs used in the Newspack UI, and sets up a couple function to return or echo them.

This is a riff off of what's in the Newspack Theme, and might be overkill/clunky - it could also work to centralize and add on the individual functions that return the markup, [like we're already doing in RAS-ACC](https://github.com/Automattic/newspack-plugin/blob/c64f20657436212b8412c5c41baa36a11e12414a/includes/reader-activation/class-reader-activation.php#L939-L946) -- open to feedback on that! 

I haven't made any changes on the Newspack Blocks side yet; the duplication seems like an okay trade off for still having the SVGs work when the Newspack Plugin is disabled, but if this seems like a good plan we could do something similar there for the checkout-related SVGs.

1. Apply the PR.
2. Check the `?ui-demo` and make sure the SVGs are loading.
3. View as a non-logged in user and confirm the Sign In link in the header still has the icon.

### Other information:


adekbadek (approved):
Nice! I think it's a needed change, makes the code much more readable and abstracts away the clunkiness of SVG markup.

laurelfulford:
Thanks @adekbadek! 🙌

PR #3122, opened 2024-05-20
feat(cli): enable running the setup with a site import

adekbadek:
### All Submissions:

Enables running the `wp newspack setup` command a `--site` param to specify a (WP) site to import initial posts from. 

1. Run `wp newspack setup --site=<wordpress-site-url>` and observe the initial posts are not auto-generated, but fetched from the site

### Other information:

### DIFF
--- a/includes/cli/class-setup.php
+++ b/includes/cli/class-setup.php
@@ -92,7 +92,7 @@ class Setup {
 	 * Populates the site with initial content
 	 *
 	 * @param array $assoc_args Assoc args passed to the CLI invocation.
-	 * @return void
+	 * @return void|WP_Error Error on failure.
 	 */
 	private function initial_content( $assoc_args ) {
 		WP_CLI::line( 'Creating Initial Content' );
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works! Non-blocking suggestions while we're here:

- It wasn't clear to me at first that the `https://` is required as part of the arg. Maybe we could just append it if the arg is passed with just the domain?
- It failed to create any content but gave no feedback if the site URL given doesn't exist or isn't a WP site. Maybe we could check the [response](https://github.com/Automattic/newspack-plugin/pull/3122/files#diff-82295c79c9036534607a98913bbb10e06ef8c3f418766076a0a945cc7da92dd1R108) and abort with an error if the `Starter_Content_WordPress` class fails to fetch any post data?

adekbadek:
@dkoo – fixed both in 749b32ba58146ea091acd90ce32ab269768275bf

dkoo (approved):
Thanks for the updates!

PR #3126, opened 2024-05-22
feat: add support for reCAPTCHA v2

dkoo:
### All Submissions:

Adds optional support for reCAPTCHA v2 ([invisible](https://developers.google.com/recaptcha/docs/invisible) flavor only) in addition to the already-supported v3. 

Screenshot 2024-05-23 at 2 42 23 PM

@Automattic/newspack-design I could use a design review from you on the new admin UI for managing the different versions, as well as the positioning of the checkbox widget in the different contexts (screenshots below).

Closes `1206943664367857/1207301960751653`.

1. Start with a site that already has reCAPTCHA v3 configured and working. Check out this branch as well as https://github.com/Automattic/newspack-newsletters/pull/1514 and https://github.com/Automattic/newspack-blocks/pull/1744.

2. Visit [Google's reCAPTCHA admin](https://www.google.com/recaptcha/admin/create) and create a new account: 
  - Type set to **Challenge (v2)** + **Invisible reCAPTCHA badge**
  - After creating, go to settings and add your test site's URL as a domain. You can also optionally bump the "Security" slider all the way to the right to make it more likely that you'll see the reCAPTCHA challenge on the front-end.

Screenshot 2024-05-23 at 2 44 50 PM

3. In WP admin, visit **Newspack > Connections** and scroll down to the reCAPTCHA section. Confirm the new "reCAPTCHA version" option, and that it's set to v3 since you had it configured already. Leave v3 enabled for now.

4. As a reader, perform the following actions, each in a new session, and confirm that you're able to complete each one:
  - Register via the Registration block
  - Register via the "Sign In" auth modal
  - Log into an existing account via the "Sign In" auth modal (note: login will not be checked by reCAPTCHA)
  - Sign up for newsletter lists via the Newsletter Signup block
  - Purchase a product via the modal checkout (Donate or Checkout Button blocks)
  - **Note:** For v2 invisible and v3 reCAPTCHA types, you may not be presented with any challenge or feedback if the initial check succeeds. To confirm that you're being tested and verified by reCAPTCHA, you can add `error_log( print_r( $captcha_verify, true ) );` just after [this line](https://github.com/Automattic/newspack-plugin/blob/feat/recaptcha-v2/includes/class-recaptcha.php#L364) to see the results of the verification request.

5. Back in WP admin, select the "v2 invisible" option (confirming that the v3-only "Threshold" option disappears) and paste your site key + secret from the invisible account you created in step 2. Save.

6. Repeat the actions from step 4 as a reader and confirm you're able to complete each one.
  - For v2 invisible reCAPTCHA, you may be asked to complete a challenge like this if the initial verification fails upon form submission. Once you complete the challenge successfully, the form should automatically submit.

Screenshot 2024-05-22 at 12 21 48 PM

7. Smoke test: disable reCAPTCHA in the Connections wizard by toggling it off, and confirm you can still complete all actions as a reader.

### Other information:

---
  - https://app.asana.com/0/0/1207301960751653


kevinzweerink:
I thought we were going to use the solution proposed in this thread: https://a8c.slack.com/archives/C015W6BES8J/p1715880136662899

This solution is pretty far from ideal from a design perspective, and will also degrade the user experience in the new RAS-ACC flows. It adds another variable to the end user experience that we'll have to consider every time we touch these flows. I hate to be too negative when it's clear that work has already been done here, but is there any way we can avoid this?

dkoo:
@kevinzweerink @thomasguillot 3d767dd removes support for the checkbox flavor of v2.

dkoo:
To follow up, after some internal discussion, we've opted to implement support for the "invisible" flavor of v2 reCAPTCHA only, due to the visual design and UX concerns that are associated with the "I am not a robot" checkbox flavor.

This is also in lieu of implementing our own OTP-based challenge in the event of a v3 failure, which currently would only be available to sites using Reader Activation features.

assets/wizards/connections/views/main/recaptcha.js
```diff
@@ -95,18 +99,47 @@ const Recaptcha = () => {
...
+										{ __( 'Learn more about reCAPTCHA versions', 'newspack-plugin' ) }
+									</ExternalLink>
+								}
+							>
+								<SelectControl
+									label={ __( 'reCAPTCHA Version', 'newspack-plugin' ) }
+									hideLabelFromVision
+									value={ settingsToUpdate?.version || 'v3' }
+									onChange={ value => {
+										console.log( value );
```
adekbadek: Should be removed (we need to update eslint config!)

adekbadek (commented):
> we've opted to implement support for the "invisible" flavor of v2 reCAPTCHA only

This means that this PR will be getting an update, is that right?

assets/wizards/connections/views/main/recaptcha.js
dkoo: Whoops, thanks for catching! Removed in 8dd15857

dkoo:
> This means that this PR will be getting an update, is that right?

@adekbadek No, I opted to keep the back-end logic that supports the v2 Checkbox flavor in place in case we need to support it in the future. I simply removed it as an option to implement from the UI, so it's now only possible to choose between v3 vs. v2 invisible.

dkoo:
@adekbadek any other feedback on this one?

adekbadek:
I see "Needs Design Review" badge here, so I refrained from code review, just posted my comment.

dkoo:
@adekbadek I removed the "Needs Design Review" tag since [this decision](https://github.com/Automattic/newspack-plugin/pull/3126#issuecomment-2128179869) satisfied the design concerns, from what I understand.

adekbadek (commented):
I could not complete the testing steps locally or on a temporary publicly available site. I never receive a token in `getCaptchaToken` function, despite the promise resolving correctly (`token` is `null`). I'll DM the test site creds so it can be debugged.

dkoo:
Because the toggle button selector can be slightly confusing with only two options (which one is selected?), 24a6dbe5 changes the admin UI to use a standard dropdown selector instead of the toggle button.

Screenshot 2024-06-04 at 1 44 15 PM

dkoo:
> I never receive a token in `getCaptchaToken` function, despite the promise resolving correctly (`token` is `null`).

The token fetched via JS API is only needed for the V3 integration. V2 automatically appends a token to the submitted form data, so `getCaptchaToken` resolves with an empty string if using V2. 1423f10 attempts to clarify this with inline documentation + renaming the function.

adekbadek (commented):
1. When testing the donation flow using V3, after submitting the payment details the flow is blocked with a "Please complete the challenge to continue." message (I can't see any challenge):

image

2. I was confused by the V3/V2 options and misconfigured it on the first try. I think what would help is aligning the wording with what's visible when creating the credentials: "Score based (v3)" & "Challenge (v2) - invisible version". Also, I think the "Your site key and secret must match the selected reCAPTCHA version." message should be more prominent (a notice, not a help text. If it's possible to validate the keys on save, that would also help.

dkoo:
> I think what would help is aligning the wording with what's visible when creating the credentials: "Score based (v3)" & "Challenge (v2) - invisible version". Also, I think the "Your site key and secret must match the selected reCAPTCHA version." message should be more prominent (a notice, not a help text. If it's possible to validate the keys on save, that would also help.

Good ideas—done in 782ad13. Unfortunately Google doesn't provide any way to validate reCAPTCHA keys. In the future we could try adding a fake test form to check this, but for now we just clear out the key + secret fields when switching versions and show an error to explain.

> When testing the donation flow using V3, after submitting the payment details the flow is blocked with a "Please complete the challenge to continue." message (I can't see any challenge):

I was running into the dreaded `notoptions` bug when testing today, which was causing some of my options to not update correctly (so you may have changed the `version` setting to v3, but since the option didn't exist before the `update_option` function was failing). I added a cache clear in f3c4811. After this I'm able to switch versions and change key/secret as expected, and the front-end behaves as expected too.

includes/class-recaptcha.php
```diff
@@ -7,22 +7,33 @@
...
 /**
  * Class for reCAPTCHA integration.
  */
 final class Recaptcha {
-	const SCRIPT_HANDLE  = 'newspack-recaptcha';
-	const OPTIONS_PREFIX = 'newspack_recaptcha_';
+	const SCRIPT_HANDLE      = 'newspack-recaptcha';
+	const SCRIPT_HANDLE_API  = 'newspack-recaptcha-api';
+	const OPTIONS_PREFIX     = 'newspack_recaptcha_';
+	const SUPPORTED_VERSIONS = [ 'v3', 'v2_checkbox', 'v2_invisible' ];
```
adekbadek: Is `v2_checkbox` intended here? It's not available in the Connections wizard settings.

adekbadek (approved):
Pre-approving as it works now as expected, but left one question.

includes/class-recaptcha.php
dkoo: Yeah, since we originally went through the trouble of adding an implementation for the v2 Checkbox flavor, I left the infrastructure in place so it'd be easy to reenable in the future if we ever want to. All we'd have to do is readd it as an option in the Connections UI.

dkoo: 4f51b03c removes the string from the `SUPPORTED_VERSIONS` constant, but adds a note in the relevant areas of the code where we could add it to reenable support for the Checkbox flavor.

PR #3137, opened 2024-05-24
fix(tracking): handling user role in pixel

adekbadek:
### All Submissions:

Fixes an issue with how the user role was retrieved in the pixel tracking code. Also refactors the code to avoid unnecessary code execution.

1. On `trunk`
1. Create a new account via RAS and visit /my-account page
2. Observe a PHP Warning logged about retrieval of index `0` while there's none*
3. Switch to this branch, observe no issues logged

\* because `$current_user->roles` in `Meta_Pixel->get_event_params` is `[1] => 'subscriber'`

### Other information:

### DIFF
--- a/includes/tracking/class-meta-pixel.php
+++ b/includes/tracking/class-meta-pixel.php
@@ -41,7 +41,7 @@ class Meta_Pixel extends Pixel {
 		$current_user = wp_get_current_user();
 		$event_params = [
 			'page_title' => get_the_title(),
-			'user_role'  => empty( $current_user->roles ) ? 'guest' : array_shift( $current_user->roles ),
+			'user_role'  => empty( $current_user->roles ) ? 'guest' : reset( $current_user->roles ),
 			'event_url'  => home_url( $wp->request ),
 		];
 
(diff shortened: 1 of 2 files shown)


includes/tracking/class-meta-pixel.php
```diff
@@ -38,7 +41,7 @@ private static function get_event_params() {
 		$current_user = wp_get_current_user();
 		$event_params = [
 			'page_title' => get_the_title(),
-			'user_role'  => empty( $current_user->roles ) ? 'guest' : $current_user->roles[0],
+			'user_role'  => empty( $current_user->roles ) ? 'guest' : array_shift( $current_user->roles ),
```
dkoo: Nit: since there's no need to alter the original array, `reset` might be a better choice instead of `array_shift`.

dkoo (approved):
Couldn't replicate the warning, but otherwise looks fine to me. 1 non-blocking suggestion below.

includes/tracking/class-meta-pixel.php
adekbadek: Ah, good call! Updated in 28f5df2c936fb5029f9a8d95fca5ae9d3f15d150

PR #3139, opened 2024-05-24
fix(ras): prevent email address exposure via user login

adekbadek:
### All Submissions:

Fixes the possibility of reader email address exposure through comments. 

1. On `trunk`,
2. Install and activate `woocommerce-memberships-for-teams` plugin
3. Register using the Registration block
4. Find the new user in WP Admin Users view - observe the "Display name publicly as" field is set to the email address
5. Switch to this branch, repeat, observe the publicly displayed name is set to the part of the email before the `@`

### Other information:

---
  - https://app.asana.com/0/0/1207301960751638

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1727,9 +1727,7 @@ final class Reader_Activation {
 			return $user_data;
 		}
 
-		$user_login      = str_replace( '+', '_', \sanitize_user( $user_data['user_email'], true ) ); // Matches the email address, but replace + with _ to allow for Gmail aliases.
-		$random_password = \wp_generate_password();
-		$user_nicename   = self::generate_user_nicename( ! empty( $user_data['display_name'] ) ? $user_data['display_name'] : $user_data['user_email'] );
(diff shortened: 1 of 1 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -1739,8 +1739,9 @@ public static function canonize_user_data( $user_data = [] ) {
 		$user_data = array_merge(
```
claudiulodro, line 1737: It doesn't let me make a comment higher up than this, but it looks like `$user_login` isn't used anywhere any more after this change, so should be removed. Not a blocker though.

claudiulodro (approved):
Looks good and works well. I like that [the logic is ported to Network also](https://github.com/Automattic/newspack-network/commit/9f74e75d65c6abeb8bcceca30299a533ddd1aed9#diff-e63d8e9b4c7fbd03f99fd8f88f433092543005b12120dc23030e1ca82f50ab89R48). I left one non-blocking note.

includes/reader-activation/class-reader-activation.php
adekbadek: Huh, good catch! Fixed in 742c30a13477b99e6f721e889f7407e655333814

PR #3146, opened 2024-05-28
Feat/rss vs wc memberships

adekbadek:
### All Submissions:

Adds an option to skip WC Memberships' content restrictions in RSS feeds. This is needed for publication which want to restrict content on the website, but still distribute it in news aggregation services.

Also, removes the WC Memberships meta box while editing an RSS feed, to avoid confusion.

1. Set up a WC Memberships plan with content restriction*
2. Visit the `/feed`, observe that not whole post content is sent in the payload 
3. Go to WooCommerce -> Settings -> Memberships and enable the "Skip content restriction in RSS feeds" setting
4. Refresh the feed, observe the full post content is available
5. Enable the RSS module by running `wp eval "\Newspack\Settings::activate_optional_module('rss');"`
6. Create or edit an RSS feed, observe there's no WP Membership meta box

\* restrict all posts for ease of testing

### Other information:

---
  - https://app.asana.com/0/0/1205279243070215

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -16,8 +16,9 @@ defined( 'ABSPATH' ) || exit;
  */
 class Memberships {
 
-	const GATE_CPT  = 'np_memberships_gate';
+	const GATE_CPT = 'np_memberships_gate';
 	const CRON_HOOK = 'np_memberships_fix_expired_memberships';
+	const SKIP_RESTRICTION_IN_RSS_OPTION_NAME = 'newspack_skip_content_restriction_in_rss_feeds';
 
 	/**
 	 * Whether the gate has been rendered in this execution.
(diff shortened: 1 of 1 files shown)


includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -997,5 +1001,62 @@ public static function remove_frontpage_content_restriction() {
...
+	 *
+	 * @param array $settings associative array of the plugin settings.
+	 */
+	public static function wc_memberships_general_settings( $settings ) {
+		$setting = [
+			'type'    => 'checkbox',
+			'id'      => self::SKIP_RESTRICTION_IN_RSS_OPTION_NAME,
+			'name'    => __( 'Skip content restriction in RSS feeds', 'newspack-plugin' ),
+			'desc'    =>
+				'<span class="show-if-hide-content-only-restriction-mode">' . __( 'If enabled, full content will be availabe in RSS feeds.', 'newspack-plugin' ) . '</span>',
```
dkoo:
```suggestion
				'<span class="show-if-hide-content-only-restriction-mode">' . __( 'If enabled, full content will be available in RSS feeds.', 'newspack-plugin' ) . '</span>',
```

dkoo (approved):
I'm approving since this mostly works, but I was having some trouble that I'm not sure is due to my test site setup or something else:

- On my local dev site, I had trouble getting my feeds to display full content instead of excerpts no matter which settings I enabled in WooCommerce Memberships settings or general WP > Reading settings.
- On another test site, updates seem to take a few minutes to take effect. Probably expected since RSS feeds are only regenerated on an interval, right?

adekbadek:
> On my local dev site, I had trouble getting my feeds to display full content instead of excerpts no matter which settings I enabled in WooCommerce Memberships settings or general WP > Reading settings.

To clarify, this only happened on a local dev site, but did not when testing on another test site?

> updates seem to take a few minutes to take effect. Probably expected since RSS feeds are only regenerated on an interval, right?

In my testing the updates were not delayed. Perhaps some caching was in place on the site you've tested with?

dkoo:
> To clarify, this only happened on a local dev site, but did not when testing on another test site?

That's right—since I didn't see it on a fresh test site, I think it's a peculiarity of my dev environment, but wanted to mention just in case we see it happen elsewhere after rollout.

dkoo (approved):

PR #3149, opened 2024-05-29
fix(ras-acc): fix focus, Google button styles

laurelfulford:
### All Submissions:

This PR fixes a couple smaller display issues I noticed while working on RAS-ACC:

* It fixes the checkbox/radio button focus style, which was getting overridden by the theme.
* It fixes the Google button in the registration block, which I broke by making the Newspack UI button styles more specific 😅 

See: 1205234045751551-as-1207417756735456 and 1205234045751551-as-1207417756735446

_To test, you need to be on a test site with Google OAuth set up._ 

1. If not already, navigate to Customizer > Colors, and set your colours to a non-default colour -- ideally something bright but darker so it's not switched to grey by the theme's contrast logic. 
2. Edit a page, and add the reader registration block; publish and append the `?ui-demo` query string at the end of the URL.
3. Check the registration block and note the funky appearance of the Google button; the size will be all wrong and it will be using the site's secondary colour:

image

4. Navigate down to the checkboxes and click them to see the focus style - rather than using the dark grey from Newspack UI, it'll use the theme's secondary colour:

image

5. Apply this PR and run `npm run build`.
6. Confirm that the Google button in the reader registration block is full width and light grey:

image

7. Confirm that the 'focus' style isn't applied when the checkboxes are clicked, but it is applied when you tab-focus on them. The focus style should be 2px and black; when the checkbox is selected it should have a slight gap between the outline and the background, but when they're not it should be right on the line:

image

image

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/elements/forms/_checkbox-radio.scss
+++ b/assets/newspack-ui/scss/elements/forms/_checkbox-radio.scss
@@ -26,6 +26,10 @@
 			}
 		}
 
+		&:focus {
+			outline: unset;
+		}
+
 		&:focus-visible {
 			outline: 2px solid var( --newspack-ui-color-neutral-90 );
 			outline-offset: 1px;
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1424,12 +1424,14 @@ final class Reader_Activation {
 			return;
 		}
 		?>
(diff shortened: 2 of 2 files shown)


chickenn00dle (approved):
Looks good!

laurelfulford:
Thanks @chickenn00dle! 🙌

PR #3155, opened 2024-05-31
fix(pwa): disable post request interception

jaredrethman:
### All Submissions:

This pull request resolves an issue in the WordPress PWA plugin where the service worker incorrectly handles WP REST API errors by intercepting requests and returning misleading HTTP statuses. By removing `wp-offline-post-request-handling`, REST API requests directly reach the server, ensuring accurate error propagation and improving reliability for developers and end-users.

1. Ensure PWA plugin is installed and active.
2. Navigate to `/wp-admin/admin.php?page=newspack-connections-wizard#/`
3. Open Network tab in Developer Tools
4. Disconnect MailChimp by clicking "Disconnect" to the right of the card.
5. Click Connect to open Modal.
6. Attempt connection with incorrect API key (i.e. force an error). See below:
Screenshot 2024-05-31 at 09 26 58
7. Click "Connect".
8. Observe in the Network tab that a fetch request has been made by the PWA service worker which has returned a 200 OK status code, however when previewing the response it is immediately clear that the request has resulted in a 500 server error derived from the MailChimp API package https://github.com/drewm/mailchimp-api/blob/master/src/MailChimp.php#L49.
Screenshot 2024-05-31 at 09 18 17
10. Observe the error displayed is "The response is not a valid JSON response.", indicative that there's been an additional error in parsing the HTML response as JSON.
Screenshot 2024-05-31 at 09 18 34

**Now to test fix:**

1. Checkout this branch.
2. Go back to the browser and clear all Application cache under Application tab in Developer Tools by clicking "Clear site data". Note: this action will log you out of wp-admin.
Screenshot 2024-05-31 at 09 43 08
3. Refresh your browser, this will redirect to login, login and redirect back to Connections Wizard.
4. Click Connect to open Modal.
5. Again, attempt connection with incorrect API key.
6. This time you should see the correct error being displayed:
Screenshot 2024-05-31 at 09 48 34

_Note: if you receive the same result as in test 1, attempt a hard refresh_

### Other information:


includes/class-pwa.php
```diff
@@ -25,6 +26,19 @@ public static function init() {
...
 		add_filter( 'wp_service_worker_error_messages', [ __CLASS__, 'error_messages' ] );
+
+		/**
+		 * Temporary workaround to disable the offline post request handling script.
+		 *
+		 * @see - https://github.com/GoogleChromeLabs/pwa-wp/issues/1106
+		 */
+		add_action(
+			'wp_front_service_worker',
+			function ( WP_Service_Worker_Scripts $scripts ) {
```
dkoo: This is a supreme nitpick, but let's stick with the prevailing pattern we tend to use for adding hook callbacks via static class methods.

dkoo (changes requested):
@jaredrethman strangely, with the PWA plugin active, I'm not seeing the `wp_front_service_worker` action or your callback fire. 

Also had a couple of questions:

1. Is there any known downside to adding this fix? It sounds from the [linked thread](https://github.com/GoogleChromeLabs/pwa-wp/issues/1106#issuecomment-1992829500) that the behavior this fixes is unintended, so I don't think so, right?
2. Should we rebase this fix on `trunk` since it has nothing to do really with the IA project and the main codebase would probably benefit from it?

And one very minor nitpick below.

jaredrethman:
Thanks for the review @dkoo . Some responses below:

> * Is there any known downside to adding this fix? It sounds from the [linked thread](https://github.com/GoogleChromeLabs/pwa-wp/issues/1106#issuecomment-1992829500) that the behavior this fixes is unintended, so I don't think so, right?

The behavior is unintended, but the route to remedy it is quite invasive IMO i.e. Plugin does not handle 500 errors correctly, so just disable ALL post request handling for the service worker. The down side is no POST requests are going to be handled by a service worker, this is most notable for offline experiences, since the service worker would act as Stale-While-Revalidate mechanism for any POST requests that take place offline.

> * Should we rebase this fix on `trunk` since it has nothing to do really with the IA project and the main codebase would probably benefit from it?

I was hesitant to do this on `trunk` since I want to spend the rest of the engineering effort on the IA feature testing this fix. But I totally agree with what you're saying. I tried to rebase, but failed, so I manually rebased to `trunk` :) 

> This is a supreme nitpick, but let's stick with the prevailing pattern we tend to use for adding hook callbacks via static class methods.

I bounced back and forth on this (But going forward I will always do static methods instead of anonymous functions) as I knew this would be there temporarily - until a fix is released.

Thanks again!

dkoo (approved):
Thanks for the rebase!

PR #3164, opened 2024-06-03
feat(ras-acc): improving keyboard navigation

laurelfulford:
### All Submissions:

This PR, along with https://github.com/Automattic/newspack-blocks/pull/1760, adds some basic keyboard accessibility improvements to the reader auth and modal checkout windows. Specifically: 

* Each modal should have the attributes `role="dialog"` and `aria-modal="true"`.
* Each modal should have the modal header given an `id`, and that id should be referenced in the attribute `aria-labelledby` on the modal.
* Each modal has JavaScript that traps focus when they're opened, so when you're tabbing or shift-tabbing through the elements, the focus remains inside of the modal and doesn't go behind it.
* Each modal should be close-able with  the `esc` key.
* When each modal is closed, the focus should return to the button that opened the modal.

See 1205234045751551-as-1206049351621098.

1. Apply this PR and https://github.com/Automattic/newspack-blocks/pull/1760 and run `npm run build` for each.
2. On a page, add a donate block, a checkout button block with a single product, and a checkout button block with a product with variations. 
3. Test the following in an incognito window; do your best not to use your mouse to navigate/interact with the site. 

#### Testing Reader Registration

1. Using the 'Tab' key, navigate to the Sign In button in the header, and hit 'Enter'.
2. Confirm that the modal that opens gets focus (indicated by the bold outline on the button):

image

3. Tab through all of the elements, and confirm once you hit the bottom, you're brought back up to the close button at the top.
5. Shift-Tab backwards through the modal, and confirm once you hit the Close button, you're brought down around to the bottom.
6. Tab back to the Close button and hit 'Enter'. Confirm that the keyboard focus is returned to the Sign In button.
7. Open the Sign In button again, and proceed to sign up for a new account.
8. Once done, close the incognito window, and open a new one.
9. Use the keyboard to again navigate to the Sign In button.
10. Sign in with the same email address you just created.
11. Confirm that you get the OTP screen.
12. Repeat steps 3 and 4 to Tab and Shift tab through the modal, and confirm that focus is still trapped.
13. Confirm that the modal has the attributes `role="dialog"` and `aria-modal="true"`.
14. Confirm that the modal header has an `id`, and that id should be referenced in the attribute `aria-labelledby` on the modal.
15. Click the Esc key to close the modal, and confirm that focus returns to the Sign In button.

#### Testing Donations

1. Using the Tab key, navigate to the Donate block, and use the Tab key and arrow keys to pick a specific donation.
2. Tab to the "Donate" button, and click Enter to submit it.
3. Confirm that the modal that opens gets focus; if you're not logged in, log in.
4. On the first Billing details screen, tab through all of the elements, and confirm once you hit the bottom, you're brought back up to the close button at the top.
5. Shift-Tab backwards through the modal, and confirm once you hit the Close button, you're brought down around to the bottom.
6. Tab back to the Close button, and hit enter; confirm the focus is returned to the Donate block.
7. Reopen the Donate modal, and fill in the first screen with billing details; tab to the Continue button and hit 'Enter'.
8. On the next screen, repeat steps 4 and 5 and confirm you loop through only the visible elements.
9. With focus inside of the checkout fields (and not on the Close button), hit the 'Esc' key and confirm the focus returns to the Donate block button.
10. Reopen the Donate modal.
11. Confirm that the modal has the attributes `role="dialog"` and `aria-modal="true"`.
12. Confirm that the modal header has an `id`, and that id should be referenced in the attribute `aria-labelledby` on the modal. 
13.  Fill in the second screen. Tab to the Submit button, and click 'Enter'.
14. On the confirmation window, repeat steps 4 and 5 and confirm that focus remains in the modal.
15. Tab to the Continue button, and click 'Enter'. 
16. On the newsletter Coda screen, repeat steps 4 and 5 and confirm that focus remains in the modal.

#### Testing Checkout Button Block - simple product 

Repeat the testing steps from the Donate block, but using a Checkout Button block with a simple product.

#### Testing Checkout Button Block - variable product

Repeat the testing steps from the Donate block, but with a Checkout Button block with a variable product. The one difference is tab and shift-tab through the first variation-picker screen, and confirm that the focus remains inside of the modal.

### Other information:


src/reader-activation-auth/accessibility.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Trap focus in the modal when opened.
+ * See: https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700
+ */
+export function trapFocus( currentModal ) {
+	const focusableEls =
+		'button, [href], input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])';
+	const visibleFocusableEls = [];
+
+	const firstFocusableEl = currentModal.querySelectorAll( focusableEls )[ 0 ]; // get first element to be focused inside modal
```
chickenn00dle, line 10: We probably won't run into this case since we control the markup of our modals, but generally, we'd want to confirm the length of this query before accessing one of its elements.

```suggestion
	const firstFocusableEl = currentModal.querySelectorAll( focusableEls )?.[ 0 ]; // get first element to be focused inside modal
```

chickenn00dle: We probably won't run into this case since we control the markup of our modals, but generally, we'd want to confirm the length of this query before accessing one of its elements.

```suggestion
	const firstFocusableEl = currentModal.querySelectorAll( focusableEls )?.[ 0 ]; // get first element to be focused inside modal
```

src/reader-activation-auth/accessibility.js
```diff
@@ -0,0 +1,42 @@
...
+	const focusableElsAll = currentModal.querySelectorAll( focusableEls );
```
chickenn00dle, line 10: Non-blocking nit: Is there a reason we are running the same call to `querySelectorAll` here? Seems like we can run it once, then just get the first element.

chickenn00dle (changes requested):
Nice work @laurelfulford!

Works great for the most part, but did leave a few comments and ran into a few issues:

1. Looks like I am unable to check or uncheck any checkboxes using the enter key. For example, the subscriptions gifting checkbox in the checkout flow:

Screenshot 2024-08-20 at 12 31 37

Or the newsletters checkboxes post auth/checkout:
Screenshot 2024-08-20 at 12 35 44

2. Step 1 of the checkout modal (billing details) does not allow me to cycle backwards with `shift` + `tab` once I get to the close modal icon:
Screenshot 2024-08-20 at 12 41 13

3. I am unable to select the saved payment method/new payment method radios by tabbing. Only the preselected option is tabbable, every other option is skipped:
Screenshot 2024-08-20 at 12 42 47

4. Should the variations modal have the same `role`, `aria-modal`, and `aria-labeledby` attributes as the other modals?
Screenshot 2024-08-20 at 12 36 27

src/reader-activation-auth/accessibility.js
laurelfulford: Oh yeah, that's not great -- this has been updated!

src/reader-activation-auth/accessibility.js
laurelfulford: Thanks Raz! This has been updated!

laurelfulford:
Thanks for the review @chickenn00dle!

For items 1 and 3, I should've included more information about how these are expected to work:

1 - When navigating checkboxes, you need to use the `space` bar to select/deselect.
3 - When navigating radio buttons, tabbing will just bring you to the selected radio option, but that also puts focus inside of the set of radio buttons. Once focus is on the selected radio, you need to use the arrow keys to navigate between the other options, and the `space` bar to select a new option.

There's a bit more information about how keyboards navigate those and other form elements [here](https://webaim.org/techniques/keyboard/#testing). Some are definitely not what you'd expect -- I submitted the form a few times when trying to select that shipping checkbox! 

2 - The Shift-Tab for the first screen on the checkout modal should be fixed in the Newspack Blocks version of the PR.

4 - The variations modal should definitely have those attributes, too -- thanks for catching that 🤦‍♀️ They've been added in the Newspack Blocks version of the PR.

Just let me know if you have questions about any of these!

src/reader-activation-auth/accessibility.js
```diff
@@ -0,0 +1,42 @@
...
+
+	// Get rid of elements that aren't visible:
+	focusableElsAll.forEach( function ( el, index ) {
+		if ( el.offsetParent !== null ) {
+			visibleFocusableEls[ index ] = el;
+		}
+	} );
+
+	const lastFocusableEl = visibleFocusableEls[ visibleFocusableEls.length - 1 ]; // get last element to be focused inside modal
+	firstFocusableEl.focus();
```
chickenn00dle, line 27: Looking at this again, it looks like this all depends on `focusableElsAll` not being empty. I think we can return early before these lines if that is empty. Otherwise we will have to add multiple checks for empty or `undefined` values throughout the code.

chickenn00dle (changes requested):
Nice! Thanks for clarifying @laurelfulford!

This works as expected then 🎉 

Just one final comment about handling potentially `undefined` or empty values in the a11y script.

src/reader-activation-auth/accessibility.js
laurelfulford: Ah good call, @chickenn00dle! I've added a check to this PR, and one to the Newspack Blocks as well. I'll wait and see if this one looks good before merging either -- thanks!

chickenn00dle (approved):
Thanks @laurelfulford! Looks good 👍

laurelfulford:
Thank you! 🎉

PR #3163, opened 2024-06-03
feat(ia): added `useWizardApiFetch` hook and `<WizardsActionCard />` & `<WizardsPluginCard />`

jaredrethman:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

#### Redux store:
- Added `getWizardData` selector to allow data retrieval without making a request i.e. `getWizardAPIData`.
- Added `isLocalError` to flag errors needing to be thrown and bubble up to be caught by consuming component/hook.

#### (new) `useWizardApiFetch` hook:
- Provides a way to make API fetch requests using the Wizard API Redux store. It handles caching, error management, request status and state updates efficiently.
- Caches promise and responses to avoid unnecessary API calls

#### `<WizardActionCard />`:
- Houses a Wizard specific abstraction of the `<Card />` component.
- Utilizes `notification` property to display errors specific to that instance of the component.

#### `<WizardPluginCard />`:
- Unifies common card+plugin usages for easier reuse.
- Utilizes `useWizardApiFetch` to make plugin requests.

#### Connections:
- Contains updates relevant to testing i.e. demo purposes.
- Note: all demo content will be removed on future iterations of the IA feature.

### How to test the changes in this Pull Request:

1. Checkout this branch and build assets
2. Navigate to `/wp-admin/admin.php?page=newspack-settings#/`
3. Observe the JSON output at the top of the page (beneath the "Connections" title)
Screenshot 2024-06-03 at 09 21 23
4. Once the request has finished, you should see the below
np-ia local_wp-admin_admin php_page=newspack-settings (4)
5. Observations (confirm all):
   - We have forced an error when hitting "/newspack/v1/plugins/google-site-kit**s**", this should be present under `error` property. 
   - Confirm `statusTwo` property value is `"Error"` too, this represents the outcome of the forced error.
   - When request runs, ensure `isFetching` state updates according to the async fetch cycle of the request.
   - Confirm when switching tabs that the cache/state is not purged and all info (including errors) is retained between transitions

### Other information:

[...]

PR #3167, opened 2024-06-05
feat(ia): Settings / Connections - APIs

jaredrethman:
### All Submissions:

* Adds Google OAuth and Mailchimp components to APIs section.
* Google OAuth: modernized `getURLParams` function and eliminated the need for the `qs` package.
* Replaced deprecated conditional props on  `<Button />` with preferred `variant` prop.
* Updated `/oauth/mailchimp` to have consistent return type i.e. `{}` not `{}|[]`. 

**Original location:**
* Connections Wizard: _/wp-admin/admin.php?page=newspack-connections-wizard_

1. Checkout `feat/ia-settings-connections__apis` and build assets.
3. For the below APIs confirm you see one (or both, requires Connect/Disconnect)

#### Google OAuth:

**_NOTE:_**
_I was unable to authenticate with Google OAuth (due to closed-source dependencies). I circumvented all server-side conditionals/guardrails to test._

**Unauthenticated:**
Screenshot 2024-06-05 at 16 21 53

**Authenticated:**
Screenshot 2024-06-05 at 16 15 09

#### Mailchimp:

**Unauthenticated:**
Screenshot 2024-06-05 at 16 25 01

**Authenticated:**
Screenshot 2024-06-05 at 16 24 08

**Additional steps:**
- Confirm Disconnect & Connect call to actions work as intended.

### Other information:


assets/wizards/newspack/views/settings/connections/google-oauth.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Internal dependencies.
+ */
+import { Button } from '../../../../../components/src';
+import WizardsActionCard from '../../../../wizards-action-card';
+import { useWizardApiFetch } from '../../../../hooks/use-wizard-api-fetch';
+import { WizardError, WIZARD_ERROR_MESSAGES } from '../../../../errors';
+
+function getURLParams() {
```
chickenn00dle, line 19: Non-blocking question: Can we use react routers [useSearchParams](https://reactrouter.com/en/main/hooks/use-search-params) or [useParams](https://reactrouter.com/en/main/hooks/use-params) here?

assets/wizards/newspack/views/settings/connections/google-oauth.tsx
```diff
@@ -0,0 +1,169 @@
...
+		);
+		wizardApiFetch< string >(
+			{
+				path: '/newspack/v1/oauth/google/start',
+			},
+			{
+				onSuccess( url ) {
+					/** authWindow can be 'null' due to browser's popup blocker. */
+					if ( authWindow ) {
+						authWindow.location = url;
```
chickenn00dle: On Firefox, the auth window opens but the location is never updated, and so I end up with a popup of the homepage:

Screenshot 2024-06-24 at 16 04 39

We should probably not trigger the opening of the popup until the api fetch is successful and we are certain there are no popup blocker issues, etc.

chickenn00dle (changes requested):
Another one looking good @jaredrethman! 

Left one bit of feedback about the auth window then this should be good to go.

assets/wizards/newspack/views/settings/connections/google-oauth.tsx
jaredrethman: Thanks for the suggestion @chickenn00dle !

- `useSearchParams` is only available in react-router-dom v6, and we are currently using v5.
- `useParams` handles path parameters, not query parameters.

Given our current setup, using a custom function like getURLParams is straightforward and avoids adding unnecessary complexity or dependencies.

assets/wizards/newspack/views/settings/connections/google-oauth.tsx
```diff
@@ -0,0 +1,170 @@
...
+	function openAuth() {
+		let authWindow: Window | null = null;
+		wizardApiFetch< string >(
+			{
+				path: '/newspack/v1/oauth/google/start',
+			},
+			{
+				onSuccess( url ) {
+					authWindow = window.open(
+						'about:blank',
```
chickenn00dle: We should be able to just open with `url` directly here. I don't think there is a reason to wait until the popup opens to update the location, is there?

assets/wizards/newspack/views/settings/connections/google-oauth.tsx
chickenn00dle: Thanks @jaredrethman! 

This is working as expected now. It looks like there was an issue with my local site config where the `NEWSPACK_GOOGLE_OAUTH_PROXY_OVERRIDE` was not correct which was causing the popup location not to update.

This did get me thinking though, right now if the popup fails to open and `authWindow` is null we simply abort with no additional error message. We should probably handle this case by notifying the publisher that they should disable any popup blocker, etc. and try again.

WDYT?

chickenn00dle (changes requested):

assets/wizards/newspack/views/settings/connections/google-oauth.tsx
jaredrethman: @chickenn00dle great idea, I have gone ahead and added some error reporting for null `authWindow` and `url` values.

assets/wizards/newspack/views/settings/connections/google-oauth.tsx
jaredrethman: @chickenn00dle have updated.

chickenn00dle (approved):
Nice! Thanks for the changes @jaredrethman! Looks good 👍

PR #3168, opened 2024-06-05
feat(ia): Settings / Connections - reCAPTCHA v3

jaredrethman:
### All Submissions:

* Adds reCAPTCHA v3 section/component.
* Fixed issue where inputs would clear between saves.

**Original location:**
* reCAPTCHA V3: _/wp-admin/admin.php?page=newspack-connections-wizard_

1. Checkout `feat/ia-settings-connections__recaptcha` and build assets.
3. Depending on your stored reCAPTCHA settings you should see one of the following:

**Correct:**
Screenshot 2024-06-05 at 20 37 53

**Missing key (same result for missing secret)**
Screenshot 2024-06-05 at 20 39 05

**Disabled**
Screenshot 2024-06-05 at 20 45 28

4. Test Saving: input changes, input clearing and disabling, and confirm interaction works as expected.

**Additional steps:**

### Other information:


assets/wizards/newspack/views/settings/connections/recaptcha.tsx
```diff
@@ -0,0 +1,171 @@
...
+								onChange={ ( value: string ) =>
+									setSettingsToUpdate( { ...settingsToUpdate, site_secret: value } )
+								}
+								disabled={ isFetching }
+								autoComplete="off"
+							/>
+							<TextControl
+								type="number"
+								step="0.05"
+								min="0"
```
chickenn00dle: Question: Looks like it's possible to save this as 0 or empty, while on `trunk` this needs to be a non-zero value. I suspect we want to retain this non-zero behavior, but am not sure if there was some discussion around this. Is this an expected change?

chickenn00dle (changes requested):
Looks good for the most part @jaredrethman 👍 

I left a question about the threshold value before giving this an approval.

Aside from that, I have one more non-blocking nitpick regarding the error message that can be addressed outside of this PR. I know this is the existing behavior, but it bothers me that the error message pushes the form down whenever it appears:

Screenshot 2024-06-24 at 15 50 22

I know this has come up in another PR, so maybe this is something we should get feedback from @kevinzweerink or @thomasguillot on? I feel like moving the error message below the form is an easy fix, but it might even be worth creating an error popover that can serve to render all errors for these settings. WDYT?

assets/wizards/newspack/views/settings/connections/recaptcha.tsx
jaredrethman: Good question... did a little digging.

Found that there's some code missing between `trunk` and this branch introduced in https://github.com/Automattic/newspack-plugin/commit/876d0a9a1baa26480e399d52199d62d8a8f6bed5#diff-1833fd29b967510299e86f14a13280961ec87bb2a1813386f741228cbc2b12fc. I have synced these changes.

**RE `threshold` value and the misalignment between presentation and storage on `trunk` i.e. _/wp-admin/admin.php?page=newspack-connections-wizard#/:_**

* The default value of `threshold` is [0.5](https://github.com/Automattic/newspack-plugin/blob/feat/ia-settings-connections__recaptcha/includes/class-recaptcha.php#L126)
* Attempting to save `0` will save `0` in the DB (Side note: the error with inputs clearing when saved has been fixed in this PR). `wp option get newspack_recaptcha_threshold` will yield `0`. Screenshot of Adminer below.
Screenshot 2024-06-26 at 10 14 41
* When a zero value is retrieved for the `newspack/v1/recaptcha` route it is evaluated as `empty()` and the default value of `0.5` is applied and sent.
Screenshot 2024-06-26 at 10 22 03

Enforcing a non-negative float shouldn't be too difficult. Let me know if you'd like me to proceed?

chickenn00dle: Truthfully, I'm not too familiar with recaptcha, and so am not really sure what the expected behavior is here. Based on the `get_setting` code, it seems as though we always want a non-zero threshold (despite allowing this to be saved as zero). I would say let's enforce a non-negative, non-zero float, but lets let @dkoo chime in here since he is probably more familiar with the requirements.

dkoo: Hey guys, thanks for laying this all out. I think the intended behavior should be to enforce a non-zero float value, since `0` would effectively mean disabling reCAPTCHA entirely.

jaredrethman: Makes sense @dkoo , will update now. Will make `min="0.1"`

dkoo: Perfect, thanks!

jaredrethman: @chickenn00dle @dkoo I have added code introduced in #3126 and stringent validation for input fields in b7119362109bb0d9fe8857ee5932e526ad6336f4

A more verbose approach was necessary since only adding `min="0.1"` to the input field prevented values `<0.1` when using up/down arrow keys and up/down input UI Screenshot 2024-07-05 at 12 45 45.

chickenn00dle: > A more verbose approach was necessary since only adding min="0.1" to the input field prevented values <0.1 when using up/down arrow keys and up/down input UI Screenshot 2024-07-05 at 12 45 45.

Thanks for working on this @jaredrethman 

I'm a bit confused as to why we can't just set the min value here to `0.1`. It looks like the updated implementation returns an error anyway whenever we submit any value < 0.1 anyway, so we should probably just not let publishers input anything less than this in the first place. 

In any case, I'm happy to retain the additional validation you've added, but we should definitely still set this min value.

chickenn00dle (approved):
Thanks for working on this @jaredrethman, I'm gonna go ahead and give this one a thumbs up since the added validation gets this working as expected and the only change I'm asking for is pretty minimal. Assuming conflicts are resolved first of course.

PR #13, opened 2019-04-10
Basic plugin management through Plugins screen

claudiulodro:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes #10 .

Adds plugin recommendations/management through the admin Plugins screen. This interface is designed for power users or self-hosted users to be able to take advantage of and easily install our recommended plugins without requiring a Jetpack connection or hosting plan.

This initial implementation only recommends AMP and Jetpack; we should add other managed plugins after this is merged as a separate PR for easier reviewing.

#### Our managed plugins are nested under the Newspack plugin. Uninstalled plugins can be installed directly from the screen.
Screen Shot 2019-04-10 at 10 36 27 AM

#### AMP plugin just installed through this screen.
Screen Shot 2019-04-10 at 10 36 44 AM

#### AMP plugin just activated. It just uses existing Plugins screen functionality for plugins once they've been installed.
Screen Shot 2019-04-10 at 10 37 12 AM

#### Calypsoified view/WPCOM plugin management screen view. Everything works naturally on WPCOM since WPCOM just Calypsoifies the Plugins screen.
Screen Shot 2019-04-10 at 10 39 06 AM

### How to test the changes in this Pull Request:

1. Go to Plugins screen. Install and manage recommended plugins on the screen.
2. Activate Jetpack. On the plugins screen, add `?calypsoify=1` to the URL. Install and manage recommended plugins on the Calypsoified screen.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #11, opened 2019-03-28
Added setup instructions for the development environment

iuravic:
Added  DEV_ENV.md containing setup instructions for the development environment.
This is the first version of the development environment, and it will be updated soon.

### All Submissions:

* [x ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes # .

### How to test the changes in this Pull Request:

1. By reading the given setup instructions in the committed file.
2.
3.

### Other information:

* [x ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x ] Have you written new tests for your changes, as applicable?
* [ x] Have you successfully ran tests with your changes locally?

Issue #10, opened 2019-03-27
Add managed plugins to Plugins page


claudiulodro:
Add the managed plugins to the WP Admin Plugins screen. Give them some visual differentiation to make it clear they are the Newspack recommended plugins, and put them at the top of the list. Hook them up to the plugin management features and enable them to be installed from this screen if not already installed.

Here is a rough example (with a gray bg). Pick a nicer color than this, though (Newspack color if it makes sense):

<img width="907" alt="Screen Shot 2019-03-26 at 12 27 20 PM" src="https://user-images.githubusercontent.com/7317227/55091679-8638ba00-506e-11e9-8a21-92252a122bd0.png">

PR #3169, opened 2024-06-05
feat(ia): Settings / Connections - Webhooks

jaredrethman:
### All Submissions:

* Adds Webhooks section/component.
* Replaced conditional (`isEditing`, `isDeleting` etc) action props with a single prop `action` to delegate Modal component state.
* Added empty endpoints notification.
* Modularize: Abstracted functionality and components into separate files.
* Added Loading status to Button during API requests.

**Original location:**
* Connections Wizard: _/wp-admin/admin.php?page=newspack-connections-wizard_

1. Checkout `feat/ia-settings-connections__webhooks` and build assets.
3. Depending on whether you have endpoints stored in your environment, confirm you see one of the below:

**Some Endpoints:**
Screenshot 2024-06-05 at 22 10 32

**No Endpoints:**
Screenshot 2024-06-05 at 22 06 01

4. Perform CRUD on endpoints (i.e. Add, edit and delete) and confirm all changes are reflected.

#### CRUD examples:
1. Click edit on an Endpoint and update Authentication Token, Actions, Label.
2. Click Delete, + confirm delete when prompted, an Endpoint.
3. Click "Add New Endpoint".
   * Add "https://66612b4263e6a0189fe8b689.mockapi.io/np/v1/test" as URL.
   * Click "Send a test request"
   * Add "Test Label" as Label.
   * Click "Save".

**Additional steps:**

### Other information:


assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
```diff
@@ -0,0 +1,69 @@
...
+			className="newspack-webhooks__endpoint mt16"
+			toggleChecked={ ! endpoint.disabled }
+			toggleOnChange={ () => setAction( 'toggle', endpoint.id ) }
+			key={ endpoint.id }
+			title={ getEndpointTitle( endpoint ) }
+			disabled={ endpoint.system }
+			description={ () => {
+				if ( endpoint.disabled && endpoint.disabled_error ) {
+					return `${ __(
+						'This endpoint is disabled due to excessive request errors',
```
chickenn00dle: non-blocking question: Are there other reasons the endpoint might be disabled? If so, let's make this message more generic and not specify the reason.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
```diff
@@ -0,0 +1,69 @@
...
+} ) {
+	return (
+		<WizardsActionCard
+			isMedium
+			className="newspack-webhooks__endpoint mt16"
+			toggleChecked={ ! endpoint.disabled }
+			toggleOnChange={ () => setAction( 'toggle', endpoint.id ) }
+			key={ endpoint.id }
+			title={ getEndpointTitle( endpoint ) }
+			disabled={ endpoint.system }
```
chickenn00dle: I think this needs to be `endpoint.disabled`. Endpoints aren't able to be toggled off right now.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
```diff
@@ -0,0 +1,424 @@
...
+const CheckboxControl: React.FC< WpCheckboxControlPropsOverride< typeof WpCheckboxControl > > = ( {
+	...props
+} ) => {
+	return <WpCheckboxControl { ...props } />;
+};
+
+/**
+ * Endpoints main GET cache key.
+ */
+const NAMESPACE = '/newspack-settings/connections/webhooks';
```
chickenn00dle: non-blocking nitpick: Might be worth moving these constants into a separate constants file so they are easier to track down in the future.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
```diff
@@ -0,0 +1,424 @@
...
+							noticeText={ __(
+								"This endpoint hasn't received any requests yet.",
+								'newspack-plugin'
+							) }
+						/>
+					) }
+				</Modal>
+			) }
+			{ [ 'edit', 'new' ].includes( action ?? '' ) && (
+				<Modal
```
chickenn00dle: I think it's worth separating out these different modals into their own components especially considering this file is over 400 lines long. WDYT?

chickenn00dle (changes requested):
Nice work @jaredrethman!

This one is mostly ready to go. Left a few comments to address first though.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
jaredrethman: You are correct. I have removed the `disable` prop entirely since disabling won't let you toggle at all once disabled. The issue with toggling was unrelated to this, I have remedied the issue.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
jaredrethman: Done 👍

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
jaredrethman: It might not be noticeable, but webhooks has already been broken into 7 files from the initial [1 file](https://github.com/Automattic/newspack-plugin/blob/edabd0cc74b23773e88215fed1841ae835cb1750/assets/wizards/connections/views/main/webhooks.js).  Happy to break out further, let me know :)

chickenn00dle: Personally, I don't see a problem with breaking this down even more so long as the result is intuitive and easier to parse. IMO, moving these two modals into their own components will help make this file cleaner, as well as make updating the modals simpler in the future.

That said, I'm not too adamant about this, so happy to leave things as is as well.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
chickenn00dle: Thanks! This is working now 👍

chickenn00dle (changes requested):
Nice! This is looking good now for the most part.

While testing the second time around, I ran into an edge case where an unhandled wp error renders on the front end though:

Screenshot 2024-06-20 at 12 24 01

This happens when I attempt to add a webhook URL that already exists. Here is the full error:

```
[20-Jun-2024 16:12:03 UTC] PHP Fatal error:  Uncaught Error: Cannot use object of type WP_Error as array in /Users/raz/Sites/newspack/plugin/includes/data-events/class-api.php:283
Stack trace:
#0 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api/class-wp-rest-server.php(1230): Newspack\Data_Events\Api::upsert_endpoint(Object(WP_REST_Request))
#1 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api/class-wp-rest-server.php(1063): WP_REST_Server->respond_to_request(Object(WP_REST_Request), '/newspack/v1/we...', Array, NULL)
#2 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api/class-wp-rest-server.php(439): WP_REST_Server->dispatch(Object(WP_REST_Request))
#3 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api.php(428): WP_REST_Server->serve_request('/newspack/v1/we...')
#4 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp-hook.php(324): rest_api_loaded(Object(WP))
#5 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)
#6 /Users/raz/Sites/newspack/sites/site/wp-includes/plugin.php(565): WP_Hook->do_action(Array)
#7 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp.php(418): do_action_ref_array('parse_request', Array)
#8 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp.php(813): WP->parse_request('')
#9 /Users/raz/Sites/newspack/sites/site/wp-includes/functions.php(1336): WP->main('')
#10 /Users/raz/Sites/newspack/sites/site/wp-blog-header.php(16): wp()
#11 /Users/raz/Sites/newspack/sites/site/index.php(17): require('/Users/raz/Site...')
#12 /Users/raz/.composer/vendor/laravel/valet/server.php(110): require('/Users/raz/Site...')
#13 {main}
  thrown in /Users/raz/Sites/newspack/plugin/includes/data-events/class-api.php on line 283
```

I think we should account for this, and then this will be good to go.

jaredrethman:
> Nice! This is looking good now for the most part.
> 
> While testing the second time around, I ran into an edge case where an unhandled wp error renders on the front end though:
> 
> Screenshot 2024-06-20 at 12 24 01
> 
> This happens when I attempt to add a webhook URL that already exists. Here is the full error:
> 
> ```
> [20-Jun-2024 16:12:03 UTC] PHP Fatal error:  Uncaught Error: Cannot use object of type WP_Error as array in /Users/raz/Sites/newspack/plugin/includes/data-events/class-api.php:283
> Stack trace:
> #0 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api/class-wp-rest-server.php(1230): Newspack\Data_Events\Api::upsert_endpoint(Object(WP_REST_Request))
> #1 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api/class-wp-rest-server.php(1063): WP_REST_Server->respond_to_request(Object(WP_REST_Request), '/newspack/v1/we...', Array, NULL)
> #2 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api/class-wp-rest-server.php(439): WP_REST_Server->dispatch(Object(WP_REST_Request))
> #3 /Users/raz/Sites/newspack/sites/site/wp-includes/rest-api.php(428): WP_REST_Server->serve_request('/newspack/v1/we...')
> #4 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp-hook.php(324): rest_api_loaded(Object(WP))
> #5 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)
> #6 /Users/raz/Sites/newspack/sites/site/wp-includes/plugin.php(565): WP_Hook->do_action(Array)
> #7 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp.php(418): do_action_ref_array('parse_request', Array)
> #8 /Users/raz/Sites/newspack/sites/site/wp-includes/class-wp.php(813): WP->parse_request('')
> #9 /Users/raz/Sites/newspack/sites/site/wp-includes/functions.php(1336): WP->main('')
> #10 /Users/raz/Sites/newspack/sites/site/wp-blog-header.php(16): wp()
> #11 /Users/raz/Sites/newspack/sites/site/index.php(17): require('/Users/raz/Site...')
> #12 /Users/raz/.composer/vendor/laravel/valet/server.php(110): require('/Users/raz/Site...')
> #13 {main}
>   thrown in /Users/raz/Sites/newspack/plugin/includes/data-events/class-api.php on line 283
> ```
> 
> I think we should account for this, and then this will be good to go.

Nice catch! I had accounted for this in the [original branch](https://github.com/Automattic/newspack-plugin/blob/4b047e8aaef4eab1f531bbf023dc71da1e76cd20/includes/data-events/class-api.php#L283-L296) but missed it when creating this PR. This should be fixed in cc5f4ed586579c4fb913a886443fd6e751f03c48

Screenshot 2024-06-20 at 13 44 04

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
jaredrethman: I'll granularize it further, might result in some ugly prop-drilling (part in due to use of redux vs context).

Stay tuned, will provide update later today.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
jaredrethman: Unable to find anything that pertains specifically to Wizards. However, I found that there is some logic in Webhooks internally: https://github.com/Automattic/newspack-plugin/blob/5f34145d37437e41fc260884df7eac9ca90e42cd/includes/data-events/class-webhooks.php#L360-L369 which attempts to recover from errors (x 15) by gradually increasing the delay between attempts seen here - https://github.com/Automattic/newspack-plugin/blob/5f34145d37437e41fc260884df7eac9ca90e42cd/includes/data-events/class-webhooks.php#L752-L768 based on this I've decided to make the error message more generic, as you've recommended.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
jaredrethman: @chickenn00dle this has been done 2dcee95f11a3e60304e2015851e1433af298555f, let me know if you feel further refinement is necessary?

chickenn00dle: Nice! This LGTM 👍 

> I'll granularize it further, might result in some ugly prop-drilling (part in due to use of redux vs context).

I don't think it's necessary in this case, but as an FYI, you can use redux's [useSelector](https://react-redux.js.org/api/hooks#useselector) (which uses context API under the hood I believe) to avoid prop-drilling.

chickenn00dle (changes requested):
Thanks for moving the modals into separate components @jaredrethman!

That bit LGTM. I am noticing testing the endpoint seems to result in a `400` bad request now though. When I `curl` the endpoint I get a `200` though.

Screenshot 2024-06-24 at 13 00 08

Also, as an aside, maybe we should render errors from the test in the same place we are rendering errors at the top of the modal? We can tackle this in a separate PR though if you prefer.

jaredrethman:
>  I am noticing testing the endpoint seems to result in a 400 bad request now though. When I curl the endpoint I get a 200 though.

That endpoint has reached it's limit. You can test using https://httpbin.org

> Also, as an aside, maybe we should render errors from the test in the same place we are rendering errors at the top of the modal? We can tackle this in a separate PR though if you prefer.

I decided to tackle this. Two things to note:
1. Have moved both success and error messages to top.
<details>
<summary>Succes - OK 200</summary>

Screenshot 2024-06-25 at 22 30 17
</details>
<details>
<summary>Error - 404: Not Found</summary>

Screenshot 2024-06-25 at 22 32 26
</details>

Tested several status codes from https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

Made several other changes/refactors.

assets/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-modals.tsx
jaredrethman: > I don't think it's necessary in this case, but as an FYI, you can use redux's [useSelector](https://react-redux.js.org/api/hooks#useselector) (which uses context API under the hood I believe) to avoid prop-drilling.

Thanks for the feedback. I should have been more specific, we're using [WordPress's version of redux](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#createreduxstore). The equivalent is [useSelect](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#useselect), which we are using. 

Also, the need for prop-drilling has nothing to do with Redux (I must have bumped my head when I said that, apologies). What requires the prop drilling on the modal components is the need to use the same instance of `useWizardApiFetch` methods & state throughout the component and it's children.

chickenn00dle: Thanks for clarifying @jaredrethman

> I should have been more specific, we're using [WordPress's version of redux](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#createreduxstore). The equivalent is [useSelect](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#useselect), which we are using.

Ah. That's right.

> the need for prop-drilling has nothing to do with Redux (I must have bumped my head when I said that, apologies). What requires the prop drilling on the modal components is the need to use the same instance of useWizardApiFetch methods & state throughout the component and it's children.

I haven't worked too deeply with `@wordpress/data` for a few years, but IIRC there should be feature parity between it and redux (since I thought it was built on redux?). So it should still be possible to create a new store, etc., and populate a components children with some state without the use of prop-drilling. Again, I don't think that's necessary here though. Was just bringing this up since I saw a mention of "ugly prop-drilling"

chickenn00dle (approved):
Nice! Thanks for addressing all of my feedback @jaredrethman.

This looks good now so giving this a thumbs up. I did notice a very small issue where the top margin for the test request success message appears to be missing its top margin:

Screenshot 2024-06-27 at 11 22 20

For reference, here is the error message top margin:

Screenshot 2024-06-27 at 11 22 41

But I'll leave it to you to sneak this fix in here or address in another PR. 

In any case, this is good to go 👍

jaredrethman:
> But I'll leave it to you to sneak this fix in here or address in another PR.

Snuck it in, removed `mt0` `className` from all `<Notice />` components.
Screenshot 2024-06-27 at 22 17 23

PR #3175, opened 2024-06-08
feat(ia): Settings / Connections - Custom Events & server-side Section class

jaredrethman:
### All Submissions:

* Adds Custom Events section/component.
* Updated `autocomplete` attributes from `off` to `one-time-code` to avoid browser prompts to save password.
* Moved all Newspack related wizardry to a directory `./newspack`.
* Added new Section Class to provide server-side functionality on a section-by-section basis. This, for now, is mostly necessary when you need to add REST api endpoint(s) to a specific section. 
* Added `includes/wizards/class-wizard.php` include to `includes/class-newspack.php` `init` method to avoid including it on every `Wizard` inherited class.
* Added Measurement ID validation to Frontend (avoids invalid requests from hitting backend) and refactored server-side validation. Rules are must start with `"G-"` followed by 10 (we use `{10,}` to allow for future lengthening) alphanumeric characters. 
* Added more descriptive, and granular, error feedback.
* Replaced deprecated conditional props on  `<Button />` with preferred `variant` prop.

**Original location:**
* Newspack Custom Events: _/wp-admin/admin.php?page=newspack-analytics-wizard#/newspack-custom-events_

1. Checkout `feat/ia-settings-connections__custom-events` and build assets.
3. Scroll to the bottom where "Activate Newspack Custom Events" section is located.
4. Observe that saving/updating data works as expected.

Screenshot 2024-06-06 at 14 10 14

5. Removing ID displays a notification and disables saving.

Screenshot 2024-06-08 at 09 31 30

**Additional steps:**
- Confirm stored data is persistent between tab changes.

### Other information:


assets/wizards/newspack/views/settings/connections/custom-events.tsx
```diff
@@ -0,0 +1,143 @@
...
+function CustomEvents() {
+	const [ ga4Credentials, setGa4Credentials ] = useState< Ga4Credentials >(
+		window.newspackSettings.connections.sections.customEvents
+	);
+	const { wizardApiFetch, errorMessage, resetError, setError } = useWizardApiFetch(
+		'newspack-settings/connections/custom-events'
+	);
+
+	useEffect( () => {
+		if ( ! isValidGA4MeasurementID( ga4Credentials.measurement_id ) ) {
```
chickenn00dle, line 51: This results in an error always being shown even when a publisher does not want to enable custom events. This feels a bit strange. At the very least we should not show the error message when both custom event fields are empty. But IMO an error message should only appear in response to some invalid action taken by the user (i.e. perform some validation before saving and outputting the error at this stage).

assets/wizards/newspack/views/settings/connections/custom-events.tsx
```diff
@@ -0,0 +1,143 @@
...
+					__(
+						'You need a valid Measurement ID (e.g. "G-ABCDE12345") to activate Newspack Custom Events.',
+						'newspack-plugin'
+					),
+					'ga4_invalid_measurement_id'
+				)
+			);
+			return;
+		}
+		if ( ! ga4Credentials.measurement_protocol_secret ) {
```
chickenn00dle, line 63: See previous comment

assets/wizards/newspack/views/settings/connections/custom-events.tsx
```diff
@@ -0,0 +1,143 @@
...
+			<div className="newspack__analytics-configuration__header">
+				<p>
+					{ __(
+						"Newspack already sends some custom event data to your GA account, but adding the credentials below enables enhanced events that are fired from your site's backend. For example, when a donation is confirmed or when a user successfully subscribes to a newsletter.",
+						'newspack-plugin'
+					) }
+				</p>
+			</div>
+
+			{ errorMessage && <Notice isError noticeText={ errorMessage } /> }
```
chickenn00dle: Nitpick: Rendering the error message above the form fields results in the form moving around as the measurement ID field is being modified. I would suggest placing the error below the form inputs instead to avoid this behavior

assets/wizards/newspack/views/settings/connections/custom-events.tsx
```diff
@@ -0,0 +1,143 @@
...
+						setGa4Credentials( { ...ga4Credentials, measurement_protocol_secret: value } )
+					}
+					autoComplete="one-time-code"
+				/>
+			</Grid>
+			<Button
+				className="newspack__analytics-newspack-custom-events__save-button"
+				variant="primary"
+				onClick={ updateGa4Credentials }
+				disabled={ !! errorMessage }
```
chickenn00dle: There is some circular logic here when the fields start off empty (when a publisher wants to enter credentials for the first time). The error message will appear by default, and so the save button is disabled. However the error message hook will only reset the error message when the credentials state is updated, which can only be done by clicking the save button.

includes/wizards/class-wizard.php
```diff
@@ -45,12 +47,25 @@ abstract class Wizard {
...
+	 * Array to store instances of section objects.
+	 *
+	 * @var Wizards\Section[]
+	 */
+	protected $sections = [];
+
 	/**
 	 * Initialize.
+	 *
+	 * @param array $args Array of optional arguments. i.e. `sections`.
```
chickenn00dle, line 60: Nitpick: It might be worth adding some additional details for the sections entry of `$args` here to make it more clear what this should look like without needing to sift through the code in the future.

includes/wizards/class-wizard.php
```diff
@@ -247,4 +262,15 @@ public function get_style_dependencies( $dependencies = [] ) {
...
 	abstract public function get_name();
+
+	/**
+	 * Load wizard sections.
+	 * 
+	 * @param Section[] $sections Array of Section objects.
+	 */
+	public function load_wizard_sections( $sections ) {
+		foreach ( $sections as $section_slug => $section_class ) {
+			$this->sections[ $section_slug ] = new $section_class();
```
chickenn00dle, line 279: Non-blocking question: Is it worth doing any validation here before instantiating?

includes/wizards/newspack/class-custom-events-section.php
```diff
@@ -0,0 +1,99 @@
...
+				'newspack_analytics_wizard_invalid_params',
+				esc_html__( 'Invalid Measurement Protocol API Secret.', 'newspack' ),
+				[ 'status' => 400 ]
+			);
+		}
+
+		update_option( 'ga4_measurement_id', sanitize_text_field( $measurement_id ) );
+		update_option( 'ga4_measurement_protocol_secret', sanitize_text_field( $measurement_protocol_secret ) );
+
+		return rest_ensure_response( $this->get_data() );
```
chickenn00dle: I'm not familiar with how option caching works, but is it necessary to trigger two calls to `get_option` via `get_data` when we are already have the values to return in the response stored in variables above?

includes/wizards/newspack/class-custom-events-section.php
```diff
@@ -0,0 +1,99 @@
...
+		if ( ! $measurement_protocol_secret ) {
+			return new WP_Error(
+				'newspack_analytics_wizard_invalid_params',
+				esc_html__( 'Invalid Measurement Protocol API Secret.', 'newspack' ),
+				[ 'status' => 400 ]
+			);
+		}
+
+		update_option( 'ga4_measurement_id', sanitize_text_field( $measurement_id ) );
+		update_option( 'ga4_measurement_protocol_secret', sanitize_text_field( $measurement_protocol_secret ) );
```
chickenn00dle, line 81: I think these are being sanitized twice given the `sanitize_callback` specified in while registering the route.

chickenn00dle (changes requested):
Thanks for this Jared!

The core functionality works as expected, but I left some comments for some issues I noticed.

assets/wizards/newspack/views/settings/connections/custom-events.tsx
chickenn00dle: Related to this, there is currently no way to remove credentials once they are entered. We should either allow saving empty fields, or add a reset button (preferably a reset button)

assets/wizards/newspack/views/settings/connections/custom-events.tsx
jaredrethman: Thanks for the feedback @chickenn00dle, I have added an early exit to avoid errors displaying when both inputs are empty.

includes/wizards/class-wizard.php
jaredrethman: I have added an `@example`, let me know if this is _not_ what you had in mind.

includes/wizards/class-wizard.php
jaredrethman: IMO, errors are important to drive engineering. I have made it more presentable.
Screenshot 2024-06-19 at 23 07 17

includes/wizards/newspack/class-custom-events-section.php
jaredrethman: I agree, however, I am following [late escaping](https://docs.wpvip.com/security/validating-sanitizing-and-escaping/#h-always-escape-late) principles. FWIW - for the most part this code was [cloned](https://github.com/Automattic/newspack-plugin/blob/129db37d81d3f4b4eff10803e89aaed4c2804b08/includes/wizards/class-analytics-wizard.php#L106-L122). 

Let me know if you'd still like me to remove?

jaredrethman (commented):
@chickenn00dle thanks for the review and feedback.

I have addressed most of your concerns and provided my own feedback where necessary.

includes/wizards/newspack/class-custom-events-section.php
jaredrethman: While this is not strictly related to cache - options are auto-loaded (Unless specified in the 3rd arg in `*_option` functions) which means they're always in memory. I originally left the `$this->get_data()` call for these reasons + it also ensures the integrity of the data and only returning what is stored in the DB. 

Have removed and reused existing variables within the method.

assets/wizards/newspack/views/settings/connections/custom-events.tsx
jaredrethman: I like the `reset` suggestion. Have added:
* Registered endpoint
* Endpoint callback
* UI to facilitate reseting inputs

assets/wizards/newspack/views/settings/connections/custom-events.tsx
jaredrethman: This was from existing UI. I have moved the error notice to below the inputs as you've suggested.

includes/wizards/class-wizard.php
chickenn00dle: I was thinking more along the lines of [explicitly documenting](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#docblock-formatting) (see Parameters That Are Arrays) `sections`, but this works as well!

assets/wizards/newspack/views/settings/connections/custom-events.tsx
chickenn00dle: Great! This is much better 👍

includes/wizards/newspack/class-custom-events-section.php
chickenn00dle: Thanks for clarifying. I'm fine with the double sanitizing as it's not breaking anything 👍

assets/wizards/newspack/views/settings/connections/custom-events.tsx
chickenn00dle: Thanks! Working great now 👍

chickenn00dle (approved):
Great! Thanks for addressing all of my feedback. Looks good now 👍

PR #3179, opened 2024-06-11
feat(ras-acc): add modal checkout GA4 events

laurelfulford:
### All Submissions:

Along with https://github.com/Automattic/newspack-blocks/pull/1768, adds GA4 tracking events to the modal checkout.

This is my first time working with our GA4 Data Events, so the exact information being tracked when and why could use some special attention. I'm adding @kmwilkerson as a tester just to make sure it gets her eyes, too!

The following should be tracked:
* When a reader triggers a new modal checkout, including:
     * What block type triggered the checkout.
     * What product was purchased.
     * How much money was spent.
* When a reader fills out the information on the first screen and clicks "Continue".
* When a reader clicks the button to complete the transaction -- for this, we're just tracking intent, and not whether or not the transaction was successful, since that's already captured. 

See 1205234045751551-as-1206525965459152

1. Ideally start on a test site connected to GA4.
2. Apply this PR and https://github.com/Automattic/newspack-blocks/pull/1768, and run `npm run build`.
3. Optionally log events by adding `error_log( print_r( $event, true ) );` after [this line](https://github.com/Automattic/newspack-plugin/blob/9b372e9e866c9584c62e9aec04c250ce3396486d/includes/data-events/connectors/ga4/class-ga4.php#L154).
4. Set up a couple checkout situations -- ideally a Donate Block, and at least one Checkout Button Block.
5. Run through the checkout and confirm what's being tracked -- the following params should be tracked for the `np_modal_checkout_interaction` event, for each step. Some are already baked in to the Data Events, so I bolded the ones added for this event, and I added notes where I wasn't 100% sure what was being captured was correct:

**When the modal checkout is opened from a Donation or Checkout Button block:**

* logged_in
* ga_session_id
* is_reader
* email_hash
* is_newsletter_subscriber
* is_donor
* is_subscriber
* **action** (is set to always be `form_submission_received`, since the modal is triggered by a successful form submission). 
* **action_type** (is `donation` for any donation purchases, and `paid_membership` for any other purchase type)
* **referer**
* **amount**
* **currency**
* **product_id**
* **recurrence**

**When the Continue Button is clicked:**

* logged_in
* ga_session_id
* is_reader
* email_hash
* is_newsletter_subscriber
* is_donor
* is_subscriber
* **action** (is always just `form_submission_received`)
* **modal_pagination** (will return `2` when you click Continue or in a weird case where, if you finish up page 1, refresh/navigate away from the page, and restart the checkout since it drops you on page 2).
* **amount**
* **currency**
* **product_id**
* **recurrence**

**Click button to complete purchase:**
* logged_in
* ga_session_id
* author
* is_reader
* email_hash
* is_newsletter_subscriber
* is_donor
* is_subscriber
* **action** (always `form_submission_received` since we're not tracking success/failure here)
* **order_id**
* **amount**
* **currency**
* **product_id**
* **recurrence**

### Other information:


includes/data-events/class-modal-checkout.php
```diff
@@ -0,0 +1,166 @@
...
+	public static function checkout_attempt( $order ) {
+		$order_data = \Newspack\Data_Events\Utils::get_order_data( $order->get_id(), true );
+
+		$data = [
+			'action'     => self::FORM_SUBMISSION,
+			'order_id'   => $order->get_id(),
+			'amount'     => $order->get_total(),
+			'currency'   => $order->get_currency(),
+			'recurrence' => $order_data['recurrence'],
+			'product_id' => $order_data['platform_data']['product_id'],
```
chickenn00dle, line 165: Looks like these lines are triggering the following PHP warning sometimes:

```
[10-Jul-2024 19:46:05 UTC] PHP Warning:  Trying to access array offset on value of type null in /Users/raz/Sites/newspack/plugin/includes/data-events/class-modal-checkout.php on line 160
[10-Jul-2024 19:46:05 UTC] PHP Warning:  Trying to access array offset on value of type null in /Users/raz/Sites/newspack/plugin/includes/data-events/class-modal-checkout.php on line 161
```

Since `\Newspack\Data_Events\Utils::get_order_data` can return `null`, we should validate `$order_data` before setting these values.

includes/data-events/class-modal-checkout.php
```diff
@@ -0,0 +1,166 @@
...
+	 */
+	public static function checkout_button_purchase( $checkout_button_metadata ) {
+		$metadata = [];
+		foreach ( $checkout_button_metadata as $key => $value ) {
+			$metadata[ $key ] = $value;
+		}
+
+		$data = [
+			'action'      => self::FORM_SUBMISSION_SUCCESS,
+			'action_type' => 'paid_membership',
```
chickenn00dle, line 87: This default is a bit misleading for simple product purchases. Or is the intention to ignore non-membership, non-donation purchases altogether (which is not the current behavior).

Also, I have a clarifying question. Are regular subscription products considered `paid_membership`'s as well? Or does this refer to subscriptions that are tied to WC Membership plans only?

includes/data-events/class-modal-checkout.php
```diff
@@ -0,0 +1,166 @@
...
+
+	/**
+	 * The name of the action for form submissions
+	 */
+	const FORM_SUBMISSION = 'form_submission_received';
+
+	/**
+	 * The name of the action for form submissions
+	 */
+	const FORM_SUBMISSION_SUCCESS = 'form_submission_success';
```
chickenn00dle, line 26: Nitpick: The naming for this is a bit misleading. At first glance it seems to indicate checkout was successful, but it really indicates the modal checkout form has been opened (I think based on this comment below: `* Fires when a reader opens the modal checkout from a checkout button block.`).

I think we can stand to make these action names more clear. For example:

`MODAL_CHECKOUT_OPEN`
`MODAL_CHECKOUT_CONTINUE`
`MODAL_CHECKOUT_SUCCESS`

includes/data-events/class-modal-checkout.php
```diff
@@ -0,0 +1,166 @@
...
+			[ __CLASS__, 'modal_pagination' ]
+		);
+	}
+
+	/**
+	 * Returns whether a product is a one time purchase, or recurring and when.
+	 *
+	 * @param string $product_id Product's ID.
+	 */
+	public static function get_purchase_recurrence( $product_id ) {
```
chickenn00dle, line 64: Nonblocking nitpick: I believe we refer to this as `frequency` pretty consistently within the blocks codebase. Maybe we can maintain this terminology?

includes/data-events/class-modal-checkout.php
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	/**
+	 * Fires when a reader opens the modal checkout from a donate block.
+	 *
+	 * @param array $donation_metadata Information about the donation.
+	 *
+	 * @return ?array
+	 */
+	public static function donate_button_purchase( $donation_metadata ) {
```
chickenn00dle, line 103: Just noting that I haven't been able to test donations yet as there seems to be a problem with them in the blocks PR for this.

chickenn00dle (changes requested):
Nice work on this one as well @laurelfulford!

Again, checkout button seems to be working as expected. I did leave some comments for a few questions and issues I ran into while testing though.

includes/data-events/class-modal-checkout.php
laurelfulford: Yeah I'm unsure about this bit, too -- I'll talk to @kmwilkerson to clarify what the options should be here (and do more digging myself!).

includes/data-events/class-modal-checkout.php
laurelfulford: I opted to leave this one -- it's a good point, but both GA4 and a lot of our other code for GA4 in Newspack Plugin use "recurrence", so I opted to leave it GA4-centric 🙂

includes/data-events/class-modal-checkout.php
laurelfulford: I'll double check this one with @kmwilkerson! I agree it's a bit confusing given what's actually being tracked, but it's consistent with how stuff is being tracked (like as a successful form submission) and how it's captured otherwise in the plugin -- I'm not sure how closely we want to follow that though!

includes/data-events/class-modal-checkout.php
laurelfulford: Good catch, @chickenn00dle! This should be fixed in [17c683d](https://github.com/Automattic/newspack-plugin/pull/3179/commits/17c683dfb2756e44f1fe276195a4ca24a285d4b6).

includes/data-events/class-modal-checkout.php
laurelfulford: This should now be fixed!

laurelfulford:
Thanks for the review @chickenn00dle! This should (largely) be ready for review again, but I'm going to confirm some of the GA4 data stuff -- if you'd like to leave it 'til then, I can update here when that bit has been confirmed 🙂

laurelfulford:
Leaving this one open for now, but I've started reworking this PR [here](https://github.com/Automattic/newspack-plugin/pull/3297) to have a fresh start and add front-end events.

laurelfulford:
Closing in favour of the reworked draft: https://github.com/Automattic/newspack-plugin/pull/3297

PR #3181, opened 2024-06-13
 fix: update dependencies to avoid conflict in WP 6.6

dkoo:
### All Submissions:

Refactoring required for changes in https://github.com/Automattic/newspack-scripts/pull/209. Note that this PR will not pass CI tests and should NOT be merged until https://github.com/Automattic/newspack-scripts/pull/209 is released as an NPM package, so this repo can install it as a dependency.

**Note:** this PR also renames the `assets` folder to `src` to bring this repo in line with all other Newspack repos' directory structure.

Follow instructions in https://github.com/Automattic/newspack-scripts/pull/209.

### Other information:


package.json
```diff
@@ -42,33 +33,29 @@
 		"watch": "newspack-scripts watch",
 		"test": "newspack-scripts test",
 		"lint": "npm run lint:scss && npm run lint:js",
-		"lint:js": "eslint --ext .js,.jsx,.ts,.tsx assets",
-		"lint:js:staged": "eslint --ext .js,.jsx,.ts,.tsx",
+		"lint:js": "newspack-scripts proxy eslint 'src/**/*.{js,jsx,ts,tsx}'",
```
adekbadek: This script exits with success state regardless of any JS issues. It should report linting issues, but it appears like it's not doing anything.

src/components/src/image-upload/index.test.js
```diff
@@ -1,8 +1,23 @@
...
 
+/**
+ * Mocks
+ */
+jest.mock( 'react-router-dom', () => ( {
+	...jest.requireActual( 'react-router-dom' ),
+	useHistory: () => ( {
+		push: jest.fn(),
+	} ),
+} ) );
```
adekbadek: When I removed all the `jest.mock` calls (three in this PR), tests still passed. Are these necessary?

package.json
```diff
@@ -42,33 +33,29 @@
...
+		"lint:js:staged": "newspack-scripts proxy eslint --ext .js,.jsx,.ts,.tsx",
+		"fix:js": "newspack-scripts proxy eslint 'src/**/*.{js,jsx,ts,tsx}' --fix",
 		"lint:php:staged": "./vendor/bin/phpcs --filter=GitStaged",
-		"format:js": "prettier 'assets/**/*.{js,jsx,ts,tsx}' --write",
-		"typescript:check": "newspack-scripts typescript-check",
-		"lint:scss": "stylelint '**/*.scss' --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
-		"format:scss": "prettier --write '**/*.scss'",
-		"lint:scss:staged": "stylelint --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
+		"typescript:check": "newspack-scripts proxy tsc",
+		"lint:scss": "newspack-scripts proxy stylelint '**/*.scss' --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
```
adekbadek: As with the JS lint, running this script does not seem to have any effect.

package.json
```diff
@@ -42,33 +33,29 @@
...
+		"lint:js": "newspack-scripts proxy eslint 'src/**/*.{js,jsx,ts,tsx}'",
+		"lint:js:staged": "newspack-scripts proxy eslint --ext .js,.jsx,.ts,.tsx",
+		"fix:js": "newspack-scripts proxy eslint 'src/**/*.{js,jsx,ts,tsx}' --fix",
 		"lint:php:staged": "./vendor/bin/phpcs --filter=GitStaged",
-		"format:js": "prettier 'assets/**/*.{js,jsx,ts,tsx}' --write",
-		"typescript:check": "newspack-scripts typescript-check",
-		"lint:scss": "stylelint '**/*.scss' --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
-		"format:scss": "prettier --write '**/*.scss'",
-		"lint:scss:staged": "stylelint --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
+		"typescript:check": "newspack-scripts proxy tsc",
```
adekbadek: As with the JS lint, running this script does not seem to have any effect.

package.json
```diff
@@ -6,25 +6,16 @@
...
-		"@types/wordpress__blocks": "^12.5.14",
-		"@types/wordpress__components": "^23.0.12",
-		"@wordpress/browserslist-config": "^5.40.0",
-		"eslint": "^7.32.0",
+		"@wordpress/browserslist-config": "^6.0.0",
+		"eslint": "^8.57.0",
 		"lint-staged": "^15.2.2",
-		"newspack-scripts": "^5.3.0",
-		"postcss-scss": "^4.0.9",
-		"prettier": "npm:wp-prettier@^2.6.2-beta-1",
```
adekbadek, line 26: Removing this dependency has the effect of not applying custom (WP) prettier rules to eslint. Running `eslint` directly (`./node_modules/.bin/eslint src`) would raise formatting errors because of this.

adekbadek (dismissed):
The only issue I've found is that the linting and typescript checking is broken with these changes. Are these `proxy` scripts necessary?

package.json
dkoo: This should be fixed with `v5.6.0-alpha.*` versions of newspack-scripts.

src/components/src/image-upload/index.test.js
dkoo: Not sure! If you force a condition that would cause the test(s) to fail, does it still report the failure?

package.json
dkoo: This should be fixed with `v5.6.0-alpha.*` versions of newspack-scripts.

package.json
dkoo: This should be fixed with `v5.6.0-alpha.*` versions of newspack-scripts.

package.json
dkoo: I'm not sure how Prettier and `wp-scripts` should be interacting. I found that when I kept the Prettier dependency, it was reporting very different style/formatting rules than `wp-scripts`. When using `wp-scripts lint-js` it seems to have built-in formatting rules (I think it already invokes Prettier under the hood), so I'm not sure a separate Prettier dependency or config is still necessary?

dkoo: With `v5.6.0-alpha.*` versions of newspack-scripts, `lint:js` and `format:js` scripts seem to be working with WP formatting standards, and it's also formatting correctly in my VSCode.

dkoo:
All linting, testing, and Typescript check scripts should now be working. This should be ready for production pending QA.

package.json
adekbadek: In my testing, running `npm run format:js` reformats JS, introducing changes to most of the files. Running `prettier` directly (`./node_modules/.bin/prettier src --write`) has the same effect (though it will also format SCSS).

src/components/src/image-upload/index.test.js
adekbadek: Yes

package.json
dkoo: Yes, `@wordpress/scripts` enforces different formatting rules than what we've been using as our in-house style (which is closer to ESLint's formatting, e.g. with spaces in brackets like `( foo, bar )` and `[ a, b ]`. So this is technically working, but I'm not sure we want to change our team's coding style to conform to this just yet, which is why I [removed Prettier as a requirement to pass the `lint` command](https://github.com/Automattic/newspack-scripts/pull/209/files#diff-1bd145b27e5bc507fcffdc88b445a1b341284a8f581d53375452838316832a41R75) in our own ESLint config.

What's nice about `@wordpress/scripts` is that linting and formatting can be treated separately. So if `npm run lint:js` throws autofixable errors, we should be able to run `wp-scripts lint-js --fix` to autofix those without enforcing Prettier formatting styles. Would it be helpful in the short term to replace the `format:js` commands with a `fix:js` command that runs this, and then reintroduce the Prettier-based `format:js` script in the future if we decide to follow the Prettier rules from `@wordpress/scripts`?

dkoo: c580196 temporarily removes the `format:js` script and introduces a `fix:js` script which basically runs `eslint --fix` via @wordpress/scripts. This should cover autofixable code quality errors without addressing the conflicting Prettier formatting rules at this time.

dkoo: Oh! In the flurry I missed your initial mention of `"prettier": "npm:wp-prettier@^2.6.2-beta-1",`. Restoring the wp-prettier dependency in [dcd4d33](https://github.com/Automattic/newspack-scripts/pull/209/commits/dcd4d33ee7c803f99332ae5273202e313be9a916) does mostly fix the formatting conflicts! So I restored the `format:js` script in ebf4197a while also keeping the separate `fix:js` script for code quality fixes only. However, running this still results in a lot of (but fewer) changes to JS src files, so I won't include a wholesale reformatting of JS in this changeset.

miguelpeixe (commented):
No major issues while testing, the only problem I found is what appears to be a conflict between the auto js formatter and the typescript check, which is probably what @adekbadek also identified in his first review.

`npm run typescript:check` passes with the current formatting, but after the changes from `npm run format:js` I get a bunch of errors like this on the ts check:

```
src/wizards/engagement/components/prompt.tsx:245:20 - error TS2339: Property 'indexOf' does not exist on type 'string | number | boolean | number[] | string[]'.
  Property 'indexOf' does not exist on type 'number'.

245                 ]?.indexOf(
                       ~~~~~~~
``` 

I don't think it's a blocking issue for now but we'll need to figure out how to adapt the rules so we don't break ts once we commit the rules from the auto formatter.

dkoo:
@miguelpeixe I'd say the conflict between formatting and TS rules is non-blocking for now, as I don't think we should reformat our JS at this point because the newer rules in `@wordpress/scripts` would result in a huge number of changes in our JS across all repos. If we decide to change our JS formatting to match, we should do so in future PRs and separately from these other changes, and then we can resolve the conflict with our TS rules.

miguelpeixe (approved):

PR #3188, opened 2024-06-19
feat(everlit): added Everlit to plugins + refactor plugins in connections wizard

jaredrethman:
### All Submissions:

Everlit plugin management. Due to the nature of the UX implementation of this feature it was necessary to do a rewrite of some of the logic within the `<Plugins />` component used within Connections Wizard, namely:
* Ability to activate plugin when inactive and reload browser when activated.
* Different status, per/plugin, functionality.
* HTML within `<ActionCard />` descriptions i.e. `<a>Learn more</a>`.
* Plugin Manager configuration for Everlit
* Configuration Manager Class for Everlit to request setup status.
* Abstracted individual Plugins into a separate component `<Plugin />` to avoid re-renders on all ActionCards when a single plugins state changes.
* Loading indicators during requests.

1. Checkout this branch and build assets i.e. `git checkout origin/feat/wizard-everlit-plugin && npm run build`
2. Navigate to _/wp-admin/admin.php?page=newspack-connections-wizard_
3. Observer the new ActionCard for Everlit is present.

Screenshot 2024-06-19 at 09 47 32

4. Click "Install Everlit", once request has completed observe you see the following:

Screenshot 2024-06-19 at 09 48 53

5. Click "Activate Everlit". Observe:
   - Once request has completed, page will reload (Necessary for Everlit Plugin settings to show once activated) 
   - "Everlit" admin menu item exists after reload.
   - The Everlit plugin action card should resemble the below - prompting further setup. 

Screenshot 2024-06-19 at 09 55 33

6. Right click "Complete Setup" and open in a new tab, this will open a tab with Everlit initial setup. Add your details and click "Submit".

Screenshot 2024-06-19 at 09 58 23

7. Navigate back to Connections tab and refresh page. Observe that further setup is still necessary. 

Screenshot 2024-06-19 at 09 55 33

8. Navigate back to Everlit settings tab and refresh page. Observe that different configuration is now available e.g. "Everlit Token". Add a "Everlit Token" (for testing purposes this can be anything) and click "Save Changes"
9. Navigate back to Connections tab and refresh. You should now see a successfully configured Everlit plugin.

Screenshot 2024-06-19 at 10 10 48

### Other information:

### DIFF
--- a/includes/configuration_managers/class-everlit-configuration-manager.php
+++ b/includes/configuration_managers/class-everlit-configuration-manager.php
@@ -38,11 +38,23 @@ class Everlit_Configuration_Manager extends Configuration_Manager {
 	 * @return bool Plugin ready state.
 	 */
 	public function is_configured() {
+		if ( ! static::is_enabled() ) {
+			return false;
+		}
 		$toc = get_option( 'everlit_clickwrap_accepted', [] ) ?? [];
 		[ 'token' => $token ] = get_option( 'everlit_settings', [] ) ?? [];
 		if ( $this->is_active() && $token !== null && isset( $toc['timestamp'] ) ) {
 			return true;
 		}
 		return false;
+	}
+
+	/**
+	 * Newspack Manager conditional check.
+	 *
+	 * @return bool If Newspack Manager is installed and connected.
+	 */
+	public static function is_enabled() {
+		return method_exists( 'Newspack_Manager', 'is_connected_to_manager' ) && \Newspack_Manager::is_connected_to_manager();
 	}
 }
--- a/includes/configuration_managers/class-everlit-configuration-manager.php
+++ b/includes/configuration_managers/class-everlit-configuration-manager.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Everlit plugin configuration manager.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+require_once NEWSPACK_ABSPATH . '/includes/configuration_managers/class-configuration-manager.php';
+
+/**
+ * Provide an interface for configuring and querying the configuration of Everlit.
+ */
+class Everlit_Configuration_Manager extends Configuration_Manager {
+
+	/**
+	 * The slug of the plugin.
+	 *
+	 * @var string
+	 */
+	public $slug = 'everlit';
+
+	/**
+	 * Configure Everlit for Newspack use.
+	 *
+	 * @return bool Enforce configurability.
+	 */
+	public function configure() {
+		return true;
+	}
+
+	/**
+	 * Is Everlit installed and connected?
+	 *
+	 * @return bool Plugin ready state.
+	 */
+	public function is_configured() {
+		$toc = get_option( 'everlit_clickwrap_accepted', [] ) ?? [];
+		[ 'token' => $token ] = get_option( 'everlit_settings', [] ) ?? [];
+		if ( $this->is_active() && $token !== null && isset( $toc['timestamp'] ) ) {
+			return true;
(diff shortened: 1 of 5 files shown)


kmwilkerson:
Nice! We can remove the Install step (as it'll be installed by default) and just start right with the activation step

jaredrethman:
> Nice! We can remove the Install step (as it'll be installed by default) and just start right with the activation step

@kmwilkerson I can absolutely remove it. Here's some motivations for keeping it:
- If the plugin is already "installed by default" user will never see the prompt to install.
- If there is ever a situation where the plugin wasn't able to be installed by default (network issues etc) the below wouldn't be accurate.
image
- Make testing this review that much easier.

Let me know your thoughts on the above.

kmwilkerson:
I gotcha -- in that case, keeping it makes total sense @jaredrethman

includes/configuration_managers/class-everlit-configuration-manager.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @var string
+	 */
+	public $slug = 'everlit';
+
+	/**
+	 * Configure Everlit for Newspack use.
+	 *
+	 * @return bool Enforce configurability.
+	 */
+	public function configure() {
```
dkoo, line 31: Is this function necessary? I'm not seeing it used anywhere. Would there be any case we wouldn't want Everlit to be configurable? e.g. for non-hosted (OSS) Newspack sites?

dkoo (approved):
This is working as described. I also tested the flow in a single tab (rather than opening the Everlit settings screen in a new tab as described in the testing instructions) and the setup flow makes sense: after entering contact details, accepting the terms, and clicking continue, you're automatically redirected to the Everlit plugin settings page to enter a token.

I have one non-blocking question below.

dkoo (changes requested):
Whoops, I double-checked with @kmwilkerson and she said the Everlit plugin card should only appear in Connections for hosted Newspack clients. We'll need to make use of that `configure()` method after all. Will reach out with technical details.

includes/configuration_managers/class-everlit-configuration-manager.php
jaredrethman: `configure` method is called within `Newspack\Configuration_Managers` class. 
https://github.com/Automattic/newspack-plugin/blob/7ec149468ba0dfe000b9ce6d8ccfa22d66fd84bc/includes/configuration_managers/class-configuration-managers.php#L94-L106

Its required in every inheriting class due to abstract keyword:
https://github.com/Automattic/newspack-plugin/blob/7ec149468ba0dfe000b9ce6d8ccfa22d66fd84bc/includes/configuration_managers/class-configuration-manager.php#L104-L109

jaredrethman (commented):
> Whoops, I double-checked with @kmwilkerson and she said the Everlit plugin card should only appear in Connections for hosted Newspack clients. We'll need to make use of that configure() method after all. Will reach out with technical details.

@dkoo I'm not sure that `configure()` is where conditional logic should go, as it's for programmatic configuration, please confirm. 

Instead of adding the conditional logic to `configure` I have:
* Added the conditional to a static method
* Called above method in localized variable `window.newspack_connections_data.can_use_everlit`
* Referenced the localize variable before rendering.
* Referenced conditional in `is_configured` to mitigate invalid API requests.

includes/configuration_managers/class-everlit-configuration-manager.php
dkoo: Got it, thanks for clarifying!

dkoo (approved):
Working well now! Thanks for the updates, @jaredrethman. I'm going to merge this since it's a requirement for the next release cycle.

PR #3199, opened 2024-06-25
feat: update reader registration block to use newspack-ui

thomasguillot:
### All Submissions:

The main goal of this PR is to update the Reader Registration block to use Newspack UI, aligning it with the rest of the RAS-ACC flow.

I've removed the styles (Stacked / Columns) as they're no longer relevant, and simplified the "Sign in" field to just a link/button (removing the need for "Already have an account?"), to further align with RAS-ACC.

On the Newspack UI end, I've made a few changes: I've replaced Inter with a new version and added Italic. I've enqueued the script into the editor to make sure the styles are available there.

image

1. Add a Reader Registration block to a page.
2. Switch to this branch and make sure the block still looks fine.
3. Play with the different settings, etc...

### Other information:


adekbadek (changes requested):
This looks much better! Some issues I've found:

1. After clicking "Sign Up" or "Sign In with Google", the SSO button (Google login) disappears, which causes a layout shift. On the base branch it's greyed out along with the rest of the UI.
2. Custom Registration success state is not holding well. This is also the case on the base branch, BTW, but it's less broken on this branch:

| Editor | `epic/ras-acc` | This PR  
|--------------|-----------|---|
| image |  image | image

thomasguillot:
@adekbadek both issues should be fixed now.

With the Editor styles, I pushed some minor changes but it looked ok to me

| Editor | Front-end |
| ------- | -------- | 
| editor | front-end |

laurelfulford:
@adekbadek Thanks for keeping this one moving! It was on my list to follow up on it but I didn't get to it. Let me know if it'd help at all to have a second set of eyes reviewing since you're also making commits 🙂

laurelfulford:
Hey @adekbadek! I just wanted to follow up on this one again to see if there's anything still blocking it from a thumbs up? I'm happy to help with merge conflicts/a second set of eyes/making outstanding fixes, whatever! I think it'd be a nice refresh with the rest of the RAS-ACC changes -- thanks! 🙂

adekbadek (approved):

adekbadek:
@laurelfulford – sorry it took so long! 🙈 I've added some tweaks and approved the changes. Since I've contributed code, do you want to give it another review?

src/newspack-ui/scss/elements/forms/_buttons.scss
```diff
@@ -5,20 +5,22 @@
...
 		color: var(--newspack-ui-color-neutral-90);
+		cursor: pointer;
 		display: inline-flex;
 		font-family: var(--newspack-ui-font-family);
 		font-size: var(--newspack-ui-font-size-s);
 		font-weight: 600;
-		letter-spacing: initial;
-		text-transform: none;
-		gap: var(--newspack-ui-spacer-2);
+		gap: calc(var(--newspack-ui-spacer-base) / 2);
```
laurelfulford, line 13:
```suggestion
		gap: calc(var(--newspack-ui-spacer-base) / 2);
		letter-spacing: initial;
		text-transform: none;
```

These two styles should stay -- they seem redundant, but were to override some common Custom CSS interference we saw when testing clone sites 😅

src/newspack-ui/scss/elements/_typography.scss
```diff
@@ -16,7 +16,6 @@
 	h5,
 	h6 {
 		font-family: var(--newspack-ui-font-family);
```
laurelfulford:
```suggestion
		font-family: var(--newspack-ui-font-family);
		font-weight: var(--newspack-ui-font-weight-strong);
```

This font weight should stay -- the headings should be `600` and without it, we get a weird mix of `600` and `700`.

src/newspack-ui/scss/elements/_tables.scss
```diff
@@ -13,6 +13,9 @@
 			font-size: var(--newspack-ui-font-size-xs); // Prevent relative font size
 			padding: calc(var(--newspack-ui-spacer-base) / 2) 0;
 			vertical-align: top;
+			> * {
+				white-space: nowrap;
+			}
```
laurelfulford:
```suggestion
```
This `white-space` style should be removed (see https://github.com/Automattic/newspack-plugin/pull/3317) -- it was deliberately done to fix an issue with stuff wrapping in the Transaction details table.

src/reader-activation-newsletters/style.scss
```diff
@@ -5,16 +5,11 @@
 	}
 
 	p.details {
-		margin-bottom: 0;
+		margin-bottom: 4px;
```
laurelfulford:
```suggestion
		margin-bottom: 0;
```

Another little regression -- this was all done in https://github.com/Automattic/newspack-plugin/pull/3273

src/reader-activation-newsletters/style.scss
```diff
@@ -5,16 +5,11 @@
...
 	}
 
 	p.recipient {
-		margin-top: 4px;
-		margin-bottom: 0;
-	}
-
-	.newspack-newsletters-signup {
-		margin-top: var(--newspack-ui-spacer-5);
+		margin-top: 0;
```
laurelfulford:
```suggestion
		margin-top: 4px;
		margin-bottom: 0;
	}

	.newspack-newsletters-signup {
		margin-top: var(--newspack-ui-spacer-5);
```

So many nits! I'm sorry, Adam!

src/other-scripts/emails/style.scss
```diff
@@ -2,15 +2,7 @@
...
-	}
-
-	.wp-block-button {
-		margin-left: 0 !important;
-		margin-right: 0 !important;
-		padding: 0 !important;
-	}
+.editor-styles-wrapper .block-editor-block-list__layout .wp-block-button {
+	margin-left: 0 !important;
+	margin-right: 0 !important;
```
laurelfulford:
```suggestion
.editor-styles-wrapper .block-editor-block-list__layout {
	:not(code) {
		font-family: Arial, Helvetica, sans-serif !important;
	}

	.wp-block-button {
		margin-left: 0 !important;
		margin-right: 0 !important;
		padding: 0 !important;
	}
```

Small regression -- see https://github.com/Automattic/newspack-plugin/pull/3326

src/newspack-ui/scss/elements/woocommerce/_overrides.scss
```diff
@@ -21,18 +21,3 @@
 		}
 	}
 }
```
laurelfulford, line 23:
```suggestion
}

.woocommerce-account {
	// Inline sign up form max-width.
	.woocommerce-notices-wrapper:has(+ .newspack-reader-auth__inline-wrapper),
	.newspack-reader-auth__inline-wrapper {
		margin-left: auto;
		margin-right: auto;
		max-width: var(--newspack-ui-modal-width-s);
	}

	.newspack-reader-auth__inline-wrapper h2 {
		font-size: var(--newspack-ui-font-size-m);
		font-weight: 600;
	}
}
```

This all should stay -- I double-checked on this branch, and we're still using the hard-coded form (and not the block) on the /my-account page; this CSS just makes sure it looks a little less blown out and weird (see https://github.com/Automattic/newspack-plugin/pull/3292)

src/newspack-ui/scss/elements/misc/_reader-auth.scss
```diff
@@ -1,8 +1,7 @@
 // Re-adds some modal-like spacing for when the sign up form ends up embedded in a page.
-// Ignore a tag buttons as they will never be first..
 .newspack-ui {
 	&.newspack-reader-auth {
-		.newspack-ui__button:not(:first-child, a.newspack-ui__button):first-of-type {
+		.newspack-ui__button:not(:first-child):first-of-type {
```
laurelfulford:
```suggestion
		.newspack-ui__button:not(:first-child, a.newspack-ui__button):first-of-type {
```

This just affects the modal, not the block, and should be changed back -- see https://github.com/Automattic/newspack-plugin/pull/3267.

src/newspack-ui/scss/elements/forms/_checkbox-radio.scss
```diff
@@ -22,11 +26,14 @@
...
 		&:checked {
 			&::before {
+				content: "";
 				opacity: 1;
 			}
 		}
 
 		&:focus {
+			border-color: var(--newspack-ui-color-input-border);
+			box-shadow: none;
```
laurelfulford:
```suggestion
```

These got moved to the `:focus-visible` pseudoselector with some other accessiblity changes, so aren't needed here.

src/newspack-ui/scss/elements/forms/_checkbox-radio.scss
```diff
@@ -22,11 +26,14 @@
 
 		&:checked {
 			&::before {
+				content: "";
```
laurelfulford:
```suggestion
```

This is redundant to the `&::before` above this, so can be removed.

includes/templates/reader-activation/login-form.php
```diff
@@ -16,7 +16,7 @@
 
 \do_action( 'woocommerce_before_customer_login_form' );
 ?>
-<div class="newspack-ui newspack-reader-auth__inline-wrapper">
+<div class="newspack-ui">
```
laurelfulford:
```suggestion
<div class="newspack-ui newspack-reader-auth__inline-wrapper">
```

This CSS class should get re-added -- it's part of this fix and won't affect the block: https://github.com/Automattic/newspack-plugin/pull/3292

src/blocks/reader-registration/index.php
```diff
@@ -261,6 +240,9 @@ function render_block( $attrs, $content ) {
 					<div class="newspack-registration__main">
 						<div>
 							<?php Reader_Activation::render_third_party_auth(); ?>
+							<?php if ( Recaptcha::can_use_captcha( 'v2' ) ) : ?>
+								<?php Recaptcha::render_recaptcha_v2_container(); ?>
+							<?php endif; ?>
```
laurelfulford:
```suggestion
```

This bit is throwing a fatal when you have reCAPTCHA set to v2 because `render_recaptcha_v2_container()` doesn't exist in the `epic/ras-acc` branch. The approach to get the two versions working were different in the RAS-ACC branch than trunk (PR: https://github.com/Automattic/newspack-plugin/pull/3283); not sure if this PR happened to be made at some weird mid-point?

Best I can tell, just getting rid of this is enough to fix it (v2 works without it), but I'd appreciate a double-check against the PR where the reCAPTCHA stuff was changed, in case I'm missing something!

(v3 seems to work fine!)

laurelfulford (changes requested):
Thank you @adekbadek! 🙌 

It looks like a weird regression happened with the not-block-specific styles at some point, though I don't get how 😬 I picked through and cross checked against PRs, and I think I found everything that needs to be changed back. Let me know if I can help take care of those! 

There was also an issue with reCAPTCHA v2, also probably a merge thing -- I think I found the bit that needs to be changed, but I would definitely trust your eyes more than mine if anything else needs to be moved over.

Otherwise, the block itself looks good! The only visual issue I noticed was a weird side-scroll thing when the block's in a prompt and using reCAPTCHA v2, but that's also happening in the registration modal, and should be fixed with this PR: https://github.com/Automattic/newspack-plugin/pull/3401

CleanShot 2024-09-06 at 12 09 04

Just let me know if you have any questions about my comments, and if I can help with anything (we can always loop in another reviewer! 😄 ) Thanks!

laurelfulford:
Hey @adekbadek! Sorry for the late follow up, this completely fell off my radar. Is there anything I can do to help with the changes requested for this one? Happy to help make changes, or answer questions -- just let me know, thanks!

src/blocks/reader-registration/index.php
adekbadek: Ah, makes sense. Let's just follow what's in `epic/ras-acc`.

adekbadek:
Sorry for _my_ late follow up here. I've applied your suggestions – thank you!

src/newspack-ui/style.scss
```diff
@@ -3,15 +3,14 @@
...
 	// Need better spot for this
 	&__color-text-gray {
 		color: var(--newspack-ui-color-neutral-60);
 	}
 
 	// Overrides for theme styles -- to move to theme?
-	h2,
-	h3 {
+	h2:not([class*="font-size"]),
+	h3:not([class*="font-size"]) {
```
laurelfulford, line 13: It looks like these styles are affecting header Registration Block: 

CleanShot 2024-10-07 at 11 53 01

Compared to the mockup:

CleanShot 2024-10-07 at 11 50 08

It looks like the title is set to the large font size [here](https://github.com/Automattic/newspack-plugin/pull/3199/files#diff-073e276f3eb477174d7918af604a017b086f70e9e9ffc0d305c57158c23a24aeR98) so I think it's just a specificity thing.

[The theme uses classes with `font-size`](https://github.com/Automattic/newspack-theme/blob/0ccc70d6e686575b04ee2e24555b96af0a34dca2/newspack-theme/sass/blocks/_blocks.scss#L1420), but WordPress adds `!important` to those as part of Gutenberg. [Newspack UI just uses `font`](https://github.com/Automattic/newspack-plugin/blob/epic/ras-acc/src/newspack-ui/scss/elements/_typography.scss#L51C4-L51C12) -- I think it may be safe to remove the `:not` unless this is meant to protect against something else?

laurelfulford (changes requested):
Thanks @adekbadek! 🙌 I checked the fixes, and you took care of everything I spotted! 

I noticed one more small visual thing but it's minor. Once that's taken care of this should be good to go!

src/newspack-ui/style.scss
adekbadek: These styles were added because the less-specific selectors were overriding font sizes set in the editor:

| Editor | `epic/ras-acc` | this branch|  
|--------------|-----------|---|
| image | image | image |

I've added `!important`s in 96afca6e1a342dc2976cdd0c7caad8f5acadb755, would that be ok?

laurelfulford (approved):
> These styles were added because the less-specific selectors were overriding font sizes set in the editor

Ah! Thanks @adekbadek -- I totally missed that 🤦‍♀️ 

I think `!important` makes sense here if we want to control the font size that closely -- I'm still not sure if we need the `:not()` selectors but they're not causing other issues; if they do in the future we can deal with them then 🙂 

Thanks for wrangling this one to the finish! It looks good to me!

PR #3233, opened 2024-07-09
feat(co-authors-plus): option to block the Guest Authors menu item

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Newspack discourages the use of Co-Authors Plus' Guest Authors feature and offers a way to migrate them to WP Users (#3068). This PR adds an optional (behind `NEWSPACK_DISABLE_GUEST_AUTHORS` feature flag) blocking of the "Guest Authors" menu item.

### How to test the changes in this Pull Request:

1. Install Co-Authors Plus and visit the Users -> Guest Authors link, observe that Guest Authors can be added
2. Set `NEWSPACK_DISABLE_GUEST_AUTHORS` environment variable to true
3. Refresh the page, observe that the Users -> Guest Authors link leads to this message and Guest Authors can't be added here:

image

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3238, opened 2024-07-11
Updating and reorganizing CAP tweaks

leogermani:
### All Submissions:

Improves our implementation of an alternative to CAP Guest Authors.

1. Checkout this branch
2. Add `define( 'NEWSPACK_DISABLE_CAP_GUEST_AUTHORS', true );` to wp-config
3. Visit Users > Guest Authors and see the replacement page
4. Click Add new Guest Author and confirm you see the regular form to add a new user but with some modifications (no username, no password)
5. Create a user and confirm you can assign it as coauthor
6. Confirm you can not login as this user
7. Confirm the edit page for this user does not have a bunch of fields (username, password, application password, woo fields, profile colors, etc)
8. @adekbadek confirm that this transitions well from a site using the previous approach. (the user role has the same slug and the capability is no longer checked)

-- test author cap ---

Add the edit_cap_posts capability to a subscribe user and confirm you can assign them as coauthors. Remove the capability and make sure you no longer can.

Make sure all other roles that can edit posts still can be assigned as coauthors.

-- test user account deletion --

Login as subscriber with the special cap and try to delete its account. Confirm you get a message saying you can't

### Other information:

### DIFF
--- a/includes/plugins/co-authors-plus/class-coauthor-user-cap.php
+++ b/includes/plugins/co-authors-plus/class-coauthor-user-cap.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Co-Authors Plus integration class.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * This class creates a new capability and adds it to all roles that have the edit_posts capability
+ *
+ * Then it filters the capability used by Co-Authors Plus to determine which users can be assigned as co-authors.
+ */
+class Coauthor_User_Cap {
+
+	/**
+	 * Custom capability name.
+	 */
+	const ASSIGNABLE_TO_POSTS_CAPABILITY_NAME = 'edit_cap_posts';
+
+	/**
+	 * Option name to mark the version of the settings. If the implementation details
+	 * change, the expected option value should be updated to trigger a reset of the settings.
+	 */
+	const SETTINGS_VERSION_OPTION_NAME = 'newsroomnz_coauthors_plus_tweaks_settings_version';
+
+	/**
+	 * Initializes the class.
+	 *
+	 * @return void
+	 */
+	public static function init() {
+		add_filter( 'coauthors_edit_author_cap', [ __CLASS__, 'coauthors_edit_author_cap' ] );
+		add_action( 'admin_init', [ __CLASS__, 'setup_custom_capability' ] );
+	}
+
+	/**
+	 * Override the capability required to assign a user as a co-author.
+	 *
+	 * @param string $edit_cap Capability required for a user to be assigned as a co-author.
+	 */
+	public static function coauthors_edit_author_cap( $edit_cap ) {
+		return self::ASSIGNABLE_TO_POSTS_CAPABILITY_NAME;
+	}
+
+	/**
+	 * Add custom capability to the roles that should be allowed to author posts.
+	 */
+	public static function setup_custom_capability() {
+		$current_settings_version = '1';
+		if ( \get_option( self::SETTINGS_VERSION_OPTION_NAME ) === $current_settings_version ) {
+			return;
+		}
+
+		$wp_roles = wp_roles();
+		foreach ( $wp_roles->roles as $role_name => $role ) {
+			$role = $wp_roles->get_role( $role_name );
+			if ( $role->has_cap( 'edit_posts' ) || $role_name === Guest_Author_Role::CONTRIBUTOR_NO_EDIT_ROLE_NAME ) {
+				$role->add_cap( self::ASSIGNABLE_TO_POSTS_CAPABILITY_NAME );
+			}
+		}
+
+		\update_option( self::SETTINGS_VERSION_OPTION_NAME, $current_settings_version );
+	}
+}
+
+Coauthor_User_Cap::init();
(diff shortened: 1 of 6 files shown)


includes/plugins/co-authors-plus/class-guest-author-role.php
```diff
@@ -0,0 +1,277 @@
...
+
+		if ( \get_option( self::SETTINGS_VERSION_OPTION_NAME ) === $current_settings_version ) {
+			return;
+		}
+
+		\remove_role( self::CONTRIBUTOR_NO_EDIT_ROLE_NAME );
+
+		\add_role( // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.custom_role_add_role
+			self::CONTRIBUTOR_NO_EDIT_ROLE_NAME,
+			__( 'Guest Author', 'newspack-plugin' ),
```
adekbadek, line 78: This role is called "Non-Editing Contributor" now, it was already communicated as such to our publishers. It's intentionally different than "Guest Authors" to underline that it's something else than CoAuthors-Plus' Guest Authors. Someone used to creating Guest Authors should be directed to use this new feature instead, not confused by the familiar label cropping up in a different place.

includes/plugins/co-authors-plus/class-guest-author-role.php
```diff
@@ -0,0 +1,277 @@
...
+	 * @param string           $password The user's password.
+	 * @return WP_Error|WP_User
+	 */
+	public static function wp_authenticate_user( $user, $password ) {
+		if ( ! is_a( $user, 'WP_User' ) ) {
+			return $user;
+		}
+
+		if ( in_array( self::CONTRIBUTOR_NO_EDIT_ROLE_NAME, $user->roles, true ) ) {
+			return new WP_Error( 'guest_authors_cannot_login', __( 'Guest authors cannot login.', 'newspack-plugin' ) );
```
adekbadek, line 239: Why shouldn't these users be allowed to log in? The goal of the original implementation, in line with the requirements, was to allow _any_ user to become assignable to posts. Many sites have users who are subscribers (should be able to manage their subscriptions), but they've contributed to an article and are listed as authors. They should still be able to log in and manage their subscriptions. The only thing they should not be able to do is editing posts.

includes/plugins/co-authors-plus/class-coauthor-user-cap.php
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 * Custom capability name.
+	 */
+	const ASSIGNABLE_TO_POSTS_CAPABILITY_NAME = 'edit_cap_posts';
+
+	/**
+	 * Option name to mark the version of the settings. If the implementation details
+	 * change, the expected option value should be updated to trigger a reset of the settings.
+	 */
+	const SETTINGS_VERSION_OPTION_NAME = 'newsroomnz_coauthors_plus_tweaks_settings_version';
```
adekbadek: Should be:

```suggestion
	const SETTINGS_VERSION_OPTION_NAME = 'newspack_coauthors_plus_settings_version';
```

adekbadek (changes requested):
I like how this improves what's proposed in #3233 and the blocking of the account deletion request. The login blocking and role name changes are not needed here.

includes/plugins/co-authors-plus/class-guest-author-role.php
leogermani: I understand that, and the change here was also intentional.

The new overridden Guest Authors page point the user to use the Guest Authors role instead, so this change makes it consistent.

We are actually moving a feature the user knows from one place to another. Guest Authors used to live here, and now it lives there, but it's the same thing.

What could/should be called "Non-Editing Contributors" are users that have another "non editor" role but are cherry picked to be able to assigned as co authors (get the custom capability). Maybe we could add a section in their user profile to allow admins to grand this capability without the need of third-party plugins.

But these are just names. If you want we can keep the name and discuss it further. We just need to also change it in the new Guest Authors page replacement

includes/plugins/co-authors-plus/class-guest-author-role.php
leogermani: We are blocking the users with this role from logging in, just as we are not event allowing a password to be set.

User with the custom capability that allows them to be assigned as co authors can still login. They are not affected.

adekbadek: > User with the custom capability that allows them to be assigned as co authors can still login.

But users with multiple roles (e.g. `subscriber,contributor_no_edit`) won't be able to log in, and should.

adekbadek:
Should have asked earlier – the testing instructions state:

> Add the edit_cap_posts capability to a subscribe user

How to do it in WP Admin UI, as a "non-technical user"?

leogermani:
> Should have asked earlier – the testing instructions state:
> 
> > Add the edit_cap_posts capability to a subscribe user
> 
> How to do it in WP Admin UI, as a "non-technical user"?

Using a plugin that manages roles and capabilities.

But as I said I think we should add an option in the edit user screen.

adekbadek:
> Using a plugin that manages roles and capabilities.

I'm using `capability-manager-enhanced` for role-related stuff. Can't see any option to add a capability to a single user. Capabilities can only be added to roles with this plugin. Can you recommend a plugin to achieve that goal?

Another issue with this approach is that it makes it impossible for the publisher to see which users are assignable to posts in the WP Admin Users view. It can be filtered by roles, but not by capabilities.

leogermani:
closed in favor of https://github.com/Automattic/newspack-plugin/pull/3277 See pemhSX-Sz-p2

PR #3248, opened 2024-07-15
feat(cli-co-authors-plus): ensure correct author slug is set

adekbadek:
### All Submissions:

With the way CoAuthors-Plus works, the term corresponding to the author must have `slug` (with `'cap-'` prefix) set to the linked WP User's `slug` (aka `user_nicename`). This change ensures that this is always the case. 

1. Create a CAP Guest Author and link them to an existing WP User, take note of the Guest Author slug and archive URL
2. Run the migration: `wp newspack migrate-co-authors-guest-authors --verbose --live` 
3. Observe the author archive still works and that the term slug (the term ID will be printed in the output) matches the WP User's `user_nicename` (`wp user get <ID> --field=user_nicename`). 

### Other information:

### DIFF
--- a/includes/cli/class-co-authors-plus.php
+++ b/includes/cli/class-co-authors-plus.php
@@ -294,7 +294,7 @@ class Co_Authors_Plus {
 				WP_CLI::warning( sprintf( 'No term found for user %d.', $user_id ) );
 			}
 
-			$author_slug = str_replace( 'cap-', '', $guest_term->slug );
+			$author_slug = preg_replace( '/^cap-/', '', $guest_term->slug );
 
 			if ( self::$live ) {
 				if ( $wp_user_term ) {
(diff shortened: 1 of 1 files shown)


includes/cli/class-co-authors-plus.php
```diff
@@ -294,6 +294,8 @@ private static function migrate_linked_guest_authors() {
 				WP_CLI::warning( sprintf( 'No term found for user %d.', $user_id ) );
 			}
 
+			$author_slug = str_replace( 'cap-', '', $guest_term->slug );
```
leogermani: There is an edge case where a username can start with `cap-` and then its term slugs will start with `cap-cap-`. I know that because of this -> https://github.com/Automattic/Co-Authors-Plus/blob/1c29766eeda11276bcf6f20e07171b3efcbc7c03/php/class-coauthors-plus.php#L1355

e.g. "Cap Ri" -> "cap-cap-ri".

I think we need to account for that here, right? And only remove the first occurrence of `cap-` and only the one at the beginning of the string

adekbadek: Huh, good catch! Fixed in 17b7d10

leogermani (approved):

PR #3261, opened 2024-07-19
feat: add log context

leogermani:
### All Submissions:

Adds contextual logging to ESP data sync

This PR also makes the `put` method of the connectors `protected` as they are not meant to be used outside of this scope

See https://github.com/Automattic/newspack-manager/pull/240

### Other information:

### DIFF
--- a/includes/data-events/connectors/class-activecampaign.php
+++ b/includes/data-events/connectors/class-activecampaign.php
@@ -103,7 +103,7 @@ class ActiveCampaign extends Connector {
 			'email'    => $data['email'],
 			'metadata' => $metadata,
 		];
-		self::put( $contact, sprintf( 'Updating the account and newsletter_selection fields after a change in the subscription lists. User ID: %d', $data['user_id'] ) );
+		self::put( $contact, 'Updating the account and newsletter_selection fields after a change in the subscription lists.' );
 	}
 }
 new ActiveCampaign();
--- a/includes/data-events/connectors/class-activecampaign.php
+++ b/includes/data-events/connectors/class-activecampaign.php
@@ -103,7 +103,7 @@ class ActiveCampaign extends Connector {
 			'email'    => $data['email'],
 			'metadata' => $metadata,
 		];
-		self::put( $contact, 'Updating the account and newsletter_selection fields after a change in the subscription lists' );
+		self::put( $contact, sprintf( 'Updating the account and newsletter_selection fields after a change in the subscription lists. User ID: %d', $data['user_id'] ) );
 	}
 }
 new ActiveCampaign();
--- a/includes/data-events/connectors/class-activecampaign.php
+++ b/includes/data-events/connectors/class-activecampaign.php
@@ -51,14 +51,15 @@ class ActiveCampaign extends Connector {
 	/**
 	 * Upsert the contact.
 	 *
-	 * @param array $contact Contact info to sync to ESP.
+	 * @param array  $contact Contact info to sync to ESP.
+	 * @param string $context Context of the update for logging purposes.
 	 */
-	public static function put( $contact ) {
+	protected static function put( $contact, $context ) {
 		$master_list_id = Reader_Activation::get_setting( 'active_campaign_master_list' );
 		if ( ! $master_list_id ) {
 			return;
 		}
-		return \Newspack_Newsletters_Contacts::upsert( $contact, $master_list_id );
+		return \Newspack_Newsletters_Contacts::upsert( $contact, $master_list_id, false, $context );
 	}
 
 	/**
@@ -102,7 +103,7 @@ class ActiveCampaign extends Connector {
 			'email'    => $data['email'],
 			'metadata' => $metadata,
 		];
-		self::put( $contact );
+		self::put( $contact, 'Updating the account and newsletter_selection fields after a change in the subscription lists' );
 	}
 }
 new ActiveCampaign();
--- a/includes/data-events/connectors/class-connector.php
+++ b/includes/data-events/connectors/class-connector.php
@@ -118,7 +118,7 @@ abstract class Connector {
 			return;
 		}
 
-		static::put( $contact, 'RAS Woo Subscription updated' );
+		static::put( $contact, sprintf( 'RAS Woo Subscription updated. Status changed from %s to %s', $data['status_before'], $data['status_after'] ) );
 	}
 
 	/**
@@ -162,6 +162,7 @@ abstract class Connector {
 
 		$contact['metadata']['network_registration_site'] = $registration_site;
 
-		static::put( $contact, 'RAS Newspack Network: user propagated from another site in the network' );
+		$site_url = get_site_url();
+		static::put( $contact, sprintf( 'RAS Newspack Network: User propagated from another site in the network. Propagated from %s to %s.', $registration_site, $site_url ) );
 	}
 }
--- a/includes/data-events/connectors/class-connector.php
+++ b/includes/data-events/connectors/class-connector.php
@@ -51,7 +51,7 @@ abstract class Connector {
 			'metadata' => $metadata,
 		];
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Reader registration' );
 	}
 
 	/**
@@ -75,7 +75,7 @@ abstract class Connector {
 
 		$contact = WooCommerce_Connection::get_contact_from_customer( $customer );
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Reader login' );
 	}
 
 	/**
@@ -97,7 +97,7 @@ abstract class Connector {
 			return;
 		}
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Order completed' );
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


includes/data-events/connectors/class-connector.php
```diff
@@ -118,7 +118,7 @@ public static function subscription_updated( $timestamp, $data, $client_id ) {
 			return;
 		}
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Woo Subscription updated' );
```
claudiulodro:
```suggestion
		static::put( $contact, sprintf( 'RAS Woo Subscription updated. Subscription ID: %d. Status changed from %s to %s', $data['subscription_id'], $data['status_before'], $data['status_after'] ) );
```

includes/data-events/connectors/class-connector.php
```diff
@@ -51,7 +51,7 @@ public static function reader_registered( $timestamp, $data, $client_id ) {
 			'metadata' => $metadata,
 		];
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Reader registration' );
```
claudiulodro, line 54:
```suggestion
		static::put( $contact, sprintf( 'RAS Reader registration. User ID: %d. Data: %s.', $data['user_id'], wp_json_encode( $metadata ) ) );
```

includes/data-events/connectors/class-activecampaign.php
```diff
@@ -102,7 +103,7 @@ public static function newsletter_updated( $timestamp, $data ) {
 			'email'    => $data['email'],
 			'metadata' => $metadata,
 		];
-		self::put( $contact );
+		self::put( $contact, 'Updating the account and newsletter_selection fields after a change in the subscription lists' );
```
claudiulodro:
```suggestion
		self::put( $contact, sprintf( 'Updating the account and newsletter_selection fields after a change in the subscription lists. User ID: %d', $data['user_id'] ) );
```

includes/data-events/connectors/class-connector.php
```diff
@@ -75,7 +75,7 @@ public static function reader_logged_in( $timestamp, $data, $client_id ) {
 
 		$contact = WooCommerce_Connection::get_contact_from_customer( $customer );
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Reader login' );
```
claudiulodro, line 78:
```suggestion
		static::put( $contact, sprintf( 'RAS Reader login. User ID: %d', $data['user_id'] ) );
```

includes/data-events/connectors/class-connector.php
```diff
@@ -97,7 +97,7 @@ public static function order_completed( $timestamp, $data, $client_id ) {
 			return;
 		}
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Order completed' );
```
claudiulodro, line 100:
```suggestion
		static::put( $contact, sprintf( 'RAS Order completed. Order ID: %d', $order_id ) );
```

includes/data-events/connectors/class-connector.php
```diff
@@ -130,7 +130,7 @@ public static function subscription_updated( $timestamp, $data, $client_id ) {
 	 */
 	public static function reader_deleted( $timestamp, $data, $client_id ) {
 		if ( true === \Newspack\Reader_Activation::get_setting( 'sync_esp_delete' ) ) {
-			return Newspack_Newsletters_Contacts::delete( $data['user_id'] );
+			return Newspack_Newsletters_Contacts::delete( $data['user_id'], 'RAS Reader deleted' );
```
claudiulodro, line 133:
```suggestion
			return Newspack_Newsletters_Contacts::delete( $data['user_id'], sprintf( 'RAS Reader deleted. User ID: %d', $data['user_id'] ) );
```

includes/data-events/connectors/class-connector.php
```diff
@@ -162,6 +162,6 @@ public static function network_new_reader( $timestamp, $data, $client_id ) {
 
 		$contact['metadata']['network_registration_site'] = $registration_site;
 
-		static::put( $contact );
+		static::put( $contact, 'RAS Newspack Network: user propagated from another site in the network' );
```
claudiulodro:
```suggestion
		$site_url = get_site_url();
		static::put( $contact, 'RAS Newspack Network: User propagated from another site in the network. Propagated from %s to %s. User ID on %s: %d', $registration_site, $site_url, $site_url, $user_id ) );
```

claudiulodro (commented):
This is looking really solid and work great in my testing. I left some suggestions about how we can make the messages more useful. What do you think?

includes/data-events/connectors/class-connector.php
claudiulodro: On this one, not sure if we need all the metadata in my suggestion, since that data is also available in the logs.

includes/data-events/connectors/class-activecampaign.php
leogermani: Looks good

includes/data-events/connectors/class-connector.php
leogermani: Not sure about adding the metadata here. The context should not be too long and not human readable. It's the "why". The metadata should be added in that extra field. Isn't it there already? I'll have a look

includes/data-events/connectors/class-connector.php
leogermani: I just merged one similar suggestion but something occurred me.

One thing this makes it more difficult is to filter the logs by context. Let me see all the log entries for "RAS Order completed" for a certain site... for example. The user ID should be present already in the `data`. Same thing for the previous and new status of the subscriptions I just approved...

leogermani:
@claudiulodro I thought about the above discussion (https://github.com/Automattic/newspack-plugin/pull/3261#discussion_r1693585177) and this is my conclusion so far:

* I think we should not add too much information in this message, like the json encoded contact metadata, to keep it short and human readable
* I think we should not add information that would make the message too unique, making it impossible to filter by it (for example adding user or subscription IDs) - we want to be able to filter by this field
* I think it's ok to add site urls and previous and new statuses, as these result in a limited number of possible values.

So, for example, [this](https://github.com/Automattic/newspack-plugin/pull/3261#discussion_r1693480858) change looks good if we keep the previous and new statuses, but remove the subscription id.

[This change](https://github.com/Automattic/newspack-plugin/pull/3261#discussion_r1693484818) does not look good as it creates an unique string.

What do you think, sounds good?

claudiulodro:
I'm comfortable with either direction. My suggestions were aimed towards making the MailChimp notes more useful, so that a client or support can debug/understand the data. It would come with a tradeoff of making the logs harder to filter, but we would still be able to filter in the logs by e.g. `RAS Reader deleted*`.

includes/data-events/connectors/class-activecampaign.php
```diff
@@ -51,14 +51,15 @@ public static function register_handlers() {
 	/**
 	 * Upsert the contact.
 	 *
-	 * @param array $contact Contact info to sync to ESP.
+	 * @param array  $contact Contact info to sync to ESP.
+	 * @param string $context Context of the update for logging purposes.
 	 */
-	public static function put( $contact ) {
+	protected static function put( $contact, $context ) {
```
dkoo, line 57: Making these `protected` is probably a good idea, but some of our internal tools for syncing and importing subscriptions rely on these methods for now (I understand they should be refactored to use the new `Newspack_Newsletters_Contact` class methods, but that might not happen for a little while)

claudiulodro:
The latest version looks good! Just a couple little logging bits.
- Did you mean to do [this commit](https://github.com/Automattic/newspack-plugin/pull/3261/commits/8dd5456de2889983199d8eb330932c9c6acf0705)? It has a user ID is why I'm asking. :)
- The two logging suggestions I think we should keep are `Propagated from %s to %s` on the network logging and `Status changed from %s to %s` on the subscription logging.

includes/data-events/connectors/class-activecampaign.php
leogermani: which tools? I did try to search all of our repos.

We should change them before we release this, because if we just leave the method public, using them will trigger many actions that were not triggered before.

I think the best solution is to create a method in those tooling, wherever we need them. It's  simple call to the `Mailchimp_API` class that is still there. This will ensure that we make a direct request to Mailchimp without triggering any undesired hooks

leogermani:
@claudiulodro https://github.com/Automattic/newspack-plugin/pull/3261/commits/b30b4208b10cc377b113260cf35bb018647ab0da does the adjustments to the messages as we discussed.

When this is over, I'll add the possibility for us to add additional logging data to the `$contact`, so we can add more stuff to the `extra` data in the logs and also more information to Mailchimp Notes if we want

includes/data-events/connectors/class-activecampaign.php
leogermani: @dkoo if you point me where we were relying on this I can help migrate. And good thing is that we can migrate right away, we don't need to wait this PR

PR #3266, opened 2024-07-23
fix(ras-acc): qa testing design feedback

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207878733150039, https://app.asana.com/0/1207817176293825/1207878733150043, and https://app.asana.com/0/1207817176293825/1207878733150047

This PR adds various style fixes to RAS-ACC auth form:

1. Spacing between google icon and button text should be 12px:

Screenshot 2024-07-23 at 12 35 10

2. Normalize button font styles in newspack ui:

Screenshot 2024-07-23 at 12 36 07

3. Consistent font weights:

Screenshot 2024-07-23 at 12 36 58

This PR should be tested with the following Blocks PR checked out as well: https://github.com/Automattic/newspack-blocks/pull/1807

1. On a test site, add the following additional CSS via the Appearance > Customize menu:

```
button {	
  border-radius: 0;
  font-size: 0.7rem;
  letter-spacing: 0.1em;
  padding: 1rem 2.5rem;
  text-transform: uppercase;
}
```
2. As a reader trigger the auth modal by clicking the Sign In button at the top right of the page.
3. Verify the 3 points above
  - Spacing between Google G and button text is 12px
  - Button text is not capitalized and custom letter spacing is not being used
  - All font weights for bolded text are 600 

### Other information:

### DIFF
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -72,7 +72,6 @@
 			h2 {
 				font-family: var( --newspack-ui-font-family );
 				font-size: var( --newspack-ui-font-size-s );
-				font-weight: 600;
 				line-height: var( --newspack-ui-line-height-s );
 				margin: 0;
 			}
@@ -122,6 +121,10 @@
 			> *:last-child {
 				margin-bottom: 0;
 			}
+		}
+
+		h1, h2, h3, h4, h5, h6, strong {
+			font-weight: 600;
 		}
 	}
 }
--- a/assets/newspack-ui/scss/_modals.scss
+++ b/assets/newspack-ui/scss/_modals.scss
@@ -72,6 +72,7 @@
 			h2 {
 				font-family: var( --newspack-ui-font-family );
 				font-size: var( --newspack-ui-font-size-s );
+				font-weight: 600;
 				line-height: var( --newspack-ui-line-height-s );
 				margin: 0;
 			}
--- a/assets/newspack-ui/scss/elements/forms/_buttons.scss
+++ b/assets/newspack-ui/scss/elements/forms/_buttons.scss
@@ -9,6 +9,8 @@
 		font-family: var( --newspack-ui-font-family );
 		font-size: var( --newspack-ui-font-size-s );
 		font-weight: 600;
+		letter-spacing: initial;
+		text-transform: none;
 		gap: var( --newspack-ui-spacer-2 );
 		justify-content: center;
 		line-height: var( --newspack-ui-line-height-s );
--- a/assets/newspack-ui/scss/elements/forms/_buttons.scss
+++ b/assets/newspack-ui/scss/elements/forms/_buttons.scss
@@ -9,7 +9,7 @@
 		font-family: var( --newspack-ui-font-family );
 		font-size: var( --newspack-ui-font-size-s );
 		font-weight: 600;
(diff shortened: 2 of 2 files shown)


laurelfulford (changes requested):
Thanks for fixing this @chickenn00dle! 

Most of the issues look to be fixed - the only one that's not is the price summary bit -- it's still using the 700 weight:

image

In the PR it looks like the `h2` in the `newspack-ui__modal__header` was targeted for the `600` fotn weight fix -- maybe we can do something more generic where all headings and `strong` tags inside of `.newspack-ui` (or at least `newspack-ui__modal`) use `600`?

chickenn00dle:
> Most of the issues look to be fixed - the only one that's not is the price summary bit -- it's still using the 700 weight

Just checking, did you also check out https://github.com/Automattic/newspack-blocks/pull/1807 while testing this. That PR renders this text with a paragraph/strong tag instead.

> In the PR it looks like the h2 in the newspack-ui__modal__header was targeted for the 600 fotn weight fix -- maybe we can do something more generic where all headings and strong tags inside of .newspack-ui (or at least newspack-ui__modal) use 600?

This sounds good to me! Done in https://github.com/Automattic/newspack-plugin/pull/3266/commits/d7d5da2226b4acec3afc1277b1900ea07d631ee9

laurelfulford:
> Just checking, did you also check out https://github.com/Automattic/newspack-blocks/pull/1807 while testing this.

🤦‍♀️ Nope! I missed that step in the testing -- taking a second look!

laurelfulford (changes requested):
Ok, the more I look at this the more I'm confusing myself!! So please let me know if this doesn't make sense:

The block PR changing the markup from `h2` to `p` does fix the font weight issue but it looks like we have this price header thing coming from two different spots, so it's not fixed everywhere, and my CSS recommendation wasn't specific enough.

If you are not logged in and try to make a donation, [you get the one that's now wrapped in a paragraph tag](https://github.com/Automattic/newspack-blocks/blob/fix/qa-testing-design-feedback/src/modal-checkout/modal.js#L332):

image

However, if you are logged in and go right into the checkout flow, [it still uses an `h2` so it's still `700`](https://github.com/Automattic/newspack-blocks/blob/fix/qa-testing-design-feedback/includes/class-modal-checkout.php#L1126-L1132):

image

The modal checkout doesn't use the `newspack-ui__modal` inside of the iframe, just `newspack-ui__modal__content`, so the more general CSS fix isn't getting picked up there. 

Looking at the Components Figma doc, it uses 400, 600, and 700 weights; 700 is rare but we can't assume it won't be used at all. I think `<p><strong>` is probably the safest fix for both for now, but let me know what you think!

(Sorry this was so long winded! Let me know if you have any questions about this!)

chickenn00dle:
Thanks for digging into this @laurelfulford!

> However, if you are logged in and go right into the checkout flow, [it still uses an h2 so it's still 700](https://github.com/Automattic/newspack-blocks/blob/fix/qa-testing-design-feedback/includes/class-modal-checkout.php#L1126-L1132):

Aha! Good catch. Pushed up a change in the blocks PR to address this case: https://github.com/Automattic/newspack-blocks/pull/1807/commits/a31d9ca98ec2b5bc5bab7311023794f2fd9770f0

laurelfulford (approved):
That did it -- thanks @chickenn00dle! 🚢

PR #3268, opened 2024-07-23
fix: force active membership if the user has active subs

dkoo:
### All Submissions:

This is a third attempt at addressing the problem described in #3060 and #3050, where if a user has more than one subscription and not all are active, any of the user's memberships tied to the subscriptions' products could be incorrectly expired at any time.

This implements two new checks at runtime:

1. When checking the user membership's status
2. When checking whether the user has access to a restricted post

The first should run only for memberships with non-active statuses, and only if the membership's plan is tied to subscription product ownership. If it is and the user does have an active subscription with the required subscription product(s), it will update the membership to `active` status and relink the membership to the active subscription. 

The second check is a fallback of sorts at the point where the user is trying to view restricted content. It checks if the membership requires any subscription products and allows access as long as the user has an active subscription with one of those products. Since the membership updates from the first check could take longer than it takes for the restricted content to load in the browser, this second check should avoid a race condition the first time any affected users try to view restricted content, but maybe it's not needed since the first check should permanently fix the membership status.

The major piece that was missing from #3050 was `WC_Memberships_User_Membership::unschedule_expiration_events()`—without calling this, updating an expired membership to `active` will result in an infinite loop of active/expire actions that will severely impact site performance. If we call that method before updating the status to `active`, the check should run exactly once for any impacted memberships, so in theory this PR should have no negative effect on site performance.

1. Set up a membership plan that requires ownership of a subscription product(s) and set it to restrict content (e.g. of a post category).
3. As a reader, purchase the required subscription and confirm that you can view the restricted content.
4. As an admin, manually update the reader subscription to "Expired" status.
5. As the reader, confirm that you can no longer view the restricted content.
6. Still as the reader, purchase another subscription to the required product.
7. On `trunk`, observe that you still can't view restricted content despite owning an active subscription with the required product.
8. As an admin, observe that the reader's membership still shows "Expired" status and is tied to the reader's expired subscription, instead of the newer active one.
9. Check out this branch and refresh the membership either in the admin or the restricted content as the reader.
10. Confirm that the reader can now view restricted content, and that the membership is updated to "Active" status, relinked to the active subscription, and that it doesn't continually try to expire and reactivate itself in an infinite loop that eventually crashes your entire site.

### Other information:

---
  - https://app.asana.com/0/0/1207858308890295

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -933,14 +933,14 @@ class Memberships {
 	 * @return string
 	 */
 	public static function check_membership_status( $post_status, $post ) {
-		if ( 'wc_user_membership' !== $post->post_type || ! self::has_memberships() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
+		if ( 'wc_user_membership' !== $post->post_type || 'wcm-active' !== $post->post_status || ! self::has_memberships() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
 			return $post_status;
 		}
 		$integrations = wc_memberships()->get_integrations_instance();
 		$integration  = $integrations ? $integrations->get_subscriptions_instance() : null;
 		$membership   = wc_memberships_get_user_membership( $post->ID );
 		$plan_id      = $membership->get_plan_id();
-		if ( ! $membership->has_status( 'active' ) && $integration && $integration->has_membership_plan_subscription( $plan_id ) ) {
+		if ( $integration && $integration->has_membership_plan_subscription( $plan_id ) ) {
 			$subscription_plan    = new \WC_Memberships_Integration_Subscriptions_Membership_Plan( $plan_id );
 			$required_products    = $subscription_plan->get_subscription_product_ids();
 			$active_subscriptions = WooCommerce_Connection::get_active_subscriptions_for_user( $membership->get_user_id(), $required_products );
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -933,7 +933,7 @@ class Memberships {
 	 * @return string
 	 */
 	public static function check_membership_status( $post_status, $post ) {
-		if ( 'wc_user_membership' !== $post->post_type || 'wcm-active' !== $post->post_status || ! self::has_memberships() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
+		if ( 'wc_user_membership' !== $post->post_type || 'wcm-active' === $post->post_status || ! self::has_memberships() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
 			return $post_status;
 		}
 		$integrations = wc_memberships()->get_integrations_instance();
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -83,7 +83,7 @@ class Memberships {
 	/**
 	 * Check if Memberships is available.
 	 */
-	public static function has_memberships() {
+	public static function is_active() {
 		return class_exists( 'WC_Memberships' ) && function_exists( 'wc_memberships' );
 	}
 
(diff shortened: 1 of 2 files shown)


includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -78,6 +80,13 @@ public static function init() {
 		include __DIR__ . '/class-metering.php';
 	}
 
+	/**
+	 * Check if Memberships is available.
+	 */
+	public static function has_memberships() {
```
adekbadek: Nit: the method name suggests to me that it checks if an entity (site/user) has "memberships", not whether a plugin is active.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -160,20 +195,18 @@ public static function get_last_successful_order( $customer ) {
...
+
 		// Prioritize any currently active subscriptions.
-		$user_subscriptions = \wcs_get_users_subscriptions( $customer->get_id() );
-		if ( ! empty( $user_subscriptions ) ) {
-			foreach ( $user_subscriptions as $subscription ) {
-				if ( $subscription->has_status( self::ACTIVE_SUBSCRIPTION_STATUSES ) ) {
-					return $subscription;
-				}
-			}
+		$active_subscriptions = self::get_active_subscriptions_for_user( $user_id() );
```
adekbadek: 🤔 
```suggestion
		$active_subscriptions = self::get_active_subscriptions_for_user( $user_id );
```

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -160,20 +195,18 @@ public static function get_last_successful_order( $customer ) {
...
+		if ( ! empty( $active_subscriptions ) ) {
+			return reset( $active_subscriptions );
 		}
 
 		// If no active subscriptions, get the most recent completed order.
 		// See https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query for query args.
 		$args = [
-			'customer_id' => $customer->get_id(),
+			'customer_id' => $user_id(),
```
adekbadek:
```suggestion
			'customer_id' => $user_id,
```

adekbadek (changes requested):
Works as described! There are two typos to fix in `get_last_successful_order` method.

The testing instructions mention comparing to the `release` branch. Should this PR be a hotfix?

dkoo:
> The testing instructions mention comparing to the `release` branch. Should this PR be a hotfix?

It was originally a hotfix, but I walked that back since it's a pretty major change that warrants more careful QA, I think. I updated the testing instructions.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
dkoo: Whoops! Fixed in 01d8e275

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
dkoo: Whoops! Fixed in 01d8e275

includes/plugins/wc-memberships/class-memberships.php
dkoo: Good point! 01d8e275 updates this to `is_active()` instead.

adekbadek (approved):

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -160,20 +195,18 @@ public static function get_last_successful_order( $customer ) {
...
+
 		// Prioritize any currently active subscriptions.
-		$user_subscriptions = \wcs_get_users_subscriptions( $customer->get_id() );
-		if ( ! empty( $user_subscriptions ) ) {
-			foreach ( $user_subscriptions as $subscription ) {
-				if ( $subscription->has_status( self::ACTIVE_SUBSCRIPTION_STATUSES ) ) {
-					return $subscription;
-				}
-			}
+		$active_subscriptions = self::get_active_subscriptions_for_user( $user_id );
```
miguelpeixe, line 201: This method returns an array of integers, which breaks the expected return of `get_last_successful_order()`. This causes the `get_contact_from_customer()` to throw a fatal error by trying to `->get_id()` from the integer and breaks data event handlers that rely on that method.

dkoo: Thanks for catching! I'll have a look and push an "alpha hotfix" for it

dkoo: PR here: https://github.com/Automattic/newspack-plugin/pull/3299

PR #3270, opened 2024-07-24
feat(ia): advertising > display ads

jaredrethman:
### All Submissions:

* New Advertising admin page.
* New sub menu pages; Display Ads (Full) & Sponsors (Initial).
* Migrate legacy advertising wizard to new IA location. 
* Migrated Add-Ons tab to Settings tab.
* Migrated Suppressions tab to Settings tab.
* Added titles to each tab and section.
* CSS changes to accommodate title additions and tighten up component spacing.

1. Checkout this code and build assets.
2. Navigate to _/wp-admin/admin.php?page=advertising-display-ads#/_
3. Observe each tab; Providers, Placement and Settings all load as expected.
4. On Settings tab confirm Suppressions and Plugins (previously Add-Ons) are present.

### Other information:

### DIFF
--- a/includes/wizards/advertising/class-advertising-display-ads.php
+++ b/includes/wizards/advertising/class-advertising-display-ads.php
@@ -411,7 +411,7 @@ class Advertising_Display_Ads extends Wizard {
 			$ad_units = $configuration_manager->get_ad_units();
 		} catch ( \Exception $error ) {
 			$message = $error->getMessage();
-			$error   = new WP_Error( 'newspack_ad_units', $message ? $message : __( 'Ad Units failed to fetch.', 'newspack' ) );
+			$error   = new WP_Error( 'newspack_ad_units', $message ? $message : __( 'Ad Units failed to fetch.', 'newspack-plugin' ) );
 		}
 
 		if ( \is_wp_error( $ad_units ) ) {
--- a/includes/wizards/advertising/class-advertising-display-ads.php
+++ b/includes/wizards/advertising/class-advertising-display-ads.php
@@ -1,0 +1,561 @@
+<?php
+/**
+ * Newspack's Advertising Wizard
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+use WP_Error;
+
+use Newspack_Ads\Providers\GAM_Model;
+
+defined( 'ABSPATH' ) || exit;
+
+require_once NEWSPACK_ABSPATH . '/includes/wizards/class-wizard.php';
+
+/**
+ * Easy interface for setting up general store info.
+ */
+class Advertising_Display_Ads extends Wizard {
+
+	const NEWSPACK_ADVERTISING_SERVICE_PREFIX = '_newspack_advertising_service_';
+
+	const SERVICE_ACCOUNT_CREDENTIALS_OPTION_NAME = '_newspack_ads_gam_credentials';
+
+	// Legacy network code manually inserted.
+	const OPTION_NAME_LEGACY_NETWORK_CODE = '_newspack_ads_service_google_ad_manager_network_code';
+
+	// GAM network code pulled from user credentials.
+	const OPTION_NAME_GAM_NETWORK_CODE = '_newspack_ads_gam_network_code';
+
+	/**
+	 * The slug of this wizard.
+	 *
+	 * @var string
+	 */
+	protected $slug = 'advertising-display-ads';
+
+	/**
+	 * The capability required to access this wizard.
+	 *
+	 * @var string
+	 */
+	protected $capability = 'manage_options';
+
+	/**
+	 * Supported services.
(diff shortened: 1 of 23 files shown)


includes/wizards/advertising/class-advertising-sponsors.php
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	protected $capability = 'manage_options';
+
+	/**
+	 * Get the name for this wizard.
+	 *
+	 * @return string The wizard name.
+	 */
+	public function get_name() {
+		return esc_html__( 'Sponsors', 'newspack' );
```
dkoo:
```suggestion
		return \esc_html__( 'Sponsors', 'newspack-plugin' );
```

includes/wizards/advertising/class-advertising-display-ads.php
```diff
@@ -0,0 +1,561 @@
...
+		add_action( 'rest_api_init', array( $this, 'register_api_endpoints' ) );
+	}
+
+	/**
+	 * Get the name for this wizard.
+	 *
+	 * @return string The wizard name.
+	 */
+	public function get_name() {
+		return \esc_html__( 'Advertising', 'newspack' );
```
dkoo:
```suggestion
		return \esc_html__( 'Advertising', 'newspack-plugin' );
```

includes/wizards/advertising/class-advertising-display-ads.php
```diff
@@ -0,0 +1,561 @@
...
+	private function retrieve_data() {
+		$configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'newspack-ads' );
+
+		$services = $this->get_services();
+		$error    = false;
+		try {
+			$ad_units = $configuration_manager->get_ad_units();
+		} catch ( \Exception $error ) {
+			$message = $error->getMessage();
+			$error   = new WP_Error( 'newspack_ad_units', $message ? $message : __( 'Ad Units failed to fetch.', 'newspack' ) );
```
dkoo:
```suggestion
			$error   = new WP_Error( 'newspack_ad_units', $message ? $message : __( 'Ad Units failed to fetch.', 'newspack-plugin' ) );

dkoo (changes requested):
Working as advertised (heh). 

The only thing I caught was the lack of whitespace between the ad units when configuring GAM:

**`trunk`**

Screenshot 2024-07-30 at 12 07 56 PM

**`feat/ia-ads`**

Screenshot 2024-07-30 at 12 08 11 PM

All other functionality works as expected, except the Plugin cards (e.g. the Configure links after installing—I expect this will be fixed in a future PR since the testing instructions here only mention that they should be appearing in the right place, right).

Something non-blocking I just noticed in this PR but haven't been looking out for: we have a mismatch in some of our translation text domains throughout the Newspack Plugin due to legacy code. The text domain should be `newspack-plugin`, matching the plugin slug, but there are many instances of `newspack` throughout. It's not a blocker at all, but when editing surrounding code I try to update those if I see them. I highlighted all the instances I saw in this changeset, but like I said I don't consider these blockers.

jaredrethman:
@dkoo thanks for the feedback, responses below.

> Working as advertised (heh).

🤣 

> The only thing I caught was the lack of whitespace between the ad units when configuring GAM:

Added back.

> All other functionality works as expected, except the Plugin cards (e.g. the Configure links after installing—I expect this will be fixed in a future PR since the testing instructions here only mention that they should be appearing in the right place, right).

Have added in page anchor navigation.

> Something non-blocking I just noticed in this PR but haven't been looking out for: we have a mismatch in some of our translation text domains throughout the Newspack Plugin due to legacy code. The text domain should be newspack-plugin, matching the plugin slug, but there are many instances of newspack throughout. It's not a blocker at all, but when editing surrounding code I try to update those if I see them. I highlighted all the instances I saw in this changeset, but like I said I don't consider these blockers.

Yeah, I've noticed these too. I've replaced all text domains in the newer stuff, but neglected to do so for the migration of the older stuff. Will keep an eye out for this going forward.

dkoo:
> Have added in page anchor navigation.

Works well after a page refresh, but since some of the settings sections aren't rendered if we don't have the plugin requirements, if you click "Configure" immediately after installing it will throw a JS error about the element being undefined. Maybe we could trigger a page refresh after the install to get around that?

jaredrethman:
> Works well after a page refresh, but since some of the settings sections aren't rendered if we don't have the plugin requirements, if you click "Configure" immediately after installing it will throw a JS error about the element being undefined. Maybe we could trigger a page refresh after the install to get around that?

Nice catch. Have added `shouldRefreshAfterUpdate` property to both plugin objects.

dkoo (approved):
Looking good now, thanks for the updates!

PR #3272, opened 2024-07-24
fix(ras-acc): redirect to homepage when accessing my account in logged out state

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207836618077219/f

This PR redirects readers to the homepage after logging out via my account. This is because we now provide a signin option from every other page and it's not necessary to visit the my-account page to login anymore when RAS is setup:

https://github.com/user-attachments/assets/a150fcc9-f6d6-4931-97f3-e6b1ddefa998

This also handles cases where folks are logging in from the my account page, forcing a redirect to trigger the updated my account dashboard.

1. Log in as a reader
2. Visit the my account page via the my account button or appending `/my-account` to the site url
3. Log in and confirm the page redirects to the my account dashboard 
4. Log out and confirm you redirected to the home page

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1152,12 +1152,8 @@ final class Reader_Activation {
 		}
 
 		/** Do not render link for authenticated readers if account page doesn't exist. */
-		if ( empty( $account_url ) ) {
-			if ( \is_user_logged_in() ) {
-				return '';
-			} else {
-				$account_url = '#';
-			}
+		if ( empty( $account_url ) && \is_user_logged_in() ) {
+			return '';
 		}
 
 		$class = function( ...$parts ) {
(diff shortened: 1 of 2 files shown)


chickenn00dle:
I brought this idea up a while back here: p1716953700550499-slack-C0390J7K0KY

The downside is we lose access to the WC notice that says you have been logged out:

Screenshot 2024-07-24 at 15 09 26

But given this matches our login experience (no notice), I don't think this is a problem.

The upside is you no are no longer directed to the awkward sign in form immediately after logging out.

Not sure what is best here, but thought I'd get this PR up and get some design review.

cc @kevinzweerink

laurelfulford (changes requested):
This works as described from a functional perspective, but making /my-account redirect to home for not logged in users makes me a little nervous 😬 I'm not sure we can say for certain no one is linking to that page as a means to expose a log in path, since publishers don't have a way to trigger the modal directly.

What do you think about limiting this to logging out? 

(BTW I may have missed some discussion around this! If so, please let me know!)

chickenn00dle:
> I'm not sure we can say for certain no one is linking to that page as a means to expose a log in path, since publishers don't have a way to trigger the modal directly.

Ah. Okay. If it helps, the redirect would only happen when Reader Activation is enabled. But I'm happy to play it safe here and limit this to only logging out.

> What do you think about limiting this to logging out?

WFM! I've updated this in https://github.com/Automattic/newspack-plugin/pull/3272/commits/1901a4386c1767bee0d74fd5e10be7571e1e45c4

I also had to apply a fix for the auth form submission when already on my account in https://github.com/Automattic/newspack-plugin/pull/3272/commits/5c59e39db9dd1f96ad7aef2874b3868e315e622b since its now possible for readers to access the login form via this page when logged out.

I've updated the testing instructions. Let me know if you have any questions!

laurelfulford (approved):
Thanks @chickenn00dle! I'm probably more paranoid about this kind of stuff than I should be, I appreciate you making this change 😅 Especially because:

> I also had to apply a fix for the auth form submission when already on my account in https://github.com/Automattic/newspack-plugin/commit/5c59e39db9dd1f96ad7aef2874b3868e315e622b since its now possible for readers to access the login form via this page when logged out.

I absolutely didn't consider this, so I'm so glad you caught it and fixed it! 

Tested, and this works well!

kevinzweerink:
Looks good from my end — agree with Laurel's mild hesitation around making `/my-account` redirect to home for all logged out users, but if we're only applying the redirect after the user logs out I think this is a good, uncontroversial change.

PR #3206, opened 2024-06-27
fix: display transaction fee as an amount instead of a percentage

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Displays the transaction fee as an amount instead of a percentage in the Donate modal checkout flow.

### How to test the changes in this Pull Request:

1. Check out this branch
2. While using the Newspack Reader Revenue platform, in **Newspack > Reader Revenue > Stripe Settings**, enable the "Allow donors to cover transaction fees" option but leave the custom message field blank.
3. As a reader, start a donation transaction and proceed to the payment part of the modal.
4. Confirm that the default message shows the fee in an absolute currency amount instead of as a percentage:

Screenshot 2024-06-27 at 4 21 55 PM

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207669785215348

PR #3275, opened 2024-07-25
fix(ras-acc): remove html markup from transaction fee order review table entry

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207836618927981/f

This PR removes the HTML markup from the Stripe transaction fee name. We need to do this because Woo will run the name string through an escaping function that causes the HTML markup to render in the WC order review table.

Before:
Screenshot 2024-07-25 at 10 56 34

After:
Screenshot 2024-07-25 at 10 57 05

1. Make sure stripe is enabled and the `Allow donors to cover transaction fees` checkbox is checked via the Newspack > Reader Revenue > Stripe menu
2. As a reader, go to any page with a donate block and initiate checkout
3. On the payment step of the modal, choose stripe as the payment method and check the "cover fees" checkbox
4. Look for the transaction table (pictured above), and confirm the Transaction Fee entry is added and there is no stray HTML markup

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
@@ -76,11 +76,7 @@ class WooCommerce_Cover_Fees {
 			return;
 		}
 		$cart->add_fee(
-			sprintf(
-				// Translators: %s is the fee percentage.
-				__( 'Transaction fee (%s)', 'newspack-plugin' ),
-				self::get_cart_fee_display_value()
-			),
+			__( 'Transaction fee', 'newspack-plugin' ),
 			self::get_cart_fee_value()
 		);
 	}


includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -76,11 +76,7 @@ public static function add_transaction_fee( $cart ) {
 			return;
 		}
 		$cart->add_fee(
-			sprintf(
```
chickenn00dle, line 79: I'm actually not sure in what other contexts we might be showing this fee name within Newspack, so it's possible we may need to go with another solution if the `($12.34)` bit of the fee name is needed elsewhere. I wasn't able to find anywhere else this was used, but let me know if I've missed something.

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -76,11 +76,7 @@ public static function add_transaction_fee( $cart ) {
...
-				// Translators: %s is the fee percentage.
-				__( 'Transaction fee (%s)', 'newspack-plugin' ),
-				self::get_cart_fee_display_value()
```
chickenn00dle, line 82: I'm actually not sure in what other contexts we might be showing this fee name within Newspack, so it's possible we may need to go with another solution if the `($12.34)` bit of the fee name is needed elsewhere. I wasn't able to find anywhere else this was used, but let me know if I've missed something.

laurelfulford:
@chickenn00dle I'm still trying to figure out "why" but it looks like this issue is happening in the release version of the plugin, too:

image

(I'll follow up in the Asana task with a link to the site in question).

What do you think about reworking this against trunk, possibly as a hotfix?

laurelfulford:
It looks like this issue was introduced here: https://github.com/Automattic/newspack-plugin/pull/3206

It was meant to affect the total displayed with the checkbox, to change it from a percentage to a total amount being covered -- from this:

image

to this: 

image

... but the value's used in the table as well, and there it should still be a percentage:

image

(I'm also noticing looking at this that we've lost a `(` for the GST value at some point, for some reason 😅 ).

chickenn00dle:
Thanks for catching this @laurelfulford!

> What do you think about reworking this against trunk, possibly as a hotfix?

I'm happy to do this. This would be targeting `release` not `trunk` right? Or is it possible to push a hotfix to `trunk` then `release`?

> but the value's used in the table as well, and there it should still be a percentage

It looks like the [same functionality used to generate the fee display value](https://github.com/Automattic/newspack-plugin/blob/93806e5a38f2f0d620dfcd5cd647718da3715bd5/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php#L288) is used when [setting the fee name](https://github.com/Automattic/newspack-plugin/blob/93806e5a38f2f0d620dfcd5cd647718da3715bd5/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php#L82), which is used by Woo when populating the transaction order review table.

So we can get a fix up in release to either simplify the fee name, or to create a new helper function to generate the fee percentage as a string and use this when setting the fee name instead.

My preference is to remove the additional information from the fee name as it seems Woo expects a simple fee name here, and we are already providing a total amount in the table. But wdyt?

laurelfulford:
> I'm happy to do this. This would be targeting release not trunk right? 

Yep, against `release` 🤦‍♀️ I was trying to say either fix in `trunk` _or_ as a hotfix, but ended up with something pretty garbled! 

> My preference is to remove the additional information from the fee name as it seems Woo expects a simple fee name here, and we are already providing a total amount in the table. But wdyt?

The table's currently showing percentages for the other values (like tax) so I'd be inclined to keep it (especially since that's what we have been showing there up until now) but either is better than what it's doing now! 🙂

chickenn00dle:
Closing in favor of this hotfix: https://github.com/Automattic/newspack-plugin/pull/3280

PR #3276, opened 2024-07-25
fix(ras-acc): correctly render auth modal response message and success button spacing

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207878733150134/f and https://app.asana.com/0/1207817176293825/1207878733150140/f

This PR ensures the auth modal response messages render correctly:

Screenshot 2024-07-25 at 13 02 39

It also removes extra padding from success modal Continue button when logging in with a password:

Before:
Screenshot 2024-07-25 at 15 08 43

After:
Screenshot 2024-07-25 at 15 05 55

1. As a reader (with an account but without a password) trigger the auth modal via the Sign in button (top right navigation)
2. Enter an invalid OTP code and confirm an error message appears and the input label and message is styled red
3. Wait for the OTP resend button timer to run down, then click the Resend code button
4. Confirm helper text appears styled in gray letting you know the code has been resent (pictured above)
5. Close the auth form without logging in
6. As a reader (with an account that has a password set) trigger the auth modal again
7. Complete the sign in process and stop at the success modal
8. Confirm the Continue button does not have additional padding at the bottom

### Other information:

### DIFF
--- a/assets/reader-activation-auth/style.scss
+++ b/assets/reader-activation-auth/style.scss
@@ -7,6 +7,10 @@
 	input[type='email'].nphp {
 		@include mixins.visuallyHidden;
 	}
+
+	.response-container {
+		margin-bottom: 12px;
+	}
 }
 
 .entry-content .newspack-reader,
@@ -179,8 +183,4 @@ li.menu-item .newspack-reader__account-link {
  */
 .woocommerce-account .entry-header .entry-title {
 	display: none;
-}
-
-.response-container {
-	margin-bottom: 12px;
 }
--- a/assets/reader-activation-auth/style.scss
+++ b/assets/reader-activation-auth/style.scss
@@ -180,3 +180,7 @@ li.menu-item .newspack-reader__account-link {
 .woocommerce-account .entry-header .entry-title {
 	display: none;
 }
+
+.response-container {
+	margin-bottom: 12px;
+}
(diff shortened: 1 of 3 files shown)


assets/reader-activation-auth/style.scss
```diff
@@ -7,6 +7,10 @@
 	input[type='email'].nphp {
 		@include mixins.visuallyHidden;
 	}
+
+	.response-container {
+		margin-bottom: 12px;
```
laurelfulford, line 12: I was going to comment that this could use the Newspack UI variables, then realized it's using a completely different CSS file 😅 

This is probably a question that goes beyond this PR, but is this stylesheet needed in the new modals? When I remove the `newspack-reader-auth` class from the modal and it doesn't seem to change the styles but I may be missing something! 

Just to keep stuff together, I think it may make sense to make a non-error version of the `newspack-ui__inline-error` that got removed from this element, and use that to handle the spacing in the Newspack UI styles instead -- what do you think?

laurelfulford (changes requested):
This works great! I left one comment about keeping the CSS with the Newspack UI stuff since we may need this message-like style for something else. Let me know if that doesn't make sense, or if I can help make that tweak!

chickenn00dle:
Thanks @laurelfulford!

> Just to keep stuff together, I think it may make sense to make a non-error version of the newspack-ui__inline-error that got removed from this element, and use that to handle the spacing in the Newspack UI styles instead -- what do you think?

I'm absolutely fine with this! That said, I think we can handle this in a follow-up PR so as not to hold up the response message rendering fix. If that's cool with you, we can probably create a low priority task in the main RAS-ACC project board.

laurelfulford (approved):
> I think we can handle this in a follow-up PR so as not to hold up the response message rendering fix

Ah, that's a fair point -- I'm good with that! 🙂

PR #3277, opened 2024-07-25
feat: enhancements for the Non editing contributor role

leogermani:
### All Submissions:

Enhances the Non-Editing Contributor role with a few new features

* If the user is only under this role, block login and remove some options from their profile
 * remove username and password, woo fields, application password, etc
 * Make email optional

* Add new UI to add this role to users under other roles

* Dont allow users with this role to delete their profiles from My Account

1. Checkout this branch
2. Add `define( 'NEWSPACK_DISABLE_CAP_GUEST_AUTHORS', true );` to wp-config
3. Visit Users > Guest Authors and see the replacement page
4. Click "Add new Non-Editing Contributor" and confirm you see the regular form to add a new user but with some modifications (no username, no password)
5. Create a user and confirm you can assign it as coauthor
6. Confirm you can not login as this user
7. Confirm the edit page for this user does not have a bunch of fields (username, password, application password, woo fields, profile colors, etc)
8. Edit a user under a subscriber-like role
9. Check the new UI and mark "Allow this user to be assigned as a co-author of a post. " - Save
10. Login as that user
11. Confirm you can't delete your account from the My Account page
12. Confirm you can assign this user as co-author

### Other information:


assets/other-scripts/co-author-plus/index.js
```diff
@@ -0,0 +1,75 @@
+/* globals guestAuthorRole, jQuery */
```
adekbadek: This directory name has a typo – should be `co-authors-plus/index.js` (not the missing `s`).

includes/plugins/class-co-authors-plus.php
```diff
@@ -103,5 +170,228 @@ public static function setup_custom_role_and_capability() {
...
+		 * @return WP_Error
+		 */
+	public static function user_profile_update_errors( $errors, $update, $user ) {
+
+		if ( self::CONTRIBUTOR_NO_EDIT_ROLE_NAME !== $user->role ) {
+			return $errors;
+		}
+
+		if ( ! empty( $errors->errors['empty_email'] ) ) {
+			$errors->remove( 'empty_email' );
```
adekbadek, line 194: Unfortunately I can't recall the details, but I've ran into issues in the past with email-less users. WP expects all users to have emails. Can a dummy email be added instead of ignoring the error?

assets/other-scripts/co-author-plus/index.js
adekbadek: Fixed in https://github.com/Automattic/newspack-plugin/pull/3278

adekbadek (commented):
Added a filepath fix and small tweaks in https://github.com/Automattic/newspack-plugin/pull/3278. 
Also, a question about assigning dummy emails. 
With that question resolved, and the tweaks PR merged in, I'll approve since everything is working well.

includes/plugins/class-co-authors-plus.php
leogermani: Done in https://github.com/Automattic/newspack-plugin/pull/3277/commits/d8848d3b066ebdd984af7afdda7a94397fe9e91b

This will most certainly be needed for Newspack Network, since we use the email as the user id across the network. thanks!

includes/plugins/class-co-authors-plus.php
```diff
@@ -209,6 +209,9 @@ public static function user_profile_update_errors( $errors, $update, $user ) {
 
 			// Create user name from Display name.
 			$user->user_login = self::generate_username( $user->display_name );
+
+			// Create a placeholder email address to avoid any issues with empty emails.
+			$user->user_email = $user->user_login . '@example.com';
```
adekbadek, line 216: The email in not required, but it _can_ be provided. It should be set to the dummy email only if it's not provided. Or simply inside the `! empty( $errors->errors['user_email'] )` conditional above.

adekbadek (changes requested):

includes/plugins/class-co-authors-plus.php
leogermani: yes of course. Also I was not creating the placeholder email on user updates. Fixed now

adekbadek (approved):

PR #3278, opened 2024-07-26
Feat/non editing enhancements tweaks

adekbadek:
Changes to https://github.com/Automattic/newspack-plugin/pull/3277:

- fix JS file path
- move handling of the default value of role selection to the backend, for compatibility with other plugins
- add handling of multi-role UI added by the `capability-manager-enhanced` plugin


leogermani (approved):
Nice! Works well, thanks!

And thanks for the fix on the filename, I didn't notice it because I already had that file on my dist folder I guess

PR #3281, opened 2024-07-26
fix(ras-acc): handle invalid removechild js error

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207836897783913/f

This resolves some js errors that appear whenever transitioning from the auth modal to the checkout modal. It looks like there is some logic that tries to remove a loading state spinner in the auth modal which throws an error since the modal is closed and transitioned to the checkout modal.

This fixes this by adding a check to confirm the spinner is a child of the submitted element before attempting to remove the spinner.

1. As a logged out reader go to any page with a donate block
2. Open dev tools and monitor console errors
3. Go through checkout, logging in via the auth modal first, and stopping at the checkout modal
4. Confirm none of the following errors are logged:

```
DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at n.endLoginFlow
```

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/view.js
+++ b/assets/blocks/reader-registration/view.js
@@ -83,7 +83,9 @@ window.newspackRAS.push( function ( readerActivation ) {
 					messageElement.appendChild( messageNode );
 					messageElement.classList.remove( 'newspack-registration--hidden' );
 				}
-				submitElement.removeChild( spinner );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Noticed this while working on another task—thanks for fixing!

PR #3283, opened 2024-07-27
fix: reCAPTCHA v2 + v3 for RAS-ACC flows

dkoo:
### All Submissions:

This PR (along with https://github.com/Automattic/newspack-blocks/pull/1811 and https://github.com/Automattic/newspack-newsletters/pull/1581) fixes reCAPTCHA v2 and v3 integrations for RAS-ACC flows, which are currently broken in `epic/ras-acc`. It also implements some improvements and standardizations to the reCAPTCHA implementations:

- All v3 token fetching now happens via the same JS methods, instead of requiring different methods for Woo checkout
- The API in `/assets/other-scripts/recaptcha` is a bit easier to follow now
- v2 widgets are rendered more intentionally only when needed, instead of `onload`
- a new `destroy` method (for v3) and `skip-recaptcha` data attributes (for v2) can now be used to skip reCAPTCHA verification on flows where it's not needed, but which share the same form and submit button elements as registration flows which do need verification
- In the admin, v2 and v3 credentials are now stored separately, so switching reCAPTCHA version won't cause you to lose stored credentials

Still not implemented in this PR, but maybe in a future PR: we should figure out a way to validate credentials in the admin based on the selected reCAPTCHA version.

1. On `epic/ras-acc`, connect reCAPTCHA v2 or v3 via **Newspack > Connections**. Observe that sign-in flows trigger unnecessary reCAPTCHA checks, and that not all sign-in, registration, subscribe, or donation flows can be completed due to reCAPTCHA errors.
2. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1811 and https://github.com/Automattic/newspack-newsletters/pull/1581. Visit **Newspack > Connections** and confirm that your credentials are still there. Switch versions (from v2 to v3 or vice versa) and enter a new set of matching credentials, and save. After saving, confirm you can switch back and forth between versions and that the appropriate credentials for both are shown when switching.
3. Test the following flows with v3 enabled (set Threshold to `1.0` and save to ensure that verification fails, then to `0.1` to allow verification to pass):
  - Register via the Registration block (should trigger reCAPTCHA verification)
  - Sign up for newsletters via the Newsletter Subscription Form block (should trigger reCAPTCHA verification)
  - Sign in via the auth modal, using OTP and password (should NOT trigger reCAPTCHA verification)
  - Register for a new account via the auth modal (should trigger reCAPTCHA verification)
  - Complete a transaction via modal checkout
    - Continuing from Billing Details (screen 1) should NOT trigger reCAPTCHA verification
    - Submitting payment details (screen 2) SHOULD trigger verification
4. Switch to v2 and repeat all flows—verification should and shouldn't happen at the same points of all flows.
5. Turn off reCAPTCHA entirely and repeat all flows again—they should all be completable without reCAPTCHA-related errors.

### Other information:

---
  - https://app.asana.com/0/0/1207791057370559

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -30,7 +30,6 @@ final class Recaptcha {
 
 		// Add reCAPTCHA to the Woo checkout form.
 		\add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_recaptcha_v2_to_checkout' ] );
-		\add_action( 'woocommerce_checkout_after_customer_details', [ __CLASS__, 'add_recaptcha_v3_to_checkout' ] );
 
 		// Verify reCAPTCHA on checkout submission.
 		\add_action( 'woocommerce_checkout_process', [ __CLASS__, 'verify_recaptcha_on_checkout' ] );
@@ -111,17 +110,14 @@ final class Recaptcha {
 	 * Register the reCAPTCHA script.
 	 */
 	public static function register_scripts() {
-		// Styles only apply to the visible v2 widgets.
-		if ( self::can_use_captcha( 'v2' ) ) {
+		if ( self::can_use_captcha() ) {
 			\wp_enqueue_style(
 				self::SCRIPT_HANDLE,
 				Newspack::plugin_url() . '/dist/other-scripts/recaptcha.css',
 				[],
 				NEWSPACK_PLUGIN_VERSION
 			);
-		}
 
-		if ( self::can_use_captcha() ) {
 			// Enqueue the reCAPTCHA API from Google's servers.
 			// Note: version arg Must be null to avoid the &ver param being read as part of the reCAPTCHA site key.
 			\wp_register_script(
@@ -428,58 +424,11 @@ final class Recaptcha {
 	}
 
 	/**
-	 * Add reCAPTCHA v3 to Woo checkout.
-	 */
-	public static function add_recaptcha_v3_to_checkout() {
-		if ( ! self::can_use_captcha( 'v3' ) ) {
-			return;
-		}
-		$site_key = self::get_site_key();
-		?>
-		<script src="<?php echo \esc_url( self::get_script_url() ); ?>"></script>
-		<script>
-			grecaptcha.ready( function() {
-				var field;
-				function refreshToken() {
-					grecaptcha.execute(
-						'<?php echo \esc_attr( $site_key ); ?>',
-						{ action: 'checkout' }
-					).then( function( token ) {
-						if ( field ) {
-							field.value = token;
-						}
-					} );
-				}
-				setInterval( refreshToken, 30000 );
-				( function( $ ) {
-					if ( ! $ ) { return; }
-					$( document ).on( 'updated_checkout', refreshToken );
-					$( document.body ).on( 'checkout_error', refreshToken );
-				} )( jQuery );
-				grecaptcha.execute(
-					'<?php echo \esc_attr( $site_key ); ?>',
-					{ action: 'checkout' }
-				).then( function( token ) {
-					field       = document.createElement('input');
-					field.type  = 'hidden';
-					field.name  = 'g-recaptcha-response';
-					field.value = token;
-					var form = document.querySelector('form.checkout');
-					if ( form ) {
-						form.appendChild( field );
-					}
-				} );
-			} );
-		</script>
-		<?php
-	}
-
-	/**
 	 * Verify reCAPTCHA on checkout submission.
 	 */
 	public static function verify_recaptcha_on_checkout() {
 		$url                   = \home_url( \add_query_arg( null, null ) );
-		$should_verify_captcha = apply_filters( 'newspack_recaptcha_verify_captcha', self::can_use_captcha(), $url );
+		$should_verify_captcha = apply_filters( 'newspack_recaptcha_verify_captcha', self::can_use_captcha( 'v3' ), $url );
 		if ( ! $should_verify_captcha ) {
 			return;
 		}
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -27,10 +27,6 @@ final class Recaptcha {
 	public static function init() {
 		\add_action( 'rest_api_init', [ __CLASS__, 'register_api_endpoints' ] );
 		\add_action( 'wp_enqueue_scripts', [ __CLASS__, 'register_scripts' ] );
-		\add_action( 'newspack_newsletters_subscribe_block_before_email_field', [ __CLASS__, 'render_recaptcha_v2_container' ] );
-
-		// Add reCAPTCHA to the Woo checkout form.
-		\add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_recaptcha_v2_to_checkout' ] );
 
 		// Verify reCAPTCHA on checkout submission.
 		\add_action( 'woocommerce_checkout_process', [ __CLASS__, 'verify_recaptcha_on_checkout' ] );
@@ -428,25 +424,6 @@ final class Recaptcha {
 		}
 
 		return true;
-	}
-
-	/**
-	 * Render a container for the reCAPTCHA v2 checkbox widget.
-	 */
-	public static function render_recaptcha_v2_container() {
-		if ( ! self::can_use_captcha( 'v2' ) ) {
-			return;
-		}
-		?>
-		<div id="<?php echo \esc_attr( 'newspack-recaptcha-' . uniqid() ); ?>" class="grecaptcha-container"></div>
-		<?php
-	}
-
-	/**
-	 * Add reCAPTCHA v2 to Woo checkout.
-	 */
-	public static function add_recaptcha_v2_to_checkout() {
-		self::render_recaptcha_v2_container();
 	}
 
 	/**
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -7,9 +7,6 @@
 
 namespace Newspack;
 
-use Error;
-use Google\Service\Docs\PinTableHeaderRowsRequest;
-
 defined( 'ABSPATH' ) || exit;
 
 /**
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -121,8 +121,6 @@ final class Recaptcha {
 				null, // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
 				false
 			);
-			\wp_script_add_data( self::SCRIPT_HANDLE_API, 'async', true );
-			\wp_script_add_data( self::SCRIPT_HANDLE_API, 'defer', true );
 
 			\wp_enqueue_script(
 				self::SCRIPT_HANDLE,
@@ -131,8 +129,6 @@ final class Recaptcha {
 				NEWSPACK_PLUGIN_VERSION,
 				true
 			);
-			\wp_script_add_data( self::SCRIPT_HANDLE, 'async', true );
-			\wp_script_add_data( self::SCRIPT_HANDLE, 'defer', true );
 
 			\wp_localize_script(
 				self::SCRIPT_HANDLE,
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -232,7 +232,7 @@ final class Recaptcha {
 				)
 			) {
 			$legacy_key      = \get_option( self::OPTIONS_PREFIX . 'site_key', false );
-			$legacy_secret   = \get_option( self::OPTIONS_PREFIX . 'site_key', false );
+			$legacy_secret   = \get_option( self::OPTIONS_PREFIX . 'site_secret', false );
 
 			if ( ! empty( $legacy_key ) ) {
 				$settings['credentials'][ $current_version ]['site_key'] = $legacy_key;
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -8,6 +8,7 @@
 namespace Newspack;
 
 use Error;
+use Google\Service\Docs\PinTableHeaderRowsRequest;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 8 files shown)


chickenn00dle (changes requested):
Thanks for getting this fix up @dkoo!

I went through the testing steps and ran into some issues. Saving and switching between recaptcha versions in admin seems to work perfectly! However, going through the flows is a bit inconsistent:

1. On v3 with threshold at `0.1`, the reader registration and newsletter blocks, as well as modal checkout work as expected, however the auth form registration flow still returns a recaptcha error. I'm getting the following console errors and the response returns the following message:
Screenshot 2024-07-29 at 14 08 09

```
Could not complete this request. Please try again later.
```

2. On v2, the newsletters form triggers the following alert on submit:
Screenshot 2024-07-29 at 14 10 56

Also, the remaining flows succeed with no problems, but it's unclear to me how to confirm recaptcha has triggered or not.

3. With recaptcha disabled, I am seeing the following error wherever the RAS-ACC signin button appears:

```
 Uncaught ReferenceError: newspack_grecaptcha is not defined
    setFormAction https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    9073 https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    9073 https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    i https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    9073 https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    K https://site.test/wp-content/plugins/newspack-plugin/dist/reader-activation.js?ver=4.6.0:1
    K https://site.test/wp-content/plugins/newspack-plugin/dist/reader-activation.js?ver=4.6.0:1
    9073 https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    a https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    o https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    O https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    <anonymous> https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
    <anonymous> https://site.test/wp-content/plugins/newspack-plugin/dist/reader-auth.js?ver=4.6.0:1
```

This looks to be preventing the auth and checkout modal scripts from running and so checkout/donation buttons and the sign in button don't work anymore.

I haven't looked through the code yet since I'm running into these issues. I'll have a look through once we can sort out these problems. Let me know if I'm missing something, or if you think this review would benefit from a call.

Thanks!

dkoo:
@chickenn00dle whoops, thanks for catching those issues. 8c42f3f should fix those, but you'll want to retest step 2 by checking out `epic/ras-acc` again, deleting the `newspack_recaptcha_credentials` option, resaving credentials in the Connections wizard, then checking out this branch once more. There was a typo in the original code that was migrating the site key to both key and secret fields.

> it's unclear to me how to confirm recaptcha has triggered or not.

For v3, you can verify that reCAPTCHA has been triggered and is working correctly by error logging [this object](https://github.com/Automattic/newspack-plugin/blob/fix/recaptcha-for-ras-acc/includes/class-recaptcha.php#L399). It should have a `score` param if it worked correctly.

For v2, you can force it to trigger a challenge on every request if you visit the [settings page](https://www.google.com/recaptcha/admin/) of your v2 account and move the "Security Preference" slider all the way to the right:

Screenshot 2024-07-29 at 4 27 37 PM

assets/other-scripts/recaptcha/index.js
```diff
@@ -34,89 +24,180 @@ function domReady( callback ) {
...
 }
 
+window.newspack_grecaptcha = window.newspack_grecaptcha || {
+	destroy,
+	render,
+	version: newspack_recaptcha_data.version,
+};
+
+const isV2 = 'v2' === newspack_recaptcha_data.version.substring( 0, 2 );
+const isV3 = 'v3' === newspack_recaptcha_data.version;
```
chickenn00dle: Nonblocking nitpick: I guess we can save the v2 calculation by assuming if `isV3` is false, `isV2` must be true? Not sure if `newspack_recaptcha_data.version` can be anything else though.

assets/other-scripts/recaptcha/index.js
```diff
@@ -34,89 +24,180 @@ function domReady( callback ) {
...
+const siteKey = newspack_recaptcha_data.site_key;
+const isInvisible = 'v2_invisible' === newspack_recaptcha_data.version;
```
chickenn00dle: Nonblocking nitpick: Similar to the last comment, are we supporting any other v2 version? If not, can we just assume if `isV2` is true, `isInvisible` is also true?

assets/other-scripts/recaptcha/index.js
```diff
@@ -34,89 +24,180 @@ function domReady( callback ) {
...
-			grecaptcha
-				.execute( siteKey, { action: 'submit' } )
-				.then( token => res( token ) )
-				.catch( e => rej( e ) );
+		// Callback when reCAPTCHA passes validation.
+		const callback = () => {
+			form.requestSubmit( button );
+			refreshWidget( button );
+		};
+		const widgetId = grecaptcha.render( button, {
```
chickenn00dle: Might be a good idea to re-focus the form for the reader to ensure recaptcha is visible on longer forms like modal checkout.

assets/reader-activation-auth/auth-form.js
```diff
@@ -162,40 +170,28 @@ window.newspackRAS.push( function ( readerActivation ) {
...
+							body,
+						} )
+							.then( () => {
+								form.setMessageContent( newspack_reader_activation_labels.code_resent );
+								container.setFormAction( 'otp' );
+								readerActivation.setOTPTimer();
 							} )
 							.catch( e => {
-								console.log( { e } ); // eslint-disable-line no-console
+								console.log( e ); // eslint-disable-line no-console
```
chickenn00dle: Might be a good time to remove this `console.log`. Is it necessary? If the purpose is for debugging, I think we can always re-add this on our end temporarily while working through any issues.

assets/reader-activation/index.js
```diff
@@ -383,23 +383,28 @@ function attachAuthCookiesListener() {
...
  */
 function attachNewsletterFormListener() {
-	const forms = document.querySelectorAll(
-		'.newspack-newsletters-subscribe,.newspack-subscribe-form,.mc4wp-form'
-	);
-	if ( ! forms.length ) {
+	const newspackForms = [ '.newspack-newsletters-subscribe', '.newspack-subscribe-form' ];
+	const thirdPartyForms = [ '.mc4wp-form' ];
+
+	if ( ! newspackForms.length && ! thirdPartyForms.length ) {
```
chickenn00dle: Isn't this always going to return false? It looks like we are creating two arrays here with hard-coded entries, and then checking for lengths.

chickenn00dle (changes requested):
Thanks for addressing my feedback so quickly @dkoo 🙌 
Most of the issues have been resolved 🎉 

I did run into a few problems though:

> Register via the Registration block (should trigger reCAPTCHA verification)

I'm not sure if this is expected, but with v2 (slider at minimum), only this still triggers recaptcha. I expected none to trigger whenever the slider is at its lowest setting.

> Register for a new account via the auth modal (should trigger reCAPTCHA verification)

With V2 enabled, and the slider maxed out to the right, the registration flow via the auth modal does not trigger recaptcha

> Submitting payment details (screen 2) SHOULD trigger verification

With V2, recaptcha does trigger, however readers have to scroll to the bottom of the form to submit, and there is nothing to let the user know they need to scroll back up to complete recaptcha. Maybe we can focus back to the top of the form for the reader?

Also, when I attempt to checkout with V2 (slider at minimum) with a logged out reader, and go through the auth+checkout flow, I get the following alert

Screenshot 2024-07-30 at 15 06 49

assets/other-scripts/recaptcha/index.js
dkoo: We originally did build in support for the non-invisible (checkbox) flavor of v2, which you might have seen on sites where you have to click a button that says "I am not a robot" to trigger the verification. So that's why all of the logic is written as if there are more than two options. It feels like it might be more futureproof to expect that there could be more than two options, too, because maybe Google will release a v4 at some point?

assets/reader-activation-auth/auth-form.js
dkoo: 39c4e67e removes the `console.log` and implements error handling in case the request to send an OTP email fails. Currently the method on the backend [always returns JSON](https://github.com/Automattic/newspack-plugin/blob/fix/recaptcha-for-ras-acc/includes/reader-activation/class-reader-activation.php#L1526) so that's why the error isn't caught by the `catch` block.

assets/other-scripts/recaptcha/index.js
dkoo: [Same response as above](https://github.com/Automattic/newspack-plugin/pull/3283#discussion_r1697562186)!

assets/reader-activation/index.js
dkoo: You're right—this is leftover from before I did the hardcoded arrays thing. Updated in 5ec8bba

assets/other-scripts/recaptcha/index.js
dkoo: It seems like the reCAPTCHA widget styles aren't playing nice with some styles in our modal checkout iframe. The widget is supposed to be fixed to the top of the viewport, so it shouldn't ever be rendering outside of the visible scroll. I spent some time trying to figure out why, but couldn't easily, so 62c60e87 adds a scroll to the top when the widget is rendered, but only in the modal checkout. I also don't know why `window.scroll` wasn't working here, hence the use of `document.body.scrollIntoView`.

dkoo:
@chickenn00dle:

> I'm not sure if this is expected, but with v2 (slider at minimum), only this still triggers recaptcha. I expected none to trigger whenever the slider is at its lowest setting.

I'm also seeing the challenge getting triggered even with the lowest slider setting. From what I've read the API itself decides whether to show the challenge, so maybe because we're doing so much non-human-like testing it's causing us to be challenged more.

> With V2 enabled, and the slider maxed out to the right, the registration flow via the auth modal does not trigger recaptcha

This was a regression that should be fixed in 39c4e67.

> With V2, recaptcha does trigger, however readers have to scroll to the bottom of the form to submit, and there is nothing to let the user know they need to scroll back up to complete recaptcha. Maybe we can focus back to the top of the form for the reader?

Agreed, however the solution turns out to be trickier than expected (maybe due to the way the checkout "locking" prevents scrolling?). [See response here](https://github.com/Automattic/newspack-plugin/pull/3283#discussion_r1697715831)

> Also, when I attempt to checkout with V2 (slider at minimum) with a logged out reader, and go through the auth+checkout flow, I get the following alert

I couldn't replicate this, but it sounds like an issue with the reCAPTCHA API not loading correctly. Can you replicate it reliably by taking specific steps?

chickenn00dle:
Nice @dkoo!

Pretty much all of the issues have been resolved. The only problem remaining is the sporadic recaptcha error alert. 

> I couldn't replicate this, but it sounds like an issue with the reCAPTCHA API not loading correctly. Can you replicate it reliably by taking specific steps?

Unfortunately this seems to happen pretty sporadically and it seems like its happening across all of the forms randomly for me. Whenever the alert appears, the response from `POST https://www.google.com/recaptcha/api2/userverify` looks something like this:

Screenshot 2024-07-31 at 11 42 41

I am using the following email address format when submitting the forms: `raz+subscriberX@automattic.com` where `X` is some number, and this only happens when I'm using v2 with the slider set to minimum.

I wonder if this has something to do with the json response from Google. Are we perhaps stripping certain characters which might be breaking some responses and not others?

dkoo:
@chickenn00dle 9c405f3 resolves the connection error by adding the `isolated` param on the `render` call, as we found that this resolved this issue. 

When testing, I also found that there was a race condition in the modal checkout flow. If you initiate a modal checkout by logging into an existing account that has prior orders and billing info, the modal checkout skips the first screen and ends up calling `newspack_grecaptcha.render`, but this could happen before the reCAPTCHA API is ready to render anything. 9c405f3 [fixes this](https://github.com/Automattic/newspack-plugin/pull/3283/commits/9c405f354e7c18f149805033658bb62e9cb8e26a#diff-fdb5afeef588626d675f1dc11e34a8939619289ac9529452e6743b88a24e884fR180-R183) by checking for the existence of the global `grecaptcha` object in our own `render` function and passing itself to the `grecaptcha.ready` callback if necessary.

chickenn00dle (changes requested):
Good catch on the race condition @dkoo!

I'm unfortunately seeing this issue again on my end, even with the latest changes checked out. For example, here is a screen recording of the auth+checkout flow on FF triggering the recaptcha multiple times to see what happens:

https://github.com/user-attachments/assets/48c81861-276f-490c-a012-8c372c2f3365

I also tried using a publicly accessible URL via ngrok but still ran into the same issue:

Screenshot 2024-08-01 at 10 09 12

Here is one other note I'm seeing while testing this:

There seems to be many duplicate google scripts loaded when I view sources in the browser console:

Screenshot 2024-08-01 at 10 31 15

I don't think this would be causing the problem, but just noting as it seems our enqueue method is running multiple times.

chickenn00dle (approved):

chickenn00dle (approved):

chickenn00dle:
Hey @dkoo, I'm still working on trying to get the sporadic could not connect error resolved. That said, I noticed with current changes there is no longer an issue when recaptcha is not enabled. WDYT about merging these into epic/ras-acc now that that's working and then I can tackle that error in a separate PR?

PR #3293, opened 2024-07-29
fix(ras-acc): fix order of elements in registration block

laurelfulford:
### All Submissions:

This PR moves the location of the Sign In with Google button in the Registration block, to fix a bug that came from changing the order for RAS-ACC, and to make it more consistent with the built-in RAS-ACC sign-in form. 

See 1205234045751551-as-1207882141458899.

1. Start on a test site running epic/ras-acc branch, with Google connected via Newspack Manager so the Sign In with Google button appears.
2. In an incognito window, click around until the campaign-based RAS sign up modal appears; this modal uses the Registration Block. Note the order:

image

3. Apply this PR.
4. Confirm that the order is fixed, and the spacing looks more normal:

image

5. Confirm that the order is also correct in the editor. (The styles there overall aren't 100% 1:1, but the block is also being reworked [here](https://github.com/Automattic/newspack-plugin/pull/3199), so I just focused on the Google button order):

image

### Other information:

### DIFF
--- a/assets/blocks/reader-registration/edit.js
+++ b/assets/blocks/reader-registration/edit.js
@@ -345,6 +345,19 @@ export default function ReaderRegistrationEdit( {
 									</div>
 								) : null }
 								<div className="newspack-registration__main">
+									{ newspack_blocks.has_google_oauth && (
+										<div className="newspack-reader__logins">
+											<button className="newspack-reader__logins__google">
+												<span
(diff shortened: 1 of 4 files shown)


laurelfulford:
Throwing back into draft mode so I can fix the editor preview, too 😅 

Edited to add: done!

chickenn00dle (approved):
Looks good!

laurelfulford:
Thanks @chickenn00dle! 🎉

PR #3294, opened 2024-07-30
fix(ras-acc): handle content gate auth checkout flow

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207836897783913/f

This PR addresses an issue where the registration/login+checkout flow via content gate is cutoff after the auth step. This happens because the content gate is watching for the reader to be authenticated before reloading the page. However, in the case of auth+checkout, we don't want to reload until after checkout has been completed.

We fix this by adding a cookie flag to indicate auth checkout status, and only reload when that flag is no longer set (after checkout).

This PR addresses another smaller issue where the newsletters modal should not triggered by the auth modal when this checkout flag is set since the checkout flow will manage which modals should appear next.

Screenshot 2024-07-30 at 12 49 45

**Important**: You will also need to checkout the changes in https://github.com/Automattic/newspack-blocks/pull/1817 to test.

1. Set up a membership plan that restricts content to users who have made a recurring donation
2. Set up a content gate for this membership plan via Newspack > Engagement > Show Advanced Settings
3. Ensure the `Present newsletter signup after checkout and registration` toggle is active via this same menu
4. As a logged out reader, visit a restricted page and make a donation to gain access to the restricted content
5. Confirm you are able to sign in as well as complete checkout and the page does not reload until AFTER all modals have closed
6. Confirm the newsletters modal is the final modal that appeared in the flow
7. Confirm you have gained access to the content once the page has reloaded
8. Smoke test the normal donation checkout flow as a signed out and signed in user (Not the content gate donation block)
9. Smoke test the normal signin and registration flow via the RAS-ACC sign in button at the top right of the site navigation

### Other information:

### DIFF
--- a/assets/reader-activation-auth/auth-form.js
+++ b/assets/reader-activation-auth/auth-form.js
@@ -236,7 +236,11 @@ window.newspackRAS.push( function ( readerActivation ) {
 					}
 
 					let callback;
-					if ( container.authCallback && data?.registered ) {
+					if (
+						container.authCallback &&
+						data?.registered &&
+						! readerActivation.getCheckoutStatus()
+					) {
 						callback = ( authMessage, authData ) =>
 							openNewslettersSignupModal( {
 								callback: container.authCallback( authMessage, authData ),
--- a/assets/reader-activation/index.js
+++ b/assets/reader-activation/index.js
@@ -326,6 +326,26 @@ export function getAuthStrategy() {
 	}
 	return getCookie( 'np_auth_strategy' );
 }
+/**
+ * Set the reader checkout status.
+ *
+ * @param {boolean} status Checkout status. Default is false.
+ *
+ * @return {void}
+ */
+export function setCheckoutStatus( status = false ) {
+	setCookie( 'np_auth_checkout_status', status );
+	emit( EVENTS.reader, getReader() );
+	return status;
+}
+/**
+ * Get the reader checkout status.
+ *
+ * @return {boolean} Reader checkout status.
+ */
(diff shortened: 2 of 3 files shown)


laurelfulford (changes requested):
This PR looks like it fixes both the issue with the content gate in most case, and with the newsletter form showing up too soon.

I only ran into one weird case where the page refreshed too soon:

1. Start as a reader who's already logged in, but does not have any donation products.
2. Navigate to a single post to see the content gate.
3. Make a donation from the single post (I made my content gate a Donation Block).
4. Run through the checkout.
5. At the end, you see the Newsletter form briefly, but closing the 'Thank you' message triggers the refresh, so you don't have a chance to use it.

This only happens when you're already logged in when you start the transaction -- if you sign up as part of the transaction, or even log into an existing account as part of the checkout process, it doesn't happen.

Just let me know if you have any questions about this!

chickenn00dle:
Thanks for the review @laurelfulford!

> This only happens when you're already logged in when you start the transaction

Good catch! We were previously only setting the checkout flag for logged out users. I've updated this in the blocks PR to add the flag for logged in users as well: https://github.com/Automattic/newspack-blocks/pull/1817/commits/9b9b0f1e246b5647e5a1475e5d52607ed60d4934

Let me know if you still see an issue on your end!

laurelfulford (approved):
That looks like it did it! I'm no longer seeing the issue, and the other flows (donating as a new user, buying another product, signing in, signing up) seem to work as expected! 🙌

PR #3303, opened 2024-08-02
fix: recaptcha tweaks for reliability & performance

dkoo:
### All Submissions:

#3283 mostly works but has some sporadic/unpredictable issues with connectivity to the reCAPTCHA API, especially with the v2 implementation. This PR (along with https://github.com/Automattic/newspack-blocks/pull/1824) attempts to simplify the v2 implementation to avoid those issues. Mainly, instead of rendering and handling separate reCAPTCHA v2 widgets for every form that requires verification, this PR renders just one per HTML document and shares it across all forms that need verification. Initial tests indicate that this is more reliable and performant due to needing fewer roundtrip requests to the Google API.

1. On `fix/recaptcha-for-ras-acc`, connect both a v2 and a v3 reCAPTCHA account on your site. Run through some form flows using both implementations, switching back and forth between v2 and v3 between tests.
2. Eventually, you might encounter a mysterious error like this when using v2, especially in Firefox, or in any browser while blocking third-party cookies:

Screenshot 2024-08-01 at 4 36 30 PM

3. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1824 and clear your browser cache and cookies, and maybe wait a little while before testing again (in case there's any caching on the Google API side as well that we don't have control over).
4. Try testing flows with both v2 and v3, switching back and forth a few times as well, and confirm that the "Cannot contact reCAPTCHA" error doesn't occur at all, or at least doesn't occur as often as before.
5. Also test all flows without reCAPTCHA enabled to make sure there are no errors in those flows.

### Other information:


chickenn00dle (changes requested):
Tested on `ras-acc-dev.newspackstaging.com` and v2 looks to be working now 🙌 ! I'm also seeing we've fixed the issue where checkout was breaking when recaptcha was not active 🎉 

Looks like v3 isn't working on the staging site though:

Screenshot 2024-08-06 at 13 52 51

Last week, I know there was an issue with the v2 credentials for the test site. Can you confirm the v3 creds are correct as well?

PR #3305, opened 2024-08-02
fix(ras-acc): reset otp flow on sign in modal close

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207922674964422/f

This PR "resets" the OTP state of the login form once the OTP timer runs out. Previously, if a reader opened the login form with an account with a password set, but then chose to sign in via OTP, the auth form would "remember" this OTP login state and prompt for OTP the next time the auth modal was opened.

This PR fixes this by resetting OTP cookies whenever the auth form is opened AND the OTP timer has run out and setting the auth form the the default signin state.

Screenshot 2024-08-06 at 12 30 04

This PR also fixes a problem where the OTP code message always mentioned "resent" even when the code was sent for the first time:

1. Set up a reader account that has a password set, and a reader account that does not
2. As the reader account with a password:
  - Trigger the signin flow via the sign-in button
  - Enter the relevant email address and select continue
  - Choose to sign in via OTP via the OTP button
  - Confirm the auth modal transitions to the OTP view, the message says "code sent!...", and an OTP email was actually sent. DON'T login.
  - Wait 60 seconds for the resend code button timer to finish running down, then click the button
  - Confirm the auth modal message updates to say "code resent!..." and another OTP email was sent.
  - Confirm you are able to login with the OTP code
3. As the reader account WITHOUT a password:
  - Repeat the above steps. The only difference should be the auth form opens into the OTP screen by default after entering the relevant email address and selecting continue.
4. Smoke test the rest of the login form to confirm nothing seems broken

### Other information:

### DIFF
--- a/includes/class-magic-link.php
+++ b/includes/class-magic-link.php
@@ -375,6 +375,38 @@ final class Magic_Link {
 	}
 
 	/**
+	 * Check for active magic link tokens.
+	 *
+	 * @param \WP_User $user User to check the active magic link token for.
+	 *
+	 * @return bool|\WP_Error
+	 */
+	public static function has_active_token( $user ) {
+		if ( ! self::can_magic_link( $user->ID ) ) {
+			return new \WP_Error( 'newspack_magic_link_invalid_user', __( 'Invalid user.', 'newspack' ) );
+		}
+
+		$now    = time();
+		$tokens = \get_user_meta( $user->ID, self::TOKENS_META, true );
+
+		$expire = $now - MINUTE_IN_SECONDS;
+		if ( ! empty( $tokens ) ) {
+			foreach ( $tokens as $index => $token_data ) {
+				/** Clear expired tokens. */
+				if ( $token_data['time'] < $expire ) {
+					unset( $tokens[ $index ] );
+				}
+			}
+		}
+
+		if ( empty( $tokens ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	 * Generate a magic link.
 	 *
 	 * @param \WP_User $user User to generate the magic link for.
--- a/includes/class-magic-link.php
+++ b/includes/class-magic-link.php
@@ -375,7 +375,7 @@ final class Magic_Link {
 	}
 
 	/**
-	 * Check for active magic link tokens.
+	 * Check for magic link tokens generated within the last 60 seconds.
 	 *
 	 * @param \WP_User $user User to check the active magic link token for.
 	 *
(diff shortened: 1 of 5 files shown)


laurelfulford (approved):
I think this looks good! I hammered it a bit, so I ran into a couple other odd things,but I don't think either are big enough to block it -- they're pretty edge-case-y, and we can tackle them separately. They are:

**Weird button state:**

1. Try logging in with an existing account with a password.
2. Opt to get a code sent to you.
3. During the countdown click 'Back'.
4. Click 'Continue'; when you get back, the "Email me a one-time code" button is greyed out, but is displaying the default text instead of the timer.

This persists even after you close the modal even after waiting a minute; you have to refresh the page to get the button back. You can still do the other actions, though (entering the password, or resetting it).

CleanShot 2024-08-09 at 16 23 00

**Getting "Stuck" on the first screen:**

1. Try logging in with an existing account without a password.
2. From the OTP screen, click "Back".
3. Click "Continue" again.
4. You get an error message to wait before requesting another code. If you wait, it DOES work again without refreshing/reopening if you wait for 60 seconds, there's just no indication how long the wait is, and you can't get back to the code page to enter your existing code. 

CleanShot-2024-08-09-at-16 31 49

Also as an observation: when you log in with a email without a password and hit the OTP screen, it doesn't show the same "Code sent!" message. Not sure if it's really needed there, though!

Just let me know if you have questions about either of these, and if I should spin up something in Asana!

chickenn00dle:
Thanks for digging into this @laurelfulford!

> I hammered it a bit, so I ran into a couple other odd things,but I don't think either are big enough to block it -- they're pretty edge-case-y, and we can tackle them separately. They are:

I think I'll have a look at these cases and address them in this PR if that works for you. I'll re-request review once I've done that!

laurelfulford:
That works too! 🙂 I'll keep an eye out for updates!

chickenn00dle:
I think this is now ready for review again @laurelfulford!

It turns out accounting for these edge cases calls for some pretty big changes to how the form behaves, so we will want to make sure this is tested pretty thoroughly. In particular, we want to make sure we test things like:

- Switching email addresses after going back in the auth form
- Allowing the resend code timer to expire and resending a code
- Confirming emails are only sent when the form is initially opened for an email address as well as when the resend code button is selected. For cases where a reader selects back and forth with no change to the email input, no email should be sent

Let me know if you have any questions!

laurelfulford (approved):
Thanks for taking a swing at those issues, @chickenn00dle!

I've been hammering the heck out of this, and I haven't managed to trip it up yet. I retested the original steps, plus:

* Clicking back and continue and monitoring for unneeded emails.
* Clicking back and entering a new email address.
* Resending the code after the timer runs out.

I also tried to recreate the two issues I originally reported, and they both look good 🙂

PR #3327, opened 2024-08-09
fix(data-events): gate interaction for registration form

miguelpeixe:
### All Submissions:

1200550061930446-as-1207615072285532

#2740 introduced tracking capabilities for forms inside the gate, which covers the Reader Registration Block and Checkout Button Block.

There's the unhandled case of registrations initiated via a hash link on the gate, which triggers the RAS' global auth form:

image

To handle those cases, this PR proposes that the tracking input be added to RAS' auth form when visiting gated content.

This means registrations made on gated content, but initiated outside of the gate (like the header "Sign In" button), will also be treated as a gate interaction. To improve accuracy, the tracking input should now only be added once the gate is **seen** on the page. With that strategy, once the page loads the reader may register through the header "Sign In" and never realize that they are on gated content, which will not be a gate interaction. Once they scroll down and hit the gate, they may scroll back up, register from "Sign In", and it will be a gate interaction.

1. Check out this branch and make sure you have RAS, Woo Memberships and a configured content gate
2. Edit your content gate to use the "Paywall with One Tier and Metering" pattern (Patterns -> Newspack Memberships)
3. In a fresh session, visit a gated article and, without scrolling down, register through the header "Sign In"
4. Confirm the registration goes through with the `registration_method` metadata as `auth-form`
5. Reset your session and visit a gated article
6. Scroll down to the gate and register through the "Register for free" link
7. Confirm the registration goes through with `registration_method` metadata as `registration-block-content-gate`
8. Also confirm the GA events goes through with the following:
```json
{"name":"np_gate_interaction","action":"form_submission_received","action_type":"registration"}
```
```json
{"name":"np_gate_interaction","action":"form_submission_success","action_type":"registration"}
```

### Other information:

### DIFF
--- a/includes/data-events/class-memberships.php
+++ b/includes/data-events/class-memberships.php
@@ -41,7 +41,7 @@ final class Memberships {
 		add_action( 'init', [ __CLASS__, 'register_listeners' ] );
 		add_filter( 'newspack_blocks_modal_checkout_cart_item_data', [ __CLASS__, 'checkout_cart_item_data' ], 10, 2 );
 		add_action( 'woocommerce_checkout_create_order_line_item', [ __CLASS__, 'checkout_create_order_line_item' ], 10, 4 );
-		add_filter( 'newspack_register_reader_form_metadata', [ __CLASS__, 'register_reader_metadata' ], 10, 2 );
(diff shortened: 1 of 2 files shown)


leogermani (approved):
Works well

PR #3338, opened 2024-08-14
fix(woocommerce-memberships): membership status check

adekbadek:
### All Submissions:

Fixes an issue introduced in #3268. Memberships were switched to active status too eagerly, causing free-trial memberships to become active under certain conditions. 

1. On `release`, update an active membership (with an active subscription attached) to `wcm-free_trial` status 
2. Load the Memberships view (`/wp-admin/edit.php?post_type=wc_user_membership`), reload
3. Observe the membership's status was updated to `wcm-active`
4. Switch to this PR and update the status again to `wcm-free_trial`
5. Repeat step 2 – observe the status was not changed

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -934,7 +934,7 @@ class Memberships {
 	 * @return string
 	 */
 	public static function check_membership_status( $post_status, $post ) {
-		if ( 'wc_user_membership' !== $post->post_type || 'wcm-cancelled' !== $post->post_status || ! self::is_active() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
+		if ( 'wc_user_membership' !== $post->post_type || ! in_array( $post->post_type, [ 'wcm-cancelled', 'wcm-expired', 'wcm-paused' ], true ) || ! self::is_active() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
 			return $post_status;
 		}
 		$integrations = wc_memberships()->get_integrations_instance();
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -925,7 +925,8 @@ class Memberships {
 
 	/**
 	 * Check if a user has an active subscription with the required products when checking membership status.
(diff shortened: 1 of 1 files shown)


miguelpeixe:
I'm about to review this now, could you change the base to `release`?

miguelpeixe (approved):
I couldn't find the "Free trial" option but switching to "Complimentary" had the same effect.

Works as described.

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -925,15 +925,16 @@ private static function user_has_content_access_from_rules( $user_id, array $rul
...
+	 * reset inactive memberships to active link to the active subscription.
 	 *
 	 * @param string  $post_status Post status.
 	 * @param WP_Post $post Post object.
 	 *
 	 * @return string
 	 */
 	public static function check_membership_status( $post_status, $post ) {
-		if ( 'wc_user_membership' !== $post->post_type || 'wcm-active' === $post->post_status || ! self::is_active() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
+		if ( 'wc_user_membership' !== $post->post_type || 'wcm-cancelled' !== $post->post_status || ! self::is_active() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
```
leogermani: There are other statuses that can be considered "inactive". I think to keep the original intention here this should read:

```suggestion
		if ( 'wc_user_membership' !== $post->post_type || ! in_array( $post->post_type, [ 'wcm-cancelled', 'wcm-expired', 'wcm-paused' ], true ) || ! self::is_active() || ! function_exists( 'wc_memberships_get_user_membership' ) ) {
```

leogermani: overall this is a very heavy routine to be executed in every `get_post_status` call... we should work on an alternative after we fix this...

miguelpeixe: Do we know if any status other than `cancelled` (or other) is automatically set for memberships that got deactivated due to a subscription that failed to renew?

All status could be manually set (including cancelled), and would still get overwritten by this hook.

> overall this is a very heavy routine to be executed in every get_post_status call... we should work on an alternative after we fix this...

It bails early on `'wc_user_membership' !== $post->post_type`, I'm not sure it has a significant impact.

PR #3343, opened 2024-08-15
feat(ras-acc): handle checkout redirect params

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207817176293828/f

This PR updates the magic link redirect url when auth is triggered during the checkout process. We do this so the blocks plugin can handle the magic link by opening into modal checkout.

Make sure to also checkout the changes in this Blocks PR before testing: https://github.com/Automattic/newspack-blocks/pull/1841

1. On one or more test page or post, set up the following:
  - A donate block
  - A checkout button block for a simple product
  - A checkout button block for a variable product, allowing the reader to choose variation
  - A checkout button block for a variable product, with a preset variation
  - A checkout button block for a simple subscription
  - A checkout button block for a variable subscription
Screenshot 2024-08-19 at 16 45 15
2. As a logged out reader, visit the page and for each do the following:
  - Select the block to trigger modal checkout
  - When the auth modal appears, if otp isn't default, choose to receive an otp code via email
  - Check the email, but do not copy and use the OTP code. Instead click the magic link (or button) in the email
  - Confirm a new browser tab opens up to the test page and the checkout modal automatically opens to the same product choice:
Screenshot 2024-08-19 at 16 48 00
  - Either complete checkout or close the modal
  - Go to my account and log out
3. As admin, set the donation block layout to the other layout that was not yet tested (either frequency or tiered) and again, repeat the above test steps.
4. Finally, via Newspack > Reader Revenue > Donations, set the main donations page donation block to untiered
5. Visit the donations page as a logged out reader, and again repeat the test steps above

BONUS: Smoke test both modal checkout while logged in and the standalone auth form while logged out to make sure there are no regressions.

### Other information:

### DIFF
--- a/src/reader-activation-auth/auth-form.js
+++ b/src/reader-activation-auth/auth-form.js
@@ -98,9 +98,9 @@ window.newspackRAS.push( function ( readerActivation ) {
 			container.setFormAction( 'signin' );
 
 			/**
-			 * Get the redirect URL.
+			 * Get the checkout redirect URL.
 			 */
-			const getRedirectUrl = () => {
+			const getCheckoutRedirectUrl = () => {
 				const checkoutType = readerActivation.getCheckoutData( 'type' );
 				if ( ! checkoutType ) {
 					return;
@@ -183,8 +183,8 @@ window.newspackRAS.push( function ( readerActivation ) {
 						body.set( 'reader-activation-auth-form', 1 );
 						body.set( 'npe', emailInput.value );
 						body.set( 'action', 'link' );
-						if ( readerActivation.getCheckoutStatus() ) {
-							const redirectUrl = getRedirectUrl();
+						if ( readerActivation.isPendingCheckout() ) {
+							const redirectUrl = getCheckoutRedirectUrl();
 							if ( redirectUrl ) {
 								body.set( 'redirect_url', redirectUrl );
 							}
@@ -272,7 +272,7 @@ window.newspackRAS.push( function ( readerActivation ) {
 					if (
 						container.authCallback &&
 						data?.registered &&
-						! readerActivation.getCheckoutStatus()
+						! readerActivation.isPendingCheckout()
 					) {
 						callback = ( authMessage, authData ) =>
 							openNewslettersSignupModal( {
@@ -414,8 +414,8 @@ window.newspackRAS.push( function ( readerActivation ) {
 						if ( ! body.has( 'npe' ) || ! body.get( 'npe' ) ) {
 							return form.endLoginFlow( newspack_reader_activation_labels.invalid_email, 400 );
 						}
-						if ( readerActivation.getCheckoutStatus() ) {
-							const redirectUrl = getRedirectUrl();
+						if ( readerActivation.isPendingCheckout() ) {
+							const redirectUrl = getCheckoutRedirectUrl();
 							if ( redirectUrl ) {
 								body.set( 'redirect_url', redirectUrl );
 							}
--- a/src/reader-activation-auth/auth-form.js
+++ b/src/reader-activation-auth/auth-form.js
@@ -98,6 +98,32 @@ window.newspackRAS.push( function ( readerActivation ) {
 			container.setFormAction( 'signin' );
 
 			/**
+			 * Get the redirect URL.
+			 */
+			const getRedirectUrl = () => {
+				const checkoutType = readerActivation.getCheckoutData( 'type' );
+				if ( ! checkoutType ) {
+					return;
+				}
+				const redirectUrl = new URL( window.location.href );
+				redirectUrl.searchParams.set( 'newspack_modal_checkout', 1 );
+				redirectUrl.searchParams.set( 'type', checkoutType );
+				// Add checkout button params.
+				if ( checkoutType === 'checkout_button' ) {
+					redirectUrl.searchParams.set( 'product_id', readerActivation.getCheckoutData( 'product_id' ) ?? '' );
+					redirectUrl.searchParams.set( 'variation_id', readerActivation.getCheckoutData( 'variation_id' ) ?? '' );
+				}
+				// Add donate params.
+				if ( checkoutType === 'donate' ) {
+					redirectUrl.searchParams.set( 'layout', readerActivation.getCheckoutData( 'layout' ) ?? '' );
+					redirectUrl.searchParams.set( 'frequency', readerActivation.getCheckoutData( 'frequency' ?? '' ) );
+					redirectUrl.searchParams.set( 'amount', readerActivation.getCheckoutData( 'amount' ) ?? '' );
+					redirectUrl.searchParams.set( 'other', readerActivation.getCheckoutData( 'other' ) ?? '' );
+				}
+				return redirectUrl.href;
+			}
+
+			/**
 			 * Handle reader changes.
 			 */
 			const handleReaderChanges = () => {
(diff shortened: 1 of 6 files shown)


laurelfulford:
@chickenn00dle when testing this, I'm running into the same issue with Chrome and Firefox, and wanted to double-check if it was something I was doing weird before giving specific feedback:

* I've started with both the website and Gmail opened in the same browser window in separate tabs, and both of them opened in different browser windows. 
* When I try to make a purchase, the email correctly goes through with a sign up link with new parameters.
* When I click that link, it opens a new tab with a URL like `[website.com]/?np_auth_link_result=1&modal_checkout=1` -- if the checkout was on a subpage, the page slug is carried over, too. On this page, the modal isn't opened, and the header is funky (specifically the links on the right -- the logo size is my weird site settings):

CleanShot 2024-08-20 at 12 01 42

* The original tab DOES update the modal as described, but it's just not obvious because it's not happening in the tab I get when clicking the email link.

From your steps it sounds like that new tab that opens should have the updated checkout. Does anything I'm doing sound wrong?

chickenn00dle:
Thanks for the review @laurelfulford!

> From your steps it sounds like that new tab that opens should have the updated checkout. Does anything I'm doing sound wrong?

That's right. The new tab should also have the modal open with the updated checkout. Can you confirm:

- What is the magic link url in the otp email you received?
- Have you checked out both the blocks and plugin PRs?
- Can you try clearing your browsers local storage for your test site?

On my end, this is what I see when I click the magic link in the email:

https://github.com/user-attachments/assets/5681844b-91d3-49c8-97ad-ff47425403a8

laurelfulford:
Thanks @chickenn00dle! 

I did double-check that I had both checked out yesterday by deleting them and re-checking them out yesterday, but no dice. Today, I restarted Chrome and re-deleted and re-re-checked out PRs, and I'm getting the updated modal in the right tab now! 🙌 I'll continue with doing some functional testing to make sure everything works as described.

laurelfulford:
I went through all the test scenarios and didn't run into any problems! 

I also tried the smoke tests (logging in using the OTP, and running through a checkout while logged in) and things are functioning as expected.

Overall this is looking good to me from a functional perspective, and I didn't into any errors, but the code changes are complex enough I think it could be worth getting a second set of eyes from the engineering team just to make sure I'm not missing anything. I'm leaving this feedback as comment because of that; just let me know if you have any questions about that, or see any issues with it 🙂 Thanks!

src/reader-activation/index.js
```diff
@@ -422,6 +402,60 @@ function attachNewsletterFormListener() {
 	} );
 }
 
+
+/**
+ * Checkout functions.
+ */
```
miguelpeixe: Given that this set of functions is expanding for checkout-specific handling, would it make sense to move it to a separate `checkout.js`?

src/reader-activation/index.js
```diff
@@ -422,6 +402,60 @@ function attachNewsletterFormListener() {
...
+ * @return {Object} Checkout data.
+ */
+export function getCheckout() {
+	return store.get( 'checkout' ) || {};
+}
+
+/**
+ * Set the current checkout data.
+ *
+ * @param {Object} data Optiona. Checkout data.
```
miguelpeixe:
```suggestion
 * @param {Object} data Optional. Checkout data.
```

miguelpeixe: There's a bit of inconsistency in the doc blocks for the checkout functions regarding spacing and phrasing. Here it starts with "Optional.", other end with "Optional".

src/reader-activation-auth/auth-form.js
```diff
@@ -157,6 +183,12 @@ window.newspackRAS.push( function ( readerActivation ) {
 						body.set( 'reader-activation-auth-form', 1 );
 						body.set( 'npe', emailInput.value );
 						body.set( 'action', 'link' );
+						if ( readerActivation.getCheckoutStatus() ) {
```
miguelpeixe: NIT suggestion: I'd assume `getCheckoutStatus()` would return a string rather than a bool. Maybe this is more of a `hasCheckout()` or `isPendingCheckout()` type of function for a bool return?

src/reader-activation-auth/auth-form.js
```diff
@@ -97,6 +97,32 @@ window.newspackRAS.push( function ( readerActivation ) {
 			};
 			container.setFormAction( 'signin' );
 
+			/**
+			 * Get the redirect URL.
+			 */
+			const getRedirectUrl = () => {
```
miguelpeixe:
```suggestion
			const getCheckoutRedirectUrl = () => {
```

We can be more explicit that this redirect URL is meant for checkout redirection.

Also, if you like the idea of the `checkout.js` file, it looks like a nice home for it as well.

miguelpeixe (changes requested):
This is such a nice feature!

It's working well for most cases, the only problem I had was selecting the "Other" option for recurring donations in the Donate Block. The checkout opens with the middle option and not the custom value. It worked well for the One-time tab in the Donate Block.

Regarding the redirect URL, if we're using the reader data store to keep the checkout parameters, what do you think of just using a `?checkout` URL param for the redirect? Then read from the stored values instead of adding all the parameters to the URL. I think it'd result in a nicer URL and less code to maintain.

src/reader-activation/index.js
chickenn00dle: Great idea! Done in https://github.com/Automattic/newspack-plugin/pull/3343/commits/f68ba174858d8213dacda9f734355bd60f9fd638

src/reader-activation/index.js
chickenn00dle: Oops. Good catch. Fixed the typo and moved the instances of `Optional` to the end in each doc block in https://github.com/Automattic/newspack-plugin/pull/3343/commits/f68ba174858d8213dacda9f734355bd60f9fd638

src/reader-activation-auth/auth-form.js
chickenn00dle: Great idea. I went with `isPendingCheckout` in https://github.com/Automattic/newspack-plugin/pull/3343/commits/f68ba174858d8213dacda9f734355bd60f9fd638

src/reader-activation-auth/auth-form.js
chickenn00dle: Great idea! Moved in https://github.com/Automattic/newspack-plugin/pull/3343/commits/1ce3631b59bb4f3288484ec82641665631927b50

chickenn00dle:
Thanks for the review @miguelpeixe!

> the only problem I had was selecting the "Other" option for recurring donations in the Donate Block. The checkout opens with the middle option and not the custom value.

Good catch! Fixed this in the blocks PR in https://github.com/Automattic/newspack-blocks/pull/1841/commits/e3355634736b359dd92a119569faba54cf6dbe53

> Regarding the redirect URL, if we're using the reader data store to keep the checkout parameters, what do you think of just using a ?checkout URL param for the redirect? Then read from the stored values instead of adding all the parameters to the URL. I think it'd result in a nicer URL and less code to maintain.

This is a good idea. I considered this, but decided to add the URL params anyway so that the links in the emails could be reused even after the stored checkout data is cleared. That said, I'm happy to go this route if we feel this isn't necessary. Just let me know!

miguelpeixe (approved):
Thank you for the revisions! It's working flawlessly now.

> add the URL params anyway so that the links in the emails could be reused even after the stored checkout data is cleared.

I see the checkout more as a navigation state rather than something attached to the authentication process. That URL being accessed another time would not really work because that token has expired and would trigger a checkout for something that the reader have already paid.

I'd go as far as not even adding any param to the redirect URL and just detecting whether the checkout state is pending on any reader authentication and pop the modal.

This is a non-blocking suggestion, though!

chickenn00dle:
> I see the checkout more as a navigation state rather than something attached to the authentication process. That URL being accessed another time would not really work because that token has expired and would trigger a checkout for something that the reader have already paid.

Hmm. This is a good point. I think I'll address this in a follow-up PR though. For now I've opened a new task for this here: https://app.asana.com/0/1205234045751551/1208148789097630/f

PR #3345, opened 2024-08-15
feat: automatically disable guest authors

leogermani:
### All Submissions:

After the introduction of the Guest Contributor role, we now want to be more proactive and disable CAP's guest authors if the site is not using the feature yet.

CAP's Guest Authors feature will be disabled by default if there are no Guest Authors in the site. If you want to force enabling it, add the NEWSPACK_ENABLE_CAP_GUEST_AUTHORS constant to your wp-config.php file.

1. In a site that has some Guest Authors
2. Enable this branch and confirm the feature is still enabled and working
3. Check that the option was set: `wp option get newspack_check_site_has_cap_guest_authors` should return `yes`
4. Delete all your guest authors (or start with a new site)
5. Make sure to delete the option: `wp option delete newspack_check_site_has_cap_guest_authors`
6. Load any page of the site again
7. Confirm Guest Authors are disabled and you see our replacement page instead
8. Confirm the option now has the value `no`
9. Add `define( 'NEWSPACK_ENABLE_CAP_GUEST_AUTHORS', true );` to wp-config
10. Confirm CAP's Guest Authors are re-enabled

Testing CRON

1. Repeat steps 1 to 4
2. Run  wp cron event run newspack_re_check_guest_authors
3. Confirm Guest Authors are disabled and you see our replacement page instead
4. Confirm the option now has the value `no`

### Other information:


includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -75,6 +74,47 @@ public static function init() {
...
+			return;
+		}
+		if ( ! self::site_has_cap_guest_authors() ) {
+			add_filter( 'coauthors_guest_authors_enabled', '__return_false' );
+			add_action( 'admin_menu', [ __CLASS__, 'guest_author_menu_replacement' ] );
+		}
+	}
+
+	/**
+	 * Checks if the site has any guest authors. Will check it once in the database and store the result in an option.
```
adekbadek, line 122: I wonder if the once-ever check is not too restrictive. Without it (or with a periodical check, until there are no GAs), the migration path is simple: delete GAs and profit. The added step of having to delete the option necessitates the involvement of a TAM or developer.

adekbadek (approved):
Works as described, but I wonder if the migration won't be too involving.

includes/plugins/co-authors-plus/class-guest-contributor-role.php
leogermani: hmm good point. I'll add a wp_cron to check it periodically, I just don't want doing this query on every request

leogermani: @adekbadek Added the cron check and updated test instructions.

I also added a condition to refresh the check every time a guest author is deleted, so we would immediately update the UI, but the experience is not great. When you delete the last Guest Author, WP still redirects you to the Guest Authors page, but since we disable GAs, you end up in an error page because that page is not registered...

I removed it, but if you have an idea on how to make it better, here's what's needed:

```
add_action( 'delete_post', [ __CLASS__, 'guest_author_deleted' ], 10, 2 );

public static function guest_author_deleted( $post_id, $post ) {
	if ( 'guest-author' === $post->post_type ) {
		self::clear_site_has_cap_guest_authors_check();
	}
}
```

includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -78,6 +83,13 @@ public static function initialize() {
 
 		// Hide author email on the frontend, if it's a placeholder email.
 		\add_filter( 'theme_mod_show_author_email', [ __CLASS__, 'should_display_author_email' ] );
+
+		// Make sure we check again if the site has guest authors evey hour.
```
adekbadek: typo

```suggestion
		// Make sure we check again if the site has guest authors every hour.
```

adekbadek (approved):

PR #3346, opened 2024-08-16
feat(ga): disable tracking for editors regardless of RA status

adekbadek:
### All Submissions:

Removes the Reader Activation check when overriding Site Kit's tracking-disabled condition. With this change, tracking will be enabled for _all users except editors_ (as opposed to being disabled for _all logged-in users_).

#### Rationale

Regardless of Reader Activation status, sites might offer log-in for readers. These reader should be tracked in GA just like anonymous visitors. This change is based on the assumption that on a Newspack site, a logged-in users who can't edit post is a reader. 

1. Deactivate Reader Activation and visit Site Kit -> Settings -> Analytics view
1. Observe that the "Excluded from Analytics" setting is set to "Users that can write posts"

### Other information:

---
  - https://app.asana.com/0/0/1207899141354695

### DIFF
--- a/includes/plugins/google-site-kit/class-googlesitekit.php
+++ b/includes/plugins/google-site-kit/class-googlesitekit.php
@@ -34,11 +34,8 @@ class GoogleSiteKit {
 	 * @param array $googlesitekit_analytics_settings GA settings.
 	 */
 	public static function filter_ga_settings( $googlesitekit_analytics_settings ) {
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Works as described

Do you think it's worth adding a bit more info on the "why" of this change since it has an impact to change the current behavior of analytics in live sites?

adekbadek:
Updated the PR description, so it can be later used in the release notes.

PR #3325, opened 2024-08-08
feat: add woo team sync metadata (WIP)

leogermani:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds a new metadata field for the ESP sync containing the Woo Teams the user is part of - if any, and if the Teams for Memberships plugin is installed.

Note that a sync to ESP is not triggered when a user is granted a membership or is added to a team. This will be subject for another PR

### How to test the changes in this Pull Request:

1. Activate Teams for Memberships
2. Create two teams
3. Add a user to the two teams (teams need to be assigned to different membership plans)
4. Log in as that user and perform an action that will trigger a sync - for example, subscribe to a newsletter or make a donation
5. Confirm that  a new field "Woo Team" was added to the contact in the ESP with the slugs of the teams the user is a part of.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207955183933546

PR #3352, opened 2024-08-21
fix(esp-meta): handle state of the 'Woo Team' meta

adekbadek:
### All Submissions:

Fixes a small issue with #3325.

1. On `trunk`, activate the Teams for Memberships plugin
2. In the Engagement wizard, ensure the "Woo Team" meta field is **deselected**
3. Add a user to a team, then log in as them and make a donation
4. Observe that the synced contact in the ESP has the "NP_Woo Team" field value set
5. Switch to this branch, repeat step 3 & 4 
6. Observe the ESP contact has the field value not set
7. Enable the field in the Engagement wizard and re-test – observe the field is now set

### Other information:

### DIFF
--- a/includes/plugins/class-newspack-newsletters.php
+++ b/includes/plugins/class-newspack-newsletters.php
@@ -183,6 +183,23 @@ class Newspack_Newsletters {
 	}
 
 	/**
+	 * Get enabled fields which match provided keys.
+	 * Will return key-value pairs of enabled fields which match the keys provided.
+	 *
+	 * @param string[] $keys Array of keys to match.
+	 */
+	public static function get_applicable_fields( $keys ) {
+		$enabled_fields = self::get_metadata_fields();
+		return array_filter(
+			self::get_metadata_keys(),
+			function( $val, $key ) use ( $keys, $enabled_fields ) {
+				return in_array( $key, $keys ) && in_array( $val, $enabled_fields );
+			},
+			ARRAY_FILTER_USE_BOTH
+		);
+	}
+
+	/**
 	 * Update the list of fields to be synced to the connected ESP.
 	 *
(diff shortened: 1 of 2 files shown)


leogermani (approved):
Approving as it works as expected, but I think this interface could be a little better. Ideas:

* `applicable_fields` sounds weird to me. Maybe a better name there
* Also offer a `should_sync_field( $field_name )` that would be an easier method to use if you are just checking for one field

or 

* the `newspack_esp_sync_normalize_contact` filter could pass the `applicable_filters` as the second argument, then the callback could use it to do its own check...

wdyt?

adekbadek:
> `applicable_fields` sounds weird to me. Maybe a better name there

Agreed, changed the name to "filtered enabled fields" in c77e150.

> Also offer a `should_sync_field( $field_name )` that would be an easier method to use if you are just checking for one field

Hm, not a fan of adding a method just to handle a single value. Unnecessary complexity IMO.

> the `newspack_esp_sync_normalize_contact` filter could pass the `applicable_filters` as the second argument, then the callback could use it to do its own check

That's an alternative, but I'm not sure if one is better than the other ¯\_(ツ)_/¯

leogermani:
@adekbadek I was looking at this again while reviewing https://github.com/Automattic/newspack-plugin/pull/3360 and I noticed that this is a redundant check...

The metadata will already be excluded from the contact if it is disabled in admin. It happens [here](https://github.com/Automattic/newspack-plugin/blob/5cea128bbc250d39a942a19b464fa1dd8dce3cd1/includes/plugins/class-newspack-newsletters.php#L337).

I've tested this again and I can safely remove the check added in this PR. The meta will only be synced if it's active.

Is there another scenario where this will not be respected?

cc @miguelpeixe

PR #3359, opened 2024-08-22
feat(ras): esp sync tools

miguelpeixe:
### All Submissions:

1200530782742699-as-1207717004877609/f

Moves the logic for syncing Woo customers and orders to the ESP from the `newspack-subscription-migrations` plugin to here.

The original class was refactored so it provides generic methods to be extended by the existing CLI and new admin functionality. Its name has also changed from "Woo Sync" to "ESP Sync", given Woo is part of the contact data but the goal is to sync readers to an ESP.

Some changes to the generic logic are worth mentioning:

 - `can_esp_sync( bool $return_errors )` was implemented so it can be called across CLI and admin functionality with proper error handling
 - `get_batch_of_customers()` was changed to `get_batch_of_readers()`. It's now using `\Newspack\Reader_Activation::get_reader_roles()` because although they are customers, every WP user can be a customer and the goal is to sync RAS readers
 - The `sync( $contact, $context )` method (previously `sync_contact()`) was modified to support any ESP through the `Newspack_Newsletters_Contacts::upsert()` method. For that, [a new helper method is proposed for RAS](https://github.com/Automattic/newspack-plugin/pull/3355) so we can get the master list given any provider

The CLI command changed from `wp newspack woo resync` to `wp newspack esp sync`, but preserves all the original functionality and args support. The exception is for the flag `--migrated-subscriptions`, which requires the `newspack-subscription-migrations` plugin to be installed, otherwise it returns an error. A new flag was added for `--sync-context` to be passed down to `Newspack_Newsletters_Contacts::upsert()` ([`--context` is reserved](https://make.wordpress.org/cli/handbook/references/config/#context-configuration)).

The new functionality allows an admin user to sync contacts through the dashboard user table, both individually and through the bulk action dropdown:

| Single | Bulk |
| --- | --- |
| image | image |

#### CLI

1. Run `wp newspack esp sync --dry-run` for a full dry run
2. Confirm the command runs successfully, going through all the readers in your database
3. Choose a reader and run `wp newspack esp sync --user-ids=USER_ID` with their user ID
4. Confirm via PHP logs that `[Newspack_Newsletters_Contacts::upsert]: Adding contact to list(s): LIST_ID. Provider is PROVIDER.` is called and you can inspect the normalized contact data
5. Do the same but now with 2 user IDs: `wp newspack esp sync --user-ids=USER_ID1,USER_ID2`
6. Confirm via PHP logs the 2 entries as in step 4
7. Grab 2 Woo Subscription IDs and run: `wp newspack esp sync --subscription-ids=ID1,ID2`
8. Confirm the command runs without issues and the PHP log entries for the `::upsert()` is correct
9. Grab 2 Woo Order IDs and run: `wp newspack esp sync --order-ids=ID1,ID2`
10. Confirm the command runs without issues and the PHP log entries for the `::upsert()` is correct
11. Cancel a reader subscription and run a sync for active subs only, along with a reader with an active subscription: `wp newspack esp sync --active-only --user-ids=USER_ID1,USER_ID2`
12. Confirm only the reader with an active sub got synced
13. Test dry run syncs modifying `--offset`, `--max-batches`, and `--batch-size` to confirm these values are applied correctly

#### Admin

1. Navigate to "Users", in WP dashboard
2. Hover the rows and confirm you see a "Resync contact to ESP" row action link
3. Click and confirm you get a success message and the PHP logs show `::upsert()` was called correctly
4. Select multiple readers, click on the Bulk actions dropdown, select "Resync to the ESP", and apply
5. Confirm you get a success message and the PHP logs show `::upsert()` was called correctly for every selected contact

### Other information:

### DIFF
--- a/includes/cli/class-ras-esp-sync.php
+++ b/includes/cli/class-ras-esp-sync.php
@@ -85,14 +85,14 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 
 		static::log( __( 'Running ESP contact sync...', 'newspack-plugin' ) );
 
-		$can_sync = static::can_esp_sync( true );
+		$can_sync = self::can_esp_sync( true );
 		if ( ! $config['is_dry_run'] && $can_sync->has_errors() ) {
 			return $can_sync;
 		}
 
 		// If syncing only migrated subscriptions.
 		if ( $config['migrated_only'] ) {
-			$config['subscription_ids'] = static::get_migrated_subscriptions( $config['migrated_only'], $config['batch_size'], $config['offset'], $config['active_only'] );
+			$config['subscription_ids'] = self::get_migrated_subscriptions( $config['migrated_only'], $config['batch_size'], $config['offset'], $config['active_only'] );
 			if ( \is_wp_error( $config['subscription_ids'] ) ) {
 				return $config['subscription_ids'];
 			}
@@ -118,7 +118,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 					continue;
 				}
 
-				$result = static::sync_contact( $subscription, $config['is_dry_run'] );
+				$result = self::sync_contact( $subscription, $config['is_dry_run'] );
 				if ( \is_wp_error( $result ) ) {
 					static::log(
 						sprintf(
@@ -139,7 +139,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 					}
 
 					$next_batch_offset = $config['offset'] + ( $batches * $config['batch_size'] );
-					$config['subscription_ids'] = static::get_migrated_subscriptions( $config['migrated_only'], $config['batch_size'], $next_batch_offset, $config['active_only'] );
+					$config['subscription_ids'] = self::get_migrated_subscriptions( $config['migrated_only'], $config['batch_size'], $next_batch_offset, $config['active_only'] );
 				}
 			}
 		}
@@ -162,7 +162,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 					continue;
 				}
 
-				$result = static::sync_contact( $order, $config['is_dry_run'] );
+				$result = self::sync_contact( $order, $config['is_dry_run'] );
 				if ( \is_wp_error( $result ) ) {
 					static::log(
 						sprintf(
@@ -180,8 +180,8 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 		if ( ! empty( $config['user_ids'] ) ) {
 			static::log( __( 'Syncing by customer user ID...', 'newspack-plugin' ) );
 			foreach ( $config['user_ids'] as $user_id ) {
-				if ( ! $config['active_only'] || static::user_has_active_subscriptions( $user_id ) ) {
-					$result = static::sync_contact( $user_id, $config['is_dry_run'] );
+				if ( ! $config['active_only'] || self::user_has_active_subscriptions( $user_id ) ) {
+					$result = self::sync_contact( $user_id, $config['is_dry_run'] );
 					if ( \is_wp_error( $result ) ) {
 						static::log(
 							sprintf(
@@ -204,13 +204,13 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 			false === $config['migrated_only']
 		) {
 			static::log( __( 'Syncing all readers...', 'newspack-plugin' ) );
-			$user_ids = static::get_batch_of_readers( $config['batch_size'], $config['offset'] );
+			$user_ids = self::get_batch_of_readers( $config['batch_size'], $config['offset'] );
 			$batches  = 0;
 
 			while ( $user_ids ) {
 				$user_id = array_shift( $user_ids );
-				if ( ! $config['active_only'] || static::user_has_active_subscriptions( $user_id ) ) {
-					$result = static::sync_contact( $user_id, $config['is_dry_run'] );
+				if ( ! $config['active_only'] || self::user_has_active_subscriptions( $user_id ) ) {
+					$result = self::sync_contact( $user_id, $config['is_dry_run'] );
 					if ( \is_wp_error( $result ) ) {
 						static::log(
 							sprintf(
@@ -231,7 +231,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 						break;
 					}
 
-					$user_ids = static::get_batch_of_readers( $config['batch_size'], $config['offset'] + ( $batches * $config['batch_size'] ) );
+					$user_ids = self::get_batch_of_readers( $config['batch_size'], $config['offset'] + ( $batches * $config['batch_size'] ) );
 				}
 			}
 		}
@@ -355,7 +355,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 		$config['max_batches']      = ! empty( $assoc_args['max-batches'] ) ? intval( $assoc_args['max-batches'] ) : 0;
 		$config['context']          = ! empty( $assoc_args['sync-context'] ) ? $assoc_args['sync-context'] : static::$context;
 
-		$processed = static::sync_contacts( $config );
+		$processed = self::sync_contacts( $config );
 
 		if ( \is_wp_error( $processed ) ) {
 			WP_CLI::error( $processed->get_error_message() );
--- a/includes/cli/class-ras-esp-sync.php
+++ b/includes/cli/class-ras-esp-sync.php
@@ -25,7 +25,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 	 *
 	 * @var string
 	 */
-	protected static $context = 'CLI ESP Sync';
+	protected static $context = 'Contact sync manually triggered via CLI';
 
 	/**
 	 * The final results object.
--- a/includes/cli/class-ras-esp-sync.php
+++ b/includes/cli/class-ras-esp-sync.php
@@ -39,7 +39,7 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 	}
 
 	/**
-	 * Resync reader contact data to the connected ESP.
+	 * Sync reader contact data to the connected ESP.
 	 *
 	 * @param array $config {
 	 *   Configuration options.
@@ -56,9 +56,9 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 	 *   @type bool        $config['is_dry_run'] True if a dry run.
 	 * }
 	 *
-	 * @return int|\WP_Error Number of resynced contacts or WP_Error.
+	 * @return int|\WP_Error Number of synced contacts or WP_Error.
 	 */
-	private static function resync_contacts( $config ) {
+	private static function sync_contacts( $config ) {
 		$default_config = [
 			'active_only'      => false,
 			'migrated_only'    => false,
@@ -72,14 +72,14 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 		];
 		$config = \wp_parse_args( $config, $default_config );
 
-		static::log( __( 'Running WooCommerce-to-ESP contact resync...', 'newspack-plugin' ) );
+		static::log( __( 'Running ESP contact sync...', 'newspack-plugin' ) );
 
 		$can_sync = static::can_sync_contacts( true );
 		if ( ! $config['is_dry_run'] && $can_sync->has_errors() ) {
 			return $can_sync;
 		}
 
-		// If resyncing only migrated subscriptions.
+		// If syncing only migrated subscriptions.
 		if ( $config['migrated_only'] ) {
 			$config['subscription_ids'] = static::get_migrated_subscriptions( $config['migrated_only'], $config['batch_size'], $config['offset'], $config['active_only'] );
 			if ( \is_wp_error( $config['subscription_ids'] ) ) {
@@ -107,12 +107,12 @@ class RAS_ESP_Sync extends Reader_Activation\ESP_Sync {
 					continue;
 				}
 
-				$result = static::resync_contact( $subscription, $config['is_dry_run'] );
+				$result = static::sync_contact( $subscription, $config['is_dry_run'] );
 				if ( \is_wp_error( $result ) ) {
 					static::log(
 						sprintf(
 							// Translators: %1$d is the subscription ID arg passed to the script. %2$s is the error message.
-							__( 'Error resyncing contact info for subscription ID %1$d. %2$s', 'newspack-plugin' ),
+							__( 'Error syncing contact info for subscription ID %1$d. %2$s', 'newspack-plugin' ),
 							$subscription_id,
 							$result->get_error_message()
 						)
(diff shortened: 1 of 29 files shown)


includes/reader-activation/class-esp-sync.php
```diff
@@ -0,0 +1,202 @@
...
+		if ( ! $is_dry_run && $can_sync->has_errors() ) {
+			return $can_sync;
+		}
+
+		$is_order = $user_id_or_order instanceof \WC_Order;
+		$order    = $is_order ? $user_id_or_order : false;
+		$user_id  = $is_order ? $order->get_customer_id() : $user_id_or_order;
+		$user     = \get_userdata( $user_id );
+
+		// Backfill Network Registration Site field if needed.
```
leogermani, line 150: this looks weird here. Newspack Network should handle this, adding the field via filters as it already does

includes/reader-activation/class-esp-sync-admin.php
```diff
@@ -0,0 +1,208 @@
...
+			\wp_die( \esc_html__( 'You do not have permission to do that.', 'newspack-plugin' ) );
+		}
+
+		$user = \get_user_by( 'id', $user_id );
+
+		if ( ! $user || \is_wp_error( $user ) ) {
+			\wp_die( \esc_html__( 'User not found.', 'newspack-plugin' ) );
+		}
+
+		$config = [
```
leogermani: is this used?

includes/reader-activation/class-esp-sync-admin.php
```diff
@@ -0,0 +1,208 @@
...
+			'user_ids' => [ $user_id ],
+		];
+		$result = static::sync_contact( $user_id );
```
leogermani: this should be `self`. We are using the parent's method

includes/reader-activation/class-esp-sync-admin.php
```diff
@@ -0,0 +1,208 @@
...
+		}
+		$can_sync = self::can_esp_sync( true );
+		if ( $can_sync->has_errors() ) {
+			\wp_die( $can_sync ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		}
+		if ( ! \current_user_can( 'edit_users' ) ) {
+			\wp_die( \esc_html__( 'You do not have permission to do that.', 'newspack-plugin' ) );
+		}
+		foreach ( $items as $user_id ) {
+			$result = static::sync_contact( $user_id );
```
leogermani: `self`

includes/cli/class-ras-esp-sync.php
```diff
@@ -0,0 +1,376 @@
...
+		$config['migrated_only']    = ! empty( $assoc_args['migrated-subscriptions'] ) ? $assoc_args['migrated-subscriptions'] : false;
+		$config['subscription_ids'] = ! empty( $assoc_args['subscription-ids'] ) ? explode( ',', $assoc_args['subscription-ids'] ) : false;
+		$config['user_ids']         = ! empty( $assoc_args['user-ids'] ) ? explode( ',', $assoc_args['user-ids'] ) : false;
+		$config['order_ids']        = ! empty( $assoc_args['order-ids'] ) ? explode( ',', $assoc_args['order-ids'] ) : false;
+		$config['batch_size']       = ! empty( $assoc_args['batch-size'] ) ? intval( $assoc_args['batch-size'] ) : 10;
+		$config['offset']           = ! empty( $assoc_args['offset'] ) ? intval( $assoc_args['offset'] ) : 0;
+		$config['max_batches']      = ! empty( $assoc_args['max-batches'] ) ? intval( $assoc_args['max-batches'] ) : 0;
+		$config['context']          = ! empty( $assoc_args['sync-context'] ) ? $assoc_args['sync-context'] : static::$context;
+
+		$processed = static::sync_contacts( $config );
```
leogermani: can also be `self`

includes/cli/class-ras-esp-sync.php
```diff
@@ -0,0 +1,376 @@
...
+
+		// Default behavior: sync all readers.
+		if (
+			false === $config['user_ids'] &&
+			false === $config['order_ids'] &&
+			false === $config['subscription_ids'] &&
+			false === $config['migrated_only']
+		) {
+			static::log( __( 'Syncing all readers...', 'newspack-plugin' ) );
+			$user_ids = static::get_batch_of_readers( $config['batch_size'], $config['offset'] );
```
leogermani: "all" readers here is not accurate, and can lead the user to believe something went wrong and the command is not respecting the chosen filter.

includes/reader-activation/class-esp-sync.php
leogermani: Network does this using the `newspack_data_events_reader_registered_metadata` filter currently present in our Connectors. That should be enough

https://github.com/Automattic/newspack-network/blob/trunk/includes/class-esp-metadata-sync.php#L26

includes/reader-activation/class-esp-sync.php
```diff
@@ -0,0 +1,202 @@
...
+			);
+		}
+
+		// Ensure the customer has a billing address.
+		if ( ! $customer->get_billing_email() && $customer->get_email() ) {
+			$customer->set_billing_email( $customer->get_email() );
+			$customer->save();
+		}
+
+		$contact = $is_order ? WooCommerce_Connection::get_contact_from_order( $order ) : WooCommerce_Connection::get_contact_from_customer( $customer );
```
leogermani, line 180: Seems that all the decisions of what metadata to sync is done in `get_contact_from_order`. This is one of the things you are getting out of there and bringing to the scope of sync, right?

leogermani (changes requested):
Now I see why you fell on a rabbit hole

It's really confusing how things are duplicate. The current connectors and what you have here... the normalize_contact filter acting at the end of it all...

I'm not sure how much repetition is between this and the what happens at `normalize_contact_data` if any. I bet you are making it all better in the upcoming refactors... we need to look at all usages of the `newspack_esp_sync_normalize_contact` filter.

I left a few minor comments.

CLI works great. wp-admin works great

Bulk action on wp-admin, as expected, times out if I try to run it on a full page of users... I'm not sure we want that... IMO, either we set up a proper way to do it async, or via Scheduled actions, or we should leave it out

includes/reader-activation/class-esp-sync-admin.php
miguelpeixe: You're right, all method calling in this class should be self. Fixed in 97b8056918b22ba37358742c1fc73b1e57936b6f

includes/reader-activation/class-esp-sync-admin.php
miguelpeixe: 97b8056918b22ba37358742c1fc73b1e57936b6f

includes/reader-activation/class-esp-sync-admin.php
miguelpeixe: Nope, got left behind after some more refactoring. Removed in 97b8056918b22ba37358742c1fc73b1e57936b6f

includes/reader-activation/class-esp-sync.php
miguelpeixe: It's part of the original tool in `newspack-subscription-migrations`, I didn't want to change it. So you're saying it's redundant logic and should no longer be in this sync?

includes/cli/class-ras-esp-sync.php
miguelpeixe: This one I thought some extension could be useful but you're right. We don't want more of that :)

includes/cli/class-ras-esp-sync.php
miguelpeixe: This is the case here. If no arguments are passed, it'll batch through every reader account.

includes/cli/class-ras-esp-sync.php
miguelpeixe: Fixed this and other method calling instances in a8656cdc8f3baf472e4e05effb5617758a0bbb0b

includes/reader-activation/class-esp-sync.php
miguelpeixe: Yep, you can see [here](https://github.com/Automattic/newspack-plugin/pull/3362/files#diff-c1aa6e451da797c3ebd6c6fa112527eefc89443140e173e276d25af8adc135b8) that the draft PR proposes a WooCommerce class inside the Sync namespace for handling contact generation based on Woo data.

miguelpeixe:
Thank you for the review!

> Now I see why you fell on a rabbit hole

Right? 😄 

> I'm not sure how much repetition is between this and the what happens at normalize_contact_data if any. I bet you are making it all better in the upcoming refactors... we need to look at all usages of the newspack_esp_sync_normalize_contact filter.

Yes, in my draft PR that method and all other metadata related methods from the `Newspack_Newsletters` class [are moving to a `Metadata` class](https://github.com/Automattic/newspack-plugin/pull/3362/files#diff-36a03a01f8ed1b00b692519c1faa46431f8983d15032f3c6324d6088d61717be). The `normalize_contact_data` should no longer be a filter and should [be called right before the `::upsert()` inside the `sync()` method](https://github.com/Automattic/newspack-plugin/pull/3362/files#diff-54a027d8960b76963f30ce437f31ebe5e011ea94e6860c2b60423378f5330731R101) for the `ESP_Sync` class. Less fragile and more readable.

> Bulk action on wp-admin, as expected, times out if I try to run it on a full page of users... I'm not sure we want that... IMO, either we set up a proper way to do it async, or via Scheduled actions, or we should leave it out

I like the idea of using AS for this, but we should also care for the feedback. How will we notify the state of the async request to the admin? Certainly something for another PR.

I'm ok leaving it as is but adding an error message if more than, let's say, 5 users are selected. The admin will receive an error notice: "Exceeded selection limit. Please select no more than 5 users per execution." They'll adjust and be able to use _some_ batch action.

includes/cli/class-ras-esp-sync.php
leogermani: hmm when I ran it with "active only" (or some other filter) I got the "Syncing all readers... " message. will try to reproduce it

includes/reader-activation/class-esp-sync.php
leogermani: Need to check. If that filter is still being called in this flow, things should be fine. but need to test

leogermani:
> I'm ok leaving it as is but adding an error message if more than, let's say, 5 users are selected. The admin will receive an error notice: "Exceeded selection limit. Please select no more than 5 users per execution." They'll adjust and be able to use some batch action.

This could work. I don't love it but... maybe get a third opinion?

miguelpeixe:
As we discussed internally, b98bf91ee447b198880c3f76ab42fbaa43cb3ef1 implements AS without feedback on batch success. The admin should receive the following notice when running the batch action:

image

The link takes the admin to Woo's AS status page, querying by the hook name (`newspack_sync_admin_batch`).

If AS is not available (Woo not installed), the bulk action is not available.

includes/reader-activation/class-esp-sync.php
leogermani: I've just tested it and it doesn't work.

I think we need to make sure that this sync goes through the same filters the current one goes. Or we need to adapt Newspack Network and add another filter to do the same thing. I prefer if Network could do this using only one filter.

Another integration to look at and see if still works is the Woo team ID -> https://github.com/Automattic/newspack-plugin/pull/3325

includes/cli/class-ras-esp-sync.php
```diff
@@ -0,0 +1,376 @@
...
+	 * Get a batch of readers' IDs.
+	 *
+	 * @param int $batch_size Number of readers to get.
+	 * @param int $offset     Number to skip.
+	 *
+	 * @return array|false Array of user IDs, or false if no more to fetch.
+	 */
+	private static function get_batch_of_readers( $batch_size, $offset = 0 ) {
+		$roles = Reader_Activation::get_reader_roles();
+		if ( defined( 'NEWSPACK_NETWORK_READER_ROLE' ) && ! in_array( NEWSPACK_NETWORK_READER_ROLE, $roles, true ) ) {
```
leogermani: This is not necessary. Newspack Network, if active, will filter `get_reader_roles` and add its role there

includes/reader-activation/class-esp-sync-admin.php
```diff
@@ -0,0 +1,216 @@
...
+		$update = sanitize_text_field( wp_unslash( $_GET['update'] ) );
+		if ( self::ADMIN_ACTION !== $update ) {
+			return;
+		}
+		$message = __( 'Contact resynced to the ESP.', 'newspack-plugin' );
+		if ( isset( $_GET['scheduled-sync-contacts'] ) ) {
+			$scheduled_syncs = absint( wp_unslash( $_GET['scheduled-sync-contacts'] ) );
+			$message         = sprintf(
+				// translators: %1$d: number of contacts resynced. %2$d: estimated time in seconds.
+				__( '%1$d contacts scheculed for resync to the ESP. This should take approximately %2$d seconds.', 'newspack-plugin' ),
```
leogermani: Looking good. Nice that we don't show the option if ActionScheduler is not present.

As a suggestion, I think we should just say that the contacts are going to be synced in background and not give an estimate.

There is only 10 contacts in a page, so the max amount we are going to show is 20 seconds
1. 20 seconds doesn't sound that long to justify all the hassle
2. If it really were only 20 seconds it would not timeout.

Each action is processed by AS in an independent request, so there is the overhead of making a request, etc. It takes a little more.... and how long it takes is a bit unpredictable...

idk, just feel that this info there doesn't help.

"They will be processed in a little while..." idk

includes/reader-activation/class-esp-sync.php
miguelpeixe: You mean it doesn't work without it, right? So we leave it for now. I also prefer for this highly specific piece code to not live inside this class, but I don't want to change more things in this PR.

> Another integration to look at and see if still works is the Woo team ID -> https://github.com/Automattic/newspack-plugin/pull/3325

This should work the same, the filters didn't change and the `Newspack_Newsletters` class, which runs its normalization after the `::upsert()` call, still behaves as intended.

includes/cli/class-ras-esp-sync.php
miguelpeixe: Well, another reason why this should've been reader roles all along 😄

miguelpeixe: Fixed in e6253280f7a81f5284cde4b3d3d3e21f52fd2d00

includes/reader-activation/class-esp-sync-admin.php
miguelpeixe: We can say it'll take "a couple of minutes". It's a good enough estimate.

includes/cli/class-ras-esp-sync.php
leogermani: Here:

image

includes/reader-activation/class-esp-sync.php
leogermani: > So we leave it for now. I also prefer for this highly specific piece code to not live inside this class, but I don't want to change more things in this PR.

I would sleep better if we fixed this in this PR, but it's ok if you promise this is being fixed in the follow up PRs - but they can't take long then. This is really ugly and can lead to problems if we decide to do any refactors on the network plugin

includes/reader-activation/class-esp-sync-admin.php
miguelpeixe: Updated in 0d554eb994eb23d7b5f5df1b5aec61a5f6cce717

includes/cli/class-ras-esp-sync.php
miguelpeixe: Fixed in 8b3ccad84e1530843261f73e27d198402ac02b3b

leogermani (approved):

PR #3360, opened 2024-08-22
feat(ras): unify ESP connector strategy for data events

miguelpeixe:
### All Submissions:

Following the work from #3359, this PR unifies the ESP connectors into one `ESP_Connector` using the recently implemented `Newspack\Reader_Activation\ESP_Sync::sync()` method for syncing contacts via data events handlers.

The only new behavior from this change is that the `newsletter_selection` field will no longer be exclusive to ActiveCampaign. It's a compromise for having a single connector handling any ESP the Newspack Newsletters plugin supports. If we find that this field should not live anywhere outside ActiveCampaign, we have 2 options:

 - Add a provider name check inside the handlers that populate that field
 - Manually toggle it off in the **Metadata field settings** page, inside Engagement -> Reader Activation -> Advanced settings, for publishers that don't want it.

I'd go with the latter, favoring the more generic approach.

The contact status handling for Mailchimp also moved to a filter inside the `Reader_Activation` class. Works but I don't love it. I'd appreciate suggestions on how to implement that requirement without hurting this effort's goal.

By using the new `ESP_Sync::can_esp_sync()` check, unless Newspack Manager says you're a live site, the sync will not go through without the new `NEWSPACK_ALLOW_READER_SYNC`. This constant should be set for testing purposes. It's important to inspect how this method works because it aims to be the single point where we ask if syncs should occur.

Regarding [this filter](https://github.com/Automattic/newspack-plugin/blob/373e157e98f3cd97c3da07e549fd647cb2f66761/includes/data-events/connectors/class-esp-connector.php#L68-L76), it was [built for Newspack Network](https://github.com/Automattic/newspack-network/pull/87). It's preserved in this PR but I wonder if we could do this differently and not have another filter messing with sync data. Perhaps the `newspack_esp_sync_contact`? cc @leogermani

## Update

1d198b9c3dc931af8fe943baccce35be7308268a proposes a change regarding the filter mentioned above. The `newspack_data_events_reader_registered_metadata` filter was removed to favor the new `newspack_esp_sync_contact` filter. This change, followed by https://github.com/Automattic/newspack-network/pull/87,  also allows for the backfilling logic in place for Newspack Network to no longer live in the ESP_Sync class, which was also removed in the commit.

To also accommodate that change, the Teams for Memberships integration filter was changed from `newspack_esp_sync_normalize_contact` to `newspack_esp_sync_contact`, which should result in no functional changes.

1. Set the `NEWSPACK_ALLOW_READER_SYNC` constant to `true`
2. Make sure you have RAS setup with Mailchimp and an audience ID as master list
3. Visit the **Metadata field settings** in Engagement -> Reader Activation -> Advanced and confirm all fields are enabled
4. Register a new reader, watch the logs, and confirm the sync goes through with the correct metadata
5. Visit the contact page in Mailchimp and confirm the merge fields match the sent data and contact subscription status matches the configured value in RAS settings
6. Signed in as the new reader, donate and confirm the same process
7. Make a recurring donation and confirm the `last_payment_date`, `recurring_payment`, `billing_cycle`, `subscription_start_date`, `next_payment_date`, `membership_status`, and `total_paid` fields are updated with the new data
8. Purchase a non-donation subscription and confirm `product_name` and the above fields update with the new data
9. Subscribe to newsletters using the Newsletter Subscription Form block and confirm the `newsletter_selection` field updates
10. Associate the reader to a membership plan and confirm the `membership_status`, `membership_plan`, `membership_start_date`, and `membership_end_date` updates according to the membership data
11. Switch your ESP to ActiveCampaign and repeat steps 4-10

### Other information:


leogermani:
> Regarding [this filter](https://github.com/Automattic/newspack-plugin/blob/373e157e98f3cd97c3da07e549fd647cb2f66761/includes/data-events/connectors/class-esp-connector.php#L68-L76), it was https://github.com/Automattic/newspack-network/pull/87. It's preserved in this PR but I wonder if we could do this differently and not have another filter messing with sync data. Perhaps the newspack_esp_sync_contact? cc @leogermani

You decide! In this new strutcture, what should be the right way for a plugin (or another part of the code) to register a new metadata and add it to the payload. Like[ this does](https://github.com/Automattic/newspack-plugin/pull/3325). Later improved by [this](https://github.com/Automattic/newspack-plugin/pull/3352).

miguelpeixe:
Thanks, @leogermani!

Just pushed a change with 1d198b9c3dc931af8fe943baccce35be7308268a and https://github.com/Automattic/newspack-network/pull/132. I updated the PR description. This allows for the backfilling logic in place for Newspack Network to no longer live in the ESP_Sync class, which feels good.

leogermani (approved):
This all looks great!

I think the mailchimp status filter is fine. It's better than some other mailchimp specific things that are inserted in the middle of some methods (like the first/last name handling in the middle of `normalize_contact`)

miguelpeixe:
> I think the mailchimp status filter is fine. It's better than some other mailchimp specific things that are inserted in the middle of some methods (like the first/last name handling in the middle of normalize_contact)

I know: https://github.com/Automattic/newspack-newsletters/pull/1628

PR #3362, opened 2024-08-22
feat(ras): sync class

miguelpeixe:
### All Submissions:

Following #3359 and #3360, this PR continues to work on centralizing how we perform reader sync. Sync data is currently handled in a few different places:

 - [`Newspack\WooCommerce_Connection`](https://github.com/Automattic/newspack-plugin/blob/b5f0f37395485add7e643c6daec07cbad1bb5e4b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php): generating contact data from customer or order ID
 - [`Newspack\Newspack_Newsletters`](https://github.com/Automattic/newspack-plugin/blob/b5f0f37395485add7e643c6daec07cbad1bb5e4b/includes/plugins/class-newspack-newsletters.php): managing the metadata and normalizing contact data in syncs

The `Newspack\Newspack_Newsletters` class has been deprecated and its metadata handling moved to [`Newspack\Reader_Activation\Sync\Metadata`](https://github.com/Automattic/newspack-plugin/blob/5b06af1e401594b1cd34c53949778495eaec0c21/includes/reader-activation/sync/class-metadata.php). A few notable changes in its behavior:
 - The `normalize_contact_data` is no longer applied as a filter to `newspack_newsletters_contact_data`. It's now called inside `Newspack\Reader_Activation\ESP_Sync::sync()`, right after the `newspack_esp_sync_contact` filter and before the `Newspack_Newsletters_Contacts::upsert()` call:

https://github.com/Automattic/newspack-plugin/blob/5b06af1e401594b1cd34c53949778495eaec0c21/includes/reader-activation/sync/class-esp-sync.php#L97-L109

 - The `normalize_contact_data` should no longer handle provider-specific adjustments, like Mailchimp's First Name and Last Name merge fields. See https://github.com/Automattic/newspack-newsletters/pull/1628
 - [Log when a metadata key is being ignored for not being allowed in](https://github.com/Automattic/newspack-plugin/blob/5b06af1e401594b1cd34c53949778495eaec0c21/includes/reader-activation/sync/class-metadata.php#L381)
 - Allow UTM keys to be sent raw, like `signup_page_utm_campaign`, and be formatted during normalization

From `Newspack\WooCommerce_Connection`, the methods `should_sync_order()`, `get_order_metadata()`, `get_contact_from_customer()`, and `get_contact_from_order()` moved to a new [`Newspack\Reader_Activation\Sync\WooCommerce`](https://github.com/Automattic/newspack-plugin/blob/5b06af1e401594b1cd34c53949778495eaec0c21/includes/reader-activation/sync/class-woocommerce.php) class.

The metadata generation on `get_order_metadata()` and `get_contact_from_customer()` no longer formats the keys with the prefixes from `Metadata`. Instead, it uses the raw keys to be later transformed via `normalize_contact_data`. This is also the case in other places where metadata is populated, like the [`Newspack\Data_Events\Connectors\ESP_Connector`](https://github.com/Automattic/newspack-plugin/blob/5b06af1e401594b1cd34c53949778495eaec0c21/includes/data-events/connectors/class-esp-connector.php) and [the Teams for Memberships integration](https://github.com/Automattic/newspack-plugin/blob/32584d801e5ae74dcc22c88a234ab22f755f13e1/includes/plugins/class-teams-for-memberships.php#L99).

All sync methods should be tested. First, repeat the testing steps from #3359 and #3360.

Finally, we should also test the UTM integration:

1. In a fresh session, visit your site with UTM parameters: `?utm_source=website&utm_medium=homepage&utm_campaign=header`
2. Donate and confirm the sync goes through including the UTM parameters

### Other information:


includes/plugins/class-newspack-newsletters.php
```diff
@@ -1,446 +0,0 @@
...
-			return $lists;
-		}
-		if ( empty( $lists ) ) {
-			return [ $master_list_id ];
-		}
-		if ( array_search( $master_list_id, $lists ) === false ) {
-			$lists[] = $master_list_id;
-		}
-		return $lists;
-	}
```
miguelpeixe, line 444: This method is no longer necessary and is duplicating logic, the master list handling happens on the `sync()` method.

includes/plugins/class-newspack-newsletters.php
```diff
@@ -1,446 +0,0 @@
...
-	 */
-	public static function init() {
-		\add_action( 'init', [ __CLASS__, 'setup_hooks' ], 15 );
-	}
-
-	/**
-	 * Setup hooks.
-	 */
-	public static function setup_hooks() {
-		\add_filter( 'newspack_newsletters_contact_data', [ __CLASS__, 'normalize_contact_data' ], 99 );
```
miguelpeixe, line 45: This should no longer be applied as a filter inside the newsletters plugin. It should be called immediately after the `newspack_esp_sync_contact` filter and before the `::upsert()` call inside the `::sync()` method:

https://github.com/Automattic/newspack-plugin/blob/45534a692f81b280a5a5dfd83ddcbbe9362a5406/includes/reader-activation/sync/class-esp-sync.php#L93-L103

This improves readability and makes the sync logic more predictable.

tests/unit-tests/newsletters.php
```diff
@@ -1,99 +0,0 @@
...
-use Newspack\Reader_Activation;
-
-/**
- * Tests the Newsletters integration.
- */
-class Newspack_Test_Newspack_Newsletters extends WP_UnitTestCase {
-	/**
-	 * Test the "master list" feature.
-	 */
-	public function test_newsletters_master_list_feature() {
```
miguelpeixe, line 18: Removed because the `add_activecampaign_master_list` method is being deprecated in this PR.

tests/unit-tests/newsletters.php
```diff
@@ -1,99 +0,0 @@
...
-			[ 1, 3 ],
-			Newspack_Newsletters::get_lists_without_active_campaign_master_list( $list_ids ),
-			'Expected to get the same lists back when no master list is set.'
-		);
-	}
-
-	/**
-	 * Contact handling.
-	 */
-	public static function test_newsletters_contact_handling() {
```
miguelpeixe, line 64: Removed because this is redundant to tests already available in `tests/unit-tests/reader-activation-sync.php`, before this refactor.

includes/plugins/class-newspack-newsletters.php
```diff
@@ -1,446 +0,0 @@
...
-				if ( ! isset( $contact['metadata'] ) ) {
-					$contact['metadata'] = [];
-				}
-				$name_fragments                    = explode( ' ', $contact['name'], 2 );
-				$contact['metadata']['First Name'] = $name_fragments[0];
-				if ( isset( $name_fragments[1] ) ) {
-					$contact['metadata']['Last Name'] = $name_fragments[1];
-				}
-			}
-		}
```
miguelpeixe, line 380: This should now live here: https://github.com/Automattic/newspack-newsletters/pull/1628

includes/reader-activation/class-reader-activation.php
```diff
@@ -2063,41 +2059,5 @@ public static function rate_limit_lost_password( $errors, $user_data ) {
...
-				$cleaned_subscriptions_site_url = strtolower( untrailingslashit( str_ireplace( '_[wc_subscriptions_siteurl]_', '', $subscriptions_site_url ) ) );
-				if ( $cleaned_subscriptions_site_url !== $site_url ) {
-					return false;
-				}
-			}
-		}
-
-		return $value;
-	}
 }
```
miguelpeixe, line 2062: Removing this as we should always rely on `Newspack\Reader_Activation\ESP_Sync::can_esp_sync()`.

tests/unit-tests/reader-activation-sync-woocommerce.php
```diff
@@ -58,34 +58,34 @@ public function test_woocommerce_connection_payment_metadata_basic() {
...
+					'sub_end_date'              => '',
+					'billing_cycle'             => '',
+					'recurring_payment'         => '',
+					'next_payment_date'         => '',
+					'total_paid'                => '$' . ( self::USER_DATA['meta_input']['wc_total_spent'] + $order_data['total'] ),
+					'account'                   => self::$user_id,
+					'registration_date'         => $today,
+					'membership_plan'           => '',
+					'membership_start_date'     => '',
+					'membership_end_date'       => '',
```
miguelpeixe, line 88: We are now sending raw keys to be later transformed before syncing.

includes/reader-activation/sync/class-metadata.php
```diff
@@ -0,0 +1,397 @@
...
+
+use Newspack\Reader_Activation\Sync;
+use Newspack\Reader_Activation\ESP_Sync;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Metadata Class.
+ */
+class Metadata extends Sync {
```
leogermani: I don't see why this class needs to extend Sync. Right now the only thing it's using from the parent class is the `log` method... but that's not enough and could be a class on its own... a helper method... or even a trait

includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -0,0 +1,264 @@
...
+use Newspack\Donations;
+use Newspack\WooCommerce_Connection;
+use Newspack\WooCommerce_Order_UTM;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WooCommerce Class.
+ */
+class WooCommerce extends Sync {
```
leogermani: Same as Metadata. I don't see why extend

includes/reader-activation/sync/class-metadata.php
```diff
@@ -0,0 +1,397 @@
...
+/**
+ * Reader Activation Sync Metadata.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Reader_Activation\Sync;
+
+use Newspack\Reader_Activation\Sync;
+use Newspack\Reader_Activation\ESP_Sync;
```
dkoo: Nit: unused `use`

includes/reader-activation/sync/class-metadata.php
```diff
@@ -0,0 +1,397 @@
...
+						if ( ! isset( $metadata[ $key ] ) || empty( $metadata[ $key ] ) ) {
+							$metadata[ $key ] = $value;
+						}
+					}
+				}
+			}
+		}
+
+		// Keys allowed to pass through without prefixing.
+		$allowed_keys = [ 'status', 'status_if_new' ];
```
dkoo, line 368: Do you think this is a more appropriate place to add the `origin_newspack` key which is currently added in the `upsert` method in [Newsletters](https://github.com/Automattic/newspack-newsletters/pull/1602/files#diff-2c51024c343e46470774b97265cded37f57e841b1db7a53a0f5a6f3e93053224R156)? Or maybe in the `ESP_Sync::sync()` method. Just in the interest of consolidating all of the synced contact metadata here.

dkoo (commented):
@miguelpeixe I love the consolidation and reorg of all of these various sync-related features. I don't have any blocking feedback, just a couple of minor nits. I'll let @leogermani handle the follow-up since he reviewed first.

includes/reader-activation/sync/class-metadata.php
miguelpeixe: You're right! I believe at some point during the refactor some more inherited methods were being used but as I worked abstracting it further it became more self-contained.

We talked about it being a trait, but it doesn't really have to, being a static class. We would also lose the possibility of having the consts in `Metadata`.

Removed in d22dd80cf82f03ebe88b42d7f811cc7a0eacaefc

includes/reader-activation/sync/class-woocommerce.php
miguelpeixe: Removed in d22dd80cf82f03ebe88b42d7f811cc7a0eacaefc

includes/reader-activation/sync/class-metadata.php
miguelpeixe: Removed in d22dd80cf82f03ebe88b42d7f811cc7a0eacaefc

includes/reader-activation/sync/class-metadata.php
miguelpeixe: Good call! I thought the `origin_newspack` was only added when creating a new contact, justifying it not being handled here. It's a good idea to move it.

dkoo: I'm not sure what the purpose of `origin_newspack` is—it doesn't seem like we use it in our code at all. Must be related to automations in the ESP?

miguelpeixe: Here's its birth: https://github.com/Automattic/newspack-newsletters/pull/1091

Maybe @leogermani can share more, but I think it's an identifier so we can treat migration cases or other necessary bulk changes without messing with non-Newspack data from an audience.

includes/reader-activation/sync/class-metadata.php
leogermani: I meant the Logger being a trait, so you can call `self::log`, and it could add a prefix to the log based on a static property, so each class would have its own prefix

But no need for that, we have a Logger class already

includes/reader-activation/sync/class-metadata.php
dkoo: After some internal discussion it seems we can drop the use of `origin_newspack` entirely as it's no longer needed, but there's nothing needed to do that in this PR.

miguelpeixe:
@dkoo @leogermani is there anything left to revise here?

leogermani (approved):

PR #3363, opened 2024-08-23
fix(esp-wc-metadata): last payment date & amount handling

adekbadek:
### All Submissions:

Fixes the handling of of "last payment" ESP metadata (`NP_Last Payment Date`, `NP_Last Payment Amount`). 

1. On `trunk`, make a recurring donation as a new reader
2. Call the `\Newspack\WooCommerce_Connection::get_contact_from_customer(new WC_Customer(<user-id>))` in `wp shell` – observe the `NP_Last Payment Date` and `NP_Last Payment Amount` fields are empty 
3. Switch to this branch, call the code again, observe the expected fields are there

### Other information:

---
  - https://app.asana.com/0/0/1207471874427924

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -377,12 +377,11 @@ class WooCommerce_Connection {
 	 * Get data for a customer to sync to the connected ESP.
 	 *
 	 * @param int|\WC_Customer $customer Customer object or customer ID.
-	 * @param \WC_Order|false  $order Order object to sync with. If not given, the last successful order will be used.
 	 * @param bool|string      $payment_page_url Payment page URL. If not provided, checkout URL will be used.
 	 *
 	 * @return array|false Contact data or false.
 	 */
-	public static function get_contact_from_customer( $customer, $order = false, $payment_page_url = false ) {
+	public static function get_contact_from_customer( $customer, $payment_page_url = false ) {
 		if ( ! is_a( $customer, 'WC_Customer' ) ) {
 			$customer = new \WC_Customer( $customer );
 		}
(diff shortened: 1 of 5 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -14,9 +14,10 @@
  */
 class WooCommerce_Connection {
 	/**
-	 * Statuses considered active subscriptions.
+	 * Statuses considered active.
 	 */
 	const ACTIVE_SUBSCRIPTION_STATUSES = [ 'active', 'pending', 'pending-cancel' ];
+	const ACTIVE_ORDER_STATUSES = [ 'processing', 'completed' ];
```
miguelpeixe, line 20: I know it's been like this before, but why are we including the `processing` status here? At this stage, the order could fail and the `order_completed` data event only sends orders with the status `completed`.

miguelpeixe: Nevermind! Just read on Woo docs:
> Processing: Payment has been received (paid), and the stock has been reduced. The order is awaiting fulfillment.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -233,11 +234,10 @@ public static function get_last_successful_order( $customer ) {
 	 *
 	 * @param \WC_Order|int $order WooCommerce order or order ID.
 	 * @param bool|string   $payment_page_url Payment page URL. If not provided, checkout URL will be used.
-	 * @param bool          $is_new Whether the order is new and should count as the last payment amount.
 	 *
 	 * @return array Contact order metadata.
 	 */
-	public static function get_contact_order_metadata( $order, $payment_page_url = false, $is_new = false ) {
+	private static function get_contact_order_metadata( $order, $payment_page_url = false ) {
```
miguelpeixe, line 240: I like that we're deprecating this parameter, but that seems to make "Last Payment Amount" and "Last Payment Date" values whichever order passed to `get_contact_from_customer()`. It's best to query it instead of relying on the passed order.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -368,36 +376,34 @@ public static function get_contact_order_metadata( $order, $payment_page_url = f
...
 		}
 
 		// Get the order metadata.
+		$order_metadata = [];
 		if ( $order ) {
-			$order_metadata = self::get_contact_order_metadata( $order, $payment_page_url, $is_new );
+			$order_metadata = self::get_contact_order_metadata( $order, $payment_page_url );
 		} else {
-			// If the customer has no successful orders, clear out subscription-related fields.
+			// If the customer has no successful orders or if no order is provided???, clear out subscription-related fields.
```
miguelpeixe: That seems like an unreasonable amount of question marks 😄

miguelpeixe (changes requested):
Tested well, but I have some concerns regarding how we're treating the "latest" order:

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
adekbadek: Huh, good call! I was wondering if any of the contact metadata could rely on an order _different_ than the last successful order (or the active subscription), but I don't think so.

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
adekbadek: That was more of a note to self :D Fixed in 93428490db4ebdbaaff954524cc71ae1d4d3fe39

includes/reader-revenue/woocommerce/class-woocommerce-connection.php
miguelpeixe: It used to but, from what I can tell, not since #3086. Turns out the method ended up with a superfluous param.

miguelpeixe (approved):
These methods are looking much cleaner now!

PR #3369, opened 2024-08-26
fix(ras-acc): don't close modal when redirecting

laurelfulford:
### All Submissions:

This PR, along with https://github.com/Automattic/newspack-blocks/pull/1855, stops the modal checkout from fully closing when the last step is redirecting to a different URL, so the process feels smoother.

The testing steps are all in https://github.com/Automattic/newspack-blocks/pull/1855.

1.
2.
3.

### Other information:

### DIFF
--- a/src/reader-activation/index.js
+++ b/src/reader-activation/index.js
@@ -159,7 +159,7 @@ export function openAuthModal( config = {} ) {
 			},
 			content: null,
 			trigger: null,
-			skipClose: false,
+			closeOnSuccess: true,
 		},
 		...config,
 	};
--- a/src/reader-activation/index.js
+++ b/src/reader-activation/index.js
@@ -159,6 +159,7 @@ export function openAuthModal( config = {} ) {
 			},
 			content: null,
 			trigger: null,
+			skipClose: false,
 		},
 		...config,
 	};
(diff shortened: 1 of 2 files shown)


src/reader-activation/index.js
```diff
@@ -159,6 +159,7 @@ export function openAuthModal( config = {} ) {
 			},
 			content: null,
 			trigger: null,
+			skipClose: false,
```
chickenn00dle: This is a nonblocking nit, but the name `skipClose` feels a bit strange to me. It's not really something that is being skipped, but more like something that is just not being done. It might be better to call this something like `closeOnSuccess` or something along those lines with a default value of `true`.

chickenn00dle (approved):
Looks good @laurelfulford!

Left a non-blocking nit, but this is a thumbs up from me.

src/reader-activation/index.js
laurelfulford: Ah thanks @chickenn00dle! I am notoriously bad at naming stuff, this is much nicer 😆 It has been updated in [4ec56a9](https://github.com/Automattic/newspack-plugin/pull/3369/commits/4ec56a97e5ccca5899743a7123f4403b4c2cf897)!

PR #3371, opened 2024-08-27
Fix/payment utm meta fields

adekbadek:
### All Submissions:

Newspack will set the `utm` order meta, but when importing data, a one-to-one relation might be needed. In such a case, a field in the imported data would correspond to a string meta field, instead of the UTM params being stored as serialized values.

In other words, when an order has `utm_campaign` meta field set, but no value at `utm` meta field, the former value should be used for `NP_Payment UTM: campaign` ESP meta field.

Change should be verifiable by reading the tests, but here are the reproduction steps if need be:

1. Make a donation on a page with **no** UTM params in the URL 
1. Update the order meta, setting a `utm_campaign` meta field to some value
2. On `trunk`, observe that `\Newspack\WooCommerce_Connection::get_contact_from_order(<order-id>)` won't return `metadata.NP_Payment UTM: campaign` field
3. On this branch, observe the field is present, with the expected value

### Other information:

---
  - https://app.asana.com/0/0/1207898544020915

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -357,16 +357,16 @@ class WooCommerce_Connection {
 
 		// Clear out any payment-related fields that don't relate to the current order.
 		$payment_fields = array_keys( Newspack_Newsletters::get_payment_metadata_fields() );
+		$utm_meta_field = Newspack_Newsletters::get_metadata_key( 'payment_page_utm' );
+		foreach ( WooCommerce_Order_UTM::$params as $param ) {
+			if ( ! isset( $metadata[ $utm_meta_field . $param ] ) ) {
+				$metadata[ $utm_meta_field . $param ] = '';
+			}
+		}
 		foreach ( $payment_fields as $meta_key ) {
 			$meta_field = Newspack_Newsletters::get_metadata_key( $meta_key );
-			if ( ! isset( $metadata[ $meta_field ] ) ) {
-				if ( 'payment_page_utm' === $meta_key ) {
-					foreach ( WooCommerce_Order_UTM::$params as $param ) {
-						$metadata[ $meta_field . $param ] = '';
-					}
-				} else {
-					$metadata[ $meta_field ] = '';
-				}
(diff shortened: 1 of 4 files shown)


miguelpeixe (approved):
Works as described. The only caveat is that it won't fetch the order UTM if it's not the latest order or the reader has an active subscription, which [will be prioritized](https://github.com/Automattic/newspack-plugin/blob/d68ae864989cd1ca899a93429e2d4977bc22668f/includes/reader-revenue/woocommerce/class-woocommerce-connection.php#L201-L205).

miguelpeixe:
We could change how the UTM is fetched by updating a user meta when an order comes in with a UTM. Then, fetch the user meta instead of the order this logic uses.

adekbadek:
> The only caveat is that it won't fetch the order UTM if it's not the latest order or the reader has an active subscription

I might be misreading, but I thought this metadata _should_ always be sourced from the latest order/sub. The goal is to have the UTMs available in the ESP for targeted campaigns. If a reader donates on a `utm_source=Xmas 2024` page, then donates again on a `utm_source=Summer 2025` page, the latter should be available in the ESP. 

Anyway, this PR does not change this logic, only adds handling of additional order meta to facilitate WC data importing. If the discussion above results in any changes requested, they should be done in a separate PR.

miguelpeixe:
>  If a reader donates on a utm_source=Xmas 2024 page, then donates again on a utm_source=Summer 2025 page, the latter should be available in the ESP.

This is fine and should behave as expected fetching the latest order. The problem is when the reader is a long-time sub and purchases something or donates with UTM params. The existing subscription will not allow the UTM parameters to go in.

Moving to a user meta will still have the intended goal, but ensure that the latest UTM will go in regardless of how this logic decides which order should be prioritized for other reasons.

I understand that the existing logic to prioritize a subscription order is to ensure that the membership status and product name fields reflect their subscription, regardless of which new order is coming in. That does not necessarily translate to how we want all fields to behave. Perhaps instead of user meta, the logic for "active subscription" should be isolated and not replace the parsed order entirely.

adekbadek:
This was rolled into #3362 and released in https://github.com/Automattic/newspack-plugin/releases/tag/v5.5.0

PR #3377, opened 2024-08-28
fix(ras-acc): handle recaptcha submit errors

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207878733150051/f

This PR addresses a problem where Firefox could end up with a strange recaptcha connection error and prevent v2 form submissions. This PR addresses this by adding an error callback and forcing a recaptcha refresh when an error is detected.

Screenshot 2024-08-28 at 13 34 57

Be sure to also checkout the following PRs before testing:
- https://github.com/Automattic/newspack-blocks/pull/1859

1. As logged out user, try to trigger the connection error (see below note)
2. Confirm a recaptcha error appears within the auth or checkout modal and not as an alert
3. Confirm resubmitting the form is successful

Note: This one is a bit tricky to reproduce. I found the best way to get the connection error to appear is to:

1. Use Firefox
4. Enable v2 recaptcha
5. Set recaptcha slider all the way down via Admin console: https://www.google.com/recaptcha/admin
6. Repeatedly go through the auth registration + checkout flow via a donation block, logout, go through the regular auth registration flow via signin button and repeat a handful of times

But truthfully not sure exactly what causes the issue...

### Other information:

### DIFF
--- a/src/other-scripts/recaptcha/index.js
+++ b/src/other-scripts/recaptcha/index.js
@@ -107,9 +107,10 @@ function refreshWidget( el ) {
 /**
  * Render reCAPTCHA v2 widget on the given form.
  *
- * @param {HTMLElement} form The form element.
+ * @param {HTMLElement}   form         The form element.
+ * @param {Function|null} errorHandler Function to handle errors. Optional.
  */
-function renderWidget( form ) {
+function renderWidget( form, errorHandler = null ) {
 	const submitButtons = [
 		...form.querySelectorAll( 'input[type="submit"], button[type="submit"]' ),
 	];
@@ -118,7 +119,7 @@ function renderWidget( form ) {
 	const options = {
 		sitekey: siteKey,
 		size: isInvisible ? 'invisible' : 'normal',
-		isolated: true, // Seems to avoid a "Cannot contact reCAPTCHA" connection error in some browsers.
+		isolated: true,
 	};
 
 	submitButtons.forEach( button => {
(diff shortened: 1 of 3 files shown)


src/other-scripts/recaptcha/index.js
```diff
@@ -118,7 +119,7 @@ function renderWidget( form ) {
 	const options = {
 		sitekey: siteKey,
 		size: isInvisible ? 'invisible' : 'normal',
-		isolated: true, // Seems to avoid a "Cannot contact reCAPTCHA" connection error in some browsers.
+		isolated: true,
```
chickenn00dle, line 122: The connection error still happens so removing the comment. I don't think it hurts to keep the option though so didn't remove. Let me know if you think we can get rid of it though!

src/other-scripts/recaptcha/index.js
```diff
@@ -139,11 +140,24 @@ function renderWidget( form ) {
...
 		};
 
+		const errorCallback = () => {
+			refreshWidget( button );
+			const message = wp.i18n.__( 'There was an error with reCAPTCHA. Please try again.', 'newspack-plugin' );
+			if ( errorHandler ) {
+				errorHandler( message );
+			} else {
+				// eslint-disable-next-line no-alert
+				alert( message );
```
chickenn00dle, line 150: I decided to ignore the eslint rule and trigger an alert here because the scripts default behavior is to alert with a generic connection error message.

chickenn00dle:
I hope its okay @dkoo, but I've assigned you for review since you have the most context with recaptcha!

src/other-scripts/recaptcha/index.js
dkoo: Sounds good to me!

dkoo (approved):
I did run into the "There was an error with reCAPTCHA. Please try again" message once, but after refreshing and trying again it worked!

dkoo:
@chickenn00dle I like how simple this fix ended up being. It doesn't necessarily avoid the issue, but it does create a workaround that seemingly works. And it feels like an edge case anyway so hopefully not many users will be affected.

PR #2835, opened 2024-01-05
fix: alpha fixes

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes some issues resulting from a bad hotfix merge.

### How to test the changes in this Pull Request:

1. Follow testing instructions in #2820 and make sure they work
2. Follow testing instructions in #2733 and make sure they work
3. Re: [this comment](https://github.com/Automattic/newspack-plugin/pull/2835#pullrequestreview-1805563860), enable some but not all billing address fields in **Newspack > Reader Revenue > Donations**. 
4. Complete a purchase of a shippable product, then log into/verify My Account, then visit Addresses.
5. Confirm that when editing the **billing** address, only the fields enabled in Reader Revenue in step 3 are required, and that you can save with empty "optional" fields (but not with empty "required" fields).
6. Confirm that when editing the **shipping** address, all default address fields are still required.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3389, opened 2024-09-04
fix: required billing fields in My Account

dkoo:
### All Submissions:

Fixes a regression that broke the address-editing solution in #2835, and adds a new `is_my_account` hidden billing field to checkouts originating from My Account so that we can determine post-checkout if the checkout originated from My Account (because we can only get the referrer URL from the immediately preceding page, not two pages earlier).

Also fixes a bug where clicking "Renew Now" or "Pay Now" from My Account redirects to the My Account main page instead of allowing you to complete the renewal.

See https://github.com/Automattic/newspack-blocks/pull/1869 too, which uses a fix from this PR.

1. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1869.
2. In **Newspack > Reader Revenue > Donations**, enable some but not all billing address fields.
3. As a reader, purchase a subscription via a modal checkout flow (Donate or Checkout Button blocks). Confirm that the post-checkout thank you page shown here uses the modal checkout template.
4. Log into My Account and visit **Addresses**. Edit your billing address and confirm that the fields here are only required if they're also required in **Newspack > Reader Revenue > Donations** (make sure to submit with empty optional fields to confirm the field validation, too).
5. Also in My Account, visit **My Subscription**  then **Renew Now**. Confirm that you're redirected to a Checkout form and NOT the My Account main page.
6. In the checkout page that loads, confirm that billing address fields are only required if they're also required in **Newspack > Reader Revenue > Donations**. (In the future, we should make the checkout flow from My Account a modal checkout flow to avoid this UI discrepancy.)
7. After completing the renewal, confirm that the "thank you" confirmation page does NOT use the modal checkout template. It should have the full site header and footer, and contain a link back to the main "My Account" page, like so:

Screenshot 2024-09-03 at 6 38 46 PM

8. Smoke test a transaction via modal checkout using an express checkout method such as Apple Pay (see https://github.com/Automattic/newspack-blocks/pull/1573 for testing instructions) and confirm that the post-checkout thank you page shown here uses the modal checkout template.

### Other information:

---
  - https://app.asana.com/0/0/1208207058134668

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -31,6 +31,8 @@ class WooCommerce_My_Account {
 		\add_filter( 'woocommerce_account_menu_items', [ __CLASS__, 'my_account_menu_items' ], 1000 );
 		\add_filter( 'woocommerce_default_address_fields', [ __CLASS__, 'required_address_fields' ] );
 		\add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'required_address_fields' ] );
+		\add_filter( 'woocommerce_get_checkout_url', [ __CLASS__, 'get_checkout_url' ] );
+		\add_filter( 'woocommerce_get_checkout_payment_url', [ __CLASS__, 'get_checkout_url' ] );
 
 		// Reader Activation mods.
 		if ( Reader_Activation::is_enabled() ) {
@@ -353,6 +355,7 @@ class WooCommerce_My_Account {
 		$is_resubscribe_request       = isset( $_REQUEST['resubscribe'] ) ? 'shop_subscription' === \get_post_type( absint( $_REQUEST['resubscribe'] ) ) : false;
 		$is_renewal_request           = isset( $_REQUEST['subscription_renewal'] ) ? true : false;
 		$is_cancel_membership_request = isset( $_REQUEST['cancel_membership'] ) ? true : false;
+		$is_checkout_request          = isset( $_REQUEST['my_account_checkout'] ) ? true : false;
 		// phpcs:enable WordPress.Security.NonceVerification.Recommended
 
 		if (
@@ -361,7 +364,8 @@ class WooCommerce_My_Account {
 			function_exists( 'wc_get_page_permalink' ) &&
 			! $is_resubscribe_request &&
 			! $is_renewal_request &&
-			! $is_cancel_membership_request
+			! $is_cancel_membership_request &&
+			! $is_checkout_request
 			) {
 			global $wp;
 			$current_url               = \home_url( $wp->request );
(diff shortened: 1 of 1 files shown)


adekbadek (changes requested):
> In the checkout page that loads, confirm that billing address fields are only required if they're also required in Newspack > Reader Revenue > Donations.

Should all fields be displayed, but only some marked as required? In my testing, only the fields requried in Newspack > Reader Revenue > Donations are displayed. Not sure if that's how it's supposed to be.

> After completing the renewal, confirm that the "thank you" confirmation page does NOT use the modal checkout template.

In my testing, the confirmation page _does_ use the checkout template:

image

dkoo:
> Should all fields be displayed, but only some marked as required? In my testing, only the fields requried in Newspack > Reader Revenue > Donations are displayed. Not sure if that's how it's supposed to be.

Yes, all billing fields should be displayed in the non-modal checkout page, but only the ones selected in Newspack > Reader Revenue > Donations should be required. (I didn't want to change the look of these flows too much at this point.)

> In my testing, the confirmation page _does_ use the checkout template:

~I can't replicate, but maybe there's some scenario I'm not considering. Did you also check out the corresponding [Blocks PR](https://github.com/Automattic/newspack-blocks/pull/1869)? If yes, and you're still seeing the modal checkout template after renewing from My Account, then what Woo version and payment gateway are you using?~

Aha, something broke with donation subscriptions. Revisiting...

dkoo:
@adekbadek looks like the Blocks plugin has a redundant filter for checkout fields that affects donate checkouts in all contexts (not just modal checkout). This is removing the `is_my_account` hidden field that lets us show the non-modal thank you template after a My Account transaction.

[788faa4](https://github.com/Automattic/newspack-blocks/pull/1869/commits/788faa4069dd38c2fda845238920aa830349458c) should remove that check, so that all fields continue to be shown for Donate products in non-modal checkouts, too.

~The only scenario that isn't fixed is the one mentioned [here](https://github.com/Automattic/newspack-blocks/pull/1869#discussion_r1744526802) (express checkouts via non-modal checkouts started from My Account but for renewals of subscriptions initially purchased via modal checkout), but this is such a specific edge case and one that has proven to be extremely difficult to solve that I don't think it should be a blocker for this fix.~ fixed by 146917f

dkoo:
146917f and [5bbe764](https://github.com/Automattic/newspack-blocks/pull/1869/commits/5bbe76472beebf4d71a64898917316bdadba68da) change the approach to filter the checkout and payment URLs to add a `my_account_checkout` query param which is used to determine if the request started from a My Account page. This seems to be more reliable and also fixes the issue for express checkout methods as well.

adekbadek (changes requested):
>  Edit your billing address and confirm that the fields here are only required if they're also required in Newspack > Reader Revenue > Donations

This didn't work in my testing – WC uses the standard set of required fields when editing an address:

image

image

The checkout behaves as expected, though, with all fields listed but only the ones selected in the RR wizard marked as required:

image

> After completing the renewal, confirm that the "thank you" confirmation page does NOT use the modal checkout template. I

Yes, this works as described now.

dkoo:
@adekbadek doh, 183e4e4b should re-fix for editing billing address.

dkoo:
Any further feedback on this, @adekbadek?

adekbadek (approved):
Sorry for the late reply! Re-tested and it works fine now.

dkoo:
Thanks, @adekbadek! Holding for a Monday release.

PR #3393, opened 2024-09-05
Hotfix/woo subs vs memberships expiration

adekbadek:
### All Submissions:

Alternative solution to the problem originally tackled in #3050, then amended in #3060, #3268, #3338, and #3380. The issue was that a re-activated membership would transition to expired status shortly after reactivation (which happens by buying a new subscription). 

The original solution was acting retroactively on the expired memberships (in a CRON hook, and when reading post status), causing OOM issues on sites with a lot of memberships (10k+). 

The solution proposed takes a proactive approach, unscheduling a membership expiration when a related subscription is reactivated.

1. Checkout the first commit (bbf0992379d57a58de6fe2467eedf409ba98f370), which reverts the previous solution – this way the problem is not addressed in any way
2. Set up a Membership Plan that's tied to a product purchase and set "Subscription length"
3. As a reader, purchase the product 
4. In WP Admin, confirm that the membership was granted
5. Cancel the subscription and observe the membership was cancelled, too
6. Again as a reader, purchase the subscription again
7. Observe that the membership was reactivated, and then set to "Expired" again (this might happen in a quick succession, but you can trace the story in the membership notes)
8. Checkout this branch (the second commit) and repeat the test. Observe that the membership is not expired after the new subscription reactivates it 

### Other information:

---
  - https://app.asana.com/0/0/1208230183010467
  - https://app.asana.com/0/0/1207858308890295

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -57,6 +57,7 @@ class Memberships {
 		add_filter( 'newspack_reader_activity_article_view', [ __CLASS__, 'suppress_article_view_activity' ], 100 );
 		add_filter( 'user_has_cap', [ __CLASS__, 'user_has_cap' ], 10, 3 );
 		add_action( 'wp', [ __CLASS__, 'remove_unnecessary_content_restriction' ], 11 );
+		add_action( 'woocommerce_subscription_status_updated', [ __CLASS__, 'handle_subscription_status_change' ], 10, 3 );
 
 		/** Add gate content filters to mimic 'the_content'. See 'wp-includes/default-filters.php' for reference. */
 		add_filter( 'newspack_gate_content', 'capital_P_dangit', 11 );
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
Cool idea with the proactive approach but I couldn't make it work as the test describes it.

With the first commit checked out I can confirm the note appears in the user membership after a new subscription reactivates the cancelled membership:

> Membership re-activated due to subscription re-purchase (Premium Access, Order 7576). Membership status changed from Active to Expired.

After retesting with the branch I can confirm that the new subscription reaches line `1019` of `includes/plugins/wc-memberships/class-memberships.php` but I still get the note on the membership when it reactivates it:

> Membership re-activated due to subscription re-purchase (Premium Access, Order 7578). Membership status changed from Active to Expired.

I wondered if it could be a timing thing and tested changing this:

```diff
-		if ( $new_subscription_status !== 'active' || ! function_exists( 'wc_memberships' ) ) {
+		if ( $new_subscription_status !== 'pending' || ! function_exists( 'wc_memberships' ) ) {
```

But no dice.

There's also this very confusing cascade of status update when there's a cancellation, which I don't think this PR aims to tackle, but I'd like to flag:

image

miguelpeixe:
I kept looking at this and I understand better now. It attempts to prevent this action after a new subscription reactivates the membership:

> Membership status changed from Active to Expired.

It's different from the other note, which is fired between cancelled -> active from the new subscription.

Is that correct? I got intermittent results. Sometimes it gets fired immediately, sometimes after a minute, and sometimes it doesn't. I'll do more tests and report back.

miguelpeixe:
It seems this all comes down to the `$membership->is_in_active_period()` method. This is used by the `wc_memberships_user_membership_expiry` event trigger and the bug is that it will always return `false` after a membership has been reactivated.

I believe the reason for that is because a subscription-tied membership does not set an expiry date but its cancellation does. Reactivating it doesn't set, nor clear, the previously set expiry date.

It should be fine for the scheduled event to run if the `$membership->is_in_active_period()` returns the correct value. Maybe we can tweak the approach to reset the expiry date once a new subscription comes up?

You can test it by running this in wp shell for a membership with a new activation and reactivation:

```
$user_id = <USER ID>;
$memberships = wc_memberships_get_user_memberships( $user_id );
$memberships[0]->is_in_active_period();
```

It'll be `true` for the fresh membership and always `false` for a reactivated membership.

adekbadek:
Thank you for looking into that! When testing multiple times I did indeed get some inconsistent results. faf56b5b2e77ffa39b9723c422a23dad0a3a5fbb adds an additional expiry-prevention using the `wc_memberships_expire_user_membership` filter. 

The membership log still has an issue, I think this is because of the override introduced in the code here:

image

adekbadek:
I did some more testing to see the `wc_memberships_expire_user_membership` added in faf56b5b2e77ffa39b9723c422a23dad0a3a5fbb would be enough, without the need for `woocommerce_subscription_status_updated` hook usage. I my testing, it is, so I've removed the redundant code in e3bfe2689f473a5d7e165cd089728cd4e28505e7.

miguelpeixe (approved):
Thank you for revising! The approach works by preventing the membership status change but the membership remains with an incorrect "Expiry date". Updating the incorrect data would be preferable as it allows Woo Memberships to run its intended flow with less interference on our end.

Anyway, this is urgent, works, and we can always revise later. I'll look into submitting a bugfix to Woo Memberships so we no longer have to do this ourselves.

miguelpeixe:
Note: patching Woo Membership to reset the expiry date on reactivation properly also fixes the multiple status changes:

image

PR #3396, opened 2024-09-05
feat(ras-acc): make sure modal checkout still works without RAS

laurelfulford:
### All Submissions:

Is a redo of https://github.com/Automattic/newspack-blocks/pull/1866 based on feedback -- I'm not 100% sure about my approach here, so any feedback on that is welcome!

This PR makes it possible to use the modal checkout when RAS isn't enabled, to match the current behaviour on trunk.

The main difference between the RAS-based checkout is that the initial checkout screen will require you to fill our the email address with the other billing details, since it won't be pre-populated from the Sign In/Sign Up screen.

See: 1207817176293825-as-1208164200882316

1. Start on a test site without RAS enabled, or disable it by running `wp option delete newspack_reader_activation_enabled`.
2. Set up three blocks: a regular Checkout Button block, one with a variable product, and a Donate block.
3. On epic/ras-acc in an incognito window, try to run the checkout in any of the blocks; only the initial variation picker will open.
4. Apply this PR and run `npm run build`.
5. In the incognito window, run through each checkout and confirm it works.
6. Enable RAS (run `wp option update newspack_reader_activation_enabled 1` or use the wizard), or re-test this PR on a second test site with RAS enabled.
7. Smoke test the checkout flows as a new reader and as a logged in reader and confirm that you get the Sign In prompts at the appropriate points, and that the checkout works as usual.

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -130,6 +130,7 @@ final class Reader_Activation {
 				'otp_auth_action'       => Magic_Link::OTP_AUTH_ACTION,
 				'otp_rate_interval'     => Magic_Link::RATE_INTERVAL,
 				'account_url'           => function_exists( 'wc_get_account_endpoint_url' ) ? \wc_get_account_endpoint_url( 'dashboard' ) : '',
+				'ras_is_enabled'        => self::is_enabled(),
 			];
 
 			if ( Recaptcha::can_use_captcha() ) {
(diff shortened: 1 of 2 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -130,6 +130,7 @@ public static function enqueue_scripts() {
 				'otp_auth_action'       => Magic_Link::OTP_AUTH_ACTION,
 				'otp_rate_interval'     => Magic_Link::RATE_INTERVAL,
 				'account_url'           => function_exists( 'wc_get_account_endpoint_url' ) ? \wc_get_account_endpoint_url( 'dashboard' ) : '',
+				'ras_is_enabled'        => self::is_enabled(),
```
chickenn00dle: non-blocking nit: Would `is_ras_enabled` be a better name here? It follows the naming convention for other boolean flags.

chickenn00dle (approved):
This works @laurelfulford

There might be some other unnecessary `readerActivation` globals that aren't needed, but I don't think its a blocker since the goal is to simply get checkout working when RAS isn't enabled.

Approving with a non-blocking nit!

includes/reader-activation/class-reader-activation.php
laurelfulford: Ah good call! Fixed in [4a120d0](https://github.com/Automattic/newspack-plugin/pull/3396/commits/4a120d030ccf3d784f2a1a91114b380efe4b434b).

PR #3397, opened 2024-09-06
fix(ras-acc): prevent invalid auth flow errors

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208214895993711/f

This PR addresses two auth modal error cases:

1. clears any auth modal message or error when transitioning to the sign in or success steps since these are the default first and final steps of the auth flow.

2. fixes an issue where recaptcha V2 event listeners would cause OTP submission logic to fire twice, resulting in OTP erroring, and then succeeding in quick succession

https://github.com/user-attachments/assets/268fc5d4-9e0d-4019-bfce-8117fa1ab127

1. As admin, ensure recaptcha v2 is enabled via Newspack > Connections menu
2. As a logged out reader, open the auth modal via the Sign in button in the site header
3. Attempt to create an account that already exists via the modal (via the Create an account button)
4. Confirm an error appears here (can be an account already exists error, or in some cases a recaptcha error)
5. Once you've confirmed the error, select the `Sign in to an existing account` button
6. Confirm the error message is cleared
7. Now attempt to login via OTP by copy/pasting the OTP code you receive by email into the auth modal
8. Confirm you are signed in as expected and the success modal appears with no errors
9. Repeat the above steps, but this time using an account that has a password setup. The only difference should be you do not receive an OTP code and can login with a password on step 7
10. Smoke test the auth modals other buttons and inputs

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1959,7 +1959,7 @@ final class Reader_Activation {
 
 		$user_id = false;
 
-		if ( $existing_user ) {
+		if ( $existing_user && self::is_reader_without_password( $existing_user ) ) {
 			// Don't send OTP email for newsletter signup.
 			if ( ! isset( $metadata['registration_method'] ) || false === strpos( $metadata['registration_method'], 'newsletters-subscription' ) ) {
 				Logger::log( "User with $email already exists. Sending magic link." );
--- a/src/other-scripts/recaptcha/index.js
+++ b/src/other-scripts/recaptcha/index.js
@@ -171,11 +171,11 @@ function renderWidget( form, errorHandler = null ) {
 		} )( jQuery );
 
 		button.addEventListener( 'click', e => {
+			e.preventDefault();
 			// Skip reCAPTCHA verification if the button has a data-skip-recaptcha attribute.
 			if ( button.hasAttribute( 'data-skip-recaptcha' ) ) {
 				callback();
 			} else {
-				e.preventDefault();
 				grecaptcha.execute( widgetId );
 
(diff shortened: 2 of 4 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -1959,7 +1959,7 @@ public static function register_reader( $email, $display_name = '', $authenticat
 
 		$user_id = false;
 
-		if ( $existing_user ) {
+		if ( $existing_user && self::is_reader_without_password( $existing_user ) ) {
```
chickenn00dle, line 1962: We don't want to send magic links to accounts with passwords so the auth form doesn't load into the OTP state for these the next time they attempt to login

src/other-scripts/recaptcha/index.js
```diff
@@ -171,11 +171,11 @@ function renderWidget( form, errorHandler = null ) {
 		} )( jQuery );
 
 		button.addEventListener( 'click', e => {
+			e.preventDefault();
```
chickenn00dle, line 174: Prevents form submission listeners from firing twice. I'm not entirely sure this doesn't cause some strange side effects though, so it would be great to thoroughly smoke test the auth form.

laurelfulford (approved):
This looks good to me, @chickenn00dle! I confirmed that it fixes the issue. 

I also tried a bunch of different approaches, including signing in/signing up when making purchases, resetting passwords, letting the OTP code count down, etc. I _did_ find a couple bugs, but I was able to recreate them on `epic/ras-acc` too, so I filed them in Asana -- there's nothing new here that I found 😄

PR #3402, opened 2024-09-06
fix: close newsletter on complete

laurelfulford:
### All Submissions:

This PR fixes an issue where the newsletter modal wasn't closing on submit when it was triggered by the Sign In button. 

I don't entirely get _why_ this fixes it, since `true` is the default for this config value, but I'm just rolling with it! 

See: 1207817176293825-as-1208232700780031

1. On epic/ras-acc with the Newsletter modal enabled, try running through a sign up, and signing up to at least one newsletter at the end. Note that it doesn't close when the form is finished submitting.
2. Apply this PR and run `npm run build`.
3. Repeat step one, signing up again using the Sign In button and signing up for at least one newsletter. Confirm that the modal closes as expected.
4. Smoke test how the newsletter form works with purchases, specifically testing a donate block, a checkout button block, and one or the other with the after-checkout behaviour set to redirect to a different page to confirm that the modal remains open until the page redirects.

### Other information:

### DIFF
--- a/src/reader-activation-auth/auth-form.js
+++ b/src/reader-activation-auth/auth-form.js
@@ -250,7 +250,6 @@ window.newspackRAS.push( function ( readerActivation ) {
 						callback = ( authMessage, authData ) =>
 							openNewslettersSignupModal( {
 								callback: container.authCallback( authMessage, authData ),
-								closeOnSuccess: true,
 							} );
 					} else {
 						callback = container.authCallback;
--- a/src/reader-activation-auth/auth-form.js
+++ b/src/reader-activation-auth/auth-form.js
@@ -250,6 +250,7 @@ window.newspackRAS.push( function ( readerActivation ) {
 						callback = ( authMessage, authData ) =>
 							openNewslettersSignupModal( {
 								callback: container.authCallback( authMessage, authData ),
+								closeOnSuccess: true,
 							} );
 					} else {
 						callback = container.authCallback;
--- a/src/reader-activation/index.js
+++ b/src/reader-activation/index.js
@@ -159,7 +159,6 @@ export function openAuthModal( config = {} ) {
 			},
 			content: null,
 			trigger: null,
-			closeOnSuccess: true,
 		},
 		...config,
 	};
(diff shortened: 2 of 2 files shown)


chickenn00dle (changes requested):
Good catch @laurelfulford! 

I dug into this a bit and see what the problem is. The `openNewslettersSignupModal` function being used here is different from the readerActivation helper method (where the default is defined and that is used by external components like the checkout block): https://github.com/Automattic/newspack-plugin/blob/f45331fd0786ae7aead00b9f1eaa50dec8ddebf7/src/reader-activation/index.js#L188

This is why we need to re-add the flag here. With that in mind, I think we need to correctly add the `closeOnSuccess` flag to the helper method default config linked above: https://github.com/Automattic/newspack-plugin/blob/f45331fd0786ae7aead00b9f1eaa50dec8ddebf7/src/reader-activation/index.js#L191-L196

and remove it from the `openAuthModal` helper method default config: https://github.com/Automattic/newspack-plugin/blob/f45331fd0786ae7aead00b9f1eaa50dec8ddebf7/src/reader-activation/index.js#L162


Then this is good to go!

chickenn00dle (approved):
Since I'm AFK the rest of the week and the requested changes are pretty straightforward, I'm gonna go ahead and pre-approve this so I don't slow this one down!

laurelfulford:
Thanks @chickenn00dle! 

Unfortunately, moving the `closeOnSuccess: true,` to `openNewslettersSignupModal` in [newspack-plugin/src/reader-activation/index.js](https://github.com/Automattic/newspack-plugin/blob/f45331fd0786ae7aead00b9f1eaa50dec8ddebf7/src/reader-activation/index.js#L162) doesn't fix this for me -- the modal remains open after submitting.

I'm not sure if you're going to see this now, but did it work for you? (I'll grab someone else to take a look if I don't hear back!).

laurelfulford:
Nope, I shouldn't be looking at this so late. I had also removed it from the auth-form.js 🤦‍♀️ 

I think I have it right but I'll still get a second set of eyes before I merge in case I've messed this up somehow -- thanks again, @chickenn00dle 😆

laurelfulford:
Revisiting with fresh eyes, I'm more confident that I got this right (and I retested a bunch of sign up/purchase cases just in case!). I'm going to go ahead and merge 🙂

PR #3403, opened 2024-09-07
fix: expedite email template fix

claudiulodro:
https://github.com/Automattic/newspack-plugin/pull/3341 as a hotfix since the full release was delayed


dkoo (commented):
Changeset looks good!

dkoo (approved):

PR #3353, opened 2024-08-21
feat(esp-sync): sync membership data regardless of subscription

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

- Adds ESP-data syncing for all memberships, regardless of whether or not a membership is linked to a subscription
- Adds new ESP metadata fields:
    - `membership_plan` - the plan name
    - `membership_start_date`
    - `membership_end_date`
- Adds a `Newspack_Newsletters::get_applicable_fields` method, also added in #3352

### How to test the changes in this Pull Request:

1. In the Engagement wizard, "Email Service Provider (ESP) Advanced Settings" section, ensure "Sync contacts to ESP" is toggled on
2. Observe new fields listed: `Membership Plan`, `Current Membership Start Date`, `Current Membership End Date`
3. Toggle them on, and also ensure `Membership Status` is toggled on
4. Create a new memberships, with a new contact – observe it's synced to the ESP, with all the applicable fields

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207471874427936

PR #3411, opened 2024-09-10
feat: remove Woo Membersip sync fields

leogermani:
### All Submissions:

We have decided to push the addition of these fields until we revisit all the fields and maybe change our current schema.

See pamTN9-9St-p2

1. Confirm that sync is still working
2. Confirm Woo Membership fields are not available
3. Confirm Woo Membership fields are not synced

### Other information:

---
  - https://app.asana.com/0/0/1208262205728468


miguelpeixe (changes requested):
I think we can also remove the `membership_saved` listener, which [was added in that PR](https://github.com/Automattic/newspack-plugin/pull/3353/files#diff-78fed9cc2ddfcee2e5179e81bc86fa694d4863cacf30d59500b2bd84da6995a7) is now no longer used.

leogermani:
> I think we can also remove the membership_saved listener, which [was added in that PR](https://github.com/Automattic/newspack-plugin/pull/3353/files#diff-78fed9cc2ddfcee2e5179e81bc86fa694d4863cacf30d59500b2bd84da6995a7) is now no longer used.

Yes I saw that and opted to leave it in, because we'll need them again in a bit and they don't do any harm there. But I can remove them if you feel strongly about it @miguelpeixe

miguelpeixe (approved):
Not strongly at all :)

PR #3414, opened 2024-09-10
fix(esp-sync): sync Connected Account field

dkoo:
### All Submissions:

Restores the syncing of the `NP_Connected Account` field, which was dropped at some point in the past.

1. In WP admin > Engagement > Reader Activation > Advanced Settings, enable the **Connected Account** field for syncing.
2. As a reader, complete a purchase via Woo checkout (modal or not) using a real live Gmail address.
3. Log out, then log in via Google SSO using the Gmail address.
4. In the connected ESP, confirm that contact's metadata has the following:
  - `NP_Registration Method` is `woocommerce`
  - `NP_Connected Account` is `google`
5. In a new session, register a new account via Google SSO using a different Gmail address.
6. Complete another purchase via Woo checkout.
7. In the connected ESP, confirm that contact's metadata has the following:
  - `NP_Registration Method` is `google`
  - `NP_Connected Account` is `google`

### Other information:

---
  - https://app.asana.com/0/0/1208262205728462

### DIFF
--- a/includes/reader-activation/sync/class-metadata.php
+++ b/includes/reader-activation/sync/class-metadata.php
@@ -320,36 +320,42 @@ class Metadata {
 	}
 
 	/**
-	 * Normalizes contact metadata keys before syncing to ESP.
+	 * Add user's registration-related data to the given metadata.
+	 * These won't be included in every sync request, but they might be stored as user meta.
 	 *
-	 * @param array $contact Contact data.
-	 * @return array Normalized contact data.
+	 * @param array $metadata Metadata to add to.
+	 *
+	 * @return array Metadata with registration data added.
 	 */
-	public static function normalize_contact_data( $contact ) {
-		if ( ! isset( $contact['metadata'] ) ) {
-			$contact['metadata'] = [];
+	private static function add_registration_data( $metadata ) {
+		$user = self::has_key( 'account', $metadata ) ? \get_user_by( 'id', self::get_key_value( 'account', $metadata ) ) : false;
+		if ( ! $user ) {
+			return $metadata;
 		}
 
-		$metadata            = $contact['metadata'];
-		$user                = ! empty( $metadata['account'] ) ? \get_user_by( 'id', $metadata['account'] ) : false;
-		$normalized_metadata = [];
-		$raw_keys            = self::get_raw_keys();
-		$prefixed_keys       = self::get_prefixed_keys();
-
-		// Registration-related fields.
-		if ( $user ) {
-			$registration_method             = ! empty( $metadata['registration_method'] ) ? $metadata['registration_method'] : \get_user_meta( $user->ID, Reader_Activation::REGISTRATION_METHOD, true );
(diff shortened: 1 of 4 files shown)


includes/reader-activation/sync/class-metadata.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Newspack\Reader_Activation\Sync;
 
+use Error;
```
miguelpeixe: What's this for?

miguelpeixe (changes requested):
It worked well with `google` as the registration method.

When registering via the modal checkout, the user is created without the `np_reader_registration_method` meta, maybe an existing bug on trunk?

includes/reader-activation/sync/class-metadata.php
```diff
@@ -330,10 +332,25 @@ public static function normalize_contact_data( $contact ) {
...
+			$connected_account               = ! empty( $metadata['connected_account'] ) ? $metadata['connected_account'] : \get_user_meta( $user->ID, Reader_Activation::CONNECTED_ACCOUNT, true );
+			if ( ! empty( $registration_method ) ) {
+				$metadata['registration_method'] = $registration_method;
+			}
+			if ( $connected_account && in_array( $connected_account, Reader_Activation::SSO_REGISTRATION_METHODS ) ) {
+				$metadata['connected_account'] = $connected_account;
+			} elseif ( $registration_method && in_array( $registration_method, Reader_Activation::SSO_REGISTRATION_METHODS ) ) {
+				$metadata['connected_account'] = $registration_method;
+			}
+		}
```
miguelpeixe: Given this method has so many different steps, I think we'd have better readability by running this in a separate method.

miguelpeixe: We should probably do the same with the UTM capturing bit.

includes/reader-activation/sync/class-metadata.php
dkoo: That's a mistake—my VSCode likes to "helpfully" try to insert dependencies if it thinks I'm missing any (such as if I'm typing `error_log`)

includes/reader-activation/sync/class-metadata.php
dkoo: Good idea! 7037b6c4 creates separate methods for both, and also refactors the `registration_method` and `connected_account` checks to use the class's built-in helpers for `has_key()` and `get_key_value()`.

miguelpeixe (changes requested):
Thank you for the revisions! Connected Account and UTM tested well.

The issue I had with the `woocommerce` registration method is no longer reproducible, not sure what happened there.

I did find another issue, with Google oauth. It seems the referrer is not passed along so it's unable to set the registration page and, naturally, the UTM params as well. Do you think it's something we can address in this PR? If not I can approve and we address it separately.

dkoo:
> I did find another issue, with Google oauth. It seems the referrer is not passed along so it's unable to set the registration page and, naturally, the UTM params as well. Do you think it's something we can address in this PR? If not I can approve and we address it separately.

25294c7 should fix the Registration Page and Signup UTM fields for SSO registrations!

miguelpeixe (approved):
Thanks for revising! Working well with Google + UTM!

PR #1648, opened 2022-05-17
fix(reader-revenue): initial order state with total of 0

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Because the order was created as completed, an email was sent right away to the site admin, with a 0 as order total.

### How to test the changes in this Pull Request:

1. Choose "Stripe" as the Reader Revenue platform and make sure your site is available publicly (so Stripe webhook can be called)
2. On `master`, make a donation using the Donate block
3. Observe an email is sent to site admin with no order details and 0 as total
4. Switch to this branch, observe the email sent has order details and correct amount

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3415, opened 2024-09-10
fix(esp-sync): normalize contact data on direct upsert calls

dkoo:
### All Submissions:

#3362 removed `normalize_contact_data` as a filter on the `newspack_newsletters_contact_data` hook. However, this results in `upsert` calls that happen outside of this plugin's `Sync` classes to sync un-normalized metadata.

This PR restores the filter, but this will result in the data from `upsert` calls from the `Sync` classes being normalized twice. I'm open to feedback on how this can be improved, but for now this fixes the issue.

1. On `trunk` in both this plugin and Newsletters, sign up for newsletter lists as a new reader via the Newsletter Subscription Form block.
2. Observe in the Logger output and in the connected ESP contact that the metadata synced upon registration contains raw, un-normalized metadata keys (e.g. `connected_account` instead of `NP_Account` or `current_page_url` instead of `NP_Registration Page`).
3. Check out this branch, repeat with a different email address and confirm that the synced metadata contains only normalized, prefixed keys.

### Other information:

### DIFF
--- a/includes/reader-activation/sync/class-metadata.php
+++ b/includes/reader-activation/sync/class-metadata.php
@@ -35,6 +35,14 @@ class Metadata {
 	public static $keys = [];
 
 	/**
+	 * Initialize hooks.
+	 */
+	public static function init_hooks() {
+		// Normalize contact data on upsert calls directly from Newsletters plugin (e.g. Subscription Form block).
+		add_filter( 'newspack_newsletters_contact_data', [ __CLASS__, 'normalize_contact_data' ] );
+	}
+
+	/**
 	 * Get the metadata keys map for Reader Activation.
 	 *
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Thank you for catching that!

I'll approve because this fixes the regression, but I think we could avoid the double normalization by having a separate filter for the `::subscribe()`, which are the instances in which the Newsletters plugin updates a contact with metadata.

All other cases that use `::upsert()` are list updates, which do not have metadata.

leogermani:
Here's an alternative https://github.com/Automattic/newspack-newsletters/pull/1648

dkoo:
Closing in favor of #1648

PR #3416, opened 2024-09-11
fix: change the current product criteria for sync

leogermani:
### All Submissions:

Updates the criteria to get what is considered the "Current product" for sync purposes

1. Just for testing, make the `get_current_product_order_for_sync` method public
2. Create a user with different sets of purchases, and make sure the method always return the expected result, according to tis description
3. Make sure to test subscriptions, expired subscriptions, one time donations, recurring donations and a user with only a regular purchase (not subscription and not donation)

### Other information:

---
  - https://app.asana.com/0/0/1208262205728470

### DIFF
--- a/includes/reader-activation/sync/class-woocommerce.php
+++ b/includes/reader-activation/sync/class-woocommerce.php
@@ -102,7 +102,7 @@ class WooCommerce {
 			array_keys( \wcs_get_users_subscriptions( $user_id ) ),
 			function( $acc, $subscription_id ) {
 				$subscription = \wcs_get_subscription( $subscription_id );
-				if ( $subscription->has_status( [ 'on-hold', 'cancelled', 'expired' ] ) ) {
+				if ( $subscription->has_status( WooCommerce_Connection::FORMER_SUBSCRIBER_STATUSES ) ) {
 					$acc[] = $subscription_id;
 				}
 				return $acc;
--- a/includes/reader-activation/sync/class-woocommerce.php
+++ b/includes/reader-activation/sync/class-woocommerce.php
@@ -92,7 +92,7 @@ class WooCommerce {
 	private static function get_most_recent_cancelled_or_expired_subscription( $user_id ) {
 		$subcriptions = array_reduce(
 			array_keys( \wcs_get_users_subscriptions( $user_id ) ),
-			function( $acc, $subscription_id ) use ( $product_ids ) {
+			function( $acc, $subscription_id ) {
 				$subscription = \wcs_get_subscription( $subscription_id );
 				if ( $subscription->has_status( [ 'on-hold', 'cancelled', 'expired' ] ) ) {
 					$acc[] = $subscription_id;
@@ -116,7 +116,10 @@ class WooCommerce {
 	 */
 	private static function get_one_time_donation_order_for_user( $user_id ) {
 		$donation_product = Donations::get_donation_product( 'once' );
-		$user_has_donated = wc_customer_bought_product( null, $user_id, $donation_product->get_id() );
+		if ( ! $donation_product ) {
+			return;
+		}
+		$user_has_donated = \wc_customer_bought_product( null, $user_id, $donation_product );
 		if ( ! $user_has_donated ) {
 			return;
 		}
--- a/includes/reader-activation/sync/class-woocommerce.php
+++ b/includes/reader-activation/sync/class-woocommerce.php
@@ -79,7 +79,15 @@ class WooCommerce {
 			return $one_time_donation_order;
 		}
 
-		return false;
+		/**
+		 * Filter the order containing what we consider to be the "Current Product" for a given user when nothing is found.
+		 *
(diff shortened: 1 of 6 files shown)


includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -40,6 +40,118 @@ public static function should_sync_order( $order ) {
...
+	 * @param int $user_id User ID.
+	 *
+	 * @return ?WCS_Subscription A Subscription object or null.
+	 */
+	private static function get_most_recent_cancelled_or_expired_subscription( $user_id ) {
+		$subcriptions = array_reduce(
+			array_keys( \wcs_get_users_subscriptions( $user_id ) ),
+			function( $acc, $subscription_id ) {
+				$subscription = \wcs_get_subscription( $subscription_id );
+				if ( $subscription->has_status( [ 'on-hold', 'cancelled', 'expired' ] ) ) {
```
dkoo: Nit: to keep things consistent, should we use [this constant](https://github.com/Automattic/newspack-plugin/blob/fix/current-product-criteria/includes/reader-revenue/woocommerce/class-woocommerce-connection.php#L19)? e.g.

```suggestion
				if ( ! $subscription->has_status( WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES ) ) {
```

tests/unit-tests/reader-activation-sync.php
```diff
@@ -74,8 +74,6 @@ public function test_can_esp_sync() {
 		Reader_Activation::update_setting( 'mailchimp_audience_id', '123' );
 		$errors = ESP_Sync::can_esp_sync( true );
 		$this->assertNotContains( 'ras_esp_master_list_id_not_found', $errors->get_error_codes(), 'Master list ID is set' );
-
-		$this->assertTrue( ESP_Sync::can_esp_sync(), 'Reader data should be syncable after conditions are met' );
```
dkoo: Did you mean to remove this test?

dkoo (changes requested):
Works as described! A minor nitpick, and a non-blocking question below. Also, do you think it's worth adding a unit test where we mock some users who own various types of products (and some who own none), to test the logic? Or is that overkill?

tests/unit-tests/reader-activation-sync.php
leogermani: Yes, this test won't pass and was already broken in trunk. Because it checks for `class_exists( 'Newspack_Newsletters_Contacts' ) `.

But we can just declare a mock class instead as well...

includes/reader-activation/sync/class-woocommerce.php
leogermani: That was actually intentional, but open for discussion.

My idea is that maybe we don't want to consider all non-active statuses there. Because what we are looking for are former donors/subscribers.

For example, "pending" could represent someone who never actually finished a subscription, right? or even "trash" or "deleted".

btw, in some cases "on-hold" could be the status of a subscription that was never active, it just had a failed payment and never completed...

wdyt?

leogermani:
> Also, do you think it's worth adding a unit test where we mock some users who own various types of products (and some who own none), to test the logic? Or is that overkill?

It's not overkill, I did some drafts but it takes a lot of mocking... and also it would basically require that we write tests for methods such as `get_active_subscriptions_for_user`, `get_most_recent_cancelled_or_expired_subscription` and `get_one_time_donation_order_for_user`... and all of them rely heavily on woocommerce plugins. I just thought it was too much effort for this one PR that needed to be shipped - and since the methods are small and simple enough, I didn't feel it was critical

tests/unit-tests/reader-activation-sync.php
dkoo: I see, up to you! We can go back and add it in a separate PR too if we decide we want it

includes/reader-activation/sync/class-woocommerce.php
dkoo: These are good points! In that case maybe it'd be a good idea to create a constant for `FORMER_SUBSCRIPTION_STATUSES` or something similar so that we can more easily explain why we're looking at these statuses in particular?

dkoo (approved):
I think it's okay to proceed without the extra tests—if we decide we do want them, that can be another PR in the future so they don't block this release.

includes/reader-activation/sync/class-woocommerce.php
leogermani: https://github.com/Automattic/newspack-plugin/pull/3416/commits/e0b2778acb012af23faadbe67b924766e3928429

tests/unit-tests/reader-activation-sync.php
leogermani: https://github.com/Automattic/newspack-plugin/pull/3416/commits/2f1a30551521f28ac7376826917fffc4c46375e4

dkoo (approved):

PR #3421, opened 2024-09-15
Feat IA-Network: Newspack Network wizard navigation

ronchambers:
### All Submissions:

This PR completes the IA Epic task "(NW1) IA Network / Foundational" - incorporating the Newspack Network plugin into the new Information Architecture for the admin menu and Newspack Dashboard.  The PR relies on already merged PRs (https://github.com/Automattic/newspack-plugin/pull/3428, https://github.com/Automattic/newspack-plugin/pull/3432, https://github.com/Automattic/newspack-plugin/pull/3448).  (Also solves Asana task: "(N2) Dashboard: Hide sections if required plugin isn't available").

1. Install/symlink this branch into a wordpress site.
2. Run `composer install` and `npm run build`.
4. Log into wp-admin in your browser.
5. Activate this "Newspack Plugin".
6. Test each sections below:

#### Newspack Network plugin non-activated

With this Newspack Plugin activated, but the Newspack Network plugin **not activated**, verify the admin menu does not show any Network items, and also that the Newspack Dashboard does not show the Network section nor cards.

image

#### Newspack Network plugin activated

Install/activate the Newspack Network plugin.  Upon initial activation, the Network "site role" will be blank.  Verify the Network plugin appears in the Newspack admin area (at position 3.9) with menu title "Network" and globe svg/icon.  Also verify on the Dashboard page, the Network section with "settings" card appears.

image

Then click the Network menu item and verify the "admin header" is shown at the top of the page.

image

#### Newspack Network plugin site role "Node"

Set the Network "site role" to "this is a node in a network".  Verify Network "Node" admin menu items, each menu item has an "Admin Header", and the Newspack Dashboard has "Node" cards. 

image

#### Newspack Network plugin site role "Hub"

Change the Network "site role" to "this site acts as the Network Hub".  Verify Network "Hub" admin menu items, each menu item has an "Admin Header", and the Newspack Dashboard has "Hub" cards.

image

**Subscriptions and Orders: (FIXED) ** Screenshot above is missing Orders.  That is now fixed.  This PR's code has been updated for the new refactored/released Network pages and post_types.  

### Other information:


ronchambers:
Thanks @jaredrethman for the review and [poc branch](https://github.com/Automattic/newspack-plugin/compare/epic/ia...feat/ia-network__jr).  

Seeing the following line in your code helps a lot.  

`includes/class-wizards.php` [line 57](https://github.com/Automattic/newspack-plugin/commit/e928ee2e3b31bace99ed6882307bf56ff331c4bb#diff-2e15e44ce845a52442bc8cd351f44967ac393db7217143db8155cc848d7f3b11R57):
`  'network-nodes'           => new Network(),`

So each Network admin page does not need to be added to the `Wizards::$wizards` array, just one entry into the array is fine?  If that's true, then I totally see how you were able to condense all my separate wizard files into one single wizard :-)

Let me know for sure and I can integrate your POC into my branch.  Thanks!

jaredrethman:
> Seeing the following line in your code helps a lot.
> 
> `includes/class-wizards.php` [line 57](https://github.com/Automattic/newspack-plugin/commit/e928ee2e3b31bace99ed6882307bf56ff331c4bb#diff-2e15e44ce845a52442bc8cd351f44967ac393db7217143db8155cc848d7f3b11R57): ` 'network-nodes' => new Network(),`
> 
> So each Network admin page does not need to be added to the `Wizards::$wizards` array, just one entry into the array is fine? If that's true, then I totally see how you were able to condense all my separate wizard files into one single wizard :-)

💯 , from what I can see `Wizards::$wizards` provides a way to instantiate the Wizard classes and is used in various default (to `Newspack\Wizards`) methods. Having taken a closer look at your code, I now see that there is less code there then I initially assumed. I think what you've done is great and I'm totally fine with you keeping the wizards separate (like you have).

I'm going to pull down your branch and test it out. Will update this PR with a review during the course of today.

ronchambers:
I'm cool either way, but my preference is to go the way you went (one-file). 

I would even go a step further and do the instantiation like this if we can:

```
        self::$wizards = [
                'setup'                   => new Setup_Wizard(),
                ...
                'advertising-sponsors'    => new Advertising_Sponsors(),
        ];

        // This doesn't seem to be needed at all in the $wizards array since it's just using the Admin Header, not a full
        // react page, and it's for a "child" plugin, not the main Newspack plugin.
        new Network();
}
```

Or another idea if we do need to add to the $wizards array...in order to keep the one-file approach is to pass in the type in the constructor.  

```
	self::$wizards = [
		'setup'                   => new Setup_Wizard(),
					....
		'advertising-sponsors'    => new Advertising_Sponsors(),
		'network-nodes'             => new Network( 'nodes' ),
		'network-subscriptions'  => new Network( 'subscriptions' ),
		'network-event-log'       => new Network( 'event-log' ),
	];
}
```

I just really like the one-file approach :-)  But up to you!

ronchambers:
@jaredrethman this is up for review again now that it's "one-file".  I also updated the subscriptions/orders "screens array" based on the recently released Network plugin.  And I update the Dashboard too to completely hide the Network section and cards when the plugin is not installed/active.

includes/wizards/class-network-wizard.php
```diff
@@ -0,0 +1,162 @@
...
+		}
+		
+		// Verify a key was found.
+		if ( empty( $network_key ) ) {
+			return;
+		}
+		
+		// Adjust the network menu attributes.
+		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+		$menu[ $network_key ][0] = 'Network';
```
jaredrethman:
```suggestion
		$menu[ $network_key ][0] = __( 'Network', 'newspack-plugin' );
```

jaredrethman (changes requested):
@ronchambers thanks for the changes and using my one file recommendation. 

In addition to my suggestion below on missed translation, I came across an issue where when adding a Order, Subscription or Nodes, i.e.:
* _/wp-admin/post-new.php?post_type=np_hub_subscriptions_
* _/wp-admin/post-new.php?post_type=np_hub_orders_
* _/wp-admin/post-new.php?post_type=newspack_hub_nodes_

This URLs highlight "Site Role" sub menu item. I was able to avoid this by [removing the relevant hook entirely](https://github.com/Automattic/newspack-plugin/blob/45958bec68f41778e51a7d8822c0a1191139c334/includes/wizards/network/class-network.php#L87) and [then re-adding it](https://github.com/Automattic/newspack-plugin/blob/45958bec68f41778e51a7d8822c0a1191139c334/includes/wizards/network/class-network.php#L171) where I want it.

Screenshot 2024-10-15 at 09 11 56

Small nitpick - [loose comparisons should be avoided](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#clever-code) where possible.

includes/wizards/class-network-wizard.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * Current screen callback to detect Network admin screens.
+	 *
+	 * @return void
+	 */
+	public function current_screen(): void {
+
+		global $current_screen, $plugin_page;
+
+		$set_and_show = function( $slug, $screen_type ) {
```
jaredrethman: @ronchambers what are your thoughts on turning this into a method i.e. `protected function set_and_show_screen( string $slug, string $screen_type ): void`, some motivations below:

- **Readability:** Improves the overall readability of the code. Instead of having to process both the logic of `current_screen()` and an anonymous function at the same time, a developer can simply refer to a clearly named method.
- **Reusability:** Useful if, in the future, we need to set up the screen in different contexts i.e. Taxonomies.
- **Testability:** We can directly unit-test `set_and_show_screen()` in isolation. anonymous functions embedded in another function would otherwise require us to write a more complex and less isolated test case.
- **Debugging and Traceability:**  If something goes wrong, a named method can be traced much more easily by developers.

includes/wizards/class-network-wizard.php
ronchambers: Thanks!  Fixed.

includes/wizards/class-network-wizard.php
ronchambers: Thanks!  Refactored.

ronchambers:
@jaredrethman thanks for the review!  

I fixed the missing translation, I refactored the anonymous display function, and I updated the two loose comparisons to `===`.  

In regards to the missing highlight on the sub menu on `post-new.php` admin pages, this seems to be a "bug/oversight" in wordpress when a CPT is not the first menu item (or maybe some other reason, not exactly sure ;-)  But since this is an issue that already exists within the Network plugin itself for all 3 CPTs, my preference would be to fix it in the Network plugin instead.  Hope that's cool.

jaredrethman (approved):

jaredrethman:
Thanks @ronchambers , fixes look good! I wasn't aware the menu highlighting issue existed - I think fixing it in Newspack Network makes total sense. 

Honestly, I prefer the `remove_action` [approach](https://github.com/Automattic/newspack-plugin/blob/feat/ia-network__jr/includes/wizards/class-network-wizard.php#L84), as its more stable, predictable and readable. But since your code works and I believe the risk is low - I’ll stand down. Code approved!

PR #3425, opened 2024-09-17
feat: command to initialize cron job to slowly backfill CAP term data

eddiesshop:
### All Submissions:

This cron (and by extension, the backfill script it executes) is crucial for our efforts to improve performance across our network of sites. This cron will slowly backfill CAP author term data on an hourly basis. This PR relies on [this separate PR in Co-Authors Plus](https://github.com/Automattic/Co-Authors-Plus/pull/1060) which creates the new backfill command. Once this data backfill is complete, we can then implement the [CAP Author Archive filter](https://lobby.vip.wordpress.com/2017/11/07/co-authors-plus-global-filter/) to greatly improve author archive page performance.

Closes # .

1. Run this query and note the number returned: `SELECT COUNT(*) FROM wp_posts WHERE post_type IN ( 'post' ) AND post_status IN ( 'publish' ) AND post_author <> 0 AND ID NOT IN ( SELECT tr.object_id FROM wp_term_relationships tr LEFT JOIN wp_term_taxonomy tt ON tr.term_taxonomy_id = tt.term_taxonomy_id WHERE tt.taxonomy = 'author' GROUP BY tr.object_id ) ORDER BY ID`
2. Run `wp newspack schedule-co-authors-author-term-backfill`
3. Check that the cron job has been adding by running `wp cron event list` and searching for `newspack_cap_author_term_backfill`
4. Force the cron job to run immediately with: `wp cron event run newspack_cap_author_term_backfill`
5. Re-run the query from Step 1, and confirm that the number has gone down.

### Other information:


includes/cli/class-co-authors-plus.php
```diff
@@ -124,6 +126,46 @@ public function backfill_non_editing_contributor( $args, $assoc_args ) {
...
+	 * @return void
+	 * @throws WP_CLI\ExitException When the command fails.
+	 */
+	public function schedule_author_term_backfill() {
+		WP_CLI::line( '' );
+
+		if ( has_action( self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL ) ) {
+			remove_action(
+				self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL,
+				function () {
```
dkoo: I believe `remove_action` must receive the [exact same callable](https://developer.wordpress.org/reference/functions/remove_action/) that was added by `add_action` and with the same priority in order to remove the callback from the hook. So I'm not sure the anonymous functions are doing what's intended here—we may want to move the functionality to a separate method and call `add/remove_action( self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL, [ __CLASS__, 'method_name' ] )` just to be sure. [Like this](https://github.com/Automattic/newspack-plugin/pull/3429/files#diff-b65e071fdfa99840cde1dd0efb55eb0946499ba61621b3548606badc9e701519R172)

includes/cli/class-co-authors-plus.php
```diff
@@ -124,6 +126,46 @@ public function backfill_non_editing_contributor( $args, $assoc_args ) {
...
+					// Do Nothing.
+				}
+			);
+		}
+
+		add_action(
+			self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL,
+			function () {
+				WP_CLI::runcommand( 'co-authors-plus create-author-terms-for-posts --batched --records-per-batch=50' );
```
dkoo: Can this log errors or success in case we need to check whether a scheduled job succeeded after the fact?

dkoo: Update: looks like it doesn't? When I try to run `wp co-authors-plus create-author-terms-for-posts --batched --records-per-batch=50` at the command line I get an error:

```
Error: 'create-author-terms-for-posts' is not a registered subcommand of 'co-authors-plus'. See 'wp help co-authors-plus' for available subcommands.
```

Running that help command, it looks like the command should be this:

```suggestion
				WP_CLI::runcommand( 'co-authors-plus create-terms-for-posts --batched --records-per-batch=50' );
```

It would be nice to be able to pipe the CLI command's output somewhere—at the very least, the debug log, but better yet using our `newspack_log` action like so:

```php
do_action(
	'newspack_log',
	'newspack_cap_author_terms_backfill',
	$message,
	[
		'type' => 'debug',
		'data' => [], // Other data relevant for the log entry like a timestamp, success/error messages, etc.
		'file' => 'newspack_cap_author_terms_backfill', // Optional flat file to write to in the /tmp directory.
	]
);
```

includes/cli/class-co-authors-plus.php
```diff
@@ -124,6 +126,46 @@ public function backfill_non_editing_contributor( $args, $assoc_args ) {
...
+			}
+		);
+
+		if ( ! wp_next_scheduled( self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL ) ) {
```
dkoo, line 199: Ideally the backfill should clean up after itself by unscheduling the recurring cron job once all posts have been backfilled.

includes/cli/class-co-authors-plus.php
dkoo: I couldn't figure out how to capture the `STDOUT` output from WP_CLI, but if you can figure it out that would make this a very powerful tool.

dkoo (changes requested):
@eddiesshop this is a really cool concept for running a backfill slowly via cron/CLI! I can see this idea being turned into a more abstract tool that would let us schedule any CLI command—something like `wp newspack-schedule-command <CLI command name> --args=<CLI command args in stringified format> --batch-size=50 --interval=hourly` that would let us set the schedule and batch sizes via args.

No need to pursue that idea in this PR, though. For now I see some issues and potential improvements that we should consider for this specific backfill command:

- The code as written isn't executing for me when I run the cron job, because the `add_action( self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL )` is only executed in the callback that gets called on `WP_CLI::add_command`. We need to hook into the action on every process, [like this](https://github.com/Automattic/newspack-plugin/pull/3429/files#diff-b65e071fdfa99840cde1dd0efb55eb0946499ba61621b3548606badc9e701519R29).
- Best practice for cron jobs scheduled by a main plugin is to register a deactivation hook to unschedule the job if the plugin gets deactivated, that way the cron job won't continue to execute (and fail with an error due to the callback not existing) if the plugin is inactive. [Example](https://github.com/Automattic/newspack-plugin/pull/3429/files#diff-b65e071fdfa99840cde1dd0efb55eb0946499ba61621b3548606badc9e701519R30)
- Instead of failing with an error if the job is already scheduled, we could simply unschedule future jobs and reschedule, [like this](https://github.com/Automattic/newspack-plugin/pull/3429/files#diff-b65e071fdfa99840cde1dd0efb55eb0946499ba61621b3548606badc9e701519R185-R186)
- I tried for a bit to figure out how to capture the `STDOUT` output from the `runcommand` calls, but WP_CLI seems to flush its buffer after every `log` or `line` command. If we could somehow capture this output it would make this pattern really powerful as an abstract tool. For example, we could `WP_CLI::runcommand( 'co-authors-plus list-posts-without-terms' );` and unschedule future jobs if there are no more posts to process.

I pushed some suggestions to a [draft PR here](https://github.com/Automattic/newspack-plugin/pull/3429) just for convenience

Another major issue I saw with this implementation (not your fault but due to limitations in the CAP plugin): looking at the [`create-terms-for-posts` command in the CAP plugin](https://github.com/Automattic/Co-Authors-Plus/blob/develop/php/class-wp-cli.php#L55), I don't see support for the `--batched` or `--records-per-batch` args. In fact, the command doesn't seem to support batching at all—if I understand correctly it just always runs the command for all posts 100 at a time. And looking at the output of the command it does seem to query all posts on every run, but skips processing the posts that have terms. So based on that, is the scheduled CLI approach even getting us any performance benefits? It seems like for a site with 1 million posts, every cron job will query 1 million posts. This would make things a lot more complicated, but maybe you could try hooking into the WP query args and adding our own batching/offsetting—WDYT? That would also make this tool useful for scheduling any CLI command, whether or not the command supports batching out of the box.

includes/cli/class-co-authors-plus.php
eddiesshop: For now I'd like to leave this as is. The reason being that, even though the backfill will create the missing data we need, we don't quite understand yet why this data would be missing in the first place. Some more research is needed into this. In the meantime, we can rely on this background process to backfill where needed, and help us with the performance improvement we need.

includes/cli/class-co-authors-plus.php
eddiesshop: In the last commit, we are now logging all output to the Newspack log.

includes/cli/class-co-authors-plus.php
eddiesshop: Thanks for creating the draft PR with the example. I've gone ahead and merged it into this PR.

eddiesshop:
**NOTE:** This PR relies on [this separate PR in Co-Authors Plus](https://github.com/Automattic/Co-Authors-Plus/pull/1060) to function.

includes/cli/class-co-authors-plus.php
```diff
@@ -15,11 +15,64 @@
...
+	/**
+	 * Execute the co-authors-plus create-terms-for-posts CLI command.
+	 */
+	public static function run_cap_cli_command() {
+		if ( method_exists( 'WP_CLI', 'runcommand' ) ) {
+			$result = WP_CLI::runcommand(
+				'co-authors-plus create-author-terms-for-posts --batched --records-per-batch=50',
+				[
+					'launch' => false, // This keeps any formatting that's been set.
+					'return' => 'all', // This ensures we get stdout, stderr, and return code.
```
dkoo: By default, if `WP_CLI::runcommand()` encounters errors during script execution it will exit without reporting the results unless `'exit_error' => false` is set in the options array. I tested by manually adding a `WP_CLI::error()` command in the CAP plugin CLI command, and confirmed that adding this arg lets us log the error message.

```suggestion
					'exit_error' => false, // This allows us to capture any errors that occur during script execution.
					'launch'     => false, // This keeps any formatting that's been set.
					'return'     => 'all', // This ensures we get stdout, stderr, and return code.
```

includes/cli/class-co-authors-plus.php
```diff
@@ -15,11 +15,64 @@
...
+
+			do_action(
+				'newspack_log',
+				$result->return_code,
+				self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL,
+				[
+					'type'      => 0 === $result->return_code ? 'success' : 'error',
+					'stdout'    => $result->stdout,
+					'timestamp' => gmdate( 'c', time() ),
+					'file'      => 'newspack_cap_author_terms_backfill',
```
dkoo: According to our `newspack_log` handler, the first arg is an error or debug code string (to let us filter by code in our logging system) and the second arg is a message string, so I think this would be more useful for logging:

```suggestion
				self::NEWSPACK_SCHEDULE_AUTHOR_TERM_BACKFILL,
				! empty( $result->stderr ) ? $result->stderr : $result->stdout,
				[
					'type' => 0 !== $result->return_code || ! empty( $result->stderr ) ? 'error' : 'success',
					'data' => [
						'stdout'    => $result->stdout,
						'timestamp' => gmdate( 'c', time() ),
					],
					'file' => 'newspack_cap_author_terms_backfill',
```

This results in logs like this for success and error, respectively:

```
[2024-10-04 21:41:22] [335]  [newspack_cap_author_term_backfill][Newspack\CLI\Co_Authors_Plus::run_cap_cli_command][SUCCESS]: Found 0 posts with missing author terms.
Processing page 2.
0 records affected
Updating author terms with new counts
Success: Done! - Data: {
    "stdout": "Found 0 posts with missing author terms.\nProcessing page 2.\n0 records affected\nUpdating author terms with new counts\nSuccess: Done!",
    "timestamp": "2024-10-04T21:41:22+00:00"
}

[2024-10-04 21:41:44] [342]  [newspack_cap_author_term_backfill][Newspack\CLI\Co_Authors_Plus::run_cap_cli_command][ERROR]: Error: A fake error occurred! - Data: {
    "stdout": "Found 0 posts with missing author terms.",
    "timestamp": "2024-10-04T21:41:44+00:00"
}
```

dkoo (changes requested):
@eddiesshop this is working nicely now, thanks for the changes! I confirmed that with the changes from https://github.com/Automattic/Co-Authors-Plus/pull/1060 the command doesn't do unnecessary queries or processing of all posts, so I think it's okay to let it keep running as a background process in that case. However, I think we should wait to merge this PR until that change gets deployed to the CAP plugin, otherwise the command will query all posts on every cron execution.

I also have a couple of minor suggestions for improving the logging in case of errors, but after that I think this is ready.

dkoo (approved):
Thanks for the updates! Approved pending release of the [dependent PR in CAP](https://github.com/Automattic/Co-Authors-Plus/pull/1060).

dkoo:
@eddiesshop looks like https://github.com/Automattic/Co-Authors-Plus/pull/1060 has been merged! Do we know when it might get deployed to the plugin repository?

leogermani:
> Do we know when it might get deployed to the plugin repository?

I'm working on this

PR #3442, opened 2024-09-20
fix: wizards - update type check conditional for `custom_logo`

jaredrethman:
### All Submissions:

Changes introduced in #3433 produced a bug relating to how `custom_logo` and `newspack_footer_logo` theme mods are saved. 

`PHP Fatal error:  Uncaught TypeError: Cannot access offset of type string on string`

This PR fixes this issue by correctly testing incoming values by the correct type.

1. Checkout this branch i.e. `git checkout origin/fix/wizard-custom-logo-save`
2. Navigate to _/wp-admin/admin.php?page=newspack-site-design-wizard#/settings_
3. Scroll to Header and Footer sections. Test the following:
   * You're able to add/update Logos and changes persist between page refreshes.
   * You're able to remove Logos (saving without error) and changes persist between page refreshes.

### Other information:

### DIFF
--- a/includes/wizards/class-setup-wizard.php
+++ b/includes/wizards/class-setup-wizard.php
@@ -520,7 +520,7 @@ class Setup_Wizard extends Wizard {
 				continue;
 			}
 
-			if ( ! empty( $value ) && in_array( $key, $this->media_theme_mods ) ) {
+			if ( ! empty( $value['id'] ) && in_array( $key, $this->media_theme_mods ) ) {
 				$value = $value['id'];
 			}
 			set_theme_mod( $key, $value );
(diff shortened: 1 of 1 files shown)


includes/wizards/class-setup-wizard.php
```diff
@@ -520,7 +520,7 @@ public function api_update_theme_with_mods( $request ) {
 				continue;
 			}
 
-			if ( null !== $value && in_array( $key, $this->media_theme_mods ) ) {
+			if ( '' !== $value && in_array( $key, $this->media_theme_mods ) ) {
```
leogermani: if `$value` needs to be an array there, what about a more explicit check like ` ! empty( $value['id'] ) `

jaredrethman: Thanks for the feedback. I have updated with your recommendation 22c742ab0b0de7ccf315fdb9f02af030b6546f7b

leogermani (approved):

PR #3447, opened 2024-09-25
chore(ia-refactor-get-name): changed abstract to function, added this->name

ronchambers:
### All Submissions:

Changed the `Wizard` class to use a normal `function` instead of `abstract` for `get_name()`.  Added a `$name` property to be overridden by each class.  This also fixes text domain from `newspack` to `newspack-plugin` in many cases.  

Advantages:
* For each class, moved the wizard name to the top of the code instead of in a function farther down in the code.
* Simpler creation of new classes.
* Standardizes return value and text domain to newspack-plugin. 
* Function can still be overridden if need be.

1. Pull branch.
2. Click in web browser on wizard pages to verify name still displays.

### Other information:

### DIFF
--- a/includes/wizards/advertising/class-advertising-display-ads.php
+++ b/includes/wizards/advertising/class-advertising-display-ads.php
@@ -31,6 +31,13 @@ class Advertising_Display_Ads extends Wizard {
 	const OPTION_NAME_GAM_NETWORK_CODE = '_newspack_ads_gam_network_code';
 
 	/**
+	 * The name of this wizard.
+	 *
+	 * @var string
+	 */
+	protected $name = 'Advertising';
+
+	/**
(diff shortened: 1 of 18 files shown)


ronchambers:
On hold.  For this to work I'd have to adjust for i18n:

Child classes:
```
remove:
-       /**
-        * The name of this wizard.
-        *
-        * @var string
-        */
-       protected $name = 'Advertising';

add:
                parent::__construct();
+
+               $this->name = __( 'Advertising', 'newspack-plugin' );
+
```

Wizard class:

```
        public function get_name() {
+               return esc_html( $this->name );
        }
```

ronchambers:
Also, for classes that don't have a constructor, one would have to be added in order to set the $name value.

PR #3448, opened 2024-09-27
Refactor (IA) for body css for react vs partial react pages

ronchambers:
### All Submissions:

This code cleans up the body `scss/css` to one selector that will cover all wizards.  Because the new `epic/ia` wizards all call `add_body_class`, this means all wizards will have `newspack-wizard-page` class on the body.  Also, since all old wizards (which I assume won't be used in `epic/ia`), but still, since they extend from the same `wizards` abstract class they too will get the `newspack-wizard-page` class on the body.

So the `css` selector simply becomes `body.newspack-wizard-page` in order to apply a white background and other padding resets.

In cases where only a "partial react" wizard is needed (like when overriding Newspack Ads, Newspack Network, or anywhere where we only need the `Admin_Header`), there is a `:not(.newspack-wizard-partial-react)` psuedo-class (*renamed* to `:not(.newspack-admin-header`) that can use by ~overriding `add_body_class` on a per-wizard basis~ (*changed* to automatically on the `Admin_Header` trait).

Please see `src/components/src/with-wizard/style.scss` in this PR.

There were tweaks that needed to be applied to get all the wizards standardized.

* `includes/wizards/newspack/class-newspack-dashboard.php` - was missing the the `newspack-wizard-page` body css
* ~`includes/wizards/advertising/class-advertising-sponsors.php` - needed to have the "not" classes `newspack-wizard-page newspack-wizard-partial-react` added.~ (*changed* to automatically on the `Admin_Header` trait).

Let me know if this doesn't make sense or the css class names need to change.  Overall this will make it easier when adding more Wizards from existing plugins like Newspack Network (my current task).  Thanks!

1. Pull this branch.
2. `npm run build` or `npm run watch` 
3.  Use web browser to click through all Wizards.

### Other information:

### DIFF
--- a/includes/wizards/advertising/class-advertising-sponsors.php
+++ b/includes/wizards/advertising/class-advertising-sponsors.php
@@ -203,4 +203,17 @@ class Advertising_Sponsors extends Wizard {
 	
 		return $submenu_file;
 	}
+
+	/**
+	 * Add body class for wizard pages.
+	 * 
+	 * @param string $classes The current body classes.
+	 */
+	public function add_body_class( $classes ) {
+		if ( ! $this->is_wizard_page() ) {
+			return $classes;
+		}
+		$classes .= parent::add_body_class( $classes ) . ' newspack-wizard-partial-react';
+		return $classes;
+	}
 }
--- a/includes/wizards/advertising/class-advertising-sponsors.php
+++ b/includes/wizards/advertising/class-advertising-sponsors.php
@@ -203,17 +203,4 @@ class Advertising_Sponsors extends Wizard {
 	
 		return $submenu_file;
 	}
-
-	/**
-	 * Add body class for wizard pages.
-	 * 
-	 * @param string $classes The current body classes.
-	 */
-	public function add_body_class( $classes ) {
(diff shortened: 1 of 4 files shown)


jaredrethman (commented):
Thanks for this refactor. I like how this simplifies css block in `src/components/src/with-wizard/style.scss`.

It appears `newspack-wizard-partial-react` class is being applied in scenarios where the Admin Header is being used? I'm curious what your thoughts are in:
* Moving the `admin_body_class` hook into the Admin_Header trait?
* Rename `newspack-wizard-partial-react` to `newspack-admin-header`.
* Add `add_filter( 'admin_body_class', [ $this, 'admin_header_body_class' ] );` to Admin_Header constructor.
* Add method:
```php
function admin_header_body_class( $classes ) {
	return $classes . ' newspack-admin-header';
}
```

Mainly this will help with classes that use the trait, by automatically applying the necessary class without needing to override `add_body_class` in `\Newspack\Wizard`.

ronchambers:
Thanks @jaredrethman - great idea!  Done and tested.

jaredrethman (approved):

PR #3450, opened 2024-09-30
fix(oauth): validate ID and prevent early deletion

miguelpeixe:
### All Submissions:

p1727694639714059-slack-C055SRT7WN4

This PR tweaks the custom OAuth transients strategy to prevent the transient ID from being stored as an empty string and prevent the token's early deletion when verifying an existing token for reuse.

Update: dc1840f11b64e8b3b104f0276260446fb593a98b fixes the cleanup method, which is currently using `now()` as time reference. This will use the system's time instead of UTC, which is the stored value. The commit changes it to `utc_timestamp()`. 1e724967e19116347b0e7a0a59a6aae887de0978 tweaks the cron from weekly to hourly. If it's supposed to clear anything older than 30 minutes, it doesn't seem like "weekly" is the best interval.

20fe0094a7ad77eb743fbec27f16cd3b7d4fc6f9 tweaks the logging strategy to add support for `newspack_log`.

#### Empty ID token

1. While on the `release` branch, make sure you have RAS setup with Google OAuth via NM
2. Navigate to a page with the auth form (`/my-account`)
3. Delete the `newspack-cid` cookie (DevTools -> Application -> Cookies)
4. Without refreshing the page, to ensure the cookie is not generated, start the auth flow
5. Via SQL, select the created transient and confirm the ID is an empty string
6. Check out this branch, repeat steps 3-4 and confirm you are unable to continue the flow without the cookie

#### Early deletion 

1. While on the `release` branch, start the Google OAuth flow
2. Before continuing, close the auth popup
3. Confirm via SQL that the transient token is in the database
4. Start the flow again
5. Confirm via SQL that the transient token is now gone
6. Attempt to finish the flow and confirm you get the "Authentication failed" error
7. Check out this branch, repeat steps 1-6 and confirm via the transient token persists and you are able to authenticate
 
### Other information:

### DIFF
--- a/includes/oauth/class-oauth-transients.php
+++ b/includes/oauth/class-oauth-transients.php
@@ -207,7 +207,7 @@ class OAuth_Transients {
 	public static function cleanup() {
 		global $wpdb;
 		$table_name = self::get_table_name();
-		$wpdb->query( "DELETE FROM $table_name WHERE created_at < now() - interval 30 MINUTE LIMIT 1000" ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$wpdb->query( "DELETE FROM $table_name WHERE created_at < utc_timestamp() - interval 30 MINUTE LIMIT 1000" ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 	}
 }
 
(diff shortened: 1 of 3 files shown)


includes/oauth/class-oauth-transients.php
```diff
@@ -165,9 +166,13 @@ public static function delete( $id, $scope ) {
 	 * @return mixed The value if it was set, false otherwise.
 	 */
 	public static function set( $id, $scope, $value ) {
+		if ( empty( $id ) ) {
+			return false;
```
leogermani, line 174: nit: but let's add some logging in this case so we can surface this error?

Maybe add the check and logging in the `generate_csrf_token` method instead

miguelpeixe: Good call, I'll keep both, since `set()` is a public method.

miguelpeixe: Added in 5670bfbb2f1ad0391b1625f5a96b8653cf977b1d

leogermani (approved):

miguelpeixe:
@leogermani could you review this again? I added the new log functionality.

leogermani (approved):
Tested well. Logs also look good

PR #3452, opened 2024-10-01
fix: handle zero cover fee percentage

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1200550061930446/1208449408463805/f

Addresses an issue where the logic for rendering cover fee percentage text causes a fatal when switching subscriptions via My Account.

Screenshot 2024-10-01 at 13 19 54

1. Ensure name your price is enabled
2. As a reader make a recurring donation
3. Visit My account and select the subscription management tab
4. If not already in the subscription view, select the subscription that was just purchased
5. Click "Change price"
6. Choose any subscription, input a new price, and submit
7. On `release` you will get a fatal. On this branch you will not.

Bonus: Go through the normal checkout flow for a donation with cover fees to confirm nothing has broken.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
@@ -270,7 +270,7 @@ class WooCommerce_Cover_Fees {
 	 */
 	public static function get_fee_percentage( $subtotal ) {
 		$flat_percentage = 0;
-		if ( $subtotal > 0 ) {
+		if ( is_numeric( $subtotal ) && (float) $subtotal > 0 ) {
 			$total = self::get_total_with_fee( $subtotal );
 			// Just one decimal place, please.
 			$flat_percentage = (float) number_format( ( ( $total - $subtotal ) * 100 ) / $subtotal, 1 );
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -269,9 +269,12 @@ public static function get_fee_value( $subtotal ) {
 	 * @return string
 	 */
 	public static function get_fee_percentage( $subtotal ) {
-		$total = self::get_total_with_fee( $subtotal );
-		// Just one decimal place, please.
-		$flat_percentage = (float) number_format( ( ( $total - $subtotal ) * 100 ) / $subtotal, 1 );
+		$flat_percentage = 0;
+		if ( $subtotal > 0 ) {
```
leogermani: very NIT: but since we're here, why not also check if this is actually a number?

chickenn00dle: Sure thing! Done in https://github.com/Automattic/newspack-plugin/pull/3452/commits/846bdc2bb47e2eaa4101c3b5af69baddd704bfb4

leogermani (approved):

PR #3462, opened 2024-10-08
feat(ia): settings additional brands

jaredrethman:
### All Submissions:

Adds Newspack Multibranded Site wizard to Newspack Plugin Settings Wizard. The majority of functionality is merely a port of the Multibranded plugin.

1. Checkout this branch and build assets.
2. Ensure Newspack Multibranded Site plugin is installed and active.
3. Navigate to _/wp-admin/admin.php?page=newspack-settings#/additional-brands_
4. Ensure Additional Brands UI loads.
Screenshot 2024-10-21 at 09 14 41
5. Perform Tests. For each test confirm; 1. data loaded is correct 2. Data persists between tab changes and page refreshes:
   - Add New Brand - Click "Add New Brand", input desired configuration and click "Save".
   - Edit an existing brand - Click Screenshot 2024-10-21 at 09 22 32, select "Edit", perform desired edits and click "Save".
   - Delete a brand - Click Screenshot 2024-10-21 at 09 22 32, select "Delete" and confirm.

### Other information:


src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
```diff
@@ -0,0 +1,409 @@
...
+
+import './style.scss';
+
+const { useParams } = Router;
+
+const {
+	themeColors: registeredThemeColors,
+	menuLocations,
+	menus: availableMenus,
+} = window.newspackSettings[ 'additional-brands' ].sections.additionalBrands;
```
miguelpeixe, line 39: The `additional-brands` property is only added if the plugin is active, which will crash here if it's not. It's probably best to add a property to be checked in the frontend so it doesn't render in case the plugin is not active.

includes/wizards/newspack/class-newspack-settings.php
```diff
@@ -90,25 +90,46 @@ public function get_local_data() {
...
 			],
-			'display-settings'  => [
+			'display-settings' => [
 				'label' => __( 'Display Settings', 'newspack-plugin' ),
 			],
-			'additional-brands' => [
-				'label' => __( 'Additional Brands', 'newspack-plugin' ),
-			],
 		];
+		if ( is_plugin_active( 'newspack-multibranded-site/newspack-multibranded-site.php' ) ) {
```
miguelpeixe: This check is a bit fragile. There are cases in which the plugin is installed with a different path, sometimes for testing or other reasons. WDYT of changing this to `defined( 'NEWSPACK_MULTIBRANDED_SITE_PLUGIN_FILE' )`?

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
```diff
@@ -0,0 +1,409 @@
...
+	wizardApiFetch: WizardApiFetch;
+	fetchLogoAttachment: ( brandId: number, logoId: number ) => void;
+} ) {
+	const { brandId = '0' } = useParams();
+	const selectedBrand = brands.find( ( { id } ) => id === Number( brandId ) );
+
+	const [ brand, updateBrand ] = hooks.useObjectState< Brand >( {
+		id: 0,
+		name: 'Brand Name',
+		slug: 'brand-slug',
```
miguelpeixe: If unchanged by the user, this default value will not allow for the brand name to have an auto-generated slug. E.g.: Brand name becoming `brand-name`. This should be an input placeholder so it doesn't intefere with the form input values.

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
```diff
@@ -0,0 +1,409 @@
...
+	upsertBrand: ( brandId: number, brand: Brand ) => void;
+	wizardApiFetch: WizardApiFetch;
+	fetchLogoAttachment: ( brandId: number, logoId: number ) => void;
+} ) {
+	const { brandId = '0' } = useParams();
+	const selectedBrand = brands.find( ( { id } ) => id === Number( brandId ) );
+
+	const [ brand, updateBrand ] = hooks.useObjectState< Brand >( {
+		id: 0,
+		name: 'Brand Name',
```
miguelpeixe: Perhaps it might also be a good idea to have this as a placeholder instead of default input value.

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
```diff
@@ -0,0 +1,409 @@
...
+			{ /* Action Buttons */ }
+			<div className="newspack-buttons-card">
+				<Button
+					disabled={ ! isBrandValid }
+					variant="primary"
+					onClick={ () => upsertBrand( Number( brandId ), brand ) }
+				>
+					{ __( 'Save', 'newspack-plugin' ) }
+				</Button>
+				<Button variant="secondary" href="#/">
```
miguelpeixe: I expected to reach `#/additional-brands` from here.

src/wizards/newspack/views/settings/additional-brands/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+							_logo:
+								brand.meta._logo instanceof Object
+									? brand.meta._logo.id
+									: brand.meta._logo,
+						} ),
+					},
+				},
+			},
+			{
+				onSuccess( result ) {
```
miguelpeixe, line 102: When updating, we should rehydrate the local store with `result` so it's guaranteed to reflect what's in the DB.

When creating a brand with an empty slug, the slug will be automatically generated and returned by the API. Currently, this will only be visible in the UI if I do a page refresh.

miguelpeixe (changes requested):
It's looking great! I've encountered a few issues, which I commented inline.

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
jaredrethman: Nice catch. Used your advice and made it a little more deliberate 0c754a23de6313c500e7458ed5820a084183ceea

includes/wizards/newspack/class-newspack-settings.php
jaredrethman: Updated c256a232f032dc023096f78544691b1a0acae821

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
jaredrethman: Updated 3bd7e88c729bf50189c85b2fe5e580c2d22df6a9

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
jaredrethman: Updated 75833a3372b61af7bf1d0ae927aabc0acc145db0

src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx
jaredrethman: Nice catch, this has been fixed. Implementation is a little verbose but better in the long run. See 78c3e3c71c53bca27b70aaf4f571912bc3112b39 & cfeac38ef8807db035678c1668a643a9454f9a2e

src/wizards/newspack/views/settings/additional-brands/index.tsx
jaredrethman: Can you share some reproducible steps, I'm unable to replicate this, unless I've misunderstood?

Here is a screen-capture of me adding a brand and it appearing in the brand list after clicking "Save" (I also navigate to a different tab and back to confirm the data is persistent).

https://github.com/user-attachments/assets/39e723e7-30e0-40f8-a04c-2ff8f5840e91

Here is another screen capture showing a persistent update:

https://github.com/user-attachments/assets/3c35f889-9388-443a-9134-4914ba072607

The code that takes care of this is:
https://github.com/Automattic/newspack-plugin/blob/75833a3372b61af7bf1d0ae927aabc0acc145db0/src/wizards/newspack/views/settings/additional-brands/index.tsx#L39-L41

jaredrethman:
> It's looking great! I've encountered a few issues, which I commented inline.

@miguelpeixe thanks for the thorough review. I've refactored & fixed most of your feedback. Let me thoughts on https://github.com/Automattic/newspack-plugin/pull/3462#discussion_r1809471915

src/wizards/newspack/views/settings/additional-brands/index.tsx
miguelpeixe: I noticed that when editing the brand, the slug is auto-generated via:

https://github.com/Automattic/newspack-plugin/blob/75833a3372b61af7bf1d0ae927aabc0acc145db0/src/wizards/newspack/views/settings/additional-brands/brand-upsert.tsx#L149-L153

I'm not sure why this did not trigger during my testing, but it is now.

Another example to test what I mean is creating 2 brands with different names but the same slug. The second brand will have its slug modified with the `-2` suffix on the backend, which will not be reflected on the front-end.

Even though this is a very edge case, it shows that the backend can make changes to the data and the front-end should be aware of that.

It seems like changing line 101 of `src/wizards/newspack/views/settings/additional-brands/index.tsx` to use the full `result` should do the trick, but I could be missing something.

jaredrethman: > Even though this is a very edge case, it shows that the backend can make changes to the data and the front-end should be aware of that.

@miguelpeixe 100% agree! I went ahead and ran two test scenarios and fixed both issues I encountered:

1. Two brands exist. Attempt to save one brand using the other brands slug e.g. "slug-one". 
   - This resulted in an error from the backend - "The slug &#8220;slug-one&#8221; is already in use by another term."
   - Solution: Added error message handling to the upsert component - c10e4d73f5e0a0ce8cfeabf943a5d1ef395962c2.
Screenshot 2024-10-23 at 09 42 48

2. One brand exists. A new brand is created using the existing brands slug e.g. "slug-one" (these scenario closely resemble the experiences you've mentioned).
   - This saved without error.
   - After save the brand appears to use the same "slug-one".
   - BUT: when page is refreshed the slug changes to "slug-one-2".
   - Solution: Use `result.slug` when updating cache - 46b001885e6e970db5ad4af4822cd48be05e62f5.
Screenshot 2024-10-23 at 10 05 24

miguelpeixe: Nice catch with the error handling, works nicely!

> Solution: Use result.slug when updating cache

This works for this scenario, but is there a reason not to use the entire backend response object here? The UI has to match the full response, controlling it by property adds some maintainability burden for future developments.

jaredrethman: > but is there a reason not to use the entire backend response object here

No particular reason. I was trying to avoid a bigger footprint in changes from [original implementation](https://github.com/Automattic/newspack-multibranded-site/blob/1ddbba3899889320f1cf4f5b21cd939865906bb8/src/admin/views/brands/index.js#L62). I have updated to use entire `result` d95e89daf575711299af7cfe6455460ad7ecd03e

miguelpeixe: Ah, I see! Thank you for clarifying and revising the approach.

miguelpeixe (changes requested):
Thank you for all the revisions! It's looking great!

One small adjustment is needed for the new error handling. It's not getting cleared:

After receiving an error due to duplicate slugs, I adjust my slug name and click save. If edit a brand - the same or any other - the error message is still there.

jaredrethman:
> Thank you for all the revisions! It's looking great!
> 
> One small adjustment is needed for the new error handling. It's not getting cleared:
> 
> After receiving an error due to duplicate slugs, I adjust my slug name and click save. If edit a brand - the same or any other - the error message is still there.

This shouldn't be happening, it should be clearing automatically after a successful subsequent request. I can reproduce, looking into why this is happening, will revert shortly.

jaredrethman:
> > Thank you for all the revisions! It's looking great!
> > One small adjustment is needed for the new error handling. It's not getting cleared:
> > After receiving an error due to duplicate slugs, I adjust my slug name and click save. If edit a brand - the same or any other - the error message is still there.
> 
> This shouldn't be happening, it should be clearing automatically after a successful subsequent request. I can reproduce, looking into why this is happening, will revert shortly.

I was mistaken, at one stage this one the plan - at the time it wasn't feasible. Anyway - I have added code that will reset the error on location change 920a27a43909dc3bcbc4a5d55ca98c9ef5e1c685

miguelpeixe (approved):
Thanks for the revisions, Jared! Working great now 💯

PR #3465, opened 2024-10-08
feat: add prop to conditionally trigger initial fetch

jaredrethman:
### All Submissions:

Adds a prop to the `<Wizard />` component to prevent the state resolver `*getWizardAPIData` from firing. This avoids prematurely setting loading state variables (`isLoading` and `isQuietLoading`) to `false`.

_Why is this needed?_

Setting `isQuietLoading` to `false` this way overrides any loading status triggered by child components via Redux, which may cause issues with the expected loading state.

1. Checkout this branch and build assets `git checkout origin/feat/ia-wizard-initial-fetch && npm run build`
2. Navigate to _/wp-admin/admin.php?page=newspack-settings#/_
3. Confirm the page loads with loading screen:
Screenshot 2024-10-08 at 22 13 51

### Other information:

### DIFF
--- a/src/components/src/wizard/index.js
+++ b/src/components/src/wizard/index.js
@@ -34,14 +34,15 @@ const { HashRouter, Redirect, Route, Switch } = Router;
 
 /**
  * @typedef  {Object}     WizardProps
- * @property {string}     headerText            The header text.
- * @property {string}     [subHeaderText]       The sub-header text, optional.
- * @property {string}     [apiSlug]             The API slug, optional.
- * @property {string}     [className]           CSS classes, optional.
- * @property {any[]}      sections              Array of sections.
- * @property {boolean}    [hasSimpleFooter]     Indicates if a simple footer is used, optional.
- * @property {() => void} [renderAboveSections] Function to render content above sections, optional.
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Not sure I fully understand why, but it works! 😆

jaredrethman:
> Not sure I fully understand why, but it works! 😆

😄 It's definitely a hack, because I don't fully understand why the below exists:
https://github.com/Automattic/newspack-plugin/blob/0b7f37e2b388feb68405734677b0d0ab18f92528/src/components/src/wizard/index.js#L48-L50

In the below `actions.finishLoadingData();` is always called. 
https://github.com/Automattic/newspack-plugin/blob/0b7f37e2b388feb68405734677b0d0ab18f92528/src/components/src/wizard/store/index.js#L127-L135

Resulting in fetch status's being `false`
https://github.com/Automattic/newspack-plugin/blob/0b7f37e2b388feb68405734677b0d0ab18f92528/src/components/src/wizard/store/index.js#L40

Needed an easy way to opt out of setting fetching status to `false` (inside `getWizardAPIData`), which normally fires after wizards api calls are made.

PR #3469, opened 2024-10-09
fix(oauth): start php session if no session available

adekbadek:
### All Submissions:

In WordPress, the PHP session might not be always started. This change ensures that it always will be.

I was not able to reproduce the conditions under which a session would not be started. A smoke test should be enough.

### Other information:

### DIFF
--- a/includes/oauth/class-google-login.php
+++ b/includes/oauth/class-google-login.php
@@ -123,7 +123,7 @@ class Google_Login {
 
 		if ( ! wp_verify_nonce( sanitize_text_field( $_GET[ self::AUTH_CALLBACK ] ), self::AUTH_CALLBACK ) ) {
 			/* translators: %s is a unique user id */
-			sprintf( __( 'Nonce verification failed for id: %s', 'newspack-plugin' ), OAuth::get_unique_id() );
+			self::handle_error( sprintf( __( 'Nonce verification failed for id: %s', 'newspack-plugin' ), OAuth::get_unique_id() ) );
 			wp_die( esc_html__( 'Invalid nonce.', 'newspack-plugin' ) );
 			return;
 		}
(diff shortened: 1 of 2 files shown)


leogermani:
Interesting! If this solves the empty ID issues it will be amazing!

I think it's cleaner if you do just

```
if ( session_status() !== PHP_SESSION_ACTIVE ) {
session_start();
}
$id = session_id()

```

I've seen some weird behaviors sometimes when you call `session_id()` before `session_start` that it actually prevents a session from being started.. ( I think this happens on CLI though).. but anyway, I think that's the most PHPy way of doing it

adekbadek:
@leogermani – done! Thanks for the tip

adekbadek:
ca27f36b622309daf1ca66ba43fc202961ebbb9b adds the uid logging to the "Nonce verification failed…" message. This way we'll be able to see if nonces fail in that no-uid case.

includes/oauth/class-google-login.php
```diff
@@ -122,7 +122,8 @@ public static function oauth_callback() {
 		}
 
 		if ( ! wp_verify_nonce( sanitize_text_field( $_GET[ self::AUTH_CALLBACK ] ), self::AUTH_CALLBACK ) ) {
-			self::handle_error( __( 'Nonce verification failed.', 'newspack-plugin' ) );
+			/* translators: %s is a unique user id */
+			sprintf( __( 'Nonce verification failed for id: %s', 'newspack-plugin' ), OAuth::get_unique_id() );
```
leogermani: Looks like you accidentally removed the call to `self::handle_error

leogermani (changes requested):

includes/oauth/class-google-login.php
adekbadek: Good catch! Added back in 2b3a8b9e1cda7d6898c293e195c9d804259d5e87

leogermani (approved):

PR #3471, opened 2024-10-11
fix(auth): simplify pending checkout

miguelpeixe:
### All Submissions:

https://github.com/Automattic/newspack-blocks/pull/1896 changes how the cart generation is handled for anonymous sessions. The cart is expected to be added to the session before authentication, allowing for this logic to be simplified.

The only stored value should now be the redirect URL so the checkout state can be recovered on `?checkout=1`.

Read more at https://github.com/Automattic/newspack-blocks/pull/1896

### Other information:

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -66,7 +66,11 @@ class Donations {
 	public static function init() {
 		self::$donation_product_name = __( 'Donate', 'newspack-plugin' );
 
-		add_action( 'wp_loaded', [ __CLASS__, 'process_donation_form' ], 99 );
+		// Process donation request.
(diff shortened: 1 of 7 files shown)


chickenn00dle (changes requested):
Looks good for the most part. But FYI, we were using the `isPendingCheckout` method here as well to flag that content gated content shouldn't be reloaded during checkout: https://github.com/Automattic/newspack-plugin/blob/bfab5b77db7ce142c8649508cadf9c975c3c3442/includes/plugins/wc-memberships/class-memberships.php#L773

This will need to be updated.

miguelpeixe:
Thank you for catching that! Updated in 09454248fe30b188cbde642530f3ed5c86117f35

chickenn00dle (approved):
Great! Thanks for the update. Looks good!

PR #3476, opened 2024-10-14
fix(ras-acc): replace email param with secret for magic links

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208379852549384/f

This PR removes the email param from magic links and instead uses a random secret which is stored as user meta when the magic link is generated.

Screenshot 2024-10-14 at 16 25 38

Note: With these changes the login link will no longer be good after logging in via the auth flow. I'm not sure if this is expected, but I imagine these links should not work permanently. Let me know if this is not the case and the secret should be stored long term.

- As a logged out reader, attempt to login via the auth modal
- In the OTP email, verify the magic link no longer contains an email address as a param
- Click the magic link and verify you are logged in
- Log out
- Click the magic link in the same email once again and verify you are not able to be authenticated this time

### Other information:

### DIFF
--- a/includes/class-magic-link.php
+++ b/includes/class-magic-link.php
@@ -65,7 +65,7 @@ final class Magic_Link {
 
 		/** Replace Newspack Newsletters Verification Email */
 		\add_filter( 'newspack_newsletters_email_verification_email', [ __CLASS__, 'newsletters_email_verification_email' ], 10, 3 );
-		\add_action( 'wp_logout', [ __CLASS__, 'clear_user_data' ], 10, 1 );
+		\add_action( 'wp_logout', [ __CLASS__, 'clear_user_tokens' ], 10, 1 );
 	}
 
 	/**
@@ -275,33 +275,6 @@ final class Magic_Link {
 	}
 
 	/**
-	 * Clear user secret.
-	 *
-	 * @param \WP_User|int $user User or user ID to clear tokens for.
-	 */
-	public static function clear_user_secret( $user ) {
-		$user_id = $user instanceof \WP_User ? $user->ID : $user;
-		\delete_user_meta( $user_id, self::USER_SECRET_META );
-
-		/**
-		 * Fires after user secret is cleared.
-		 *
-		 * @param \WP_User $user User for which tokens were cleared.
-		 */
-		do_action( 'newspack_magic_link_user_secret_cleared', $user );
-	}
-
-	/**
-	 * Clear all user data.
-	 *
-	 * @param \WP_User|int $user User or user ID to clear tokens for.
-	 */
-	public static function clear_user_data( $user ) {
-		self::clear_user_tokens( $user );
-		self::clear_user_secret( $user );
-	}
-
-	/**
 	 * Get a random OTP code.
 	 *
 	 * @return string Random OTP code.
@@ -415,7 +388,11 @@ final class Magic_Link {
 	 * @return string
 	 */
 	public static function generate_secret( $user ) {
-		$secret = str_replace( '-', '', wp_generate_uuid4() );
+		$secret = \get_user_meta( $user->ID, self::USER_SECRET_META, true );
+		if ( ! empty( $secret ) ) {
+			return $secret;
+		}
+		$secret = wp_hash( $user->user_email );
 		\update_user_meta( $user->ID, self::USER_SECRET_META, $secret );
 		return $secret;
 	}
@@ -730,9 +707,6 @@ final class Magic_Link {
 		// Authenticate the reader.
 		Reader_Activation::set_current_reader( $user->ID );
 
-		// Clear user secret after login.
-		self::clear_user_secret( $user );
-
 		/**
 		 * Fires after a reader has been authenticated via magic link.
 		 *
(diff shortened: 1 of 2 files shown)


includes/class-magic-link.php
```diff
@@ -379,6 +407,19 @@ public static function generate_token( $user ) {
...
 
+	/**
+	 * Generate secret.
+	 *
+	 * @param \WP_User $user User to generate the secret for.
+	 *
+	 * @return string
+	 */
+	public static function generate_secret( $user ) {
+		$secret = str_replace( '-', '', wp_generate_uuid4() );
```
chickenn00dle: Seems its possible for `wp_generate_uuid4` to generate duplicates in some rare instances, but given we are only storing these secrets between magic link creation and login and at most until logout, I think it should be fine.

But let me know if you think we should use something else here!

includes/class-magic-link.php
```diff
@@ -1153,7 +1197,7 @@ public static function newsletters_email_verification_email( $email, $user, $url
 		$verification_url   = \add_query_arg(
 			[
 				'action'   => self::AUTH_ACTION,
-				'email'    => urlencode( $user->user_email ),
+				'email'    => self::generate_secret( $user ),
```
chickenn00dle: I'm not sure how to trigger `newsletters_email_verification_email` to test in the UI. Would love some direction on this.

includes/class-magic-link.php
```diff
@@ -729,39 +772,41 @@ function () {
...
-			$email = \sanitize_email( $_GET['email'] );
-			if ( $email ) {
-				$user = \get_user_by( 'email', $email );
-				if ( ! $user ) {
+			$secret = filter_input( INPUT_GET, 'secret', FILTER_SANITIZE_STRING );
+			if ( $secret ) {
+				$users = \get_users(
+					[
+						'meta_key'   => self::USER_SECRET_META,
+						'meta_value' => $secret, // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
```
chickenn00dle: Not sure if there is a more performant way to handle this. Let me know!

includes/class-magic-link.php
miguelpeixe: This is Newspack Newletters' email verification for managing newsletters subscriptions. Under RAS this verification is automatically done when the reader sets a password or connects via magic link/otp.

You can test it by manually removing the newsletter verification in wp shell: `wp user meta delete {user_id} newspack_newsletters_email_verified`. Visit "My Account -> Newsletters" and click to verify.

includes/class-magic-link.php
miguelpeixe: With a rotating secret, I also think this is more than enough entropy.

includes/class-magic-link.php
chickenn00dle: Thanks! Just tested and seems to be working. Also noting I added a quick small commit to fix updating the key here: https://github.com/Automattic/newspack-plugin/pull/3476/commits/6e813075b6c7d6e608fa15f76cab0ce5c85bc539

miguelpeixe:
With the proposed approach only the last magic link is going to work because the user meta value is replaced every token request. We currently support [5 tokens](https://github.com/Automattic/newspack-plugin/blob/bfe0b8769856f26c57cb82dbed058e1474229e39/includes/class-magic-link.php#L353-L354) and the reader should be able to log in from any of the last 5.

Given this is for obfuscating the reader's email from the link, there isn't a security feature by rotating it, so I suggest having a single permanent token created once. This increases our chance of collision with uuid, so we can use `wp_hash( $user_email )` instead, which ensures uniqueness and is safely salted.

miguelpeixe:
> We currently support [5 tokens](https://github.com/Automattic/newspack-plugin/blob/bfe0b8769856f26c57cb82dbed058e1474229e39/includes/class-magic-link.php#L353-L354) and the reader should be able to log in from any of the last 5.

We should add a unit test for that 😄

chickenn00dle:
Thanks for the review @miguelpeixe!

> Given this is for obfuscating the reader's email from the link, there isn't a security feature by rotating it, so I suggest having a single permanent token created once. This increases our chance of collision with uuid, so we can use wp_hash( $user_email ) instead, which ensures uniqueness and is safely salted.

Alright. That makes sense. I've updated to persist the secret in https://github.com/Automattic/newspack-plugin/pull/3476/commits/8bdcb1b52650a6340fcb508c064715f8cd95e3d7

> We should add a unit test for that 😄

I did not add a unit test for this specific case, but did get one added for the `generate_secret` method added in this PR: https://github.com/Automattic/newspack-plugin/pull/3476/commits/129d8d4eca8396f41e899ace26e92c256ec4368d

chickenn00dle:
> We should add a unit test for that 😄

Nevermind, I went ahead and added a unit test for the 5+ tokens case: https://github.com/Automattic/newspack-plugin/pull/3476/commits/7f505def12990af323850ac768e58fb2f116cce7

includes/class-magic-link.php
```diff
@@ -729,39 +754,40 @@ function () {
...
+					]
+				);
+				if ( empty( $users ) || count( $users ) > 1 ) {
```
miguelpeixe: Non-blocking performance tweak:

```suggestion
				$user_query = new \WP_User_Query(
					[
						'meta_key'    => self::USER_SECRET_META,
						'meta_value'  => $secret, // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
						'number'      => 1,
						'count_total' => false,
					]
				);
				$users = $user_query->get_results();
				if ( empty( $users ) ) {
```

miguelpeixe (approved):
Thank you for adding unit tests! It's working great. Left a non-blocking suggestion for performance.

includes/class-magic-link.php
chickenn00dle: Thanks for the suggestion! Updated in https://github.com/Automattic/newspack-plugin/pull/3476/commits/81af75c89b982391bd23766d67ede80822d4ea10

PR #3479, opened 2024-10-16
feat(ia): network settings

ronchambers:
### All Submissions:

This PR moves "Settings" screens to tabs.  ( The previously merged "Foundational PR" is here: https://github.com/Automattic/newspack-plugin/pull/3421  ). Much of this new PR's code has been refactored based on newer techniques we're now using for these types of wizards.

The changes that need review are:

#### When the Site Role is "Node", then the "Node Settings" screen is moved to a tab.  Attached.

epic-ia-network-settings-node-settings

#### When the Site Role is "Hub", then the "Distributor Settings" screen is moved to a tab.  Attached.

epic-ia-network-settings-distributor-settings

Note: if you see that any of the 3 CPTs (Nodes, Subscriptions, Orders) are not highlighting the correct submenu items, this is fixed in the Network plugin itself, so don't worry about that here.  ( See Network PR: https://github.com/Automattic/newspack-network/pull/146 ) 

1. Pull branch and make sure you have the Network plugin active.
2. Use wp-admin in your browser to click the different menu items in the Network menu verifying the admin header shows up and the correct submenu item is highlighted.
3. Do this twice: once as a Network "Node" and once as a Network "Hub".  
4. Verify the Settings pages (Distributor and Node Settings) are in tabs. 

### Other information:


includes/wizards/class-network-wizard.php
```diff
@@ -55,83 +54,189 @@ public function __construct() {
...
-		if ( ! empty( $current_screen->post_type )
-			&& ! empty( $this->admin_screens['post_type'][ $current_screen->post_type ] )
-			&& false === $current_screen->is_block_editor ) {
-			$this->slug = $current_screen->post_type;
-			$this->screen_type = 'post_type';
-			return $this->enable_display();
+		$sanitized_action    = sanitize_text_field( $_GET['action'] ?? '' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$sanitized_page      = sanitize_text_field( $_GET['page'] ?? '' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$sanitized_post_type = sanitize_text_field( $_GET['post_type'] ?? '' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$sanitized_post_id   = sanitize_text_field( $_GET['post'] ?? '' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
jaredrethman:
```suggestion
		$sanitized_action    = filter_input( INPUT_GET, 'action', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
		$sanitized_page      = filter_input( INPUT_GET, 'page', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
		$sanitized_post_type = filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
		$sanitized_post_id   = filter_input( INPUT_GET, 'post', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
```

ronchambers:
In preparation for https://github.com/Automattic/newspack-plugin/pull/3563

I need to add a tweak to this PR:
```
+               parent::__construct();
```

ronchambers:
@jaredrethman small tweak: [90e723a: change individual hooks to use parent construct instead](https://github.com/Automattic/newspack-plugin/pull/3479/commits/90e723ac0b0417a62b52502a10622109a803a80c) committed above.  

This will fix the issue with https://github.com/Automattic/newspack-plugin/pull/3563 once both PRs are merged.

includes/wizards/class-network-wizard.php
```diff
@@ -55,93 +61,203 @@ public function __construct() {
...
+		} elseif ( 'post-new.php' === $pagenow && 'newspack_hub_nodes' === $sanitized_post_type ) {
+			// add new node screen: post-new.php?post_type=newspack_hub_nodes .
+			// note: assumes non-block editor, otherwise we need to not set this.
+			$screen_slug = $sanitized_post_type;
+		} elseif ( 'post.php' === $pagenow && 'edit' === $sanitized_action && 'newspack_hub_nodes' === get_post_type( $sanitized_post_id ) ) {
+			// edit node screen: post.php?post={ID}&action=edit
+			// note: assumes non-block editor, otherwise we need to not set this.
+			$screen_slug = 'newspack_hub_nodes';
 		}
+
```
jaredrethman: Readability improvement/recommendation. Below uses PHP8 `match` expression, switch can also be used.
```suggestion
		$screen_slug = match ( true ) {
			'admin.php' === $pagenow && isset( $this->admin_screens[ $sanitized_page ] ) 
				=> $sanitized_page,
			'edit.php' === $pagenow && isset( $this->admin_screens[ $sanitized_post_type ] ) 
				=> $sanitized_post_type,
			'post-new.php' === $pagenow && 'newspack_hub_nodes' === $sanitized_post_type 
				=> $sanitized_post_type,
			'post.php' === $pagenow && 'edit' === $sanitized_action && 'newspack_hub_nodes' === get_post_type( $sanitized_post_id ) 
				=> 'newspack_hub_nodes',
			default => '',
		};
```

includes/wizards/class-network-wizard.php
```diff
@@ -55,93 +61,203 @@ public function __construct() {
...
-		$this->admin_header_init( [ 'title' => $this->get_name() ] );
+		if ( is_callable( $is_node_function ) && call_user_func( $is_node_function ) ) {
+			$site_role = 'node';
+		} elseif ( is_callable( $is_hub_function ) && call_user_func( $is_hub_function ) ) {
+			$site_role = 'hub';
+		} else {
+			$site_role = '';
+		}
+
+		return $site_role;
```
jaredrethman: This can be simplified.
```suggestion
		static $site_role;
		if ( isset( $site_role ) ) {
			return $site_role;
		}
		$site_role = Newspack_Network_Site_Role::get();
		return $site_role;
```
Below is an even simpler method, slightly less readable:
```php
public static function get_site_role() {
	static $site_role = null;
	$site_role ??= Newspack_Network_Site_Role::get();
	return $site_role;
}
```

ronchambers:
@jaredrethman  - this has been successfully retested after the "hide notifications" was merged.  Please review when you can.  Thanks!

jaredrethman (approved):
@ronchambers your code works well, I can see you put a lot of effort into this one. 

I made a few optional suggestions. Another thing I noticed is high usage of `is_callable` to protect against inactive Network plugin, is this necessary considering the `__construct` method has this guard clause:
```php
if ( ! is_plugin_active( 'newspack-network/newspack-network.php' ) ) {
	return;
}
```

includes/wizards/class-network-wizard.php
ronchambers: Thanks!  Updated and resolved.

includes/wizards/class-network-wizard.php
ronchambers: Thanks!  Updated and resolved.

includes/wizards/class-network-wizard.php
ronchambers: Thanks for sharing `??=` that's very cool.  I do like the simplified solutions but I have two concerns.  The first concern about not using `is_callable` I'll share in the main comment thread here in a minute.  The second concern is the `Newspack_Network_Site_Role::get()` function could return `false` if the option isn't set yet.  So we'd need to make sure `false` gets changed to `''`.  I went with an in-between solution that reduces the `is_callable`'s down to just one, but still tests that the function exists and changes false to blank.  Please see: https://github.com/Automattic/newspack-plugin/pull/3479/commits/703a8d7e3d2d1c19164765de14d4fb58ba045619

ronchambers:
@jaredrethman the reason I prefer to keep `is_callable` is that even if the plugin is active, it could have been refactored to remove the expected function.  I had this happen one time with one of the Newspack plugins where I checked to make sure the required plugin was active, but I didn't check to make sure the class/function still existed.  Everything was fine for months until someone refactored the included plugin.  When the class/function disappeared, and because I didn't have an `is_callable`/`class_exists`/etc,  PHP threw a FATAL.  This brought the publisher's website down for 30 minutes why we tried to figure out why.  So in the cases where one plugin requires something from a different plugin, I try to do as much verification as I can prior to including/calling the external class/function.  Hope it's cool to keep!

jaredrethman:
> @jaredrethman the reason I prefer to keep `is_callable` is that even if the plugin is active, it could have been refactored to remove the expected function. I had this happen one time with one of the Newspack plugins where I checked to make sure the required plugin was active, but I didn't check to make sure the class/function still existed. Everything was fine for months until someone refactored the included plugin. When the class/function disappeared, and because I didn't have an `is_callable`/`class_exists`/etc, PHP threw a FATAL...

@ronchambers valid argument, I hadn't considered this particular scenario.

PR #3481, opened 2024-10-17
feat(auth-modal): onSuccess, onDismiss, and onError callbacks

miguelpeixe:
### All Submissions:

Adds support for a callback when the auth modal closes without authenticating.

Context and instructions at https://github.com/Automattic/newspack-blocks/pull/1908

### Other information:

### DIFF
--- a/src/reader-activation-auth/auth-modal.js
+++ b/src/reader-activation-auth/auth-modal.js
@@ -22,24 +22,24 @@ export function openAuthModal( config = {} ) {
 	const modalTrigger = config.trigger;
 
 	if ( reader?.authenticated ) {
-		if ( config.callback ) {
-			config.callback();
+		if ( config.onSuccess && typeof config.onSuccess === 'function' ) {
+			config.onSuccess();
 		}
 		return;
 	}
 
 	const container = getModalContainer();
 	if ( ! container ) {
-		if ( config.callback ) {
-			config.callback();
+		if ( config.onSuccess && typeof config.onSuccess === 'function' ) {
+			config.onSuccess();
 		}
 		return;
 	}
 
 	const modal = container.closest( '.newspack-reader-auth-modal' );
 	if ( ! modal ) {
-		if ( config.callback ) {
-			config.callback();
+		if ( config.onSuccess && typeof config.onSuccess === 'function' ) {
+			config.onSuccess();
 		}
 		return;
 	}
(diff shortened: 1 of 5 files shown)


chickenn00dle (approved):
Approving this one but just noting I had a non-blocking nit/question about this here: https://github.com/Automattic/newspack-blocks/pull/1908#discussion_r1805335681

src/reader-activation-auth/auth-modal.js
```diff
@@ -22,34 +22,34 @@ export function openAuthModal( config = {} ) {
 	const modalTrigger = config.trigger;
 
 	if ( reader?.authenticated ) {
-		if ( config.callback ) {
-			config.callback();
+		if ( config.onSuccess && typeof config.onSuccess === 'function' ) {
+			config.onSuccess();
```
chickenn00dle, line 26: Non-blocking suggestion: You can remove the need for these checks throughout the PR (`config.onSuccess && ... === 'function'`) by just calling `config?.onSuccess?.()`

src/reader-activation-auth/index.js
```diff
@@ -63,13 +63,14 @@ window.newspackRAS.push( readerActivation => {
...
-				callback = () => {
+				onSuccess = () => {
 					window.location.href = redirect;
 				};
 			}
 
 			openAuthModal( {
-				callback,
+				onSuccess,
+				onError: onSuccess,
```
chickenn00dle: Let's rename the callback to something other than `onSuccess` since its meant to be called even when auth is unsuccessful. `callback` seemed fine to me.

chickenn00dle (changes requested):

src/reader-activation-auth/auth-modal.js
miguelpeixe: This will still throw an error if `onSuccess` is truthy an not a function.

src/reader-activation-auth/index.js
miguelpeixe: Updated in ba879387fa25db4eac9fd3eb016d7749952013e0

src/reader-activation-auth/auth-modal.js
chickenn00dle: Hmm, this bit should cause this to resolve as undefined when not a function: `onSuccess?.()`

miguelpeixe: I double-checked running `const notAFn = true; notAFn?.();` in the console and it errored.

chickenn00dle: Right. This won't work on every browser, but our build process will be sure to compile this to something universally compatible. This isn't a blocker though so feel free to keep what you have!

chickenn00dle (approved):

src/reader-activation-auth/auth-modal.js
chickenn00dle: For reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls

miguelpeixe: I'd rather keep it because the optional chaining operator will not do type validation.

miguelpeixe:
Thanks for reviewing, @chickenn00dle!

PR #3484, opened 2024-10-17
fix: command to fix active subs w/ missing next_payment dates

dkoo:
### All Submissions:

Adds a CLI command to report and and optionally fix active or pending subscriptions that have a `next_payment` date in the past, or are missing a `next_payment` date entirely. This could happen if Action Scheduler jobs failed to update or renew subscriptions as scheduled, or as a result of missing data on migrated subscriptions.

1. Create some subscriptions with missing `next_payment` dates. This must be done using `wp shell` (replace `888` with a subscription ID on your test site) as the UI has safeguards to prevent the deletion of next payment dates. Make sure to do this for at least a few subscriptions that meet the following criteria:

  - A subscription started within the past 90 days
  - A subscription started more than 90 days ago
  - A subscription with no end date
  - A subscription with an end date BEFORE when the next payment date should be
  - A subscription with an end date AFTER when the next payment date
  - A subscription with at least one completed order
  - A subscription with no successful orders (you may need to create this one manually in the admin dashboard)

```bash
> $sub = wcs_get_subscription( 888 )
= WC_Subscription {#15359
    +refunds: null,
    +order_type: "shop_subscription",
  }

> $sub->delete_date( 'next_payment' ) // OR, update the `next_payment` date to a date in the past with $sub->update_dates( [ 'next_payment' => '2024-09-01 21:13:36' ] )
= null

> $sub->save()
= 888
```

2. Run the new CLI command in dry run mode and confirm it reports the subscriptions and missed total revenue, but without actually updating the subscriptions' `next_payment` dates:

```bash
$ wp newspack-woocommerce fix_missing_next_payment_dates --dry-run

===================
=     DRY RUN     =
===================


Fetching active subscriptions with missing or missed next_payment dates...

+-----+--------+---------------------+---------------------+---------------------+----------------+----------------+--------------+
| ID  | status | start_date          | next_payment_date   | end_date            | billing_period | missed_periods | missed_total |
+-----+--------+---------------------+---------------------+---------------------+----------------+----------------+--------------+
| 969 | active | 2023-11-04 20:25:12 | 0                   | 2024-11-04 22:25:44 | month          | 12             | 2400         |
| 888 | active | 2024-06-01 21:13:36 | 2024-12-01 21:13:36 | 0                   | month          | 5              | 250          |
| 475 | active | 2024-08-05 20:36:11 | 2024-11-11 19:24:45 | 0                   | month          | 0              | 0            |
+-----+--------+---------------------+---------------------+---------------------+----------------+----------------+--------------+
Success: Finished processing 3 subscriptions. Total missed revenue: $2,650.00
```

3. In the reported results, ensure the following are true:

  - Only subscriptions started in the past 90 days are handled
  - Subscriptions with no end date get a newly calculated `next_payment` date (reported in the `next_payment_date` column)
  - Subscriptions with a far-future end date should get a newly calculated `next_payment` date before the end date
  - Subscriptions with an imminent end date don't get a new `next_payment` date (they should show `0` in the `next_payment_date` column)
  - Missed periods and total values accurately reflect the number of renewals that should have occurred since the most recent completed renewal order (if any) or the subscription start (if no orders), and based on the subscription product's billing interval and period (e.g. every month/year, every 6 months, etc.).

4. Re-run the command with a `--start-date` arg in `YYYY-MM-DD` format and confirm that the command only reports subscriptions started after the given date.
5. Re-run the command without `--dry-run` and confirm that the command provides the same report data as in step 2, but also updates the subscriptions with the reported `next_payment` dates.

### Other information:

---
  - https://app.asana.com/0/0/1206214865188773

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
@@ -194,7 +194,7 @@ Fetching active subscriptions with missing or missed next_payment dates...
 
 		$query_args    = [
 			'subscriptions_per_page' => $batch_size,
-			'subscription_status'    => WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES,
+			'subscription_status'    => [ 'active', 'pending' ],
 			'offset'                 => 0,
 		];
 		$processed     = 0;
@@ -223,6 +223,7 @@ Fetching active subscriptions with missing or missed next_payment dates...
 
 				$result   = [
 					'ID'                => $subscription->get_id(),
+					'status'            => $subscription->get_status(),
 					'start_date'        => $subscription_start,
 					'next_payment_date' => $next_payment_date,
 					'billing_period'    => $subscription->get_billing_period(),
@@ -269,6 +270,7 @@ Fetching active subscriptions with missing or missed next_payment dates...
 				$results,
 				[
 					'ID',
+					'status',
 					'start_date',
 					'next_payment_date',
 					'billing_period',
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
@@ -221,7 +221,7 @@ Fetching active subscriptions with missing or missed next_payment dates...
 					continue;
 				}
 
-				$result   = [
+				$result = [
 					'ID'                => $subscription->get_id(),
 					'status'            => $subscription->get_status(),
 					'start_date'        => $subscription_start,
@@ -230,10 +230,14 @@ Fetching active subscriptions with missing or missed next_payment dates...
 					'missed_periods'    => 0,
 					'missed_total'      => 0,
 				];
-				$min_date = strtotime( $subscription_start );
-				while ( $min_date <= $now ) {
-					$result['missed_periods']++;
-					$min_date = strtotime( '+1 ' . $result['billing_period'], $min_date );
+				if ( ! empty( $result['billing_period'] ) && ! empty( $result['billing_interval'] ) ) {
+					$period   = $result['billing_period'];
+					$interval = (int) $result['billing_interval'];
+					$min_date = strtotime( "+$interval $period", strtotime( $subscription_start ) ); // Start after first period so we don't count in-progress periods as missed.
+					while ( $min_date <= $now ) {
+						$result['missed_periods']++;
+						$min_date = strtotime( "+$interval $period", $min_date );
+					}
 				}
 
 				if ( $result['missed_periods'] ) {
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cli.php
@@ -163,6 +163,131 @@ class WooCommerce_Cli {
 	}
 
 	/**
+	 * Fixes or reports active subscriptions that have missed next payment dates.
+	 * By default, will only process subscriptions started in the past 90 days.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--dry-run]
+	 * : If set, will report results but will not make any changes.
+	 *
+	 * [--batch-size=<batch-size>]
+	 * : The number of subscriptions to process in each batch. Default: 50.
+	 *
+	 * [--start-date=<date-string>]
+	 * : A date string in YYYY-MM-DD format to use as the start date for the script. Default: 90 days ago.
+	 *
+	 * @param array $args Positional args.
+	 * @param array $assoc_args Associative args.
+	 */
+	public function fix_missing_next_payment_dates( $args, $assoc_args ) {
+		$dry_run    = ! empty( $assoc_args['dry-run'] );
+		$now        = time();
+		$batch_size = ! empty( $assoc_args['batch-size'] ) ? intval( $assoc_args['batch-size'] ) : 50;
+		$start_date = ! empty( $assoc_args['start-date'] ) ? strtotime( $assoc_args['start-date'] ) : strtotime( '-90 days', $now );
+
+		\WP_CLI::log(
+			'
+Fetching active subscriptions with missing or missed next_payment dates...
+		'
+		);
+
+		$query_args    = [
+			'subscriptions_per_page' => $batch_size,
+			'subscription_status'    => WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES,
+			'offset'                 => 0,
+		];
+		$processed     = 0;
+		$subscriptions = \wcs_get_subscriptions( $query_args );
+		$total_revenue = 0;
+		$results       = [];
+
+		while ( ! empty( $subscriptions ) ) {
+			foreach ( $subscriptions as $subscription_id => $subscription ) {
+				array_shift( $subscriptions );
+				$subscription_start = $subscription->get_date( 'start_date' );
+
+				// If the subscription start date is before the $args start date, we're done.
+				if ( strtotime( $subscription_start ) < $start_date ) {
+					$subscriptions = [];
+					break;
+				}
+
+				$next_payment_date = $subscription->get_date( 'next_payment' );
+				$is_in_past        = ! strtotime( $next_payment_date ) || strtotime( $next_payment_date ) < $now;
+
+				// Subscription has a valid next payment date and it's in the future, so skip.
+				if ( $next_payment_date && ! $is_in_past ) {
+					continue;
+				}
+
+				$result   = [
+					'ID'                => $subscription->get_id(),
+					'start_date'        => $subscription_start,
+					'next_payment_date' => $next_payment_date,
+					'billing_period'    => $subscription->get_billing_period(),
+					'missed_periods'    => 0,
+					'missed_total'      => 0,
+				];
+				$min_date = strtotime( $subscription_start );
+				while ( $min_date <= $now ) {
+					$result['missed_periods']++;
+					$min_date = strtotime( '+1 ' . $result['billing_period'], $min_date );
+				}
+
+				if ( $result['missed_periods'] ) {
+					$result['missed_total'] += $subscription->get_total() * $result['missed_periods'];
+					$total_revenue += $result['missed_total'];
+				}
+
+				if ( ! $dry_run ) {
+					$subscription->update_dates(
+						[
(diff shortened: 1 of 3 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,133 @@ function( $subscription ) {
...
+					'start_date'        => $subscription_start,
+					'next_payment_date' => $next_payment_date,
+					'billing_period'    => $subscription->get_billing_period(),
+					'missed_periods'    => 0,
+					'missed_total'      => 0,
+				];
+				$min_date = strtotime( $subscription_start );
+				while ( $min_date <= $now ) {
+					$result['missed_periods']++;
+					$min_date = strtotime( '+1 ' . $result['billing_period'], $min_date );
```
adekbadek: I've a broken subscription, which has no billing period set. The while loop runs indefinitely. There should be a check for the billing period value.

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,133 @@ function( $subscription ) {
...
+		'
+		);
+
+		$query_args    = [
+			'subscriptions_per_page' => $batch_size,
+			'subscription_status'    => [ 'active', 'pending' ],
+			'offset'                 => 0,
+		];
+		$processed     = 0;
+		$subscriptions = \wcs_get_subscriptions( $query_args );
```
adekbadek, line 208: On larger sites, this might take a while – it's querying for _all_ active & pending subs. Direct DB query will likely be way more performant. It has to take into account that some sites use HPOS, while some don't, though.

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,133 @@ function( $subscription ) {
...
+					'billing_period',
+					'missed_periods',
+					'missed_total',
+				]
+			);
+			\WP_CLI::success(
+				sprintf(
+					'Finished processing %d subscriptions. %s',
+					$processed,
+					$total_revenue ? 'Total missed revenue: ' . \wp_strip_all_tags( html_entity_decode( \wc_price( $total_revenue ) ) ) : ''
```
adekbadek, line 263: This is not exactly true: if a sub is starting _now_ (had one payment), and does not have next payment date, no revenue was missed (it will be _after_ the next billing period).

adekbadek (changes requested):

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: Didn't realize this could happen! 6e5c96f4 adds a check for both billing period and billing interval (and correctly handles intervals greater than 1).

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: Good point! 6e5c96f4 changes this so that we don't start counting missed periods until _after_ the first one. I think this should be more accurate, right?

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: It only queries for `$batch_size` (50 by default) at a time, and stops as soon as it finds a subscription with a creation date older than the `--start-date` date (90 days ago by default). So it'll take a while if you give it a start date far in the past and/or there are a ton of subscriptions since the start date, but I think for most sites this shouldn't be much of an issue.

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
adekbadek: Makes sense!

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
adekbadek: Yeah, got it. Should be fine, we'll see on these bigger sites 🤞

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,137 @@ function( $subscription ) {
...
+				$result = [
+					'ID'                => $subscription->get_id(),
+					'status'            => $subscription->get_status(),
+					'start_date'        => $subscription_start,
+					'next_payment_date' => $next_payment_date,
+					'billing_period'    => $subscription->get_billing_period(),
+					'missed_periods'    => 0,
+					'missed_total'      => 0,
+				];
+				if ( ! empty( $result['billing_period'] ) && ! empty( $result['billing_interval'] ) ) {
```
adekbadek: Not satisfying these conditions should result in skipping the current subscription. It can't be fixed if we don't know the billing period & interval.

adekbadek (changes requested):

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: Ah, I see. 6e91f69 skips the subscription if it's missing either billing interval or period.

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,141 @@ function( $subscription ) {
...
+				if ( $next_payment_date && ! $is_in_past ) {
+					continue;
+				}
+
+				$result = [
+					'ID'                => $subscription->get_id(),
+					'status'            => $subscription->get_status(),
+					'start_date'        => $subscription_start,
+					'next_payment_date' => $next_payment_date,
+					'billing_period'    => $subscription->get_billing_period(),
```
adekbadek: Because `'billing_interval'` is never set in the `$result`, all items will bail on the `if` on line 235. 

```suggestion
					'billing_period'    => $subscription->get_billing_period(),
					'billing_interval'  => $subscription->get_billing_interval(),
```

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,141 @@ function( $subscription ) {
...
+					'missed_total'      => 0,
+				];
+
+				// Can't process a broken subscription (missing a billing period or interval).
+				if ( empty( $result['billing_period'] ) || empty( $result['billing_interval'] ) ) {
+					continue;
+				}
+
+				$period   = $result['billing_period'];
+				$interval = (int) $result['billing_interval'];
```
adekbadek: In my local data, I've also subs with `billing_interval` of `0`. Seems like a weird edge case, but better to code defensively :)

adekbadek (changes requested):

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: Whoops, you're right! Sorry for this oversight.

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: Shouldn't matter because of the check for `empty( $result['billing_interval'] )` above (on [line 236](https://github.com/Automattic/newspack-plugin/pull/3484/files#diff-1edfd3af2c626848fa527735509ca45036fd27cfe620169fb98f63740fcf8056R236)). If the `billing_interval` is 0 we won't get to this line.

adekbadek (approved):

dkoo:
@adekbadek just thought of a case that we need to handle with this. The script should check subscriptions for an end date, and if the end date is before the calculated next payment date, it should bail. Otherwise those subscriptions _may_ get renewed instead of properly expiring.

adekbadek:
Ah, good catch! Makes sense. Since this is "business-sensitive", would mind adding unit tests here?

dkoo:
Great idea to add unit tests! I'll keep this PR alive until we add those and handle the "end dates" scenario above.

dkoo:
@adekbadek 9686c86 adds handling for end dates (if a subscription has an end date that will occur before the calculated next payment date, a next payment date will not be set). It also calculates "missed" billing periods by starting from the last successful order, if one exists.

ff7c665 adds a new suite of unit tests and fixes a couple of minor issues exposed by the test suite.

tests/unit-tests/woocommerce-cli.php
```diff
@@ -0,0 +1,196 @@
...
+							'date_completed' => gmdate( 'Y-m-d H:i:s', strtotime( '-1 month' ) ),
+						]
+					),
+				],
+			]
+		);
+		$result = WooCommerce_Cli::calculate_next_payment_date( $subscription );
+
+		// Subscription wasn't processed.
+		$this->assertFalse( $result );
```
adekbadek: Nit: instead of a code comment, the description can be an argument in the assertion. This will make test failures more readable. 

```suggestion
		$this->assertFalse( $result, Subscription wasn't processed. );
```

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
```diff
@@ -162,6 +162,188 @@ function( $subscription ) {
...
+
+	/**
+	 * Given a subscription, calculates the next payment date and missed payments.
+	 *
+	 * @param WC_Subscription $subscription The subscription.
+	 * @param bool            $dry_run If set, will not make any changes.
+	 *
+	 * @return array|false The result array or false if the subscription is broken.
+	 */
+	public static function calculate_next_payment_date( $subscription, $dry_run = false ) {
```
adekbadek: Nit: the function name suggests the the return value is a date. Could be something along the lines of `validate_subscription_dates`.

adekbadek (approved):

tests/unit-tests/woocommerce-cli.php
dkoo: Ah, good call! 2603e8a updates the test assertions.

includes/reader-revenue/woocommerce/class-woocommerce-cli.php
dkoo: Updated to `validate_subscription_dates` in 879ba1a

dkoo:
@adekbadek Merging this one since you preapproved it!

PR #3489, opened 2024-10-21
fix(ras-acc): add onSuccess/onError callbacks to recaptcha render method

chickenn00dle:
### All Submissions:

Part of https://github.com/Automattic/newspack-blocks/pull/1912.

See https://github.com/Automattic/newspack-blocks/pull/1912 for instructions

### Other information:

### DIFF
--- a/src/other-scripts/recaptcha/index.js
+++ b/src/other-scripts/recaptcha/index.js
@@ -136,50 +136,39 @@ function renderWidget( form, onSuccess = null, onError = null ) {
 		}
 
 		// Callback when reCAPTCHA passes validation.
-		const callback = () => {
+		const successCallback = () => {
+			onSuccess?.()
 			form.requestSubmit( button );
 			refreshWidget( button );
 		};
 
-		const errorCallback = () => {
-			refreshWidget( button );
-			const message = wp.i18n.__( 'There was an error with reCAPTCHA. Please try again.', 'newspack-plugin' );
-			if ( onError ) {
-				onError?.( message );
-			} else {
-				// Recaptcha's default error behavior is to alert with the above message.
-				// eslint-disable-next-line no-alert
-				alert( message );
-			}
-		}
-
 		// Render reCAPTCHA widget. See https://developers.google.com/recaptcha/docs/invisible#js_api for API reference.
 		const widgetId = grecaptcha.render( button, {
 			...options,
-			callback,
-			'error-callback': errorCallback,
+			callback: successCallback,
+			'error-callback': () => {
+				refreshWidget( button );
+				const message = wp.i18n.__( 'There was an error with reCAPTCHA. Please try again.', 'newspack-plugin' );
+				if ( onError ) {
+					onError( message );
+				} else {
+					// Recaptcha's default error behavior is to alert with the above message.
+					// eslint-disable-next-line no-alert
+					alert( message );
+				}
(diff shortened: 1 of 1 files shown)


src/other-scripts/recaptcha/index.js
```diff
@@ -135,53 +136,39 @@ function renderWidget( form, errorHandler = null ) {
...
 		setInterval( () => refreshWidget( button ), 120000 ); // Refresh widget every 2 minutes.
 
-		// Refresh reCAPTCHAs on Woo checkout update and error.
-		( function ( $ ) {
-			if ( ! $ ) {
-				return;
-			}
-			$( document ).on( 'updated_checkout', () => renderWidget( form ) );
-			$( document.body ).on( 'checkout_error', () => renderWidget( form ) );
-		} )( jQuery );
```
miguelpeixe, line 171: Won't this break the reCaptcha integration with legacy checkout? My understanding is that our reCaptcha feature is not exclusive to the modal experience.

https://github.com/Automattic/newspack-plugin/blob/f9d4c5389ff6d1824844e64f91a70932efe0ecca/includes/class-recaptcha.php#L29

miguelpeixe (changes requested):
Great changes! Small question below

src/other-scripts/recaptcha/index.js
chickenn00dle: It looks like this only [triggers for v3](https://github.com/Automattic/newspack-plugin/blob/f9d4c5389ff6d1824844e64f91a70932efe0ecca/includes/class-recaptcha.php#L430) where the strategy is to [add a hidden recaptcha field](https://github.com/Automattic/newspack-plugin/blob/f9d4c5389ff6d1824844e64f91a70932efe0ecca/src/other-scripts/recaptcha/index.js#L60). The `renderWidget` method is only called for v2 so changes here shouldn't have an impact.

miguelpeixe: Thank you for explaining!

src/other-scripts/recaptcha/index.js
```diff
@@ -147,8 +152,16 @@ function renderWidget( form, onSuccess = null, onError = null ) {
...
+				const retryCount = parseInt( button.getAttribute( 'data-recaptcha-retry-count' ) ) || 0;
+				if ( retryCount < 3 ) {
+					button.setAttribute( 'data-recaptcha-retry-count', retryCount + 1 );
+					refreshWidget( button );
+				} else {
+					clearInterval( refreshIntervalId );
+				}
+				const message = retryCount < 3
+					? wp.i18n.__( 'There was an error with reCAPTCHA. Please try again.', 'newspack-plugin' )
+					: wp.i18n.__( 'There was an error with reCAPTCHA. Please reload the page and try again.', 'newspack-plugin' );
```
chickenn00dle, line 164: I wonder if we should instead nudge the reader to contact the site admin instead. I have seen cases where the API fails for other unknown reasons though (possibly due to some dev site strangeness), so am just asking for a reload.

Let me know if the site admin messaging would be better though!

miguelpeixe: This should be fine. This will happen on any page load so the admins would become aware instantly.

miguelpeixe (approved):

PR #3491, opened 2024-10-21
fix: let "Cover fees" checkbox work with any payment method

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-blocks/pull/1915, uses a custom hook to render the "cover transaction fees" checkbox for every payment method rendered in the Woo checkout flow. This allows the "cover transaction fees" option to be shown after any payment fields instead of relying on a custom hook in the Stripe payment gateway plugin that's tied to the "legacy" Stripe checkout experience.

1. Start on `release` for this repo and Newspack Blocks.
2. In Newspack > Reader Revenue > Stripe Settings, ensure that "Allow donors to cover transaction fees" is checked.
3. In WooCommerce > Settings > Payments, enable the Stripe payment gateway, ~the WooPayments gateway,~ and at least one other payment method (such as "Check payments").
4. In WooCommerce > Settings > Payments > Stripe > Manage > Settings > Advanced Settings, make sure the "Enable the legacy checkout experience" option is **unchecked**:

Screenshot 2024-10-21 at 3 16 07 PM

5. As a reader, start a donation checkout via the Donate block and proceed to the second screen (payment methods).
6. Select "credit/debit card" for the Stripe payment gateway and observe that the transaction fees checkbox doesn't appear here.
7. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/1915 and refresh the modal checkout, or start a new donation and proceed to the payment screen again. Confirm that the transaction fees checkbox DOES appear for the Stripe ~and WooPayments~ payment methods, but not any other payment methods. 
8. Confirm that the fee is displayed in a price breakdown when the checkbox is selected, and that the price breakdown is hidden when the checkbox is unselected. ~Also confirm that the fee status persists when switching between Stripe and WooPayments, and that the checkbox states for each payment method always match between the two.~
9. Complete a donation with the checkbox both checked and unchecked, and confirm that the order total includes the fee only if the checkbox is checked.
10. Also complete a donation by checking the box, then switching to a payment method other than Stripe ~or WooPayments~. Confirm that the fee is hidden when switching to an unsupported payment method, and that the order total upon completion with that payment method does not include the fee.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
@@ -13,7 +13,8 @@ defined( 'ABSPATH' ) || exit;
  * WooCommerce Order UTM class.
  */
 class WooCommerce_Cover_Fees {
-	const CUSTOM_FIELD_NAME = 'newspack-wc-pay-fees';
+	const CUSTOM_FIELD_NAME  = 'newspack-wc-pay-fees';
+	const SUPPORTED_GATEWAYS = [ 'stripe', 'woocommerce_payments' ];
 
 	/**
 	 * Initialize hooks.
@@ -134,7 +135,7 @@ class WooCommerce_Cover_Fees {
 		if ( ! self::should_allow_covering_fees() ) {
 			return false;
 		}
-		if ( ! isset( $data['payment_method'] ) || 'stripe' !== $data['payment_method'] ) {
+		if ( ! isset( $data['payment_method'] ) || ! in_array( $data['payment_method'], self::SUPPORTED_GATEWAYS, true ) ) {
 			return false;
 		}
 		if ( ! isset( $data[ self::CUSTOM_FIELD_NAME ] ) || '1' !== $data[ self::CUSTOM_FIELD_NAME ] ) {
@@ -149,7 +150,7 @@ class WooCommerce_Cover_Fees {
 	 * @param string $payment_gateway The slug for the payment gateway rendering these payment fields.
 	 */
 	public static function render_stripe_input( $payment_gateway ) {
-		if ( ! self::should_allow_covering_fees() || 'stripe' !== $payment_gateway ) {
+		if ( ! self::should_allow_covering_fees() || ! in_array( $payment_gateway, self::SUPPORTED_GATEWAYS, true ) ) {
 			return;
 		}
 		?>
--- a/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
@@ -23,7 +23,7 @@ class WooCommerce_Cover_Fees {
 		\add_action( 'woocommerce_checkout_update_order_review', [ __CLASS__, 'persist_fee_selection' ] );
 		\add_action( 'woocommerce_cart_calculate_fees', [ __CLASS__, 'add_transaction_fee' ] );
 		\add_action( 'woocommerce_checkout_order_processed', [ __CLASS__, 'add_order_note' ], 1, 3 );
-		\add_action( 'wc_stripe_payment_fields_stripe', [ __CLASS__, 'render_stripe_input' ] );
+		\add_action( 'newspack_blocks_after_payment_fields', [ __CLASS__, 'render_stripe_input' ] );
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
```diff
@@ -145,16 +146,18 @@ private static function should_apply_fee( $data ) {
 
 	/**
 	 * Render the "cover fees" input for Stripe.
+	 *
+	 * @param string $payment_gateway The slug for the payment gateway rendering these payment fields.
 	 */
-	public static function render_stripe_input() {
-		if ( ! self::should_allow_covering_fees() ) {
+	public static function render_stripe_input( $payment_gateway ) {
```
chickenn00dle: Non-blocking question: I understand WCPay fees are via Stripe, but perhaps we should rename this method since we are not only supporting Stripe gateway anymore? WDYT?

chickenn00dle (changes requested):
Nice work @dkoo!

The behavior works as expected, however there is an issue where we are overcharging WCPay transactions for the transaction fee:
Screenshot 2024-10-24 at 16 53 15

I'm assuming we will need to adjust our calculations here to account for the different fee amounts between the gateways.

On a similar note, I also noticed we look to be undercharging for Stripe gateway transactions:

Screenshot 2024-10-24 at 16 53 29

This is less of an issue and not a blocker for this specific PR so we can probably look into the Stripe gateway case in a follow up PR.

includes/reader-revenue/woocommerce/class-woocommerce-cover-fees.php
dkoo: That's a good point. ba8d687a updates this method name plus a couple of others, some inline comments, and the order note added to donation orders when the donor opts in to covering fees, to remove mentions of Stripe and clarify that the option applies to all Newspack-supported payment gateways (currently just Stripe and WooPayments) .

dkoo:
@chickenn00dle about the difference in fee amounts between Stripe vs. WooPayments, I think addressing that could be considered an enhancement rather than a fix as publishers can adjust the fee calculation so that it matches their transaction fees. I'll have a closer look at how the defaults are being calculated vs. the actual Stripe fees in the order total in another PR, though, as the default should match Stripe's default behavior...

dkoo:
@chickenn00dle re: the fee amounts, the fee is calculated on the pre-tax total. Maybe WooPayments and Stripe are treating the tax amount differently in terms of the transaction fee, which would account for the difference in payout. Otherwise, without taxes I saw that the fees were identical for both Woo and Stripe gateways, and the fee we charged matched the fees deducted by the payment gateways:

<table>
<tr>
<td>

**Stripe:**

Screenshot 2024-10-24 at 3 58 34 PM
</td>
<td>

**WooPayments:**

Screenshot 2024-10-24 at 3 58 30 PM
</td>
</tr>
</table>

Now, we probably _should_ be calculating the transaction fee on top of any other fees and taxes collected, but I think almost all Newspack sites are not collecting taxes as part of these transactions because donations are generally not taxable purchases in most countries. So I think fixing that is a relatively low priority.

chickenn00dle (approved):
> Now, we probably should be calculating the transaction fee on top of any other fees and taxes collected, but I think almost all Newspack sites are not collecting taxes as part of these transactions because donations are generally not taxable purchases in most countries.

I'll give this a thumbs up with this in mind, however I will note that I would much rather see us remove WCPay from this PR altogether and add support for this when we are willing to look further into the calculation logic.

Ultimately, I just don't want to expose our publishers to any chance of unknowingly overcharging their readers, even if a very small amount. But if its the case that we are certain none will be impacted by this, then I'm okay to see the fix for this go up as is if you feel strongly about adding WCPay now and are going to work on adjusting the calculation logic right away.

dkoo:
@chickenn00dle I understand your concerns. 411ab7a4 removes WooPayments from the array of "supported gateways" for now, so it won't get the transaction fee checkbox.

chickenn00dle (approved):
Great! Thanks for the change. Looks good 👍

dkoo:
Thanks, @chickenn00dle! Holding for a Monday release.

dkoo:
Note: d36d6dd fixes another bug we found for sites that had never saved the "allow covering fees" option. The `get_option( 'newspack_donations_allow_covering_fees', true );` check should always include `true` as the default second arg.

PR #3492, opened 2024-10-23
fix(webhooks): deprecate global endpoint

miguelpeixe:
### All Submissions:

1200550061930446-as-1208602681752393/f

Global webhook endpoints can be very disruptive to the site as our data events base grows. There's no clear use for such a feature, so we should deprecate it for now.

1. Make sure you have `define( 'NEWSPACK_EXPERIMENTAL_WEBHOOKS', true );` in your `wp-config.php`
2. Checkout this branch and visit "Newspack -> Connections"
3. Create a new endpoint and confirm the global checkbox is no longer an option
4. Attempt to save without selecting any actions and confirm you get an error message
5. Set the endpoint URL (use https://webook.site), select the `reader_logged_in` action and save
6. In a new session, authenticate as a reader
7. In the dashboard, click the endpoint's card "View Requests" and confirm the request is created
8. Wait until it's processed and confirm the endpoint receives the data

### Other information:

---
  - https://app.asana.com/0/0/1208602681752393

### DIFF
--- a/includes/data-events/class-api.php
+++ b/includes/data-events/class-api.php
@@ -102,7 +102,7 @@ final class Api {
 		if ( ! current_user_can( 'manage_options' ) ) {
 			return new \WP_Error(
 				'newspack_rest_forbidden',
-				esc_html__( 'You cannot use this resource.', 'newspack' ),
+				esc_html__( 'You cannot use this resource.', 'newspack-plugin' ),
 				[
 					'status' => 403,
 				]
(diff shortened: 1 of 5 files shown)


adekbadek (approved):
Works as described.

> Attempt to save without selecting any actions and confirm you get an error message

Sidenote: there should be front-end validation here.

miguelpeixe:
Good call! Added in 2bb2cdba550e1c5e02dd566396eb77d7cafd01ac

In case of an error, the modal scrolls to the top. To achieve that I had to refactor our Modal wrapper a bit. It wasn't forwarding the ref so the modal DOM wasn't accessible. It's a small refactor, but please also test the modal in other contexts, like Campaigns, to ensure it doesn't introduce any regression.

adekbadek (approved):

PR #3494, opened 2024-10-23
feat: support for WooPayments as a payment gateway

dkoo:
### All Submissions:

Adds support for payment gateways beyond Stripe in the Reader Revenue > Payment Methods dashboard page. This PR adds WooPayments as a first-party payment gateway option and upgrades this plugin to Newspack's list of "managed" plugins that can be auto-installed in one click.

Because the Transaction Fee options apply to both Stripe and WooPayments gateways, these options have been moved outside of the Stripe component, as well as the data payload for Stripe, and into its own component/endpoints.

Screenshot 2024-10-22 at 3 16 29 PM

This PR also updates the UI shown under "Reader Revenue" during the setup wizard. Previously, this component showed the Donation amounts UI, plus the UI to input Stripe API keys. Now, this component will show the platform selector plus the Donation amounts UI (this is so we can at least show something—otherwise if you get to the screen and you hadn't already installed Woo and Woo Subscriptions plugins, this component would be totally empty).

Lastly, this PR removes the plugin installer component shown if the required payment gateway plugin isn't installed or active on your site. Since our handler to enable plugin gateways will automatically install and activate the required plugin if it's missing, this is an unnecessary UI step.

1. Start with a fresh testing site and install this branch.
2. Proceed through the initial setup steps. In the "Settings" step make sure to select a Country and a Currency before proceeding to the Services step. This will avoid a fatal error when enabling WooPayments for the first time (seems to be a bug with WooPayments, which assumes you've completed WooCommerce setup before enabling it).
3. On the "Services" page, enable the Reader Revenue toggle and confirm you see the platform selector here. Confirm that you can switch back and forth between platforms:

Screenshot 2024-10-22 at 4 47 47 PM

4. Select "Newspack" and confirm that you see the required plugins component:

Screenshot 2024-10-22 at 4 49 09 PM

5. Confirm that you can install WooCommerce using the "Install" link. You won't be able to install Woo Subscriptions here, so install it using CLI then refresh the page (you may have to exit the WooCommerce onboarding screen and proceed through the initial Newspack setup steps again—for this reason for most site setups it probably makes sense to preinstall the required Woo plugins before initiating the setup wizard).
6. Now when enabling Reader Revenue in the Services page, you should see the basic donation amounts UI (this should be the same as before, just minus the Stripe settings):

Screenshot 2024-10-23 at 3 45 54 PM

6. Complete the setup wizard, then go to **Newspack > Reader Revenue > Payment Methods**. Confirm you see the new payment gateways UI:

Screenshot 2024-10-23 at 3 49 17 PM

7. Toggle ON both payment gateways. Confirm that both enter an enabled but "not connected" state, and confirm that the "Additional Settings" card appears below them. Confirm that you can update and save the transaction fee fields in "Additional Settings".

Screenshot 2024-10-23 at 3 51 07 PM

9. Click on "Connect" for each payment gateway and confirm that each leads to the screen to initiate the connection process for that payment gateway.
10. Complete the connection flows for both Stripe and WooPayments in test mode (you will need to log into a Stripe test account to complete both flows).
  - For Stripe, click the "Create or connect a test account instead" link:
Screenshot 2024-10-23 at 3 52 09 PM

  - For WooPayments, expand the "I'm setting up a store for someone else" card and click the "Enable sandbox mode" button:
Screenshot 2024-10-23 at 3 51 43 PM

11. After completing both connection flows, return to **Newspack > Reader Revenue > Payment Methods** and confirm that both payment gateways are shown as active and connected in test mode:

Screenshot 2024-10-23 at 3 55 32 PM

12. Click the "Configure" button for each payment gateway and confirm that each leads to the settings page for that payment gateway plugin. 
13. Publish a Donate block on a post or page and visit as a reader. Start a donation checkout and confirm that you can complete a test transaction using both payment gateways (you should see two options for "Credit card"):

Screenshot 2024-10-23 at 3 58 59 PM

14. Confirm that after toggling OFF either payment gateway in the Reader Revenue > Payment Methods page, that payment gateway is no longer available for checkout on the front-end.

### Other information:


adekbadek (changes requested):
1. I don't see the "cover fees" section (the "Collect transaction fees" section in RR wizard is toggled on). The "Cover Fees" UI is only visible after trying to save the "Collect transaction fees" section – see next item,

2. When clicking on the "Collect transaction fees" card in the RR wizard, the card reloads and the toggle is still toggled on. It can't be untoggled.

dkoo:
@adekbadek looks like both this branch and its parent needed an update with `trunk`. It should be working as described, now. Also note that in #3491 we [temporarily removed support for WooPayments](https://github.com/Automattic/newspack-plugin/pull/3491#issuecomment-2438186719) due to inconsistent fee calculations between WooPayments vs. Stripe when the cart total also included taxes. Although this is an edge case because most countries won't require taxes on donation transactions, 8ac42ab should fix this by including taxes in the fee calculation, and also restores support for WooPayments.

adekbadek (approved):

PR #3490, opened 2024-10-21
Feat / IA for Newsletters

ronchambers:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Once the "TODOs" below are done, this PR will complete the Information Architecture Wizard for the Newsletters Plugin. This code will update the existing Newsletters Plugin header and menu.  

See screenshot:

newsletters-tabs

* The Newsletters custom blue header is deactivated.
* A new Admin Header is added.
* The menu is moved and compressed.
* Some menu items are now combined into Tabs.

Additionally the new Dashboards page is updated so that if the Newsletters plugin is not installed, then the dashboard will not show Newsletters items.  

See screenshot:

newsletters-dashboard-new

### TODOs:

- We may need to change the menu icon.  Export the SVG from Figma or talk to Thomas/designer.  See code comment: `// @TODO get SVG from Figma? This one is "envelope" from: https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library`

- I was having a hard time setting the Newsletters CPT parent menu item to a decimal value.  It seems that CPTs only support an integer value for `menu_positon`.  So I created a function that will loop through `global $menu` and then move the CPT to position 3.3 or "3.33" if something is already at 3.3.  See the current code in function `move_cpt_menu()`.  You may want to look at function `registered_post_type_newsletters( $post_type )` to see how I'm changing the `menu_icon` but when I try to change the `menu_position` to 3.3 in that function it doesn't work.

- I didn't have chance to fully test what happens if someone goes to a "bookmarked" category or tag page that was removed by this Wizard.  I recommended we remove categories and tags from the Newsletters menu in this PR ( https://github.com/Automattic/newspack-newsletters/pull/1678 ) but still, even with the links removed, someone could have bookmarked any of the following links, so we may want to view them in the browser and see what the user would see.  We'll probably want to remove the "blue" Newsletters header and apply the new Admin Header like this code does for the other screens.  
[...]

PR #3501, opened 2024-10-25
refactor: wizards menu order handling

miguelpeixe:
### All Submissions:

I started investigating the different ways we're managing menu order for wizard, prompted by @ronchambers [comment](https://github.com/Automattic/newspack-plugin/pull/3490#issue-2603331787) regarding the Newsletters' wizard menu order:

> I was having a hard time setting the Newsletters CPT parent menu item to a decimal value. It seems that CPTs only support an integer value for menu_positon. So I created a function that will loop through global $menu and then move the CPT to position 3.3 or "3.33" if something is already at 3.3. See the current code in function move_cpt_menu(). You may want to look at function registered_post_type_newsletters( $post_type ) to see how I'm changing the menu_icon but when I try to change the menu_position to 3.3 in that function it doesn't work.

Turns out we have multiple strategies in place:
- [hook priority](https://github.com/Automattic/newspack-plugin/blob/26a610637adb00022e7c59bf3108987391dc5b06/includes/wizards/class-wizard.php#L66)
- [`add_menu_page()` argument](https://github.com/Automattic/newspack-plugin/blob/26a610637adb00022e7c59bf3108987391dc5b06/includes/wizards/advertising/class-advertising-display-ads.php#L550)
- [custom reordering logic](https://github.com/Automattic/newspack-plugin/blob/26a610637adb00022e7c59bf3108987391dc5b06/includes/wizards/class-network-wizard.php#L148-L159)

This is hard to maintain and unpredictable. In this refactor PR I propose a consolidated strategy applied in the `class-wizards.php`:

https://github.com/Automattic/newspack-plugin/blob/406bc54174789f7f83cddc1cefede77b4caee5dd/includes/class-wizards.php#L141-L172

Each wizard with a parent menu item can use the public vars `parent_menu` and `menu_order` to determine how it should render relative to the `newspack-dashboard` wizard:

https://github.com/Automattic/newspack-plugin/blob/406bc54174789f7f83cddc1cefede77b4caee5dd/includes/wizards/class-network-wizard.php#L33-L45

Result:

image

This PR also deprecates the `menu_priority` variable from the abstract class. It's a bit misleading, as it only applies the hook priority to the `admin_menu` hook. This is not a guaranteed way of setting the menu order because it gets overridden by the `add_menu_page` argument.

1. Checkout this PR, make sure you have Ads, Network, and Newsletters plugin active
2. Confirm the menu order reflects the image above
3. Click each menu and confirm it behaves as expected

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -127,7 +127,7 @@ final class Newspack {
 
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/class-setup-wizard.php';
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/class-components-demo.php';
-		
+
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/traits/trait-wizards-admin-header.php';
 
 		// Newspack Wizards and Sections.
@@ -140,8 +140,8 @@ final class Newspack {
 
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/class-setup-wizard.php';
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/class-components-demo.php';
-		
-		// Advertising Wizard. 
+
+		// Advertising Wizard.
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/advertising/class-advertising-display-ads.php';
 		include_once NEWSPACK_ABSPATH . 'includes/wizards/advertising/class-advertising-sponsors.php';
 
--- a/includes/class-wizards.php
+++ b/includes/class-wizards.php
@@ -44,22 +44,25 @@ class Wizards {
 			'settings'                => new Settings(),
 			// v2 Information Architecture.
 			'newspack-dashboard'      => new Newspack_Dashboard(),
-			'newspack-settings'       => new Newspack_Settings( 
+			'newspack-settings'       => new Newspack_Settings(
 				[
 					'sections' => [
 						'custom-events' => 'Newspack\Wizards\Newspack\Custom_Events_Section',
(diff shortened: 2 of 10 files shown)


jaredrethman (approved):
@miguelpeixe have tested the code and works as described. 

Additionally: I really like your approach, first time i’ve come across [custom_menu_order](https://developer.wordpress.org/reference/hooks/custom_menu_order/) hook - looks like the right tool for the job! 

Approved!

miguelpeixe:
Thanks, @jaredrethman! Since this touches on a core aspect of the project, I'll also wait for @ronchambers feedback before merging.

leogermani:
Not sure if this has any impact here, but for awareness: https://core.trac.wordpress.org/ticket/52035

I've been meaning to circle back to this and help it get to the finish line

ronchambers:
I didn't test locally, but I trust the two of you 😃   Overall, I agree that moving the menus around for the 3 different plugins (Listings, Newsletters, and Network - my original tasks ) became cumbersome where each plugin had a different approach...  So yes, a standard approach is great!  Since Listings and Newsletters were re-assigned to @miguelpeixe I won't worry about testing those.  I do need to do some final testing on the Network plugin wizard so I'll be sure to merge this (or epic/ia after it's merged) to my Network branch and verify it's all good there too.  Approved for me!  Thanks!

miguelpeixe:
> Not sure if this has any impact here, but for awareness: https://core.trac.wordpress.org/ticket/52035

Thanks for flagging! The strategy doesn't cover submenu pages. I'm assuming we have tighter control over the execution order for submenu pages so it's less problematic.

PR #3510, opened 2024-10-30
feat(ras): login reader after successful password reset

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208631669643213/f

This PR logs in a reader after they have successfully reset their password when RAS is enabled:

Screenshot 2024-10-30 at 11 22 27

1. As admin, ensure RAS is enabled
2. As a reader, go to my account and go through the reset password flow
3. Ensure you are automatically logged in after resetting the password at the password reset form pictured above
4. Log out
5. Trigger the auth flow via the sign in button
6. Enter the same reader email as from step 1, but don't input your password. Instead click the Forgot Password button
7. Go through the reset flow and again ensure you are automatically logged in after resetting the password at the password reset form pictured above

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -84,6 +84,7 @@ final class Reader_Activation {
 		\add_action( 'wp_footer', [ __CLASS__, 'render_auth_modal' ] );
 		\add_action( 'wp_footer', [ __CLASS__, 'render_newsletters_signup_modal' ] );
 		\add_action( 'wp_ajax_newspack_reader_activation_newsletters_signup', [ __CLASS__, 'newsletters_signup' ] );
+		\add_action( 'woocommerce_customer_reset_password', [ __CLASS__, 'login_after_password_reset' ] );
 
 		if ( self::is_enabled() ) {
 			\add_action( 'clear_auth_cookie', [ __CLASS__, 'clear_auth_intention_cookie' ] );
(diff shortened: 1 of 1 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -2359,5 +2354,31 @@ private static function get_logged_in_reader_email_address() {
...
+	 *
+	 * @param WP_User $user WP_User object.
+	 */
+	private static function log_in_reader( $user ) {
+		\wp_clear_auth_cookie();
+		\wp_set_current_user( $user->ID );
+		\wp_set_auth_cookie( $user->ID, true );
+		\do_action( 'wp_login', $user->user_login, $user );
+		Logger::log( 'Logged in user ' . $user->ID );
+	}
```
miguelpeixe: Why not use `set_current_reader()`?

miguelpeixe (changes requested):
Works as described, but it's not clear to me why we are decoupling the `set_current_reader()` method without the reader check. `log_in_reader()`, despite its name, ends up logging in anyone.

includes/reader-activation/class-reader-activation.php
chickenn00dle: > Why not use set_current_reader()?

I wanted to avoid the redundant validation and potential returning of WP_Error that comes with `set_current_reader` since in this case we are always going to be logging in the user regardless of whether they are a `reader` or not at successful password reset.

That said, I'm fine going with `set_current_reader` instead if you prefer! Updated in https://github.com/Automattic/newspack-plugin/pull/3510/commits/bfe042f16fe1bd115c788f86799bfbf58da37d14

miguelpeixe: > in this case we are always going to be logging in the user regardless of whether they are a reader or not at successful password reset.

To be safe, it's better to have this only for readers, which I believe is how it's now after bfe042f16fe1bd115c788f86799bfbf58da37d14. Publishers may have additional layers of security added to the login form aimed at users with higher capabilities that this could bypass (IP/proxy, brute force protection, 2fa, etc.). A bad actor could find a way to reset an account's password, but if there's additional protection, that doesn't mean they're allowed in instantly.

That's also why I hesitate to have a method that doesn't check if it's a reader before logging in. It could get accidentally misused.

miguelpeixe (approved):

PR #3514, opened 2024-10-31
feat: also trigger subscription change events on switch

leogermani:
### All Submissions:

Also trigger the data events for subscription changes when there's a subscription switch.

Questions:

1. Do the docs look ok?
2. Do you think we should add an additional param to the event identifying the type of change (status or switch)? - Note that we don't have access to the previous state of the subscription to add to the event exactly what changed
3. How do I configure a non donation subscription that can be downgraded/upgraded to another product?
4. Should we keep this only to donations for now?

Testing donations:

1. As a reader, make a recurring donation
2. Go to My Account > My Subscription
3. Click on "Change price"
5. choose a product and a value and submit
6. Visit card and proceed to checkout
7. Complete checkout
8. Check that the donation_changed event is fired with the correct info (you can set a webhook using https://webhook.site/ for that if you want)

### Other information:

---
  - https://app.asana.com/0/0/1208541920545689

### DIFF
--- a/includes/data-events/listeners.php
+++ b/includes/data-events/listeners.php
@@ -253,6 +253,46 @@ Data_Events::register_listener(
 );
 
 /**
+ * For when a donation subscription is switched (recurrence/amount change).
+ */
+Data_Events::register_listener(
+	'woocommerce_subscriptions_switch_completed',
+	'donation_subscription_changed',
+	function( $order ) {
+
+		if ( ! function_exists( 'wcs_get_objects_property' ) ) {
+			return;
+		}
+
+		$switch_order_data = wcs_get_objects_property( $order, 'subscription_switch_data' );
+
+		if ( empty( $switch_order_data ) || ! is_array( $switch_order_data ) ) {
+			return;
+		}
+
+		$subscription_id = array_key_first( $switch_order_data );
+
+		$subscription = wcs_get_subscription( $subscription_id );
+
+		if ( ! $subscription instanceof \WC_Subscription ) {
+			return;
+		}
+
+		$data = \Newspack\Data_Events\Utils::get_recurring_donation_data( $subscription );
+		if ( ! is_array( $data ) ) {
+			return;
+		}
+		return array_merge(
+			$data,
+			[
+				'status_before' => $subscription->get_status(),
+				'status_after'  => $subscription->get_status(),
+			]
+		);
+	}
+);
+
+/**
  * When a non-donation subscription status changes.
  * The subscription will be removed from the user's list of active subscriptions.
  */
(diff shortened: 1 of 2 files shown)


dkoo (approved):
@leogermani this works as described. The "switch" functionality is related to the Name Your Price extension—to create a group of non-donation products that can be switched, you can either create a variable subscription and mark its variations as "Name Your Price" enabled, or create a bunch of products with NYP enabled and group them under a parent product (like with our Donation products).

includes/data-events/listeners.php
```diff
@@ -252,6 +252,46 @@ function( $subscription, $status_to, $status_from ) {
...
+		}
+
+		$data = \Newspack\Data_Events\Utils::get_recurring_donation_data( $subscription );
+		if ( ! is_array( $data ) ) {
+			return;
+		}
+		return array_merge(
+			$data,
+			[
+				'status_before' => $subscription->get_status(),
```
dkoo, line 288: Nit: I'm not sure we need to add `status_before` and `status_after` params to the switch handlers. These are only relevent if the status has changed

dkoo:
To address your Qs:

> 1. Do the docs look ok?

They look fine to me—we might consider mentioning Name Your Price as the source of the "switch" feature.

> 2. Do you think we should add an additional param to the event identifying the type of change (status or switch)? - Note that we don't have access to the previous state of the subscription to add to the event exactly what changed

Not sure this is necessary if you take my [suggestion](https://github.com/Automattic/newspack-plugin/pull/3514#discussion_r1830202091) because the status changes will get `status_before` and `status_after` params. But it can't hurt to add `'subscription_switch' => true` to these handlers.

> 4. Should we keep this only to donations for now?

Nah, I think it's good to handle both. I'm not sure if any publishers use NYP for anything other than donations, but you never know.

includes/data-events/listeners.php
leogermani: I did that to keep consistency and make sure the event always has the same parameters in all cases. I was thinking about making it easier for people building integrations that rely on this event

leogermani:
> @leogermani this works as described. The "switch" functionality is related to the Name Your Price extension—to create a group of non-donation products that can be switched, you can either create a variable subscription and mark its variations as "Name Your Price" enabled, or create a bunch of products with NYP enabled and group them under a parent product (like with our Donation products).

I think NYP works on top of it, but Subscription Switch is a concept that exist in Woo Subscriptions itself -> https://woocommerce.com/document/subscriptions/switching-guide/

Or am I missing something?

dkoo:
Ah, you're right—I think the only requirement is that the product is part of a group or a variation in a variable subscription product.

includes/data-events/listeners.php
dkoo: That makes sense! In that case, it can't hurt to add a `'type' => 'switch' | 'status'` param too

leogermani:
I'll merge this as it is. Thinking that adding a `type` with a value of `switch` could add more confusion instead of helping. What is a "switch"? What is "status"?... 

They can simply check all the fields and they'll know what have changed

PR #3520, opened 2024-11-04
fix: use user email for esp sync purposes

leogermani:
### All Submissions:

When a user purchases something on a site with RAS activated and its info gets synced to the ESP, we were prioritizing the billing email over the user registered email.

This means that if the user registers with one email, but then uses a different email to make a purchase, it would create a new user in the ESP - but not in WP.

This PR makes sure we always use the user email for sync purposes.

Was there any specific reasons why we wanted to use billing email there? Do we really want this change?

1. on `trunk`
2. Register with a user
3. Make a donation or purchase something, but change your email in the checkout form
4. Confirm ESP syncs info about this purchase using the modified email
5. On this branch, confirm that all ESP syncs are tied to the original user email
6.
7.

### Other information:

---
  - https://app.asana.com/0/0/1208677634068371

### DIFF
--- a/includes/reader-activation/sync/class-woocommerce.php
+++ b/includes/reader-activation/sync/class-woocommerce.php
@@ -341,7 +341,7 @@ class WooCommerce {
 		$last_name  = $customer->get_billing_last_name();
 		$full_name  = trim( "$first_name $last_name" );
 		$contact    = [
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
> Was there any specific reasons why we wanted to use billing email there? Do we really want this change?

I don't think it was intentional, it was rather assumed that the billing email will be the same as the user email.

PR #3525, opened 2024-11-05
feat(plugins): add url redirect for wc subscription renewals

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1206274567818686/1208437232706180/f

This PR adds handling for ~~a `np_renewal` param~~ a new `my-account/my-renewals` endpoint to redirect readers to the first available pending renewal checkout, otherwise the subscriptions my account page:

Screenshot 2024-11-05 at 16 39 08

1. As a reader, purchase multiple subscriptions
2. As admin, go to any subscription edit page for one of these subscriptions via WooCommerce > Subscriptions
3. Create a renewal order for the subscription
Screenshot 2024-11-05 at 16 41 21
4. As the logged in reader, visit the `my-account/my-renewals` endpoint of your test site: `site.url/my-account/my-renewals`
5. Confirm you are redirected to checkout for the renewal order in question
6. As admin again, create another renewal order for another subscription belonging to this reader (there should now be two subscriptions that need payment)
7. As reader again, visit the url from step 3 again
8. This time confirm you are directed to the my account subscriptions page
Screenshot 2024-11-06 at 18 19 15
9. Complete checkout to re-activate the subscription. Then do the same for the second pending renewal (All subscriptions should be active).
10. Visit the url from step 3 again.
11. Confirm you are directed to the my account subscriptions page again
12. Log out
13. Visit the url from step 3 again.
14. Confirm you are presented with a login form
15. Login and confirm you are redirected to the my account subscriptions page once again.

### Other information:


src/reader-activation/auth.js
```diff
@@ -113,7 +113,9 @@ window.newspackRAS.push( function ( readerActivation ) {
 						/** If there's a pre-auth, signing in redirects to the reader account. */
 						if ( reader?.email && ! reader?.authenticated ) {
 							link.setAttribute( 'data-redirect', link.getAttribute( 'href' ) );
-							redirectInput.value = link.getAttribute( 'href' );
+							if ( ! redirectInput?.value ) {
```
chickenn00dle: We should respect any pre-defined redirect values.

src/reader-activation/auth.js
```diff
@@ -177,7 +179,7 @@ window.newspackRAS.push( function ( readerActivation ) {
 				emailInput.value = reader?.email || '';
 			}
 
-			if ( redirectInput && ev?.target?.getAttribute( 'data-redirect' ) ) {
+			if ( ev?.target?.getAttribute( 'data-redirect' ) && ! redirectInput?.value ) {
```
chickenn00dle: Same here re: respecting pre-defined redirect values

chickenn00dle:
Got some feedback here so removing the `Needs Review` label to work on this: 0-as-1208437232706180/1208702687864921/f

chickenn00dle:
Added some changes to make the redirect happen via a new my-account endpoint `my-renewals`. Also updated the testing instructions. 

Just awaiting feedback to verify the expected behavior before marking this one ready for review.

dkoo (approved):
Working well! Conditionally approved, with just a couple of questions:

1. I had to refresh permalinks before the new endpoint would redirect. Is there anything we need to do to ensure this happens automatically with no downtime for live sites?
2. Based on the discussion in https://github.com/Automattic/newspack-plugin/pull/3560#issuecomment-2498249591 it sounds like we're moving to a `woocommerce-subscriptions` directory with smaller classes inside, so should we refactor to this structure here to avoid a future conflict? This could probably just be moved to the `woocommerce-subscriptions` directory since it's fairly self-contained.

leogermani (changes requested):
as per @dkoo review, we need to add a rewrite rule refresh to be execute once after this change is deployed. Also agree with his comment on the file structure.

And just to confirm that naming things is one of the hardest problems in computes science. WDYT of `renew-subscription` or `subscription-renew` (in the singular form, because that's the main intent of this link, redirect to Subscriptions is a fallback)

chickenn00dle:
Thanks for the feedback @leogermani and @dkoo!

> we need to add a rewrite rule refresh to be execute once after this change is deployed. Also agree with his comment on the file structure.

Flushed permalinks and reorganized in https://github.com/Automattic/newspack-plugin/pull/3525/commits/fa7cdc2fc45c25e07d31ddb922f6766c6a8e5262

> And just to confirm that naming things is one of the hardest problems in computes science. WDYT of renew-subscription or subscription-renew (in the singular form, because that's the main intent of this link, redirect to Subscriptions is a fallback)

I like `renew-subscription`! I went with that in https://github.com/Automattic/newspack-plugin/pull/3525/commits/fa7cdc2fc45c25e07d31ddb922f6766c6a8e5262

leogermani:
This looks good and works nicely.

One question though. My expectation was - I think it would be awesome - that if I'm not logged in and visit `renew-subscription`, I will be redirected to the checkout page, the same thing that happens when I'm logged in. Do you think that's possible?

chickenn00dle:
> that if I'm not logged in and visit renew-subscription, I will be redirected to the checkout page, the same thing that happens when I'm logged in. Do you think that's possible?

We can accomplish this by adding a redirect param with a value of the new endpoint and handling this when rendering the auth form. Added in https://github.com/Automattic/newspack-plugin/pull/3525/commits/a01a0b5ba429ec9504f286c497aed2f74fa71fa4

Let me know what you think!

leogermani (approved):

PR #3529, opened 2024-11-06
fix(ia): deprecate webhooks global endpoint

miguelpeixe:
### All Submissions:

1205919985867982-as-1208628153846581/f

Implements changes from #3492 to the IA project.

1. Navigate to "Newspack -> Settings" and at the Connections tab, scroll to the Webhooks section 
2. Click to create a new endpoint and confirm the "Global" option is no longer available
3. Attempt to save without a URL or selecting any actions
4. Confirm the modal scrolls to the top and you see an error message
5. Fill a random string (invalid URL) and attempt to save
6. Confirm you get an error due to invalid format
7. Put a valid URL and select some actions
8. Confirm you can save the endpoint
9. Refresh the page and confirm the endpoint persist

### Other information:

### DIFF
--- a/src/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
+++ b/src/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
@@ -33,27 +33,22 @@ function EndpointActionsCard( {
 			description={ () => {
 				if ( endpoint.disabled && endpoint.disabled_error ) {
 					return `${ __(
-						'Endpoint disabled due to error:',
+						'Endpoint disabled due to error',
 						'newspack-plugin'
 					) }: ${ endpoint.disabled_error }`;
 				}
 				return (
 					<Fragment>
 						{ __( 'Actions:', 'newspack-plugin' ) }{ ' ' }
-						{ endpoint.global ? (
-							<span className="newspack-webhooks__endpoint-action">
-								{ __( 'global', 'newspack-plugin' ) }
+						{ endpoint.actions.map( action => (
+							<span
+								key={ action }
+								className="newspack-webhooks__endpoint-action"
+							>
+								{ action }
 							</span>
-						) : (
-							endpoint.actions.map( action => (
-								<span
-									key={ action }
-									className="newspack-webhooks__endpoint-action"
-								>
-									{ action }
-								</span>
-							) )
-						) }
+						) )
+					 }
 					</Fragment>
 				);
 			} }
--- a/src/wizards/newspack/views/settings/connections/webhooks/modals/upsert.tsx
+++ b/src/wizards/newspack/views/settings/connections/webhooks/modals/upsert.tsx
@@ -6,7 +6,7 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState, Fragment } from '@wordpress/element';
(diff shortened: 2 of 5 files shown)


src/wizards/newspack/views/settings/connections/webhooks/endpoint-actions-card.tsx
```diff
@@ -33,27 +33,22 @@ function EndpointActionsCard( {
 			description={ () => {
 				if ( endpoint.disabled && endpoint.disabled_error ) {
 					return `${ __(
-						'Endpoint disabled due to error:',
+						'Endpoint disabled due to error',
 						'newspack-plugin'
 					) }: ${ endpoint.disabled_error }`;
```
miguelpeixe, line 38: This was printing two `:`, removed the one from the localized string

src/wizards/newspack/views/settings/connections/webhooks/modals/upsert.tsx
```diff
@@ -78,9 +87,14 @@ const Upsert = ( {
 	}
 
 	function testEndpoint(
-		url: string | undefined,
+		url: string,
```
miguelpeixe, line 90: Not sure why this would ever be `undefined` so I removed it

src/wizards/newspack/views/settings/connections/webhooks/utils.tsx
```diff
@@ -81,3 +82,40 @@ export function getRequestStatusIcon(
...
 	return endpoint.requests.some( request => request.errors.length );
 }
+
+/**
+ * Validate endpoint URL.
+ *
+ * @param url The URL to validate.
+ * @return    Error message if URL is invalid, false otherwise.
+ */
+export function validateUrl( url: string ): string | false {
```
jaredrethman, line 92: This can be modernized to use [URL interface](https://developer.mozilla.org/en-US/docs/Web/API/URL) for example:
```ts
export function validateUrl( url: string ): string | false {
	if ( ! url ) {
		return __( 'URL is required.', 'newspack-plugin' );
	}
	try {
		const urlObject = new URL( url );
		// Ensure the protocol is either http or https
		if ( ! [ 'http:', 'https:' ].includes( urlObject.protocol ) ) {
			return __( 'URL must use HTTP or HTTPS protocol.', 'newspack-plugin' );
		}
		return false;
	} catch ( error ) {
		return __( 'Invalid URL format.', 'newspack-plugin' );
	}
}
```
There is also a function inside `@wordpress/url` called `isUrl` which also validates urls using the above pattern.

jaredrethman (approved):
@miguelpeixe code looks good, I made one non-blocking comment about validating URLs below. Feel free to keep or change as you think!

src/wizards/newspack/views/settings/connections/webhooks/utils.tsx
miguelpeixe: Good call! Updated in 9eb25894d8850e5054d0ea0998e89d7eb8a9181e. While at it, I also restricted to HTTPS, which [is enforced in the backend](https://github.com/Automattic/newspack-plugin/blob/9eb25894d8850e5054d0ea0998e89d7eb8a9181e/includes/data-events/class-api.php#L248).

jaredrethman (approved):
Code+changes look great, approved!

PR #3530, opened 2024-11-06
docs: add documentation for WP CLI commands

dkoo:
### All Submissions:

Adds missing documentation for custom WP CLI commands registered by this plugin and standardizes documentation for all commands into the callback methods' docblocks for use with the global `--help` flag. 

Test each of the commands with the `--help` flag and confirm that the output accurately describes the script's purpose and available params, e.g.

Screenshot 2024-11-11 at 4 00 16 PM

Commands covered: 

- `wp newspack ras setup`
- `wp newspack verify-reader`
- `wp newspack migrate-co-authors-guest-authors`
- `wp newspack backfill-non-editing-contributors`
- `wp newspack schedule-co-authors-author-term-backfill`
- `wp newspack esp sync`

### Other information:


includes/cli/class-co-authors-plus.php
```diff
@@ -79,20 +79,6 @@ public static function unschedule_author_term_backfill() {
...
-	 *
-	 * [--verbose]
-	 * : Produce more output.
-	 *
-	 * [--user_logins]
-	 * : Comma-separated list of user logins. If provided, only WP Users with these logins will be processed.
-	 *
-	 * [--guest_author_ids]
-	 * : Comma-separated list of Guest Author IDs. If provided, only Gues Authors with these IDs will be processed.
-	 *
```
adekbadek: Removing inline documentation makes the `--help` flag (`wp newspack migrate-co-authors-guest-authors --help`) unhelpful:

| Before | After | 
|--------------|-----------|
| image | image |

I'm not a fan of decoupling the docs from the code. When all commands are properly documented inline, WP CLI provides an excellent summary by itself (`wp newspack --help`), so there should be no need for a README:

image

adekbadek (changes requested):

includes/cli/class-co-authors-plus.php
dkoo: Ah, good point! Didn't realize that those inline docblocks were actually picked up by WP CLI. I'll move the documentation from the README to the docblocks instead.

dkoo: @adekbadek a78f11b standardizes the use of docblocks instead of a separate README.md file.

adekbadek (approved):

PR #3534, opened 2024-11-07
fix(wizard-admin-menu-priority): reinstate admin menu action hook priority

ronchambers:
### All Submissions:

We lost the ability to set the `admin_menu` action hook priority in the parent Wizard class.  This PR reinstates that ability.

Somewhere in the midst of the [previous two merges](https://github.com/Automattic/newspack-plugin/commits/epic/ia/includes/wizards/class-wizard.php) the variable `$menu_priority` was removed, but the variable is still needed for the priority for the `admin_menu` hook.

I found this issue because the parent menu for "Advertising > Sponsors" wasn't being expended properly locally but on the staging site the menu was fine.

**Local (prior to this PR fix):**

image

**Staging site is currently working:**

image

### Variable renamed:

To be a bit most specific about what the variable does, I renamed it from `menu_priority` to `admin_menu_hook_priority`.  It's a bit wordy, but I wanted to differentiate it from the variables that are used by the `custom_menu_order`.

I also looked through the previous commits and added the variable back to the classes that had it prior.  Basically any class that was using it with a priority above 10.  For the classes that had the priority at 1 (newspack-dashboard) or 2 (newspack-settings), I didn't add it back since 2 is already the default.

For `includes/wizards/advertising/class-advertising-display-ads.php` I did keep the 2 in the code with a note: `* Important: keep this at 2 otherwise parent menu won't expand/highlight for class-advertising-sponsors.php`

1. Pull branch on top of `epic/ia`.
2. Use WP-Admin in browser to verify menu functions correctly for IA related items.

### Other information:

### DIFF
--- a/includes/wizards/advertising/class-advertising-sponsors.php
+++ b/includes/wizards/advertising/class-advertising-sponsors.php
@@ -48,6 +48,13 @@ class Advertising_Sponsors extends Wizard {
 	protected $capability = 'manage_options';
 
 	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 99;
+
+	/**
 	 * Advertising_Sponsors Constructor.
 	 */
 	public function __construct() {
--- a/includes/wizards/class-components-demo.php
+++ b/includes/wizards/class-components-demo.php
@@ -31,6 +31,13 @@ class Components_Demo extends Wizard {
 	protected $capability = 'manage_options';
 
 	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 100;
+
+	/**
 	 * Whether the wizard should be displayed in the Newspack submenu.
 	 *
 	 * @var bool.
--- a/includes/wizards/class-wizard.php
+++ b/includes/wizards/class-wizard.php
@@ -61,6 +61,13 @@ abstract class Wizard {
 	public $menu_order = 0;
 
 	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 2;
+
+	/**
 	 * Initialize.
 	 *
 	 * @param array $args Array of optional arguments. i.e. `sections`.
@@ -70,7 +77,7 @@ abstract class Wizard {
 	 * $my_wizard = new My_Wizard( [ 'sections' => [ 'my-wizard-section' => 'Newspack\Wizards\My_Wizard\My_Wizard_Section' ] ] );
 	 */
 	public function __construct( $args = [] ) {
-		add_action( 'admin_menu', [ $this, 'add_page' ] );
+		add_action( 'admin_menu', [ $this, 'add_page' ], $this->admin_menu_hook_priority );
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_scripts_and_styles' ] );
 		if ( isset( $args['sections'] ) ) {
 			$this->load_wizard_sections( $args['sections'] );
--- a/includes/wizards/advertising/class-advertising-display-ads.php
+++ b/includes/wizards/advertising/class-advertising-display-ads.php
@@ -70,6 +70,15 @@ class Advertising_Display_Ads extends Wizard {
 	public $menu_order = 4;
 
 	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * Important: keep this at 2 otherwise parent menu won't expand/highlight for class-advertising-sponsors.php
+	 * 
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 2;
+
+	/**
 	 * Constructor.
 	 */
 	public function __construct() {
--- a/includes/wizards/class-listings-wizard.php
+++ b/includes/wizards/class-listings-wizard.php
@@ -56,6 +56,13 @@ class Listings_Wizard extends Wizard {
 	public $menu_order = 3;
 
 	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 99;
+
+	/**
 	 * Constructor.
 	 */
 	public function __construct() {


includes/wizards/class-components-demo.php
```diff
@@ -30,6 +30,13 @@ class Components_Demo extends Wizard {
 	 */
 	protected $capability = 'manage_options';
 
+	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 100;
+
```
miguelpeixe, line 39: Do we need one for the components demo? This doesn't render a menu item.

includes/wizards/class-wizard.php
```diff
@@ -70,7 +77,7 @@ abstract class Wizard {
 	 * $my_wizard = new My_Wizard( [ 'sections' => [ 'my-wizard-section' => 'Newspack\Wizards\My_Wizard\My_Wizard_Section' ] ] );
 	 */
 	public function __construct( $args = [] ) {
-		add_action( 'admin_menu', [ $this, 'add_page' ] );
+		add_action( 'admin_menu', [ $this, 'add_page' ], $this->admin_menu_hook_priority );
```
miguelpeixe, line 80: If the hook order is supposed to follow the same logic as the `$menu_order`, we could avoid an additional variable and go with:

```suggestion
		add_action( 'admin_menu', [ $this, 'add_page' ], 2 + $this->menu_order );
```

wdyt?

includes/wizards/advertising/class-advertising-sponsors.php
```diff
@@ -47,6 +47,13 @@ class Advertising_Sponsors extends Wizard {
 	 */
 	protected $capability = 'manage_options';
 
+	/**
+	 * Admin Menu Hook Priority when calling 'admin_menu' action hook.
+	 *
+	 * @var int.
+	 */
+	protected $admin_menu_hook_priority = 99;
```
miguelpeixe, line 55: Is 99 an arbitrary high number?

miguelpeixe (changes requested):
Thank you for catching that! A feel comments:

includes/wizards/class-components-demo.php
ronchambers: I don't know...I just added it back to be safe since it was removed.  I'm not sure how to test that.  I think it's part of the old (non IA) wizards.  I could checkout `trunk` or `release` branch locally then try to test it.  Maybe @jaredrethman knows if it's needed?

includes/wizards/class-wizard.php
ronchambers: I think we need to keep them separate.  The `$menu_order` has to do with the `custom_order` of wp-admin display menu order.  The values will be in the range of `0 through 5`.

Whereas `$admin_menu_hook_priority` has to do with when `add_page` function is run.  The values here would be above or below the default action hook priority of `10`.

includes/wizards/class-components-demo.php
miguelpeixe: It's a hidden page accessible only via url (admin.php?page=newspack-components-demo)

includes/wizards/advertising/class-advertising-sponsors.php
ronchambers: Yes, I think 11 would also work.  Anything above the default hook priority of 10 should be fine.  I just left it as 99 since that is what I saw in the prior code.

includes/wizards/class-wizard.php
miguelpeixe: The first wizard has a hook priority of 2, which is why I'm assuming that higher than 2 is sufficient and respects the desired execution logic  for following wizards.

I'm still hesitant to add boilerplate code that might not be necessary. Let's defer to @jaredrethman 😀

jaredrethman:
@ronchambers really well done finding that regression. The weird part I'm finding is that the submenu item is highlighted but the group of submenu items isn't expanding.

Your proposal resolves the issue and reinstates what was there originally - that's great. I just can't help but wonder why we're able to move all those other menu items (specifically the ones with a post type `edit.php` as destination) without causing this issue.

@miguelpeixe your rational of not adding in properties unnecessarily really resonates with the idea of keeping the code readable, simple, maintainable. 

I'd like to spend some time debugging this to understand why other post type menu items are working and see if we can re-align and unify all implementations.

ronchambers:
@jaredrethman - while you're debugging you might want to look at this code block in `wp-admin/menu-header.php`:

https://github.com/WordPress/WordPress/blob/1013750a458823a7c01df8efbfc3eb6b25c639da/wp-admin/menu-header.php#L38-L50

You'll notice that even after the `parent_file` and `submenu_file` filters have fired, the function `get_admin_page_parent()` might "undo" whatever was set by filters (namely the `parent_file` value).  If you step-debug through the function:

https://github.com/WordPress/WordPress/blob/1013750a458823a7c01df8efbfc3eb6b25c639da/wp-admin/includes/plugin.php#L1956

You'll see that there are lots of pathways where the global `$parent_file` will be changed to whatever this function thinks it should be 😢 It seems to me that core should respect the original `apply_filters( 'parent_file', ...` return value and not do this extra "override" function...  But that doesn't seem to be the case.

As you look into the many pathways, you'll see checks like `if ( 'admin.php' === $pagenow...` which is why, I believe, we get different parent file highlighting/expanding in the wp-admin menu based on if the URL `wp-admin/{file_name}.php` contains `edit.php` vs `admin.php`.   

Also of note is that when a submenu page is "hidden" like in the following code:

```
add_submenu_page(
    '', // hidden.
    ...
```

The global `$submenu` will store that submenu item into a "blank" index like `$submenu[''] = array( hidden items )`.  So based on if something is hidden or not, could produce a different "pathway" result in the `get_admin_page_parent()` function.

The only way I could get all this to work for Network Wizard's "hidden" Settings pages is:

1) Create a pathway that doesn't change `$parent_file`.  See code in the Network Settings draft PR here: https://github.com/Automattic/newspack-plugin/blob/a9d7592df2e419fb0587fa34cfdd2ce60c7b77c8/includes/wizards/class-network-wizard.php#L333

2) Also of note is that when hiding menu items using the `add_submenu_page( `` // hidden ...)` technique, this can also cause the HTML Title to get "lost".  See this note in the same draft PR: https://github.com/Automattic/newspack-plugin/blob/a9d7592df2e419fb0587fa34cfdd2ce60c7b77c8/includes/wizards/class-network-wizard.php#L345

So....with all that said, if you can't find a suitable solution, possibly we don't use `add_submenu_page( `` // hidden ...)` technique and instead just hide the menu items using CSS? This could avoid the two issues I linked to in my draft PR and also it might make it easier for the `parent_file` and `submenu_file` filters to work since the `get_admin_page_parent()` function won't go down the wrong path (hopefully) 😃

jaredrethman:
@ronchambers spent quite a bit of time looking into this and I think what you've done here is the way to go - still don't know exactly why adding a priority to that hook makes it work (even if `2` without a property it appears to work). Not worth the all the extra code to make this work without. You got my 👍 

Feel like this code base needs a stable API to standardize menus+pages within and across plugins.

ronchambers:
I...just...couldn't...let it go...   Since the Advertising menu was the basis for the other menus (Newsletters, Listings, Network), I needed to understand what the issue was.  If/when we create a standardized API for this we'll have to know.  So I got it figured out 😃 

New draft PR here: https://github.com/Automattic/newspack-plugin/pull/3535

PR #3541, opened 2024-11-12
Tests/remove network dependent

adekbadek:
### All Submissions:

Removes and adjusts unit tests, so they are not network-dependent. Unit tests should not rely on remote services and servers.

1. Observe the tests are passing on CI and locally

### Other information:


tests/unit-tests/api-plugins-controller.php
```diff
@@ -85,159 +82,6 @@ public function test_get_plugins_authorized() {
 		$this->assertEquals( $expected_jetpack_info, $data['jetpack'] );
 	}
 
-	/**
```
leogermani, line 88: afaik, all these tests are not making any requests. `server->dispatch` is emulating the server receiving a request

leogermani: It's the way to test Rest API endpoints without needing http requests

tests/unit-tests/emails.php
```diff
@@ -51,26 +51,6 @@ private static function get_test_email( $type ) {
 	 * Email setup & defaults generation.
 	 */
 	public function test_emails_setup() {
-		self::assertEquals(
-			Emails::get_emails( [ 'test-email-config' ] ),
-			[],
-			'Emails are empty until configured.'
-		);
-		self::assertFalse(
-			Emails::can_send_email( 'test-email-config' ),
```
leogermani, line 60: can we short circuit `wp_mail` to make sure no requests are made?

tests/unit-tests/api-plugins-controller.php
adekbadek: It's not about testing the REST API endpoints. The endpoint _handler_ is making requests to the plugin repository to fetch plugin code.

tests/unit-tests/emails.php
adekbadek: This is not about `wp_mail`. The issue with this test is that it installed `newspack-newsletters` from the plugin repository. I've replaced it with a mock of `Newspack_Newsletters` class, and to simplify removed the test assertions related to the plugin being inactive. This is not ideal, I know, but I don't think making it ideal is worth the effort.

tests/unit-tests/api-plugins-controller.php
leogermani: gotcha

leogermani (approved):

PR #3545, opened 2024-11-13
fix(woocommerce-emails): use the default email if there are no donation products

adekbadek:
### All Submissions:

Fixes the custom receipt emails, so that they are not sent for order which do not contain donation products.

1. Ensure you have donation products set up, and at least one non-donation product 
2. Ensure the custom receipt email is enabled in Reader Revenue wizard -> Emails
3. Donate and observe the custom email is sent
4. Buy a non-donation product and observe a default WC email is sent

### Other information:

---
  - https://app.asana.com/0/0/1208741538452766

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -267,14 +267,7 @@ class WooCommerce_Connection {
 	 */
 	public static function send_customizable_receipt_email( $enable, $order, $class ) {
 		// If there are no donation products in the order, do not override the default WC receipt email.
-		$has_donation_product = false;
-		$donation_product_ids = array_values( \Newspack\Donations::get_donation_product_child_products_ids() );
-		foreach ( $order->get_items() as $item_id => $item ) {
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-connection.php
```diff
@@ -266,6 +266,19 @@ public static function force_disable_order_attribution( $should_allow ) {
 	 * @return bool
 	 */
 	public static function send_customizable_receipt_email( $enable, $order, $class ) {
+		// If there are no donation products in the order, do not override the default WC receipt email.
+		$has_donation_product = false;
```
leogermani: In other places we check this using 

```
$product_id = \Newspack\Donations::get_order_donation_product_id( $order_id );
```

which is basically the same thing you do here. Isn't it simples if we just use that?

adekbadek: Huh, thanks! Changed in b3e62f020de78e416dd5efaa905a7c2ee6ee09bb

leogermani (approved):

PR #3549, opened 2024-11-15
feat(ia): audience campaigns

miguelpeixe:
### All Submissions:

1205919985867982-as-1206833421212801/f

Moves the Campaigns wizard to Audience -> Campaigns.

image

image

image

image

image

#### Campaigns tab

1. Confirm you have RAS setup (via Audience -> Configuration)
2. Checkout this branch, https://github.com/Automattic/newspack-popups/pull/1365, and visit Audience -> Campaigns
3. Confirm the Campaigns tab renders as expected (see image above)
4. Click "Preview Segment" on any of the segments and confirm the preview renders with the expected prompts
5. Click on a prompt settings button and confirm the modal renders
6. Tweak the settings, save, and confirm the changes persist
7. Click a prompt's "more" button, then "Edit"
8. Confirm the editor renders with the correct sidebar menu active
9. Back to the wizard, click on a prompt's "more" button, then "Preview", and confirm the prompt preview renders as expected
10. Do the same with "Duplicate", "Deactivate", and "Delete"
11. Click "Add New Campaign", add the campaign, and confirm the campaign is created
12. Add a new prompt to a segment, confirm the correct menu is active, save the prompt, and go back to the wizard
13. Confirm the created prompt renders for the campaign
14. Switch the view from "Active Prompts" to all the other options and confirm the filters are applied correctly

#### Segments tab

1. Switch to the Segments tab and confirm it renders as expected (see image above)
2. Toggle segments and confirm the changes persist
3. Use the chevrons to move their priority and confirm it also persists
4. Edit a segment and confirm the changes persist
5. Create a new segment and confirm it's created
6. Delete the created segment and confirm it's deleted

#### Analytics tab

1. Confirm the tab renders as expected and the links go to the correct destination

#### Settings tab

1. Confirm the tab renders as expected (image above)
2. Change all the options to something else and save
3. Confirm the changes persist

### Other information:


dkoo (changes requested):
@miguelpeixe Functionally, working really well! I noticed a couple of tiny style bugs, one of which isn't even related to these changes (but since we're here, may as well fix it):

1. If a segment contains no prompts in the current view, the styles look a little off. Looks like we're missing some padding, compared to `trunk`:

| `trunk` | this branch |
| --- | --- |
| Screenshot 2024-11-19 at 4 51 12 PM | Screenshot 2024-11-19 at 4 37 30 PM |

2. This is on `trunk` too, but now's a good time to fix it. The dropdown for the filters seems to be missing some CSS. Note the unnecessary padding in the highlighted/unselected items, and the extra black checkmark in the selected item.

Screenshot 2024-11-19 at 4 43 39 PM



Otherwise looking really clean 👍

miguelpeixe:
Thank you for the review, @dkoo!

> If a segment contains no prompts in the current view, the styles look a little off. Looks like we're missing some padding, compared to trunk:

Fixed in a236248778b867e290a35bee660795820b7daaee:

image

> This is on trunk too, but now's a good time to fix it. The dropdown for the filters seems to be missing some CSS. Note the unnecessary padding in the highlighted/unselected items, and the extra black checkmark in the selected item.

Good call! Fixed in 50342d9d832989e199735d6641d80fbc15fd6194

image

I've decided to use the component's native icon with adjusted color instead of our custom pseudo-element. I also added a min-width to the list so it looks better when the selected filter is a shorter word.

includes/wizards/audience/class-audience-campaigns.php
```diff
@@ -63,4 +114,929 @@ public function add_page() {
...
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		return $this->api_get_settings();
+	}
+
+	/**
+	 * Update settings for a Pop-up.
```
dkoo:
```suggestion
	 * Update settings for a prompt.
```

ABSOLUTELY UNNECESSARY NIT: is now a good time to do the ol' Find/Replace to rename `pop-up` to `prompt`? 😄

dkoo (approved):
Looking really nice—thanks for the updates, @miguelpeixe!

includes/wizards/audience/class-audience-campaigns.php
miguelpeixe: Can't think of a better time! Done in 3a06b084b3b99934c97231d85e7288e00c434711.

I've left the CM methods unchanged since they are wrappers for the plugin's methods. The plugin still uses `popup`.

I also removed the `api_set_popup_terms()` which was unused (also unused in `trunk`).

dkoo (approved):

PR #3555, opened 2024-11-18
feat(wc): duplicate orders admin notice

adekbadek:
### All Submissions:

Adds an admin notice when duplicate orders are detected:

image

1. Donate two times with the same amount, within 10 minutes
2. Do that again, but change the dates of these two to more than a month ago
3. Trigger the CRON job: `wp cron event run newspack_wc_check_order_duplicates`
4. Visit the admin panel, observe the admin notice about potentially duplicated orders – from the last month (the first two created)
5. Dismiss one of the cases, observe the page reloads and the list item is no longer displayed
6. Run `wp newspack detect-order-duplicates --cutoff_time='1 year' --save` and refresh the admin page
7. Observe three items are present on the list: two new ones from more than a month ago

### Other information:

---
  - https://app.asana.com/0/0/1208711277186607

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
@@ -40,16 +40,26 @@ class WooCommerce_Duplicate_Orders {
 	 * Duplicates will be detected if it the same amount, same day, from the same customer.
 	 *
 	 * @param number $cutoff_time The cutoff time in the past (how many seconds ago).
+	 * @param number $current_page Current page of results.
+	 * @param array  $results Results to be merged with new results.
 	 */
-	private static function get_order_duplicates( $cutoff_time ): array {
-		$orders = wc_get_orders(
+	public static function get_order_duplicates( $cutoff_time, $current_page = 0, $results = [] ): array {
+		$per_page = 100;
+		$order_result = wc_get_orders(
 			[
-				'limit'          => -1,
+				'paginate'       => true,
+				'limit'          => $per_page,
 				'status'         => [ 'wc-completed' ],
+				'offset'         => $current_page * $per_page,
 				'date_completed' => '>' . ( time() - $cutoff_time ),
 			]
 		);
 
+		if ( defined( 'WP_CLI' ) && WP_CLI && $order_result->max_num_pages > 0 ) {
+			\WP_CLI::line( sprintf( 'Processing page %d/%d of orders.', $current_page + 1, $order_result->max_num_pages ) );
+		}
+
+		$orders = $order_result->orders;
 		$order_duplicates = [];
 
 		foreach ( $orders as $order ) {
@@ -76,8 +86,6 @@ class WooCommerce_Duplicate_Orders {
 			$order_duplicates[ $email ][ $amount ][ $date ][] = $order->get_id();
 		}
 
-		$results = [];
-
 		foreach ( $order_duplicates as $email => $amounts ) {
 			foreach ( $amounts as $amount => $dates ) {
 				foreach ( $dates as $date => $order_ids ) {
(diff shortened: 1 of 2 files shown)


leogermani (commented):
Even though the SQL queries are pretty efficient, they are huge, hard to read and hard to maintain.

And also hard to test, because I need two different sites to test it.

Doing this with simple code, looping though recent orders would be less efficient, but much more readable, maintainable and testable... And considering that this check runs daily, we don't need to look at all the orders in the history, just the most recent ones.

adekbadek:
Yeah, that makes sense. I've updated with the following:

1. Instead of SQL query, iterate order objects retrieved with `wc_get_orders` (in batches of 100, so the site does not fall over)
2. Added a `wp newspack detect-order-duplicates` CLI command to enable filling in historical data (older than a month). This will have to be ran on live sites after the PR is deployed.

Update the PR description accordingly.

includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
```diff
@@ -0,0 +1,247 @@
...
+			$dismissed_duplicates[] = $_POST['dismiss_order_ids']; // phpcs:ignore WordPress.Security.NonceVerification.Missing, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			update_option( self::DISMISSED_DUPLICATES_OPTION_NAME, $dismissed_duplicates );
+			// Refresh the page to reflect changes.
+			wp_safe_redirect( isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : admin_url() );
+			exit;
+		}
+	}
+
+	/**
+	 * CLI handler to upsert the order duplicates with a specified timeframe.
```
leogermani: is it detect or upsert?

includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
```diff
@@ -0,0 +1,247 @@
...
+	 *
+	 * [--cutoff_time=<time-string>]
+	 * : The cutoff time in the past (e.g. "2 months").
+	 *
+	 * [--save]
+	 * : Whether to save the results for display in the admin panel.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     wp newspack detect-order-duplicates --cutoff_time='2 months' --save
```
leogermani, line 230: is it detect or upsert?

includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
leogermani: Let's add a clearer description of what the command does
- search for duplicates and optionally store this info to be displayed in the admin panel

leogermani (commented):
Works as described! Thanks for the changes! I left a few minor comments.

I just noticed one thing. I first ran the CLI command without the `--save` param. It detected the duplicated but didn't save. However, it ended up displaying the admin notice, but with no orders listed. Just the summary. When you clicked, it was an empty list.

Another question: Since we're running this daily, and cutting off based on completion date, why default to check the last 30 days? Looks like we will be doing a lot of unnecessary checks every day. Isn't it enough to check just the last one or two days?

includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
adekbadek: Without `--save`, it's only detection. With it, it's upsertion.

includes/reader-revenue/woocommerce/class-woocommerce-duplicate-orders.php
adekbadek: Updated in 520f49c1367f00aebf272734c855943de2c7080a

adekbadek:
> However, it ended up displaying the admin notice, but with no orders listed.

Yeah, I failed to account for an empty state, thanks for catching that! Fixed in 520f49c1367f00aebf272734c855943de2c7080a

> Isn't it enough to check just the last one or two days?

That's right, changed to a day in 520f49c1367f00aebf272734c855943de2c7080a

leogermani (approved):

PR #3560, opened 2024-11-20
feat(subscriptions): add setting to reattempt payment after final retry

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208702687864922/1208702687865959/f

This PR adds a new 'On-hold duration' setting for subscriptions. This setting adds a new retry rule set to the specified number of days in WooCommerce > Settings > Subscriptions > On-hold Duration:

Screenshot 2024-11-25 at 15 53 34

1. As admin, go to WooCommerce > Settings > Subscriptions and scroll down. Confirm there is a new section added for Newspack subscriptions settings. Also confirm the On-hold Duration setting is present.
2. Select a non-zero value and save. Confirm the setting persists on reload
3. As a reader, purchase any subscription product
4. As admin, go to the edit subscriptions page for this subscription (WooCommerce > Subscriptions)
5. Follow these steps to manually trigger 5 payment retries: https://woocommerce.com/document/subscriptions/develop/failed-payment-retry/#section-7
6. Run one more time, verify a sixth job is scheduled set to run the number of days set in the On-hold Duration setting:
Screenshot 2024-11-20 at 17 06 12
7. Visit the edit subscription page for the subscription. Confirm the subscription is still on-hold. Also confirm the renewal order is still in a pending status:
Screenshot 2024-11-20 at 17 10 12
8. Run the scheduled action one more time. Confirm no new scheduled action is added (this should be the final retry rule)
9. Go back to the edit subscription page and verify the subscription is now expired and the renewal order status is set to failed:
Screenshot 2024-11-20 at 17 12 03
10. Set the On-hold duration setting to 0 and repeat the above steps. This time the fifth scheduled action should trigger the subscription to expire and the renewal order to fail.

### Other information:

### DIFF
--- a/includes/plugins/class-woocommerce-subscriptions.php
+++ b/includes/plugins/class-woocommerce-subscriptions.php
@@ -17,7 +17,8 @@ class WooCommerce_Subscriptions {
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
-		add_filter( 'woocommerce_subscription_settings', [ __CLASS__, 'add_post_retry_payment_attempt_setting' ], 11, 1 );
+		add_filter( 'woocommerce_subscription_settings', [ __CLASS__, 'add_post_retry_setting' ], 11, 1 );
+		add_filter( 'wcs_default_retry_rules', [ __CLASS__, 'maybe_apply_post_retry' ], 99, 1 );
 	}
 
 	/**
@@ -36,7 +37,7 @@ class WooCommerce_Subscriptions {
 	 *
 	 * @return array
 	 */
-	public static function add_post_retry_payment_attempt_setting( $settings ) {
+	public static function add_post_retry_setting( $settings ) {
 		if ( self::is_active() ) {
 			return array_merge(
 				$settings,
@@ -50,9 +51,9 @@ class WooCommerce_Subscriptions {
 					[
 						'name'              => __( 'Post-retry Payment Attempt', 'newspack-plugin' ),
 						'desc'              => __( 'The number of days after final retry to reattempt payment when automatic retries are enabled', 'newspack-plugin' ),
-						'id'                => 'newspack_subscriptions_post_retry_payment_attempt',
+						'id'                => 'newspack_subscriptions_post_retry_days',
 						'css'               => 'max-width:80px;',
-						'value'             => self::get_post_retry_payment_attempt(),
+						'value'             => self::get_post_retry_days(),
 						'type'              => 'number',
 						'custom_attributes' => array(
 							'min'  => 0,
@@ -74,8 +75,32 @@ class WooCommerce_Subscriptions {
 	 *
 	 * @return int
 	 */
-	public static function get_post_retry_payment_attempt() {
-		return absint( get_option( 'newspack_subscriptions_post_retry_payment_attempt', 0 ) );
+	public static function get_post_retry_days() {
+		return absint( get_option( 'newspack_subscriptions_post_retry_days', 0 ) );
+	}
+
+	/**
+	 * Conditionally adds post-retry rule to retry rules.
+	 *
+	 * @param array $retry_rules Subscriptions retry rules.
+	 */
+	public static function maybe_apply_post_retry( $retry_rules ) {
+		if ( self::is_active() ) {
+			$post_retry_days = self::get_post_retry_days();
+			if ( $post_retry_days > 0 ) {
+				$final_retry_rule                                    = end( $retry_rules );
+				$post_retry_rule                                     = $final_retry_rule;
+				$final_order_status                                  = $final_retry_rule['status_to_apply_to_order'];
+				$final_subscription_status                           = $final_retry_rule['status_to_apply_to_subscription'];
+				$final_retry_rule['status_to_apply_to_order']        = 'pending';
+				$final_retry_rule['status_to_apply_to_subscription'] = 'on-hold';
+				$post_retry_rule['status_to_apply_to_order']         = $final_order_status;
+				$post_retry_rule['status_to_apply_to_subscription']  = $final_subscription_status;
+				$post_retry_rule['retry_after_interval']             = DAY_IN_SECONDS * $post_retry_days;
+				return array_merge( $retry_rules, [ $post_retry_rule ] );
+			}
+		}
+		return $retry_rules;
 	}
 }
 WooCommerce_Subscriptions::init();
--- a/includes/plugins/class-woocommerce-subscriptions.php
+++ b/includes/plugins/class-woocommerce-subscriptions.php
@@ -17,8 +17,8 @@ class WooCommerce_Subscriptions {
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
-		add_filter( 'woocommerce_subscription_settings', [ __CLASS__, 'add_post_retry_setting' ], 11, 1 );
-		add_filter( 'wcs_default_retry_rules', [ __CLASS__, 'maybe_apply_post_retry_rule' ], 99, 1 );
+		add_filter( 'woocommerce_subscription_settings', [ __CLASS__, 'add_on_hold_duration_setting' ], 11, 1 );
+		add_filter( 'wcs_default_retry_rules', [ __CLASS__, 'maybe_apply_on_hold_duration_rule' ], 99, 1 );
 	}
 
 	/**
@@ -31,13 +31,13 @@ class WooCommerce_Subscriptions {
 	}
 
 	/**
-	 * Add post-retry payment attempt setting.
+	 * Add on-hold duration setting.
 	 *
 	 * @param array $settings Subscription settings.
 	 *
 	 * @return array
 	 */
-	public static function add_post_retry_setting( $settings ) {
+	public static function add_on_hold_duration_setting( $settings ) {
 		if ( self::is_active() ) {
 			return array_merge(
 				$settings,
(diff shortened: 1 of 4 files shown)


chickenn00dle:
Removing from review after feedback in 1208702687864922-as-1208702687865959/f

chickenn00dle:
Ready for review again!

includes/plugins/class-woocommerce-subscriptions.php
```diff
@@ -0,0 +1,113 @@
...
+		if ( self::is_active() && count( $retry_rules ) > 0 ) {
+			$on_hold_duration = self::get_on_hold_duration();
+			if ( 0 < $on_hold_duration ) {
+				$retry_rules[] = [
+					'retry_after_interval'            => $on_hold_duration * DAY_IN_SECONDS,
+					'status_to_apply_to_order'        => 'pending',
+					'status_to_apply_to_subscription' => 'on-hold',
+				];
+			}
+			$retry_rules[] = [
```
leogermani: have you considered simply adding one rule with `$on_hold_duration` as interval and `expired` as status, instead of two rules?

And if `$on_hold_duration` is zero, change the status of the last rule?

I ask this because I think it would look a little cleaner in the admin. In my case, my on hold period was 3 days:

image

It's probably because I'm rushing the scheduled actions, but the last action shows out of the order, as if it happened before. Also, there is one unnecessary additional attempt to process the payment right after one failed.

includes/plugins/class-woocommerce-subscriptions.php
```diff
@@ -0,0 +1,113 @@
...
+		return absint( get_option( 'newspack_subscriptions_on_hold_duration', 0 ) );
+	}
+
+	/**
+	 * Conditionally adds on-hold duration rule to retry rules.
+	 *
+	 * @param array $retry_rules Subscriptions retry rules.
+	 */
+	public static function maybe_apply_on_hold_duration_rule( $retry_rules ) {
+		if ( self::is_active() && count( $retry_rules ) > 0 ) {
```
leogermani: What are we checking with `count( $retry_rules )` ? In which case there will be no rules? If another plugin is tweaking it? Or if retries are disabled? If the latter, is there a more direct check for that option?

leogermani (commented):
Works like a charm! I also confirmed that this works with pre-existing subscriptions

I left two non blocking questions. But before we merge it, let's check if we are ok to shipping this or if we will want to give publishers a heads up and maybe hold until they are ready to receive this change.

includes/plugins/class-woocommerce-subscriptions.php
chickenn00dle: > have you considered simply adding one rule with $on_hold_duration as interval and expired as status, instead of two rules?

The problem with this is these retry rules are parsed at the time of failed payment before the actual retry happens which means the status for a given retry rule will be applied at the time the rule is created and before the retry is actually triggered. So adding a single rule with `$on_hold_duration` as interval and `expired` as status would result in the subscription being set to `expired` on the fifth retry with the sixth retry attempt doomed to fail since the subscription has already been set to `expired`.

I hope this makes sense. Let me know if you'd like to huddle to talk it out.

includes/plugins/class-woocommerce-subscriptions.php
chickenn00dle: > What are we checking with count( $retry_rules ) ? In which case there will be no rules? If another plugin is tweaking it? Or if retries are disabled? If the latter, is there a more direct check for that option?

I added the check mostly to account for the possibility of another plugin tweaking the rules. This particular hook would never be triggered if retries are disabled.

That said, I realize now that I had been assuming we enforce retries to be enabled for our sites. If this isn't the case, we would also need to force subscriptions to expire when this isn't enabled I think?

chickenn00dle:

> let's check if we are ok to shipping this or if we will want to give publishers a heads up and maybe hold until they are ready to receive this change.

Should I maybe create a new epic branch for this?

includes/plugins/class-woocommerce-subscriptions.php
leogermani: Ah, gotcha, that makes sense

leogermani:
> Should I maybe create a new epic branch for this?

Let's ask Katie before.

Can I suggest one last thing? I see that in https://github.com/Automattic/newspack-plugin/pull/3568 you are creating the same file. WDYT of creating a folder instead and having one file for each feature.

chickenn00dle:
> I see that in https://github.com/Automattic/newspack-plugin/pull/3568 you are creating the same file. WDYT of creating a folder instead and having one file for each feature.

IMO, this isn't necessary quite yet since we aren't adding too much and I feel a single subscriptions integration class is enough to house all of the logic for both `on-hold duration` and cancellation reason metadata, at least until we end up adding more functionality. 

That said, if you prefer we divide this further, I can make the change!

leogermani:
> That said, if you prefer we divide this further, I can make the change!

I always prefer small files over large files, small methods over large methods, etc :) 

Also, it makes the history looks better when, in the future, we finally decide to break things into more than one file and all the git history is less accessible.

Why save on files? :)

leogermani:
ah, and let's do a feature flag instead of an epic? based on our recent experiences? wdyt?

chickenn00dle:
> Also, it makes the history looks better when, in the future, we finally decide to break things into more than one file and all the git history is less accessible.

Sounds good!

> ah, and let's do a feature flag instead of an epic? based on our recent experiences? wdyt?

Great idea!

I've reorganized the code and added a feature flag in https://github.com/Automattic/newspack-plugin/pull/3560/commits/612c960fdc2590267d2ddfb52566d1b3daae8b41

Note I've also added a check for reader activation being enabled which I realized I forgot to include in this PR. 

Let me know if this works!

includes/plugins/class-woocommerce-subscriptions.php
leogermani: > If this isn't the case, we would also need to force subscriptions to expire when this isn't enabled I think?

I don't think we enforce that. That's a good question. Let's open a ticket for that, as Katie and handle this in another PR if needed

leogermani (approved):

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * Check if WooCommerce Subscriptions Integration is enabled.
+	 *
+	 * Enalbed if Reader activation is enabled and the feature flag is defined.
+	 *
+	 * @return bool
+	 */
+	public static function is_enabled() {
+		return Reader_Activation::is_enabled() && defined( self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG );
```
leogermani: you also need to check if the value of the constant is true? It can be defined as false

leogermani (changes requested):

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
chickenn00dle: Good catch! Added in https://github.com/Automattic/newspack-plugin/pull/3560/commits/31f273ce720da5d2572bb100d39055e7dc6d8666

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * Check if WooCommerce Subscriptions Integration is enabled.
+	 *
+	 * Enalbed if Reader activation is enabled and the feature flag is defined.
+	 *
+	 * @return bool
+	 */
+	public static function is_enabled() {
+		return Reader_Activation::is_enabled() && defined( self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG ) && self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG;
```
leogermani: this check doesn't look right.

`defined( self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG )` looks ok, because the class constant is a string and it will check for a global constant named after that string

but `self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG` will resolve as a string and thus always return `true`. 

Any reason to have this proxy constant in the class and not check the global constant directly?

If you want to keep it (no strong opinions here), I think you need to use `constant( self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG )`

chickenn00dle: > but self::NEWSPACK_SUBSCRIPTIONS_EXPIRATION_FEATURE_FLAG will resolve as a string and thus always return true.

Ah. Whoops! That's my bad. Shouldn't have rushed that last commit.

> Any reason to have this proxy constant in the class and not check the global constant directly?

Not really, aside from explicitly defining the feature flag in the code for search-ability and future reference. That said, its fairly self-contained so removed in https://github.com/Automattic/newspack-plugin/pull/3560/commits/89dea71ea838b1dce1ecb05fd9e2d899fa97ef9c

leogermani (approved):

includes/plugins/woocommerce-subscriptions/class-on-hold-duration.php
```diff
@@ -0,0 +1,103 @@
...
+					'retry_after_interval'            => $on_hold_duration * DAY_IN_SECONDS,
+					'status_to_apply_to_order'        => 'pending',
+					'status_to_apply_to_subscription' => 'on-hold',
+				];
+			}
+			$retry_rules[] = [
+				'retry_after_interval'            => 0,
+				'status_to_apply_to_order'        => 'failed',
+				'status_to_apply_to_subscription' => 'expired',
+			];
```
claudiulodro, line 99: Just double checking, if I've read this right it looks like the expiration date will be the date of the last failed payment retry? If so, that is the behavior publishers want!

chickenn00dle: That's correct! Sorry for the late reply. Totally missed this comment.

PR #3562, opened 2024-11-20
refactor(ia): advertising display ads

jaredrethman:
### All Submissions:

Refactors Advertising slugs. The initial intention was to structure slugs as a representation of location i.e. `{top-level-admin-page-name}-{submenu-name}`. 
* Newspack / Dashboard - `newspack-dashboard`
* Newspack / Settings - `newspack-settings`
* Advertising / Display Ads - `advertising-display-ads`

As the project progressed a similar pattern was followed with the addition of a `newspack-` prefix.
This PR introduces `newspack-` prefix to Advertising. 

1. Checkout branch and build assets - `git checkout origin/feat/ia-consistent-slugs && npm run build`
2. Navigate to _/wp-admin/admin.php?page=newspack-advertising-display-ads#/_
3. Ensure Advertising / Display Ads loads and functions as expected.
4. Navigate to _/wp-admin/admin.php?page=newspack-listings-settings_
5. Ensure Listings / Settings loads and functions as expected.

### Other information:


ronchambers:
We might want to ignore the Ad Blocker fix at this time.  We could just make a rule that publishers must turn off Ad Blockers for the wp-admin.  Otherwise, I've tested that the following slugs will work:

`newspack-ads-display-ads`
`newspack-ads-wizard`

We just can't have 'advertising' in the slug because this causes the menu link class to become:

`<a class="wp-has-submenu wp-has-current-submenu wp-menu-open menu-top toplevel_page_newspack-advertising-display-ads" ...`

which is what DuckDuckGo blocker seems to look at.  But even if we change the slug, this only fixes the DuckDuckGo blocker.  There is no guarantee it will work for other blockers.

I don't see a way in menu-header.php to filter the classes, only [add additional classes](https://github.com/WordPress/WordPress/blob/ff3fde39ee4de6893beb59d87afb3d6ac63446ad/wp-admin/menu-header.php#L106).

I say let's not worry about it.  Publishers can turn off add blockers if they don't see the "Advertising" menu icon :-)

(Other issues I'll add below in a few minutes...)

ronchambers:
@jaredrethman I previously captured some mismatched naming issues.  I'll list them here if you're able to include in this PR?

1)

https://github.com/Automattic/newspack-plugin/blob/d6c79d89c7935bc84a469d460f0c98219670eb5c/includes/configuration_managers/class-newspack-ads-configuration-manager.php#L190

```
- return get_option( Advertising_Wizard::NEWSPACK_ADVERTISING_SERVICE_PREFIX . $service, false );
+ return get_option( Advertising_Display_Ads::NEWSPACK_ADVERTISING_SERVICE_PREFIX . $service, false );
```

2)

https://github.com/Automattic/newspack-plugin/blob/d6c79d89c7935bc84a469d460f0c98219670eb5c/includes/wizards/class-setup-wizard.php#L624-L630

Same as item 1, three lines need class name changed.

3)

https://github.com/Automattic/newspack-plugin/blob/d6c79d89c7935bc84a469d460f0c98219670eb5c/src/wizards/advertising/components/ad-unit-size-control/index.js#L23

I think that line is OK.  

But this line may need to change:

https://github.com/Automattic/newspack-plugin/blob/d6c79d89c7935bc84a469d460f0c98219670eb5c/src/wizards/advertising/components/ad-unit-size-control/index.js#L90

4)

You may want to remove the old `advertising` wizard from the class-wizards.php file and also delete the `includes/wizards/class-advertising-wizard.php` file itself so it's not autoloaded.  I assume those files will be deleted in the future anyway, so it might be good to do that now so that when testing now, it will be the same testing like when the files are removed.  I also view the debug.log alot when testing.  That's how I found the class name issues (1 and 2) above.  Hope that helps!

ronchambers:
As for the Listings slug change, my concern would be that the url changes if a website is using the Listings plugin with or without the Newspack plugin.  In the case where the listings plugin is used alone the url is:

`/wp-admin/admin.php?page=newspack-listings-settings-admin`

When the Newspack plugin is activated, that page will then error with message `Sorry, you are not allowed to access this page.`, since the `-admin` was removed.  (This would happen if someone had a bookmark or clicked a link in an old help document or something).

I assume the odds of someone running Listings without the Newspack plugin is most likely really rare, so this case would probably never happen, but in general my preference would be to only change the urls if it's really needed.  

Up to you ;-)

ronchambers (changes requested):
Reviewed.  Pleases see comments above.  Thanks!

jaredrethman:
@ronchambers appreciate the review and thorough feedback. I've gone with all your feedback. 

> I'll list them here if you're able to include in this PR?

[feat(ia): pr feedback #issuecomment-2492010662](https://github.com/Automattic/newspack-plugin/pull/3562/commits/7cdb2ad64bc13705fa1743bd6d6ea7d3f9db9423)

> I assume the odds of someone running Listings without the Newspack plugin is most likely really rare, so this case would probably never happen, but in general my preference would be to only change the urls if it's really needed.

Nice catch with this. This should be done in unison with the same change in Newspack Listing if it's to happen at all, to that end I've reverted the `admin` suffix removal - [revert(ia): admin suffix removal from listings slug](https://github.com/Automattic/newspack-plugin/pull/3562/commits/a5dfeac45155a19a9c1eb31c417031a54bb770f2). It also keeps this PR specific to Advertising refactoring.

> We might want to ignore the Ad Blocker fix at this time. We could just make a rule that publishers must turn off Ad Blockers for the wp-admin.

This was my initial thought when the bug surfaced. I've since changed my mind. Without some ad blocker detection functionality - to tell editors to turn off their ad blockers - I believe we should fix this. I went with `newspack-ads-display-ads` - [fix(ia): duckduckgo blocking display of advertising menu](https://github.com/Automattic/newspack-plugin/pull/3562/commits/bdcf4d8e52a46bfb59b841ec2c76fa334f4d1e82). I've also tested this particular bug on the following Chrome extensions:
* [DuckDuckGo Privacy Essentials](https://chromewebstore.google.com/detail/duckduckgo-privacy-essent/bkdgflcldnnnapblkhphbgpggdiikppg?hl=en-US&utm_source=ext_sidebar).
* [AdBlock](https://chromewebstore.google.com/detail/adblock-%E2%80%94-block-ads-acros/gighmmpiobklfepjocnamgkkbiglidom?hl=en-US&utm_source=ext_sidebar) (This extension had no affect on Advertising)

ronchambers (approved):
AWESOME!  Approved! 🚀

PR #3531, opened 2024-11-07
fix(ia): core update nag margins

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

1205919985867982-as-1208354721118905/f

Fixes the core update nag margins when navigating Newspack wizards for both full wizard pages and header-only:

| | Before | After |
| --- | --- | --- |
| Header-only | image | image |
| Wizard page | image | image |

### How to test the changes in this Pull Request:

1. Checkout this branch, navigate to the wizard pages and admin-header-only pages
2. Confirm the margins are correct, as in the images above

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3563, opened 2024-11-21
fix(ia): hide notifications

ronchambers:
### All Submissions:

The current "hide notifications" code needed some fixes to work for the current wizards and new wizards too.  

This new solution moves `remove_notifications()` from the newspack class to the wizard abstract class.

To get an idea of what notifications are being removed, please see [this PR](https://github.com/Automattic/newspack-plugin/pull/3531) that was leaving the notices in place, but just adding margin.  We've now since decided/fixed that we'll actually remove the notices instead.

Another fix/change that was made was to move the rendering hook for the "admin-header-only" wizards to `in_admin_header` instead of `all_admin_notices` because we were removing the action that was also being used for rendering.  Because of this hook change, the location in the HTML where the admin-header was being rendered changed, so we needed to adjust the CSS based on this new HTML location.  Mainly this was needed to fix the mobile layout.

image

Finally, the CSS that originally added margins to the notifications (instead of removing notifications) was allowed to stay in the CSS (with some refactoring).  This is kind of a "fall-back" in case the PHP doesn't hide a notice for some reason it will still look OK on the page.

## Testing:

When testing, no notices should appear on any wizards (except WordPress core notices like "1 post moved to trash" and "saved").  

For Wizards that are "admin-header-only" on CPT pages, notices such as "post moved to trash" and other "bulk/edit" actions will still show.  These are internal to WordPress and provide helpful feedback to users/admins.

image

Settings pages such as Network > Site Role (`admin.php?page=newspack-network`) will show a "saved" notice like the attached since these are rendered by core and are not removed.

image

#### How to test the changes in this Pull Request:

1. Pull this branch.
2. `npm run build`
3. Click on a variety of Wizards pages in wp-admin and verify no notices are shows.

### Other information:


ronchambers:
I'm going to look into this today:

> We could move this PR's php code to the abstract Wizards class so that each wizard could control it's own ability to hide notifications or not. We could set the default to "hide" but wizards could unhide on a per-screen basis. Just an idea if we ever need more granularity for hiding.

ronchambers:
PR write-up above has been updated for moving code into the wizard class.  Ready for review again.

jaredrethman (changes requested):
Appreciate the refactor @ronchambers , this is looking great! One thing I noticed is _this_ change isn't affecting Network Wizards. Looks like the constructor in `Network_Wizard` is missing `parent::__construct();`

ronchambers:
thanks @jaredrethman - the parent construct issue was just fixed in this PR: https://github.com/Automattic/newspack-plugin/pull/3479#issuecomment-2499247829

Once both PRs are merged, the issue will solve itself; the Network Wizard will hide notices.

jaredrethman (approved):
@ronchambers thanks for addressing my feedback. Code is approved!

PR #3568, opened 2024-11-22
feat(subscriptions): add cancellation reason metadata

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208702687864922/1208702687865961

This PR adds meta data to cancelled woo subscriptions and uses this new meta to sync cancellation reason to ESPs for cancelled subscriptions.

Before testing be sure to set the feature flag in wp-config: `define( 'NEWSPACK_SUBSCRIPTIONS_EXPIRATION', true );`. Also ensure Newspack logging is active on the site: `define( 'WP_NEWSPACK_DEBUG', true )`

1. As admin ensure an ESP is connected and sync is working
2. Go to Newspack > Engagement > Show Advanced Settings > Sync contacts to ESP and ensure `Subscription Cancellation Reason` is checked:
Screenshot 2024-11-22 at 16 21 40
3. As a reader purchase a new subscription. **This should be the only active subscription for the reader since our ESP sync will only sync cancelled subscription data if there are no other active subscriptions**
4. As admin, go to the edit subscription page for the above subscription and cancel (WooCommerce > Subscriptions)
5. After page reload scroll down to the custom fields section of the edit subscription page and verify a new cancellation reason field exists with a value of `manually-cancelled`:
Screenshot 2024-11-22 at 14 36 22
6. As a reader again, purchase another subscription
7. Still as reader go to my account and cancel the subscription (My Account > Subscriptions > View then click the cancel button)
8. Via wp cli, update the pending cancellation subscription end date to one minute from now:

```
$subscription = wcs_get_subscription( SUBSCRIPTION_ID );
$subscription->update_dates( array( 'end_date' => "2024-11-22 19:35:01" ) );
```
9. Wait for the end date time to lapse, then visit the site as the reader to ensure the transition from pending cancel to cancel happens
10. As admin, visit the edit subscriptions page for the subscription and verify the cancellation reason meta field exists with a value of `user-cancelled`
Screenshot 2024-11-22 at 14 35 35
 11. Check the logs and ensure the `Subscription Cancellation Reason` field with the expected value is present in the normalized ESP sync data:

Screenshot 2024-12-02 at 19 00 37

### Other information:


includes/plugins/woocommerce-subscriptions/class-subscriptions-meta.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @param string          $from_status   The status the subscription is changing from.
+	 * @param string          $to_status     The status the subscription is changing to.
+	 * @param WC_Subscription $subscription  The subscription object.
+	 */
+	public static function maybe_record_cancelled_subscription_meta( $id, $from_status, $to_status, $subscription ) {
+		if ( ! WooCommerce_Subscriptions::is_active() ) {
+			return;
+		}
+		if ( 'cancelled' === $to_status && ! in_array( $from_status, [ 'cancelled', 'expired' ], true ) ) {
+			$meta_value = is_admin() ? self::CANCELLATION_REASON_ADMIN_CANCELLED : self::CANCELLATION_REASON_USER_CANCELLED;
```
leogermani: We want a third `expired` option here

leogermani (changes requested):

includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -285,6 +285,12 @@ private static function get_order_metadata( $order, $payment_page_url = false )
 					$metadata['product_name'] = reset( $subscription_order_items )->get_name();
 				}
 			}
+
+			// Record the cancellation reason if the subscription was cancelled.
+			$cancellation_reason = $current_subscription->get_meta( WooCommerce_Subscriptions::CANCELLATION_REASON_META_KEY, '' );
```
leogermani: I'm getting an error that I believe is coming from this line

```
Class "Newspack\Reader_Activation\Sync\WooCommerce_Subscriptions" not found
```

The meta was not added neither to the subscription nor to the ESP

includes/plugins/woocommerce-subscriptions/class-subscriptions-meta.php
leogermani: As we discussed, let's leave this for another PR, but I left another comment because it didn't work for me

includes/reader-activation/sync/class-woocommerce.php
chickenn00dle: Ah. Good catch! Fixed in https://github.com/Automattic/newspack-plugin/pull/3568/commits/2a51cf403dab1e46609e7eb615499e62d3d02f86

> The meta was not added neither to the subscription nor to the ESP

Strange that the meta was not being added to the subscription though. I just updated the testing instructions, but make sure you've defined the ff constant to see the changes at work.

Also, I've run into this problem in the past so thought I'd mention it here as well, but if you have a contact with too many meta fields in MC, the newer fields will simply not be added. You can still check the logs for what will be synced with Newspack logging enabled in these cases:

Screenshot 2024-11-26 at 16 52 16

leogermani: > make sure you've defined the ff constant to see the changes at work.

I'd never make such a mistake........ :grimacing:

leogermani: > Also, I've run into this problem in the past so thought I'd mention it here as well, but if you have a contact with too many meta fields in MC, the newer fields will simply not be added. 

Yes. This is a known issue. It's a limitation we have to deal with Mailchimp, that's why we can choose which fields we want. If you scroll the logs up a little you will see that there was an attempt to create the merge field and it failed because it reached the limit.

includes/plugins/woocommerce-subscriptions/class-subscriptions-meta.php
```diff
@@ -0,0 +1,44 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Main class.
+ */
+class Subscriptions_Meta {
+	const CANCELLATION_REASON_META_KEY        = 'newspack_subscriptions_cancellation_reason';
+	const CANCELLATION_REASON_USER_CANCELLED  = 'user-cancelled';
+	const CANCELLATION_REASON_ADMIN_CANCELLED = 'manually-cancelled';
```
leogermani: Every time I see this value I have to ask myself if manually means by the admin or by the user. Why not call this admin-cancelled?

leogermani (changes requested):
@chickenn00dle I'm not getting an infinite loop when I cancel the subscription.

I added some logging and I can see the `maybe_record_cancelled_subscription_meta` method being called over and over again...

the woo_meta was added successfully though

includes/plugins/woocommerce-subscriptions/class-subscriptions-meta.php
chickenn00dle: I used the names @kmwilkerson decided on here. But am happy to change this if there is a strong preference otherwise.

I will say that from the perspective of a non-technical publisher, I think `manual` probably makes more sense than `admin`.

chickenn00dle:
> I added some logging and I can see the maybe_record_cancelled_subscription_meta method being called over and over again...

That's odd. I can't reproduce this on my end when I cancel as admin or as the user. This hook should only trigger when a subscription status is actually updated and once a subscription is cancelled it should no longer be possible to update the status again?

That said, I went ahead and added a preliminary check for cancellation reason meta before updating in the method in https://github.com/Automattic/newspack-plugin/pull/3568/commits/7c582fb851ffe5bb983c0df1cf3a35a74e43a1de. Maybe this can help resolve things on your end? If not, would you mind sharing your steps to reproduce.

leogermani:
> Wait for the end date time to lapse, then visit the site as the reader to ensure the transition from pending cancel to cancel happens

This will be triggered by any request. If I close the reader window and navigate as an admin, it will trigger the cancellation and store `manually-cancelled` as the meta. We need a better strategy to make sure we capture the original intent at the moment the reader clicks the cancel button

chickenn00dle:
> This will be triggered by any request. If I close the reader window and navigate as an admin, it will trigger the cancellation and store manually-cancelled as the meta. We need a better strategy to make sure we capture the original intent at the moment the reader clicks the cancel button

Hmm. The retry system uses Action Scheduler which runs on `WP_Cron`. I thought `is_admin` only returned true if the request is made from an admin screen? If so, scheduled actions wouldn't cause this to resolve as `true`. 

I tested on my end by only reloading in an admin screen until the scheduled action resolved and I am correctly seeing `user-cancelled` meta added:

Screenshot 2024-11-27 at 17 01 06

Are you seeing otherwise?

chickenn00dle:
> This will be triggered by any request. If I close the reader window and navigate as an admin, it will trigger the cancellation and store manually-cancelled as the meta. We need a better strategy to make sure we capture the original intent at the moment the reader clicks the cancel button

Hmm, the scheduled cancellation uses action scheduler which uses WP_Cron. I thought that `is_admin` only returned true if the request comes from an admin screen? If so, wouldn't `is_admin` resolve to `false` for scheduled actions? (I could be misunderstanding this though)

I tested this and when I am on an admin screen at the time a subscription is cancelled by a user reaches its end date, the meta still shows `user-cancelled`:
Screenshot 2024-11-27 at 17 01 06

Are you seeing otherwise?

leogermani:
> Are you seeing otherwise?

Yes. As I user, I cancelled the subscription and then closed the window... did everything else having only the admin tab active

tests/unit-tests/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -0,0 +1,41 @@
...
+		$is_active = WooCommerce_Subscriptions::is_active();
+		$this->assertTrue( $is_active, 'WooCommerce Subscriptions integration should be active when WC_Subscriptions class exists.' );
+	}
+
+	/**
+	 * Test WooCommerce_Subscriptions::is_enabled.
+	 */
+	public function test_is_enabled() {
+		$is_enabled = WooCommerce_Subscriptions::is_enabled();
+		$this->assertTrue( $is_enabled, 'WooCommerce Subscriptions integration should be enabled when Feature Flag is present.' );
```
chickenn00dle, line 39: I wanted to test the case where FF is not defined, as well as when Reader Activation is not enabled, but could not find a way to modify globally defined constants between test runs. I tried `@preserveGlobalState` paired with `@runInSeparateProcess`, but this ended up removing other globals the tested method relied on.

Would love any advice on how to approach this if possible

chickenn00dle:
This is now ready for another round of review @leogermani.

I updated the strategy for determining admin vs. reader cancellations by adding an additional set of meta values for the `pending-cancel` state (`manually-pending-cancel` and `user-pending-cancel`, both ignored by ESP sync logic). These are checked when the subscription reaches the `cancelled` state to determine which cancelled meta value to use.

This means that if a reader "cancels" to `pending-cancel` and then the admin fully cancels prior to the end date, the meta value will still show `user-cancelled`. I think this is alright though since the users decision to cancel is what I think we want to capture here.

tests/unit-tests/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
leogermani: there's no good way to test constants... The best we could do would be to change the approach we use for Feature Flags and use something else to handle these things

leogermani (approved):
I've pushed a commit to avoid that infinite loop again. Everything looks good now :)

PR #3570, opened 2024-11-23
feat: add constant to disable Google OAuth in RAS

laurelfulford:
### All Submissions:

To enable Google as a Sign-In option in RAS, all you need to do is connect Google to the site under Newspack > Connections.

Once that's done, though, you can't disable Google as a sign in option without severing that connection; that connection can be used for other things, like ads. You need to use CSS to hide the button. 

This PR adds a environmental constant, `NEWSPACK_DISABLE_GOOGLE_OAUTH`, that can be used to disable Google OAuth once Google is connected to a site. Down the road, it will probably make sense to have a separate control to disable OAuth specifically in the UI.

See: 1207817176293825-as-1208782453256681.

1. Set up your site to have Google as a Sign In option (there are some steps to do this on a test site here: PamTN9-9RR-p2
2. Run through a sign in and confirm the Google Sign In option appears:

CleanShot 2024-11-22 at 16 26 36

3. In wp-config.php, add `define( 'NEWSPACK_DISABLE_GOOGLE_OAUTH', true );`
4. Run through a sign in again and confirm Google no longer appears:

CleanShot 2024-11-22 at 16 26 15

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -1705,9 +1705,11 @@ final class Reader_Activation {
 
 	/**
 	 * Render third party auth buttons for an authentication form.
+	 *
+	 * If Google is connected to the site, this can be overridden by setting the `NEWSPACK_DISABLE_GOOGLE_OAUTH` environment constant.
 	 */
 	public static function render_third_party_auth() {
-		if ( ! Google_OAuth::is_oauth_configured() ) {
+		if ( ! Google_OAuth::is_oauth_configured() || ( defined( 'NEWSPACK_DISABLE_GOOGLE_OAUTH' ) && NEWSPACK_DISABLE_GOOGLE_OAUTH ) ) {
 			return;
 		}
 		?>


dkoo (approved):

dkoo (commented):
One nit: right now we only support Google as an SSO service, but in the future we might add more. Should the constant be more generalized to reflect this? e.g. `NEWSPACK_DISABLE_READER_SSO` or something similar?

laurelfulford:
> right now we only support Google as an SSO service, but in the future we might add more. Should the constant be more generalized to reflect this? e.g. `NEWSPACK_DISABLE_READER_SSO` or something similar?

Good question! 🤔 In this specific case, the issue is that there's no UI to turn Google off once connected, and you may not connect it for OAuth, you may connect it for something unrelated (like ads). So I can see a case where you may want to just turn off Google as an SSO, even if another's available. 

(Ideally when we add another, we'll also incorporate a UI to turn off each instead of doing this -- this was kind of a quick fix😅). 

Let me know if that makes sense, @dkoo! I'm going to hold off merging so we don't end up having to chase down flags to change them if we change our minds!

dkoo:
That sounds good, thanks for clarifying!

laurelfulford:
Thanks for the quick reply, @dkoo!

PR #3590, opened 2024-12-02
fix(emails): account for false order value in email hook

chickenn00dle:
### All Submissions:

Looks like `$order` can be false in some cases:

```
An error of type E_ERROR was caused in line 271 of the file /srv/htdocs/wp-content/plugins/newspack-plugin/includes/reader-revenue/woocommerce/class-woocommerce-connection.php. Error message: Uncaught Error: Call to a member function get_id() on null in /srv/htdocs/wp-content/plugins/newspack-plugin/includes/reader-revenue/woocommerce/class-woocommerce-connection.php:271
```

This PR fixes this by checking the value of `$order` BEFORE calling `get_id`

1. Go through the checkout flow for a donation product and for a non-donation product and monitor logs.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -267,14 +267,14 @@ class WooCommerce_Connection {
 	 * @return bool
 	 */
 	public static function send_customizable_receipt_email( $enable, $order, $class ) {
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
LGTM! 👍 I ran a couple transactions (donation and regular), and didn't see anything off in the logs!

PR #3593, opened 2024-12-03
feat(wcs): add expired subscription cli tool

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208702687864922/1208702687865965/f

This PR adds a cli command to migrate the status of all on-hold subscriptions that have failed all retries to expired:

```
NAME

  wp newspack migrate-expired-subscriptions

DESCRIPTION

  Migrate status of on-hold WooCommerce subscriptions that have failed all payment retries to expired.

SYNOPSIS

  wp newspack migrate-expired-subscriptions [--live] [--verbose] [--ids]

OPTIONS

  [--live]
    Run the command in live mode, updating the subscriptions.

  [--verbose]
    Produce more output.

  [--ids]
    Comma-separated list of subscription IDs. If provided, only subscriptions with these IDs will be processed.
```

1. Make sure you DO NOT have the `NEWSPACK_SUBSCRIPTIONS_EXPIRATION` constant defined
2. If you don't already have any active subscriptions, as a reader purchase a handful of subscriptions
3. Follow the steps here to trigger scheduled retry for a handful of subscriptions: https://woocommerce.com/document/subscriptions/develop/failed-payment-retry/#section-7
4. Ensure you have at least one subscription of each:
  - An on-hold subscription with all payment retries completed
  - An on-hold subscription that is awaiting another payment retry
  - An on-hold subscription that has no retries scheduled (manually move this to on-hold)
  - A non-on-hold subscription (active, expired, etc.)
5. Via terminal, run the `wp newspack migrate-expired-subscriptions --verbose` command. Confirm an error appears letting you know subscription integration is not enabled.
6. Define the `NEWSPACK_SUBSCRIPTIONS_EXPIRATION` constant: `define( 'NEWSPACK_SUBSCRIPTIONS_EXPIRATION', true );`
7. Via terminal, run the command again. Confirm the output shows you are in dry run mode, 1 subscription has been updated, and the other subscriptions are skipped with the expected reasons. Also confirm the non-on-hold subscription is not included.
8. Now include some ids to confirm the command will work on only the provided ids: `wp newspack migrate-expired-subscriptions --verbose --ids=1,2,3,4,5`
9. Finally, run the command with the live flag: `wp newspack migrate-expired-subscriptions --live`. Confirm the status of the on-hold subscription has updated to expired.
10. Smoke test the command with some more subscriptions!

### Other information:

### DIFF
--- a/includes/cli/class-woocommerce-subscriptions.php
+++ b/includes/cli/class-woocommerce-subscriptions.php
@@ -1,0 +1,122 @@
+<?php
+/**
+ * WooCommerce Subscriptions Integration CLI commands.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\CLI;
+
+use WP_CLI;
+use Newspack\Woocommerce_Subscriptions as WooCommerce_Subscriptions_Integration;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WooCommerce Subscriptions Integration CLI commands.
+ */
+class WooCommerce_Subscriptions {
+	/**
+	 * Flag for live mode.
+	 *
+	 * @var bool
+	 */
+	private static $live = false;
+
+	/**
+	 * Flag for verbose output.
+	 *
+	 * @var bool
+	 */
+	private static $verbose = false;
+
+	/**
+	 * Migrate on-hold WooCommerce subscriptions with failed renewal orders to expired status.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--live]
+	 * : Run the command in live mode, updating the subscriptions.
+	 *
+	 * [--verbose]
+	 * : Produce more output.
+	 *
+	 * @param array $args Positional arguments.
+	 * @param array $assoc_args Assoc arguments.
+	 * @return void
+	 */
+	public function migrate_expired_subscriptions( $args, $assoc_args ) {
+		WP_CLI::line( '' );
+		if ( ! WooCommerce_Subscriptions_Integration::is_enabled() ) {
+			WP_CLI::error( 'WooCommerce Subscriptions Integration is not enabled.' );
+			WP_CLI::line( '' );
+			return;
+		}
+		self::$live    = isset( $assoc_args['live'] ) ? true : false;
+		self::$verbose = isset( $assoc_args['verbose'] ) ? true : false;
+		if ( self::$live ) {
+			WP_CLI::line( 'Live mode - subscription statuses will be updated.' );
+		} else {
+			WP_CLI::line( 'Dry run. Use --live flag to run in live mode.' );
+		}
+		$updated       = 0;
+		$page          = 1;
+		$per_page      = 25;
+		$subscriptions = wcs_get_subscriptions(
+			[
+				'paged'                  => $page,
+				'subscriptions_per_page' => $per_page,
+				'subscrition_status'     => 'on-hold',
+			]
+		);
+		if ( empty( $subscriptions ) ) {
+			WP_CLI::success( 'No on-hold subscriptions to process.' );
+			WP_CLI::line( '' );
+			return;
+		}
+		WP_CLI::line( 'Processing subscriptions...' );
+		WP_CLI::line( '' );
+		while ( ! empty( $subscriptions ) ) {
+			foreach ( $subscriptions as $subscription ) {
+				$id = $subscription->get_id();
+				if ( self::$verbose ) {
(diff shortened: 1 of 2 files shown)


includes/cli/class-woocommerce-subscriptions.php
```diff
@@ -0,0 +1,182 @@
...
+						WP_CLI::line( 'Last retry does not have a failed status. Moving to next subscription...' );
+						WP_CLI::line( '' );
+					}
+					continue;
+				} else {
+					if ( self::$verbose ) {
+						WP_CLI::line( 'Updating subscription status to expired...' );
+					}
+					if ( self::$live ) {
+						$subscription->update_status( 'expired', __( 'Subscription status updated by Newspack CLI command.', 'newspack-plugin' ) );
```
leogermani, line 133: Let's also add one of those Subscriptions Notes that show up in the subscription page?

claudiulodro: In discussion with Tracy and Katie, it would be best if the subscription had the expiration date of the last failed payment rather than the expiration date of when the script was run.

claudiulodro (dismissed):

includes/cli/class-woocommerce-subscriptions.php
chickenn00dle: > Let's also add one of those Subscriptions Notes that show up in the subscription page?

The second argument to `$subscription->update_status` is added as a subscription note. Is this what you were referring to?

 Screenshot 2024-12-06 at 11 44 11

> In discussion with Tracy and Katie, it would be best if the subscription had the expiration date of the last failed payment rather than the expiration date of when the script was run.

Sure thing @claudiulodro! I updated to script to use the date of the last retry object. This should be the same as the last failed payment: https://github.com/Automattic/newspack-plugin/pull/3593/commits/b29bdd17b357d01af080d0d7e237201734f9d1fc

When testing though, since we are manually triggering the scheduled actions, this end date will be set as though the scheduled action was not triggered manually, using the expected end date for the retry. I think this should be fine since our publishers shouldn't be doing this, however let me know if we need to be more specific here.

leogermani (changes requested):
The detection and the update seems to be working fine.

But the script runs in an infinite loop because `get_subscriptions` always return the same subs.

I dont see a `paged` param in wcs_get_subscriptions -> https://github.com/riclain/woocommerce-subscriptions/blob/b0dd2efd869b0639ad54131fba9c13b4ed5a09a2/wcs-functions.php#L332

Also, the End date didn't work for me. It was set to the current date. I did have some old expired subs in my DB so tit's easy to spot it

chickenn00dle:
> I dont see a paged param in wcs_get_subscriptions -> https://github.com/riclain/woocommerce-subscriptions/blob/b0dd2efd869b0639ad54131fba9c13b4ed5a09a2/wcs-functions.php#L332

@leogermani, `paged` is here in the official WCS core repo: https://github.com/Automattic/woocommerce-subscriptions-core/blob/1692c41cf43e2c3c5ef0816a48196ffb241b9e72/wcs-functions.php#L402

Looks like I forgot to add a call to `$subscription->save()`, although its strange that on my end I don't see the infinite loop without this. Does this fix that issue for you?

Looking into the end date issue now.

leogermani:
> Looks like I forgot to add a call to $subscription->save(), although its strange that on my end I don't see the infinite loop without this. Does this fix that issue for you?

I don't think it will, because it happens on dry mode as well.

In fact, I tried on a shell to call `wcs_get_subscriptions` passing `paged` as big number, just to see if I got an empty array as a result.. and I didn't

----------------

Looks like I was runing a pretty old version of woo subscriptions :) upgrading it fixed the infinite loop. I'try to test the data again in a bit

leogermani:
Tested again and the "End date" is still set to the date I ran the script.

chickenn00dle:
> Tested again and the "End date" is still set to the date I ran the script.

Oof. This was because the call to `$subscription->set_end_date` needs to happen after `$subscription->update_status` since the status update will also set an end date. 

When I initially made the changes and tested locally, this was the case, but before pushing the commit, I thought I'd "tidy up the changes" before pushing and reversed the order. 

Sorry about that! Should be fixed now.

leogermani (approved):
> When I initially made the changes and tested locally, this was the case, but before pushing the commit, I thought I'd "tidy up the changes" before pushing and reversed the order.

been there. done that! :) 

All good now!

PR #3603, opened 2024-12-03
fix(esp-sync): schedule second sync upon subscription reactivation, just in case

dkoo:
### All Submissions:

We fire an ESP contact sync whenever a Woo subscription status is updated. When a subscription undergoes automatic renewal, it rapidly changes status to `on-hold` when payment processing begins, and then to `active` once successful payment is confirmed. However, since this can happen so quickly it's possible that sync after the `active` update can happen before the `on-hold` sync completes (a sort of race condition).

Rather than trying to untangle the Woo Subscriptions logic here, the easiest thing to do is to schedule a second "backup" sync a few minutes later when a subscription status changes to `active`. This should ensure that the ESP contact is updated to `active` after a successful renewal even if the race condition occurs.

1. Check out this branch.
2. Manually renew a subscription in WP admin.
3. Confirm that the ESP contact is synced when the subscription is changed to `on-hold` status and again when the status changes to `active` (but note that it might not always happen in this order).
4. After the `active` sync completes, run `wp cron event list` and confirm that a non-repeating `newspack_scheduled_esp_sync` job is scheduled roughly two minutes in the future.
5. Wait ~2 minutes or run `wp cron event run newspack_scheduled_esp_sync` and confirm that the contact is synced to the ESP again with the correct "active" status.
6. (BONUS) Renew several subscriptions in rapid succession (within a single 2-minute window) and confirm that all subscriptions get a second sync ~2 minutes later, with the correct contact data.

### Other information:

---
  - https://app.asana.com/0/0/1208887537927842


includes/data-events/connectors/class-esp-connector.php
```diff
@@ -140,7 +140,18 @@ public static function subscription_updated( $timestamp, $data, $client_id ) {
...
 		}
 
-		self::sync( $contact, sprintf( 'RAS Woo Subscription updated. Status changed from %s to %s', $data['status_before'], $data['status_after'] ) );
+		// Schedule a backup sync in two minutes if the subscription is being reactivated, to get around potential race conditions.
+		$delay = 'active' === $data['status_after'] ? 120 : 0;
+
+		self::sync(
+			$contact,
+			sprintf(
+				'RAS Woo Subscription updated. Status changed from %s to %s',
```
leogermani, line 147: Let's add something to the message to identify these requests

leogermani: I was not able to test this yet. But a few comments/questions

Here it seems you are replacing the sync with a scheduled sync. So it's not a backup sync. I would leave the sync as is and schedule one additional backup sync.

Also, I'm not sure about triggering this only when a subscription is set to active. A renewal could end up with a subscription in another state, and the race condition can also happen in that case.

I think we need to identify situations when two syncs are triggered too close to one another, and then schedule a third one just to be sure. (or if there's an easy and safe way, delay the second one as you did here. But the second one is not necessarily settingg the status to active)

dkoo: > Here it seems you are replacing the sync with a scheduled sync. So it's not a backup sync. I would leave the sync as is and schedule one additional backup sync.

It doesn't just delay the sync, it actually schedules another sync after completing the regular sync (`upsert` [call here](https://github.com/Automattic/newspack-plugin/pull/3603/files#diff-54a027d8960b76963f30ce437f31ebe5e011ea94e6860c2b60423378f5330731R114)).

> Also, I'm not sure about triggering this only when a subscription is set to active. A renewal could end up with a subscription in another state, and the race condition can also happen in that case.

We can certainly add a backup sync in other instances, too—we would just need to add that third param to the `sync` call to trigger the schedule after the given number of seconds.

dkoo: > Let's add something to the message to identify these requests

I did [add a log](https://github.com/Automattic/newspack-plugin/pull/3603/files#diff-54a027d8960b76963f30ce437f31ebe5e011ea94e6860c2b60423378f5330731R118) whenever a backup sync is scheduled. Maybe these logs should be done through the `newspack_log` action instead?

leogermani: > We can certainly add a backup sync in other instances, too—we would just need to add that third param to the sync call to trigger the schedule after the given number of seconds.

wonder if we can identify when a renewal happens. And then schedule a few minutes after the renew is done

dkoo: 3e28d39 adds an optional `$data` param to the Sync class's logger method. If it receives any data, it will log it to `newspack_log`.

dkoo: > wonder if we can identify when a renewal happens. And then schedule a few minutes after the renew is done

Sure, this is probably a more flexible solution. 3e8bce46 adds a new data event when a subscription renews, and a new handler that schedules a sync two minutes after that event fires. It also fetches the contact data at the moment of the scheduled sync (instead of when it's scheduled), which should help more to avoid the race condition.

leogermani (approved):
Beautiful. It works.

I just have one  comment/question/suggestion... Maybe we could use a different name for the Data Even, instead of `subscription_renewed`... 

This name gives the impression that it will be triggered after a successful renewal, and might lead to confusion when people hook into it in the future. Maybe "renewal_attempt"?

dkoo:
@leogermani good suggestion! 660fe4d6 updates the data event name (and the handlers/methods that use it) to clarify that distinction. (EDIT: and 0d6b92a0 removes some unintended changes that got swept up in this commit.)

includes/data-events/connectors/class-esp-connector.php
```diff
@@ -140,7 +141,38 @@ public static function subscription_updated( $timestamp, $data, $client_id ) {
...
+	 * @param int   $timestamp Timestamp of the event.
+	 * @param array $data      Data associated with the event.
+	 * @param int   $client_id ID of the client that triggered the event.
+	 */
+	public static function subscription_renewal_attempt( $timestamp, $data, $client_id ) {
+		if ( empty( $data['subscription_id'] ) || empty( $data['user_id'] ) || empty( $data['email'] ) ) {
+			return;
+		}
+
+		self::schedule_sync(
```
leogermani, line 171:
```suggestion
		/** 
		* When a renewal happens, it triggers two syncs to the ESP, one setting the subscription as on hold, and a 
		* second one setting it back to active. This sometimes creates a race condition on the ESP side. 
		* This third request will make sure the ESP always has the correct and most up to date data about the reader.
		*/
		self::schedule_sync(
```

PR #3604, opened 2024-12-06
fix: duplicate orders save on cron

leogermani:
### All Submissions:

There were a fee oversights in my review in https://github.com/Automattic/newspack-plugin/pull/3555

First. The cron job will not save the duplicates, so the notices will never appear. This fixes it.

Also, the CLI command name says it is an upsert, but it was passing a parameter that would override the whole option. Fixed it as well

Perform the tests on https://github.com/Automattic/newspack-plugin/pull/3555

### Other information:


dkoo (commented):
Maybe I'm doing something wrong, but following the instructions in #3555, after changing the order dates of the second pair of orders to more than a month ago, I still see them in the notification shown at step 4:

Screenshot 2024-12-10 at 10 32 29 AM

dkoo (approved):
It's possible the above was due to user error (e.g. I created the orders, then switched to the branch, then changed the dates, which would cause the first cron job to pick them up when switching to the branch). After testing with more orders, this is working as expected.

PR #3605, opened 2024-12-07
fix(ras-acc): re-add recaptcha to the WooCommerce checkout

laurelfulford:
### All Submissions:

The code that adds the g-recaptcha field to the WooCommerce cart checkout was removed as part of [this PR](https://github.com/Automattic/newspack-plugin/pull/3283).

Unfortunately this causes the regular cart not to be able to process orders when reCAPTCHA v3 is enabled. V2 still seems to work, but it's possibly because reCAPTCHA doesn't work at all. I've reinstated both.

1. Set up your site to use reCAPTCHA v3. 
2. Run through a regular modal checkout of a subscription product.
3. Go to My Account > Subscriptions, and edit the monthly amount for your subscription.
4. On the final step of trying to checkout, note the reCAPTCHA error.
5. Apply this PR. 
6. Repeat steps 3 and 4 and confirm that the reCAPTCHA error doesn't return.
7. Try a modal checkout purchase and confirm there are no errors.
8. Try a regular /shop to /cart purchase and confirm there are no errors.
9. Switch to reCAPTCHA v2, and repeat all the tests. 

### Other information:

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -25,6 +25,10 @@ final class Recaptcha {
 		\add_action( 'rest_api_init', [ __CLASS__, 'register_api_endpoints' ] );
 		\add_action( 'wp_enqueue_scripts', [ __CLASS__, 'register_scripts' ] );
 
+		// Add reCAPTCHA to the Woo checkout form.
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
Looks good. Good catch @laurelfulford!

laurelfulford:
Thanks @chickenn00dle! Updating the alphas...

PR #3631, opened 2024-12-13
fix(recaptcha): various optimizations

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208950357031079/f

This PR makes some optimizations to our recaptcha implementation:

1. Only render recaptcha once a form is focused
2. Auto retry on recaptcha error/expire
3. Specify recaptcha container so we don't break form layouts

Screenshot 2024-12-16 at 14 26 17

1. Enable recaptcha v2
2. As reader open any page/post that contains a newsletters subscribe block but do not focus the form yet.
4. In browser dev tools, search for an element with a `data-recaptcha-widget-id` attribute. There should be none
5. Now click anywhere in the form
6. In dev tools again, search for an element with a `data-recaptcha-widget-id` attribute. The button in the focused form should have this.
7. Now smoke test all of the following, each time making sure the layout of the forms don't change as you focus:
    - [x] Checkout Button or Donate block
    - [x] Newsletters subscribe block
    - [x] Registration block
    - [x] RAS-ACC auth registration modal
8. Enable recaptcha v3
9. Repeat steps 2-6, this time searching for a hidden input with a name of `g-recaptcha-response`.
10. Now smoke test all of the following:
    - [x] Checkout Button or Donate block
    - [x] Newsletters subscribe block
    - [x] Registration block
    - [x] RAS-ACC auth registration modal
11. Disable recaptcha and smoke test the following
    - [x] Checkout Button or Donate block
    - [x] Newsletters subscribe block
    - [x] Registration block
    - [x] RAS-ACC auth registration modal

**Testing recaptcha errors:**
To test auto-retry on recaptcha error, you can force the error callback by replacing this line with `callback: errorCallback`: https://github.com/Automattic/newspack-plugin/blob/ba6f10f15127d5c016d7557da6684270749f385c/src/other-scripts/recaptcha/index.js#L177

1. Re-enable recaptcha v2
2. As reader, go through any relevant form
3. After recaptcha is triggered, instead of succeeding the error callback should be triggered, re-prompting the widget (in normal circumstances a render failure would not trigger the widget to appear)
4. This should happen 3 times before an error is rendered asking for a page reload.

### Other information:

### DIFF
--- a/src/other-scripts/recaptcha/index.js
+++ b/src/other-scripts/recaptcha/index.js
@@ -96,6 +96,39 @@ function addHiddenField( form ) {
 }
 
 /**
+ * Append a generic error message above the given form.
+ *
+ * @param {HTMLElement} form    The form element.
+ * @param {string}      message The error message to display.
+ */
+function addErrorMessage( form, message ) {
+	const errorText = document.createElement( 'p' );
+	errorText.textContent = message;
+	const container = document.createElement( 'div' );
+	container.classList.add( 'newspack-recaptcha-error' );
+	container.appendChild( errorText );
+	// Newsletters block errors render below the form.
+	if ( form.parentElement.classList.contains( 'newspack-newsletters-subscribe' ) ) {
+		form.append( container );
+	} else {
+		container.classList.add( 'newspack-ui__notice', 'newspack-ui__notice--error' );
+		form.insertBefore( container, form.firstChild );
+	}
+}
+
+/**
+ * Remove generic error messages from form if present.
+ *
+ * @param {HTMLElement} form The form element.
+ */
+function removeErrorMessages( form ) {
+	const errors = form.querySelectorAll( '.newspack-recaptcha-error' );
+	for ( const error of errors ) {
+		error.parentElement.removeChild( error );
+	}
+}
+
+/**
  * Remove the hidden reCAPTCHA v3 token field from the given form.
  *
  * @param {HTMLElement} form The form element.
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
@chickenn00dle all is working well, and the rendering of the widgets/hidden fields during field focus is seamless and causes no detectable form layout changes or delays in form submission. 🎉 

The only thing I see is that when displaying the v2 widget on the modal checkout second screen, the modal is no longer scrolling up to bring it into the viewport. As a reader, if I don't know to expect a reCAPTCHA v2 widget, it's hard to tell what's happening as the form inside modal simply becomes greyed-out and non-interactive. Could we ensure that if the widget is displayed, we're scrolling the modal viewport so that it's visible?

dkoo:
Also, I pushed 80cb8506 because I saw an issue in my local environment while testing this in which the site key/secret fields in the admin dashboard were appearing empty even though I knew the data existed in my site DB.

chickenn00dle:
> The only thing I see is that when displaying the v2 widget on the modal checkout second screen, the modal is no longer scrolling up to bring it into the viewport. As a reader, if I don't know to expect a reCAPTCHA v2 widget, it's hard to tell what's happening as the form inside modal simply becomes greyed-out and non-interactive. Could we ensure that if the widget is displayed, we're scrolling the modal viewport so that it's visible?

That's odd. On my end it looks to be scrolling:

https://github.com/user-attachments/assets/1b6c4cff-e0af-4b96-b15f-7c353d402d29

dkoo:
fe03f44 restores a couple of lines from #3283 which trigger a re-render of v2 widgets when Woo checkout fires `updated_checkout` or `checkout_error` jQuery events. These events indicate that entire form elements (including our widgets and the submit buttons they're attached to) may have been replaced by new DOM elements, so re-rendering ensures that the reCAPTCHA check will continue to happen afterward.

dkoo (approved):
This is working now on my end, and because it's in response to some time-sensitive publisher requests, I'm going to go ahead and merge without a second PR reviewer. 🤞

PR #3638, opened 2024-12-18
feat(corrections): add corrections and clarifications behind feature flag

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208894036683495/1208945901567236 and https://app.asana.com/0/1208894036683495/1208945901567238

This PR adds the corrections feature behind the `NEWSPACK_CORRECTIONS_ENABLED` feature flag. Note this is a quick and simple implementation based on an already existing custom plugin strictly for the purpose of getting ready for a site migration. 

**Editor:**
Screenshot 2025-01-08 at 16 35 44

**Frontend:**
Screenshot 2025-01-08 at 16 34 54

1. Enable the corrections feature flag in wp-config (`NEWSPACK_CORRECTIONS_ENABLED`)
2. Edit any post as admin
3. Scroll to the bottom of the content editor and verify the corrections metabox is present
4. Select the activate corrections checkbox, select a location, and add some corrections
5. As a reader visit the post on the front end
6. Verify the corrections are added to the post
7. As admin again, select a different corrections location
8. As reader again, visit the post and verify the corrections have moved
9. As admin again, uncheck the activate corrections checkbox and save
10. As reader again, visit the post and verify the corrections are NOT added to the post
12. As admin, delete a revision and save
13. As reader, verify the revisions appear as expected on the frontend
14. As admin, remove the feature flag from wp-config and edit any post
15. Verify corrections are no longer present

### Other information:

### DIFF
--- a/includes/corrections/class-corrections.php
+++ b/includes/corrections/class-corrections.php
@@ -1,0 +1,314 @@
+<?php
+/**
+ * Newspack Corrections and Clarifications
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+/**
+ * Class to handle Corrections and Clarifications.
+ */
+class Corrections {
+	/**
+	 * Meta key for storing corrections.
+	 */
+	const CORRECTION_META = 'article-corrections';
+
+	/**
+	 * Meta key for storing whether corrections are active.
+	 */
+	const CORRECTION_ACTIVE_META = 'has_corrections';
+
+	/**
+	 * Initializes the class.
+	 */
+	public static function init() {
+		if ( ! self::is_enabled() ) {
+			return;
+		}
+		add_action( 'add_meta_boxes', [ __CLASS__, 'add_corrections_metabox' ] );
+		add_action( 'save_post', [ __CLASS__, 'save_corrections_metabox' ] );
+		add_filter( 'the_content', [ __CLASS__, 'output_corrections_on_post' ] );
+		add_action( 'init', [ __CLASS__, 'add_corrections_shortcode' ] );
+	}
+
+	/**
+	 * Checks if the feature is enabled.
+	 *
+	 * True when:
+	 * - NEWSPACK_CORRECTIONS_ENABLED is defined and true.
+	 *
+	 * @return bool True if the feature is enabled, false otherwise.
+	 */
+	public static function is_enabled() {
+		return defined( 'NEWSPACK_CORRECTIONS_ENABLED' ) && NEWSPACK_CORRECTIONS_ENABLED;
+	}
+
+	/**
+	 * Adds the corrections shortcode.
+	 */
+	public static function add_corrections_shortcode() {
+		add_shortcode( 'corrections', [ __CLASS__, 'handle_corrections_shortcode' ] );
+	}
+
+	/**
+	 * Handles the corrections shortcode.
+	 *
+	 * @return string The shortcode output.
+	 */
+	public static function handle_corrections_shortcode() {
+		global $wpdb;
+
+		$post_ids = get_posts(
+			[
+				'posts_per_page' => -1,
+				'meta_key'       => self::CORRECTION_ACTIVE_META, // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key
+				'meta_value'     => 1, // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+				'fields'         => 'ids',
+				'orderby'        => 'date',
+				'order'          => 'DESC',
+			]
+		);
+
+		ob_start();
+		foreach ( $post_ids as $post_id ) :
+			$corrections = get_post_meta( $post_id, self::CORRECTION_META, true );
+			if ( empty( $corrections ) ) {
+				continue;
+			}
+
+			?>
+			<!-- wp:group {"className":"is-style-default correction-shortcode-item"} -->
+			<div class="wp-block-group is-style-default correction-shortcode-item">
+				<div class="wp-block-group__inner-container">
+					<!-- wp:newspack-blocks/homepage-articles {"showExcerpt":false,"showDate":false,"showAuthor":false,"mediaPosition":"left","specificPosts":["<?php echo intval( $post_id ); ?>"],"imageScale":2,"specificMode":true} /-->
+
+					<div class="correction-list">
+						<?php foreach ( $corrections as $correction ) : ?>
+							<?php $correction_heading = ! empty( $correction['date'] ) ? 'Correction on ' . gmdate( 'M j, Y', strtotime( $correction['date'] ) ) : 'Correction'; ?>
+							<p>
+								<span class="correction-date"><?php echo esc_html( $correction_heading ); ?><span>:</span></span>
+								<?php echo esc_html( $correction['correction'] ); ?>
+							</p>
+						<?php endforeach; ?>
+					</div>
+
+			</div></div>
+			<!-- /wp:group -->
+			<?php
+		endforeach;
+		return do_blocks( ob_get_clean() );
+	}
+
+	/**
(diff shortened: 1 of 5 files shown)


chickenn00dle:
TODO:

- [x] Update corrections
- [x] Refresh on corrections submit

src/other-scripts/corrections/index.js
```diff
@@ -0,0 +1,91 @@
...
+			}
+
+			if ( ! isSavingPost && hasSavedPost ) {
+				// Unsubscribe from the store.
+				unsubscribe();
+				window.location.href = window.location.href;
+			}
+		} );
+	}
+} );
```
chickenn00dle, line 90: This is a super hacky way to force reload the page. We need to reload the page when adding new corrections so the metabox fields can repopulate with the respective correction ids (needed for updates and deleting). Ideally we remove this when we are able to fully work on the feature

chickenn00dle:
Gonna mark this one ready for review since the core functionality is working and this should at least get us to a point where we can begin migrating corrections.

After working on this, my thought is we should re-implement corrections as a new block. I did start working on this a bit, but realized this would imply some design decisions, so for the time being I've gone ahead and just pushed up a rework of our existing corrections custom plugin.

cc @leogermani

includes/corrections/class-corrections.php
```diff
@@ -0,0 +1,503 @@
...
+		update_post_meta( $correction_id, self::CORRECTION_DATE_META, sanitize_text_field( $correction['date'] ) );
+	}
+
+	/**
+	 * Delete corrections for post.
+	 *
+	 * @param array $correction_ids correction ids.
+	 */
+	public static function delete_corrections( $correction_ids ) {
+		$stored_correction_ids = get_post_meta( $post_id, self::CORRECTIONS_IDS_META, true );
```
leogermani, line 227: `$post_id` doesn't seem to be defined here

includes/corrections/class-corrections.php
```diff
@@ -0,0 +1,503 @@
...
+		];
+		$args = array(
+			'labels'           => $labels,
+			'description'      => 'Post type used to store corrections and clarifications.',
+			'has_archive'      => true,
+			'public'           => true,
+			'public_queryable' => true,
+			'query_var'        => true,
+			'rewrite'          => [ 'slug' => 'correction' ],
+			'show_ui'          => true,
```
leogermani: I don't think we want the UI for that

includes/corrections/class-corrections.php
```diff
@@ -0,0 +1,503 @@
...
+		foreach ( $corrections as $correction ) {
+			$id = wp_insert_post(
+				[
+					'post_title'   => 'Correction for ' . get_the_title( $post_id ),
+					'post_content' => $correction['content'],
+					'post_type'    => self::POST_TYPE,
+					'post_status'  => 'publish',
+					'meta_input'   => [
+						self::CORRECTION_POST_ID_META => $post_id,
+						self::CORRECTION_DATE_META    => $correction['date'],
```
leogermani: why not store the date using post_date?

includes/corrections/class-corrections.php
```diff
@@ -0,0 +1,503 @@
...
+		);
+	}
+
+	/**
+	 * Save corrections for post.
+	 *
+	 * @param int   $post_id     The post ID.
+	 * @param array $corrections The corrections.
+	 */
+	public static function save_corrections( $post_id, $corrections ) {
```
leogermani: nit: naming it `add_corrections` (or insert) would make the purpose of this method clearer

leogermani (changes requested):
It works and serves the purpose of unblocking migration.

There is just one warning being thrown at `delete_corrections` method.

I left a few other non blocking comments

Another non-blocking comment: I think "activate corrections" and "position" should be a global setting, and not something per post... We can discuss it and I don't think this will be a blocker for migration now.... well... maybe we can get rid of the "activate" so it doesn't get in the way.

Finally, WDYT of adding a short README to the folder describing how the data is stored, to inform LEs that will work on the migration?

includes/corrections/class-corrections.php
chickenn00dle: Oops! Good catch. Fixed in https://github.com/Automattic/newspack-plugin/pull/3638/commits/fea98909ba23a62584c9f97bd3a2adc97bf1be2e

includes/corrections/class-corrections.php
chickenn00dle: Also fixed in https://github.com/Automattic/newspack-plugin/pull/3638/commits/fea98909ba23a62584c9f97bd3a2adc97bf1be2e

includes/corrections/class-corrections.php
chickenn00dle: I thought about this, but also considered corrections will be viewed in other contexts outside of a post (shortcode, archive, etc.) and so it might be useful to distinguish between the actual date a correction is published and the date assigned to a correction by a publisher as these might not always be the same.

includes/corrections/class-corrections.php
chickenn00dle: Good suggestion! I am not very good at naming things. Fixed in https://github.com/Automattic/newspack-plugin/pull/3638/commits/fea98909ba23a62584c9f97bd3a2adc97bf1be2e

chickenn00dle:

> Another non-blocking comment: I think "activate corrections" and "position" should be a global setting, and not something per post... We can discuss it and I don't think this will be a blocker for migration now.... well... maybe we can get rid of the "activate" so it doesn't get in the way.

I think position makes sense as a global setting, but do think it would be useful for publishers to be able to set whether corrections are active at the post level. For example, being able to publish a correction but not set it to active until it has been approved/verified. That said, I am open to any decision here.

> Finally, WDYT of adding a short README to the folder describing how the data is stored, to inform LEs that will work on the migration?

Great idea! Added one here: https://github.com/Automattic/newspack-plugin/blob/fea98909ba23a62584c9f97bd3a2adc97bf1be2e/includes/corrections/README.md

Let me know if I need to expand anywhere!

includes/corrections/class-corrections.php
leogermani: Right, I actually thought that the date the correction was posted is not relevant, so we could use this field as the correction date.. no need to have two.

But I get your point. The only thing is that if we go this route, `get_corrections` will have to order posts by a postmeta value, that will be a string. If we used the post_date this would be working out of the box

leogermani:
> For example, being able to publish a correction but not set it to active until it has been approved/verified. That said, I am open to any decision here.

Makes sense. We don't need to decide this now, it has little impact on the migration. Can I just suggest that this should be active by default?

I'm thinking about LEs migrating data and wanting to look at the site to see if the data is there. Ideally they wouldn't need to add a "active = true" for every post and we'll think about it later

includes/corrections/class-corrections.php
chickenn00dle: That's a fair point. I've updated this to use `post_date` instead in https://github.com/Automattic/newspack-plugin/pull/3638/commits/13dbe298c90b8fd99b6766138d43d8a3cb91179d

chickenn00dle:
> Makes sense. We don't need to decide this now, it has little impact on the migration. Can I just suggest that this should be active by default?

> I'm thinking about LEs migrating data and wanting to look at the site to see if the data is there. Ideally they wouldn't need to add a "active = true" for every post and we'll think about it later

That's fair. I've updated this in https://github.com/Automattic/newspack-plugin/pull/3638/commits/0017f165072eeff547e70d7efd5fc08893749b88 so when the meta is not present corrections will be shown.

leogermani (approved):
Thanks for the revisions!

chickenn00dle:
Removing needs-changelog label since this feature is strictly to prep for a migration and won't actually launch quite yet

PR #3639, opened 2024-12-18
fix: add max limit of 5 consecutive scheduled syncs

dkoo:
### All Submissions:

Adds a mechanism to the `schedule_sync` method so that consecutive scheduled syncs for a single contact will only happen up to a max of five consecutive times. This is to avoid an infinite loop if a contact is scheduled to retry the sync after a failure, but the sync keeps failing.

The counter is reset upon a successful sync to the ESP.

Follow instructions in https://github.com/Automattic/newspack-newsletters/pull/1729.

### Other information:


dkoo:
Since we're coming right up against the holidays at this point, we're going to hold this change for when the team is back at full strength.

PR #2934, opened 2024-02-15
fix(my-account): always update all subs when updating payment method

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

When updating a Stripe Credit Card, the new card should be used for all active subscriptions. This ensure that when the donor updates their payment method, it will be used for renewing the subscriptions.

| Before | After | 
|--------------|-----------|
| image | image |

See 1200550061930446-as-1206327869657181/f

### How to test the changes in this Pull Request:

1. Make a recurring donation using the `4242 4242 4242 4242` credit card
2. Go to /my-account, "Payment Methods" section and update the payment method to `4000056655665556` CC
3. Save the payment method and go to "My Subscription" section -> observe the payment method used is the new one

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3643, opened 2024-12-20
fix(stripe): when adding new card, update subs of all statuses

dkoo:
### All Submissions:

By default, when a reader adds a new payment method to the Stripe gateway, the checkbox to "Update the payment method for all of my current subscriptions" is only shown if the reader has an _active_ subscription. If the reader only has non-active subscriptions (e.g. a subscription in `on-hold` status due to a failed payment), then those subscriptions' payment methods won't be updated automatically when a new payment method is added.

Since we visually hid this checkbox in #2934 so that it's always active when present, it was hard to tell that it wasn't always there.

1. On `release`, as a reader, purchase a subscription.
2. As an admin, set the subscription to "on hold" status (or select the "create pending renewal order", which simulates a subscription after a failed payment).

Screenshot 2024-12-19 at 5 23 04 PM

3. As the reader, add a new different payment method in My Account.
4. As the admin, manually renew the subscription. Observe that the renewal order created charges the old card used in step 1, not the one added in step 3. Also observe in the reader's My Account > My Subscription page that the subscription is still set to use the old payment method.
5. Check out this branch.
6. Repeat 1–4 on this branch, but after renewing the subscription, confirm that it charges the new payment method and that in the reader's My Account > My Subscription page, the subscription is set to use the new payment method.

### Other information:

---
  - https://app.asana.com/0/0/1208948912819349

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -35,6 +35,7 @@ class WooCommerce_My_Account {
 		\add_filter( 'woocommerce_billing_fields', [ __CLASS__, 'required_address_fields' ] );
 		\add_filter( 'woocommerce_get_checkout_url', [ __CLASS__, 'get_checkout_url' ] );
 		\add_filter( 'woocommerce_get_checkout_payment_url', [ __CLASS__, 'get_checkout_url' ] );
+		\add_filter( 'wc_stripe_update_subs_payment_method_card_statuses', [ __CLASS__, 'update_payment_methods_for_all_subs' ] );
 
 		// Reader Activation mods.
 		if ( Reader_Activation::is_enabled() ) {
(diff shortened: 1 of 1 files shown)


leogermani (approved):
It works as described, even though the flow was a litle confusing to me.

When the subscription is On-hold, I don't see the "Change payment method" in the subscription page (My Account). So I have to go to "Payment methods" page (which seems to be a silly default behavior from Woo)

I added a new payment method and confirmed this was applied to the existing on hold subscription, however, the new payment method was not set as the default method. Should we maybe set it as default then?

image

dkoo:
> It works as described, even though the flow was a litle confusing to me.
> 
> When the subscription is On-hold, I don't see the "Change payment method" in the subscription page (My Account). So I have to go to "Payment methods" page (which seems to be a silly default behavior from Woo)

Yes, it is a little confusing and it's all standard WooCommerce behavior. 

> I added a new payment method and confirmed this was applied to the existing on hold subscription, however, the new payment method was not set as the default method. Should we maybe set it as default then?

I thought about making this change in this PR too! But it's a bit more involved and doesn't really have any affect on the issue itself (of on-hold subscriptions not getting updated with the new payment method). "Default payment" in this case just means that it's the one that's preselected in the checkout flow if you have more than one saved payment method.

leogermani:
yeah, I think it's fine to move ahead... but it feels like a major gap in the Woo default experience! If I have an On-hold subscription I can't change the payment method for it? Strange

PR #3616, opened 2024-12-11
perf(data-events): queue dispatches to execute on shutdown

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

p1733922470397489-slack-C055SRT7WN4

Depending on the setup, a site can execute 5 dispatches when a reader registers:

 - `membership_status_active`
 - `membership_saved`
 - `reader_logged_in`
 - `reader_registered`
 - `network_user_updated`

The list grows if the registration happens during a subscription purchase.

Other actions on the site can also dispatch multiple events.

Each dispatch runs a `wp_remote_post()`. Even though it's a non-blocking execution with a timeout of 0.01, calling this function for every dispatch is unnecessary and costs performance.

This PR proposes queueing the dispatches to be executed in a single `wp_remote_post()` called on the `shutdown` hook. This hook runs right before PHP shuts down execution, including on fatal errors.

### How to test the changes in this Pull Request:

1. Make sure you have RAS configured and `NEWSPACK_LOG_LEVEL` set to at least `2` on `wp-config.php`
2. Add a webhook endpoint (https://webhook.site/ is your friend) assigned to the actions listed above
3. On a fresh session, register a new reader
4. Confirm you get the following log: (actions may vary according to your setup, but you should get at least `reader_registered` and `reader_logged_in` in the same run

```
Dispatching actions: "membership_status_active, membership_saved, reader_logged_in, reader_registered, network_user_updated".
```

5. Confirm you get the logs for handler execution as well:

```
[Newspack\Data_Events::handle]: Executing global action handlers for "membership_status_active".
[Newspack\Data_Events::handle]: Executing action handlers for "membership_status_active".
[Newspack\Data_Events::handle]: Executing global action handlers for "membership_saved".
[Newspack\Data_Events::handle]: Executing action handlers for "membership_saved".
[Newspack\Data_Events::handle]: Executing global action handlers for "reader_logged_in".
[...]

PR #3659, opened 2025-01-08
feat(data-events): unique event key

miguelpeixe:
### All Submissions:

Implement support for a unique event key to prevent multiple consecutive dispatches for events that always send the latest and complete payload in the same execution.

The key is determined when registering the action or listener:

```php
Data_Events::register_action( 'my_action', 'email' );
// or:
Data_Events::register_listener( 'my_hook', 'my_action', 'my_callback', 'email' );
```

In the example above, events dispatched matching the value in the property `email` of its payload data will only be executed once with the final payload data:

```php
Data_Events::dispatch( $action_name, [ 'email' => 'test@test.com', 'foo' => 'bar' ] );
Data_Events::dispatch( $action_name, [ 'email' => 'test@test.com', 'foo' => 'baz' ] );
```

Only 1 dispatch will be executed, containing `[ 'email' => 'test@test.com', 'foo' => 'baz' ]`.

This PR depends on #3616

1.
2.
3.

### Other information:

### DIFF
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -38,13 +38,20 @@ final class Data_Events {
 	private static $global_handlers = [];
 
 	/**
+	 * Dispatches queued for execution on shutdown.
+	 *
+	 * @var array[]
+	 */
+	private static $queued_dispatches = [];
+
+	/**
 	 * Initialize hooks.
 	 */
 	public static function init() {
 		\add_action( 'wp_ajax_' . self::ACTION, [ __CLASS__, 'maybe_handle' ] );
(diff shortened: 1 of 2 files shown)


miguelpeixe:
After some thought, this feature seems overkill and adds unnecessary complexity to data events. Each feature can handle their own deduplication to avoid consecutive requests or dispatches.

This is currently being done in [Network's content distribution](https://github.com/Automattic/newspack-network/blob/956219dad76cd2b020fc96004b45d9075380b514/includes/class-content-distribution.php#L76-L88) and soon we'll have something similar for ESP syncs.

PR #3660, opened 2025-01-08
fix(cli): verify-reader CLI command

dkoo:
### All Submissions:

Fixes the `wp newspack verify-reader` CLI command:

- Example parameter values cannot contain spaces in the docblock synopsis
- `--user` is a [global flag](https://make.wordpress.org/cli/handbook/references/config/#global-parameters) and can't be used as an associative arg for specific commands

1. On `trunk`, run `wp newspack verify-reader --user=<user ID or email address>` and observe that it fails with an error:

```
Warning: The `wp newspack verify-reader` command has an invalid synopsis part: [--user=<user
Warning: The `wp newspack verify-reader` command has an invalid synopsis part: ID|email>]
Error: Please provide a user ID or email address.
```

2. Check out this branch and test `wp newspack verify-reader --user-id=<user ID or email address>` with both a user ID and an email address and confirm that it works as expected. (Note the new flag name `--user-id` instead of `--user`)

### Other information:

### DIFF
--- a/includes/cli/class-ras.php
+++ b/includes/cli/class-ras.php
@@ -55,14 +55,14 @@ class RAS {
 	 *
 	 * ## OPTIONS
 	 *
-	 * [--user=<user ID|email>]
+	 * [--user-id=<id|email>]
 	 * : The user ID or email address associated with the reader account to verify.
 	 *
 	 * @param array $args Positional args.
 	 * @param array $assoc_args Associative args.
 	 */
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Works well.

Naming it `user-id` is not great because you can also pass the user email. I guess it's better than conflicting with a global argument, so I'm ok with it.

dkoo:
I had the same thought, but the way the command is written you _could_ actually pass any flag and it will work...

PR #3664, opened 2025-01-10
feat: update CLI to migrate guest authors

leogermani:
Closed in favor of https://github.com/Automattic/newspack-plugin/pull/3724

### All Submissions:

This PR changes the logic of the guest author migration script.

Basically, the script was checking if a guest user already existed as a regular WP user based on the user_login. I'm changing that to check for the user_nicename.

In a regular situation, the nicename will usually be the same thing as the login, but there are cases where they can be different. One of them is when you migrated users from another system and treated these fields in a different way in the database.

CAP uses the `$user->user_nicename` to create and match the author term, and that's why this is the field we should look at.

1. Create a few guest authors and assign posts to them. 
2. Run the script and make sure the guest contributors users are created and the posts are correctly assigned.
3. Create a guest author and make sure there is a wp user that has its nicename with the same value of the GA's login
4. Run the script and make sure no new WP user is created, and the posts that were assigned to that GA is assigned to this WP user

### Other information:

### DIFF
--- a/includes/cli/class-co-authors-plus.php
+++ b/includes/cli/class-co-authors-plus.php
@@ -155,6 +155,9 @@ class Co_Authors_Plus {
 		$unlinked_guest_authors = get_posts( $get_posts_args );
 		WP_CLI::line( sprintf( 'Found %d guest author(s) not linked to any WP User.', count( $unlinked_guest_authors ) ) );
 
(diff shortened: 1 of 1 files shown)


leogermani:
Closed in favor of https://github.com/Automattic/newspack-plugin/pull/3724

PR #3667, opened 2025-01-11
feat: add custom bylines

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208894036683495/1208945901567231 and https://app.asana.com/0/1208894036683495/1208945901567229

This PR adds a barebones custom byline feature to posts behind a `NEWSPACK_BYLINES_ENABLED` feature flag.

Screenshot 2025-01-13 at 19 17 44

1. Go to any post edit page
2. Confirm a new `Newspack Custom Byline` panel is present in the post editor sidebar
3. Enable the custom byline
4. Enter some text and confirm the byline appears in the post editor above the content
5. Add an author link to the byline by using the `<Author id=USER_ID>Name</Author>` tag
6. Confirm a link appears in the byline
7. Change the author ID to one that is not on the site
8. Confirm the link disappears but the author name is still present in the byline
9. Try to add any invalid tag (`<Invalid>`) and confirm an alert appears warning you only `Author` is supported
10. Disable custom byline and confirm the byline disappears from the post editor
11. Re-enable and confirm byline reappears as it was before

### Other information:

### DIFF
--- a/src/bylines/index.js
+++ b/src/bylines/index.js
@@ -1,7 +1,8 @@
+/* globals newspackBylines */
+
 /**
  * WordPress dependencies
  */
-import apiFetch from '@wordpress/api-fetch';
 import { ToggleControl, TextareaControl } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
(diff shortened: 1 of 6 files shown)


src/bylines/index.js
```diff
@@ -0,0 +1,111 @@
...
+			if ( /<Author id=(\d+)>/.test( text ) ) {
+				console.info( 'Updating byline…' ); // eslint-disable-line no-console
+				text = text.replace( /<Author id=(\d+)>([^<]+)<\/Author>/g, ( match, authorId, authorName ) => {
+					const authorData = authors.find( a => a.id === parseInt( authorId ) );
+					return authorData ? `<a href="/author/${ authorData.name }">${ authorName }</a>` : authorName;
+				} );
+			}
+			bylineEl.innerHTML = text;
+		}
+	};
```
chickenn00dle: This is another super hacky solution to getting the byline to render in the editor. Ideally we replace this when implementing the full feature.

chickenn00dle:
@leogermani, I've got this PR mostly ready for review assuming its enough to render the byline in the post editor only for the purposes of migration. I didn't want to do too much work on rendering this on the front-end without understanding the requirements fully here.

That said, if we'd rather figure this out now, I can get that added in the PR as well.

leogermani (changes requested):
The link was not added for me att first. Then I noticed you were checking to see if the ID exists, but the fetch for authors in the `useEffect` only returns 10 authors, so it will never find the ID I was using.

But even when I used one of the returned IDs, the link was broken, with nothing after the `/author/` part.

I don't think we need to verify anything at this point. Just create a link. And I think we can go `siteurl.com/?author=$ID`, no need to get their slugs.

Finally, let's also add a filter to `the_content` and render the byline on the front end as well. This will make QA easier

chickenn00dle:

> I don't think we need to verify anything at this point. Just create a link. And I think we can go siteurl.com/?author=$ID, no need to get their slugs.

Sure thing! Updated in https://github.com/Automattic/newspack-plugin/pull/3667/commits/f8ec6b84500844bfa4a368b67bbb1759d2195e3e

> let's also add a filter to the_content and render the byline on the front end as well. This will make QA easier

Makes sense. Added in https://github.com/Automattic/newspack-plugin/pull/3667/commits/f8ceb4448c89fb4ac4cb320732db26651a7dbc83

leogermani (approved):
I've added a few more things to the Readme. Thanks for the changes!

PR #3673, opened 2025-01-13
feat(my-account): reCAPTCHA check on add payment method

dkoo:
### All Submissions:

Adds a reCAPTCHA check to the form to add payment methods in My Account.

We'll want to test with both reCAPTCHA v2 and v3.

1. Enable reCAPTCHA v2 in Newspack > Connections and check out this branch.
2. At [this line](https://github.com/Automattic/newspack-plugin/blob/hotfix/protect-add-payment-method/includes/class-recaptcha.php#L406-L407), add `error_log( print_r( $captcha_verify, true ) );` so we can confirm that we're verifying with the reCAPTCHA service after form submit.
3. Register a new reader account, log in, and verify as needed to view My Account.
4. Visit `<site URL>/my-account/add-payment-method/` and add a payment method.
5. (when testing v2 only) Confirm that you get a v2 widget challenge when you click the "Add payment method" submit button.
6. After submitting the form, confirm that the verification result is logged to debug.log (after adding the `error_log` in step 2).
7. To test a rejection, add `return new \WP_Error( 'newspack_test_error', 'Test error message' );` at [this line](https://github.com/Automattic/newspack-plugin/blob/hotfix/protect-add-payment-method/includes/class-recaptcha.php#L379-L380).
8. Repeat the "Add payment method" form submission and confirm that the submission is rejected with your test error message (note that the message might disappear after a second, but this seems to be a WooCommerce behavior and not related to this PR). Visit `<site URL>/my-account/payment-methods` and confirm that a new payment method was NOT saved.
9. Repeat all testing steps with reCAPTCHA v3.

### Other information:

---
  - https://app.asana.com/0/0/1209016655921587

### DIFF
--- a/includes/class-recaptcha.php
+++ b/includes/class-recaptcha.php
@@ -28,9 +28,13 @@ final class Recaptcha {
 		// Add reCAPTCHA to the Woo checkout form.
 		\add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_recaptcha_v2_to_checkout' ] );
 		\add_action( 'woocommerce_checkout_after_customer_details', [ __CLASS__, 'add_recaptcha_v3_to_checkout' ] );
+		\add_action( 'woocommerce_add_payment_method_form_bottom', [ __CLASS__, 'add_recaptcha_v3_to_checkout' ] );
 
 		// Verify reCAPTCHA on checkout submission.
 		\add_action( 'woocommerce_checkout_process', [ __CLASS__, 'verify_recaptcha_on_checkout' ] );
+
+		// Verify reCAPTCHA when adding new payment method.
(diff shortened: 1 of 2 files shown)


includes/class-recaptcha.php
```diff
@@ -28,9 +28,13 @@ public static function init() {
 		// Add reCAPTCHA to the Woo checkout form.
 		\add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_recaptcha_v2_to_checkout' ] );
 		\add_action( 'woocommerce_checkout_after_customer_details', [ __CLASS__, 'add_recaptcha_v3_to_checkout' ] );
+		\add_action( 'woocommerce_add_payment_method_form_bottom', [ __CLASS__, 'add_recaptcha_v3_to_checkout' ] );
```
leogermani, line 31: is this change related?

leogermani: ah nvm, it is... the name of the method misled me

leogermani (approved):

PR #3678, opened 2025-01-15
feat: rate limit checkout attempts

dkoo:
### All Submissions:

Adds support for rate limiting checkout attempts by user. If a `NEWSPACK_CHECKOUT_RATE_LIMIT` environment constant is defined as an integer, checkout attempts that occur within that number of seconds of a prior checkout attempt by the same user (if logged in) or IP (if not logged in) will be rejected before WooCommerce sends payment info to any payment gateways. 

If WooCommerce detects any other checkout validation errors that would prevent the checkout from occurring (such as missing required fields), we won't count this as a checkout attempt, so that readers can attempt successive checkouts in order to correct errors without being rate limited. So in practice, only successful transactions and checkout requests which would result in declined payments should be rate-limited.

The mechanism used is WooCommerce's built-in [`WC_Rate_Limiter` utility](https://woocommerce.github.io/code-reference/classes/WC-Rate-Limiter.html). This offers a lightweight API to check and limit the rate of particular actions by arbitrary ID string.

Also adds rate-limiting for adding new payment methods in My Account, but because of how this feature is split between the main WooCommerce plugin and many payment gateway extension plugins, it doesn't necessarily rate limit declined payment methods (thus still allowing for card testing with some payment gateways). This will need to be handled in a separate PR, possibly via JS since each payment method behaves differently when deciding when/whether to contact the payment processor.

1. Check out this branch. Optionally disable reCAPTCHA in Newspack > Connections to make testing easier.
2. In `wp-config.php`, define `NEWSPACK_CHECKOUT_RATE_LIMIT` as something long enough for manual testing like `60` or `90`.
3. As an anonymous reader in a new browser session, start a Woo checkout modal checkout. After completing the first screen (payment details), click "Back" from the second screen and confirm that you can proceed to the second screen again without being rate-limited (validation-only checkout requests are not rate-limited).
4. Attempt to complete the checkout but with invalid input: for example, by using dev tools to edit the value of the email address field to an invalid email address, then submitting. Confirm that the response includes the error(s) about the invalid input, but not anything about rate limiting. You can also use WP CLI or error logging to confirm that a new transient is not created from this invalid checkout request.
5. Complete the checkout with valid inputs, but use a [test Stripe payment which will be declined](https://docs.stripe.com/testing#declined-payments).
6. After getting the "payment declined" error message and within the number of seconds defined in step 2, complete checkout again with a valid/non-declined payment method and confirm that you get an error message to `Please wait a moment before trying to complete this transaction again`.
7. Wait until the number of seconds elapses and then submit once more, and confirm that the transaction completes successfully.
8. Repeat with a new checkout session while logged into a reader account with saved payment details.
9. Repeat again with a regular (non-modal) Woo checkout.

#### Testing rate-limiting of adding new payment methods

1. Log into My Account as a reader and visit the Payment Methods menu link to add a new payment method (alternatively, just plug in `/my-account/add-payment-method` since this URL isn't blocked even if the menu item is hidden).
2. Add a new valid payment method, then try to immediately add another new payment method and confirm that the request is rejected with a `Please wait a moment before trying to add a new payment method` error message. (Note that the error message may flash and disappear after a moment, but this seems to be a Woo/My Account bug and isn't related to the changes in this PR.)
3. Wait until the number of seconds elapses and then try again, and confirm that the payment method is successfully added.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -174,14 +174,15 @@ class WooCommerce_Connection {
 	 * Check the rate limit for the current user or IP.
 	 * Currently locked behind a NEWSPACK_CHECKOUT_RATE_LIMIT environment constant, for controlled rollout.
 	 *
+	 * @param string $action_name   The action the user is trying to perform.
 	 * @param string $error_message Error message to display or return if the user should be rate-limited.
 	 * @param bool   $return_error  If true and the user should be rate-limited, return a WP_Error with the given message instead of a boolean value.
 	 *
 	 * @return bool|WP_Error True or WP_Error if the rate limit is exceeded, false otherwise.
 	 */
-	public static function rate_limit_by_user( $error_message = '', $return_error = false ) {
+	public static function rate_limit_by_user( $action_name, $error_message = '', $return_error = false ) {
 		$rate_limited = false;
-		if ( ! defined( 'NEWSPACK_CHECKOUT_RATE_LIMIT' ) ) {
+		if ( ! defined( 'NEWSPACK_CHECKOUT_RATE_LIMIT' ) || ! class_exists( 'WC_Rate_Limiter' ) ) {
 			return $rate_limited;
 		}
 		if ( ! $error_message ) {
@@ -191,7 +192,7 @@ class WooCommerce_Connection {
 		$now        = time();
 		$rate_limit = defined( 'NEWSPACK_CHECKOUT_RATE_LIMIT' ) ? (int) NEWSPACK_CHECKOUT_RATE_LIMIT : 90; // Number of seconds to wait before allowing the same user to attempt another checkout action. Default: 90.
 		if ( 0 === $rate_limit ) {
-			return $rate_limited; // If $rate_limit is 0 seconds, bail early to avoid creating a non-expiring transient.
+			return $rate_limited; // If $rate_limit is 0 seconds, no need to proceed.
 		}
 
 		// If not logged in, use IP.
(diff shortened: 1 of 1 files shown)


dkoo:
@miguelpeixe refactored this to use `WC_Rate_Limiter` in lieu of transients. I think this is ready for a review now!

chickenn00dle (commented):
Nice work @dkoo!

Works as described. Just one question regarding whether we should be including successful transactions here. 

If I do something like make a successful donation, then decide to purchase a monthly subscription product, I have to wait rate limit duration.

It's probably unlikely readers will do these things back to back like this in such a short time, but just in case, should we not trigger the rate limit on successful transactions? Or is there a scenario where a bad actor might want to flood a site with successful orders?

dkoo:
Great question! My thinking was that in a card-testing scenario, some of the tested cards might result in successful orders, so we should still rate limit those.

chickenn00dle (approved):
> My thinking was that in a card-testing scenario, some of the tested cards might result in successful orders, so we should still rate limit those.

Makes sense, although I don't think we would see back to back successful tested cards. Rate limiting the failures is probably good enough right? The only scenario where this might be necessary is if the successful card is reused, which I don't think would happen, but am not sure if there is some strategy I'm not aware of.

My concern is we inadvertently create friction for more real readers than we are for these card testing attackers.

I'm not super familiar with these attacks though, so will defer to you and approve this. Its still worth noting we should look out for reports from real readers in the future.

dkoo:
That's definitely a valid concern. I suspect that a reader is extremely unlikely to attempt two real transactions within a 60-second period, but let's stay vigilant for publisher/reader complaints.

PR #3681, opened 2025-01-16
fix(wcs): expire manual subscriptions after on-hold duration

chickenn00dle:
### All Submissions:

Part of https://app.asana.com/0/1200550061930446/1209066781737999/f

This PR adds a new scheduled action to expire manual subscriptions after the specified on-hold duration:

Screenshot 2025-01-16 at 17 19 11

1. Set at least two active subscriptions to manual, either as the reader via my-account or as admin via the edit subscriptions page in the edit billing menu.
2. Manually process renewals for both subscriptions via the edit subscriptions page.
3. In WPAdmin, WooCommerce > Status > Scheduled Actions, look for two pending `newspack_expire_manual_subscription` events that are scheduled to run X days after the next payment date, where X is what you have set as your site's on-hold duration (WooCommerce > Settings > Subscriptions > On-hold Duration) plus 7 days
4. As a reader, renew one of the on-hold subscriptions via my-account
5. Verify the subscription is now active as admin, and the matching scheduled action is no longer pending.
6. As admin, manually trigger the other `newspack_expire_manual_subscription` action
7. Verify the matching subscription is now expired

### Other information:

---
  - https://app.asana.com/0/0/1209066782957092

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-on-hold-duration.php
+++ b/includes/plugins/woocommerce-subscriptions/class-on-hold-duration.php
@@ -126,9 +126,9 @@ class On_Hold_Duration {
 	 */
 	public static function maybe_schedule_expiration( $order_id, $subscription ) {
 		if ( 'on-hold' === $subscription->get_status() ) {
-			$on_hold_duration    = self::get_on_hold_duration();
-			$on_hold_duration_ts = 0 < $on_hold_duration ? $on_hold_duration * DAY_IN_SECONDS : HOUR_IN_SECONDS;
-			$timestamp           = $subscription->get_time( 'next_payment' ) + $on_hold_duration_ts;
+			$default_grace_period = 7 * DAY_IN_SECONDS; // 7 days, the number of days the retry system normally waits before marking a subscription as expired.
+			$on_hold_duration     = self::get_on_hold_duration() * DAY_IN_SECONDS;
+			$timestamp            = $subscription->get_time( 'next_payment' ) + $default_grace_period + $on_hold_duration;
 			as_schedule_single_action( $timestamp, self::AS_HOOK, [ $subscription->get_id() ], self::AS_GROUP );
 		}
 	}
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Works as described.

I tested one other thing I'm not sure if there's an issue there or not.

While the subscription is on hold, and there's a scheduled action to expire it, if an admin manually sets the subscription as Active again it won't delete the scheduled action.

However, when that action runs, it won't expire the subscription because it's active, so no idea if it's a problem or not. thoughts? But if it goes to on hold again, then it will be expired...

Also, the expiration date is set to 7 days + on hold duration + the next payment date. So if I process a renewal but they still have 15 paid days, there will be an extra 15 days there... All good, just highlighting because it was different from the testing instructions

chickenn00dle:
Thanks for the review on this one as well @leogermani!

> While the subscription is on hold, and there's a scheduled action to expire it, if an admin manually sets the subscription as Active again it won't delete the scheduled action.

> However, when that action runs, it won't expire the subscription because it's active, so no idea if it's a problem or not. thoughts? But if it goes to on hold again, then it will be expired...

Yeah. I purposefully made the scheduled action do nothing if the subscription is not on-hold to account for any scenarios where the admin wants to handle a subscription manually. I also am only removing the scheduled action only when a successful payment is made to account for any strange situations where an admin changes a subscriptions status multiple times with the final change being on-hold. My assumption here is, if a payment isn't made and the subscription is on-hold by the time we reach the end of the total grace period, we should expire. WDYT?

> Also, the expiration date is set to 7 days + on hold duration + the next payment date. So if I process a renewal but they still have 15 paid days, there will be an extra 15 days there... All good, just highlighting because it was different from the testing instructions

Ah. I think what you've written here is the expected behavior. My bad. The reader should not lose any time on their subscription when it is renewed early.

leogermani:
LGTM

PR #3691, opened 2025-01-20
Remove support for legacy form checkout

claudiulodro:
### All Submissions:

Removes support for the legacy form-based WooCommerce checkout. Other than for backwards compatibility, I believe the main purpose it's still around is to support browsers that have JS disabled, but that's not something we support (especially since it wouldn't work with reCaptcha).

Simplest way to verify that it's been removed:

1. Add a `die( 'ATTEMPTING FORM CHECKOUT' );` [here in your Woo plugin](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-form-handler.php#L382) in the `checkout_action` method.
2. Put this script somewhere:
```
<?php

$site = 'http://newspackdev.local';

$body = [
	'woocommerce_checkout_place_order'    => 1,
];

var_dump( wp_remote_post( $site, [ 'body' => $body, 'timeout' => 50 ] ) );
```
4. Run it with `wp eval-file scriptfile.php` before applying this patch. Observe that the response contains that `ATTEMPTING FORM CHECKOUT` string within it.
5. Apply the patch. Run the script again. Observe that the response doesn't contain the string.
6. Verify regular Checkout screen and modal checkout work.

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -38,6 +38,7 @@ class WooCommerce_Connection {
 		include_once __DIR__ . '/class-woocommerce-duplicate-orders.php';
 
 		\add_action( 'admin_init', [ __CLASS__, 'disable_woocommerce_setup' ] );
+		\add_action( 'wp_loaded', [ __CLASS__, 'disable_legacy_form_checkout' ], 1 );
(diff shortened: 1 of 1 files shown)


dkoo (approved):

dkoo (commented):
@claudiulodro this could be a hotfix, if we wanted it to go out sooner

claudiulodro:
I'm OK sending it through the regular release flow just to benefit from the alpha release testing/process. It's pretty straightforward to add this code snippet to a site in an emergency in the meantime though.

Issue #123, opened 2019-06-26
Should be able to save settings when Stripe not enabled


claudiulodro:
**To Reproduce**
1. Go to Subscriptions Onboarding wizard
2. Go through settings. Try and save with Stripe not enabled. Observe error.

**Screenshots**
![localhost_wp-admin_admin php_page=newspack-subscriptions-onboarding-wizard](https://user-images.githubusercontent.com/7317227/60199556-0ea73200-97f9-11e9-82f1-26f17a25accd.png)

**Expected behavior**
Key info should not be validated when Stripe is not enabled.

PR #3692, opened 2025-01-21
fix(recaptcha): improvements for reCAPTCHA v2 + modal checkout

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-blocks/pull/2028, implements improvements to the reCAPTCHA v2 integration in a modal checkout context. Also introduces a new option (via environment constant, for now) to allow third-party WooCommerce extensions to load their assets in modal checkout, to allow for greater flexibility. Not all Woo extensions are guaranteed to work with the modal checkout, but many will.

These changes came about due to issues some publishers saw when using the [ELEX Address Verification plugin](https://wordpress.org/plugins/address-validation-address-auto-complete/)—this plugin triggers actions on the `click` event of the `#place_order` button in checkout, which is also bound to our reCAPTCHA v2 widget execution. This causes a conflict where only one of these bound events will fire on click—but because both ELEX and reCAPTCHA can block a checkout request upon failure, enabling both will make it impossible to complete checkout.

This PR avoids such conflicts by creating a clone of the `#place_order` button in our modal checkout template, hiding the original button element, and binding our reCAPTCHA v2 widget execution to the clone instead of the original. When the reCAPTCHA widget fires its success callback (either upon initial pass or challenge pass), it will trigger a `click` event on the "real" `#place_order` button, which will allow other extensions with events bound to that button to do their own thing.

To be clear, this PR doesn't indicate full Newspack support for the ELEX plugin—but it will allow for publishers to test and use plugins like ELEX alongside Newspack, if the resulting experience is acceptable.

#### Pre-testing setup

1. Add `define( 'NEWSPACK_ALLOW_ALL_CHECKOUT_SCRIPTS', true );` to wp-config.php to allow unsupported Woo extensions to load their assets in modal checkout.
2. Connect and enble reCAPTCHA v2 in Newspack > Connections.
3. Install, activate and configure an unsupported Woo extension plugin which would normally conflict with reCAPTCHA v2. To test with [ELEX](https://wordpress.org/plugins/address-validation-address-auto-complete/), install the plugin, then sign up for a free EasyPost account to obtain a test API key (or DM me for a key you can use) and enter the key in the API credentials screen (`https://siteurl/wp-admin/admin.php?page=wc-settings&tab=elex_address_autocomplete_validation&section=elex-api-credentials`). Also make sure to enable address validation via EasyPost in the Address Validation settings page:

Screenshot 2025-01-21 at 12 25 31 PM

4. If using ELEX, make sure to enable address fields in **Newspack > Reader Revenue > Donations** or ELEX will block all checkout requests due to missing address inputs. Alternatively, test using a shippable product, which will always show both billing and shipping address fields in modal checkouts.

#### Testing steps

1. Check out this branch and https://github.com/Automattic/newspack-blocks/pull/2028.
2. As a reader, start a modal checkout via Donate or Checkout Button blocks.
3. Fill in all required address fields but using a fake (non-existent) address on modal screen 1.
4. On modal screen 2, complete payment details and click "Complete transaction" to start the checkout request.
5. Confirm that you see a reCAPTCHA v2 widget challenge (make sure your reCAPTCHA settings in Google are set to highest security to ensure that you'll always be challenged).
6. After completing the reCAPTCHA challenge, confirm that you get an error response: `Address verification failed`
7. Go back to modal screen 1 and update the address fields using a real address (I tested with a US-based address—non-US addresses are currently untested).
8. Go back to modal screen 2 and submit payment details again. Confirm that you get another reCAPTCHA v2 challenge.
9. After completing the reCAPTCHA challenge, this time confirm that the address verification succeeds with a popup:

Screenshot 2025-01-21 at 12 35 02 PM

10. After accepting the suggested address, confirm that the transaction completes successfully.
11. Smoke test transactions with reCAPTCHA v3 enabled, and reCAPTCHA disabled entirely. In both cases the address verification should occur as soon as you first click "Complete transaction" with all required fields filled out.

### Other information:

---
  - https://app.asana.com/0/0/1209184701915831

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-products.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-products.php
@@ -52,7 +52,7 @@ class WooCommerce_Products {
 		return [
 			'newspack_autocomplete_orders' => [
 				'id'            => '_newspack_autocomplete_orders',
-				'wrapper_class' => 'show_if_virtual',
+				'wrapper_class' => '',
 				'label'         => __( 'Auto-complete orders', 'newspack-plugin' ),
 				'description'   => __( 'Allow orders containing this product to automatically complete upon successful payment.', 'newspack-plugin' ),
 				'default'       => 'yes',
@@ -197,11 +197,7 @@ class WooCommerce_Products {
 	 * @param WC_Product $product The product associated with this order item.
 	 */
 	public static function require_order_processing( $needs_proccessing, $product ) {
-		if ( $product->is_virtual() ) {
-			return self::get_custom_option_value( $product, 'newspack_autocomplete_orders' ) ? false : $needs_proccessing;
-		}
-
-		return $needs_proccessing;
+		return self::get_custom_option_value( $product, 'newspack_autocomplete_orders' ) ? false : $needs_proccessing;
 	}
 }
 
--- a/newspack.php
+++ b/newspack.php
@@ -14,7 +14,7 @@
 
 defined( 'ABSPATH' ) || exit;
 
-define( 'NEWSPACK_PLUGIN_VERSION', '5.12.0' );
+define( 'NEWSPACK_PLUGIN_VERSION', '5.12.0--recaptcha-test.8' );
 
 // Define NEWSPACK_PLUGIN_FILE.
 if ( ! defined( 'NEWSPACK_PLUGIN_FILE' ) ) {
--- a/newspack.php
+++ b/newspack.php
@@ -14,7 +14,7 @@
 
 defined( 'ABSPATH' ) || exit;
 
-define( 'NEWSPACK_PLUGIN_VERSION', '5.12.0--recaptcha-test.8' );
+define( 'NEWSPACK_PLUGIN_VERSION', '5.12.0' );
 
 // Define NEWSPACK_PLUGIN_FILE.
 if ( ! defined( 'NEWSPACK_PLUGIN_FILE' ) ) {
--- a/src/other-scripts/recaptcha/index.js
+++ b/src/other-scripts/recaptcha/index.js
@@ -50,11 +50,16 @@ function renderV2Widget( form, onSuccess = null, onError = null ) {
 			form.appendChild( hiddenField );
 		}
 		hiddenField.value = token;
-		const buttons = [
-			...form.querySelectorAll( 'input[type="submit"], button[type="submit"]' )
-		];
 		form.setAttribute( 'data-recaptcha-validated', '1' );
-		form.requestSubmit( buttons[ buttons.length - 1 ] );
+
+		// If the form has a #place_order button, click it.
+		const placeOrder = form.querySelector( '#place_order' );
+		if ( placeOrder ) {
+			placeOrder.click();
+		} else {
+			form.requestSubmit( form.querySelector( 'input[type="submit"], button[type="submit"]' ) );
+		}
+
 		refreshV2Widget( form );
 	};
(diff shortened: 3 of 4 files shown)


laurelfulford (changes requested):
Overall things look good! I ran into one maybe-issue that I included at the end. 

I ran these checkouts with ELEX and Braintree, and ELEX and Stripe. In most cases, things work as described; I:

* Entered the wrong address fine on the first screen
* Got and passed reCAPTCHA v2 prompt, entered my Braintree test code when using that gateway, and got the Address failed message (not sure if we can change this to remove the colon?):

CleanShot 2025-01-27 at 11 05 51

* I hit the back button, corrected it, entered the credit card info again.
* Got reCAPTCHA checked again, and then saw the email correction screen:

CleanShot 2025-01-27 at 11 07 38

* I picked the original address, entered the Braintree code again when using that gateway, and the transaction went through.
* I also smoke tested without reCAPTCHA and with v3 with both gateways.

The only issue I ran into was using Braintree and subscription products with free trials. I get this error when I try to submit the first screen (it comes from the Braintree plugin):

CleanShot 2025-01-27 at 11 18 34

I found [this suggestion](https://wordpress.org/support/topic/braintree-free-trials/#post-16251716), and it does seem to fix the issue but it means no payment method is collected on checkout; best I can tell, readers would be prompted to add payment information later on, which seems subpar:

CleanShot 2025-01-27 at 11 22 30

I don't think this will be an issue for publishers who currently need this support, but I wanted to flag in case it's something that should be fixed before this is merged. I'm making this as "request changes", but that's the only issue I ran into so it should be OK if that's OK!

dkoo:
@laurelfulford thanks for the testing and feedback!

> got the Address failed message (not sure if we can change this to remove the colon?):

The message can be customized in ELEX plugin settings ("Customise" tab).

Re: the issue with Braintree + free trial subscription products, since this seems to be a bug in the Braintree plugin itself and isn't introduced by the modal checkout flows, I don't think think we should let that block us from moving ahead. We should ensure that sites who want to use Braintree are aware of this limitation. For the record, I did attempt to tweak the priority value on the `woocommerce_cart_needs_payment` hook, but while this allowed the transaction to complete, Braintree still failed to collect payment, resulting in a failed order and on-hold subscription being created.

Also noting a bug I noticed in the ELEX plugin, which you can actually see in the screenshots of the address validation popup above: validated addresses drop the "Street 2" field, so that 

> 999 Street Name
> #123
> City, CA, 99999

becomes 

> 999 STREET NAME
> CITY, CA, 99999

Again, this is a bug originating in the ELEX plugin, so fixing it isn't within the scope of our code.

laurelfulford (approved):
Thanks for following up on all that stuff, @dkoo!

> The message can be customized in ELEX plugin settings ("Customise" tab).

Ah! That's perfect -- nice that they can add some direction there, too 🙂 

> Re: the issue with Braintree + free trial subscription products, since this seems to be a bug in the Braintree plugin itself and isn't introduced by the modal checkout flows, I don't think think we should let that block us from moving ahead. 

Agreed! Thanks for verifying this. 

> Also noting a bug I noticed in the ELEX plugin, which you can actually see in the screenshots of the address validation popup above... Again, this is a bug originating in the ELEX plugin, so fixing it isn't within the scope of our code.

Good catch, and totally fair! As an aside, Canada Post has an address autocomplete thing you can use here that always bungles my address 😆 These things aren't infallible! 

This all looks good to me - giving it a 🚢

dkoo:
The changes here are for an extreme edge case, so I don't think it needs to be documented.

PR #3707, opened 2025-01-28
fix(ia): tweak syndication plugin cards behavior

dkoo:
### All Submissions:

Tweaks the `WizardsPluginCard` component slightly:

- Allows for a `reloadOnActivation` prop to be passed. If set to `false`, the card will not trigger a page reload after installing or activating its plugin. Deactivating the plugin will still result in a page reload, to clear out dead WP admin menu links. Default: `true` so that component instances not specifically touched by this PR are not affected.
- If passed an `editLink` prop, will continue to show the "Configure" link to the plugin's settings page even after the plugin has been fully setup.
- If the plugin isn't yet installed, calls the `Plugin_Manager::activate()` method instead of the `install()` method, as `activate()` will automatically install the plugin first if not installed. This lets the user skip an extra click and REST API roundtrip to install then activate the plugin.

Then, applies these changes to instances of the `WizardsPluginCard` component in the Syndication wizard page:

- Sets both Apple News and Distributor cards to `reloadOnActivation: false`—these don't really require a refresh immediately upon activation as the "Configure" links that appear will work without a page reload.
- Updates the plugin title and description for Publish to Apple News to exactly match those from the plugin itself.
- Adds an `editLink` prop for the Distributor plugin which leads to the "External Connections" settings page.

1. Start with the Apple News and Distributor plugins not installed on your testing site.
2. Navigate to **Newspack > Settings > Syndication**.
3. Click the toggle button on the left-hand side of each plugin card. Confirm that upon success, both plugins are in a fully-activated state and show their "Configure" buttons, and that the "Configure" buttons lead to the correct plugin settings page.
5. Delete both plugins and repeat step 3, but this time use the "Install <plugin name>" button on the right-hand side of the card. Confirm that the resulting behavior is the same as using the toggle button.
4. Toggle off both plugins using the toggle button and confirm that the page reloads after toggling off each one. Also confirm that the plugins are actually set to a deactivated status.
5. Smoke test the plugin cards elsewhere throughout the wizard pages and confirm that they still work as intended. Most exist in the following Settings subpages:

  - Connections
  - Social
  - Emails

### Other information:

### DIFF
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -562,7 +562,7 @@ class Plugin_Manager {
 		}
 
 		if ( \is_plugin_active( $installed_plugins[ $plugin_slug ] ) ) {
-			return new WP_Error( 'newspack_plugin_already_active', __( 'The plugin is already active.', 'newspack-plugin' ) );
+			return true;
 		}
 
 		$activated = activate_plugin( $installed_plugins[ $plugin_slug ] );
--- a/includes/class-plugin-manager.php
+++ b/includes/class-plugin-manager.php
@@ -199,7 +199,7 @@ class Plugin_Manager {
 			],
 			'publish-to-apple-news'       => [
 				'Name'        => \esc_html__( 'Publish to Apple News', 'newspack-plugin' ),
-				'Description' => \esc_html__( 'Export and synchronize posts to Apple format', 'newspack-plugin' ),
+				'Description' => \esc_html__( 'Export and synchronize posts to Apple format.', 'newspack-plugin' ),
 				'Author'      => \esc_html__( 'Alley Interactive', 'newspack-plugin' ),
 				'AuthorURI'   => \esc_url( 'https://www.alleyinteractive.com' ),
 				'PluginURI'   => \esc_url( 'https://github.com/alleyinteractive/apple-news' ),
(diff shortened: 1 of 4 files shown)


miguelpeixe (approved):
Great changes and it's working well.

An issue I'm having on the epic branch and here, not introduced by these changes, is a weird inability to deactivate Apple News.

I followed the instructions to delete the plugin and install/activate via the card, but it is still unable to deactivate afterward.

https://github.com/user-attachments/assets/b7583443-d50e-4fa6-95f1-31204aacf4f7

I'll investigate this further and we can tackle it in a separate PR.

miguelpeixe:
I think I got it. It's this line:

https://github.com/Automattic/newspack-plugin/blob/ca1103b4fe0b489c2a4f36674d121ef9e4f0ceae/src/wizards/wizards-plugin-card.tsx#L341

I don't have the plugin configured (`pluginState.configured`), just activated.

This line needs to be `! statuses.isActive ? onActivate() : onDeactivate();`

You can test it by removing the `newspack_has_configured_plugin_publish-to-apple-news` option and trying to deactivate the plugin.

Another weird UI glitch I just found is when I click to Configure for the first time it enters the loading state with the card toggled off before redirecting.

dkoo:
@miguelpeixe it seems like updating that line fixes both issues you mentioned—does it for you?

miguelpeixe:
Yep :shipit:

PR #3726, opened 2025-02-04
fix(esp-sync): get last payment amount from actual completed order

dkoo:
### All Submissions:

Fixes a bug where the `NP_Last Payment Amount` value isn't always accurate. Currently we assume that if the subscription is active, the last payment must match its renewal price, but that's not always the case—for example, for new subscriptions that include a free trial period. In this scenario, the subscription in free trial mode will be active and have a successful completed parent order, but that parent order's amount will be `0` instead of the recurring amount.

Closes `1200550061930446/1209296877142332`.

1. On a site with RAS setup and connected, create a subscription product with a non-zero cost and a free trial period.
2. On `release`, purchase a subscription for that product as a new reader.
3. In the connected ESP or via logging, observe that the `NP_Last Payment Amount` is set to the cost of the subscription, even though the `NP_Total Paid` value is correctly `0`.
4. Check out this branch and trigger a resync of this contact via any means.
5. Confirm that `NP_Last Payment Amount` is now `0` after the resync.
6. Trigger a manual early renewal of the subscription to charge the customer, and confirm that `NP_Last Payment Amount` is now the amount charged in the renewal order.
7. Smoke test purchasing a subscription product and/or resyncing a contact who has previously purchased a subscription WITHOUT a free trial period and confirm that the `NP_Last Payment Amount` is still accurate.

### Other information:

---
  - https://app.asana.com/0/0/1209296877142332

### DIFF
--- a/includes/reader-activation/sync/class-woocommerce.php
+++ b/includes/reader-activation/sync/class-woocommerce.php
@@ -179,6 +179,42 @@ class WooCommerce {
 	}
 
 	/**
+	 * Get the amount of the last payment associated with the given subscription.
+	 *
+	 * @param \WC_Subscription $subscription Subscription object.
+	 *
+	 * @return float The amount of the last payment.
+	 */
+	private static function get_last_payment_amount( $subscription ) {
+		$last_order = $subscription->get_last_order(
+			// The whole WC_Order object, not just the ID.
+			'all',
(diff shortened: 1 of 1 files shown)


naxoc (approved):
Yup! I tested and smoke tested and it looks good. 

ok

PR #3728, opened 2025-02-05
fix(esp-sync): transform outgoing dates to site timzeone

naxoc:
### All Submissions:

We would like the dates that are synced to ESPs to be in the site's timezone instead of UTC. Since we already store (and have stored) dates as UTC, this PR suggests just transforming them. That might be over-simplified and I'm also not sure I'm doing it in the right place, but here is a shot :) Suggestions very welcome on where to place this code (in a filter somewhere?).

0. Set your site timezone to something far way from UTC (to make it easier to spot)
1. Make sure you have ESP contacts to sync
2. Run `wp newspack esp sync` and either use a debugger or log statements to see that the dates change. I used `NP_Last Payment Date` and `NP_Registration Date` as test material.

### Other information:

---
  - https://app.asana.com/0/0/1209271919879007

### DIFF
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -8,7 +8,6 @@
 namespace Newspack\Reader_Activation;
 
 use Newspack\Reader_Activation;
-use Newspack\Reader_Activation\Sync\Metadata;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -108,11 +107,6 @@ class ESP_Sync extends Sync {
 		 * @param string $context The context of the sync.
 		 */
 		$contact = \apply_filters( 'newspack_esp_sync_contact', $contact, $context );
-
-		if ( ! empty( $contact['metadata']['registration_date'] ) ) {
-			// Convert date from UTC to the site's timezone.
-			$contact['metadata']['registration_date'] = get_date_from_gmt( $contact['metadata']['registration_date'] );
-		}
 
 		$contact = Sync\Metadata::normalize_contact_data( $contact );
 
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -116,8 +116,11 @@ class ESP_Sync extends Sync {
 			$timezone     = \wp_timezone();
 			$utc_timezone = new \DateTimeZone( 'UTC' );
 			foreach ( $contact['metadata'] as $key => $value ) {
-				if ( is_string( $value ) && false !== \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value ) ) {
-					$date = \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value, $utc_timezone );
+				if ( empty( $value ) || ! is_string( $value ) ) {
+					continue;
+				}
+				$date = \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value, $utc_timezone );
+				if ( false !== $date ) {
 					$date->setTimezone( $timezone );
 					$contact['metadata'][ $key ] = $date->format( Metadata::DATE_FORMAT );
 				}
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -116,7 +116,7 @@ class ESP_Sync extends Sync {
 			$timezone     = \wp_timezone();
 			$utc_timezone = new \DateTimeZone( 'UTC' );
 			foreach ( $contact['metadata'] as $key => $value ) {
-				if ( \is_string( $value ) && \preg_match( '/^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$/', $value ) ) {
+				if ( is_string( $value ) && false !== \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value ) ) {
 					$date = \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value, $utc_timezone );
 					$date->setTimezone( $timezone );
 					$contact['metadata'][ $key ] = $date->format( Metadata::DATE_FORMAT );
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -109,23 +109,12 @@ class ESP_Sync extends Sync {
 		 */
 		$contact = \apply_filters( 'newspack_esp_sync_contact', $contact, $context );
 
-		$contact = Sync\Metadata::normalize_contact_data( $contact );
-
-		// Go over all values in the contact's metadata and if they look like dates, then convert them from UTC to the site's timezone.
-		if ( ! empty( $contact['metadata'] ) ) {
-			$timezone     = \wp_timezone();
-			$utc_timezone = new \DateTimeZone( 'UTC' );
-			foreach ( $contact['metadata'] as $key => $value ) {
-				if ( empty( $value ) || ! is_string( $value ) ) {
-					continue;
-				}
-				$date = \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value, $utc_timezone );
-				if ( false !== $date ) {
-					$date->setTimezone( $timezone );
-					$contact['metadata'][ $key ] = $date->format( Metadata::DATE_FORMAT );
-				}
-			}
+		if ( ! empty( $contact['metadata']['registration_date'] ) ) {
+			// Convert date from UTC to the site's timezone.
+			$contact['metadata']['registration_date'] = get_date_from_gmt( $contact['metadata']['registration_date'] );
 		}
+
+		$contact = Sync\Metadata::normalize_contact_data( $contact );
 
 		$result = \Newspack_Newsletters_Contacts::upsert( $contact, $master_list_id, $context );
 
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -8,6 +8,7 @@
 namespace Newspack\Reader_Activation;
 
 use Newspack\Reader_Activation;
+use Newspack\Reader_Activation\Sync\Metadata;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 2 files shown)


includes/reader-activation/sync/class-esp-sync.php
```diff
@@ -110,6 +111,19 @@ public static function sync( $contact, $context = '' ) {
 
 		$contact = Sync\Metadata::normalize_contact_data( $contact );
 
+		// Go over all values in the contact's metadata and if they look like dates, then convert them from UTC to the site's timezone.
+		if ( ! empty( $contact['metadata'] ) ) {
+			$timezone     = \wp_timezone();
+			$utc_timezone = new \DateTimeZone( 'UTC' );
+			foreach ( $contact['metadata'] as $key => $value ) {
+				if ( \is_string( $value ) && \preg_match( '/^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$/', $value ) ) {
```
dkoo: Maybe this is from past JS-based trauma, but I always shy away from using regex if I can find an alternative solution with other native functions/methods. For this check, since we know the date format we're setting in the `Metadata` class, would it be safer to use something like this to check whether the string is a date? 

```suggestion
				if ( is_string( $value ) && false !== \DateTime::createFromFormat( Metadata::DATE_FORMAT, $value ) ) {
```

Also, this is purely subjective coding style and we're admittedly not super consistent about it throughout our codebase, but we tend to use the `\` global namespace only for non-native-PHP functions, such as global WP or Woo stuff, and classes that would otherwise throw a namespacing error (such as `DateTime` in this case). Since `is_string` is a [native function](https://www.php.net/manual/en/function.is-string.php), we'd just say `is_string`. Does it make sense? I'm not sure, but it's how most of us on @Automattic/newspack-product  do it. 😆

includes/reader-activation/sync/class-esp-sync.php
```diff
@@ -110,6 +111,19 @@ public static function sync( $contact, $context = '' ) {
 
 		$contact = Sync\Metadata::normalize_contact_data( $contact );
 
+		// Go over all values in the contact's metadata and if they look like dates, then convert them from UTC to the site's timezone.
+		if ( ! empty( $contact['metadata'] ) ) {
```
dkoo: We're already filtering all contact metadata through [this `normalize_contact_data` method](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/reader-activation/sync/class-metadata.php#L397) before we pass it to the ESP, which also applies a `newspack_esp_sync_normalize_contact` filter to the data to allow for greater control over the synced data, if needed. I think we should move the timezone handling to this method so that the filtered data includes this transformation, WDYT?

dkoo (changes requested):
@naxoc I like this change, and suspect most publishers will, too! I have a suggestion regarding where this function should go, though—let me know if you agree!

leogermani:
> Since we already store (and have stored) dates as UTC

What evidence we have for that? I'd like us to have a better understanding of what's going on with the data here.

For example, I can't link because it's a private repo, but if you look at the code for `WC_Subscription::get_date( $date_type, $timezone = 'gmt')`, it hints that the value is stored in the DB is in the site's timezone. Because it will only modify the date if `$timezone` is set to `'site'`...

All Woo functions that get the dates don't seem to do any transformations...

I'd love us to have a deeper understanding of what's going on before jumping to a quick fix... these things tend to be more complicated than they look like.

as for where whatever we do should live, I have a third opinion :) : I expect that the output of the `Newspack\Reader_Activation\Sync\WooCommerce::get_order_metadata` method to give me the values as they should be

naxoc:
Good call @leogermani  I guess we don't have any guarantees, but I did take a look at the `get_date()` usage and I can't find anywhere where it's called with a timezone, so it would default to the default arg ('gmt') in all the uses we have of it. So that one should be OK.

That said – we might have to take a more careful approach. We can identify fields we know are UTC and then only process those. I don't know if I can come up with that list, but a guess would be **all fields that have to with subscriptions** (because no calls with a timezone to `get_date()` ) and the **registration date** (checked that one), and probably **payment dates** – I checked next/last payment dates.

What do you think? Should we make a list of fields to process?

naxoc:
> I expect that the output of the `Newspack\Reader_Activation\Sync\WooCommerce::get_order_metadata` method to give me the values as they should be

Cool! I'll move the code there. Just to confirm – that code is only used for ESP sync right? Would tampering with the dates cause us problems in syncing to places that expect the date to be utc?

leogermani:
> Good call @leogermani I guess we don't have any guarantees, but I did take a look at the `get_date()` usage and I can't find anywhere where it's called with a timezone, so it would default to the default arg ('gmt') in all the uses we have of it. So that one should be OK.
> 
> That said – we might have to take a more careful approach. We can identify fields we know are UTC and then only process those. I don't know if I can come up with that list, but a guess would be **all fields that have to with subscriptions** (because no calls with a timezone to `get_date()` ) and the **registration date** (checked that one), and probably **payment dates** – I checked next/last payment dates.
> 
> What do you think? Should we make a list of fields to process?

I was quickly having a look at some other methods, like the `WC_DateTime` class, which should take care of everything related to dates in Woo... and it looks like it should already return the dates in the site's timezone, if we call the correct methods... 

The reason I'm asking questions here is because it seems that Woo should already be handling this... maybe we are not getting the dates in the way we should be doing

naxoc:
I'm not sure if you mean that we should be storing dates differently or that we can use other methods converting them @leogermani ? Do you mean on the way out?

I took a look at `WC_DateTime`. It wraps PHP's `DateTime` but it's completely up to the code calling it what timezone will come out of that. I'm kinda hoping that the system saves things in UTC for sanity sake, but because of the format dates are stored in who knows? 🤷‍♀️ 

Here is how I tested my assumptions that dates are stored as UTC:
1. set the timezone on the site Australia or something far away from your timezone
2. Buy a subscription with a new user
3. `wp post meta list <new subscription post id>` and verify that the dates there are UTC.
4. `wp user get <new user ID>` and check that `user_registered` is UTC.
5. Then before running ` wp newspack esp sync` put `error_log( sprintf( '%s: %s before: %s after: %s', $contact['name'], $key ,$value, $contact['metadata'][ $key ] ) );` in the `foreach` that this PR adds and check that all before-dates there are UTC. 

All that said - maybe the safest is still to target dates we know are UTC.

leogermani:
Yes, it looks like dates are stored in UTC in the database - and it makes sense.

But I can see that all the `get_date_*()` calls done to WC_Orders, will return the date already transformed to the site's timezone. For example [this](https://github.com/Automattic/newspack-plugin/blob/954e01b792d425f670f415fc4d3142e1b4bcb8ba/includes/reader-activation/sync/class-woocommerce.php#L257).

However, the `get_date()` method of the `WC_Subscription` object returns the dates in UTC by default. 

So it seems to me that all we need to do is to add a second parameter `site` to these calls [here](https://github.com/Automattic/newspack-plugin/blob/954e01b792d425f670f415fc4d3142e1b4bcb8ba/includes/reader-activation/sync/class-woocommerce.php#L289-L298)

And that's all we need to do. Isn't it?

The only thing I'm not sure about is the user registration date. The `get_date_created()` method called [here](https://github.com/Automattic/newspack-plugin/blob/954e01b792d425f670f415fc4d3142e1b4bcb8ba/includes/reader-activation/sync/class-woocommerce.php#L350) returns a `WC_Datetime` object saying the timezone is correct, but the date is the exact same date that is stored in the DB...

```
wp> $customer->get_date_created();
=> object(WC_DateTime)#10958 (4) {
  ["utc_offset":protected]=>
  int(0)
  ["date"]=>
  string(26) "2025-01-10 18:45:16.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(17) "America/Sao_Paulo"
}

...
mysql> select user_registered from wp_users where ID = 1;
+---------------------+
| user_registered     |
+---------------------+
| 2025-01-10 18:45:16 |
+---------------------+
```

naxoc:
Thanks @leogermani - this is much simpler. If we are confident those dates will not be used in other contexts where the timezone is expected to be in UTC, then I agree that is the right place to edit them. As for the registration date – I edited that still in the sync class because I'm unsure where is best. Suggestions welcome :)

includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -231,11 +233,6 @@ private static function get_order_metadata( $order, $payment_page_url = false )
 			return [];
 		}
 
-		$is_subscription = false;
```
naxoc, line 234: This variable was unused, so I deleted this.

includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -322,16 +319,16 @@ private static function get_order_metadata( $order, $payment_page_url = false )
...
 
-			$metadata['sub_start_date']      = $current_subscription->get_date( 'start' );
-			$metadata['sub_end_date']        = $current_subscription->get_date( 'end' ) ? $current_subscription->get_date( 'end' ) : '';
+			$metadata['sub_start_date']      = $current_subscription->get_date( 'start', 'site' );
+			$metadata['sub_end_date']        = $current_subscription->get_date( 'end', 'site' ) ?? '';
 			$metadata['billing_cycle']       = $current_subscription->get_billing_period();
 			$metadata['recurring_payment']   = $current_subscription->get_total();
 			$metadata['last_payment_amount'] = self::get_last_payment_amount( $current_subscription );
-			$metadata['last_payment_date']   = $current_subscription->get_date( 'last_order_date_paid' ) ? $current_subscription->get_date( 'last_order_date_paid' ) : gmdate( Metadata::DATE_FORMAT );
+			$metadata['last_payment_date']   = $current_subscription->get_date( 'last_order_date_paid', 'site' ) ?? current_time( Metadata::DATE_FORMAT );
```
naxoc: Not quite sure I understand why we'd use `gmdate()` or now `current_time()` if it it's empty (it could just be empty), but I'm just replicating the existing code.

includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -322,16 +319,16 @@ private static function get_order_metadata( $order, $payment_page_url = false )
 				$metadata['membership_status'] = $current_subscription->get_status();
 			}
 
-			$metadata['sub_start_date']      = $current_subscription->get_date( 'start' );
-			$metadata['sub_end_date']        = $current_subscription->get_date( 'end' ) ? $current_subscription->get_date( 'end' ) : '';
+			$metadata['sub_start_date']      = $current_subscription->get_date( 'start', 'site' );
```
leogermani: This is not quite the same thing

`??` looks specifically for `NULL`. So if the date returns false or an empty string the metadata would be set to false or an empty string

```
wp> '' ?? 'no';
=> string(0) ""
wp> null ?? 'no';
=> string(2) "no"
wp> '' ? '' : 'no';
=> string(2) "no"
wp> false ?? 'no';
=> bool(false)
```

includes/reader-activation/sync/class-esp-sync.php
```diff
@@ -108,6 +109,11 @@ public static function sync( $contact, $context = '' ) {
 		 */
 		$contact = \apply_filters( 'newspack_esp_sync_contact', $contact, $context );
 
+		if ( ! empty( $contact['metadata']['registration_date'] ) ) {
```
leogermani: I think it's fine if we did this transformation where this is first created [here](https://github.com/Automattic/newspack-plugin/blob/954e01b792d425f670f415fc4d3142e1b4bcb8ba/includes/reader-activation/sync/class-woocommerce.php#L350), so it's together with all the others and we assure that that method alaready output all the dates in the right timezone

includes/reader-activation/sync/class-woocommerce.php
naxoc: You are right. My bad
I changed it to use empty for readability. Sadly WordPress does not allow the Elvis operator.

includes/reader-activation/sync/class-woocommerce.php
```diff
@@ -382,10 +383,9 @@ public static function get_contact_from_customer( $customer, $payment_page_url =
 		$metadata = [];
 
 		$customer_id                   = $customer->get_id();
-		$user                          = \get_user_by( 'id', $customer_id );
```
naxoc, line 385: This line was unused – I deleted it.

leogermani (approved):
Working as expected!

Thanks for going through all the revisions! And apologies for being so opinionated, but I hope you agree this looks good!

PR #3746, opened 2025-02-12
feat: custom byline interface

allysonsouza:
### All Submissions:

This Pull Request implements the feature to allow the selection of author tokens from authors selected on Co-Authors Plus to compound a custom byline with them.

1. Go to a post edit page and assign authors to the post on co-authors plus panel
2. Below it, you will see the Newspack Custom Byline panel. Set the toggle to active to enable custom byline.
3. You will see the custom byline field, which is a `div` with `contenteditable`. Below it you should see the authors assigned to the post as buttons that will insert the tokens on the content when clicked.
4. Click on buttons to add the token to the content
5. Click on buttons inside the byline content to remove token
6. Also, the tokens should be removed when editions are made to then with delete/backspace for example
7. Save the post, reload the page and see if the custom byline saved is what is expected
8. Check if the custom byline is editable as expected too after saving it

**Things to check:**

- The authors set on Co-Authors Plus are being displayed as tokens to be used on Custom Byline?
- When I click on token "+", the author token is inserted at the end of the byline?
- Does the token get removed after a change on it's content is made (by using delete, backspace or by adding additional characters)?
- Does the token get removed when I click on "x" on it after inserted into the byline?
- Does additional space after the token is inserted to allow more text to be typed after it?
- Does the byline get save properly?
- After saving, does the available tokens (the ones that aren't in use) are correctly displayed below the byline field?

### Other information:


claudiulodro (dismissed):
Overall I think this is coming together! The UX of it feels good. I noticed a few quirks in my testing:

1. In general it's kind of hard to read once there are a few authors in there:

Screenshot 2025-02-13 at 1 29 49 PM

Not sure if we should reduce the size of the author boxes or how best to handle that. What do you think, @thomasguillot?

2. Using the arrow keys, you can go into one of the boxes and then type to modify the content. I think it should not be editable.

Screenshot 2025-02-13 at 1 31 01 PM

3. If you select a portion of the text that includes a box and delete, the box doesn't go back into the available authors section:

Screenshot 2025-02-13 at 1 33 24 PM

Screenshot 2025-02-13 at 1 33 34 PM

4. Authors are always inserted at the end of the text box. I think the best experience would be for them to be inserted at the cursor position, otherwise it's tricky to edit existing bylines to add/change authors around.

leogermani:
This is starting to look pretty good!

Here are two things I noticed:

* This feature should also work without CoAuthors Plus. In that case, it would list only the normal post author
* The input should be pre-filled with the default by byline
* The value stored in the post meta field should follow what's described [here](https://github.com/Automattic/newspack-plugin/tree/trunk/includes/bylines#readme) 
 
The idea of storing the ID and the name is to have a fallback in case the user is deleted. When we render the byline, we'll always read the ID and get the Display name from the database, but if the ID is gone, we'll use the name that is stored there.

leogermani:
Ah, and before you get crazy and go deep in a rabbit role, we DON'T NEED TO SUPPORT Guest Authors

thomasguillot:
> Overall I think this is coming together! The UX of it feels good. I noticed a few quirks in my testing:
> 
> 1. In general it's kind of hard to read once there are a few authors in there:
> 
> Screenshot 2025-02-13 at 1 29 49 PM
> Not sure if we should reduce the size of the author boxes or how best to handle that. What do you think, @thomasguillot?

You're right, this doesn't look good. Let me think about a different approach here.

thomasguillot:
Note: The panel should simply be named "Byline", not "Newspack Custom Byline"

thomasguillot:
Can we please make sure the style matches Core's [FormTokenField](https://wordpress.github.io/gutenberg/?path=/docs/components-formtokenfield--docs)

Also we can use `@wordpress/base-styles/colors` rather than hardcoding the values.

miguelpeixe:
I've just tested on a fresh post and I still have issues with the backspace interaction:

https://github.com/user-attachments/assets/9fa6593f-ded0-4c43-9de6-4f8b7e9eb816

Another problem I faced while drafting the new post was that the additional authors only became available after I saved and refreshed the post. Ideally, the component would be aware of the edited changes and update the component's available tokens.

allysonsouza:
The latest commit fixes the backspace interaction, @miguelpeixe. About authors, it was expected that for the first version that would be acceptable (to have to refresh the page to make them available to use on byline), can you confirm that, @claudiulodro?

miguelpeixe:
Thanks for the revisions @allysonsouza!

Backspace interaction is working well for deleting the token when the cursor is at the end of it. I'm facing a couple of issues still:

 - The cursor repositions to the beginning after deleting a token
 - Removing (via delete or backspace) multiple selected text doesn't restore the token

https://github.com/user-attachments/assets/2422113b-59d1-4215-b371-46a468454e72

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+			<button
+				class="components-button components-form-token-field__remove-token token-inline-block__remove"
+				type="button"
+				data-token="$1"
+			>
+				${ close }
+			</button>
+		</span>`;
+
+		// For backwards compatibility, replace the '<Author>' tags, which were used before.
```
leogermani: no need for BC. This is brand new

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+const CustomBylineModal = ( { children } ) => {
+	const [ isOpen, setOpen ] = useState( false );
+	const openModal = () => setOpen( true );
+	const closeModal = () => setOpen( false );
+
+	return (
+		<>
+			<Button variant="secondary" onClick={ openModal }>
+				Set Custom Byline
+			</Button>
```
goldenapples, line 139: It seems like there should be a way of seeing the byline without clicking to open the modal. Does it make sense to show the byline here rather than just the button? Is there a design for this?

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+			dangerouslySetInnerHTML={ { __html: byline.current } }
+		/>
+	);
+};
+
+const TokenInlineBlock = ( { token, onInsert } ) => {
+	return (
+		<>
+			<span
+				className="components-form-token-field__token token-inline-block"
```
goldenapples: This probably needs `contenteditable="false"` since we're not trying to allow editing text within this span.

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+		getEditedPostAttribute( 'meta' )[ newspackBylines.metaKeyByline ] || '';
+
+	/**
+	 * Parse byline meta to convert custom tags (<Author></Author> or [Author][/Author]) to token markup.
+	 *
+	 * @see    {@link https://github.com/Automattic/newspack-plugin/tree/trunk/includes/bylines#readme|Custom Bylines}
+	 * @return {string} Parsed byline looking up for <Author id=1></Author> tags and replacing them.
+	 */
+	const bylineParser = () => {
+		const tokenMarkup = `<span id="token-$1" class="components-form-token-field__token token-inline-block author-token" data-token="$1">
```
goldenapples: This looks like the same component as `TokenInlineBlock` - can they be refactored to use the same function for generating markup?

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+	);
+
+	/** Fetch post author from core */
+	const { postAuthor } = useSelect( select => {
+		const { getUser, getUsers } = select( coreStore );
+		const _authorId = getEditedPostAttribute( 'author' );
+		const query = { ...AUTHORS_QUERY };
+
+		return {
+			authors: getUsers( query ),
```
goldenapples: This is being fetched but not returned; is it necessary?

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+				) {
+					// Remove token element.
+					bylineElement
+						.querySelector(
+							'span#token-' + event.target.dataset.token
+						)
+						.remove();
+
+					// Update byline meta.
+					updateBylineMeta( bylineElement );
```
goldenapples: Does this need to be called directly? I think removing the token from the field should trigger the mutation observer.

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
+
+	/**
+	 * Initialize on DOM ready
+	 *
+	 * Fill tokenInUse analyzing the byline element, and
+	 * add Mutation Observer to byline element after child element is ready.
+	 */
+	useEffect( () => {
+		// Wait for child component that will be analyzed to be ready.
+		if ( ! isTokensdReady || ! isBylineReady ) {
```
goldenapples: is this a typo "isTokensdReady"

src/bylines/style.scss
```diff
@@ -13,3 +13,47 @@
...
+	color: #1e1e1e;
+	cursor: pointer;
+	overflow: initial;
+	padding: 0 2px;
+	background: #ddd;
+	display: inline-block;
+	height: auto;
+	line-height: 24px;
+	min-width: unset;
+	transition: all 0.2 scubic-bezier (0.4, 1, 0.4, 1);
```
goldenapples: Potential typo here in the easing function?

src/bylines/index.js
```diff
@@ -3,64 +3,520 @@
...
-				} );
-			}
-			bylineEl.innerHTML = text;
+
+		// If Co-Authors Plus is active, use their authors
+		if ( newspackBylines.is_co_authors_plus_active ) {
+			const controller = new AbortController();
+
+			apiFetch( {
+				path: `/coauthors/v1/coauthors?post_id=${ postId }`,
```
goldenapples, line 439: Why can't we just use the edited coauthors attribute on the post?

src/bylines/index.js
allysonsouza: You're right, it was a reminiscence of a tentative or something, removed it

src/bylines/index.js
allysonsouza: Yes, it is, but I couldn't think about how to retrieve the render of a component to store on a variable, I guess that's not possible, but another solution could be possible, I'm trying to think what.

src/bylines/index.js
allysonsouza: You're right, they don't seen to be necessary there, the mutation observer will be already triggered and it will call it

src/bylines/index.js
allysonsouza: Nice suggestion, I didn't thought about that

allysonsouza:
After struggling to handle the state of the content, which was not being managed by React state, @goldenapples started working on a simplification to use `onInput` instead of MutationObserver, which lead to a simpler and shorter code.

He's proposing it on [this PR on our fork](https://github.com/humanmade/newspack-plugin/pull/1), I've been testing it and it seems like we need to deal with the removal of the tokens itself when there's a change to the token content (like adding characters or removing them with backspace or delete).

That seems possible to be handled by onInput too, by adding an onInput on each token to remove itself when changes happen and to keep the onInput on the wrapper contenteditable div.

leogermani:
Heads up that I've implemented the data structure changes that this PR does in trunk to allow us to move on with some tests. This might create conflicts that should be easily resolved - just ignore the changes on trunk

https://github.com/Automattic/newspack-plugin/pull/3863

miguelpeixe (approved):
Thank you for all the revisions, the code is well organized and very readable!

I ran into a couple of issues, but I'm not considering them blockers because we can address them separately:

1. When backspacing a token, my cursor gets reset to position zero
2. Making changes to the post authors does not update the available tokens. This is also the case after saving the changes. Only after a page refresh I'm able to use the new authors. This only happens when using Co-Authors Plus.

Other than that, I'm able to make changes to the byline and save with different token/text variations without issues.

PR #3750, opened 2025-02-13
fix(wcs): account for pagination and failed scheduled retries in on-hold migration script

chickenn00dle:
### All Submissions:

See https://a8c.slack.com/archives/C015W6BES8J/p1739466463068249?thread_ts=1738343578.900239&cid=C015W6BES8J

This PR fixes an issue where updating a subscription status throws off the results of `wcs_get_subscriptions` when using pagination.

This also adjusts the logic to expire a subscription that has failed to schedule a payment retry.

**Setup tips**:
    - **To setup a subscription with a failed parent order**, add the following filter `add_filter( 'newspack_subscriptions_expiration_enabled', '__return_false' );` then attempt to purchase a sub as reader using this stripe test card: `4000000000000002`
    - **To setup subscriptions that end up in the on-hold status after retries**, add the following filter `add_filter( 'newspack_subscriptions_expiration_enabled', '__return_false' );` then follow [these steps](https://woocommerce.com/document/subscriptions/develop/failed-payment-retry/#section-7).
    - **To setup a subscription with a next payment in the past**, use the following via wp shell on a currently on-hold subscription to manually set the start and/or next_payment dates:
```
$s = wcs_get_subscription(SUBSCRIPTION_ID);
$s->update_dates([ 'start' => 'YYYY-MM-DD HH:MM:SS', 'next_payment' => 'YYYY-MM-DD HH:MM:SS' ]);
```

**Testing steps**:
Note: Be sure to remove the above filter before running the script.

1. Change this value to something like `1`: https://github.com/Automattic/newspack-plugin/blob/535d8274dae721e2e1ce38689b344ee585e63da1/includes/cli/class-woocommerce-subscriptions.php#L248
2. On a test site ensure you have one of each of these kinds of on-hold subscriptions (see setup tips above):
  - with a failed parent order
  - with a next payment date set two months ago
  - that has completed all standard retries
3. I'm not sure how to force a failed scheduled retry, but you can change this line to something like `if ( true ) {` to simulate this case: https://github.com/Automattic/newspack-plugin/blob/4bd37d7dd4bf887cc376bae18479a8d441318a7b/includes/cli/class-woocommerce-subscriptions.php#L151
3. Run the script with the `--verbose --live` flag and confirm all on-hold subscriptions appear in the output
4. Check scheduled actions for a new `newspack_expire_manual_subscription` event for the subscription that has completed all standard retries (WooCommerce > Status > Scheduled Actions)

### Other information:

### DIFF
--- a/includes/cli/class-woocommerce-subscriptions.php
+++ b/includes/cli/class-woocommerce-subscriptions.php
@@ -163,7 +163,6 @@ class WooCommerce_Subscriptions {
 									$subscription->save();
 								}
 							}
-							++$scheduled;
 						} else {
 							// If there have been no retries, schedule expiration.
 							if ( self::$verbose ) {
@@ -174,8 +173,8 @@ class WooCommerce_Subscriptions {
 								$subscription->update_meta_data( '_newspack_cli_expiration_scheduled', true );
 								$subscription->save();
 							}
-							++$scheduled;
 						}
+						++$scheduled;
 					}
 				}
 				// Expire any subscriptinos that have passed the on-hold duration.
--- a/includes/cli/class-woocommerce-subscriptions.php
+++ b/includes/cli/class-woocommerce-subscriptions.php
@@ -151,17 +151,10 @@ class WooCommerce_Subscriptions {
 								remove_filter( 'wcs_is_scheduled_payment_attempt', '__return_true' );
 								if ( 0 === $subscription->get_date( 'payment_retry' ) ) {
 									if ( self::$verbose ) {
-										WP_CLI::line( 'Failed to schedule payment retry.' );
-									}
-									// There are cases on live sites where a renewed subscription status fails to transition back to active,
-									// so check to make sure the last order is not completed before flagging for expiration.
-									if ( ! $last_order || 'completed' !== $last_order->get_status() ) {
-										$should_expire = true;
-									} else {
-										WP_CLI::line( 'Subscription last order is completed. Moving to next subscription...' );
+										WP_CLI::line( 'Failed to schedule payment retry. Moving to next subscription...' );
 										WP_CLI::line( '' );
-										continue;
 									}
+									continue;
 								} else {
 									$subscription->add_order_note(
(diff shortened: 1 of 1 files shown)


includes/cli/class-woocommerce-subscriptions.php
```diff
@@ -228,11 +245,32 @@ private static function get_subscriptions( $page = 1 ) {
...
+	 * @return array
+	 */
+	private static function get_flagged_subscriptions() {
+		$subscriptions = wcs_get_subscriptions(
+			[
+				'subscriptions_per_page' => 50,
+				'subscription_status'    => 'on-hold',
+				'meta_query'             => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+					[
+						'key'     => '_newspack_cli_status_updated',
```
leogermani: Suggestion: Use a different meta key to mark subscriptions to be processed at the end of the run. Otherwise you might end up processing subscriptions marked in a previous run to the script.

then you should delete this meta after the sub is processed...

or... just store the subs that need to be processed in the memory. You can store only the ID and get the subscription at the end...

Makes sense?

chickenn00dle: > Suggestion: Use a different meta key to mark subscriptions to be processed at the end of the run. Otherwise you might end up processing subscriptions marked in a previous run to the script.

Sure. Although the `subscription_status` from previous runs would no longer be `on-hold` so I don't think it hurts to use this meta. But happy to separate this out if its cleaner.

> or... just store the subs that need to be processed in the memory. You can store only the ID and get the subscription at the end...

I was going to do this, but like Claudiu mentioned in the call, if a site has too many subscriptions to process, this can lead to memory issues.

chickenn00dle: In any case, I updated to use the `_newspack_cli_to_status` meta as the flag in https://github.com/Automattic/newspack-plugin/pull/3750/commits/9021ec6b9dde6e5ceaefd6294e1802794a07a460

leogermani (approved):
tested and added a few more logging lines

PR #3751, opened 2025-02-14
feat: add corrections customize settings

Takshil-Kunadia:
### All Submissions:

- Add a new Panel and global customizer setting “Corrections Location” (with options “Top” and “Bottom”).
- Fix a edge case in Correction Modal.

Closes # .

1. Open the WordPress Customizer and Load a single post in the theme.
2. Open Template Settings > Correction Settings > Corrections & Clarifications, and check the Radio buttons Top & Bottom to see the impact.

### Other information:

### DIFF
--- a/includes/corrections/class-corrections.php
+++ b/includes/corrections/class-corrections.php
@@ -448,7 +448,7 @@ class Corrections {
 	 * @return string the post content with corrections.
 	 */
 	public static function output_corrections_on_post( $content ) {
-		if ( is_admin() || ! is_single() ) {
+		if ( is_admin() || ! is_single() || wp_is_block_theme() ) {
 			return $content;
 		}
 
--- a/includes/corrections/class-corrections.php
+++ b/includes/corrections/class-corrections.php
@@ -52,6 +52,11 @@ class Corrections {
 	const REST_ROUTE = '/corrections';
 
 	/**
+	 * Customize settings for corrections.
+	 */
+	const CORRECTIONS_LOCATION_CUSTOMIZE_SETTING = 'corrections_location';
+
+	/**
 	 * Initializes the class.
 	 */
 	public static function init() {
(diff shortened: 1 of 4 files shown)


Takshil-Kunadia:
@leogermani 
Currently, if corrections are enabled and present, they are automatically appended to the post content. However, we are not providing users with an option to enable or disable corrections per post, as was originally intended by:

https://github.com/Automattic/newspack-plugin/blob/abd1bc2b4128415a0c0c26a861448de6d5f6b203/includes/corrections/class-corrections.php#L32

Would you like me to add a UI option to enable/disable corrections for individual posts?

Additionally, if we proceed with this implementation, corrections will be appended to all posts by default. This might not be ideal for users who may want to use blocks we are going to create via https://app.asana.com/0/1209292256643614/1209292256643625.

Would it make sense to introduce a separate setting in the modal for "Enable/Disable auto-inserting corrections" so users can choose whether corrections should be automatically appended? We could reuse(and rename it) CORRECTIONS_ACTIVE_META for this, but I wanted to check with you first—should we manage "enable/disable corrections" and "auto-insert corrections" separately, or should they share the same meta field?

leogermani:
> should we manage "enable/disable corrections" and "auto-insert corrections" separately, or should they share the same meta field?

As @laurelfulford  suggested [here](https://github.com/Automattic/newspack-theme/pull/2435#issuecomment-2666449736), let's only filter into `the_content` if the active theme is not a block theme

leogermani (approved):

naxoc (approved):
This looks good 👍  I tested "Open Template Settings > Correction Settings > Corrections & Clarifications" and it works great. It's not under Template Settings, but directly under "Corrections Settings", but I also think that makes more sense.

Takshil-Kunadia:
Thanks! @leogermani @naxoc for taking a look 🙇 . Merging this PR 🚀

PR #2923, opened 2024-02-13
fix(woocommerce): fix order-pay page when using woocommerce-membershi…

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes an issue with the `woocommerce-memberships-for-teams` plugin when attempting to complete a pending-payment order. 

See 1200550061930446-as-1206314181096660

### How to test the changes in this Pull Request:

1. Activate `woocommerce-memberships-for-teams` on top of regular Newspack & Woo setup
2. Edit the monthly donation product and tick "Team membership" checkbox:

image

3. Create a membership plan with access based on buying a recurring donation product
1. Make a monthly donation as a new user, then as admin manually update the order status to "Pending payment"
1. Visit the "Customer payment page" (when logged in as the customer)
4. On `release` -> observe the cart is emptied and you cannot pay for the order
6. Switch to this branch, visit the payment page link again, observe a regular Woo cart experience*

\* you won't be able to actually pay because of the manual setting of 'pending payment' status

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3752, opened 2025-02-14
fix(woocommerce): add team name to checkouts for memberships-for-teams

naxoc:
### All Submissions:

This adds some filtering to team name on checkout. Some checkout situations were causing errors. See https://app.asana.com/0/1207572908840041/1209019125448166

Also see #2923 - this is related and some of the code from that was changed to accommodate more scenarios. In testing I found a bunch more scenarios where things break because of the missing team name. The ones I found are described in the testing instructions

#### Questions
I made some utility functions because we already have custom code that does similar things and the team name repair thing is so widespread.
Should we move the new functions to a specific woo-memberships-for-teams class? 

Also see instructions in https://app.asana.com/0/1207572908840041/1209019125448166

* you'll need woo subscriptions, memberships, and teams for memberships.
* Set up a membership plan and a subscriptions:
    * variable subscription
    * set up some variations
    * mark it as a "Team subscription"

* Add the subscription product to a checkout button in a page. Visit as a reader and buy it.
* Confirm in admin the subscription is created, and a team is created for that user.
* As an admin, cancel the subscription.

  
As that reader,

1. go to My Account > Subscriptions and try to click "Resubscribe". Without this patch you'd get a ""Field 'Group Name' is required." error - verify you don't get the error.
2. On your account page -> My Subscriptions click "Upgrade or Downgrade" and verify that the team name is filled out (I'm unsure if we should be displaying any of these fields though 🤔 Should we hide them?).
3. On the accounts page –> My Subscriptions click "Renew now" on the subscription and verify that you are taken to checkout and not to an empty cart.
4. To check for regressions, also go through the test instructions in #2923

  
  ### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-connection.php
@@ -699,15 +699,21 @@ class WooCommerce_Connection {
 	 */
 	public static function wc_memberships_for_teams_product_team_user_input_fields( $fields ) {
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( ! empty( $_REQUEST['team_name'] ) ) {
+		if ( empty( $fields['team_name'] ) || ! empty( $_REQUEST['team_name'] ) ) {
 			return $fields;
 		}
 		global $wp;
-		if ( empty( $fields['team_name'] ) || ! isset( $wp->query_vars['order-pay'] ) ) {
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		if ( isset( $wp->query_vars['order-pay'] ) ) {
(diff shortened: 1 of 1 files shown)


leogermani:
Didn't work for me :/ the "resubscribe" flow still fails the same way.

Maybe I can send you a dump of my test site

naxoc:
@leogermani woops. My bad. I got the scenarios switched around. Could you try again please?

leogermani (approved):
Yes! Not all the scenarios work.

Not sure I was able to test Adam's PR beacuse I don't know what he means by "Customer payment page".... 

I agree with your suggestion to group and isolate these fixes for Teams. We have a `plugins` folder where we put things for third party plugins... but not sure if this one could be one file inside the reder-revenue/woocommerce folder... not a strong opinion

PR #3768, opened 2025-02-20
fix(memberships): ensure user membership is linked to the correct subscription

dkoo:
### All Submissions:

Yet another attempt to fix the WooCommerce Memberships bug addressed in #3050, #3060, #3268, #3338, #3380, and #3393. This bug exists in the [WooCommerce Memberships plugin](https://github.com/woocommerce/woocommerce-memberships/pull/115) itself, but there's no indication that the issue will be addressed upstream, so we're applying a patch here as it continues to affect live sites.

The crux of the issue: if a user purchases more than one subscription for a product required by a user membership plan, and the first subscription has an **Expired** status, their user membership will remain linked to the expired subscription. Even if they subsequently purchase a new subscription for the same product, the membership will remain linked to the expired subscription, and will remain in expired status—so the user won't be able to access restricted content despite having purchased the required subscription.

This PR addresses the issue by checking the user's memberships upon login and subscription status change. If a user membership requires a subscription to specific products and the user has active subscriptions for those products, the membership will be relinked to the most recent active subscription, and the expiration date will be cleared. This should trigger the membership to automatically reactivate, granting the user access to any restricted content defined in the membership plan.

1. Publish a membership plan that requires a subscription product and is tied the subscription length. Optionally add content restrictions to test user access to the restricted content.

Screenshot 2025-02-20 at 4 45 57 PM

2. On `release`, as a reader, purchase a subscription with the required product.
3. As an admin, change the subscription's status to `expired`.
4. As the reader, purchase a second subscription with the required product.
5. As an admin, find the user's membership (in **WooCommerce > Memberships**). 
6. Observe that the user membership remains linked to the initial expired subscription, and remains in "Expired" status, meaning that the reader won't be able to access any content restricted by the membership plan.
7. Check out this branch.
8. As the reader, log out and then log back into the site again to trigger the `reader_logged_in` data event.
9. Confirm that the user membership has been relinked to the active subscription, that the membership now has an "Active" status with no expiration date, and that the reader can now access restricted content.

Screenshot 2025-02-20 at 4 56 01 PM

10. As an admin, set the active subscription to `expired` status and confirm that the membership also changes to expired status again.
11. As the reader, purchase a subscription for the required product again. Confirm that the user membership has been once again relinked to the new active subscription, that the membership now has an "Active" status with no expiration date, and that the reader can now access restricted content.

### Other information:

---
  - https://app.asana.com/0/0/1210190746201040
  - https://app.asana.com/0/0/1209414048559497

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -10,6 +10,7 @@ namespace Newspack;
 use Newspack\Data_Events;
 use Newspack\Logger;
 use Newspack\Memberships\Metering;
+use Newspack\Reader_Activation;
 use Newspack\WooCommerce_Connection;
 
 defined( 'ABSPATH' ) || exit;
@@ -41,7 +42,6 @@ class Memberships {
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
-		add_action( 'init', [ __CLASS__, 'register_data_event_handlers' ] );
 		add_action( 'init', [ __CLASS__, 'register_post_type' ] );
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
 		add_action( 'admin_init', [ __CLASS__, 'redirect_cpt' ] );
@@ -75,6 +75,10 @@ class Memberships {
 		add_filter( 'newspack_gate_content', 'wp_replace_insecure_home_url' );
 		add_filter( 'newspack_gate_content', 'do_shortcode', 11 ); // AFTER wpautop().
 
+		/** Fixes to ensure that memberships are linked to the correct active subscription. */
+		add_action( 'woocommerce_subscription_status_updated', [ __CLASS__, 'check_user_memberships_on_subscription_update' ] );
+		add_action( 'wp_login', [ __CLASS__, 'check_user_memberships_on_login' ], 10, 2 );
+
 		include __DIR__ . '/class-block-patterns.php';
 		include __DIR__ . '/class-metering.php';
 		include __DIR__ . '/class-import-export.php';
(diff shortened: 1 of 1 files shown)


dkoo:
As discussed with @miguelpeixe, #3769 gets us close to a fix but doesn't solve the issue for user memberships linked to expired subscriptions.

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -1086,5 +1087,101 @@ public static function handle_wc_memberships_expire_user_membership( $expire, $u
...
+		$updated = false;
+		if ( ! self::is_active() ) {
+			return $updated;
+		}
+		$user_memberships = wc_memberships_get_user_memberships( $user_id );
+		foreach ( $user_memberships as $user_membership ) {
+			$membership_plan_id      = $user_membership->get_plan_id();
+			$user_membership_id      = $user_membership->get_id();
+			$subscription_membership = new \WC_Memberships_Integration_Subscriptions_User_Membership( $user_membership_id );
+			$active_subscription_id     = self::get_user_subscription_for_membership_plan( $user_id, $subscription_membership->get_plan_id() );
```
miguelpeixe: NIT

```suggestion
			$active_subscription_id  = self::get_user_subscription_for_membership_plan( $user_id, $subscription_membership->get_plan_id() );
```

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -72,6 +75,10 @@ public static function init() {
 		add_filter( 'newspack_gate_content', 'wp_replace_insecure_home_url' );
 		add_filter( 'newspack_gate_content', 'do_shortcode', 11 ); // AFTER wpautop().
 
+		/** Fixes to ensure that memberships are linked to the correct active subscription. */
+		add_action( 'woocommerce_subscription_status_updated', [ __CLASS__, 'check_user_memberships_on_subscription_update' ] );
```
miguelpeixe:
```suggestion
		add_action( 'woocommerce_subscription_status_updated', [ __CLASS__, 'check_user_memberships_on_subscription_update' ], 11 );
```

This should run after the Memberships hook, otherwise resetting the end date may not stick

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -1086,5 +1087,101 @@ public static function handle_wc_memberships_expire_user_membership( $expire, $u
...
+
+				// If the user membership is linked to the wrong subscription, relink it.
+				if ( $linked_subscription_id !== $active_subscription_id ) {
+					$updated         = $subscription_membership->set_subscription_id( $active_subscription_id );
+					$membership_plan = $subscription_membership->get_plan();
+
+					// Reset end date for plans with access set to subscription length.
+					if ( $membership_plan->is_access_length_type( 'subscription' ) && ! empty( $subscription_membership->get_end_date() ) ) {
+						$subscription_membership->set_end_date( '' );
+					}
```
miguelpeixe: With the hook priority at the default (10), the `set_subscription_id()` works and the membership is relinked, but `set_end_date()` doesn't persist and the membership remains with the expired status.

After changing the hook priority to 11 the end date resets, but the status remains "expired". Maybe we should be running a `$subscription_membership->update_status( 'active', $note );` here too?

miguelpeixe (changes requested):
I ran into a few issues that I commented inline.

Do we still need the `handle_wc_memberships_expire_user_membership()` logic here?

includes/plugins/wc-memberships/class-memberships.php
dkoo: Thanks for catching this! I can confirm this is the case. Maybe the data events approach was working because it happened after all other WC Memberships hooks had done their thing

dkoo: f436f1a is now working for me. As for whether we still need `handle_wc_memberships_expire_user_membership()`, I repurposed it as another trigger to check and possibly relink the membership, which should prevent the "Expired" status from being set if the membership is linked to the wrong subscription. WDYT?

dkoo: 3fad077 updates `handle_wc_memberships_expire_user_membership()` to `check_user_memberships_on_expire()` and groups it with the other `check_user_memberships_on_*` methods, for clarity. It also only prevents the expiration if that specific membership should be relinked to an active subscription.

dkoo:
I'm not sure why the unit test job is failing now—tests run fine locally 😕

miguelpeixe:
> I'm not sure why the unit test job is failing now

Either CI or .org is unstable and unable to download WP. It's been doing that for me all day.

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -7,7 +7,10 @@
 
 namespace Newspack;
 
+use Newspack\Data_Events;
```
miguelpeixe: This is unused now

miguelpeixe (approved):
> I repurposed it as another trigger to check and possibly relink the membership, which should prevent the "Expired" status from being set if the membership is linked to the wrong subscription. WDYT?

Good call!

It's working well for me now! 🎉

includes/plugins/wc-memberships/class-memberships.php
dkoo: Thanks, removed in 835649b5!

dkoo:
Holding this until a non-Friday

PR #3770, opened 2025-02-21
feat: add basic local storybook implementation for newspack-ui

thomasguillot:
### All Submissions:

- Introduces a new Storybook setup for Newspack UI components
  - Stories are stored inside a dedicated `/storybook` directory at the root level
  - Configures Storybook to use our CSS-only components
  - Adds comprehensive documentation for each component

- Creates initial component stories:
  1. Badge
     - Documents all variants (primary, secondary, outline, success, error, warning)

  2. Button
     - Documents all variants (primary, secondary, branded, ghost, outline, destructive)
     - Shows size options (x-small, small, medium)
     - Demonstrates layout modifiers (wide, icon, loading)
  
  3. Segmented Control
     - Shows basic implementation with tabs and panels
     - Documents size options and wide layout

- Change the default size of the icon-only button to be `medium`, just like a regular button.
     - It means I had to update Reader Activation modals and elements
     - I also added the inner spans (glyph and text) for buttons using icons

Note: My goal is to ultimately have a dedicated Github page.

1. Install and run Storybook:
   ```bash
   npm install
   npm run storybook
   ```
   Storybook should open at http://localhost:6006

2. In Storybook's sidebar, verify:
   - Components are listed under "Newspack UI"
   - Badge, Button and Segmented Control components are present

3. Test Badge component:
   - Check all variants render correctly (Primary, Secondary, etc.)
  
4. Test Button component:
   - Check all variants render correctly (Primary, Secondary, etc.)
   - Try different sizes using the controls panel
   - Test layout options (wide, icon-only)
   - Verify loading state works
   - Ensure all interactive examples work as expected

5. Test Segmented Control component:
   - Verify tabs switch content correctly
   - Test different sizes using the controls panel
   - Check the wide layout option
   - Ensure selected states are visually correct

6. Verify documentation:
   - Code examples should update when changing controls
   - All modifiers should be properly documented
   - CSS classes should match what's being used

7. Check if the Reader Activation modals and elements are sized correctly
   - Mainly buttons that are using icon + text or icon-only

### Other information:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,4 +7,13 @@ module.exports = {
 		newspack_aux_data: 'readonly',
 	},
 	ignorePatterns: [ '*/dist/', '*/node_modules/', '*/release' ],
+	overrides: [
+		{
+			files: [ '.storybook/**/*.js' ],
+			rules: {
+				'import/no-unresolved': 'off',
+				'import/no-extraneous-dependencies': 'off'
+			}
+		}
+	]
 };
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,0 +1,31 @@
+const config = {
+  stories: [
+    '../storybook/**/*.mdx',
+    '../storybook/**/*.stories.@(js|jsx|mjs|ts|tsx)',
+    '../src/**/*.mdx',
+    '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)',
+  ],
+  addons: [
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
+    '@storybook/addon-interactions',
+  ],
+  framework: {
+    name: '@storybook/react-webpack5',
+    options: {},
+  },
+  docs: {
+    autodocs: 'tag',
+  },
+  webpackFinal: async (config) => {
+    // Add SASS support
+    config.module.rules.push({
+      test: /\.scss$/,
+      use: ['style-loader', 'css-loader', 'sass-loader'],
+    });
+
+    return config;
+  },
+};
+
+export default config;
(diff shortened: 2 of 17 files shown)


thomasguillot:
@laurelfulford Why was the default size for icon-only buttons set to x-small, while regular buttons default to medium? I’ve updated this behaviour so that all buttons now default to medium.

dkoo:
When I run `npm run storybook` I get an error message:

Screenshot 2025-02-25 at 11 55 32 AM

adekbadek (changes requested):
I'm a fan of Storybook, it does a great job at showcasing a UI library, and I think it would be a great addition to our developer/designer docs. With regards to the implementation in the PR:

1. The npm packages should not be a part of the root `package.json`. Potential for dependency conflicts and other npm issue grows with the number of packages, so I think there should be a separation of concerns here. A solution would be to have a separate directory for storybook, with its own `package.json`.
2. There is now a bit of mess: 
    1. There's a `Newspack_UI::return_demo_content` method, which duplicates Storybook effort → Storybook should rather replace that, unless the two should live alongside for some reason.
    2. There's a lot of duplication due to the fact that the UI library is essentially a CSS library. The markup is created separately in PHP (twice: for the `return_demo_content` and for actual components) and in Storybook. From what I see, it seems that this UI library is to be used on the PHP-generated front-end. If that's the case, the HTML should be generated by PHP (in a single place).
    3. The "Segmented Control" component looks very familiar to the Donate Block's "modern" style. What's the plan here?

image
 


(I did not encounter the issue @dkoo mentioned in the other review, worked fine for me)

thomasguillot:
Closing this PR. This is outdated and it can be implemented in a better way.

PR #3778, opened 2025-02-24
feat(my-account): add email change cancellation option

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208993180326452/1209468700585359/f

This PR adds the option to cancel an email change request after one has been submitted:

Screenshot 2025-02-24 at 12 42 38

1. Ensure the `NEWSPACK_EMAIL_CHANGE_ENABLED` FF is added in wp-config
2. As admin, go to Newspack > Engagement > Show Advanced Settings > Transactional Email Content and reset the `Change email` template
3. Click edit next to this template and verify the last line in the main body of the email template contains the `EMAIL_CANCELLATION_URL` template tag
4. As a logged in reader, go to my account and enter a new email address, then submit
5. Confirm the page reloads and a `Cancel email change` button appears next to the `Save changes` button. 
6. Click this cancel button and verify the page reloads with a notice indicating the email change request was cancelled and the email field has reverted to the old email address
7. Repeat step 4 and this time check the readers new email for the email change verification email
8. Verify the email contains a link to cancel the email change request
9. Click the link and verify the request is cancelled as in step 6

### Other information:


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -877,6 +907,27 @@ public static function handle_verify_email_change() {
 				);
 				\wc_add_notice( __( 'Your email address has been successfully updated.', 'newspack-plugin' ) );
 			}
+		} else {
+			\wc_add_notice( $error, 'error' );
+		}
+		\wp_safe_redirect( \wc_get_endpoint_url( 'edit-account', '', \wc_get_page_permalink( 'myaccount' ) ) );
```
dkoo: I stumbled on this by accident, but if you click on the link while not logged in, it brings you to the site's homepage and doesn't seem to do anything, but the nonce/link then gets invalidated, so after logging in I can no longer click the cancellation link to cancel the request (I just get the "something went wrong" error message).

What if we tweaked this flow to check logged-in state before verifying the nonce? If not logged in, we should redirect to the My Account page so they user gets presented with a login screen, then after logging in, we can proceed with the nonce verification and cancellation logic.

dkoo: This is also the case for confirming an email change. I think the same flow during logged-out state would work for that, too.

dkoo (changes requested):
This is mostly working as described, except for one flow that I detailed below (when clicking a confirm or cancel link while not logged into the reader's account).

includes/reader-revenue/my-account/class-woocommerce-my-account.php
chickenn00dle: Oooh. Good catch @dkoo! We actually do perform the `is_user_logged_in` check before verifying the nonce in both [handle_verify_email_change](https://github.com/Automattic/newspack-plugin/blob/25754c17ee6a1ee8b685a5f20aef013340a84b00/includes/reader-revenue/my-account/class-woocommerce-my-account.php#L888) and [handle_cancel_email_change](https://github.com/Automattic/newspack-plugin/blob/25754c17ee6a1ee8b685a5f20aef013340a84b00/includes/reader-revenue/my-account/class-woocommerce-my-account.php#L921). 

But this issue will still happen since the nonce is tied to the session. So if the url is generated while the reader is logged in, then the reader logs out and logs back in, and finally the reader clicks the link, it will fail. 

To get around this case, I've switched from using nonces to `wp_hash` in https://github.com/Automattic/newspack-plugin/pull/3778/commits/cc4111edf6252b77b47a5d67d7c54449da7b2e34. This commit also makes it so logged out readers are directed to my account when clicking these links as well.

Let me know if this works!

chickenn00dle: Actually, I think I'm going to have to take this one out of review since doing away with the nonces opens us up to the issue you pointed out in slack here: https://a8c.slack.com/archives/C055SRT7WN4/p1740519700873229

I'll have to do some more thinking about this later this week. (I'm open to suggestions though!) I'll ping you once this is ready again. Sorry for the confusion @dkoo!

dkoo: Thanks, @chickenn00dle! If you want to separate out the refactoring of the nonce/logged-out state into another PR, we can proceed with this one as-is. Whatever you prefer!

chickenn00dle: Alright, I've made a small change to always use a hash of the existing email address, this way only the logged in user with that email can complete the process. I think this should cover us, but let me know what you think @dkoo!

dkoo: Nice! This is working well with the email hash, but I think we just need one more change for the handling in a logged-out state. Using the following steps:

1. From My Account, change your email address to get the verification email
2. Log out of the reader account
3. Click either the Confirm or Cancel links from the email: you'll get directed to log in at My Account, with the `?verify-email-change=token` or `?cancel-email-change=token` params still attached
4. Log into the account using this form and click "Continue" after logging in

At this point, the page reloads and you're logged into the account, but in the reload we lose the params on the URL, so the confirm/cancel request isn't handled. As the reader, I'll need to go back to the email and click the link again. 

I think we should let these params pass through after the login so that the reader won't have to take that extra step to complete the intended action. Maybe we can detect the params on the referrer URL on a `wp_login` hook to handle them?

dkoo (changes requested):
@chickenn00dle we're almost there! I just have one more suggestion below before merging this. Let me know if you want to discuss further.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
chickenn00dle: Great idea! Unfortunately the auth form hijacks the wp_login flow, so I had to make changes to the redirect flow there in https://github.com/Automattic/newspack-plugin/pull/3778/commits/ae1217f2c2a6d0a9a125cc96bbeb11a59050a4a7

Let me know if this looks good to you!

dkoo: Nice, this accomplishes the task of making the flow more seamless for the reader, so I'm good with it!

dkoo (approved):

PR #3780, opened 2025-02-24
feat(my-account): sync email change with esp

chickenn00dle:
### All Submissions:

Part of https://app.asana.com/0/1208993180326452/1209215513701719

This and the accompanying newsletters PR (https://github.com/Automattic/newspack-newsletters/pull/1764) trigger email change updates in the connected ESP after an email change is verified.

You will need to do three rounds of testing, one for MailChimp, one for Active Campaign, and one for Constant Contact

1. Make sure the `NEWSPACK_EMAIL_CHANGE_ENABLED` is added and `true` in wp-config.
2. With this and the newsletters PR checked out, ensure your test site is connected to the ESP
3. Log into a reader account that exists in the ESP
4. Go to my account, update the email address, and save changes
5. Check the verification email and click the button to verify
6. Ensure you are redirected to my account with a success message and the new email address showing
7. Log into the ESP dashboard, and search for the new contact by email
8. Also verify the old contact email is no longer present
9. **For mailchimp**, check for a contact note indicating the contact was migrated, and the old contact email is now archived

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -8,6 +8,7 @@
 namespace Newspack;
 
 use Newspack\Reader_Activation;
+use Newspack\Reader_Activation\Sync\Metadata;
 use Newspack\Reader_Activation\ESP_Sync;
 use Newspack\Stripe_Connection;
 use Newspack\WooCommerce_Connection;
@@ -982,16 +983,19 @@ class WooCommerce_My_Account {
 	 * @param int    $user_id User ID.
 	 * @param string $new_email New email address.
 	 * @param string $old_email Old email address.
-	 *
-	 * @return bool
 	 */
 	public static function sync_email_change( $user_id, $new_email, $old_email ) {
+		if ( ! class_exists( 'Newspack_Newsletters_Contacts' ) ) {
+			return;
+		}
 		$contact = ESP_Sync::get_contact_data( $user_id );
 		if ( ! $contact ) {
-			return false;
+			return;
 		}
+		$list_id          = Reader_Activation::get_esp_master_list_id();
 		$existing_contact = array_merge( $contact, [ 'email' => $old_email ] );
-		$update = ESP_Sync::sync( $contact, 'Email Change', $existing_contact );
+		$contact          = Metadata::normalize_contact_data( $contact );
+		$update           = \Newspack_Newsletters_Contacts::upsert( $contact, $list_id, 'Email_Change', $existing_contact );
 		if ( is_wp_error( $update ) ) {
 			// TODO: reschedule the sync if failure is not due to existing email.
 			Logger::error( 'Error syncing email change: ' . $update->get_error_message() );
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -979,17 +979,19 @@ class WooCommerce_My_Account {
 	/**
 	 * Sync email change with site ESPs.
 	 *
-	 * @param int $user_id User ID.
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -864,25 +866,60 @@ public static function handle_verify_email_change() {
...
+						'ID'         => $user_id,
 						'user_email' => $new_email,
 					]
 				);
-				\wc_add_notice( __( 'Your email address has been successfully updated.', 'newspack-plugin' ) );
+				if ( \is_wp_error( $update ) ) {
+					\wc_add_notice( $error, 'error' );
+				} else {
+					\delete_user_meta( \get_current_user_id(), self::PENDING_EMAIL_CHANGE_META );
+					self::sync_email_change( $user_id, $new_email, $old_email );
```
adekbadek: `sync_email_change` method might not be successful -> error state should be handled here, instead of assuming success.

adekbadek (changes requested):
The happy-paths work, left some notes. I did test with AC and MC, but had trouble accessing CC (🤦) -> I'll test there later, hoping this was an intermittent error. 

1. The instructions neglected the step of configuring the `NEWSPACK_EMAIL_CHANGE_ENABLED` environment variable. 
2. WC will require a Display Name to be set, so updating the email address only won't be possible:

image

3. While the address change is initiated, but not confirmed, the email input in My Account displays the new email address. Is that intentional? IMO the old address should be displayed, with a notice saying that a change request was initiated. 
4. An accidental edge-case: initiate the change, log out and in, click the verification link -> the link does not work ("Something went wrong" error notice), with no recourse because the email address field is disabled. Not sure if it's worth examining, but raising for completeness.

chickenn00dle:
Thanks for the review @adekbadek!

> The instructions neglected the step of configuring the NEWSPACK_EMAIL_CHANGE_ENABLED environment variable.

Sorry about that! Derrick and Miguel had been reviewing the PRs associated with this feature, so I stopped including that step in the description. I should have remembered you would be coming back and kept it though. I've re-added.

> WC will require a Display Name to be set, so updating the email address only won't be possible

Ah. Good catch. Once a display name is set, you will be able to update the email address only, but when its not you will have to set both. I'm not sure we should allow this to be empty though, so not sure we should "fix" this as Woo seems to expect this to have a value when this form is submitted. In any case, I will open another task to address this since this is unrelated to ESP sync.

> While the address change is initiated, but not confirmed, the email input in My Account displays the new email address. Is that intentional? IMO the old address should be displayed, with a notice saying that a change request was initiated.

Good question. This is intentional and outlined in the scope here: https://newspackengineering.wordpress.com/2025/02/10/project-scope-change-email/

> An accidental edge-case: initiate the change, log out and in, click the verification link -> the link does not work ("Something went wrong" error notice), with no recourse because the email address field is disabled. Not sure if it's worth examining, but raising for completeness.

This was actually fixed in [another PR](https://github.com/Automattic/newspack-plugin/pull/3778) that was merged this morning. I've rebased so this should be resolved. Let me know if you still see this happening on your end though!

includes/reader-revenue/my-account/class-woocommerce-my-account.php
chickenn00dle: Yeah. Agreed. [I mentioned in the Asana task](https://app.asana.com/0/0/1209215513701719/1209530560533168/f) that my plan was to tackle handling error cases in a follow-up PR to avoid making these PRs too complex. But if you prefer this be done in a single PR, I can work on that here!

chickenn00dle:
FYI, task opened for the display name issue: https://app.asana.com/0/1208993180326452/1209560126738155/f

adekbadek (approved):

PR #3783, opened 2025-02-25
feat(woo-member-commenting): optional module for member commenting

naxoc:
### All Submissions:

* [x] Have you followed the [[Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [[WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [[VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [[Pull Requests](https://github.com/Automattic/pulls)](../../pulls) for the same update/change?

This adds a new optional module to Newspack. It makes it possible to only allow comments for user that have a given membership.  It's very much based on `newspack-woocommerce-member-commenting.php` in our custom code repo.

It's a first stab at this optional module, so there are some things that are a bit basic:

* The module can only be enabled from CLI because we don't have the UI for it yet. I added a small class that can activate/deactivate optional modules.
* The membership required is a constant in wp-config.php. 
* The slug (path?) to tease for buying the membership is a constant for now.

#### Questions

* I'm using the `includes/wizards/class-settings.php` class that is also used from the wizard – is that too risky?
* The text that tells the user to log in and/or buy a plan is hardcoded and I wasn't really sure what to do with it. Suggestions welcome.
* We should probably add some styling to the message in italics. Suggestions welcome.

#### To test the WP_CLI command

1. Run `wp newspack optional-module activate woo-member-commenting` and verify with `wp option get newspack_settings` that the module is enabled (look for true).
2. Run it again and verify that you get an error because it's already active.
3. Run it again with a module name that does not exist. `horse` is a great option there.
4. Run `wp newspack optional-module deactivate woo-member-commenting` and verify with `wp option get newspack_settings` that the module is disabled (look for false).
5. Run it again and check that you get an error.

#### To test the membership comments

1. Make sure you have `woocommerce` and `woocommerce-memberships` active.
2. Go to Woo -> Memberships -> Membership Plans -> and click "Add New Membership Plan" to add a plan – no need for any settings other than the slug. You'll need that, so take note.
3. Now go to Woo -> Members and add that membership to a user.
4. Put the settings for the module in `wp-config.php`. See below for how the array should look.
5. With the admin user, visit a post and verify that you see the comment form under it.
6. WIth the user you added the membership to, visit that post and verify you can comment.
7. With a non-logged in user, visit the post and verify that you get a link to log in and to buy a membership if you filled in the `membership_purchase_post_id` part of the settings.

```php
const NP_WC_MEMBER_COMMENT_SETTINGS = [
  'membership_plan_slug'        => [''], // Slug(s) of membership plans. 
  'membership_purchase_post_id'     => 6, // Post id to a page/post with membershipt tease
  'membership_required_message' => "Become a member, why don't cha?",
];
```

### Other information:

---

  - https://app.asana.com/0/0/1209425721676725

### DIFF
--- a/includes/optional-modules/class-woo-member-commenting.php
+++ b/includes/optional-modules/class-woo-member-commenting.php
@@ -1,0 +1,163 @@
+<?php
+/**
+ * Woo member commenting.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Woo member commenting.
+ */
+class Woo_Member_Commenting {
+
+	/**
+	 * Initialize.
+	 *
+	 * @return void
+	 */
+	public static function init(): void {
+		add_action( 'init', [ __CLASS__, 'on_init' ] );
+	}
+
+	/**
+	 * Action callback on init.
+	 *
+	 * @return void
+	 */
+	public static function on_init() {
+		if ( ! self::should_load() ) {
+			return;
+		}
+
+		add_filter( 'comment_form_defaults', [ __CLASS__, 'filter_comment_form_defaults' ] );
+		add_filter( 'comment_form_fields', [ __CLASS__, 'filter_comment_form_fields' ] );
+		add_filter( 'comment_form_submit_field', [ __CLASS__, 'filter_comment_form_submit_field' ] );
+	}
+
+	/**
+	 * Filter callback.
+	 *
+	 * @param array $defaults The defaults array to filter.
+	 *
+	 * @return mixed
+	 */
+	public static function filter_comment_form_defaults( $defaults ) {
+		if ( ! self::require_membership_to_comment_for_user() ) {
+			return $defaults;
+		}
+
+		$defaults['comment_notes_before'] = '';
+		$defaults['must_log_in']          = sprintf( '<p class="must-log-in">%s</p>', self::get_membership_required_message() );
+
+		return $defaults;
+	}
+
+
+	/**
+	 * Filter callback.
+	 *
+	 * @param array $fields The fields array to filter.
+	 *
+	 * @return array The filtered fields array.
+	 */
+	public static function filter_comment_form_fields( $fields ) {
+		return self::require_membership_to_comment_for_user() ? [] : $fields;
+	}
+
+	/**
+	 *  Filter callback.
+	 *
+	 * @param string $field Field to filter.
+	 *
+	 * @return string The filtered field.
+	 */
+	public static function filter_comment_form_submit_field( $field ) {
+		if ( ! self::require_membership_to_comment_for_user() ) {
+			return $field;
+		}
+
+		return sprintf( '<p class="must-log-in">%s</p>', self::get_membership_required_message() );
+	}
+
+	/**
+	 * Get the message to display to users that don't have access to commenting.
+	 *
+	 * @return string The message to display.
+	 */
+	private static function get_membership_required_message(): string {
+		$message = __( 'Only Members may post a comment.', 'newspack-plugin' );
+		if ( ! is_user_logged_in() ) {
+			$message .= ' ' . __( 'If you already have a membership, then <a href="/my-account/">sign in</a>.', 'newspack-plugin' );
+		}
+		if ( defined( 'NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG' ) && ! empty( NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG ) ) {
+			/* translators: %s - is the slug to buy a membership */
+			$message .= ' ' . sprintf( __( '<a href="/%s">Become a member now</a>.', 'newspack-plugin' ), NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG );
+		}
+
+		return $message;
+	}
+
+	/**
+	 * Whether class should suppress commenting for non-members.
+	 *
+	 * @return bool True if should suppress commenting for non-members, false otherwise.
+	 */
+	public static function should_load(): bool {
+		if ( empty( self::get_plan_slugs() ) ) {
+			return false;
+		}
+
+		return Settings::is_optional_module_active( 'woo-member-commenting' ) && function_exists( 'wc_memberships_get_user_memberships' ) && ! current_user_can( 'edit_posts' );
+	}
+
+	/**
+	 * Get slugs for membership plans that should allow commenting.
+	 *
+	 * This is managed by a constant for now – note that the constant can be a single string or an array of strings.
+	 *
+	 * @return array Array of plan slugs that allow commenting.
+	 */
+	private static function get_plan_slugs(): array {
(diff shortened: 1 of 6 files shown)


includes/optional-modules/class-woo-member-commenting.php
```diff
@@ -0,0 +1,163 @@
...
+	public static function init(): void {
+		add_action( 'init', [ __CLASS__, 'on_init' ] );
+	}
+
+	/**
+	 * Action callback on init.
+	 *
+	 * @return void
+	 */
+	public static function on_init() {
```
dkoo, line 31: In the past, we've included a snippet like this to automatically transition sites that have the standalone version of the plugin to the optional module instead (to avoid conflicts/collisions):

```suggestion
	public static function on_init() {
		// If the standalone plugin is active, deactivate it and activate as a module.
		if ( is_plugin_active( 'newspack-woocommerce-member-commenting/newspack-woocommerce-member-commenting.php' ) ) {
			deactivate_plugins( 'newspack-woocommerce-member-commenting/newspack-woocommerce-member-commenting.php' );
			Settings::activate_optional_module( 'woo-member-commenting' );
		}

```

includes/optional-modules/class-woo-member-commenting.php
```diff
@@ -0,0 +1,163 @@
...
+	 * @return string The message to display.
+	 */
+	private static function get_membership_required_message(): string {
+		$message = __( 'Only Members may post a comment.', 'newspack-plugin' );
+		if ( ! is_user_logged_in() ) {
+			$message .= ' ' . __( 'If you already have a membership, then <a href="/my-account/">sign in</a>.', 'newspack-plugin' );
+		}
+		if ( defined( 'NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG' ) && ! empty( NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG ) ) {
+			/* translators: %s - is the slug to buy a membership */
+			$message .= ' ' . sprintf( __( '<a href="/%s">Become a member now</a>.', 'newspack-plugin' ), NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG );
```
dkoo: (non-blocking) maybe just a subjective nitpick?

```suggestion
			$message .= ' ' . sprintf( __( '<a href="%s">Become a member now</a>.', 'newspack-plugin' ), home_url( NP_WC_MEMBER_COMMENT_MEMBERSHIP_PURCHASE_SLUG ) );
```

If we want to go above and beyond, we could also support a constant to provide a post ID and populate the permalink for it here.

includes/optional-modules/class-woo-member-commenting.php
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * Get the message to display to users that don't have access to commenting.
+	 *
+	 * @return string The message to display.
+	 */
+	private static function get_membership_required_message(): string {
+		$message = __( 'Only Members may post a comment.', 'newspack-plugin' );
+		if ( ! is_user_logged_in() ) {
+			$message .= ' ' . __( 'If you already have a membership, then <a href="/my-account/">sign in</a>.', 'newspack-plugin' );
```
dkoo: It's possible that some sites may have changed the base slug for the My Account page. Something like this could help avoid hard-coding a wrong URL for those sites:

```suggestion
			$message .= ' ' . sprintf( __( 'If you already have a membership, then <a href="%s">sign in</a>.', 'newspack-plugin' ), function_exists( 'wc_get_account_endpoint_url' ) ? \wc_get_account_endpoint_url( 'edit-account' ) : home_url( 'my-account' ) );
```

dkoo (changes requested):
Nice work and tests well! To answer some of your questions:

>  I'm using the includes/wizards/class-settings.php class that is also used from the wizard – is that too risky?

I think it's okay for now. This file has been moved/renamed in the `epic/ia` branch, so we'll need to update that branch too after this lands. Maybe in the IA branch we should move it under the `optional-modules` directory to keep it closer to the features it controls, too, but that's not in this PR's scope. I added the Information Architecture tag so we know there's some follow-up required there after this PR.

>  The text that tells the user to log in and/or buy a plan is hardcoded and I wasn't really sure what to do with it. Suggestions welcome.

WDYT about supporting just a single constant that's basically a settings object? This could let us support additional customizations if needed with `wp_parse_args`. Something like:

```php
const NP_WC_MEMBER_COMMENT_SETTINGS = [
	'membership_plan_slug'        => '',
	'membership_purchase_url'     => 0,
	'membership_required_message' => ''
];
```

This could also serve as a template for data that can eventually be configured via UI.

Other suggestions below!

includes/optional-modules/class-woo-member-commenting.php
naxoc: I like it. I put in support for that

includes/optional-modules/class-woo-member-commenting.php
naxoc: There are only 3 sites that use the old one. Since we are also changing the settings constant, I think it's safer to fix those 3 manually. Does that make sense?

includes/optional-modules/class-woo-member-commenting.php
naxoc: Good call! 👍

naxoc:
Thanks for the feedback, @dkoo – I added support for the array settings. I like that idea. Let me know if you disagree on https://github.com/Automattic/newspack-plugin/pull/3783#discussion_r1977550426

includes/optional-modules/class-woo-member-commenting.php
dkoo: Makes sense, thanks for clarifying.

includes/optional-modules/class-woo-member-commenting.php
```diff
@@ -0,0 +1,211 @@
...
+			return '';
+		}
+
+		$post_url = get_permalink( (int) $post_id );
+		$message  = self::get_module_setting( 'membership_required_message' );
+		if ( empty( $message ) ) {
+			$message = __( 'Become a member now', 'newspack-plugin' );
+		}
+
+		return sprintf( '<a href="%s">%s</a>.', $post_url, $message );
```
dkoo: Non-blocking because I'm unsure: Not sure if we should include punctuation here, in case the message contains its own punctuation ("Become a member today!")

```suggestion
		return sprintf( '<a href="%s">%s</a>', $post_url, $message );
```

dkoo (approved):
Thanks for the changes, @naxoc!

includes/optional-modules/class-woo-member-commenting.php
naxoc: Agreed!

PR #3740, opened 2025-02-10
Feat/change nicename UI

leogermani:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This is an experimental UI to https://github.com/Automattic/newspack-plugin/pull/3725.

This is not meant to be enabled for every user, but it can be enabled if there's a special need for this feature.

### How to test the changes in this Pull Request:

1. add `define( 'NEWSPACK_CHANGE_NICENAME_UI', true );`
2. Edit a user
3. Confirm there's a new section called "Change User archive URL"
4. Check test instructions on https://github.com/Automattic/newspack-plugin/pull/3725 and perform them, but using the UI this time

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3785, opened 2025-02-27
Hotfix/add change nicename

leogermani:
Adds the change nicename CLI and experimental UI

applies https://github.com/Automattic/newspack-plugin/pull/3725 and https://github.com/Automattic/newspack-plugin/pull/3740

You can see https://github.com/Automattic/newspack-plugin/pull/3740 for testing instructions

choose an author and open its archive page on a new tab
Edit their profile
Check the new "Change user slug" section
Confirm the "Check availability" works, by typing a few options there, including some you know are taken
Change the slug to something else
Back to the author's archive tab, reload the page and confirm it redirects you to the new slug

### DIFF
--- a/includes/plugins/co-authors-plus/class-nicename-change-ui.php
+++ b/includes/plugins/co-authors-plus/class-nicename-change-ui.php
@@ -1,0 +1,208 @@
+<?php
+/**
+ * Nicename change UI class.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+
+/**
+ * This class adds a UI to the user profile page to allow users to change their nicename.
+ */
+class Nicename_Change_UI {
+
+	/**
+	 * Registers the hooks.
+	 */
+	public static function init() {
+
+		if ( ! defined( 'NEWSPACK_CHANGE_NICENAME_UI' ) || ! NEWSPACK_CHANGE_NICENAME_UI ) {
+			return;
+		}
+
+		add_action( 'edit_user_profile', [ __CLASS__, 'edit_user_profile' ] );
+
+		add_action( 'admin_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
+
+		add_action( 'wp_ajax_newspack_change_nicename', [ __CLASS__, 'change_nicename_ajax' ] );
+		add_action( 'wp_ajax_newspack_change_nicename_check', [ __CLASS__, 'change_nicename_check_ajax' ] );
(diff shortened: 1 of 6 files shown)


includes/plugins/co-authors-plus/class-nicename-change.php
```diff
@@ -0,0 +1,301 @@
...
+	/**
+	 * Registers the hooks.
+	 */
+	public static function init() {
+		add_action( 'profile_update', [ __CLASS__, 'profile_update' ], 10, 2 );
+		add_action( 'template_redirect', [ __CLASS__, 'old_nicename_redirect' ] );
+
+		// add a cli command to update the nicename.
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			\WP_CLI::add_command( 'newspack nicename-change', [ __CLASS__, 'cli_change_nicename' ] );
```
naxoc, line 34: It would be great to get synopsis on this so it's easier to see from cli what the args should look like.

includes/plugins/co-authors-plus/class-nicename-change.php
```diff
@@ -0,0 +1,301 @@
...
+
+		// Update the nicename.
+		wp_update_user(
+			[
+				'ID'            => $user_id,
+				'user_nicename' => $new_nicename,
+			]
+		);
+
+		\WP_CLI::success( 'Nicename updated.' );
```
naxoc, line 97: Mayeb output the username as well for the user – it's easy when only providing an ID to make a mistake and not realize what user was changed.

includes/plugins/co-authors-plus/class-nicename-change-ui.php
```diff
@@ -0,0 +1,208 @@
...
+	public static function change_nicename_ajax() {
+
+		check_ajax_referer( 'newspack_change_nicename_nonce', 'nonce' );
+
+		$new_nicename = isset( $_POST['new_nicename'] ) ? sanitize_title( wp_unslash( $_POST['new_nicename'] ) ) : '';
+
+		$existing = Nicename_Change::get_existing_nicenames( $new_nicename );
+
+		if ( ! empty( $existing ) ) {
+			return self::change_nicename_check_ajax();
```
naxoc, line 118: This function returns nothing, so we can remove the return.

naxoc (approved):
Tested this and it works well 👍 

I'll merge this one now because we want it released, but I added a couple of nit-comments that would be good to come back to. Especially the one about letting the user know from CLI what the usrname or email is on the usr being changed.

PR #3792, opened 2025-03-03
fix(my-account): handle email change esp sync errors

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208993180326452/1209215513701719

This and [the partner newsletters PR](https://github.com/Automattic/newspack-newsletters/pull/1769) account for errors when attempting to sync an email change with ESPs.

We account for two types of issues:

1. The email address we are updating to already exists in the ESP. In these cases, we delete the old contact and trigger an update request to the new contact.
2. Any other generic error is handled by rescheduling the sync for 24 hours.

Be sure to also checkout https://github.com/Automattic/newspack-newsletters/pull/1769 before testing.

1. Ensure the `NEWSPACK_EMAIL_CHANGE_ENABLED` constant is defined and true in wp-config.
2. Ensure your test site ESP is mailchimp and connected
3. Log into a reader account that exists in mailchimp and update the email address to a new one altogether
4. Complete verification then check mailchimp to ensure the email address updated correctly for the contact
5. Now repeat step 3, but this time update to an email address that exists in the ESP but does not exist in the test site (you may need to manually add the contact in MC).
6. Complete verification then check mailchimp to ensure the old contact has been removed and the one that you updated to still exists
7. Now go Engagement > Newsletters and replace the ESP key with something invalid
8. Repeat step 3 one more time
9. Complete verification then confirm a new `newspack_esp_sync_email_change` cron job has been scheduled (`wp cron event list`)
10. Fix the ESP key, then manually trigger the cron job (`wp cron event run newspack_esp_sync_email_change`)
11. Check mailchimp and verify the email update occurred as expected
12. Repeat the above steps for both Active Campaign and Constant Contact

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -1013,8 +1013,8 @@ class WooCommerce_My_Account {
 			return;
 		}
 		$list_id          = Reader_Activation::get_esp_master_list_id();
-		$contact          = Metadata::normalize_contact_data( $contact );
 		$existing_contact = array_merge( $contact, [ 'email' => $old_email ] );
+		$contact          = Metadata::normalize_contact_data( $contact );
 		$update           = \Newspack_Newsletters_Contacts::upsert( $contact, $list_id, 'Email_Change', $existing_contact );
 		if ( is_wp_error( $update ) ) {
 			// TODO: reschedule the sync if failure is not due to existing email.
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -1023,9 +1023,9 @@ class WooCommerce_My_Account {
 		$contact          = Metadata::normalize_contact_data( $contact );
 		$update           = \Newspack_Newsletters_Contacts::upsert( $contact, $list_id, 'Email_Change', $existing_contact );
 		if ( is_wp_error( $update ) ) {
-			// If the update failed, retry in 1 hour.
-			\wp_schedule_single_event( time() + HOUR_IN_SECONDS, self::SYNC_ESP_EMAIL_CHANGE_CRON_HOOK, [ $user_id, $new_email, $old_email ] );
-			Logger::error( 'Error syncing email change with ESP: ' . $update->get_error_message() . '. Retrying in 1 hour.' );
+			// If the update failed, retry in 24 hours.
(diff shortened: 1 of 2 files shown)


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -997,11 +1003,10 @@ public static function sync_email_change( $user_id, $new_email, $old_email ) {
 		$contact          = Metadata::normalize_contact_data( $contact );
 		$update           = \Newspack_Newsletters_Contacts::upsert( $contact, $list_id, 'Email_Change', $existing_contact );
 		if ( is_wp_error( $update ) ) {
-			// TODO: reschedule the sync if failure is not due to existing email.
-			Logger::error( 'Error syncing email change: ' . $update->get_error_message() );
-			return false;
+			// If the update failed, retry in 24 hours.
+			\wp_schedule_single_event( time() + DAY_IN_SECONDS, self::SYNC_ESP_EMAIL_CHANGE_CRON_HOOK, [ $user_id, $new_email, $old_email ] );
```
chickenn00dle: I didn't set a limit to the number of retries here. Since we are handling the case of existing email address within the various provider upsert methods, any other types of errors should be resolvable I think. But let me know if we prefer to set a limit here instead!

miguelpeixe (changes requested):
In the `ESP_Sync` class we have `sync()` and `schedule_sync()` methods. This strategy and the `sync_email_change()` method are creating a duplicated strategy for managing contact synchronization.

I worry that this creates risks in the long term, which we recently worked on mitigating by centralizing how syncs behave.

Is it possible to leverage the existing methods from the `ESP_Sync` class for this?

chickenn00dle:
> Is it possible to leverage the existing methods from the ESP_Sync class for this?

Unfortunately, the `ESP_Sync` methods only work with Mailchimp and Active Campaign, but not Constant Contact.

I can probably modify these methods to trigger for constant contact for only this very specific case, but I think its cleaner to handle this logic separately instead and keep the sync methods clean.

Happy to go the other route if we prefer though!

miguelpeixe:
Hmm, good point. Is it unsupported solely because of a missing "master list"?

If that's the case, we'd benefit more from allowing a sync to go through on CC without a master list. Perhaps return `-1` [here](https://github.com/Automattic/newspack-plugin/blob/a3493cccad6437d3e1afcb15ffd53a08c5c595e6/includes/reader-activation/class-reader-activation.php#L483-L507)?

If there are other reasons or that's not a move we'd like to make now, we could also move the `sync_email_change()` as is to the `ESP_Sync` class. It'd be easier to refactor if we decide to in the future.

leogermani:
I think there's no reason to not support Constant Contact fully in RAS, meaning also adding an option to select a "master list" for it

miguelpeixe (approved):
Thank you for revising!

PR #3796, opened 2025-03-05
feat(reader-auth): send WP login reminder email to non-reader accounts

dkoo:
### All Submissions:

Addresses some persistent confusion expressed by publishers who are trying to use reader-based login and registration flows to log into their non-reader (admin, editor, author, etc.) accounts. In this scenario, we can send the user an email to remind them to log in via the regular WP login page instead of reader-facing UIs.

1. Check out this branch. 
2. Visit **Newspack > Engagement > Reader Activation > Advanced Settings > Transactional Email Content** and confirm there's a new "Non-reader account" email at the bottom of the list.

Screenshot 2025-03-04 at 5 17 47 PM

3. Edit this email and confirm it looks something like this:

Screenshot 2025-03-04 at 5 08 55 PM

4. In a new incognito session, attempt to register a reader account via the Registration block using an email address already associated with an admin, editor, or author user account. Confirm that you see the message that the account is already registered:

Screenshot 2025-03-04 at 5 08 26 PM

5. Check the email inbox associated with the email address and confirm that you got the reminder email as in step 3. Confirm that the "Continue to [Site Name]" button leads to the WP login form page.
6. In the incognito session, attempt to login and register via the "Sign In" modal. Confirm that you see the following when attempting to login or register, and that in either case you get the reminder email in the email address's inbox:

Logging in (note: we could show a message directing the user to check their email here instead, if desired):
 
Screenshot 2025-03-04 at 5 08 06 PM

Registering:

Screenshot 2025-03-04 at 5 41 04 PM

7. Update and save the email template in step 3, repeat steps 4-5, and confirm that the sent reminder email matches the updated content.
8. Try to register via the Sign In modal with an existing reader account email address and confirm that you see the following messages depending on whether the reader account has set a password:

Has a password:
Screenshot 2025-03-04 at 5 40 50 PM

No password:
Screenshot 2025-03-04 at 5 43 34 PM

9. Smoke test registering a new reader account and logging into an existing reader account using both password and OTP. 

### Other information:


includes/templates/reader-activation-emails/non-reader.php
```diff
@@ -0,0 +1,211 @@
...
...:4px;cursor:auto;mso-padding-alt:12px 24px;background:#ff8200 !important;" valign="middle"><a href="*WP_LOGIN_URL*" style="display:inline-block;background:#ff8200 !important;color:#fff !important;font-family:Georgia;font-size:16px;font-weight:bold;line-height:1.5;margin:0;text-decoration:none;text-transform:none;padding:12px 24px;mso-padding-alt:0px;border-radius:4px;" target="_blank">Continue to *SITE_TITLE*</a></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table><![endif]--> <img src="https://dkoo.jurassic.tube/wp-content/np-newsletters-pixel.php?id=257&#038;tid=oSa1m1Ic1jUXCCuAXE5y290Fd1R7M5Eo&#038;em=*|EMAIL|*" width="1" height="1" alt="" style="display: block; width: 1px; height: 1px; border: none; margin: 0; padding: 0;"></div></body></html>';
```
adekbadek: I see instances of `https://dkoo.jurassic.tube?np_newsletters_click` in the body here. Should be removed.

includes/reader-activation/class-reader-activation.php
```diff
@@ -2080,9 +2086,16 @@ public static function register_reader( $email, $display_name = '', $authenticat
...
 		$user_id = false;
 
-		if ( $existing_user && self::is_reader_without_password( $existing_user ) ) {
+		if ( $existing_user ) {
+			if ( ! self::is_user_reader( $existing_user ) ) {
+				self::send_non_reader_login_reminder( $existing_user );
+				return false;
+			}
+
 			// Don't send OTP email for newsletter signup.
```
adekbadek: This comment should be updated to explain the new condition.

adekbadek (changes requested):
> publishers who are trying to use reader-based login and registration flows to log into their non-reader (admin, editor, author, etc.) accounts

I assume the reason for the email solution is because we don't want to enable fishing for admin email addresses. Just noting because I'd expect that noted somewhere, preferably in the code.

I think the reminder email content should be more explicit about using the `/wp-admin` link and point out that the user tried to use a login flow reserved for readers.

includes/reader-activation/class-reader-activation.php
dkoo: Added an explanation in 458121d. Also made the email message more explicit in 8f96cc4:

> Someone attempted to sign into your account on *SITE_TITLE* using a login method reserved for reader accounts. Please use the following link to sign in via the WordPress admin dashboard:

dkoo:
Thanks for the feedback, @adekbadek! I've responded in the threads above.

includes/templates/reader-activation-emails/non-reader.php
dkoo: Whoops, thanks for catching! Fixed in 8a827ac.

dkoo:
@adekbadek requesting another review on this when you have a chance!

includes/templates/reader-activation-emails/non-reader.php
```diff
@@ -0,0 +1,225 @@
...
...e:0px;padding:0;word-break:break-word;"><table border="0" cellpadding="0" cellspacing="0" role="presentation" style="border-collapse:separate;line-height:100%;"><tbody><tr><td align="center" bgcolor="#ff8200 !important" role="presentation" style="border:none;border-radius:4px;cursor:auto;mso-padding-alt:12px 24px;background:#ff8200 !important;" valign="middle"><a href="*WP_LOGIN_URL*" style="display:inline-block;background:#ff8200 !important;color:#fff !important;font-family:Georgia;font-size:16px;font-weight:bold;line-height:1.5;margin:0;text-decoration:none;text-transform:none;padding:12px 24px;mso-padding-alt:0px;border-radius:4px;" target="_blank">Continue to *SITE_TITLE*</a></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table></td></tr></table><![endif]--></td></tr></tbody></table></div><!--[if mso | IE]></td></tr></table><![endif]--></div>
```
adekbadek: The string `Sign in Someone […]` should be translatable, like the `Someone attempted […]` one later in the HTML.

includes/reader-activation/class-reader-activation.php
adekbadek: Thanks! What I meant was that the `Don't send OTP email for newsletter signup.` comment should be updated following the update of the condition, something like:

```suggestion
			// Don't send OTP email for newsletter signup, or if the reader has a password set. 
```

adekbadek (changes requested):
Sorry for the late reply 🙈

includes/templates/reader-activation-emails/non-reader.php
dkoo: Thanks! 07336c81 should fix this

adekbadek (approved):

includes/reader-activation/class-reader-activation.php
```diff
@@ -2001,10 +2001,16 @@ public static function process_auth_form() {
 		self::set_auth_intention_cookie( $email );
 
 		$user = \get_user_by( 'email', $email );
-		if ( ( ! $user && 'register' !== $action ) || ( $user && ! self::is_user_reader( $user ) ) ) {
+		if ( ! $user && 'register' !== $action ) {
 			return self::send_auth_form_response( new \WP_Error( 'unauthorized', wp_kses_post( __( 'Account not found. <a data-set-action="register" href="#register_modal">Create an account</a> instead?', 'newspack-plugin' ) ) ) );
 		}
 
+		if ( ! self::is_user_reader( $user ) ) {
```
miguelpeixe, line 2008: This is not taking into account the account registration flow and will throw a fatal error if I attempt to register via the auth modal. Needs a `! $user` check.

miguelpeixe: Fix at #3944

PR #3799, opened 2025-03-05
feat(my-account): sync admin email change with external connections

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1208993180326452/1209589552372839

This PR triggers a sync with both Stripe and the site ESP when an email address is updated by an admin

1. Make sure the `NEWSPACK_EMAIL_CHANGE_ENABLED` constant is defined and true in wp-config
2. As an admin go to Users and select a user that is associated with a Stripe account
3. Change the email address, save, then check stripe to confirm the email address was updated
4. Repeat step 1, but this time with a user that is associated with a contact from the sites ESP
5. Again, change the email address, save, then check the ESP to confirm the email address was updated

Note: error cases for ESP sync are being tackled in another PR, so be sure to only update to an email address that doesn't exist in the ESP.

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -72,6 +72,7 @@ class WooCommerce_My_Account {
 			\add_filter( 'wcs_my_account_redirect_to_single_subscription', [ __CLASS__, 'redirect_to_single_subscription' ] );
 			\add_filter( 'wc_memberships_members_area_my-memberships_actions', [ __CLASS__, 'hide_cancel_button_from_memberships_table' ] );
 			\add_filter( 'wc_memberships_my_memberships_column_names', [ __CLASS__, 'remove_next_bill_on' ], 21 );
+			\add_action( 'profile_update', [ __CLASS__, 'handle_admin_email_change_request' ], 10, 3 );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -896,6 +897,25 @@ public static function handle_email_change_request( $user_id ) {
...
+	 * Handle admin email change request.
+	 *
+	 * @param int     $user_id User ID.
+	 * @param WP_User $user    User object.
+	 * @param array   $data    User data.
+	 */
+	public static function handle_admin_email_change_request( $user_id, $user, $data ) {
+		if ( ! is_admin() || ! self::is_email_change_enabled() ) {
+			return;
+		}
```
miguelpeixe, line 910: Should this check be preventing an email update? Bailing here just means the email change happened without the required syncs.

miguelpeixe (changes requested):
Works as described! Just a question regarding the guard clause.

includes/reader-revenue/my-account/class-woocommerce-my-account.php
chickenn00dle: I believe so. We bail here if the request is not from admin since non-admin requests are handled via a different hook, or if the feature is not enabled to maintain the existing behavior.

chickenn00dle: Oops, just realized I didn't directly answer your first question:

> Should this check be preventing an email update?

No. At least we aren't doing this currently, so my assumption is this is expected.

miguelpeixe: Thanks for explaining! I believe we could go without this clause. `is_email_change_enabled()` is not whether the sync should happen, it's whether we allow a reader to change their email. Non-blocking, though!

miguelpeixe (approved):

includes/reader-revenue/my-account/class-woocommerce-my-account.php
chickenn00dle: > I believe we could go without this clause. is_email_change_enabled() is not whether the sync should happen, it's whether we allow a reader to change their email. Non-blocking, though!

I think you're correct, although the goal here is to make sure the behavior remains consistent when this feature is not enabled, so we'd need this to make sure admins can still change email addresses without triggering the sync.

I'll keep it for now, but can remove if it causes issues down the line!

PR #3794, opened 2025-03-04
fix: is-without-password condition

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

https://github.com/Automattic/newspack-plugin/pull/3747 introduced a regression, this PR fixes it. 

Also fixes an assertion in the `test_register_existing_reader` test - if a user exists, the value returned is false, not a WP Error.

### How to test the changes in this Pull Request:

1. On `trunk`, register as a new reader, log out, then attempt to sign in again using the auth modal
2. Observe that after clicking "Continue" after the email input, the state of the modal progresses to password prompt - even though a password has never been set
3. Switch to this branch, observe that the second state of the modal is an OTP input

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3800, opened 2025-03-06
Hotfix/ras sign in

adekbadek:
https://github.com/Automattic/newspack-plugin/pull/3794 as a hotfix


adekbadek:
Assuming the approval from https://github.com/Automattic/newspack-plugin/pull/3794 counts here.

PR #3801, opened 2025-03-06
feat(rss-feed): Add Feed enhancements for Republications.

Takshil-Kunadia:
### All Submissions:

Addresses https://app.asana.com/0/0/1209344511155915/f .

1. Add relevant options to the RSS Feed settings, if the Republication tracker plugin is activated:
    - A checkbox to add a republication tracker snippet to posts.
    - A checkbox to filter the feed to only include republishable posts.
    - A select field to choose between RSS and Atom feed formats.
2. When the user selects the Atom feed format, the "Wrap the content of <title> elements in CDATA tags" option is disabled and forced to false (since WordPress Atom feeds encoded title by default).
3. Adjusts the feed URL generation and query modifications to reflect the selected feed format.
4. Outputs the <republication_tracker> tag in the feed when the tracker option is enabled.
5. Modifies the meta query so that only posts not marked to hide the republishing button are included when filtering is active.

1. Install & Configure Republication Tracker Tool.
2. Create Feeds from Admin > RSS Feeds.
3. Check for new options.
   - Checkbox for adding republication tracker pixel.
   - Checkbox for only including republishable posts.
   - Select for RSS or Atom feeds.
4. Try the above settings and check the feed link generated.
6. If republication tracker pixel checkbox is clicked a `<republication_tracker>` tag is included adding the tracking pixel.
7. If only Include Republishable posts checkbox is added, only posts which have `"Hide Republishing Button"` metabox checkbox unchecked will be included.
8. RSS & Atom feeds will be displayed based on the selected feed type.

### Other information:

### DIFF
--- a/includes/optional-modules/class-rss.php
+++ b/includes/optional-modules/class-rss.php
@@ -465,7 +465,9 @@ class RSS {
 						<input type="checkbox" name="only_republishable" value="1" <?php checked( $settings['only_republishable'] ); ?> />
 					</td>
 				</tr>
-				<tr>
+
+			<?php endif; ?>
+			<tr>
 					<th><?php esc_html_e( 'Feed format', 'newspack-plugin' ); ?></th>
 					<td>
 						<select name="feed_format">
@@ -474,7 +476,6 @@ class RSS {
 						</select>
 					</td>
 				</tr>
-			<?php endif; ?>
 		</table>
 		<?php
 	}
@@ -573,9 +574,10 @@ class RSS {
 			$only_republishable             = filter_input( INPUT_POST, 'only_republishable', FILTER_SANITIZE_NUMBER_INT );
 			$settings['only_republishable'] = (bool) $only_republishable;
 
-			$feed_format             = filter_input( INPUT_POST, 'feed_format', FILTER_SANITIZE_SPECIAL_CHARS );
-			$settings['feed_format'] = in_array( $feed_format, [ 'rss', 'atom' ] ) ? $feed_format : 'rss';
 		}
+
+		$feed_format             = filter_input( INPUT_POST, 'feed_format', FILTER_SANITIZE_SPECIAL_CHARS );
+		$settings['feed_format'] = in_array( $feed_format, [ 'rss', 'atom' ] ) ? $feed_format : 'rss';
 
 		update_post_meta( $feed_post_id, self::FEED_SETTINGS_META, $settings );
 		// @todo flush feed cache here.
--- a/includes/optional-modules/class-rss.php
+++ b/includes/optional-modules/class-rss.php
@@ -376,7 +376,7 @@ class RSS {
 
 			<?php
 			// Only show this new option if the Republication Tracker Tool plugin is active.
-			if ( class_exists( 'Republication_Tracker_Tool' ) ) :
+			if ( self::is_republication_tracker_plugin_active() ) :
 				?>
 				<tr>
 					<th>
@@ -479,7 +479,7 @@ class RSS {
 			<?php endif; ?>
 			<?php
 			// Only show this new option if the Republication Tracker Tool plugin is active.
-			if ( class_exists( 'Republication_Tracker_Tool' ) ) :
+			if ( self::is_republication_tracker_plugin_active() ) :
 				?>
 				<tr>
 					<th><?php esc_html_e( 'Add republication tracker snippet to posts', 'newspack-plugin' ); ?></th>
@@ -581,7 +581,7 @@ class RSS {
 		}
 
 		// Process Republication Tracker options only if the plugin is active.
-		if ( class_exists( 'Republication_Tracker_Tool' ) ) {
+		if ( self::is_republication_tracker_plugin_active() ) {
 			$republication_tracker             = filter_input( INPUT_POST, 'republication_tracker', FILTER_SANITIZE_NUMBER_INT );
 			$settings['republication_tracker'] = (bool) $republication_tracker;
 
@@ -653,7 +653,7 @@ class RSS {
 			);
 		}
 
-		if ( class_exists( 'Republication_Tracker_Tool' ) && ! empty( $settings['only_republishable'] ) ) {
+		if ( self::is_republication_tracker_plugin_active() && ! empty( $settings['only_republishable'] ) ) {
 			$meta_query = $query->get( 'meta_query' );
 			if ( ! is_array( $meta_query ) ) {
 				$meta_query = [];
@@ -819,6 +819,16 @@ xmlns:media="http://search.yahoo.com/mrss/"
 		}
 
 		return $title;
+	}
+
+	/**
+	 * Check if the Republication Tracker Tool plugin is active.
+	 * This is used to determine whether to show additional options in the RSS feed settings.
+	 *
+	 * @return bool Whether the Republication Tracker Tool plugin is active.
+	 */
+	private static function is_republication_tracker_plugin_active() {
+		return class_exists( 'Republication_Tracker_Tool' );
 	}
 }
 RSS::init();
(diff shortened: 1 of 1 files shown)


includes/optional-modules/class-rss.php
```diff
@@ -342,9 +349,22 @@ public static function render_content_settings_metabox( $feed_post ) {
 		</table>
 
 		<script>
+			const handleCDATA = function() {
```
Takshil-Kunadia: Need to do this as [WordPress atom feed wraps title in CDATA by default](https://github.com/WordPress/WordPress/blob/ac459cf68898409a90352539dbc9b7d319f123d5/wp-includes/feed-atom.php#L74).

includes/optional-modules/class-rss.php
```diff
@@ -427,6 +447,34 @@ public static function render_technical_settings_metabox( $feed_post ) {
...
+				?>
+				<tr>
+					<th><?php esc_html_e( 'Add republication tracker snippet to posts', 'newspack-plugin' ); ?></th>
+					<td>
+						<input type="hidden" name="republication_tracker" value="0" />
+						<input type="checkbox" name="republication_tracker" value="1" <?php checked( $settings['republication_tracker'] ); ?> />
+					</td>
+				</tr>
+				<tr>
+					<th><?php esc_html_e( 'Only include republishable posts', 'newspack-plugin' ); ?></th>
```
adekbadek: This setting rather belongs in the "Content Settings" section. Also, it's pretty powerful, so a help text describing it should be added: "When toggled on, posts which have republication disabled will be excluded from the feed."

includes/optional-modules/class-rss.php
```diff
@@ -427,6 +447,34 @@ public static function render_technical_settings_metabox( $feed_post ) {
...
+					<td>
+						<input type="hidden" name="only_republishable" value="0" />
+						<input type="checkbox" name="only_republishable" value="1" <?php checked( $settings['only_republishable'] ); ?> />
+					</td>
+				</tr>
+				<tr>
+					<th><?php esc_html_e( 'Feed format', 'newspack-plugin' ); ?></th>
```
adekbadek: This setting is problematic. Currently it does two things:

1. Displays the Atom or RSS feed link in the feed settings -> the problem here is that both feed links are always available, displaying one or the other in the settings suggests to the user that only the selected one is available. On top of that, the user would expect the link to remain the same, with the format different depending on the chosen feed type.  
3. Handles the wrapping of CDATA -> but if RSS is chosen as the feed type, and "Wrap the content of <title> elements in CDATA tags" option toggled on, the Atom feed will have a double-wrapping (`<![CDATA[<![CDATA[`)

To remediate this issues I suggest:

1. Instead of the misleading "Feed format" option, display both RSS and Atom links in the feed settings. This will reflect the fact that both are always available. 
2. If the "Wrap the content of <title> elements in CDATA tags" option is toggled on, prevent it from having effect on Atom feeds.

adekbadek (changes requested):

leogermani:
oh, I made some changes before seeing @adekbadek 's comments. Please take his advice, they're better. I didn't know that both feed formats were available for all custom feeds.

One thing I did though is to only apply the query filter for republishable posts if the plugin is active

leogermani:
Also, you are including the tracking snippet as a new tag in the feed. Not sure if that's what we want or if we just want to append that to the post content itself. @claudiulodro ?

includes/optional-modules/class-rss.php
Takshil-Kunadia: Moved it to content settings & added help text . ✅

includes/optional-modules/class-rss.php
Takshil-Kunadia: >the problem here is that both feed links are always available. Instead of the misleading "Feed format" option, display both RSS and Atom links in the feed settings.

I thought of this too! but assumed this was intentional. With that in mind, I've updated the Feed URLs metabox to show both links.

> If the "Wrap the content of <title> elements in CDATA tags" option is toggled on, prevent it from having effect on Atom feeds.

I had reported [this](https://github.com/Automattic/newspack-plugin/pull/3801#discussion_r1983029512) and its already taken care of in https://github.com/Automattic/newspack-plugin/pull/3801/commits/268e93df5225b5cd2d20177854bb7f401fed223b. I have refactored that part to work with the updated changes.

cc: @leogermani

Takshil-Kunadia:
@adekbadek @leogermani Thanks for the thoughtful feedback! 
I have made the changes required. Please review 🙇

includes/optional-modules/class-rss.php
```diff
@@ -244,14 +261,31 @@ public static function render_url_metabox( $feed_post ) {
...
+				<td>
+					<h3>
+						<a href="<?php echo esc_url( $rss_feed_url ); ?>" target="_blank">
+							<?php echo esc_url( $rss_feed_url ); ?>
+						</a>
+					</h3>
+				</td>
+			</tr>
+			<tr>
+				<td><h3><?php esc_html_e( 'Atom :-', 'newspack-plugin' ); ?></h3></td>
```
adekbadek: The `:-` looks odd. How about a simple dash (here & in the other instance above)?
 ```suggestion
				<td><h3><?php esc_html_e( 'Atom -', 'newspack-plugin' ); ?></h3></td>
```

File: includes/optional-modules/class-rss.php [hunk 5]
```diff
@@ -339,6 +373,22 @@ public static function render_content_settings_metabox( $feed_post ) {
...
 			</tr>
+
+			<?php
+			// Only show this new option if the Republication Tracker Tool plugin is active.
+			if ( self::is_republication_tracker_plugin_active() ) :
+				?>
+				<tr>
+					<th>
+						<?php esc_html_e( 'Only include republishable posts', 'newspack-plugin' ); ?>
+						<p class="description"><?php echo esc_html_x( '(When toggled on, posts which have republication disabled will be excluded from the feed.)', 'help text for only republishable setting', 'newspack-plugin' ); ?></p>
```
adekbadek: Since this is already styled as a help/description text, there's no need for parentheses:
```suggestion
						<p class="description"><?php echo esc_html_x( 'When toggled on, posts which have republication disabled will be excluded from the feed.', 'help text for only republishable setting', 'newspack-plugin' ); ?></p>
```

includes/optional-modules/class-rss.php
```diff
@@ -655,6 +741,13 @@ public static function add_extra_tags() {
 				}
 			}
 		}
+
+		// If the Republication Tracker Tool is enabled and the option is checked, output the tracker snippet.
+		if ( ! empty( $settings['republication_tracker'] ) && method_exists( 'Republication_Tracker_Tool', 'create_tracking_pixel_markup' ) ) {
+			?>
+			<republication_tracker><?php echo \Republication_Tracker_Tool::create_tracking_pixel_markup( $post->ID ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></republication_tracker>
```
adekbadek: Echoing @leogermani 's [question](https://github.com/Automattic/newspack-plugin/pull/3801#issuecomment-2706644119) – why is the pixel wrapped in a custom tag?

adekbadek (changes requested):

claudiulodro:
For the tracking snippet, let's include that as a <script> tag at the end of the post content please,[ like how the pixel is included in the "regular" content](https://github.com/Automattic/republication-tracker-tool/blob/21f387166fdda64b7a6c4a95a31da360f328f5ad/republication-tracker-tool.php#L246C25-L283)

includes/optional-modules/class-rss.php
adekbadek: (answered [here](https://github.com/Automattic/newspack-plugin/pull/3801#issuecomment-2711314201))

Takshil-Kunadia: @adekbadek  I'm not entirely sure I captured the change as intended, so I've updated the code to append the tracking snippet at the end of the content in [feat(rss): address tracking snippet feedback](https://github.com/Automattic/newspack-plugin/pull/3801/commits/cc0e2a4f54060150ca3053ac9103b4f040770bf0) along with taking care of attribution setting as shared by @claudiulodro . Please do let me know if this is what was required or if further adjustments are needed.

cc: @claudiulodro

adekbadek (approved):

PR #3807, opened 2025-03-07
feat: add Lite sites feature

leogermani:
### All Submissions:

Adds the Lite site feature

Note: Built a simple Settings page for now - we can build a Newspack settings page after AI is released

1. Go to Settings > Lite Site
2. Enable the feature and save
3. Visit yoursite.com/text and confirm you see a text only version of the site
4. Visit the posts and confirm they look good
5. Confirm images and any other non-textual elements are removed from the posts
6. Confirm the HR in the header and footer follow Newspack theme's primary color
7. Tweak the settings and confirm they work as expected

### Other information:

### DIFF
--- a/includes/lite-site/class-lite-site.php
+++ b/includes/lite-site/class-lite-site.php
@@ -85,14 +85,14 @@ class Lite_Site {
 
 		add_settings_section(
 			'newspack_lite_site_main',
-			__( 'Lite Site Settings', 'newspack' ),
+			__( 'Lite Site Settings', 'newspack-plugin' ),
 			'__return_null',
 			'newspack_lite_site'
 		);
 
 		add_settings_field(
 			'enabled',
-			__( 'Enable Lite Site', 'newspack' ),
+			__( 'Enable Lite Site', 'newspack-plugin' ),
 			[ __CLASS__, 'render_enabled_field' ],
 			'newspack_lite_site',
 			'newspack_lite_site_main'
@@ -100,7 +100,7 @@ class Lite_Site {
 
 		add_settings_field(
 			'url_base',
-			__( 'URL Base', 'newspack' ),
+			__( 'URL Base', 'newspack-plugin' ),
 			[ __CLASS__, 'render_url_base_field' ],
 			'newspack_lite_site',
 			'newspack_lite_site_main'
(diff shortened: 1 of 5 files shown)


includes/lite-site/templates/archive.php
```diff
@@ -0,0 +1,56 @@
...
+
+	foreach ( $recent_posts as $current_post ) {
+		printf(
+			'<li><a href="/text/%d">%s</a></li>',
+			esc_attr( $current_post->ID ),
+			esc_html( $current_post->post_title )
+		);
+	}
+	?>
+	</ul>
```
adekbadek, line 45: What about pagination? The setting is called "Posts per Page", which suggest pagination.

includes/lite-site/class-lite-site.php
```diff
@@ -0,0 +1,469 @@
...
+		);
+	}
+
+	/**
+	 * Render settings page
+	 */
+	public static function render_settings_page() {
+		?>
+		<div class="wrap">
+			<h1><?php esc_html_e( 'Lite Site Settings', 'newspack-plugin' ); ?></h1>
```
adekbadek: IMO there should be a text here describing that the feature is about, and in what situation it might be needed.

includes/lite-site/templates/archive.php
leogermani: Yeah, I renamed the setting. There's no pagination

includes/lite-site/class-lite-site.php
leogermani: Good call, I'll propose something and ask for feedback, then I'll update the PR

leogermani: Done in https://github.com/Automattic/newspack-plugin/pull/3807/commits/2f1d58868f18edec4631d42a2356b751fa352c5a along with a few other tweaks

includes/lite-site/class-lite-site.php
```diff
@@ -150,7 +150,9 @@ public static function add_menu_page() {
 	public static function render_settings_page() {
 		?>
 		<div class="wrap">
-			<h1><?php esc_html_e( 'Lite Site Settings', 'newspack-plugin' ); ?></h1>
+			<h1><?php esc_html_e( 'Lite Site', 'newspack-plugin' ); ?></h1>
+			<p>Lite Site is a text-only version of this website that loads faster and uses less data.</p>
+			<p>It’s designed to allow your readers to still be able to access your content despite connectivity issues, poor network coverage, or in the event of natural disasters and emergencies.</p>
```
adekbadek: These strings should be translatable

adekbadek (changes requested):

includes/lite-site/class-lite-site.php
leogermani: of course

adekbadek (approved):

PR #3810, opened 2025-03-07
fix: update paths to share button, social CSS in Perfmatters

laurelfulford:
### All Submissions:

I noticed while digging into an image size issue on one of our sites that their share buttons were super slow to load, and just sat in the bulleted list form for a while. 

After digging around a bit, it looks like the file structure used for Jetpack when loading static files from the CDN is pretty different from when it's loading them the plugin. The sharing CSS loads from https://c0.wp.com/p/jetpack/14.4/modules/sharedaddy/sharing.css and the social logos load from https://c0.wp.com/p/jetpack/14.4/_inc/social-logos/social-logos.min.css. 

This PR updates the paths in the Perfmatters settings so they should work when the files are served from the CDN and when they are not. Hopefully they're not too generic now!

(I left the jetpack.css as is because it doesn't seem to be loading on my test site so I'm not sure what the path should change to! It may be defunct now, though). 

See: 1200550061930446-as-1209168099234266

1. Apply this PR.
2. Turn on the Site Accelerator, and turn it on for static files. 
3. Enable Jetpack's share buttons and Perfmatters default settings.
4. Confirm that the share buttons are loading okay.
5. Turn off Site Accelerator for static files.
6. Confirm that the share buttons are still loading okay.

### Other information:

### DIFF
--- a/includes/plugins/class-perfmatters.php
+++ b/includes/plugins/class-perfmatters.php
@@ -85,8 +85,8 @@ class Perfmatters {
 			'plugins/newspack-newsletters', // Newspack Newsletters.
 			'plugins/newspack-plugin', // Newspack main plugin.
 			'plugins/newspack-popups', // Newspack Campaigns.
-			'plugins/jetpack/modules/sharedaddy', // Jetpack's share buttons.
-			'plugins/jetpack/_inc/social-logos', // Jetpack's social logos CSS.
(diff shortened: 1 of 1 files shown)


adekbadek (approved):

laurelfulford:
Thanks @adekbadek!

PR #3811, opened 2025-03-10
feat(media): enable setting a default image for 404 images

adekbadek:
### All Submissions:

Adds a feature to designate an image in the media library as a "404 Not Found" image. This image will be served when the asset is not found.

image

1. Insert an image in a post, then remove the image in the media library
2. Visit the front-end, observe the image element is broken
3. Visit the media library, edit an image and mark it as the default image using the "Is default image?" option
4. Reload the front-end, observe the default image is now loaded
5. Load the post editor, observe that the default image is used there, too

### Other information:

---
  - https://app.asana.com/0/0/1209600985817377

### DIFF
--- a/includes/class-default-image.php
+++ b/includes/class-default-image.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * Newspack News Revenue Hub feature management.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Handles Default Image functionality.
+ */
+class Default_Image {
+	/**
+	 * Form field name.
+	 */
+	const FIELD_NAME = 'is_default_image';
+
+	/**
+	 * Option name.
+	 */
+	const OPTION_NAME = 'newspack_default_image_url';
+
+	/**
+	 * Add hooks.
+	 */
+	public static function init() {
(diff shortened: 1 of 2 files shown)


includes/class-default-image.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * Newspack News Revenue Hub feature management.
```
leogermani: update comment

includes/class-default-image.php
```diff
@@ -0,0 +1,92 @@
...
+		add_filter( 'attachment_fields_to_save', [ __CLASS__, 'save_attachement_settings' ], 10, 2 );
+		add_filter( 'attachment_fields_to_edit', [ __CLASS__, 'add_attachment_field' ], 10, 2 );
+		add_action( 'template_redirect', [ __CLASS__, 'handle_not_found_image' ] );
+	}
+
+	/**
+	 * Handle not found image.
+	 */
+	public static function handle_not_found_image() {
+		if ( is_404() ) {
```
naxoc: Non-blocking nit: The conditionals here are nested pretty deep. We could return early and have less nesting
`! is_404() || empty( $_SERVER['REQUEST_URI'] )`

naxoc (approved):
I tested this and it's working great! 

I added a non-blocking nit.

includes/class-default-image.php
adekbadek: That's a great suggestions, thanks! Changed in 320e9f82f677

PR #3815, opened 2025-03-11
fix(ia): rerender subscription lists on esp provider update

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/1/26890605006346/project/1205919985867982/task/1209464932054166?focus=true

This PR fixes an issue where updating a provider does not always refresh the subscriptions lists associated with that provider until a page reload. This happened because subscriptions lists would rerender when the provider was updated, but before api credentials were stored and settings were saved. As a result updating the provider would cause the subscriptions lists to fetch using the outdated provider details.

We fix this by:
- Only rerendering subscriptions lists when provider updates AND lists are not locked
- Correctly locking lists depending on provider update conditions

Screenshot 2025-03-12 at 11 47 40

**Onboarding**
1. Reset newspack via the `reset newspack` link at the bottom of any wizard
Screenshot 2025-03-12 at 11 48 42
2. Navigate to Audience > Setup > Email Service Provider, and click the button to do the ESP step
3. Confirm the subscriptions lists section is not visible
4. Select an ESP provider and enter valid credentials
5. Confirm the subscriptions lists section appears and renders the expected lists

**Switching Providers**
1. Select a different provider
2. Confirm the subscriptions lists section renders a warning to save before editing subscriptions lists
3. Enter valid credentials for the new ESP
4. Confirm the subscriptions lists section renders the expected lists after loading

Bonus: Smoke test the ESP and Subscriptions Lists settings with invalid credentials, etc.

### Other information:

### DIFF
--- a/src/wizards/newsletters/views/settings/index.js
+++ b/src/wizards/newsletters/views/settings/index.js
@@ -52,8 +52,13 @@ export const Settings = ( {
 		const newProvider = newslettersConfig?.newspack_newsletters_service_provider || '';
 		if ( provider !== newProvider ) {
 			setError( false );
-			setLockedLists( !! provider );
 			setProvider( newProvider );
+			if ( config?.settings?.newspack_newsletters_service_provider?.onboarding ) {
+				// We are onboarding, so we should always lock the lists until saved.
+				setLockedLists( true );
+			} else {
+				setLockedLists( !! provider );
+			}
 		}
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Fixes the issue!

The whole `lockedLists` idea is not great (my bad) and the regex for provider key detection is a bit fragile. I don't think we should refactor this right now, though.

PR #3816, opened 2025-03-11
fix(ia): audience wizard without activation

miguelpeixe:
### All Submissions:

The audience wizard does not render anything other than the Audience Management setup when its not activated. Not all publishers will set it up but should still use the Salesforce integration and the "Checkout & Payment" tab.

This PR fixes that behavior and also tweaks the labels to render "Setup" in the menu item and tab only when deactivated. It should render "Configuration" once activated.

1. Deactivate audience management: `wp option delete newspack_reader_activation_enabled`
2. Visit the "Audience -> Setup" wizard and confirm the "Salesforce" section renders as well as the Checkout & Payment tab
3. Activate audience management and confirm the menu item and tab changed from "Setup" to "Configuration" and all the other components render

### Other information:

### DIFF
--- a/src/wizards/audience/views/setup/setup.js
+++ b/src/wizards/audience/views/setup/setup.js
@@ -179,212 +179,214 @@ export default withWizardScreen(
 						skipPrerequisite={ skipPrerequisite }
 					/>
 				) ) }
-			{ config.enabled && (
-				<Card noBorder>
-					<hr />
-					<SectionHeader
-						title={ __(
-							'Newsletter Subscription Lists',
-							'newspack-plugin'
-						) }
-					/>
-					<ActionCard
-						title={ __(
-							'Present newsletter signup after checkout and registration',
-							'newspack-plugin'
-						) }
-						description={ __(
-							'Ask readers to sign up for newsletters after creating an account or completing a purchase.',
-							'newspack-plugin'
-						) }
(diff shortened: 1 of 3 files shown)


src/wizards/audience/views/setup/setup.js
```diff
@@ -179,212 +179,214 @@ export default withWizardScreen(
...
-					/>
-					{ config.use_custom_lists && (
-						<SortableNewsletterListControl
-							lists={
-								newspackAudience.available_newsletter_lists
-							}
-							selected={ config.newsletter_lists }
-							onChange={ selected =>
-								updateConfig( 'newsletter_lists', selected )
+			<Card noBorder>
```
chickenn00dle: Nonblocking nit: It's possible for this `Card` component to have no children if `config.enabled` and `can_use_salesforce` are both false. Maybe we can nest each of these sections within their own `Card` components to avoid this case.

chickenn00dle (approved):
Works as expected. Left a small nitpick but its non-blocking so feel free to ignore.

src/wizards/audience/views/setup/setup.js
miguelpeixe: That looks a lot better! Thank you for the suggestion.

Updated in 00e18fb803a79067520f0cfe9379e6d476dd13e7

PR #3817, opened 2025-03-11
fix: making `get_post()` is not `null` before attempting to read props

eddiesshop:
This line was leading to many error log entries. It seems `get_post()` can be null at times. Perhaps this requires a deeper fix to investigate when this is being run? Should `get_post()` ever be returning `null`?

### All Submissions:

* [X] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?

Stops polluting the error logs.

1. Activate this plugin.
2. Tail the error log `tail -f /tmp/php-errrors`
3. Visit any number of posts/pages.
4. Notice the warnings: `PHP Warning:  Attempt to read property "post_author" on null in /wordpress/plugins/newspack-plugin/6.0.3/includes/data-events/connectors/ga4/class-ga4.php on line 189`
5. Checkout out this branch.
6. Reload the opened posts/pages.
7. Notice that the warnings are not thrown again.

### Other information:

### DIFF
--- a/includes/data-events/connectors/ga4/class-ga4.php
+++ b/includes/data-events/connectors/ga4/class-ga4.php
@@ -184,7 +184,7 @@ class GA4 {
 					get_coauthors()
 				)
 			);
-		} else {
+		} elseif ( null !== get_post() && is_numeric( get_post()->post_author ) ) {
 			// For some reason, get_the_author() does not work here.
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Thank you! I've refactored it a bit to avoid calling `get_post` three times.

PR #3819, opened 2025-03-12
fix: auth modal issues with content gating

adekbadek:
### All Submissions:

Fixes an issue with the "Continue" button being unresponsive when the hash-triggered auth modal was used.

1. On `trunk`,
1. Ensure RAS is configured
2. Create a membership plan that is restricted by registration, and set it to restrict all posts
3. Edit the content gate and insert a registration button there – a button and set the link to `#register_modal`
1. On a different page (which won't be content gated), also insert the registration button
4. On the frontend, visit that page and click the button. Insert and email and click "Continue"
5. After the registration completes, the modal will progress to the success state
6. Observe that the "Continue" button there does not work
7. Now visit a content-gate post – use the button to register and observe the same issue
8. Switch to this branch
9. Observe that on the success state, the modal "Continue" button:
    1. Closes the modal on the regular page
    2. Closes the modal and reloads the page – so the gate is lifted – on the restricted post

### Other information:

---
  - https://app.asana.com/0/0/1209599296873194

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -759,9 +759,8 @@ class Memberships {
 	/**
 	 * Render footer JS.
 	 *
-	 * If the gate was rendered, reload the page after 2 seconds in case RAS
-	 * detects a new reader. This allows the membership purchase to unlock the
-	 * content.
+	 * If the gate was rendered, reload the page after a new reader is detected.
+	 * This allows the membership purchase to unlock the content.
 	 */
 	public static function render_js() {
 		if ( ! self::$gate_rendered ) {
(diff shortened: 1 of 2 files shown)


includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -774,15 +773,18 @@ public static function render_js() {
 				ras.on( 'reader', function( ev ) {
 					if ( ev.detail.authenticated && ! window?.newspackReaderActivation?.getPendingCheckout() ) {
 						if ( ras.overlays.get().length ) {
+							// When an overlay is added or removed,
+							// check if there are none – this means an overlay was removed and there are none left.
+							// In this case, reload the window.
 							ras.on( 'overlay', function( ev ) {
-								if ( ! ev.detail.overlays.length && ! window?.newspackReaderActivation?.openNewslettersSignupModal ) {
```
adekbadek, line 778: @chickenn00dle – please advise if this condition was needed. It appears to be a constant condition (`openNewslettersSignupModal` is always defined), so the reload would never happen. I can't see the reason for this condition, and there's no comment to shed light on it.

adekbadek:
Closing in favour of a hotfix: https://github.com/Automattic/newspack-plugin/pull/3820

PR #3820, opened 2025-03-12
hotfix: auth modal issues with content gating

adekbadek:
### All Submissions:

Fixes an issue with the "Continue" button being unresponsive when the hash-triggered auth modal was used.

1. On `release`,
1. Ensure RAS is configured
2. Create a membership plan that is restricted by registration, and set it to restrict all posts
3. Edit the content gate and insert a registration button there – a button and set the link to `#register_modal`
1. On a different page (which won't be content gated), also insert the registration button
4. On the frontend, visit that page and click the button. Insert and email and click "Continue"
5. After the registration completes, the modal will progress to the success state
6. Observe that the "Continue" button there does not work
7. Now visit a content-gate post – use the button to register and observe the same issue
8. Switch to this branch
9. Observe that on the success state, the modal "Continue" button:
    1. Closes the modal on the regular page
    2. Closes the modal and reloads the page – so the gate is lifted – on the restricted post

### Other information:

---
  - https://app.asana.com/0/0/1209599296873194

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -759,9 +759,8 @@ class Memberships {
 	/**
 	 * Render footer JS.
 	 *
-	 * If the gate was rendered, reload the page after 2 seconds in case RAS
-	 * detects a new reader. This allows the membership purchase to unlock the
-	 * content.
+	 * If the gate was rendered, reload the page after a new reader is detected.
+	 * This allows the membership purchase to unlock the content.
 	 */
 	public static function render_js() {
 		if ( ! self::$gate_rendered ) {
@@ -774,15 +773,18 @@ class Memberships {
 				ras.on( 'reader', function( ev ) {
 					if ( ev.detail.authenticated && ! window?.newspackReaderActivation?.getPendingCheckout() ) {
 						if ( ras.overlays.get().length ) {
+							// When an overlay is added or removed,
+							// check if there are none – this means an overlay was removed and there are none left.
+							// In this case, reload the window.
 							ras.on( 'overlay', function( ev ) {
-								if ( ! ev.detail.overlays.length && ! window?.newspackReaderActivation?.openNewslettersSignupModal ) {
+								if ( ! ev.detail.overlays.length ) {
 									window.location.reload();
 								}
 							} );
 						} else {
 							setTimeout( function() {
 								window.location.reload();
-							}, 2000 );
+							}, 2000 ); // HACK: 2s delay to allow the purchase to be processed.
 						}
 					}
 				} );
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -770,22 +770,22 @@ class Memberships {
 		<script type="text/javascript">
 			window.newspackRAS = window.newspackRAS || [];
 			window.newspackRAS.push( function( ras ) {
+				let hasReader = false;
+				ras.on( 'overlay', function( ev ) {
+					// When an overlay was closed, and there's a reader,
+					// reload the window, but allow other JS – which might have
+					// triggered another overlay – to be executed (setTimeout hack).
+					if ( ! ras.overlays.get().length && hasReader ) {
+						setTimeout( () => {
+							if ( ! ras.overlays.get().length ) {
+								window.location.reload();
+							}
+						}, 1 )
+					}
+				})
 				ras.on( 'reader', function( ev ) {
 					if ( ev.detail.authenticated && ! window?.newspackReaderActivation?.getPendingCheckout() ) {
-						if ( ras.overlays.get().length ) {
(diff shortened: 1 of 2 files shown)


includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -774,15 +773,18 @@ public static function render_js() {
 				ras.on( 'reader', function( ev ) {
 					if ( ev.detail.authenticated && ! window?.newspackReaderActivation?.getPendingCheckout() ) {
 						if ( ras.overlays.get().length ) {
+							// When an overlay is added or removed,
+							// check if there are none – this means an overlay was removed and there are none left.
+							// In this case, reload the window.
 							ras.on( 'overlay', function( ev ) {
-								if ( ! ev.detail.overlays.length && ! window?.newspackReaderActivation?.openNewslettersSignupModal ) {
```
adekbadek, line 778: @chickenn00dle – please advise if this condition was needed. It appears to be a constant condition (openNewslettersSignupModal is always defined), so the reload would never happen. I can't see the reason for this condition, and there's no comment to shed light on it.

chickenn00dle: Hey @adekbadek,

This value will be set when the post checkout/registration newsletters modal is active. Unfortunately when transitioning from registration modal to newsletters modal, there is a moment when overlays are empty, so if we don't add this check, the page will reload before the newsletters modal has a chance to render.

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -774,15 +773,18 @@ public static function render_js() {
...
+								if ( ! ev.detail.overlays.length ) {
 									window.location.reload();
 								}
 							} );
 						} else {
 							setTimeout( function() {
 								window.location.reload();
-							}, 2000 );
+							}, 2000 ); // HACK: 2s delay to allow the purchase to be processed.
```
miguelpeixe: This is not a hack to allow the purchase to be processed. It's for UX, so the "Thank you" message at the end of the transaction is not just flashed to the reader before the page refresh.

adekbadek: That makes more sense, I've just rephrased the original function comment:

> reload the page after 2 seconds in case RAS detects a new reader. This allows the membership purchase to unlock the content.

I'll test and update the comment.

adekbadek: I'm missing something here. At the end of the modal checkout transaction there's a success state with a button (which [is rendered unconditionally](https://github.com/Automattic/newspack-blocks/blob/trunk/src/modal-checkout/templates/thankyou.php#L69-L71)), so it'll never flash since it's displayed until user closes it:

image

Which steps does the user have to take to arrive at the condition with the `setTimeout`?

includes/plugins/wc-memberships/class-memberships.php
adekbadek: > This value will be set when the post checkout/registration newsletters modal is active.

In my testing, this value is always set, so it's a constant condition. It's [included here](https://github.com/Automattic/newspack-plugin/blob/trunk/src/reader-activation/index.js#L416-L440) where `readerActivation` is defined, and then that value is [assigned to the global](https://github.com/Automattic/newspack-plugin/blob/trunk/src/reader-activation/index.js#L484). 

> when transitioning from registration modal to newsletters modal, there is a moment when overlays are empty

Got it, makes sense. I've updated the code to accommodate this condition in c4785979268c

includes/plugins/wc-memberships/class-memberships.php
adekbadek: I've removed this code for completeness sake, but once I have the reproduction steps for the case that this was supposed to handle, I'll work on accommodating it.

miguelpeixe: Right, that js predates RAS-ACC. We had no post-checkout modal, just a thank you message.

The logic in that script is agnostic to the checkout and makes sense. If authentication is detected and there's no overlay it should be ok to refresh the page just to unlock the viewing article.

miguelpeixe: You might be able to reproduce it by using the "Sign In" to authenticate to a member account while accessing a locked article. Publishers add a sign in link in their gates too.

adekbadek: That's the scenario I'm testing: 

1. visit a reg-walled article
2. use the "Sign In" (or manually placed button) to authenticate

But on `trunk`, the `if ( ras.overlays.get().length )` is always met, so the else-branch (with the 2s delay) is never executed.

Can you test the scenario you have on mind on this branch?

miguelpeixe: Ah, I see. In that case, I can only imagine it being useful for other opened tabs, which is a very edge case that I don't think justifies the additional logic.

miguelpeixe (approved):

PR #3822, opened 2025-03-12
fix(woocommerce): my-account redirect when resubscribing

adekbadek:
### All Submissions:

Fixes an over-eager check which causes resubscription requests to fail if site is using HPOS.

1. On `trunk`, 
2. Turn on HPOS  in WooCommerce Settings -> Advanced -> Features -> Order data storage ->  High-performance order storage
3. Log in as a user with a cancelled subscription, visit the My Account page, and try to resubscribe
4. Observe clicking the "resubscribe" button gets you to main My Account page, and the resubscription does not succeed
5. Switch to this branch, observe resubscription is possible

### Other information:


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -405,7 +405,7 @@ public static function is_user_verified() {
 	 */
 	public static function redirect_to_account_details() {
 		// phpcs:disable WordPress.Security.NonceVerification.Recommended
-		$is_resubscribe_request       = isset( $_REQUEST['resubscribe'] ) ? 'shop_subscription' === \get_post_type( absint( $_REQUEST['resubscribe'] ) ) : false;
+		$is_resubscribe_request       = isset( $_REQUEST['resubscribe'] ) ? true : false;
```
chickenn00dle: Nit: We don't need the ternary, `isset` should return true or false.

chickenn00dle (approved):
Left a non-blocking nit, but this is working as expected 👍

includes/reader-revenue/my-account/class-woocommerce-my-account.php
adekbadek: You're right, I was blindly following the convention set be the subsequent lines. Fixed in 3b1d8e2f99f9

PR #3824, opened 2025-03-12
feat: enable email change for newspack users

chickenn00dle:
### All Submissions:

Goes with https://github.com/Automattic/newspack-manager/pull/331

This PR enables the email change feature for all newspack users.

Screenshot 2025-03-12 at 15 44 25

Be sure to also checkout https://github.com/Automattic/newspack-manager/pull/331 before testing.

1. Log in as a reader with a non-a8c email address (i.e. email@test.com)
2. Go to My Account > Account details and verify the email input is disabled (You may need to verify the account first)
3. Repeat steps 1 and 2, this time with an a8c email address, and confirming the email input is not disabled

### Other information:

### DIFF
--- a/includes/reader-revenue/my-account/class-woocommerce-my-account.php
+++ b/includes/reader-revenue/my-account/class-woocommerce-my-account.php
@@ -795,8 +795,8 @@ class WooCommerce_My_Account {
 	 * Whether email changes are enabled.
 	 */
 	public static function is_email_change_enabled() {
-		if ( class_exists( '\Newspack_Manager\Features' ) ) {
-			return \Newspack_Manager\Features::is_newspack_user();
+		if ( class_exists( '\Newspack_Manager\Features' ) && \Newspack_Manager\Features::is_automattician() ) {
+			return true;
 		}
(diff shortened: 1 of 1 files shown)


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -795,6 +795,9 @@ public static function update_payment_methods_for_all_subs() {
 	 * Whether email changes are enabled.
 	 */
 	public static function is_email_change_enabled() {
+		if ( class_exists( '\Newspack_Manager\Features' ) ) {
+			return \Newspack_Manager\Features::is_newspack_user();
```
leogermani: if the user is not a "newspack_user" we still want to enable the feature if the feature flag is on. Need to return here only if it's `true`, otherwise should check the constant

chickenn00dle: Ah. Good catch! Fixed in https://github.com/Automattic/newspack-plugin/pull/3824/commits/c5b53aa17edf871ba7ce37f50a361461d9099468

leogermani (approved):

leogermani:
Cherry picked to alpha

PR #3825, opened 2025-03-12
Revert "refactor(corrections): remove corrections feature flag (#3797)"

leogermani:
This reverts commit 777695726a57ee4c29e42dfb27814e2b7a4eedc9.

We'll need to change something on this feature and therefore we are keeping the feature flag.

This will need to be cherry picked to alphas once it's merged to trunk

== Testing ==

Confirm that the corrections feature is only available if the `NEWSPACK_CORRECTIONS_ENABLED` constant is set to true


naxoc (approved):
Tested and works 👍

leogermani:
Cherry picked to alpha

PR #3828, opened 2025-03-13
Fix: Guest Contributors Role Init for CLI and PHPUnit

ronchambers:
### All Submissions:

If a new WordPress website is created using only the CLI or PHPUnit without any browser navigation to `/wp-admin/`, then `admin_init` isn't run and the custom Guest Contributors role will not get created.  To avoid this, the role creation should happen on `init`.  This will make sure the role exists for `/wp-admin/`, CLI, and PHPUnit (also if newpack plugin is included as a dependency in a secondary repo's PHPUnit).

Without this change, do a `n sites-add` with Newspack Docker so that a new site will be created via CLI.  Do not open the site in the browser.  Verify the role does not exist by running `wp role list`

Now, pull the change and do the same steps on a new `n sites-add` and you'll see `wp role list` will show the Guest Contributor role.

Also, the PHPUnit test has been updated to remove `set_up_before_class` since it is no longer needed because `init` will run for PHPUnit.

### Other information:

### DIFF
--- a/includes/plugins/co-authors-plus/class-guest-contributor-role.php
+++ b/includes/plugins/co-authors-plus/class-guest-contributor-role.php
@@ -50,16 +50,11 @@ class Guest_Contributor_Role {
 	 */
 	public static function initialize() {
 		add_filter( 'coauthors_edit_author_cap', [ __CLASS__, 'coauthors_edit_author_cap' ] );
-
-		// Setup the custom role directly after global $wp_roles is initialized in wp-settings.php.
-		// Do not use 'admin_init' as this will not account for CLI commands where the admin is not initialized.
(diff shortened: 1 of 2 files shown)


ronchambers:
@leogermani - hopefully you can give this a look soon :-)  Let me know if you have questions or concerns.  Thanks!

leogermani (approved):

ronchambers:
Hi @adekbadek - you helped me in the past with the semantic release stuff, I think the last commit on this PR follows the correct pattern `fix: ...`.  But I have a question in regards to the `needs-changelog` label that was added.  I don't think this PR needs any special change log comments.  It's just an internal code fix.  Should I just remove the label? (cc @leogermani ). Thanks!

leogermani:
Hi @ronchambers We can handle that. This is a pretty minor change so I'll just remove the label

PR #3829, opened 2025-03-13
feat(corrections-location): integrate Correction location logic

Takshil-Kunadia:
### All Submissions:

Closes https://app.asana.com/1/26890605006346/project/1209292256643614/task/1209635079338293.

1. Create a Test Post > Manage Corrections.
2. Check Manage Corrections modal should have a new `LOCATION` setting. Default: `Bottom`.
3. Add at some corrections with the location set to "top" and others with the location set to "bottom".
4. Verify that each correction’s meta field for location is correctly saved. After reloading few times.
5. View the post on the front-end and confirm that corrections with the "top" location are displayed above the post content and corrections with the "bottom" location are displayed below the post content.
6. Ensure that if no corrections exist for a given location, no corrections in that location is rendered.

### Next Steps:-
- Refactor Corrections blocks to work with the location setting.

### Other information:

### DIFF
--- a/includes/corrections/README.md
+++ b/includes/corrections/README.md
@@ -18,17 +18,18 @@ define( 'NEWSPACK_CORRECTIONS_ENABLED', true );
 
 Corrections are stored as `newspack_correction` custom post type. A correction consists of the following fields:
 
-| Name                          | Type     | Stored As      | Description                                                                    |
-| ----------------------------- | -------- | -------------- | ------------------------------------------------------------------------------ |
-| `title`                       | `string` | `post_title`   | The correction title. Defaults to 'Correction for [post title]'                |
-| `content`                     | `string` | `post_content` | The correction text.                                                           |
-| `date`                        | `string` | `post_date`    | The date assigned to the correction.                                           |
-| `newspack_correction-post-id` | `int`    | `post_meta`    | The ID of the post to which the correction is associated.                      |
-| `newspack_corrections_type`   | `string` | `post_meta`    | Whether it's a correction or a clarification (`correction` or `clarification`) |
+| Name                            | Type     | Stored As      | Description                                                                    |
+| ------------------------------- | -------- | -------------- | ------------------------------------------------------------------------------ |
+| `title`                         | `string` | `post_title`   | The correction title. Defaults to 'Correction for [post title]'                |
+| `content`                       | `string` | `post_content` | The correction text.                                                           |
+| `date`                          | `string` | `post_date`    | The date assigned to the correction.                                           |
+| `newspack_correction-post-id`   | `int`    | `post_meta`    | The ID of the post to which the correction is associated.                      |
+| `newspack_corrections_type`     | `string` | `post_meta`    | Whether it's a correction or a clarification (`correction` or `clarification`) |
+| `newspack_corrections_location` | `string` | `post_meta`    | Where the correction should be displayed. (`top` or `bottom`)                  |
 
 In addition, some correction data is stored in the associated post as post meta:
 
 | Name                            | Type                    | Stored As   | Description                                                   |
 | ------------------------------- | ----------------------- | ----------- | ------------------------------------------------------------- |
 | `newspack_corrections_active`   | `bool`                  | `post_meta` | Whether the feature is enabled for the post.                  |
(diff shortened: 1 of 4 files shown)


leogermani (approved):
I did one small tweak to change "Close & save" to "Save & close"

leogermani:
One more thing. I just updated the docs -> https://github.com/Automattic/newspack-plugin/pull/3829/commits/85768ab3ca844695f56adf840df4df45369f2550

Also, @Takshil-Kunadia , I was looking at the `CORRECTIONS_ACTIVE_META`. Do we still use it? I don't think there's a UI for that and, in the code, I only see us checking for it, but I don't see us writing this value... I think we can remove it? WDYT?

leogermani (changes requested):
Removing approval while we discuss my last comment about the "is active" post meta

Takshil-Kunadia:
@leogermani I too had this [question before](https://github.com/Automattic/newspack-plugin/pull/3751#:~:text=Would%20you%20like%20me%20to%20add%20a%20UI%20option%20to%20enable/disable%20corrections%20for%20individual%20posts%3F).

Basically this can be used as a flag for cases where Editor wants to keep the corrections saved(exist/persist) for that post but turn off displaying it in the front-end for a particular post as otherwise current logic will display corrections for the posts which have corrections data. If this is a use case...otherwise it makes sense to remove it.

thomasguillot:
Design-wise, I haven't much to say 👍

leogermani:
> Basically this can be used as a flag for cases where Editor wants to keep the corrections saved(exist/persist) for that post but turn off displaying it in the front-end for a particular post as otherwise current logic will display corrections for the posts which have corrections data. If this is a use case...otherwise it makes sense to remove it.

We need to either remove it from the code or to add the UI to enable/disable it.

I think we can remove it, but then we need to change how we are handling the shortcode there.. because we are filtering by posts with this feature enabled. If we remove that we are going to quey all posts and we don't want that

Takshil-Kunadia:
> I think we can remove it, but then we need to change how we are handling the shortcode there.. because we are filtering by posts with this feature enabled. If we remove that we are going to quey all posts and we don't want that

@leogermani I don't think this should be an issue, right? we are already querying all the corrections of all posts with the get_posts(), filtering is done with the meta key but since we haven't provided UI I don't imagine it being set by anyone via a manual DB update.
moreover, eliminating meta query should be better for query instead.

leogermani:
I'm talking about this query -> https://github.com/Automattic/newspack-plugin/blob/9880b5ec3d7c3d10123a6d90e40229a2268425c5/includes/corrections/class-corrections.php#L396

We need to change that. This will end up querying for ALL posts in the database and then fetching the corrections for each of them

Takshil-Kunadia:
@leogermani ~~Ahhh! Missed that there is no post-type check in that query. makes sense. will add that and remove the meta query. That should do it.~~

Gotcha! hmmm that would be a tad bit difficult! We can do a DB query to find unique post_id meta values from all the corrections posts. If its alright!

Post ID is saved in each correction 
https://github.com/Automattic/newspack-plugin/blob/9880b5ec3d7c3d10123a6d90e40229a2268425c5/includes/corrections/class-corrections.php#L27

Takshil-Kunadia:
@leogermani DB query to reverse search the post IDs of the corrections.
```php
$post_ids = $wpdb->get_col(
	$wpdb->prepare(
		"SELECT DISTINCT meta_value FROM $wpdb->postmeta WHERE meta_key = %s",
		self::CORRECTION_POST_ID_META
	)
);
```

Takshil-Kunadia:
@leogermani Please have a look at https://github.com/Automattic/newspack-plugin/pull/3829/commits/e8898f665032c84c7a3c79eec804a78667207c24

leogermani (approved):
@Takshil-Kunadia I actually had a closer look and made a call to remove the shortcode. I think this came from the old plugin we used as a base, but we don't need. That looked like a ticking bomb because it was querying all corrections from all posts... we don't need that anywhere now that we have a proper archive in the theme and the blocks for block themes

Takshil-Kunadia:
Thanks! @leogermani that takes care of it! 🙌

PR #3831, opened 2025-03-13
Add exception to Admin view for Multisite installations

dariusk:
### All Submissions:

* [X] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?

On a Multisite Network installation, local Admins (as opposed to "Network" AKA "Super" Admins) [have reduced capabilities compared to the Admin role on standard installations](https://developer.wordpress.org/advanced-administration/multisite/administration/#user-access-capabilities). Specifically they can't install new plugins.

For Multisite users viewing the Admin page but who do not have Network Admin privileges, this commit skips over some `add_filter` calls that were causing errors on the plugin listing page (since the Admin did not actually have those capabilities).

This will still work correctly for a Network Admin with the plugin installation privileges since the `is_network_admin()` will catch that and render the correct available actions.

(co-authored by @jsdiaz)

1. Install Newspack in a Multisite Network environment
2. View the plugins page as a regular Admin user (not a Network Admin)
3. Observe php errors at the top of the page
4. Apply this patch
5. View the plugins page again as the same user
6. Observe no errors

### Other information:

### DIFF
--- a/includes/class-admin-plugins-screen.php
+++ b/includes/class-admin-plugins-screen.php
@@ -18,8 +18,10 @@ class Admin_Plugins_Screen {
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_filter( 'all_plugins', [ $this, 'inject_managed_plugins' ] );
-		add_filter( 'plugin_action_links', [ $this, 'modify_action_links' ], 10, 3 );
+		if ( ! is_multisite() || is_network_admin() ) {
+			add_filter( 'all_plugins', [ $this, 'inject_managed_plugins' ] );
+			add_filter( 'plugin_action_links', [ $this, 'modify_action_links' ], 10, 3 );
+		}
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_scripts_and_styles' ] );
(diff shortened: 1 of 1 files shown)


includes/class-admin-plugins-screen.php
```diff
@@ -18,8 +18,10 @@ class Admin_Plugins_Screen {
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_filter( 'all_plugins', [ $this, 'inject_managed_plugins' ] );
-		add_filter( 'plugin_action_links', [ $this, 'modify_action_links' ], 10, 3 );
+		if ( ! is_multisite() || is_network_admin() ) {
```
adekbadek, line 21: While it works, it's a somewhat roundabout way – it'd be more readable to add a `current_user_can( 'install_plugins' )` in both methods (the check won't work in the constructor. Like so: https://github.com/Automattic/newspack-plugin/commit/dd8946e2feaecddf3c98f4fa7ba84fa1cee5ed5d. Would that still work for your use case @dariusk ?

PR #3837, opened 2025-03-17
fix(newsletters): display correct premium lists in modal

chickenn00dle:
### All Submissions:

See https://newspackp2.wordpress.com/2025/03/10/upcoming-changes-march-17-2025/#comment-13540

This PR fixes an issue where the newsletters modal was incorrectly showing all premium newsletters. We fix this by specifying product ids that should "unlock" the newsletter in the modal before making it visible:

Screenshot 2025-03-17 at 12 45 54

**Requires https://github.com/Automattic/newspack-newsletters/pull/1782 and https://github.com/Automattic/newspack-blocks/pull/2075 to test.**

#### Setup
1. On a test site, set up three Subscription Lists via Newspack > Engagement > Newsletters
2. Set up two subscription products (non-donation) and add two checkout buttons to any page or post for these products
4. Set up two memberships tied to the two subscription products from the previous step, and restricting two of the three subscription lists from the first step, one per membership
5. Set up a third subscription product, not associated with any membership or subscription list and add a third checkout button to the same page or post
6. Enable the post-checkout/auth newsletter modal and add all three subscription lists via Newspack > Engagement > Show Advanced Settings > Newsletter Subscription Lists

#### Testing
1. As a reader, purchase the subscription product not restricting any subscription lists
2. Confirm the newsletters modal appears post-checkout only displaying one of the three subscription lists. Don't signup.
3. As the reader, purchase one of the subscription products restricting a subscription list.
4. Confirm the newsletters modal appears post-checkout displaying:
  - The non-restricted list
  - The restricted list associated with the product you just purchased
5. As the reader, purchase the final subscription product
6. Confirm all three lists appear in the newsletters modal. None should be restricted since you've purchased all products

Bonus: Smoke test the newsletters form post-auth form.

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -1087,5 +1087,24 @@ class Memberships {
 
 		return (int) reset( $user_subscriptions );
 	}
+
+	/**
+	 * Get product Ids for the given membership plan.
+	 *
+	 * @param int $plan_id Membership plan ID.
+	 *
+	 * @return array Product IDs.
+	 */
+	public static function get_product_ids_for_membership_plan( $plan_id ) {
+		if ( ! function_exists( 'wc_memberships_get_membership_plan' ) ) {
+			return [];
+		}
+		$product_ids = [];
+		$plan        = \wc_memberships_get_membership_plan( $plan_id );
+		if ( $plan ) {
+			$product_ids = $plan->get_product_ids();
+		}
+		return $product_ids;
+	}
 }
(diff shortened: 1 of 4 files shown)


leogermani:
As much as we can I think the filtering of the lists that will show up on the modal should happen in the newspack plugin, on the memberships integration there... having it here starts to spread the "premium newsletters" functionality across too many places and it will become harder to maintain

chickenn00dle:
We decided not to go with this approach. Closing.

PR #3839, opened 2025-03-17
fix(ia): refresh on audience setup complete

miguelpeixe:
### All Submissions:

1205919985867982-as-1209464932054174/f

After the audience setup, a full page refresh is required. The current approach runs only a `window.location.replace()` which will not do a full page load if the replacement is just a hash change.

This PR forces a reload after the location is replaced.

1. Go through a fresh Audience setup flow
2. Confirm you're redirected to the audience home wizard on completion and you get the following notice along with all the tabs unlocked:

image

### Other information:


chickenn00dle (changes requested):
This fixes the issue, but now this introduces some strangeness in the UI. After completing setup, the page appears to reload twice as a result of calling replace and then reload:

https://github.com/user-attachments/assets/3a0961b1-b1d6-4690-b9e0-7e96cce68c34

Maybe we can replace both of these with `window.location.href = url`?

miguelpeixe:
> Maybe we can replace both of these with window.location.href = url?

That will have the same behavior as `.replace()` for hash changes.

How about 7318b600d30de224b0e2d4a860ce31931eb7e716?

chickenn00dle (changes requested):
> How about https://github.com/Automattic/newspack-plugin/commit/7318b600d30de224b0e2d4a860ce31931eb7e716?

Thanks for clarifying! I am still seeing this strange double reload in the UI with this change:

https://github.com/user-attachments/assets/623ceb6c-7a62-4776-8dc8-48dca86cfba0

It would be great if we could remain in the completed card until the reload triggers.

miguelpeixe:
Hmm, that's strange. I can't reproduce it. Which browser is that?

chickenn00dle:
This is on Firefox 136.0.2

PR #3842, opened 2025-03-17
fix: dont sync info about failed orders

leogermani:
### All Submissions:

This PR fixes the Last Payment Date meta field.

The call for `$current_subscription->get_date( 'last_order_date_paid', 'site' );` do not filter by completed orders, so it can look at a Failed order. In those cases, the date_paid will be zero, because the order was not paid.

Later in the code, id date_paid is zero we are replacing it with the current data, which makes no sense.

This PR makes sure to fetch this info from the last successful order. The same order from which we take the last payment amount.

Note: There has been reports of the Last Payment Amount also being wrong, but I could not reproduce it...

Testing:

Not sure how to reproduce this state, but you need a subscription that has a successful parent order and optionally a few other, and then one last failed order.

Run 

1. on `release` run `Newspack\Reader_Activation\Sync\WooCommerce::get_contact_from_customer($id)` for the user ID
2. Confirm that the date will return the current date
3. Checkout this PR and confirm the date will be the date of the last successful order

### Other information:

---
  - https://app.asana.com/0/0/1209680595023887

### DIFF
--- a/includes/reader-activation/sync/class-woocommerce.php
+++ b/includes/reader-activation/sync/class-woocommerce.php
@@ -182,13 +182,13 @@ class WooCommerce {
 	}
 
 	/**
-	 * Get the amount of the last payment associated with the given subscription.
+	 * Get the successful order associated with the given subscription.
 	 *
 	 * @param \WC_Subscription $subscription Subscription object.
 	 *
-	 * @return float The amount of the last payment.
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
I've tested with a subscription that had a last payment in pending state. Worked as described.

PR #3850, opened 2025-03-21
fix: add check if product before gating content

laurelfulford:
### All Submissions:

This PR makes sure that products that are included in memberships -- either as limited to members or that have discounts limited to members -- do not get a content gateway.

It also overrides `user_has_cap()` for products; this function was causing all restricted products to return the message `This product is part of your membership, but not yet! It will become available on {date}.`, the message for `product_access_delayed_message` [when a member will get access to a product on a certain date], instead of the one tailored to what's actually happening (like a product being fully restricted to a membership). 

Note: for expediency, you can test https://github.com/Automattic/newspack-theme/pull/2465 at the same time as this one -- it also makes sure the correct membership messaging doesn't look weird!  

See: https://app.asana.com/0/1200550061930446/1209731811855442/f

1. Set up two products, and make sure to populate the Product Description and Product Short Description. This just makes it easier to see the full issue.
2. Set up a Woo Membership Plan.
3. Under 'Restrict Content', restrict all posts.
4. Under 'Restrict Products', restrict the purchase of at least one product to members only.
5. Under 'Purchasing Discounts', discount at least one specific product. Make sure to set a discount amount, and check the 'Active' checkbox at the end (no comment about how long it took me to figure this out). 
6. Under Newspack > Engagement > Show Advanced Settings, enable and set up the content gate for your membership.
7. In an incognito window, click around posts on your site and confirm that you get the membership gateway.
8. Go to /shop and navigate to view the two products restricted by your membership:

   * Your Restricted Product should show no message, but it will be missing a button to add to cart.
   * Your Purchasing Discount product will show the Short Description, buttons, then full description followed by the Content Gate:

CleanShot 2025-03-21 at 14 05 56@2x

9. Apply this PR.
10. Repeat steps 7 and 8 to confirm that the content gate still works, but is not applied to products. 
   * Your Restricted Product should show a message about the restriction:

CleanShot 2025-03-21 at 16 22 05@2x

   * Your Purchasing Discount product should not show the gate, and instead show a message (this screenshot has [this PR applied](https://github.com/Automattic/newspack-theme/pull/2465)).

CleanShot 2025-03-21 at 14 08 42@2x

11. Smoke test [the testing steps here](https://github.com/Automattic/newspack-plugin/pull/2623) to make sure the changes to the `user_has_cap()` function doesn't cause issues.

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -659,6 +659,7 @@ class Memberships {
 		if ( get_queried_object_id() !== get_the_ID() ) {
 			return '';
 		}
+
 		self::$gate_rendered = true;
 		return self::get_inline_gate_content();
 	}
@@ -854,7 +855,7 @@ class Memberships {
 	 */
 	public static function user_has_cap( $all_caps, $caps, $args ) {
 		// Bail if Woo Memberships is not active.
-		if ( ! self::is_active() ) {
+		if ( ! self::is_active() || is_product() ) {
 			return $all_caps;
 		}
 
@@ -944,6 +945,7 @@ class Memberships {
 		$has_subscription  = false;
 
 		foreach ( $rules as $rule ) {
+
 			$membership_plan_id = $rule->get_membership_plan_id();
 			$has_subscription   = ! empty( self::get_user_subscription_for_membership_plan( $user_id, $membership_plan_id ) );
 
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -659,7 +659,6 @@ class Memberships {
 		if ( get_queried_object_id() !== get_the_ID() ) {
 			return '';
 		}
-
 		self::$gate_rendered = true;
 		return self::get_inline_gate_content();
 	}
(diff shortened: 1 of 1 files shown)


laurelfulford:
@dkoo I'd love to get your feedback on the change to `user_has_cap()`, just to make sure I haven't done something breaking -- thank you!

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -648,7 +648,7 @@ public static function get_inline_gate_content() {
 	 */
 	public static function wc_memberships_notice_html( $notice, $message_body, $message_code, $message_args ) {
 		// If the gate is not available, don't mess with the notice.
-		if ( ! self::has_gate() ) {
+		if ( ! self::has_gate() || is_product() ) {
```
adekbadek, line 652: Nit: normally, the membership notice would not be displayed on product pages – only if the plan has a discount. A comment here about this quirkiness might be worthwhile, so the change is not deleted in the future due to purported redundancy.

adekbadek (approved):
Works as described!

includes/plugins/wc-memberships/class-memberships.php
laurelfulford: Ah good call -- thanks @adekbadek! Comments have been added.

laurelfulford:
Thanks for the review, @adekbadek!

PR #3858, opened 2025-03-24
fix: move optional module settings from Syndication to own class

dkoo:
### All Submissions:

#3783 introduced another optional module, but one that's not related to the Syndication section in our dashboard. Since #3783 expects the optional module settings to be available in the `Newspack\Settings` class, which has been deprecated with its methods moved to other classes throughout the wizard app, it's currently throwing a fatal.

This PR moves all methods and constants related to optional modules to a new `Optional_Modules` class for clarity.

1. On `epic/ia`, observe a fatal error when loading any WP admin page.
2. Check out this branch and confirm the fatal no longer occurs.
3. Confirm that the settings in the **Newspack > Settings > Syndication** page display and update as expected.
4. Confirm that testing instructions in #3783 work as expected.

### Other information:


tests/unit-tests/syndication.php
```diff
@@ -5,6 +5,7 @@
  * @package Newspack\Tests
```
adekbadek, line 5: Nit: this test file is now more about `Optional_Modules`, so the file name could be updated.

adekbadek (approved):
Changes look and work ok, though I don't see the fatal on `epic/ia`.

miguelpeixe:
@dkoo is there anything blocking this merge? We should get IA to alphas.

tests/unit-tests/syndication.php
dkoo: Good idea, updated in 70612a4c

PR #3861, opened 2025-03-26
feat(woocommerce): add metadata to Stripe txns

adekbadek:
### All Submissions:

Adds Stripe metadata to transactions processed via `woocommerce-gateway-stripe` plugin:

- Product - product name from WC
- Transaction Type - Donation/Subscription/Subscription Renewal
- Membership Type - the WCM plan name
- subscription_id -the ID
- subscription_status - renewed or created → this is redundant with "Transaction Type", for legacy reasons – the code here is adapted from a custom plugin which was active on one site

1. Connect the Stripe payment gateway in WooCommerce
2. Set up a membership plan to which the access is granted upon a purchase
3. As a reader, purchase this item
4. In Stripe, observe that the transaction has the following metadata: `Product`, `Transaction Type`, `Membership Type` with relevant values
5. If the item is a subscriptions, also `subscription_id` and `subscription_status` should be added

### Other information:

---
  - https://app.asana.com/0/0/1209367423001040


includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		$order_item = array_values( $order->get_items() )[0];
+		if ( ! $order_item ) {
+			return $metadata;
+		}
+		$product_id = $order_item->get_product_id();
+
+		// Product name.
+		$metadata[ __( 'Product', 'newspack-plugin' ) ] = $order_item->get_name();
```
leogermani: I know we are migrating this existing feature from another repo... but I'm not sure it's a good idea to internationalize the metadata key... 

* a wrong translation can bring unsupported characters or strings that are too long
* If they change the language they'll have data in both formats
* harder to provide support and identify which field is which
* Everything in Stripe is in english anyways, isnt it?

adekbadek: Huh, good call! Updated

adekbadek:
@leogermani – I've updated the `Transaction Type` setting to account for non-donation purchases. In such case, the value will be `Regular Purchase`.

leogermani (commented):
Not sure if I did something wrong, but I'm not seeing any of theme meta in Stripe. Here's where I'm looking:

image

This is a payment that bought a subscription that has a membership tied to it

leogermani (approved):
Doesn't work on my local but was able to confirm it's working on other sites

adekbadek:
Closing in favor of #3903 – to release as a hotfix.

PR #3863, opened 2025-03-26
feat: update custom bylines data structure

leogermani:
### All Submissions:

Implements the new data structure that is in the works in https://github.com/Automattic/newspack-plugin/pull/3746 into trunk, so we can run and QA data migrations.

1. See https://github.com/Automattic/newspack-plugin/pull/3667
2. Run the same tests, but using the new data structure with `[...]` instead of `<...>`

### Other information:

### DIFF
--- a/src/bylines/index.js
+++ b/src/bylines/index.js
@@ -35,9 +35,19 @@ const BylinesSettingsPanel = () => {
 	}
 	// Byline change handler.
 	const handleBylineChange = value => {
-		const tags = value.match( /<[^>]+>/g );
-		if ( tags && tags.some( tag => ! tag.startsWith( '<Author' ) && ! tag.startsWith( '</Author' ) ) ) {
-			alert( __( 'Only the <Author> tag is allowed.', 'newspack-plugin' ) ); // eslint-disable-line no-alert
+		const tags = value.match( /\[[^\]]+\]/g );
+		if (
+			tags &&
+			tags.some(
+				tag =>
+					! tag.startsWith( '[Author' ) &&
+					! tag.startsWith( '[/Author' )
+			)
+		) {
+			// eslint-disable-next-line no-alert
+			alert(
+				__( 'Only the [Author] tag is allowed.', 'newspack-plugin' )
+			);
 			return;
 		}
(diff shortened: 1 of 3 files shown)


src/bylines/index.js
```diff
@@ -35,9 +35,19 @@ const BylinesSettingsPanel = () => {
 	}
 	// Byline change handler.
 	const handleBylineChange = value => {
-		const tags = value.match( /<[^>]+>/g );
-		if ( tags && tags.some( tag => ! tag.startsWith( '<Author' ) && ! tag.startsWith( '</Author' ) ) ) {
-			alert( __( 'Only the <Author> tag is allowed.', 'newspack-plugin' ) ); // eslint-disable-line no-alert
+		const tags = value.match( /\[[^\]]+\]/g );
```
chickenn00dle, line 38: Non-blocking: Looking at this again after a while, this could use a code comment to clarify what we are searching for.

src/bylines/index.js
```diff
@@ -53,10 +63,13 @@ const BylinesSettingsPanel = () => {
 				contentEl.insertBefore( bylineEl, contentEl.firstChild );
 			}
 			// If there are author tags
-			if ( /<Author id=(\d+)>/.test( text ) ) {
-				text = text.replace( /<Author id=(\d+)>([^<]+)<\/Author>/g, ( match, authorId, authorName ) => {
-					return `<a href="${ newspackBylines.siteUrl }/?author=${ authorId }">${ authorName }</a>`;
-				} );
+			if ( /\[Author id=(\d+)\]/.test( text ) ) {
+				text = text.replace(
+					/\[Author id=(\d+)\]([^\[]+)\[\/Author\]/g,
```
chickenn00dle, line 68: Non-blocking: Similar to above, this could use a code comment to clarify what we are doing here.

chickenn00dle (approved):
Looks good!

Left some comments that you can feel free to ignore.

src/bylines/index.js
leogermani: This is all being completely rewritten in https://github.com/Automattic/newspack-plugin/pull/3746

chickenn00dle: Gotcha. Perfect!

PR #3874, opened 2025-03-31
chore: prevent PHP notice with is_singular

adekbadek:
### All Submissions:

Adds a check to prevent a PHP notice.

1. On `trunk`, with WooCommerce Memberships active, visit any page 
2. Observe a 

```
PHP Notice:  Function is_singular was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run. Before then, they always return false. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 3.1.0.) in /var/www/html/wp-includes/functions.php on line 6121
```

notice logged

4. Switch to this branch, reload, observe no notice logged

### Other information:


includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -855,6 +855,11 @@ private static function can_manage_woocommerce( $caps ) {
 	 * @return array Filtered capabilities.
 	 */
 	public static function user_has_cap( $all_caps, $caps, $args ) {
+		global $wp_query;
```
naxoc: I wonder if `! did_action('wp')` would be more accurate here? To know if conditional tags are available?

adekbadek: Gets the job done in a more elegant way, thanks!

naxoc (approved):
Looks good. Thanks for fixing this one - it was noisy!

PR #3878, opened 2025-03-31
Fix: remove design link

laurelfulford:
### All Submissions:

This PR removes the new Appearance > Design menu item added in WordPress 6.8.

This is an update to the Appearance > Patterns menu added in WordPress 6.6; the Design menu now has the Patterns menu item underneath it, plus an additional Style link.

When the Patterns link was originally added, [we hid it](https://github.com/Automattic/newspack-plugin/pull/3236) because we'd already added a link to a WP Admin-style page under Posts, and worried it would cause confusion (the new page also used the Site Editor, which is new to our pubs and could cause a support burden). 

With the change in 6.8, I think the risk of confusion is lower, but I also ran into a couple Newspack-specific bugs when testing -- for example, campaign popups are triggered on some of the screens under Patterns but their close buttons don't work there. And the Styles view isn't entirely accurate (it's not pulling in our theme styles consistently), and it occasionally throws an editor error. I'd like to dig into both of these, but for timing, let's hide the new link until I have time to test. 

Once WordPress 6.8 is released, the original code hiding just the Appearance > Patterns link can be removed. 

1. Apply this PR on a site running the latest WP 6.8 RC, and confirm you don't have a WP Admin > Appearance > Design link.
2. Apply this PR on a site running the latest WP release (6.7.2), and confirm you don't have a WP Admin > Appearance > Patterns link.

### Other information:

### DIFF
--- a/includes/class-patches.php
+++ b/includes/class-patches.php
@@ -20,7 +20,7 @@ class Patches {
 		add_filter( 'wpseo_enhanced_slack_data', [ __CLASS__, 'use_cap_for_slack_preview' ] );
 		add_action( 'admin_menu', [ __CLASS__, 'add_patterns_menu_link' ] );
 		add_action( 'admin_menu', [ __CLASS__, 'add_pattern_categories_menu_link' ] );
-		add_action( 'admin_menu', [ __CLASS__, 'remove_core_patterns_menu_link' ] );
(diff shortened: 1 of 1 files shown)


adekbadek (approved):

laurelfulford:
Thanks @adekbadek!

PR #3358, opened 2024-08-22
Media Kit page handling

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

1. Removes references to the `publisher-media-kit` plugin, which will now be included in `newspack-ads` plugin.
2. Change the Media Kit page handling in the Ads Wizard -> Add Ons tab

image

The three states of the UI:

image

### How to test the changes in this Pull Request:

1. Switch `newspack-ads` to `feat/expose-media-kit` branch (https://github.com/Automattic/newspack-ads/pull/908)
1. Visit the Advertising Wizard -> Add Ons tab
2. Observe the UI acts as described in the comment above.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3881, opened 2025-04-02
fix: restore Media Kit components in Ads wizard

dkoo:
### All Submissions:

Hotfix to add some IA reorganization that was missed.

- Restores the Media Kit settings component in the Advertising > Display Ads > Settings wizard page (behavior should remain as described in #3358)
- Moves the Suppression settings component from its own view to the Advertising > Display Ads > Settings wizard page
- Changes the visual style of the Suppression settings section to more closely match other settings components on the page (not in the design mockup)

Note: I did not move the Super Cool Ad Inserter plugin settings to the "Plugins" settings in this PR, as is outlined in the design mockup. These settings are more tightly integrated into the Newspack Ads plugin and moving them would require a more extensive refactor that doesn't fit the urgency of this issue. cc @thomasguillot if we'd like to flag this for a future refactor.

Screenshot 2025-04-02 at 17-20-53 Advertising ‹ dkoo — WordPress

1. Check out this branch
2. In WP admin, visit **Advertising > Display Ads > Settings**
3. Confirm that Suppression settings appear in this page, as well as Ad Refresh Control and Media Kit settings under a "Plugins" heading
4. Confirm that settings in all sections can be updated and saved and that updates persist after reloading the page

### Other information:

---
  - https://app.asana.com/0/0/1209874293448116


thomasguillot:
We need to have a "Plugins" tab because why would the page has heading that says "Settings" and then we scroll down and there's another heading, this time that says "Plugins".

laurelfulford (approved):
Thanks for tackling this, @dkoo! Everything saves as expected, and TIL Ad Refresh Control was a separate plugin? 

The only thing I noticed was that the Suppression section isn't 1:1 with how it looks in the mockup (with the header, and without the border), but I think that could be tackled in a second, less urgent iteration.

I'm marking as approved from a functional perspective, but I'm happy to take another look if you end up breaking 'Plugins' into its own tab in this PR like @thomasguillot recommended.

CleanShot 2025-04-03 at 08 29 36

dkoo:
Thanks, @laurelfulford! Yeah, re: the "Suppression" settings, I'm proposing here that we wrap the section in a card like the other settings sections so that the page looks more visually consistent. Ok with you, @thomasguillot?

Note that we've also decided to eliminate the "Plugins" heading halfway down the page. The whole page now looks like this. with all settings sections existing at the same hierarchy in the Settings page:

Screenshot 2025-04-03 at 11-15-39 Advertising ‹ dkoo — WordPress

I'll flag this as needing a design review from @thomasguillot since the end result is a bit different than the original designs.

thomasguillot (approved):
👍

PR #3883, opened 2025-04-03
feat: add unique icons repository and remove custom icons

thomasguillot:
### All Submissions:

* We now have an `/src/icons/` containing ALL the custom Newspack icons (see [Figma](https://www.figma.com/design/HTXAGh0kxIUtKTaCKTMsA2/Newspack-Blocks?node-id=131-648&t=VYExRt8f57MuIeye-1))
* The idea is in the future we create an NPM package so we can easily include these icons in our plugins and we only have to maintain them in one place
* I've removed all the custom icons I could find, and replaced them with the new icons. This includes:
    * Dashboard
    * Prompts
    * Corrections block
    * Reader registration block

1. Switch to this branch
2. Check if all the affected wizards still work and display the correct icon
3. Check if all the affected blocks still work and display the correct icon

### Other information:


dkoo (changes requested):
Not seeing anything broken, and I love the consolidation of all custom icons in a single place. I did see a few `<svg>` elements floating around the dashboard pages that were maybe missed:

- https://github.com/Automattic/newspack-plugin/blob/add/icons/src/bylines/index.js#L27
- https://github.com/Automattic/newspack-plugin/blob/add/icons/src/components/src/newspack-icon/index.js#L34
- https://github.com/Automattic/newspack-plugin/blob/add/icons/src/plugins-screen/plugins-screen.js#L56

If the ultimate goal is to create an NPM package for this, creating it as a separate repo might be a good idea instead of bundling it into this plugin (I know we do this for Newspack Components, but it causes some confusion).

thomasguillot:
@dkoo I tried updating https://github.com/Automattic/newspack-plugin/blob/add/icons/src/bylines/index.js#L27 but I'm not sure how I can use a button component in this markup (even Cursor is confused). The code for this feature is an absolute mess (inc. the CSS)
I was only able to make a small tweak https://github.com/Automattic/newspack-plugin/pull/3883/commits/449ca6b49b69bbcd40e70743b16d4076c4804098

thomasguillot:
@dkoo https://github.com/Automattic/newspack-plugin/blob/add/icons/src/components/src/newspack-icon/index.js#L34 This is the newspack icon as a component so this we're not modifying it

thomasguillot:
@dkoo Same with https://github.com/Automattic/newspack-plugin/blob/add/icons/src/plugins-screen/plugins-screen.js#L56 I wasn't able to use a component.
I did make a small change though https://github.com/Automattic/newspack-plugin/pull/3883/commits/876af500d66201115fed88024f42cc1c4c7fbf30

thomasguillot:
> If the ultimate goal is to create an NPM package for this, creating it as a separate repo might be a good idea instead of bundling it into this plugin (I know we do this for Newspack Components, but it causes some confusion).

How could that be done? Would a separate PR make sense?

I was thinking of following Gutenberg’s example, where all packages are organised in one place: https://github.com/WordPress/gutenberg/tree/trunk/packages.

dkoo:
> How could that be done? Would a separate PR make sense?
> 
> I was thinking of following Gutenberg’s example, where all packages are organised in one place: https://github.com/WordPress/gutenberg/tree/trunk/packages.

Gotcha, in that case this is a fine way to do things. Nevermind!

dkoo (approved):

PR #3895, opened 2025-04-07
fix(analytics): migrate back-end GA4 events to front-end

dkoo:
### All Submissions:

Deprecates back-end GA4 events, which we've found to be too inaccurate to be useful, possibly due to various caching layers readers might encounter in production sites.

Migrates `reader_registered` and `reader_logged_in` events to the front-end, although not with all params that were included on back-end equivalents.

Also fixes a bug where registering via a Registration block directly inside a content gate (as opposed to an auth modal triggered by a link in the content gate) fails to trigger an auto page reload after successful registration.

Also migrates all content gate GA events to the front-end. Note that the parameters passed to these events and some of the values for these parameters differ from the back-end events they replace, notably:

- `action_type=paid_membership` - This will now be `action_type=checkout_button`, which matches the parameter value given by modal checkout interaction events. This should make it easier to track conversion rates by filtering events by `action_type=checkout_button`.
- Donation interactions from within content gates are now tracked with their own parameters.

All events and their parameters are documented in a new [README file](https://github.com/Automattic/newspack-plugin/blob/d98601f48b0cc0c2f5602f6d6dd277b51657f80a/src/memberships-gate/README.md).

Requires https://github.com/Automattic/newspack-blocks/pull/2103, https://github.com/Automattic/newspack-newsletters/pull/1800, and https://github.com/Automattic/newspack-popups/pull/1424

Supersedes #3805. Closes `1200550061930446/as/1209870698187233`.

#### Content gate events

1. Check out this branch, https://github.com/Automattic/newspack-blocks/pull/2103, https://github.com/Automattic/newspack-newsletters/pull/1800, and https://github.com/Automattic/newspack-popups/pull/1424.
2. Set up a site with Google Analytics + RAS + Woo Memberships with a metered regwall + paywall setup (DM me for instructions on how). The important thing is that the Non-Member Content block shown to anonymous readers contains one or more of the Donate, Registration, or Checkout Button blocks, and the Member Content block shown to registered readers contains the others of those blocks.
3. As an anonymous reader, visit a post that will show the regwall part of the content gate. In Dev Tools > Network look for a `collect` XHR request that contains a payload that looks like this:

Screenshot 2025-04-08 at 5 04 02 PM

Note the highlighted part of the payload, and confirm that the `gate_has_donation_block`, `gate_has_registration_block`, and `gate_has_checkout_button` params are accurately set to `yes` or `no` based on which of those blocks are visible within the content gate. Also note other parameters as described in the [README](https://github.com/Automattic/newspack-plugin/blob/d98601f48b0cc0c2f5602f6d6dd277b51657f80a/src/memberships-gate/README.md) and confirm they're accurate.

4. Register (or log in) via the Registration block inside the content gate to pass the regwall. Confirm that after a couple of seconds, the page reloads automatically to show the full article content.
5. In Dev Tools > Network look for the `collect` XHR requests with event name `reader_registered` or `reader_logged_in`, and confirm their parameters are accurate. 
6. View more posts until you see the paywall content gate. In Dev Tools > Network look for the `collect` XHR requests and confirm that their parameters are accurate.
7. Start modal checkouts from Donation and Checkout Blocks inside the content gate. Again, look for the `collect` requests and confirm their parameters are accurate.
8. Dismiss the modal checkouts and complete them, and confirm that `np_modal_checkout_interaction` events are fired for these interactions, and that all events contain a `gate_post_id` with the content gate's post ID, indicating these checkouts originated from the content gate.
3. Review the `src/memberships-gate/README.md` file and confirm that it accurately describes the events and their params.

#### `registered` & `logged_in` events

1. Register new reader accounts via the Registration block, Newsletter Subscription Form block, and the auth modal. For the blocks, register accounts via block instances both inside and not inside Campaigns prompts and content gates (note: for the latter you may need to stop the auto-refresh from happening or watch real-time reports in Google Analytics to see this event). Confirm that all registrations result in `reader_registered` GA events being fired, and that the ones from inside prompts include a `popup_id` param.
2. Log into an existing account via the auth modal and confirm that it fires a `reader_logged_in` GA event.
3. Log into an existing account via a magic link and confirm that a GA event fires on the first page view after authenticating.
4. Review the `src/reader-activation/README.md` file and confirm that it accurately describes the events and their params.

### Other information:


includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -170,6 +170,100 @@ public static function setup_sitekit_ga4() {
...
+	 *
+	 * @return array
+	 */
+	public static function get_custom_event_parameters() {
+		$params = [
+			'logged_in' => is_user_logged_in() ? 'yes' : 'no',
+		];
+		$session_id = self::extract_sid_from_cookies();
+		if ( $session_id ) {
+			$params['ga_session_id'] = $session_id;
```
leogermani: For the purpose of sending custom params to the front end, we don't need that. The front-end already has and automatically adds the session id to events

leogermani (commented):
Almost everything works, the only event I couldn't see was the "action=dismiss". Is it a thing?

Other than that left a comment and let's see if we can align how this and popups measurements work as much as possible

includes/plugins/google-site-kit/class-googlesitekit.php
dkoo: Removed in 2785aeb7

dkoo:
@leogermani 6f50aa5 and c1abda3 add front-end GA events for reader registrations and logins, as well as a README file describing their params. These events are named the same as the equivalent deprecated back-end events, but don't contain as much data in their params. We can explore adding more data in a separate PR, if needed.

dkoo:
> Almost everything works, the only event I couldn't see was the "action=dismiss". Is it a thing?

For `np_gate_interaction` events, `action=dismissed` events are only fired when dismissing the auth modal if it was opened from a page containing a content gate. If dismissing a modal checkout, a `np_modal_checkout_interaction` event with `action=dismissed` is fired, and it will include the `gate_id` param if the checkout started from a content gate.

leogermani:
A few comments questions from my lates tests:

* the modal_checkout_interaction `form_submission_success` event doesn't include the gate id. I can only see the gate_id up until `form_submission`
* I only see the additional params like is_logged_in, is_reader etc, for the np_gate_interaction, but not for the modal_checkout_interaction. Is this expected?
* The refresh after registration worked well, but it didn't refresh the page after a subscription purchase that unlocked the membership.

FInally, I didin't see the `reader_registered` event being fired after a registration from within the content gate. I thought it could be because of the page redirect, but reading the PR description maybe this is expected?

I'd say that if we can't make sure that we capture ALL reader registrations from all the possible methods, maybe we shouldn't add this event just yet. Same for logged_in... it will generate incomplete stats

dkoo:
@leogermani I took another pass at this. Please update these branches, too (note the new one for Popups): https://github.com/Automattic/newspack-blocks/pull/2103, https://github.com/Automattic/newspack-newsletters/pull/1800, and https://github.com/Automattic/newspack-popups/pull/1424

> * the modal_checkout_interaction `form_submission_success` event doesn't include the gate id. I can only see the gate_id up until `form_submission`

This should now be fixed with 10c2f5e and [6f107c7](https://github.com/Automattic/newspack-blocks/pull/2103/commits/6f107c7c7f51222010c6570e82d1b8468a147564). Content gate and prompt data should now be passed in the same manner all the way through modal checkout completion.

> * I only see the additional params like is_logged_in, is_reader etc, for the np_gate_interaction, but not for the modal_checkout_interaction. Is this expected?

These extra params are now sent for all events, but only if `NEWSPACK_GA_ENABLE_CUSTOM_FE_PARAMS` is enabled in wp-config.php.

> * The refresh after registration worked well, but it didn't refresh the page after a subscription purchase that unlocked the membership.

This should be fixed by a8b3457 and [aaffc42](https://github.com/Automattic/newspack-blocks/pull/2103/commits/aaffc42796ab15ed40f4fd3cddf29ec56e4c9492), with a new reader data activity to denote a successfully completed modal checkout. Note that the page will only refresh once all RAS overlays have been closed.

dkoo:
> FInally, I didin't see the `reader_registered` event being fired after a registration from within the content gate. I thought it could be because of the page redirect, but reading the PR description maybe this is expected?

It should be firing now, but it might be hard to see with the auto-reload unless you're watching the Google Analytics Real Time Reports dashboard.

leogermani:
> It should be firing now, but it might be hard to see with the auto-reload unless you're watching the Google Analytics Real Time Reports dashboard.

Will get back to this tomorrow. Did you change anything? I use the "Preserve log" option in the debugger...

Do you think we are capturing 100% of the ways a user can register. I don't think we should add this event if we are not sure. People will look at this and ask why numbers are not matching

dkoo:
> Did you change anything? I use the "Preserve log" option in the debugger...

On that particular item, no—but it's possible the request is getting cancelled if it doesn't complete before the page reload happens. That reminded me of a fix we implemented in #1726 to prevent this from happening. It appears this fix got deprecated from our code at some point, but after reintroducing it locally and testing, it doesn't seem to prevent some pending XHR requests from getting cancelled upon a `window.location` change (you can test this by throttling your connection and clicking away from the page while there are still pending `collect` requests from GA—if they don't complete by the time the browser navigates away, they'll show as cancelled in dev tools). I also can't find any documentation on the `transport_type: 'beacon'` option for GA anymore, which makes me think this behavior is no longer supported. :(

dkoo:
> Do you think we are capturing 100% of the ways a user can register. I don't think we should add this event if we are not sure. People will look at this and ask why numbers are not matching

The new `reader_registered` Data Events activity we're dispatching across all RAS registration flows should trigger the GA event for every possible way a user can register _via our flows_. But not if they somehow create a customer or subscriber user account via other means (such as via standard WooCommerce checkout). Is that enough of a concern to warrant delaying adding this event for now?

dkoo:
@leogermani cb8d59d attempts to avoid the cancelled GA event requests by increasing the delay before the auto-page-reload to 5 seconds. Since this delay is so long, it also adds some user-facing visual feedback in the interim on the gate itself:

overlay gates:
Screenshot 2025-04-16 at 11 57 56 AM

inline gates:
Screenshot 2025-04-16 at 11 54 05 AM

dkoo:
@leogermani this is ready for another review based on the latest discussion:

- No more `form_submission_success` actions for `np_gate_interaction` events—the `form_submission` actions will represent a successful conversion intent
- Improved logic for the auto-reload: the page will auto-reload on gated content only when:
  - Successfully registering a new account, logging into an existing account, or completing a modal checkout transaction
  - Only once all modal overlays have been closed

leogermani (approved):

PR #3888, opened 2025-04-07
Fix: RAS OTP Autocomplete feature

AKSHAT2802:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes: https://app.asana.com/0/1208970929679778/1207509651211057/f

### How to test the changes in this Pull Request:

1. Add `add_filter( 'newspack_reader_activation_enabled', '__return_true' );` 
2. Open site in incognito on Mobile
3. Create a new account and logout
4. Try login with the created account
5. Autofill OTP from Keyboard

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3898, opened 2025-04-08
fix(otp-input): enhance OTP input handling for clipboard

AKSHAT2802:
### All Submissions:

Closes https://app.asana.com/0/1208970929679778/1207509651211057/f

1. Add add_filter( 'newspack_reader_activation_enabled', '__return_true' );
2. Open the site in incognito on Mobile.
3. Create a new account and log out.
4. Try logging in with the created account.
5. Copy OTP to the clipboard.
6. Try Pasting OTP from keyboard suggestion.

### Other information:

### DIFF
--- a/src/reader-activation-auth/otp-input.js
+++ b/src/reader-activation-auth/otp-input.js
@@ -23,7 +23,6 @@ domReady( function () {
 		for ( let i = 0; i < length; i++ ) {
 			const digit = document.createElement( 'input' );
 			digit.setAttribute( 'type', 'text' );
-			digit.setAttribute( 'maxlength', '1' );
 			digit.setAttribute( 'pattern', '[0-9]' );
 			digit.setAttribute( 'autocomplete', 0 === i ? 'one-time-code' : 'off' );
 			digit.setAttribute( 'inputmode', 'numeric' );
@@ -71,9 +70,20 @@ domReady( function () {
 				}
 			} );
 			digit.addEventListener( 'input', ev => {
-				const otpDigit = ev.target.value.trim();
-				if ( otpDigit.match( /^[0-9]$/ ) ) {
-					values[ i ] = otpDigit;
+				const otpInput = ev.target.value.trim();
+				if ( length === otpInput.length ) {
+					for (let index = 0; index < length; index++) {
+						const char = otpInput[index];
+						if (/^[0-9]$/.test(char)) {
+							const input = inputContainer.querySelector(`[data-index="${index}"]`);
+							input.value = char;
+							values[index] = char;
+						}
+					}
+					otpCodeInput.value = values.join('');
+					return;
+				} else if ( otpInput.match( /^[0-9]$/ ) ) {
+					values[ i ] = otpInput;
 					const next  = inputContainer.querySelector(`[data-index="${i + 1}"]`);
 					if (next) {
 						next.focus();
--- a/src/reader-activation-auth/otp-input.js
+++ b/src/reader-activation-auth/otp-input.js
@@ -74,7 +74,7 @@ domReady( function () {
 				if ( length === otpInput.length ) {
 					for ( let index = 0; index < length; index++ ) {
 						const char = otpInput[ index ];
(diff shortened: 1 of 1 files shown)


AKSHAT2802:
Here are the Screen recordings
Before: (Staging site)
https://github.com/user-attachments/assets/57fbfe50-90fc-41e8-81ef-ba96bf3638ac

After: (Local site)
https://github.com/user-attachments/assets/fee04d7d-b88b-4834-acc8-ebfe8eab7e49

miguelpeixe (changes requested):
Thank you for addressing this!

I was able to confirm the issue on iPhone's Chrome and Firefox, but not on Safari. Unfortunately, this patch doesn't seem to fix it for those cases.

I wonder if using the `navigator.credentials.get()` API directly gives us better control to fill the inputs, rather than working around the `autocomplete` parameter.

AKSHAT2802:
> I was able to confirm the issue on iPhone's Chrome and Firefox, but not on Safari. Unfortunately, this patch doesn't seem to fix it for those cases.

Thank you for checking this!
I just tested it again on iOS using Safari, Chrome, and Firefox — and in all three browsers, the OTP is getting filled correctly on my end. Would it be possible for you to recheck on your device? Maybe there's a slight variation in environment or configuration that’s affecting the behavior.

Is the testing device configured with this [[Link](https://support.apple.com/en-gb/guide/iphone/iphf9219d8c9/ios#:~:text=Turn%20Password%20AutoFill%20on%20or,up%20to%20three%20passwords%20apps.)]

> I wonder if using the `navigator.credentials.get()` API directly gives us better control to fill the inputs, rather than working around the `autocomplete` parameter.

I explored using navigator.credentials.get() with the WebOTP API as well.

I implemented it locally, but found it wasn’t a feasible solution in our case due to a few limitations:
- It relies on the WebOTP API, which is only available in some modern browsers.
- It doesn’t support OTP via email. [[link]](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/get#otp)
-Browser compatibility is limited, especially on iOS — it's only supported in Chrome on Android (no support in Firefox, Safari, etc.).
- The SMS must follow a very specific format, including the domain on the last line to work properly. [[Link](https://web-otp.glitch.me/#:~:text=Send%20an%20SMS,glitch.me%20%2312345)]
- Ref: [MDN WebOTP API](https://developer.mozilla.org/en-US/docs/Web/API/OTPCredential)

Screenshot 2025-04-09 at 7 02 54 PM


So while navigator.credentials.get() offers tighter integration in ideal cases, it didn’t feel robust enough to rely on.

miguelpeixe:
Thank you for thoroughly looking into my suggestion! I agree that it's not viable for our requirements.

Here's a screen recording of Chrome under iPhone running this branch:

https://github.com/user-attachments/assets/2a7400df-251e-4d6b-9427-e40309a800b6

AKSHAT2802:
I believe the new changes are unavailable on the staging site. Does this branch needs a redeployment to work with new code changes? (The previous merged [PR](https://github.com/Automattic/newspack-plugin/pull/3888) also had the same branch name which may have lead to stale code).

Screen recording from iOS Chrome (Mail-OTP):


https://github.com/user-attachments/assets/566bdd3e-cd7b-4e7e-b93f-793415368677

miguelpeixe:
I deployed this branch's latest commit to that site and tested it in incognito. Is my issue also reproducible for you in https://boldly-witty.jurassic.ninja so we can confirm it's an env problem?

AKSHAT2802:
Testing site: https://boldly-witty.jurassic.ninja/ is working fine for me. (Incognito mode, chrome, iOS)

Screen Recording:

https://github.com/user-attachments/assets/4b0270d9-9f66-4126-9719-73be15eb8714

miguelpeixe (approved):
I don't know what changed, but it's working consistently now. I also confirmed that the changed approach doesn't introduce regressions to the desktop experience.

I can only think that cache got in the way, even though it's strange because I was using a fresh incognito session every attempt. 🤷

PR #3908, opened 2025-04-11
Caps for RSS Feeds

adekbadek:
### All Submissions:

Adds fine-grained capability control to RSS Feeds. This is a first step (and a POC) in granular capability controls to more areas of Newspack.
Before, RSS Feeds CPT inherited from the regular `post` capabilities, so an administrator would have to allow a user to edit all posts if they wanted them to edit RSS Feeds.  

1. Confirm that as an Editor and as an Admin, you can still manage RSS Feeds
2. Install `capability-manager-enhanced` plugin
3. Go to Capabilities → Roles and add a new role
4. In Capabilities → Capabilities edit the role and give it Posts -> Edit* cap and all caps in RSS Feeds:

image

5. Also add Admin → Read capability

image

6. Create a new user with this role
7. Log in as the user and observe that you cannot see any existing posts, but you can manage RSS Feeds

\* WC will mess with admin login for users without `edit_posts` cap

### Other information:

---
  - https://app.asana.com/0/0/1204478788464366


dkoo (changes requested):
Not sure if this is related to these changes or the `capability-manager-enhanced` plugin, but a user with these capabilities can't access the main dashboard page, resulting in an error screen immediately after logging in:

Screenshot 2025-05-14 at 10 34 25 AM

If I paste the `/wp-admin/edit.php?post_type=partner_rss_feed` URL directly into the browser, I can access this page and I also have access to some other menu items. Most of these are probably tied to the `edit_posts` capability, however because the user lacks `edit_others_posts` most of the features within these menu items are unusable.

Screenshot 2025-05-14 at 10 32 22 AM

adekbadek:
Admin → Read capability was missing. 

image

I've updated the instructions.

miguelpeixe (approved):
Working as described!

It's unfortunate that we have to call `current_user_can( get_post_type_object( $post_type )->cap->edit_others_posts )` instead of the simpler `current_user_can( 'edit_others_posts' )` for these entities now.

Would it make sense for `\Newspack\Capabilities` to provide a more straightfoward way to get that value?

adekbadek:
> Would it make sense for \Newspack\Capabilities to provide a more straightfoward way to get that value?

Good idea, added `Capabilities::current_user_can` method

miguelpeixe (approved):
💯

PR #3918, opened 2025-04-14
WC Stripe: disable Link/Apple Pay/Google Pay by default

Takshil-Kunadia:
### All Submissions:

Closes https://app.asana.com/1/26890605006346/project/1209102263738351/task/1209403196706994?focus=true.

There are no filters to filter the default value of these options, hence hooking onto the pre options saving, if the values are being saved for the first time we disable the respective payment methods ( Apple Pay / Google Pay & Link by Stripe) which are covered under Express Checkout Elements in WC Stripe Payment gateway plugin. This code will run only once when it is a fresh install following this code is not executed and works as per the flow of WC Stripe Payements.

1. Freshly Install WooCommerce Stripe Gateway.
2. Start set up for connecting to Stripe account.
3. After connecting, Switch to Payment methods tab and notice the Express Checkout payments disabled by default.

### Other information:

### DIFF
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -21,12 +21,12 @@ class WooCommerce_Gateway_Stripe {
 		add_filter( 'wc_stripe_intent_metadata', [ __CLASS__, 'add_transaction_metadata' ], 10, 2 );
 
 		/**
-		 * Disable Stripe Express Checkout feature flag.
-		 * This is a workaround for the Stripe Express Checkout feature flag
+		 * Disable Stripe Express Checkout feature flags.
+		 * This is a workaround for the Stripe Express Checkout feature flags
 		 * being enabled by default on new installs.
 		 */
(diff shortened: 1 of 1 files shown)


includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -104,5 +112,58 @@ function( $plan ) use ( $product_id ) {
...
+	/**
+	 * Disable Stripe Express Checkout in main settings.
+	 *
+	 * @param array $settings Stripe settings.
+	 * @param array $old_settings Old Stripe settings.
+	 * @return array
+	 */
+	public static function disable_express_checkout_in_main_settings( $settings, $old_settings ) {
+		/**
+		 * If the old strip settings are empty, it means this is a new install.
```
dkoo:
```suggestion
		 * If the old stripe settings are empty, it means this is a new install.
```

dkoo (approved):
Works! Just a typo below, then it's good to go

includes/plugins/class-woocommerce-gateway-stripe.php
Takshil-Kunadia: Thanks! @dkoo for catching that.
Thanks! @leogermani for updating it. 🙇

PR #3923, opened 2025-04-16
fix(wizards): react render strategy and ref forwarding

miguelpeixe:
### All Submissions:

Change our react app rendering strategy to `createRoot()`, fixing the following deprecation warning:

```
Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17
```

It also addresses ref forwarding issues for function components under the `withWizard()` HOC.

```
react-dom.js:73 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

To address this, the `Wizard` component required a small refactor to be encapsulated in `forwardRef()`, which then required the `useWizardData` and `STORE_NAMESPACE` to no longer be appended to the component function. This is a good change; this is not a good way to export dependencies.

92c5fad8488784a7e79af34f15522bdcc37527ee introduces the `contenthash` to the asset chunk's filename to deal with cached assets coming from [the lazy load strategy](https://github.com/Automattic/newspack-plugin/blob/a8a590cb2b4514e7cb248e086c8464fec5b84461/src/wizards/index.tsx).

Navigate **all wizards** and confirm they render as expected.

### Other information:

### DIFF
--- a/src/components/src/wizard/index.js
+++ b/src/components/src/wizard/index.js
@@ -7,7 +7,7 @@ import classnames from 'classnames';
  * WordPress dependencies.
  */
 import { useSelect } from '@wordpress/data';
-import { useState } from '@wordpress/element';
+import { useState, forwardRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { category } from '@wordpress/icons';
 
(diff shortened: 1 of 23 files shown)


claudiulodro (approved):
Works great in my testing 👍

PR #3924, opened 2025-04-16
fix: change nonce strategy for data events

leogermani:
### All Submissions:

This PR replaces the nonce strategy for data events. In some cases, and in some sites, nonce verification can fail because the user session is not recognized in the dispatched request that handles the events. 

Since we don't really need a nonce per user, this implementation is simpler and more reliable.

1. Make some actions that will trigger some data events (for example, subscribe to a newsletter or make a purchase while RAS is enabled)
2. Inspect the error log and confirm all the events are being dispatched and handled as usual

### Other information:

### DIFF
--- a/includes/data-events/class-data-events.php
+++ b/includes/data-events/class-data-events.php
@@ -34,9 +34,24 @@ final class Data_Events {
 	const NONCE_EXPIRATION_OPTION = 'newspack_data_events_nonce_expiration';
 
 	/**
-	 * Nonce lifetime in seconds (24 hours).
+	 * Nonce lifetime in seconds (1 hour).
 	 */
-	const NONCE_LIFETIME = 86400; // 24 hours in seconds
+	const NONCE_LIFETIME = 3600; // 1 hour in seconds
+
+	/**
+	 * Grace period in seconds for accepting old nonces.
+	 */
+	const NONCE_GRACE_PERIOD = 10; // 10 seconds
+
+	/**
+	 * Option name for storing the previous nonce.
+	 */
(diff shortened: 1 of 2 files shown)


includes/data-events/class-data-events.php
```diff
@@ -23,6 +23,21 @@ final class Data_Events {
...
+
+	/**
+	 * Option name for storing the nonce expiration.
+	 */
+	const NONCE_EXPIRATION_OPTION = 'newspack_data_events_nonce_expiration';
+
+	/**
+	 * Nonce lifetime in seconds (24 hours).
+	 */
+	const NONCE_LIFETIME = 86400; // 24 hours in seconds
```
miguelpeixe: This nonce strategy is less secure than WP's, wdyt we flush it more often? 1 hour?

miguelpeixe: We could also do something other than nonces. Make it more secure by signing the payload, maybe an hmac hash? It's the server talking to itself, so we don't need a key pair, just the signature hash.

miguelpeixe: Nevermind, exactly because the server is talking to itself it doesn't matter much. I'd still lower the lifetime though

leogermani: https://github.com/Automattic/newspack-plugin/pull/3924/commits/c1fe71bcd3d5a2618ecd85ff4977b8b71632c3ae decreases lifespan to one hour.

I decided to then add a 10 seconds grace period to avoid requests from failing if they happen to be dispatched exactly when the nonce is rotated

miguelpeixe (approved):
Non-blocking: I think the "previous nonce" logic could be simplified using wp transients.

PR #3927, opened 2025-04-17
fix(corrections): Limit CPT to posts only

Takshil-Kunadia:
### All Submissions:

Addresses https://app.asana.com/1/26890605006346/task/1209996963457664?focus=true.

This PR ensures the Corrections and Clarifications panel is only displayed for the post post type. Previously, it could appear publicly as a conventional CPT applicable of Content Gate.

To resolve this:
- The custom post type registration for Content Gate is now set to 'public' => false.
- Also added checks to ensure corrections modal to supported post types.

| Before | After |
|--------|--------|
| Screenshot 2025-04-17 at 3 01 47 PM | Screenshot 2025-04-17 at 3 00 59 PM | 

1. Open configuration screen for any of the content gate content.
2. Notice Corrections not available as a CPT applicable to content gate.
3. Create or edit a post of type post or the supported post type — the Corrections and Clarifications panel should appear.
4. Open any random Page editor screen. 
5. Confirm the panel should not appear.

### Other information:

### DIFF
--- a/includes/corrections/class-corrections.php
+++ b/includes/corrections/class-corrections.php
@@ -37,11 +37,6 @@ class Corrections {
 	const CORRECTIONS_TYPE_META = 'newspack_corrections_type';
 
 	/**
-	 * Supported post types.
-	 */
-	const SUPPORTED_POST_TYPES = [ 'article_legacy', 'content_type_blog', 'post', 'press_release' ];
-
-	/**
 	 * REST route for corrections.
 	 */
 	const REST_ROUTE = '/corrections';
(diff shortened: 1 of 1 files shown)


Takshil-Kunadia:
https://github.com/Automattic/newspack-plugin/blob/0c00e7f13f915ec48982c05013e46442f25495f0/includes/corrections/class-corrections.php#L42

@leogermani Should I reduce this to only post as specified in the ticket?

includes/corrections/class-corrections.php
```diff
@@ -145,18 +150,20 @@ public static function register_post_type() {
...
-			'show_in_rest'     => true,
-			'supports'         => $supports,
-			'taxonomies'       => [],
-			'menu_icon'        => 'dashicons-edit',
+			'labels'              => $labels,
+			'description'         => 'Post type used to store corrections and clarifications.',
+			'has_archive'         => true,
+			'public'              => false,
+			'publicly_queryable'  => true,
+			'exclude_from_search' => false,
```
dkoo: Nice fix to allow corrections to appear in front-end search results.

dkoo (approved):
Approving because it does what it describes, but I'll defer to @leogermani on the question of which post types should be supported.

leogermani:
Thanks for catching that! I removed those post types and added a filter to allow people to opt in

includes/corrections/class-corrections.php
leogermani: @dkoo  I actually just changed `excluded_from_search` to `true`. Do you think it should not be excluded from search? I think it should because it will generate noise in search results... there's no `single` template for corrections

dkoo (approved):
Looks good!

includes/corrections/class-corrections.php
dkoo: Ah, gotcha—I think it can go either way. Since the corrections' content appear in article content, it would make sense to me for them to appear in search results, but not if they don't link anywhere. Let's leave as `true` for now

Takshil-Kunadia:
Merging this PR. Thanks for the inputs 🙇

PR #3938, opened 2025-04-21
feat(ga4): np_newsletter_subscribed event upon front-end signup

dkoo:
### All Submissions:

#3895 migrated back-end GA4 events to the front-end, but didn't implement events to replace back-end `np_newsletter_subscribed` events. This PR does.

1. Check out this branch and https://github.com/Automattic/newspack-newsletters/pull/1819.
2. As a new reader, sign up for newsletters using the following methods:

  - Registration block, signing up for newsletters
  - Auth modal, after registering (must enable post-checkout newsletter signup lists in Audience > Settings)
  - After completing modal checkout (must enable post-checkout newsletter signup lists in Audience > Settings)
  - My Account > Newsletters (check new lists and save)

3. In each case, confirm that an `np_newsletter_subscribed` event is dispatched to Google Analytics with the following params:
  - `lists` - array of list IDs the user subscribed to
  - `newsletters_subscription_method` - the method of signup (one of: `newsletters-subscription-block`, `reader-registration`, `auth-modal`, `post-checkout`, or `my-account`)
  - `newspack_popup_id` or `gate_post_id`, if applicable

### Other information:

---
  - https://app.asana.com/0/0/1210051336945369

### DIFF
--- a/includes/reader-revenue/my-account/index.js
+++ b/includes/reader-revenue/my-account/index.js
@@ -31,6 +31,7 @@ domReady( function () {
 	const cancelButton = document.querySelector( '.subscription_details .button.cancel' );
 	const { labels, nonce, rest_url, should_rate_limit } = newspack_my_account || {};
 
+	// Show a confirmation dialog before cancelling a subscription.
 	if ( cancelButton ) {
 		const confirmCancel = event => {
 			const message =
@@ -45,6 +46,7 @@ domReady( function () {
 		cancelButton.addEventListener( 'click', confirmCancel );
 	}
 
+	// Rate limit the add payment method form.
 	const addPaymentForm = document.getElementById( 'add_payment_method' );
 	if ( addPaymentForm && Boolean( should_rate_limit ) ) {
 		const errorContainer = document.querySelector( '.woocommerce-notices-wrapper' );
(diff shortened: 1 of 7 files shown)


includes/reader-revenue/my-account/index.js
```diff
@@ -93,4 +95,22 @@ domReady( function () {
...
+		if ( subscribed && reader?.email && reader?.authenticated ) {
+			readerActivation.dispatchActivity( 'newsletter_signup', {
+				email: reader.email,
+				lists: subscribed,
+				newsletters_subscription_method: 'my-account',
+			} );
+		}
+		params.delete( 'newspack_newsletters_subscription_subscribed' );
+		const newQueryString = params.toString() ? '?' + params.toString() : '';
+		window.history.replaceState( {}, '', window.location.pathname + newQueryString );
```
miguelpeixe: This will also fire an event with empty lists if I only remove subscriptions. I don't think we should fire for such cases.

miguelpeixe (changes requested):
There are some schema inconsistencies across different event origins. When subscribing to newsletters via the reader registration block the `lists` are not part of the payload, just `email` and `newsletters_subscription_method`.

There's also a strange behavior when accessing the My Account after registering, I got a couple of dispatches with the following payload:

```
{
  "action": "newsletter_signup",
  "data": {
    "email": "miguel@example.com",
    "lists": [
      ""
    ],
    "newsletters_subscription_method": "my-account"
  },
  "timestamp": 1745412894141
}
```

This happened to an unverified account, at the My Account dashboard landing page.

includes/reader-revenue/my-account/index.js
dkoo: Ah, the `split( ',' )` was in the wrong place, resulting in that `[ '' ]` value when the param is empty or nonexistent. b9e5937 should fix it.

dkoo:
@miguelpeixe thanks for catching those bugs!

> There are some schema inconsistencies across different event origins. When subscribing to newsletters via the reader registration block the `lists` are not part of the payload, just `email` and `newsletters_subscription_method`.

This was happening because the success callback was removing the form element before we built the activity data. 🤦 23cce1e should fix this by removing the form _after_ we dispatch the event.

> There's also a strange behavior when accessing the My Account after registering, I got a couple of dispatches with the following payload:

Is this still happening for you after b9e5937? If so, we can tighten up the JS so that this particular push only happens if the `newspack_newsletters_subscription_subscribed` query param exists AND we're on the `my-account/newsletters` path, but in my testing this doesn't seem to be necessary.

miguelpeixe (approved):
Thanks for revising! The issues were fixed and all is working well.

PR #3949, opened 2025-04-23
feat(analytics): "My Account" dashboard interactions

miguelpeixe:
### All Submissions:

Dispatch reader data activities and GA4 events on My Account interactions:

| Activity | Params | Interaction |
| --- | --- | --- |
| `subscription_cancelled` | `subscription_id` | Click and confirm the "Cancel" button on an active subscription page. |
| `subscription_reactivated` | `subscription_id` | Click the "Reactivate" button on a subscription pending cancellation. |
| `subscription_switched` | -- | Complete the checkout after clicking "Upgrade or Downgrade" on an active **variable** subscription |
| `payment_method_deleted` | -- | Click "Delete" on a payment method in the "Payment methods" page |
| `payment_method_added` | `payment_method`: the method gateway (stripe, woocommerce_payments, etc) | Submit a new payment method via the "Payment methods" page  |
| `payment_method_changed` | `subscription_id`, `update_all_subscriptions` | Submit a new payment method for an active subscription. |
| ~`address_updated`~* | ~`address_type`: Either `billing` or `shipping`~ | ~Submit an address update via the "Addresses" page~ |
| `product_reordered` | -- | Complete the checkout after clicking "Order again" on a past order from the "Orders" page |

The GA event for each activity above is `np_{activity}`

Update*: The `address_updated` was removed in favor of GA's native `form_submit`. For accuracy, the `form_name` should be `billing_address`, `shipping_address`, and `edit_account`.

This PR also refactors the analytics approach to introduce a few utility functions and reduce repetition.

1. Confirm the `reader_registered` and `reader_logged_in` events continue to work as proposed in #3895
2. Test each event described above, following the "Interaction" column of the table, and confirm the activities are dispatched and the GA4 events are pushed

### Other information:


includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -112,11 +112,67 @@ public static function api_check_rate_limit() {
...
+	 * Whether it's the standard "Switch Subscription" page.
+	 *
+	 * @return bool
+	 */
+	public static function is_switch_subscription_checkout_page() {
+		return (
+			function_exists( 'is_checkout' )
+			&& is_checkout()
+			&& function_exists( 'wcs_cart_contains_switches' )
+			&& wcs_cart_contains_switches()
```
dkoo, line 135: Non-blocking: Looks like this function returns data on the switch items, including the subscription ID being switched, if any. Would it be useful to pass some of this data to the front-end so we can include it in the activity and GA4 event payloads? At least the subscription ID being affected, but maybe also the type of switch (seemingly in a `upgraded_or_downgraded` key for the items).

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -112,11 +112,67 @@ public static function api_check_rate_limit() {
...
+	/**
+	 * Whether the cart contains reorders.
+	 */
+	public static function cart_contains_reorders() {
+		$cart = \WC()->cart;
+		if ( ! $cart ) {
+			return false;
+		}
+		foreach ( $cart->get_cart() as $item ) {
+			if ( isset( $item['newspack_order_again'] ) && $item['newspack_order_again'] ) {
```
dkoo: Non-blocking: Similar to the subscription switching event, is there any value in including info on the item being reordered? Perhaps the original order ID and the product being reordered?

includes/reader-revenue/my-account/class-woocommerce-my-account.php
```diff
@@ -112,11 +112,67 @@ public static function api_check_rate_limit() {
...
+	public static function is_payment_method_change_page() {
+		return isset( $_GET['my_account_checkout'] ) && isset( $_GET['change_payment_method'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+	}
+
+	/**
+	 * Whether it's the standard "Switch Subscription" page.
+	 *
+	 * @return bool
+	 */
+	public static function is_switch_subscription_checkout_page() {
```
dkoo: These checks evaluate whether you're on the checkout page and whether the cart contains certain items. One thing I've noticed is that if you start a flow but then abandon it, the items remain in the cart and then affect these checks as long as they're still in the cart.

For example: 

1. Start an upgrade/downgrade by adding one to your cart but then don't complete the checkout.
2. Go back to the My Subscription page and click "Change Payment".
3. Note that `newspack_my_account.is_switch_subscription_checkout_page` is truthy despite not being in that flow anymore.
4. If you complete the "Change Payment" flow, because `newspack_my_account.is_switch_subscription_checkout_page` is truthy, the activity logged is a `subscription_switched` action instead of `payment_method_changed`.

Maybe we need to add some more conditions to these checks beyond what's in the cart, or empty the cart ourselves when starting any of these flows?

src/reader-activation/analytics.js
```diff
@@ -1,16 +1,16 @@
...
  * @return {Object} Event payload.
  */
-
-const getEventPayload = ( extraParams = {} ) => {
+const getEventPayload = ( payload = {}, data = {} ) => {
 	return {
-		...extraParams,
+		...payload,
+		newspack_popup_id: data?.newspack_popup_id || null,
+		gate_post_id: data?.gate_post_id || null,
```
dkoo: We might want to avoid adding these keys to the payload if the values are `null` to a.) save a few bytes per GA request, and b.) because of GA's limit of 25 max params per event, these empty params are just taking up space.

dkoo (changes requested):
@miguelpeixe I really like the abstractions you've put in place like `registerElementActivity`, `registerCheckoutActivity`, and `registerActivityEvent` which should make the Reader Data API even more robust and extensible. Thanks for making these changes.

Feedback is inline. I think you mentioned that we should do a review of the event params we want to collect in a separate PR, so I marked a couple of my items as non-blocking.

dkoo:
Also, I neglected to include this interaction in the original proposal, but should we add an activity/event for updating account details?

src/reader-activation/analytics.js
miguelpeixe: Good call! Fixed in ac05a08e338717472adbe548a7d545487470ead7 and also removed the `referrer` parameter (see p1746557784095789-slack-C015W6BES8J)

includes/reader-revenue/my-account/class-woocommerce-my-account.php
miguelpeixe: Nice catch! This seems like a very edge case. What other case would we have for a checkout page that is not about the items in the cart?

Added a check in 674ac1bb4772d4f87e56d8307dc7ce313ea0310b

miguelpeixe:
> Also, I neglected to include this interaction in the original proposal, but should we add an activity/event for updating account details?

We could, and that reminds me that this and the address updates are already trackable via GA's `form_submit`:

image

Would it be worth looking into enhancing the existing event by populating `form_id` and `form_name` instead of a custom event?

includes/reader-revenue/my-account/class-woocommerce-my-account.php
miguelpeixe: Added `order_id` and `product_id` in fd0697a868361f00c86369f028ff5ecb6ec88030

includes/reader-revenue/my-account/class-woocommerce-my-account.php
miguelpeixe: Added `subscription_id` in 1027bfec860eb6e9054c4149f538e27d920749bb and `upgraded_or_downgraded` in cc5e0ec70c7fb39f120debfbc4c140444c0ad7b5

includes/reader-revenue/my-account/class-woocommerce-my-account.php
dkoo: I don't know of any other cases where this might occur. In My Account on my test site, the only checkout flows I could find were subscription switching, repurchasing an order, paying for a pending order, and payment method change. Theoretically if you start and then abandon any of those flows without emptying the cart, then start a different one, the existing cart items could cause some confusion about what type of the flow the second one is, but I agree that this is an edge case that real users aren't likely to encounter.

dkoo:
> Would it be worth looking into enhancing the existing event by populating form_id and form_name instead of a custom event?

That sounds like a solid idea to me, and this could help us cover other form submissions throughout My Account even if we don't write custom events for them.

miguelpeixe:
Switched the address update strategy to use `form_submit` with the `form_name` as `billing_address` and `shipping_address`. Also added a `form_name` for `edit_account`. 48552e3b8766d5ffb6c57faffa1de2a8bb3ed8da

Also updated the PR description

src/utils/index.js
```diff
@@ -54,3 +54,61 @@ export function convertFormDataToObject( formData, includedFields = [] ) {
...
+ *
+ * @param {string|Element} element The element to register the activity on. Can either be the element node or a string for the element selector.
+ * @param {string}         action  The action to dispatch an activity for.
+ * @param {Function}       cb      The callback to populate the activity data.
+ * @param {string}         event   The element event to listen for. Defaults to `submit` for form elements and `click` for other elements.
+ */
+export function registerElementActivity( element, action, cb, event ) {
+	window.newspackRAS = window.newspackRAS || [];
+	if ( typeof element === 'string' ) {
+		element = document.querySelector( element );
```
dkoo: Sorry I missed this in the first round, but I think this needs to be `document.querySelectorAll` so that the function can handle multiple instances of the same selector on a page. Right now the `payment_method_deleted` event will only fire when deleting the first payment method in the table, as each payment method in the table has its own "Delete" button.

dkoo (changes requested):
@miguelpeixe this is working great, and the use of `form_submit` to capture edit account and address updates is an excellent idea. A couple of minor things I saw and one thing that I think we need to refactor (below, inline).

* Not sure if this is related to this changeset, but when submitting updates to Edit Account and Address forms, I see a PHP warning: `PHP Warning:  Undefined variable $field_values in /var/www/html/wp-content/plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php on line 312`. This doesn't seem to affect anything functionally, AFAICT.
* Renewing an active subscription from the My Account > My Subscription page fires the same `np_product_reordered` event as if clicking "Order again" on a past order from the "Orders" page. Should we consider a separate event for subscription renewals, or is it okay to conflate these two actions?

src/utils/index.js
miguelpeixe: Ah, great catch! Fixed in 69df8db8c62e45c859fe5c08c678a85d778e5073

miguelpeixe:
> Not sure if this is related to this changeset, but when submitting updates to Edit Account and Address forms, I see a PHP warning: PHP Warning:  Undefined variable $field_values in /var/www/html/wp-content/plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php on line 312. This doesn't seem to affect anything functionally, AFAICT.

This seems unrelated to this PR and I'm unable to reproduce (on either trunk or this branch). I think this should be looked at separately.

> Renewing an active subscription from the My Account > My Subscription page fires the same np_product_reordered event as if clicking "Order again" on a past order from the "Orders" page. Should we consider a separate event for subscription renewals, or is it okay to conflate these two actions?

Oh, that's unexpected. Great find! I guess we can track that as well then 😄 
Added the `subscription_renewal_early` activity/event containing the `subscription_id` in 21306c875a67f452345f30d684ddad2869144a8d

dkoo (changes requested):
> This seems unrelated to this PR and I'm unable to reproduce (on either trunk or this branch). I think this should be looked at separately.

👍

> Added the `subscription_renewal_early` activity/event containing the `subscription_id` in 21306c875a67f452345f30d684ddad2869144a8d

_Really_ close now! I'm seeing multiple instances of the `np_subscription_renewal_early` event when I click "Renew now", load the checkout page with the early renewal, and then finally complete the renewal. I think we should maybe fire only one event when you click the button to complete the renewal transaction, right?

src/utils/index.js
dkoo: Confirmed!

miguelpeixe:
Hm... that's strange, it should be the case because it's implemented via `registerCheckoutActivity()`, just like the other ones. Taking a closer look.

miguelpeixe:
I'm unable to reproduce the issue. There's no activity attached to the "Renew Now" button, only the checkout. Can you log something at the `sendEvent()` so we can see exactly at which point multiple events are fired at once?

src/utils/index.js
```diff
@@ -54,3 +54,67 @@ export function convertFormDataToObject( formData, includedFields = [] ) {
...
+	}
+	// If no callback is provided, use a noop.
+	if ( ! cb ) {
+		cb = () => ( {} );
+	}
+	event = event || ( element.tagName === 'FORM' ? 'submit' : 'click' );
+
+	if ( element && ! Array.isArray( element ) ) {
+		element = [ element ];
+	}
```
dkoo:
```suggestion
	if ( element && ! Array.isArray( element ) ) {
		element = [ element ];
	}
	if ( ! element?.length ) {
		return;
	}
	// If no callback is provided, use a noop.
	if ( ! cb ) {
		cb = () => ( {} );
	}
	event = event || ( element[0].tagName === 'FORM' ? 'submit' : 'click' );
```

dkoo (commented):
@miguelpeixe I think I see what's happening. If passing `element` as a string to `registerElementActivity()`, `element` becomes an array due to the `querySelectorAll`, but then the condition for defining the event to listen for assumes `element` is a single item. What's happening for me is that `event` is being redefined as `click`, so every time I click on an input inside the `form[name="checkout"]` element, it's firing the checkout activity. I think the suggested changes should solve this.

src/utils/index.js
miguelpeixe: Ah, good catch! I think it's safer to just place the event detection inside the loop. Updated in 39625d592733b4c35861f705833355cab02884bf

miguelpeixe: I think I wasn't able to reproduce because I was just scrolling down and doing a single click on the checkout submit button 😅

dkoo (approved):
Looking good, now! Thanks for the changes!

PR #3951, opened 2025-04-25
feat(my-account): apply Newspack UI styles to My Account w/ env constant

dkoo:
### All Submissions:

Reorganizes My Account files to support switching between multiple different versions of the UI with an environment constant. Establishes `v0` (current UI) and `v1` classes which enqueue different assets and have the ability to load different custom templates depending on which UI version is active.

This PR doesn't attempt to fully implement new My Account designs, templates, or custom UI components yet, it just applies CSS to get global content/layouts and individual components as close as possible without altering the templates.

Some notes:

- Does the use of child "UI version" classes and version string comparisons make sense for this? I'm imagining we might eventually need to support even more than two different versions of the UI.
- The styles are split between the My Account SCSS and a new `_my-account.scss` Newspack UI module. The latter is so that we can `@extend` important WooCommerce classes with Newspack UI styles without importing and duplicating Newspack UI CSS in the compiled My Account CSS.
- The V1 class dequeues WooCommerce styles from the Newspack Theme. This is so we don't have to fight/override the theme styles as much and can start from more of a blank slate.
- Moved the My Account front-end JS from the `includes` folder to `src/my-account` with some more modular organization: 
  - JS is split into smaller modules by My Account section/feature
  - Top-level JS is enqueued and executed in all My Account pages, regardless of active UI version
  - JS and SCSS in `v0` and `v1` folders are compiled as separate files and enqueued/executed only when the matching UI version is active

1. Check out this branch.
2. Log in as a verified reader with newsletters, subscriptions and/or donations.
3. Smoke test the My Account pages—without adding an environment constant to wp-config.php, all pages should look and behave identically as in `trunk`.
4. Add an environment constant to wp-config.php: `define( 'NEWSPACK_MY_ACCOUNT_VERSION', '1.0.0' );` (this tells the plugin to load the `v1` class)
5. Hard-refresh the My Account pages and confirm that the page layouts and their components are using Newspack UI styles (make sure to test both mobile and non-mobile breakpoints). Behavior of features throughout all pages should remain identical to `v0`/`trunk`.

<table>
<tr>
<td>Screenshot 2025-04-25 at 11 46 06 AM</td>
<td>Screenshot 2025-04-25 at 11 47 26 AM</td>
</tr>
</table>

### Other information:


dkoo:
Tagging @Automattic/newspack-design for a design review. Keep in mind this PR is an incremental step toward the full redesign and the scope of changes here is limited to CSS only.

src/my-account/payment-methods.js
```diff
@@ -0,0 +1,65 @@
...
+ * Initialize functions for the Payment Methods page.
+ */
+export function init() {
+	rateLimitPaymentMethods();
+}
+
+/**
+ * Rate limit the add payment method form.
+ */
+function rateLimitPaymentMethods() {
```
miguelpeixe: Non-blocking: Separating in multiple files is a great decision but I think this `init()` layer looks unnecessary. Each module can import `domReady` and run their thing while `index.js` just imports it without having to call a function.

src/my-account/v1/_navigation.scss
```diff
@@ -0,0 +1,47 @@
...
+			padding-right: var(--newspack-ui-spacer-5);
+			width: auto;
+		}
+		ul {
+			list-style: none;
+			margin: 0;
+			padding: 0;
+		}
+		li {
+			a {
```
miguelpeixe, line 29: Are there no hover/active styles?

includes/reader-revenue/woocommerce/my-account/class-woocommerce-my-account.php
```diff
@@ -1,22 +1,16 @@
 <?php
 /**
- * Connection with WooCommerce's features.
+ * Newspack customizations of WooCommerce's My Account features.
```
miguelpeixe: We have WooCommerce stuff in `includes/plugins` and `includes/reader-revenue/woocommerce`. Maybe we use this opportunity to consolidate in a `includes/plugins/woocommerce` directory?

includes/reader-revenue/woocommerce/my-account/class-woocommerce-my-account.php
```diff
@@ -79,9 +72,28 @@ public static function init() {
 			\add_filter( 'wc_memberships_my_memberships_column_names', [ __CLASS__, 'remove_next_bill_on' ], 21 );
 			\add_action( 'profile_update', [ __CLASS__, 'handle_admin_email_change_request' ], 10, 3 );
 			\add_action( self::SYNC_ESP_EMAIL_CHANGE_CRON_HOOK, [ __CLASS__, 'sync_email_change_with_esp' ], 10, 3 );
+
+			// Decide which My Account UI version to load.
+			if ( version_compare( self::get_version(), '1.0.0', '<' ) ) {
+				include_once NEWSPACK_ABSPATH . 'includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v0.php';
+			} else {
+				include_once NEWSPACK_ABSPATH . 'includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php';
+			}
```
miguelpeixe: Do we need the full path here? I think this should be fine:

```php
include_once __DIR__ . '/class-my-account-ui-v0.php';
```

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -0,0 +1,101 @@
...
+			case 'myaccount/form-login.php':
+				if ( isset( $_GET[ self::AFTER_ACCOUNT_DELETION_PARAM ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+					return dirname( NEWSPACK_PLUGIN_FILE ) . '/includes/reader-revenue/woocommerce/my-account/templates/myaccount-after-delete-account.php';
+				}
+				return $template;
+			case 'myaccount/form-edit-account.php':
+				if ( isset( $_GET[ self::DELETE_ACCOUNT_FORM ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+					return dirname( NEWSPACK_PLUGIN_FILE ) . '/includes/reader-revenue/woocommerce/my-account/templates/myaccount-delete-account.php';
+				}
+				return dirname( NEWSPACK_PLUGIN_FILE ) . '/includes/reader-revenue/woocommerce/my-account/templates/myaccount-edit-account.php';
```
miguelpeixe: Same as the previous comment, I think we can use just `__DIR__`

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v0.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Newspack;
+
+use Newspack\WooCommerce_Connection;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Connection with WooCommerce's "My Account" page.
+ */
+class My_Account_UI_V0 extends WooCommerce_My_Account {
```
miguelpeixe: Why does this class extend `WooCommerce_My_Account`? If `My_Account_UI_XX` classes are meant to inherit shared utility, I think it should be its own `My_Account_UI` base class.

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -0,0 +1,101 @@
...
+	 * @param string $tag The tag.
+	 *
+	 * @return string The output.
+	 */
+	public static function add_newspack_ui_wrapper( $output, $tag ) {
+		if ( 'woocommerce_my_account' === $tag ) {
+			return '<div class="newspack-ui">' . $output . '</div>';
+		}
+		return $output;
+	}
```
miguelpeixe: Given that the goal is to have the My Account dashboard take over the entire page, what do you think of having `newspack-ui` as a body class instead?

miguelpeixe (changes requested):
The structure is looking great! I didn't look too much into the implemented styles because I understand it's still partial. My main focus was on how we're getting this foundation started.

I really like the directory structure for both backend and frontend. The v0 and v1 separation is a great idea and is implemented in a way that is very straightforward to figure out and navigate. Just left a minor comment for the PHP part.

In my testing I had no issues navigating the dashboard in both versions.

src/my-account/payment-methods.js
dkoo: Great point, `domReady` is right there. 😆 Updated in c6bd2e2c

src/my-account/v1/_navigation.scss
dkoo: Hard to tell from the Figma mockups. I think it looks better with hover styles matching the secondary button styles:

Screenshot 2025-05-09 at 4 24 35 PM

16822cdd

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: I tried that, but that ends up applying Newspack UI styles to more elements than we want (header, footer, etc.). Since we only want Newspack UI styles to apply to generalized components inside the post content, this was the easiest way I could figure out to do so without relying on theme-specific hooks.

dkoo: I think once we implement the full-page takeover in a separate PR, we might be able to safely move this class name to the body

includes/reader-revenue/woocommerce/my-account/class-woocommerce-my-account.php
dkoo: Good call. ae456e80 moves all of these files to `includes/plugins/woocommerce`.

includes/reader-revenue/woocommerce/my-account/class-woocommerce-my-account.php
dkoo: Updated in ae456e80

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Update in ae456e80

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v0.php
dkoo: They're extending `WooCommerce_My_Account` so they can both use constants and methods from that class as `self::`, but I suppose they could just call `WooCommerce_My_Account::` directly and not extend any class. I think the `My_Account_UI_XX` classes are separate enough that they don't need to share anything between them, so 2dd7f701 removes the extension and just makes these regular classes.

If they do eventually need to share some methods or constants, such as specific template parts, we can add the `My_Account_UI` base class.

dkoo:
Thanks for the feedback, @miguelpeixe! I'm glad the `_VX` convention made sense to you. 😆 I think this is ready for another review.

miguelpeixe (approved):
Thank you for the changes! Looks good and tests well!

src/my-account/v1/_navigation.scss
thomasguillot: The buttons should be "ghost"

thomasguillot: Screenshot 2025-05-12 at 16 21 15

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Keeping this thread open in case we want to revisit it in the future when we remove the header/footer from these pages.

src/my-account/v1/_navigation.scss
dkoo: Ah, thanks for clarifying, @thomasguillot! I've updated this to match in 67c9b7d8

dkoo:
@thomasguillot d1bc6519 also tweaks margins/spacing for the buttons inside tables:

mobile:
Screenshot 2025-05-12 at 12 01 03 PM

tablet/desktop:
Screenshot 2025-05-12 at 12 00 39 PM

thomasguillot (approved):
There are some vertical alignment issues but I'm not too concerned since this screen will eventually change.

includes/reader-revenue/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: https://github.com/Automattic/newspack-plugin/pull/3974#issuecomment-2877975641

PR #3952, opened 2025-04-25
feat(woocommerce): log error notices

miguelpeixe:
### All Submissions:

Log WooCommerce error notices, which is the primary error handling across Woo features. This includes My Account form validation, cart, and checkout issues. The payload will include the cart contents as `wc_cart`.

This PR also adds the user referrer and user agent to always be sent as part of the log payload.

1. Donate or purchase a subscription and use the following card number so it's declined: `4100000000000019`
2. Confirm you were unable to complete the checkout
3. Confirm a log entry was pushed with the following `error_code`: `newspack_woocommerce_error_notice`
4. Confirm the log entry has the following message: `Error: Your card was declined.` and the `extra` property includes the correct agent, referrer, and `wc_cart` information

### Other information:

### DIFF
--- a/includes/reader-revenue/woocommerce/class-woocommerce-logs.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-logs.php
@@ -26,6 +26,10 @@ class WooCommerce_Logs {
 	 * @param string $message The error message.
 	 */
 	public static function log_error_notices( $message ) {
+		// Only log if there is a message.
+		if ( empty( $message ) ) {
+			return $message;
+		}
 		$data = [
 			'wc_cart' => WC()->cart->get_cart_contents(),
 		];
--- a/includes/reader-revenue/woocommerce/class-woocommerce-logs.php
+++ b/includes/reader-revenue/woocommerce/class-woocommerce-logs.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Log WooCommerce events
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class for logging WooCommerce events.
+ */
+class WooCommerce_Logs {
+	/**
+	 * Initialize the class.
+	 */
+	public static function init() {
+		add_filter( 'woocommerce_add_error', [ __CLASS__, 'log_error_notices' ] );
+	}
+
+	/**
+	 * Log error notices.
+	 *
+	 * @param string $message The error message.
+	 */
+	public static function log_error_notices( $message ) {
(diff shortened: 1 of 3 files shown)


includes/reader-revenue/woocommerce/class-woocommerce-logs.php
```diff
@@ -0,0 +1,36 @@
...
+	public static function init() {
+		add_filter( 'woocommerce_add_error', [ __CLASS__, 'log_error_notices' ] );
+	}
+
+	/**
+	 * Log error notices.
+	 *
+	 * @param string $message The error message.
+	 */
+	public static function log_error_notices( $message ) {
```
dkoo, line 28: We may want to bail if `$message` is empty, otherwise it doesn't tell us much.

includes/reader-revenue/woocommerce/class-woocommerce-logs.php
```diff
@@ -0,0 +1,36 @@
...
+		$data = [
+			'wc_cart' => WC()->cart->get_cart_contents(),
```
dkoo, line 34: Is the cart info always relevant? Should we only send this data if the error happened during a checkout?

dkoo (changes requested):
I think this is a great idea! While testing this I noticed a handful of logs where `message` was empty, and I couldn't quite figure out where those were coming from, but I'm not sure it's useful to log in that case. I also think we should be more intentional about the data we attach to the log—cart info might not always be what we're looking for, as the hook can be called if an error notice is added during non-checkout My Account operations as well.

includes/reader-revenue/woocommerce/class-woocommerce-logs.php
miguelpeixe: It's additional session context that I don't think is harmful to include. Whether it has valuable information for a particular error is up to the person debugging.

I don't see value in adding code to not send it.

includes/reader-revenue/woocommerce/class-woocommerce-logs.php
miguelpeixe: Good call! Added in b128b89141507180e1b42f0700825de3ee95d396

includes/reader-revenue/woocommerce/class-woocommerce-logs.php
dkoo: I think the danger in my mind is that it could cause confusion for the person debugging. This info will be included as long as anything has been added to the cart at any point during the session and then not cleared out, so the error could have nothing to do with the cart/checkout and may send the person debugging down a rabbit hole. But maybe it's something we'll want to monitor for a bit to see if that happens, rather than preemptively solving for it.

dkoo (approved):

PR #3956, opened 2025-04-29
fix(ga4): ensure gate_post_id and newspack_popup_id are passed to activities

dkoo:
### All Submissions:

Fixes issues with `reader_registered` and `reader_logged_in` activities losing the `gate_post_id` and `newspack_popup_id` params in certain cases.

Also removes nonce verification for the Reader Registration block form, which seems to be resulting in failed registration attempts in some cases where whole-page HTML is served from a cache.

Follow testing instructions from #3895, preferably using a site that has edge caching enabled. Ensure that:

- After a successful registration by any means from inside a content gate or a Campaigns prompt, the GA events for `np_reader_registration` should have `gate_post_id` or `newspack_popup_id` params attached.
- Events for `np_reader_logged_in` should have `gate_post_id` params if triggered from a `#signin_modal` link inside a content gate, whether using password or OTP.

### Other information:

### DIFF
--- a/includes/data-events/class-memberships.php
+++ b/includes/data-events/class-memberships.php
@@ -43,6 +43,7 @@ final class Memberships {
 		add_action( 'woocommerce_checkout_create_order_line_item', [ __CLASS__, 'checkout_create_order_line_item' ], 10, 4 );
 		add_filter( 'newspack_register_reader_metadata', [ __CLASS__, 'register_reader_metadata' ] );
 		add_filter( 'newspack_auth_form_metadata', [ __CLASS__, 'register_reader_metadata' ] );
+		add_filter( 'newspack_otp_login_metadata', [ __CLASS__, 'register_reader_metadata' ] );
 		add_filter( 'newspack_register_reader_form_metadata', [ __CLASS__, 'register_reader_metadata' ] );
 		add_filter( 'newspack_newsletters_subscription_form_metadata', [ __CLASS__, 'register_reader_metadata' ] );
 	}
@@ -86,9 +87,14 @@ final class Memberships {
 	 */
 	public static function register_reader_metadata( $metadata ) {
 		$gate_post_id = filter_input( INPUT_POST, self::METADATA_NAME, FILTER_SANITIZE_NUMBER_INT );
-		if ( ! empty( $gate_post_id ) && isset( $metadata['registration_method'] ) ) {
-			$metadata['gate_post_id']        = $gate_post_id;
-			$metadata['registration_method'] = $metadata['registration_method'] . '-content-gate';
+		if ( ! empty( $gate_post_id ) && ( isset( $metadata['registration_method'] ) || isset( $metadata['login_method'] ) ) ) {
+			$metadata['gate_post_id'] = $gate_post_id;
+			if ( isset( $metadata['registration_method'] ) ) {
+				$metadata['registration_method'] = $metadata['registration_method'] . '-content-gate';
+			}
+			if ( isset( $metadata['login_method'] ) ) {
+				$metadata['login_method'] = $metadata['login_method'] . '-content-gate';
+			}
 		}
 		return $metadata;
 	}
--- a/includes/data-events/class-memberships.php
+++ b/includes/data-events/class-memberships.php
@@ -89,12 +89,6 @@ final class Memberships {
 		$gate_post_id = filter_input( INPUT_POST, self::METADATA_NAME, FILTER_SANITIZE_NUMBER_INT );
 		if ( ! empty( $gate_post_id ) && ( isset( $metadata['registration_method'] ) || isset( $metadata['login_method'] ) ) ) {
 			$metadata['gate_post_id'] = $gate_post_id;
-			if ( isset( $metadata['registration_method'] ) ) {
-				$metadata['registration_method'] = $metadata['registration_method'] . '-content-gate';
-			}
-			if ( isset( $metadata['login_method'] ) ) {
-				$metadata['login_method'] = $metadata['login_method'] . '-content-gate';
-			}
 		}
 		return $metadata;
 	}
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -2116,7 +2116,7 @@ final class Reader_Activation {
 		}
 
 		/**
-		 * Filters the metadata to be saved for a reader registered via the auth modal.
+		 * Filters the metadata to be saved for a reader going through the auth modal.
 		 *
 		 * @param array  $metadata Metadata.
 		 * @param string $email    Email address of the reader.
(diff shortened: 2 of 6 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -2116,10 +2112,25 @@ function_exists( 'wc_get_account_endpoint_url' ) ? \wc_get_account_endpoint_url(
 
 				$payload['registered']    = 1;
 				$payload['authenticated'] = 1;
-				return self::send_auth_form_response( $payload, false );
+				break;
 		}
-	}
 
+		/**
+		 * Filters the metadata to be saved for a reader registered via the auth modal.
```
miguelpeixe:
```suggestion
		 * Filters the metadata to be saved for a reader going through the auth modal.
```

src/blocks/reader-registration/index.php
```diff
@@ -377,20 +377,24 @@ function send_form_response( $data, $message = '' ) {
  */
 function process_form() {
 	// No need to process form values if Reader Activation is disabled.
-	if ( ! Reader_Activation::is_enabled() ) {
+	if ( ! Reader_Activation::is_enabled() || \is_user_logged_in() ) {
```
miguelpeixe: Will we ever have this running when the reader is logged in?

src/reader-activation-auth/auth-form.js
```diff
@@ -395,8 +402,12 @@ window.newspackRAS.push( function ( readerActivation ) {
 				}
 
 				if ( 'otp' === action ) {
+					const code = { code: body.get( 'otp_code' ) };
```
miguelpeixe: Maybe change this variable to something else, like `data` or `payload`?

includes/data-events/class-memberships.php
```diff
@@ -86,9 +87,14 @@ public static function checkout_create_order_line_item( $item, $cart_item_key, $
...
-			$metadata['gate_post_id']        = $gate_post_id;
-			$metadata['registration_method'] = $metadata['registration_method'] . '-content-gate';
+		if ( ! empty( $gate_post_id ) && ( isset( $metadata['registration_method'] ) || isset( $metadata['login_method'] ) ) ) {
+			$metadata['gate_post_id'] = $gate_post_id;
+			if ( isset( $metadata['registration_method'] ) ) {
+				$metadata['registration_method'] = $metadata['registration_method'] . '-content-gate';
+			}
+			if ( isset( $metadata['login_method'] ) ) {
+				$metadata['login_method'] = $metadata['login_method'] . '-content-gate';
+			}
```
miguelpeixe: I know this has already been merged in this format, but I think we can get more out of this if we use different parameters here instead of concatenating strings. WDYT?

Something like:
`login_method`: `otp`
`login_trigger`: `content-gate`

src/reader-activation/index.js
```diff
@@ -263,11 +263,13 @@ export function getOTPTimeRemaining() {
...
  *
- * @param {number} code OTP code.
+ * @param {Object} data                          Data.
+ * @param {number} data.code                     OTP code.
+ * @param {number} data.memberships_content_gate Gate post ID.
  *
  * @return {Promise} Promise.
  */
-export function authenticateOTP( code ) {
+export function authenticateOTP( { code, memberships_content_gate } ) {
```
miguelpeixe: I'm not a fan of this approach... Outside analytics, this metadata is not relevant to the authenticateOTP() function. If the analytics metadata grows, it can look pretty bad.

Can we do this in a more extensible way? Perhaps leveraging the store or cookies to pass this information along to the backend?

miguelpeixe (changes requested):

includes/data-events/class-memberships.php
dkoo: I agree that I'm not a fan of the string concatenation altering the values of `registration_method` and `login_method`. I might argue that adding the `-popup` or `-content-gate` suffix to the method is now no longer needed since we're always passing `newspack_popup_id` and `gate_post_id` params when the action originates from those triggers, WDYT @kmwilkerson?

src/reader-activation/index.js
dkoo: I'll think of a better way to do this

src/blocks/reader-registration/index.php
dkoo: It's not doing much, removed in c26838dd

includes/data-events/class-memberships.php
dkoo: ~1bf9122~ c26838d and https://github.com/Automattic/newspack-popups/pull/1431 propose removing these suffixes to avoid the string concatenation.

src/reader-activation-auth/auth-form.js
dkoo: Changed this approach in c26838dd, so it's no longer needed

src/reader-activation/index.js
dkoo: c26838dd avoids passing the gate post ID to the backend entirely by just looking for the presence of the `memberships_content_gate` data in the form data. Come to think of it, since all of the activities and the GA events they trigger are now on the front-end, there's really no reason to send this info the back-end at all, but I didn't want to refactor too much in this changeset. Let me know if you'd prefer to just rip off that band-aid instead.

includes/data-events/class-memberships.php
miguelpeixe: Maybe you forgot to push `1bf9122` or is it in c26838ddf631178c2055a21528c194325cff967b?

src/reader-activation/index.js
miguelpeixe: Awesome! Confirmed the gate ID is present on the OTP auth flow:

image

miguelpeixe (changes requested):
Thank you for the revisions!

Code looks good and working as per testing instructions.

I tested one other thing that didn't behave as I expected and I'm not sure if it's intentional. Signing in via a prompt doesn't include the popup ID in the reader_logged_in event like it does for gates. I'm ready to approve if that's expected.

includes/data-events/class-memberships.php
dkoo: Sorry, I meant c26838d

dkoo:
> I tested one other thing that didn't behave as I expected and I'm not sure if it's intentional. Signing in via a prompt doesn't include the popup ID in the reader_logged_in event like it does for gates. I'm ready to approve if that's expected.

I wasn't sure what to do for capturing logins via prompts. In almost every case a login from a prompt will come from the Registration block's "Sign into an existing account" button, which will simply open the auth modal. So there's no way to tell the difference between a login from clicking this button vs. logging in via the auth modal directly. So for now, this is expected, but I'm open to handling this in some other way (perhaps by adding a hidden input to the auth modal form if it's opened by clicking the button? But this seems like overengineering to me).

miguelpeixe (approved):
I agree!

I think we faced this before and at the time I also considered expanding the [modal hashes](https://github.com/Automattic/newspack-plugin/blob/74f77735a5135b004516f8d1217d8752d2c8fadd/src/reader-activation-auth/auth-modal.js#L2) to also hold additional parameters for such cases but didn't pursue...

PR #3957, opened 2025-05-02
fix: skip image error handler inside scripts

miguelpeixe:
### All Submissions:

Prevent the image error handler from replacing the markup strings inside script tags, causing a js crash.

1. While on trunk, create a post with an image and an HTML block with the following:

```
<script>
var markup = '<img src="https://picsum.photos/500" />';
</script>
```

2. Visit the page and confirm you get a js crash
4. Checkout this branch, refresh the page and confirm you don't get the crash
5. Confirm the error handler is present for the image but not for the img markup inside the script

### Other information:

### DIFF
--- a/includes/class-default-image.php
+++ b/includes/class-default-image.php
@@ -120,22 +120,37 @@ class Default_Image {
 	 * @return string Modified content with onerror attributes.
 	 */
 	public static function add_onerror_to_images( $content ) {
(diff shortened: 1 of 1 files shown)


adekbadek (commented):
Modifying content is always risky, here's an alternative that uses JS only: https://github.com/Automattic/newspack-plugin/pull/3963

PR #3961, opened 2025-05-06
feat: add Subscription Confirmation checkboxes for FTC compliance

laurelfulford:
### All Submissions:

This PR adds some Subscription options to the Audience > Configuration > Checkout & Payment page to help sites meet FTC compliance. 

This PR needs to be tested along with https://github.com/Automattic/newspack-blocks/pull/2117, which adjust some styles on the Newspack Blocks side of things.

See 1200550061930446-as-1208929637924603

1. Apply this PR and https://github.com/Automattic/newspack-blocks/pull/2117, and run `npm run build` for both. 
2. Make sure your test site is using WooCommerce as the Reader Revenue platform, and that WooCommerce Subscriptions is enabled. You'll need some non-subscription and subscription products set up to test
3. Navigate to Audience > Configuration > Checkout & Payment, and confirm there's a new Subscription section at the bottom. It should have two options, one for enabling a subscription confirmation checkbox, and one for enabling a Terms & Conditions checkbox. Both should be off by default:

CleanShot 2025-05-06 at 15 39 47

4. Toggle on the "Enable subscription confirmation checkbox", and confirm you now get a text area to edit the text label. It should be prepopulated with some default text:

CleanShot 2025-05-06 at 15 40 59@2x

5. Click 'Save Settings'.
6. On the front-end, run through a checkout using a subscription product -- this can be from the Checkout Button block, or a Monthly or Annual donation from the Donate Block. 
7. On the second screen of the checkout, confirm that you get the Subscription Confirmation checkbox at the bottom. The "Place Order" button should also be greyed out while this checkbox isn't selected:

8. Fill out some test payment details, check the Subscription Confirmation box, and confirm you can check out. 
9. Repeat the above with a non-subscription product and confirm that you don't get the Subscription Confirmation checkbox.
10. Navigate back to Audience > Configuration > Checkout & Payment, and try changing the text label for the Subscription Confirmation checkbox.
11. Click Save Settings, then repeat step 6 and 7 and confirm that your updated text appears.
12. Navigate back to Audience > Configuration > Checkout & Payment, and toggle on "Enable Terms & Conditions confirmation checkbox". "Enable subscription confirmation checkbox" should automatically toggle off, and its Label field should be hidden; a two new label fields should appear below the "Enable Terms & Conditions confirmation checkbox" toggle, one for a label that's populated, and one for a URL that's not:

CleanShot 2025-05-07 at 12 45 00@2x

13. Try clicking 'Save Settings' without making edits, and confirm you get an alert that you need to populate the URL:

14. Add a URL and click 'Save Settings'.
15. Retest steps 6 and 7 to confirm this checkbox now appears, including the link. Like with the Subscription Confirmation checkbox, the Place Order button should be greyed out and not let you submit the form until it's checked. When you click the link, it should open your T&C page in a new window:

15. Navigate back to Audience > Configuration > Checkout & Payment and try clearing all the text in either label field, and click 'Save Settings'. Confirm when you do this that the textfield is repopulated with the default text so it doesn't let you leave it empty.
16. Try changing the text label for the Terms & Conditions checkbox, including moving the link and click save. Confirm this is applied to the front-end.
17. Try changing the text label to remove the `{{ }}` entirely. Confirm that the whole label is now linked: 

18. Lastly, make sure you can complete a checkout, including checking the Terms & Conditions box. 
19. You'll need a non-donation subscription product to test this next bit: navigate to the /shop page on your site, and run through a checkout. Confirm that you see your Terms & Conditions checkbox at the bottom of the checkout page (though it doesn't have any form-blocking JS fanciness).

CleanShot 2025-05-07 at 10 42 41@2x

20. Try to complete your checkout without checking the checkbox and confirm that you get a standard WooCommerce error and that you can't complete the purchase:

CleanShot 2025-05-06 at 15 58 03@2x

22. Repeat the above with a non-subscription product (and make sure you don't have any other subscription products in the cart anymore). Confirm you don't get the checkbox at the bottom.
23. Lastly, navigate back to Audience > Configuration > Checkout & Payment and repeat the /shop testing steps with the Subscription Confirmation checkbox enabled. 

### Other information:

---
  - https://app.asana.com/0/0/1208929637924603

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
@@ -59,7 +59,7 @@ class Subscriptions_Confirmation {
 
 		if ( Reader_Activation::is_terms_confirmation_enabled() ) {
 			if ( ! isset( $_POST['newspack_subscription_terms_confirmation'] ) || '1' !== $_POST['newspack_subscription_terms_confirmation'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
-				wc_add_notice( esc_html__( 'You must agree to the terms & conditions before proceeding.', 'newspack-plugin' ), 'error' );
+				wc_add_notice( esc_html__( 'You must agree to the Terms & Conditions before proceeding.', 'newspack-plugin' ), 'error' );
 			}
 		}
 	}
@@ -84,6 +84,7 @@ class Subscriptions_Confirmation {
 		}
 		return $label;
 	}
+
 	/**
 	 * Add either the Subscription Confirmation or Terms & Conditions checkbox to the WooCommerce checkout form when enabled and when the cart contains a subscription product.
 	 */
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
@@ -90,7 +90,7 @@ class Subscriptions_Confirmation {
 				array(
 					'type'     => 'checkbox',
 					'class'    => array( 'form-row-wide', 'newspack-subscription-confirmation-checkbox' ),
-					'label'    => '<span>' . Reader_Activation::get_terms_confirmation_text() . ' <a href="' . esc_url( Reader_Activation::get_terms_confirmation_link() ) . '">' . esc_html__( 'Read More', 'newspack-plugin' ) . '</a></span>',
+					'label'    => Reader_Activation::get_terms_confirmation_text(),
 					'required' => true,
 				)
 			);
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
@@ -18,7 +18,7 @@ class Subscriptions_Confirmation {
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
-		if ( ! WooCommerce_Subscriptions::is_active() ) {
+		if ( ! class_exists( 'WooCommerce_Subscriptions' ) || ! WooCommerce_Subscriptions::is_active() ) {
 			return;
 		}
 		add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_subscription_confirmation_checkboxes' ] );
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
@@ -18,7 +18,7 @@ class Subscriptions_Confirmation {
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
-		if ( ! class_exists( 'WooCommerce_Subscriptions' ) || ! WooCommerce_Subscriptions::is_active() ) {
+		if ( ! class_exists( '\Newspack\WooCommerce_Subscriptions' ) || ! WooCommerce_Subscriptions::is_active() ) {
 			return;
 		}
 		add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_subscription_confirmation_checkboxes' ] );
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
@@ -21,8 +21,8 @@ class Subscriptions_Confirmation {
 		if ( ! WooCommerce_Subscriptions::is_enabled() ) {
 			return;
 		}
-		add_action( 'woocommerce_review_order_before_submit', [ __CLASS__, 'add_subscription_confirmation_checkbox' ] );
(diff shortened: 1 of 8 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -2650,17 +2654,66 @@ public static function get_post_checkout_registration_success_text() {
...
+
+	/**
+	 * Get the text label for the Terms & Conditions confirmation checkbox.
+	 *
+	 * @return string Returns either the default text label or a customized one.
+	 */
+	public static function get_terms_confirmation_text() {
+		return \get_option(
+			self::OPTIONS_PREFIX . 'woocommerce_terms_confirmation_text',
+			__( 'I have read and accept the <a href="#">Terms & Conditions</a>.', 'newspack-plugin' )
```
adekbadek: I see potential for user error here – the user might think toggling the checkbox is enough. There should be some validation to make sure that there's an actual link there. Also – if it makes sense – the legal pages URL set in Audience Management could be used as the default.

adekbadek: The link must have a `target="_blank"` attribute forced. Otherwise, the link will open inside of the checkout modal, without any way to get back to the checkout flow.
Better yet, the link could open in the modal, if there would be a button to get back to the checkout flow.

laurelfulford: > I see potential for user error here... There should be some validation to make sure that there's an actual link there.

Good point! Now I'm also wondering about just errors in the HTML itself 😬 

@thomasguillot What do you think about having a separate field for the link and then put the customized text and an uneditable link and text ("Read Terms & Conditions"?) together in the checkbox label?

thomasguillot: That's a good idea. I just have one question... how would you know which part of the text should have the link? Also Adam has got a point with target blank

laurelfulford: >  Better yet, the link could open in the modal, if there would be a button to get back to the checkout flow.

Yeah it'd be nice to smooth this out a bit! 

WooCommerce's baked in Terms & Conditions kind of does this -- it opens the text of the Terms & Conditions page inside of the checkout like this (I just have the page set to 'Sample Page' so the text is a little weird): 

CleanShot 2025-05-07 at 09 03 32

Flagging for @thomasguillot as well for the visual end of this. I can start with the `_blank` part at least to make sure people can get back to the checkout easier!

laurelfulford: We discussed this a bit in Slack and landed on wrapping the text in `{{ }}`. I added that in https://github.com/Automattic/newspack-plugin/pull/3961/commits/27743883462b9a8fad4ae49b9f1968caad84fb23 (along with the `_blank` target, and if the `{{ }}` is missing from the text, I made the whole thing be linked.

laurelfulford: Went with `_blank` for this one after discussing in Slack!

laurelfulford:
Thanks for your initial feedback here, @adekbadek! I left you assigned but feel free to unassign if you don't want to test again.

I made some changes based on your feedback and some discussion with @thomasguillot in Slack, and updated the steps and screenshots: 

* The Terms & Conditions are now handled separately as a text label and URL field. 
* Publishers can wrap text in `{{ }}` to add the link to specific text in the label; if that's missing, the whole label is linked. 
* The link opens in a new window. 
* I opted to leave the URL field blank by default. I like the idea of prepopulating it from the Legal Pages settings for Audience Development if set up, but the T&C linked here needs to include specific information to meet the FTC guidelines. Out of an abundance of caution, I think leaving it blank and forcing the publisher to populate it before using this makes sure they're picking a page that meets the requirements (well, hopefully!).

includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Main class.
+ */
+class Subscriptions_Confirmation {
+
+	/**
+	 * Initialize hooks and filters.
+	 */
+	public static function init() {
+		if ( ! WooCommerce_Subscriptions::is_active() ) {
```
claudiulodro:
```suggestion
		if ( ! class_exists( 'WooCommerce_Subscriptions' ) || ! WooCommerce_Subscriptions::is_active() ) {
```

claudiulodro (approved):
Code looks solid and this works well in both modal and traditional checkout in my testing! Left one tiny safety suggestion :)

claudiulodro:
We should also test this on a certain site using the Elex address validation thing, just to make sure there won't be surprises :)

adekbadek (approved):
All good, barring @claudiulodro 's suggestion which should be implemented.

includes/plugins/woocommerce-subscriptions/class-subscriptions-confirmation.php
laurelfulford: Yes! Fixed in [d009e0f](https://github.com/Automattic/newspack-plugin/pull/3961/commits/d009e0fcf2de1cbff0d0f70d2c606c8cd1aec853).

laurelfulford:
Thanks @adekbadek and @claudiulodro! 🙌 I'll hold off merging until I confirm the last bit of third party testing 🙂

PR #3964, opened 2025-05-07
fix(jetpack): handle the related posts max age option

adekbadek:
### All Submissions:

The option to change the Jetpack Related Posts feature maximum age was not handled. This PR reinstates the handling.

A casualty of the https://github.com/Automattic/newspack-plugin/pull/3857 epic PR. 

_Look in Asana for (alternative) internal testing instructions._ 

1. On `trunk`,
1. On a site with Jetpack connected, visit Newspack → Settings → Display Settings
2. Enable "Related Posts" and set Maximum Age Of Related Content In Months to “6”
3. View a post or execute `(new Jetpack_RelatedPosts())->get_for_post_id(<post-id>, []);`, observe that you may get posts older than 6 months 
4. Switch to this branch, observe that only posts published less than 6 months ago are returned

### Other information:

---
  - https://app.asana.com/0/0/1210134334697036

### DIFF
--- a/includes/plugins/class-jetpack.php
+++ b/includes/plugins/class-jetpack.php
@@ -31,37 +31,37 @@ class Jetpack {
 	 * @var string[]
 	 */
 	public static $default_active_modules = [
-		/** 
+		/**
 		 * Assets CDN
 		 *
 		 * @link https://jetpack.com/support/site-accelerator/
 		 */
 		'photon-cdn',
-		/** 
+		/**
 		 * Image CDN
 		 *
 		 * @link https://jetpack.com/support/site-accelerator/
 		 */
 		'photon',
-		/** 
+		/**
 		 * Contact Form
 		 *
 		 * @link https://jetpack.com/support/contact-form/
 		 */
 		'contact-form',
-		/** 
-		 * Brute Force Protection. 
+		/**
+		 * Brute Force Protection.
 		 *
 		 * @link https://jetpack.com/support/protect/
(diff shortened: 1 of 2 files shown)


includes/plugins/class-jetpack.php
```diff
@@ -307,12 +310,28 @@ public static function remove_google_analytics_from_available( $modules ) {
...
 	}
+
+	/**
+	 * Restrict the age of related content shown by Jetpack Related Posts.
+	 *
+	 * @param array $date_range Array of start and end dates.
+	 * @return array Filtered array of start/end dates.
+	 */
+	public static function restrict_age_of_related_posts( $date_range ) {
+		$related_posts_max_age = get_option( ( new \Newspack\Wizards\Newspack\Recirculation_Section() )->related_posts_option );
```
miguelpeixe: I think we can change the `related_posts_option` to a class constant, it's how we usually treat these, and it avoids instantiating a class that, IMO, should be a singleton.

We're also in the Newspack namespace here, you can call `Wizards\Newspack\Recirculation_Section`

miguelpeixe (changes requested):
Works! A small request inline

includes/plugins/class-jetpack.php
adekbadek: Thanks, that's indeed more elegant. Updated in f2a326c

miguelpeixe (approved):

PR #3965, opened 2025-05-08
fix(ga4): fire login/registration activities via SSO

dkoo:
### All Submissions:

Plugs a hole in the `reader_logged_in` and `reader_registered` activities/GA4 events implemented in #3895. Logins and registrations via SSO were not firing these activities, and thus not triggering GA4 events to track these user interactions.

This PR ensures that logins and registrations via SSO will fire these activities and events. It also adds a new `sso` boolean parameter which is only set if the activity originated via an SSO flow. It also refactors the front-end analytics handlers to reduce some code duplication.

1. Check out this branch.
2. In incognito sessions, register new accounts using the "Sign in with Google" button, using both the Registration block and in the Sign In modal.
3. Confirm that `reader_registered` reader activities are fired (look in Dev Tools > localStorage > `np_reader_1_activity`) and `np_reader_registered` GA4 events are fired with relevant params, including the new `sso` param.
4. In new incognito sessions, log into the accounts created in step 2 using the "Sign in with Google" button, using both the Registration block and in the Sign In modal.
5. Confirm that `reader_logged_in` reader activities are fired (look in Dev Tools > localStorage > `np_reader_1_activity`) and `np_reader_logged_in` GA4 events are fired with relevant params, including the new `sso` param.
6. Smoke test other GA4 events as described in #3895 and confirm that there are no other changes in those events' behavior or data.

### Other information:


src/reader-activation/analytics.js
```diff
@@ -1,18 +1,27 @@
...
+	if ( data?.newspack_popup_id ) {
+		payload.newspack_popup_id = data.newspack_popup_id;
+	}
+	if ( data?.gate_post_id ) {
+		payload.gate_post_id = data.gate_post_id;
+	}
+	if ( data?.sso ) {
+		payload.sso = data.sso;
+	}
+	return payload;
```
miguelpeixe: Can we change this to match what was [recently merged to trunk](https://github.com/Automattic/newspack-plugin/blob/22e959022980ac27b3f5cb8b4b3ba1a0c969edb9/src/reader-activation/analytics.js#L1-L18)? We're also no longer adding the `referrer`.

miguelpeixe (changes requested):
The event parameters look good for both auth modal and block, but I'm unable to get the SSO to trigger a `reader_registered` event. It's always `reader_logged_in` regardless of account creation.

Also, not sure if related to this PR, but creating an account via SSO is not logging me in. Only after I SSO again (with the account created) that I'm authenticated.

src/reader-activation/analytics.js
dkoo: Yes, good call! d31c04e6 updates `getEventPayload` so it matches what's in `trunk`.

dkoo:
> The event parameters look good for both auth modal and block, but I'm unable to get the SSO to trigger a `reader_registered` event. It's always `reader_logged_in` regardless of account creation.

8bd85c7d fixes this for both the Registration block and Sign In modal—now you should be able to get `reader_registered` events when registering, and `reader_logged_in` when logging in.

> Also, not sure if related to this PR, but creating an account via SSO is not logging me in. Only after I SSO again (with the account created) that I'm authenticated.

I'm not able to replicate this. After registering for a new account via SSO in either case, I'm logged in. Maybe it's related to https://github.com/Automattic/newspack-popups/pull/1436? I noticed that without that fix, I was seeing strange behavior in the reader data store.

dkoo:
@miguelpeixe I rebased to `trunk`. Looks like we lost the login/registration/newsletter events there somehow (they are still in `release`), so this PR now restores them.

miguelpeixe:
The events weren't lost, they migrated to the new strategy using the `registerActivityEvent()`:

https://github.com/Automattic/newspack-plugin/blob/22e959022980ac27b3f5cb8b4b3ba1a0c969edb9/src/reader-activation/analytics.js#L65-L79

dkoo:
@miguelpeixe whoops, I forgot, sorry! 8bb8e404 removes them again and also adds a missing piece of data to the payload when logging in using a password.

miguelpeixe (approved):
Nice! Thank you for the revisions

PR #3971, opened 2025-05-12
fix: make sure fix duplcate fields apply filters

leogermani:
### All Submissions:

The fix duplicate scripts is not accountig for fields added via a filter. This PR fixes it with as little refactor as possible.

1. Activate Newspack Network
2. Add some debug to the `get_fields_to_check_for_duplicates` method
3. On `trunk` confirm that the `NP_Network Registration Site` field is not present
4. On this branch, confirm it is present - with the `NP_` prefix.

### Other information:

### DIFF
--- a/includes/cli/class-mailchimp.php
+++ b/includes/cli/class-mailchimp.php
@@ -450,11 +450,17 @@ class Mailchimp {
 	 * @return array
 	 */
 	private static function get_fields_to_check_for_duplicates() {
+		/**
+		 * Apply the filters that allow other plugins to add their own fields.
+		 *
+		 * This filter is documented in Newspack\Reader_Activation\Sync\Metadata::get_keys().
+		 */
+		$fields = \apply_filters( 'newspack_ras_metadata_keys', Metadata::get_all_fields() );
 		$fields = array_map(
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works, so approving! But nit: couldn't we just use `Metadata::get_keys()` instead of `Metadata::get_all_fields()` and avoid having to reapply the `newspack_ras_metadata_keys` filter?

leogermani:
> This works, so approving! But nit: couldn't we just use `Metadata::get_keys()` instead of `Metadata::get_all_fields()` and avoid having to reapply the `newspack_ras_metadata_keys` filter?

@dkoo thanks!

`get_keys` has this: 

```
$fields = Donations::is_platform_wc() ? self::get_all_fields() : self::get_basic_fields();
```

So if the donation platform is not WC, we won't get all the fields... I know this should not be a problem in most cases, but I just wanted to make sure that the CLI command will always get ALL the fields no matter what, even if the site is not properly configured.

We could refactor things on the Metadata class to avoid having to reapply the filter, but I think that would be more complex, and this is just a CLI.. so I thought it was fine

PR #3972, opened 2025-05-13
feat: add compatibility to network in custom bylines

leogermani:
### All Submissions:

This makes custom bylines work with Newspack Network.

When the post is being distributed, it will add a mapping with user IDs and emails to the post meta of the distributed post. On the other end, the target site will modify the custom byline to make sure it points to the correct users on the other side.

1. Create a new post and distribute it (or use an existing one)
2. Add a custom Byline and save
3. Check that in the target site the post displays the correct byline, pointing to the correct author
4. Confirm that the ID of that author is different in each site, but that the `[author id=x` shortcode points to different IDs in each site

### Other information:

### DIFF
--- a/includes/bylines/class-bylines.php
+++ b/includes/bylines/class-bylines.php
@@ -37,6 +37,10 @@ class Bylines {
 		add_filter( 'pre_newspack_posted_by', [ __CLASS__, 'pre_newspack_posted_by' ] );
 		add_filter( 'newspack_blocks_post_authors', [ __CLASS__, 'newspack_blocks_post_authors' ] );
 		add_filter( 'newspack_blocks_post_byline', [ __CLASS__, 'newspack_blocks_post_byline' ] );
+
+		// Newspack Network compatibility.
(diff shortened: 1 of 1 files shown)


includes/bylines/class-bylines.php
```diff
@@ -294,5 +298,69 @@ public static function newspack_blocks_post_byline( $byline ) {
...
+	 * @param int   $post_id   The post ID.
+	 * @param bool  $is_linked Whether the post is linked.
+	 * @param array $payload   The post payload.
+	 */
+	public static function newspack_network_incoming_post_inserted( $post_id, $is_linked, $payload ) {
+		if ( ! $is_linked ) {
+			return;
+		}
+
+		$byline = \get_post_meta( $post_id, self::META_KEY_BYLINE, true );
```
naxoc: Small nit: The mixed use of backslashed WP functions is a bit confusing. I don't have a preference for either, but I do like it consistent.

naxoc (approved):
I tested this and it looks good. 

My nit is non-blocking :)

PR #3974, opened 2025-05-13
feat(my-account): full-site takeover template and custom nav menu

dkoo:
### All Submissions:

This PR builds on the Newspack UI styles implemented in #3951. It implements a custom no-header/no-footer page template for My Account pages while logged in, as well as a custom template part, styles, and JS for the My Account nav menu.

Closes `NPPD-107` and `NPPD-106`.

<table>
<tr>
<td>
mobile, menu closed
Screenshot 2025-05-13 at 12 33 06 PM
</td>
<td>
mobile, menu open
Screenshot 2025-05-13 at 12 33 00 PM
</td>
</tr>
<tr>
<td>
mobile, horizontal orientation (menu and content are separately scrollable)
Screenshot 2025-05-13 at 5 28 26 PM
</td>
<td>
desktop
Screenshot 2025-05-13 at 12 32 35 PM
</td>
</tr>
</table>

Some notes:

- The custom no-header/no-footer page template is based on the `no-header-footer.php` template from Newspack Theme but implemented separately in this plugin so that it's not reliant on the Newspack Theme being active. I'm open to discussing alternative approaches, but this seemed cleanest to me.
- The site logo shown in the desktop nav menu is the image chosen as the "Site icon" in the Customizer (also filterable via `newspack_my_account_site_logo_url` hook).
- The designs include an area for custom links above the Sign Out link, but the UI to add custom links is out of scope of this project. They can be added in code via a `newspack_my_account_navigation_footer_items` filter hook.
- (for @Automattic/newspack-design) Since this isn't the "true" full-screen takeover experience (where the My Account pages functions as an app overlay on the regular site), I took some liberty with adding a mobile "account" icon button on the right side of the fixed topbar element to toggle the menu open. Does the placement and behavior make sense?

1. Check out this branch and rebuild assets
2. Make sure `define( 'NEWSPACK_MY_ACCOUNT_VERSION', '1.0.0' );` is defined in wp-config.php
3. Add a site icon image in **Customize > Site Identity**
4. Log into My Account with a reader account and confirm that all My Account pages render without a site header/footer while logged in
5. Test look and behavior of the menu at various breakpoints and screen sizes
6. (Bonus) add some custom menu links using a filter and ensure that the links render above the "Sign out" link at the bottom of the menu sidebar:

```php
add_filter(
	'newspack_my_account_navigation_footer_items',
	function() {
		return [
			[
				'url'   => 'https://automattic.com',
				'label' => __( 'Automattic', 'newspack-plugin' ),
			],
			[
				'url'   => 'https://newspack.com',
				'label' => __( 'Newspack', 'newspack-plugin' ),
			],
		];
	}
);
```

### Other information:


dkoo:
This moves the `.newspack-ui` class to the `body` element as suggested [here](https://github.com/Automattic/newspack-plugin/pull/3951#discussion_r2080403294), since we no longer need to worry about Newspack UI styles bleeding into non-content elements.

thomasguillot:
The sign out button should be a `ghost` variant.

thomasguillot:
Screenshot 2025-05-15 at 10 59 48@2x

It looks broken if it's the Block Theme but I guess it's because right now the theme doesn't even support WC.

dkoo:
@thomasguillot maybe we can tackle Block Theme support in a separate PR? We may need to tweak the page template to accommodate it, too.

dkoo:
@thomasguillot e55fc8ff fixes hover styles for the Sign Out button:

Screenshot 2025-05-15 at 11 32 49 AM

thomasguillot:
> maybe we can tackle Block Theme support in a separate PR?

Sounds good.

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -19,10 +19,52 @@ class My_Account_UI_V1 {
...
+
+	/**
+	 * Render My Account pages with a no-header/no-footer page template.
+	 *
+	 * @param string $template The template.
+	 * @return string The template file path.
+	 */
+	public static function page_template( $template ) {
+		if ( function_exists( 'is_account_page' ) && \is_account_page() && \is_user_logged_in() ) {
+			return __DIR__ . '/templates/v1/myaccount.php';
```
miguelpeixe:
```suggestion
			return __DIR__ . '/templates/v1/my-account.php';
```

I think we should follow Woo's naming here.

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -19,10 +19,52 @@ class My_Account_UI_V1 {
 	 * @codeCoverageIgnore
 	 */
 	public static function init() {
+		\add_filter( 'page_template', [ __CLASS__, 'page_template' ] );
+		\add_filter( 'newspack_ads_should_show_ads', [ __CLASS__, 'suppress_ads' ] ); // Suppress ads on My Account pages.
+		\add_filter( 'newspack_popups_assess_has_disabled_popups', [ __CLASS__, 'suppress_popups' ] ); // Suppress popups on My Account pages.
```
miguelpeixe: I'm surprised we didn't have that already! I think this should be version-agnostic and under `includes/plugins/woocommerce/my-account/class-woocommerce-my-account.php`

miguelpeixe (changes requested):
Nice! All works well in my testing. Left a couple of minor requests below.

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: They actually use both `myaccount` and `my-account` in various places, but notably `myaccount` for [template parts](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/templates/myaccount) vs `my-account` for the shortcode and permalinks. Not sure if there's any logic behind the difference, though. I'm okay with picking one and sticking to it in our code, though.

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Yeah, I also thought it was strange that we weren't doing it already! I initially put this in v1 only in case any publishers are actively pushing ads or prompts in My Account pages, but I think I agree that no one should be. Updated in bda9a9e9.

dkoo: efc976c and 9c9c2d9 remove the `myaccount-` prefixes from the custom template parts—the prefix is redundant IMO since they're inside a `my-account` directory now. The `myaccount.php` page template is now `my-account.php` to match our own naming convention.

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
miguelpeixe: I was thinking of the [`my-account.php` file](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/myaccount/my-account.php), but you're right, they also use `myaccount` for the directory 🤔 

I'll let you decide 😄

dkoo: efc976c and 9c9c2d9 remove the `myaccount-` prefixes from the custom template parts—the prefix is redundant IMO since they're inside a `my-account` directory now. The `myaccount.php` custom page template is now `my-account.php` to match our own naming convention.

miguelpeixe (approved):

PR #3977, opened 2025-05-14
fix: remove non-existent filter

adekbadek:
3b7ee8ebd3c1 was a bad merge, it re-added a line that was removed in 59296184c013

Related: https://github.com/Automattic/newspack-plugin/pull/3976

## To test

1. On `trunk`,
1. Try to make a donation using the Donate block, with Stripe gateway
2. Observe it failing, switch to this branch, test again


adekbadek:
This will make it into `trunk` via https://github.com/Automattic/newspack-plugin/pull/3976

PR #3985, opened 2025-05-20
feat(my-account): custom confirmation modal flows

dkoo:
### All Submissions:

Implements custom UI for the following flows for My Account v1:

- Reset password: Instead of the usual core WP password reset flow via email, the reader should be able to enter their current password (if it exists) along with the new password right in the Account Settings page to reset their password.
- Cancel subscription: Show a styled confirmation modal instead of the native browser dialogue.
- Delete account: Show a series of styled confirmation modals instead of the native browser dialogue.

To facilitate these new flows using Newspack UI elements, implements some new Newspack UI abstractions:

- A `Newspack_UI::generate_modal()` method which will generate markup and some generalized JS behavior for a styled modal given an array of options (see docblock for supported args)
- A new "snackbar" notification component that will appear in a floating fixed position in a corner of the page and disappear automatically after five seconds (see the Newspack UI components demo to see an example).

Closes `NPPD-108`.

1. With `NEWSPACK_MY_ACCOUNT_VERSION` either unset or set to `0.0.0` in wp-config.php, smoke test the flows to reset password, cancel subscriptions, and delete account as a reader. There should be no changes in these flows.
2. Set `NEWSPACK_MY_ACCOUNT_VERSION` to `1.0.0` and repeat all three flows with a new reader account. More details below.

#### New Account Settings page template and password reset flow

1. After verifying the account, confirm that the template for Account Settings is now slightly different:
  - There should be three page sections, each with a header (Profile, Password, and Delete Account)
  - Profile should include fields for first + last names, display name, and email address

2. Test updating Profile fields and confirm that all fields other than email address are now optional.
3. For the Password section, confirm that since the reader account is brand new and hasn't yet set a password, the only fields shown are "New password" and "Re-enter new password" with the CTA in the button saying "Set password":

Screenshot 2025-05-21 at 9 53 54 AM

4. Enter a weak password (e.g. only letters) and confirm that it triggers Woo's standard password strength indicator and blocks saving the new password unless a strong password is entered:

Screenshot 2025-05-21 at 9 54 12 AM

5. Confirm that the new password can't be saved unless both password fields are filled out with the same strong password.
6. Save the new password with matching fields and confirm that the Password section now has an additional field for "Current password", and that the CTA in the button now says "Update password":

Screenshot 2025-05-21 at 9 56 27 AM

7. Confirm that you can't update the password now unless all conditions are met:

  - The "Current password" field contains the correct current password
  - Both "new password" fields contain the same strong password string

8. Click "Forgot password" and confirm that it initiates the usual email-based flow to reset the password.
9. After clicking the "Reset password" link from the email, confirm that it leads to the Account Settings page but with this modal automatically open to reset your password:

Screenshot 2025-05-21 at 5 40 33 PM

10. Confirm that the password inputs in the modal must contain a strong password to submit.
11. Submit the modal with mismatched password strings and confirm that you get a "passwords do not match error". Confirm that clicking "Create a password" again at this point opens the reset password modal instead of forcing you to send a new email.
12. Submit the modal with matching strong password strings and confirm that your password is updated successfully.
13. Click on the "Reset password" link in the email from step 9 and confirm that the link is no longer valid:

Screenshot 2025-05-21 at 5 44 09 PM

#### Cancel subscription

1. Purchase a subscription as the reader and then visit My Account > My Subscription.
2. Click the "Cancel" button in the subscription details table and confirm it shows this styled modal instead of the native browser UI:

Screenshot 2025-05-20 at 5 36 00 PM

3. Confirm that the end date in the modal matches the scheduled next payment date for the subscription (which is when a pending-cancellation subscription would transition to "cancelled" status).
4. Confirm that the × close button and the "Keep subscription" button both simply close the modal without doing anything.
5. Confirm that clicking "Cancel subscription" completes the cancellation and switches the subscription to "pending-cancellation" status.

#### Delete account

1. Go back to Account Settings and click the big red "Delete account" button. Confirm that it shows this styled modal (note that it doesn't contain the "Manage donations" section as specified in the [designs](https://www.figma.com/design/mkvHE3qozmmGrytPt9RGrV/My-Account?node-id=2636-46786&m=dev) because the separate My Account pages for managing donations vs. subscriptions is out-of-scope of this PR):

<table>
<tr>
<td>
desktop:
Screenshot 2025-05-20 at 5 36 23 PM
</td>
<td>
mobile:
Screenshot 2025-05-21 at 10 18 38 AM
</td>
</table>

2. Confirm that the "Manage subscriptions" and "Manage newsletters" buttons both link out to the correct My Account pages for those areas.
3. Confirm that the × close button and the "Cancel" button both simply close the modal without doing anything.
4. Confirm that clicking "Delete account" results in this second modal being displayed, and that it triggers the "Delete your account" email to be sent to the reader account:

Screenshot 2025-05-20 at 5 36 33 PM

5. Click the link from the email and confirm that you land on the Account Settings page with this confirmation modal automatically open:

Screenshot 2025-05-20 at 5 36 49 PM

6. Confirm that the × close button and the "Cancel" button both simply close the modal without doing anything.
7. Click the "Delete account" button again and confirm that it shows the modal from step 5 again (instead of making you send a new "Delete your account" email).
8. Click "Delete account" inside the modal and confirm that you get redirected to the homepage in a logged-out state, but with this new Newspack UI "snackbar" notification telling you that your account has been deleted (note that the snackbar will disappear automatically after 5 seconds):

Screenshot 2025-05-20 at 5 37 05 PM

14. Confirm that the account has actually been deleted on the back-end, and all of the usual side effects have occurred (active subscriptions cancelled, contact unsubscribed in the ESP).

### Other information:


dkoo:
Tagging @Automattic/newspack-design for a review here as well

miguelpeixe:
Maybe something happened on rebase? "Account settings" is rendering blank for me.

dkoo:
@miguelpeixe yes, I was missing some renamed templates after the merge from `trunk`. Should be fixed after 92bbd71c

includes/plugins/woocommerce/my-account/templates/v1/account-settings.php
```diff
@@ -0,0 +1,276 @@
...
+			<input
+				type="password"
+				class="woocommerce-Input woocommerce-Input--text input-text"
+				name="current_password"
+				id="current_password"
+				required
+				aria-required="true"
+			/>
+		</p>
+		<?php endif; ?>
```
miguelpeixe: I'm still working on my review, but I'd like to point this out before going through everything.

For security reasons, we shouldn't allow password changes without re-authentication. If the password has not yet been set, I suggest we make the reader go through the regular password reset flow via email (what we have now)

dkoo: Good point, and I also realize I missed a custom modal to actually reset the password in the email-based flow, too. Will work on a fix for this flow

dkoo: 98bab4d8 forces the user to go through the email-based flow if they've never set a password before:

Screenshot 2025-05-21 at 4 07 47 PM

Digging in a little, it's more complicated to bypass Woo's default "reset password" form since it's handled via redirects. I might work on that in a separate PR

dkoo: 46ce0a4b should do it, actually. 31cf3df and 663684f also ensure that the password reset key check from core Woo still works (so that you can't use the same "Reset password" link more than once). Updated the testing instructions.

thomasguillot:
Screenshot 2025-05-22 at 10 36 59@2x

Most likely an edge case but for some reason the htlm in this error message isn't displayed properly

thomasguillot:
Screenshot 2025-05-22 at 10 55 46@2x

The gap above "New password" is too big

dkoo:
@thomasguillot thanks for the changes and feedback!

> Most likely an edge case but for some reason the htlm in this error message isn't displayed properly

We shouldn't allow HTML tags to be displayed in these messages. cce3040 strips tags everywhere we're setting the message via URL param.

> The gap above "New password" is too big

7e19550 removes the `.woocommerce-form-row` classes from the paragraphs inside the modal form, which was adding the extra space. f1eac84 moves the styles for `.woocommerce-password-strength` outside of that parent class so the styles look the same both inside and outside the modal.

Screenshot 2025-05-22 at 10 14 11 AM
Screenshot 2025-05-22 at 10 15 58 AM

thomasguillot:
Thank you @dkoo

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -24,6 +33,16 @@ public static function init() {
 		\add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_assets' ], 11 );
 		\add_filter( 'wc_get_template', [ __CLASS__, 'wc_get_template' ], 10, 5 );
 		\add_filter( 'woocommerce_account_menu_items', [ __CLASS__, 'my_account_menu_items' ], 1001 );
+		\add_filter( 'newspack_myaccount_required_fields', [ __CLASS__, 'account_settings_required_fields' ] );
+		\add_action( 'wp_loaded', [ __CLASS__, 'maybe_generate_password_reset_key' ] );
+		\add_action( 'template_redirect', [ __CLASS__, 'redirect_reset_password_link' ], 11 );
+		\add_filter( 'validate_password_reset', [ __CLASS__, 'validate_password_reset' ], 10, 2 );
+		\add_action( 'wp_loaded', [ __CLASS__, 'maybe_password_reset_success' ] );
+		\add_action( 'newspack_woocommerce_after_edit_account_form', [ __CLASS__, 'add_reset_password_modal' ] );
```
miguelpeixe: We have a lot of code just for password reset management. WDYT we put this in a separate class? This is going to grow a lot more as the project progresses.

includes/class-newspack-ui.php
```diff
@@ -64,13 +66,130 @@ public static function colors_css_wrap() {
...
+	 *               @type string $method The HTTP method to use.
+	 *               @type string $nonce The nonce to use for the request.
+	 *               @type array $body The body to send with a POST request.
+	 *               @type array $params Parameters to send with a GET request.
+	 *               @type string $next The ID for another modal to open after the action is completed.
+	 *         }
+	 *     }
+	 * }
+	 */
+	public static function generate_modal( $args ) {
```
miguelpeixe, line 96: I think not for this PR, but we could benefit from making this a class. Perhaps the next iteration on the project that uses the modal can expand on this. WDYT?

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -119,5 +130,481 @@ public static function my_account_menu_items( $items ) {
...
+		if ( ! \is_user_logged_in() || ! Reader_Activation::is_user_reader( $user ) ) {
+			return;
+		}
+
+		// Only if updating password from Account Settings page.
+		$action = filter_input( INPUT_POST, 'action', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
+		$nonce  = filter_input( INPUT_POST, 'woocommerce-reset-password-nonce', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
+		if ( self::RESET_PASSWORD_ACTION !== $action || empty( $nonce ) ) {
+			return;
+		}
```
miguelpeixe: Shouldn't we also verify the nonce before moving forward?

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -119,5 +130,481 @@ public static function my_account_menu_items( $items ) {
...
+		$_POST['reset_key']   = $key;
+		$_POST['reset_login'] = $login;
+	}
+
+	/**
+	 * Check if the user has a valid password reset key.
+	 *
+	 * @return bool True if the user has a valid password reset key, false otherwise.
+	 */
+	public static function check_password_reset_key() {
```
miguelpeixe: I think this and the method below could benefit from the readability of guard clauses, like it's being done in other methods for this class.

miguelpeixe (changes requested):
Everything tested well and I like how the new modal feature is looking! Left a few comments inline

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Good idea! Moved to a separate `My_Account_UI_V1_Passwords` class in 67d48c6e.

includes/class-newspack-ui.php
dkoo: Yeah, this can probably be expanded upon into a more robust class where you pass arguments via the class constructor. I definitely welcome any improvements!

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: It's already being checked by Woo, but we probably shouldn't assume. Updated in 459eb1c2

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: 24392523 refactors all of the methods in the `_V1` classes to use the same guard clause patterns. In the specific case of the `My_Account_UI_V1_Passwords::check_password_reset_key()` method, I'm not sure we need the guard clauses as it's a read-only method and is only called by `My_Account_UI_V1_Passwords::redirect_reset_password_link()`, which does check all conditions.

dkoo:
@miguelpeixe thanks for the review! Let me know if you had something else in mind for the class file structure.

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -119,5 +135,286 @@ public static function my_account_menu_items( $items ) {
...
+			<div>
+				<p class="font-size newspack-ui__font--s newspack-ui__font--bold"><?php esc_html_e( 'Subscriptions', 'newspack-plugin' ); ?></p>
+				<p class="newspack-ui__helper-text"><?php esc_html_e( 'Review and cancel active subscriptions.', 'newspack-plugin' ); ?></p>
+			</div>
+			<div class="newspack-ui__width--33">
+				<a class="newspack-ui__button newspack-ui__button--secondary newspack-ui__button--wide" href="<?php echo esc_url( \wc_get_endpoint_url( 'subscriptions', '', \wc_get_page_permalink( 'myaccount' ) ) ); ?>">
+					<?php esc_html_e( 'Manage subscriptions', 'newspack-plugin' ); ?>
+				</a>
+			</div>
+		</div>
```
miguelpeixe, line 208: After testing again with a fresh reader account I realized that this is also rendering when I have no subscriptions, which points to an empty page with a weird notification markup:

image

While at it, maybe also skip the "Manage newsletters" if the reader is not subscribed to any list?

miguelpeixe (changes requested):
Looks good! Thank you for the revisions. On my second test I found a small issue with the account deletion modal:

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Good catch! a5e46dd2 tweaks this modal template so that the "Manage Subscriptions" and "Manage Newsletters" content will only be shown if the user has any active subscriptions for those entities. If not, the modal will look a lot simpler:

Screenshot 2025-05-27 at 11 30 16 AM

miguelpeixe: Nice, thanks!

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
```diff
@@ -119,5 +135,305 @@ public static function my_account_menu_items( $items ) {
...
+		}
+
+		// If the user has clicked the button from the delete account email, show the confirmation modal.
+		$delete_account_form = filter_input( INPUT_GET, WooCommerce_My_Account::DELETE_ACCOUNT_FORM, FILTER_SANITIZE_FULL_SPECIAL_CHARS );
+		if ( ! empty( $delete_account_form ) ) {
+			return self::delete_account_confirmation_modal();
+		}
+
+		$active_subscriptions     = WooCommerce_Connection::get_active_subscriptions_for_user( $user->ID );
+		$newsletter_subscriptions = method_exists( 'Newspack_Newsletters_Subscription', 'get_contact_lists' ) ? \Newspack_Newsletters_Subscription::get_contact_lists( $user->data->user_email ) : [];
```
miguelpeixe: Noting that this will increase load time for the My Account dashboard page, as it performs a remote request to the ESP API.

It's suboptimal, but I wouldn't say it's something to be optimized in this PR.

miguelpeixe (approved):
Thank you for the revisions! Tests well

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Yes, that's true. We could try looking at the stored reader data instead, WDYT?

dkoo: Here's how that might look: 2e51a2be

miguelpeixe: Good idea! I guess having to JSON decode it is a price to pay 😄

miguelpeixe (approved):

includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
dkoo: Unless we want to augment the `get_data` method so that it can optionally return decoded data? No getting around the need to decode somewhere, though.

PR #3987, opened 2025-05-21
fix: ensure compatibility with woo memberships 1.27.2

leogermani:
### All Submissions:

Woocommerce Memberhisps lowered the priority of the hook it uses to restrict content, which made our own hooks to break. The result was that we were unable to apply metering to the posts and all posts were being restricted in the first visit.

Also, our JS was correctly removing the content gate before all the free articles have been read, so the end result was just a post with one or two paragraphs.

1. Install the latest Woo Memberships (1.27.2)
2. Create a content gate with metering for anonymous visitor
3. On `trunk` notice that the content is trimmed on the first visit
4. Checkout this branch and see that the content is not trimmed, only when you read all the allowed articles
5. See that the content gate renders as expected

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-metering.php
+++ b/includes/plugins/wc-memberships/class-metering.php
@@ -36,7 +36,7 @@ class Metering {
 	 */
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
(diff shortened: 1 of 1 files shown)


claudiulodro (approved):
Confirmed this solves it on both new and old versions of Memberships

PR #3991, opened 2025-05-23
feat(my-account): payment information custom page template

dkoo:
### All Submissions:

Implements a custom page template and UI for the Payment Methods page, renamed to "Payment Information". Eventually this page will also contain UI to add, edit, and delete billing and shipping addresses, but that will come in a separate PR.

Note: I'm seeing the success messages shown after adding a new payment method disappear automatically after a few seconds—I'm not sure what's causing this, but I'm open to suggestions on how to fix.

Closes NPPD-116, NPPD-117, NPPD-119, and NPPD-120.

1. Make sure you have at least one (but ideally all) of the Newspack-supported payment gateways active and connected in Test Mode via **WooCommerce > Settings > Payments**. We currently support: [WooPayments](https://woocommerce.com/payments/), [Stripe](https://wordpress.org/plugins/woocommerce-gateway-stripe/), and [PayPal](https://woocommerce.com/products/woocommerce-paypal-payments/).
3. Purchase a subscription or make a recurring donation as a new reader. 
2. Check out this branch and make sure `NEWSPACK_MY_ACCOUNT_VERSION` is not defined or set to `0.0.0` in wp-config.php.
4. Visit My Account, verify the email if needed, and visit **My Account > Payment Methods**. Confirm that there's no difference in styles or behavior here compared to `trunk`. Smoke test adding and deleting payment methods in all active payment gateways.
5. Define `NEWSPACK_MY_ACCOUNT_VERSION` as `1.0.0` in wp-config.php and refresh the Payment Methods page. Confirm that it now looks like this:

<table>
<tr>
<td>
mobile:
Screenshot 2025-05-28 at 4 15 06 PM
</td>
<td>
desktop:
Screenshot 2025-05-28 at 4 14 54 PM
</td>
</tr>
</table>

6. Add a new payment method and confirm that the form loads in a modal:

Screenshot 2025-05-28 at 5 01 58 PM

7. Confirm that multiple payment methods can be toggled like an accordion UI, but that the first one is always expanded.
8. Test adding new payment methods for all active payment gateways.
9. Test setting one of the non-default payment methods as default using the dropdown menu, and confirm that any new checkouts default to the this payment method:

Screenshot 2025-05-28 at 5 11 08 PM

10. Test deleting payment methods. Delete all payment methods (you may need to cancel active subscriptions in order to do this) and confirm that the page now looks like this:

Screenshot 2025-05-28 at 4 29 12 PM

11. Note that if you switch back to v0.0.0, you may need to flush permalinks or the Add Payment Method screen may return a 404.

### Other information:


dkoo:
I noticed that the fix from #2934 isn't working here. The checkbox is always visible, and it's not checked by default despite the `wc_stripe_save_to_subs_checked` filter returning `true`. This only seems to be true in the V1 UI—not sure why, but since the checkbox is still visible I'm considering it non-blocking. Would love to figure out how to at least force it checked on by default as with v0, though.

Screenshot 2025-05-28 at 5 07 05 PM

rbcorrales (approved):
I tested this with Stripe and Woo, and it works as expected. Nice overhaul!

The only issue that may require further review is a double confirmation message I encountered while updating the default payment method. However, I only saw this happen once, wondering if this might be a temporary issue related to the Stripe API message. In any case, this did not seem to affect the functionality itself.

image

dkoo:
> The only issue that may require further review is a double confirmation message I encountered while updating the default payment method

I'm not able to replicate the message that asks if you'd like to update your subscriptions to use the new payment method. Are there specific steps you took to get that message? After adding a payment method, I just see this (true for both WooPay and Stripe):

Screenshot 2025-05-30 at 5 05 55 PM

rbcorrales:
I'm also unable to replicate the issue. I've tried adding and changing the default payment methods multiple times. I suspect it may be related to a temporary problem with the Stripe API as per the screenshot message.

Is there a possibility that there is an automatic retry mechanism that could have sent the request a second time, and if the first call failed, could the callback handler have been called twice?

dkoo:
Oh, I see it now! You need to click "make default" on a payment method that was added via the Stripe gateway plugin. But I'm not seeing the double message; I suspect this was an odd result of the API error in your screenshot.

Screenshot 2025-05-30 at 5 15 15 PM

dkoo:
I see what's happening now! If you click the "Make default" action multiple times in a row, it might perform the same action on the server multiple times, each time adding its own message.

Screenshot 2025-05-30 at 5 22 58 PM

We should probably block the UI after clicking these options—I'll see if that's a quick change, and if so will update this PR

dkoo:
@rbcorrales eb4ee4f3 adds some JS and CSS to block UI on certain interactions. Try the following:

- Click "Make default" or "Delete payment method" via the dropdown menus in Payment Information
- Click "Update profile", "Update password", "Forgot password" on Account Information

In all cases, the clicked element's parent container should be greyed out and non-interactive after clicking.

rbcorrales (approved):

PR #3998, opened 2025-05-27
fix(corrections): rewrite rule flush check

miguelpeixe:
### All Submissions:

The rewrite rules are flushed on every page load. This fixes the option check to only run if the option doesn't exist.

1. While on release, confirm the option is updated and the rewrite rules are flushed every page load
2. Checkout this branch and run `wp option delete newspack_corrections_rewrite_rules_updated`
3. Reload the page and confirm the option gets updated and rewrite rules are flushed
4. Reload again and confirm it no longer updates

### Other information:

### DIFF
--- a/includes/corrections/class-corrections.php
+++ b/includes/corrections/class-corrections.php
@@ -171,7 +171,7 @@ class Corrections {
 		\register_post_type( self::POST_TYPE, $args );
 
 		$rewrite_rules_updated_option_name = 'newspack_corrections_rewrite_rules_updated';
-		if ( get_option( $rewrite_rules_updated_option_name ) !== true ) {
+		if ( get_option( $rewrite_rules_updated_option_name ) !== 'true' ) {
 			flush_rewrite_rules(); //phpcs:ignore
 			update_option( $rewrite_rules_updated_option_name, true );
 		}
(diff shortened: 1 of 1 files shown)


includes/corrections/class-corrections.php
```diff
@@ -171,7 +171,7 @@ public static function register_post_type() {
 		\register_post_type( self::POST_TYPE, $args );
 
 		$rewrite_rules_updated_option_name = 'newspack_corrections_rewrite_rules_updated';
-		if ( get_option( $rewrite_rules_updated_option_name ) !== true ) {
+		if ( get_option( $rewrite_rules_updated_option_name ) !== '1' ) {
```
leogermani: WDYT checking for `false === ...` ?

rbcorrales (approved):
Looks good to me. Another failsafe option could be:
```php
if ( ! (bool) get_option( $rewrite_rules_updated_option_name ) ) {
```

includes/corrections/class-corrections.php
miguelpeixe: Alright! Done in b39599577306463e89d536a09bdcde8a7b610699

miguelpeixe:
I think the `false` check, as suggested by Leo, should be fine. We only care if the option doesn't yet exist.

PR #4000, opened 2025-05-28
feat(collections): register collection CPT and taxonomy

rbcorrales:
### All Submissions:

- Register a Collections custom post type and a collection taxonomy for managing collections.
- Adds a sync mechanism between collection posts and terms to maintain bidirectional linking relying on post and term meta.
- Add a hook management trait for consistent hook registration/unregistration to avoid triggering circular updates.
- Reorganizes test files into a more structured directory layout.

Closes [NPPD-138](https://linear.app/a8c/issue/NPPD-138/register-collection-cpt-and-taxonomy).

1. Enable the `NEWSPACK_COLLECTIONS_ENABLED` feature flag in `wp-config.php` and the Collections module in the Newspack settings. A new sidebar menu should appear:
image
2. For easier testing, you can also enable the Collection taxonomy in the menu [changing this line](https://github.com/Automattic/newspack-plugin/blob/1f320b6bef7fd1c651ff57f400d03ee07a5fb05f/includes/collections/class-collection-taxonomy.php#L87):
```php
'show_in_menu' => true,
```
3. Create a new Collection post:
   - Verify that a corresponding Collection term is automatically created.
   - Verify that editing the post title or slug updates the term accordingly.
   - Verify that trashing the post deletes the term.
   - Verify that deleting the post deletes the term.
4. Create a new Collection term (from the UI if the admin menu is enabled following the second step or from the block editor sidebar on a normal post):
   - Verify that a corresponding Collection post is automatically created with a draft status.
   - Verify that editing the term name or slug updates the post.
   - Verify that deleting the term trashes the post.
5. Edit or create a regular post:
   - There should be a new Collections sidebar panel in the block editor:
image
   - Verify that you can add existing Collections.
   - Verify that you can create a new Collection term from the sidebar and that it also creates a new Collection post.
6. Verify that there's a new Collections column in the post list in the admin, showing the assigned collections assigned to a post and that clicking them filters posts by that collection.

### Other information:


includes/collections/class-sync.php
```diff
@@ -0,0 +1,371 @@
...
+		}
+	}
+
+	/**
+	 * Delete the linked term when a collection post is deleted or trashed.
+	 *
+	 * @param int $post_id Post ID.
+	 * @return bool|int|WP_Error|void True on success, false if term does not exist. Zero on attempted deletion of default Category. WP_Error if the taxonomy does not exist. Void if no changes were made.
+	 */
+	public static function handle_post_deleted( $post_id ) {
```
miguelpeixe: Trashing a post shouldn't be a destructive action. Restoring a collection should also restore its relationships.

I think we should either implement a trashed state for the collection term via a term meta or store the term's relationship data as post meta so it can be re-established in the recreated term

includes/collections/class-sync.php
```diff
@@ -0,0 +1,371 @@
...
+		return $result;
+	}
+
+	/**
+	 * Handle untrashing of a collection.
+	 *
+	 * @param int $post_id Post ID.
+	 * @return WP_Term|array|WP_Error|void Term object if exists, array if created, WP_Error on failure, void if no changes were made.
+	 */
+	public static function handle_post_untrashed( $post_id ) {
```
miguelpeixe: When untrashing a collection I'm unable to restore its original slug and end up with a draft `{slug}__trashed`. The expected behavior from core is to have the `__trashed` suffix removed when untrashing.

In one case, an odd interaction caused the creation of duplicated terms:

image

The post's slug got restored to `my-first-collection`. I wasn't able to reproduce this duplication and am not sure how I got the restored post to also get back to the original slug.

I consistently get the restored trashed post to remain with the trash slug, as well as the re-created term.

includes/collections/class-sync.php
```diff
@@ -0,0 +1,371 @@
...
+		 *
+		 * @param int $post_id Post ID.
+		 * @param int $term_id Term ID.
+		 */
+		do_action( 'newspack_collections_before_removing_linked_post', $post_id, $term );
+
+		// Trash the post instead of deleting it permanently.
+		Post_Type::unregister_hooks();
+		$result = wp_trash_post( $post_id );
+		Post_Type::register_hooks();
```
miguelpeixe, line 382: Why keep the post in the trash here? If that was an accidental deletion and the collection gets recreated via taxonomy term we'll end up with 2 posts in the database.

miguelpeixe (changes requested):
I like this architecture and it tested mostly well! I encountered a few issues that I commented inline.

includes/collections/class-sync.php
miguelpeixe: In general, I think the two-way sync strategy should be 1:1. If we allow an entity to be in a different state than the other, we open a door for unexpected things to happen.

includes/collections/class-sync.php
rbcorrales: Since terms don't implement status OOTB, I wanted to avoid leaving orphaned relationships as people sometimes simply trash posts instead of permanently deleting them thinking they're gone. Since the only elements copied from a post to a taxonomy are the name and slug, and the taxonomy is hidden from the admin so no extra configs can be made to them, I thought it would be simpler to remove and recreate them as necessary. However, I'm open to exploring alternatives if you feel strongly about this approach.

includes/collections/class-sync.php
rbcorrales: I just tried creating and deleting collections multiple times and I cannot replicate it 🤔 

Every time I untrash a post, its slug and the corresponding term slug are restored without any `_trashed suffix`. I even tried creating a collection with the same title and slug of a post in the trash, then restoring it, and it restored the relationship to the other existing term, but I never saw the suffix.

If you can consistently replicate it, can you guide me to the steps you're taking?

includes/collections/class-sync.php
rbcorrales: > or store the term's relationship data as post meta so it can be re-established in the recreated term

Thinking a bit more about this, what data do you think would need to be stored? Given that the only relevant data for the term should be the name and slug, which in turn should exactly match the post being restored, and it should be a 1:1 relationship.

includes/collections/class-sync.php
rbcorrales: > Why keep the post in the trash here? 

Technically, you shouldn't be deleting terms directly, as this is hidden from the UI. These functions are more like a safeguard for edge cases where you delete things via code or maybe a bulk deletion process.

> If that was an accidental deletion and the collection gets recreated via taxonomy term we'll end up with 2 posts in the database.

It may be unintentional, but from a data perspective, I believe it's better having to manually delete a duplicate post than to lose any configured collection metadata permanently.

> In general, I think the two-way sync strategy should be 1:1. If we allow an entity to be in a different state than the other, we open a door for unexpected things to happen.

Agree with that, and I've been striving for it as much as possible with the creation and edits. Unfortunately, there are no taxonomy statuses to match them with posts 1:1 (as discussed in the other comment).

includes/collections/class-sync.php
miguelpeixe: > what data do you think would need to be stored? 

The term relationship with posts, so just post IDs. Restoring a collection from trash will recreate a term but without any posts assigned to it, so we permanently lose data when trashing, which shouldn't happen.

dkoo:
While testing #4001 I noticed there's an error notice being shown from the Ads plugin when editing a Collection post type. I confirmed that it happens on this branch, too.

I'm not seeing this error on any other CPT on my site, so I'm wondering what's different about this one?

Screenshot 2025-05-29 at 4 51 14 PM

includes/collections/class-sync.php
miguelpeixe: Not sure what happened here... I tested again and can't reproduce 😕 

Let's leave this for now, I'll report if it happens again.

includes/collections/class-sync.php
miguelpeixe: Sounds good, thank you for explaining!

includes/collections/class-sync.php
rbcorrales: Oh, you're totally right, I missed that detail about already tagged posts. I'll work on that, thanks for clarifying!

miguelpeixe:
@dkoo I saw that as well. That's because the Collections post type doesn't have support for `custom-fields` and there seems to be a bug in Newspack Ads [not handling the `undefined` response](https://github.com/Automattic/newspack-ads/blob/5fb1559030adb230522370711cd3853817b094f9/src/suppress-ads/index.js#L72-L73) from `select( 'core/editor' ).getEditedPostAttribute( 'meta' )` for such post types.

I'll push a PR for the Ads issue.

includes/collections/class-sync.php
rbcorrales: @miguelpeixe I submitted a fix for this. Now, when a post is trashed, it should add a `_newspack_collection_inactive` term meta instead of deleting it. Once untrashed, then the meta is deleted. I also modified the term queries to also check for that meta.

Could you please take a look at this and let me know if you see any issues with this approach? Thanks!

rbcorrales:
> That's because the Collections post type doesn't have support for `custom-fields`

I ended up adding support for this but in a newer PR, but thanks for flagging and fixing this.

rbcorrales:
BTW, I also now prefixed the object names here for consistency with other Newspack features:
- **Collection CPT:** `collection` -> `newspack_collection`
- **Collection Taxonomy:** `collection_tax` -> `newspack_collection_taxonomy`

So if you have some locally created objects, those won't appear in the UI unless you manually delete them.
Here are some handy queries if you want to update them in the db (needs cache flushing):

```sql
UPDATE wp_posts SET post_type = 'newspack_collection' WHERE post_type = 'collection';
UPDATE wp_term_taxonomy SET taxonomy = 'newspack_collection_taxonomy' WHERE taxonomy = 'collection_tax';
```

includes/collections/class-collection-taxonomy.php
```diff
@@ -0,0 +1,170 @@
...
+ */
+class Collection_Taxonomy {
+	use Hook_Management_Trait;
+
+	/**
+	 * Taxonomy for Collections.
+	 *
+	 * @var string
+	 */
+	private const TAXONOMY = 'newspack_collection_taxonomy';
```
miguelpeixe, line 25: Now that we're making the change, I have a non-blocking question/suggestion. Why not just `newspack_collection`, like the cpt?

includes/collections/class-sync.php
```diff
@@ -0,0 +1,416 @@
...
+			/**
+			 * Fires before marking a term as inactive.
+			 *
+			 * @param int $post_id Post ID.
+			 * @param int $term_id Term ID.
+			 */
+			do_action( 'newspack_collections_before_marking_term_inactive', $post_id, $linked_term_id );
+
+			Collection_Taxonomy::unregister_hooks();
+			$result = update_term_meta( $linked_term_id, Collection_Taxonomy::INACTIVE_TERM_META_KEY, '1' );
```
miguelpeixe: Non-blocking: like `is_term_inactive()`, this could be a method in the `Collection_Taxonomy` class

miguelpeixe (approved):
Thank you for addressing the trash status! Works beautifully.

Left a couple of non-blocking comments below.

includes/collections/class-collection-taxonomy.php
rbcorrales: Good question! Even though this is an internal taxonomy, it still needs to expose its REST endpoint so it can be used in contexts like tagging posts in the admin. By default, WordPress uses the object name as the REST base (e.g., `/wp-json/wp/v2/newspack_collection`).

If both the CPT and taxonomy share the same name, it causes conflicts when interacting with the taxonomy sidebar on the post editor (in particular, bad request errors on the frontend) because I guess WordPress gets confused with their definitions mixed up.

One way around this might be to explicitly set a different `rest_base` for the taxonomy when registering it, but then you have to remember there's a disconnect between its definition and the way to interact with it. Having distinct names for the CPT and taxonomy avoids ambiguity and makes things clearer overall when looking at the data. Although it might be a bit verbose, it shouldn't be exposed to the final readers, so naming it explicitly feels like an easier-to-maintain choice.

includes/collections/class-sync.php
rbcorrales: Thanks for the suggestion, I agree. Added two methods to encapsulate the inactive behavior in the `Collection_Taxonomy` class: `deactivate_term` and `reactivate_term`.

While on that, I discovered an issue: When the post was deleted from the trash, the taxonomy wasn't being deleted from the database because the `term_exists` check was ignoring inactive items, so it thought there was nothing to delete. I solved it by also creating a `term_exists` method in the class that wraps the original method with a call to unregister the inactive filter to check if the term exists regardless of its status.

PR #4004, opened 2025-05-28
fix: sync membership status to ESP when reader deletes account

laurelfulford:
### All Submissions:

This PR makes sure a reader's NP_Membership Status is updated to the ESP when a user deletes their account. 

I went down a bit of a rabbit hole with this one, thinking the missing information sent by `reader_delete_sync()` was causing it not to sync to the ESP. It turns out it was actually because the user no longer existed by the time `run_queued_syncs()` was run, so `self::get_contact_data( $user->ID )` wasn't returning anything for them. 

I ended up trying to grab more user info before delete with `store_user_data_before_deletion()` to fix what I originally thought the issue was. I've left that in place in case it's better to send as much info to the ESP as possible, but the NP_Membership Status update seems to work without it by just sending the stuff under `// If we don't have anything stored, send what we can.`.

See NPPM-387.

1. Start with a test set connected to either ActiveCampaign or MailChimp, and make sure it's set to sync reader data to the ESP under Audience > Configuration. 
2. Set up a membership product for readers to purchase, and add a Checkout button block to a page for that product. 
3. In an incognito window, run through a checkout with the membership product. 
4. When complete, confirm your reader's information was synced to the ESP:

CleanShot 2025-05-28 at 11 03 00

5. Back in your incognito window, go to My Account > Delete Account. Follow the steps (getting the deletion email, following the link, and clicking the "Delete Account" button). 
6. Check your ESP; note the account still has a NP_Membership Status of `active`, even though the account was deleted. 
7. Apply this PR.
8. Repeat steps 3-6; on step 6, confirm that your reader's NP_Membership Status was updated to `user-deleted`:

CleanShot 2025-05-28 at 11 11 06

### Other information:

### DIFF
--- a/includes/data-events/connectors/class-esp-connector.php
+++ b/includes/data-events/connectors/class-esp-connector.php
@@ -23,6 +23,8 @@ class ESP_Connector extends Reader_Activation\ESP_Sync {
 	 */
 	public static function init_hooks() {
 		add_action( 'init', [ __CLASS__, 'register_handlers' ] );
+		// Store user data before deletion.
+		add_action( 'delete_user', [ __CLASS__, 'store_user_data_before_deletion' ], 5 );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


dkoo (approved):
@laurelfulford this is working as described!

However, while testing this I realized we weren't doing what the UI is saying we do—when a user account is deleted, we should be deleting the account, or at least unsubscribing them from lists. These PRs should fix this issue:

- https://github.com/Automattic/newspack-plugin/pull/4008
- https://github.com/Automattic/newspack-newsletters/pull/1837

laurelfulford:
Thanks @dkoo!

PR #4016, opened 2025-06-04
feat(my-account): change payment method modal

miguelpeixe:
### All Submissions:

Implement the modal checkout for subscriptions' "Change payment" action.

Because the "change payment method" page is not a regular checkout and has a custom URL, the `registerModalCheckoutButton()` will now open the modal using the response URL from the button's href fetch.

1. Checkout this branch and https://github.com/Automattic/newspack-blocks/pull/2141
2. Make sure you have `define( 'NEWSPACK_MY_ACCOUNT_VERSION', '1.0.0' );` in your `wp-config.php`
3. Connected as a reader, purchase a subscription product
4. Navigate to My Account -> Subscriptions, and "View" the purchased subscription
5. Click "Change payment" and confirm that the modal renders:

image

7. Select a new payment method with a new test card (e.g. `5555555555554444` or `378282246310005`)
8. Uncheck the "Use this payment method for all of my current subscriptions" option and submit
9. Confirm you get the following success message:

image

10. Click "Continue browsing" or close and confirm that the page refreshes and the "Payment" line is updated
11. Click Change payment again, select the previous payment method, and submit (leaving the "Use this payment method for all of my current subscriptions" option checked)
12. Confirm you get the following success message:

image

13. Click "Continue browsing" or close and confirm that the page refreshes and the "Payment" line is updated

### Other information:


rbcorrales:
@miguelpeixe the modal functions correctly. However, after changing the payment method, I see a confirmation appears in a larger modal, which then switches to a smaller confirmation modal a few seconds later without any further interactions. Is this expected?
image
image

miguelpeixe:
Thanks, @rbcorrales! This is certainly not ideal.

I can reproduce it locally, but not consistently. This happens because of some latency to reach here:

https://github.com/Automattic/newspack-blocks/blob/0e1a82f6beea2fcb68becf0eef94e17b97bbede0/src/modal-checkout/modal.js#L118

We have a custom `iframeReady` utility to resolve the iframe UI quicker than we would with `iframe.addEventListener( 'load' )`, but we're still relying on `load` here:

https://github.com/Automattic/newspack-blocks/blob/0e1a82f6beea2fcb68becf0eef94e17b97bbede0/src/modal-checkout/modal.js#L145

I'm proposing some changes here: https://github.com/Automattic/newspack-blocks/pull/2146

rbcorrales (approved):
Yeah, it's appreciable if there are latency issues (my tunneled local is a bit slow).

I tested it again with the changes from the other PR, and I still can see the intermediate state, but it has a blurred overlay with the spinner on top, which should be fine as it kind of reflects that the process hasn't been completely finished yet and the modal isn't still ready for interaction. Before, it just looked like the layout suddenly shifted and the modal elements moved around.

image

PR #4022, opened 2025-06-06
feat: add option to create and display an Accessibility Statement page

laurelfulford:
### All Submissions:

This PR adds an option to create and Accessibility Statement to the Newspack Plugin, and creates a page with some boillerplate content.

With https://github.com/Automattic/newspack-theme/pull/2506, a link to that page is added to the footer of the Newspack Theme.

As part of this change, the design called for the Display Settings tab under Newspack > Settings has been renamed to Advanced Settings, and the Advanced Setting button at the bottom has been renamed to Open Customizer. This is because the page this option has been added to is no longer strictly design.

See NPPM-105.

_**Note:** I updated these test steps after addressing @thomasguillot's feedback below!_ 

1. Apply this PR and https://github.com/Automattic/newspack-theme/pull/2506, and run `npm run build` for the Newspack Plugin. 
2. Navigate to Newspack > Settings; confirm the last tab now reads Advanced Settings, and clicking it opens the Advanced Settings page:

CleanShot 2025-06-06 at 14 57 45

3. Scroll to the bottom of the page and confirm there's a new "Accessibility Statement Page" section, with a "Edit and Publish Page" button and two links to the W3 site. It should also display a message, "Your accessibility statement page is not yet published. Please review and make edits before publishing."

CleanShot 2025-06-08 at 12 30 35

4. Test the W3 links to confirm that they work, and open the links in new tabs.
6. Click the "Edit Page" link. It should bring you to a page called, "Accessibility Statement", with some boilerplate text and a bunch of elements highlighted that publishers should change, including the pub's information, a timeline for response, and today's date:

CleanShot 2025-06-08 at 12 31 57

7. Make some edits and publish the page. 
8. Return to Newspack > Settings > Advanced Settings, and confirm that the Accessibility Statement Page section has been updated, so the message says, "Your accessibility statement page is published.":

CleanShot 2025-06-06 at 15 03 41

9. View the front-end of the site; confirm that there is an 'Accessibility Statement' link in the footer. Click the link and confirm it goes to your Accessibility Statement page. 
10. Edit your Accessibility Statement page to change the title and page slug; publish. 
11. View the front-end again confirm that the Accessibility Statement link has been updated to use the new page title and slug. 
12. Navigate back to Newspack > Settings > Advanced Settings. Confirm that the "Edit Page" link still works. 
13. Navigate to WP Admin > Pages, and delete your Accessibility Statement page. It should have a post status denoting what page it is in the list, similar to the Front Page:

image

14. Navigate back to Newspack > Settings > Advanced Settings, and confirm the Accessibility Statement Page shows a new message asking you to create a page. This is a fallback in case a publisher deletes the draft page:

CleanShot 2025-06-08 at 12 41 49

15. Click the "Create Page" button; the page should reload and go back to the "Edit and Publish Page" messaging.
16. Confirm you can edit the new page successfully, and publish it. 
17. Delete the `theme_mod`, like with ` wp theme mod remove accessibility_statement_page_id`.
18. Go to Newspack > Settings > Advanced Settings.
19. Confirm you're back to the "Edit and Publish Page" status, and that a new Accessibility Statement draft page has been created. (This is just to make sure everything's working as expected under the hood: if the theme_mod doesn't exist, the draft page should be created and its ID should be saved. If it does exist but the associated page has been deleted, it should let you create a new page).

### Other information:


laurelfulford:
@thomasguillot I've added a link to a test site in the Linear task to hopefully make this easier to test the visuals/UX!

thomasguillot:
There's some weird spacing around the external links. e.g.:

Screenshot 2025-06-07 at 09 35 23@2x

thomasguillot:
I think rather than "Create Page", we should just create it automatically, but as a draft. And then the label would be "Edit and Publish Page". And then once published, it'd be just "Edit Page".

And then for the notice, we would just have 2:
* "Your accessibility statement page is not yet published. Please review and make edits before publishing." or,
* "Your accessibility statement page is published."

What do you think?

thomasguillot:
Screenshot 2025-06-07 at 09 40 23@2x

Do you think it's possible to add a "page type" (not sure how it's called) to the list of pages?

thomasguillot:
When you're in the editor, and you click the "Back" site icon button, you're redirected to the Pages.

`wp-admin/edit.php?post_type=page`

This isn't a dealbreaker but would it be possible to instead redirect the user back to where they were before? (So Newspack > Settings > Advanced Settings if they came form our plugin)

thomasguillot:
Screenshot 2025-06-07 at 09 45 06@2x

I'm not too keen on this yellow text at the top of the page. I think it's ok for now but I wonder if we can leverage the [Guide](https://wordpress.github.io/gutenberg/?path=/docs/components-guide--docs) component as an enhancement task?

thomasguillot:
This is great, thank you @laurelfulford 
It’s mostly a ✅ from me.

laurelfulford:
Thanks @thomasguillot! This is super helpful!

> There's some weird spacing around the external links.

Ah, I literally had weird spaces inside of the tags! Fixed 😆 

CleanShot 2025-06-08 at 11 39 39

> I think rather than "Create Page", we should just create it automatically, but as a draft. And then the label would be "Edit and Publish Page". And then once published, it'd be just "Edit Page".

My concern here is how we're handling the page afterwards. This is the only way to assign this page; if the page is deleted (as a draft or after publishing), we probably don't want to keep creating it (since that may have been for a reason, and the publisher doesn't want to use it). But then there wouldn't be a way for a publisher to recreate the page if this was an accident 😕

I do think skipping the initial 'Create Page' step would be nicer though, so I went with this mashup:

1. When you visit the Advanced Settings page, it checks to see if a theme_mod is saved for the accessibility statement page ID. If not, a page is created and you get the "Edit and Publish Page" messaging.
2. After the page is published, you get the "Edit Page" messaging.
3. If you visit the Advanced Settings page and it have a page ID saved but it matches a page that's been trashed or doesn't exist, you get the "Create Page" messaging.

Most people won't see that last state, but it makes sure the page can be recreated if needed!

> Do you think it's possible to add a "page type" (not sure how it's called) to the list of pages?

Turns out “yes”, using the [`display_post_states hook()`](https://developer.wordpress.org/reference/hooks/display_post_states/) — TIL! 

CleanShot 2025-06-08 at 12 07 11

> When you're in the editor, and you click the "Back" site icon button, you're redirected to the Pages.

This looks like it happens for the Support Our Publication page as well, and when you’re creating a new Newsletter (going from Add Newsletter > Pick a Template, then when you're in the Newsletter editor, clicking the Site Icon brings you to the Newsletter CPT list page). I’m a little leery about trying to update that and messing something bigger up, and I don’t really get whether this is “right” (that there’s a default ‘back’ state sometimes). 

I did check, and the regular browser back button takes you to the right spot. Hopefully that’s enough for now, but we can iterate if we get more feedback about this!

> I'm not too keen on this yellow text at the top of the page. I think it's ok for now but I wonder if we can leverage the ~[Guide](https://wordpress.github.io/gutenberg/?path=/docs/components-guide--docs)~ component as an enhancement task? 

Yeah it’s pretty fugly! I mostly wanted to make sure people scrolled down and read it/made updates. 

I think a Guide could be an ultimate fix, but for now I replace the yellow with a grey box, and left the highlighted sections to what needs to be changed.

CleanShot 2025-06-08 at 12 08 50

I'm going to update the testing steps to reflect these changes -- just let me know if you have any questions/concerns!

includes/class-advanced-settings.php
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+/**
+ * Advanced Settings class.
+ */
+class Advanced_Settings {
```
miguelpeixe: WDYT we call it `Accessibility_Statement_Page`? We don't need this class to house more advanced settings. Maybe under `includes/advanced-settings/class-accessibility-statement-page.php` if you prefer.

includes/class-advanced-settings.php
```diff
@@ -0,0 +1,188 @@
...
...hnologies working with web browsers and assistive technologies installed on your computer:</p>\n<!-- /wp:paragraph -->\n\n<!-- wp:list -->\n<ul class=\"wp-block-list\"><!-- wp:list-item -->\n<li>HTML</li>\n<!-- /wp:list-item -->\n\n<!-- wp:list-item -->\n<li>CSS</li>\n<!-- /wp:list-item -->\n\n<!-- wp:list-item -->\n<li>JavaScript</li>\n<!-- /wp:list-item --></ul>\n<!-- /wp:list -->\n\n<!-- wp:heading -->\n<h2 class=\"wp-block-heading\">Legal Framework</h2>\n<!-- /wp:heading -->\n\n<!-- wp:paragraph -->\n<p>This accessibility statement is provided in accordance with:</p>\n<!-- /wp:paragraph -->\n\n<!-- wp:list -->\n<ul class=\"wp-block-list\"><!-- wp:list-item -->\n<li>European Accessibility Act (Directive (EU) 2019/882)</li>\n<!-- /wp:list-item -->\n\n<!-- wp:list-item -->\n<li>Web Content Accessibility Guidelines (WCAG) 2.1</li>\n<!-- /wp:list-item --></ul>\n<!-- /wp:list -->\n\n<!-- wp:separator -->\n<hr class=\"wp-block-separator has-alpha-channel-opacity\"/>\n<!-- /wp:separator -->\n\n<!-- wp:paragraph -->\n<p><em>This statement demonstrates our ongoing commitment to accessibility and our compliance with the EU Accessibility Act requirements.</em></p>\n<!-- /wp:paragraph -->",
```
miguelpeixe: This can be hard to maintain. Can we include this template from a separate, well-formatted file?

includes/class-advanced-settings.php
```diff
@@ -0,0 +1,188 @@
...
+		// Register REST routes.
+		add_action( 'rest_api_init', [ __CLASS__, 'register_rest_routes' ] );
+		// Add post status to accessibility statement page.
+		add_filter( 'display_post_states', [ __CLASS__, 'accessibility_post_status' ], 10, 2 );
+	}
+
+	/**
+	 * Create an accessibility statement page.
+	 *
+	 * @return array|WP_Error The created page data or error.
```
miguelpeixe:
```suggestion
	 * @return array|null|WP_Error The created page data, null if it has been trashed or deleted, or error.
```

includes/class-advanced-settings.php
```diff
@@ -0,0 +1,188 @@
...
+	 * @param WP_Post $post The post object.
+	 * @return array The post states.
+	 */
+	public static function accessibility_post_status( $post_states, $post ) {
+		$page_id = get_theme_mod( 'accessibility_statement_page_id' );
+		if ( $page_id === $post->ID ) {
+			$post_states['accessibility_statement'] = __( 'Accessibility Statement', 'newspack-plugin' );
+		}
+		return $post_states;
+	}
```
miguelpeixe: Nice touch!

src/wizards/newspack/views/settings/advanced-settings/accessibility-statement.tsx
```diff
@@ -0,0 +1,181 @@
...
+}
+
+type PageData = {
+	editUrl: string;
+	status: string;
+	pageUrl: string;
+};
+
+export default function AccessibilityStatement( { isFetching }: AccessibilityStatementProps ) {
+	const { wizardApiFetch } = useWizardApiFetch( 'newspack-settings/display-settings/accessibility-statement' );
```
miguelpeixe:
```suggestion
	const { wizardApiFetch } = useWizardApiFetch( 'newspack-settings/advanced-settings/accessibility-statement' );
```

src/wizards/newspack/views/settings/advanced-settings/accessibility-statement.tsx
```diff
@@ -0,0 +1,181 @@
...
+	const [ localIsFetching, setLocalIsFetching ] = useState( false );
+	const [ localPageData, setLocalPageData ] = useState<PageData | null>( null );
+
+	// Function to fetch fresh data
+	const fetchFreshData = () => {
+		setLocalIsFetching( true );
+		wizardApiFetch(
+			{
+				path: `/newspack/v1/wizard/newspack-settings/accessibility-statement?_t=${Date.now()}`,
```
miguelpeixe: Is the `_t` argument needed here?

miguelpeixe (changes requested):
Everything tested great! Left a few requests inline.

includes/class-advanced-settings.php
laurelfulford: Oh yeah, that makes sense -- done! I went with adding the /advanced-settings/ directory so it could also house the boilerplate text file 🙂

includes/class-advanced-settings.php
laurelfulford: Yes, it's horrible!! I went with a separate PHP file that just returns the block markup so it needs minimal formatting (just escaping the single quotes in the content). This definitely makes it easier to read and hopefully easier to edit, since you can just copy-paste content from Gutenberg with minimal tweaks.

includes/class-advanced-settings.php
laurelfulford: Thanks Miguel -- done!

includes/class-advanced-settings.php
laurelfulford: Credit for this all goes to @thomasguillot! 😄

src/wizards/newspack/views/settings/advanced-settings/accessibility-statement.tsx
laurelfulford: Eep, thanks for catching this -- updated!

src/wizards/newspack/views/settings/advanced-settings/accessibility-statement.tsx
laurelfulford: Doesn't look like it! I added that at one point trying to get the Create > Edit and Publish Page button updating right, but that went through a few iterations, I think this got left at some point when it wasn't needed. 

I removed it and the button/message still seem to update correctly!

laurelfulford:
Thanks @miguelpeixe, this was super helpful! 🙌 I think I've addressed everything, but just let me know if anything I did looks weird - thanks!

includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		// If no page ID is stored, create a new page.
+		$page_id = wp_insert_post(
+			[
+				'post_title'   => __( 'Accessibility Statement', 'newspack-plugin' ),
+				'post_name'    => 'accessibility-statement',
+				'post_status'  => 'draft',
+				'post_type'    => 'page',
+				'post_content' => require NEWSPACK_ABSPATH . 'includes/advanced-settings/class-accessibility-statement-boilerplate.php',
```
miguelpeixe: I think this would look better if we use output buffer. Instead of returning a long string, let's just have the file print its contents normally ([example](https://github.com/Automattic/newspack-plugin/blob/c5953294b16df0ed813796b52bfceb238c3442f8/includes/plugins/wc-memberships/block-patterns/pay-wall-one-tier.php)), and get it like this:

https://github.com/Automattic/newspack-plugin/blob/c5953294b16df0ed813796b52bfceb238c3442f8/includes/plugins/wc-memberships/class-block-patterns.php#L83-L85

miguelpeixe (changes requested):
Thank you for the revisions, @laurelfulford!

Just one minor feedback inline.

includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -0,0 +1,188 @@
...
+		// Add post status to accessibility statement page.
+		add_filter( 'display_post_states', [ __CLASS__, 'accessibility_post_status' ], 10, 2 );
+	}
+
+	/**
+	 * Create an accessibility statement page.
+	 *
+	 * @return array|null|WP_Error The created page data, null if it has been trashed or deleted, or error.
+	 */
+	public static function create_accessibility_statement_page() {
```
miguelpeixe: This is non-blocking, but now that this whole class is about accessibility statement, the methods can be less verbose. `create_page()` would be fine.

includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -0,0 +1,188 @@
...
+
+	/**
+	 * Get accessibility statement page data.
+	 * The function to actually output the link lives in the Classic Theme.
+	 *
+	 * TODO: Create a block for the Block Theme that outputs the format we need there.
+	 *
+	 * @return array|false The page data or false if not set.
+	 */
+	public static function get_accessibility_statement_page() {
```
miguelpeixe: Same here with the method name, could be just `get_page()`

includes/advanced-settings/class-accessibility-statement-page.php
laurelfulford: Ah, thanks @miguelpeixe! This has been updated -- it's much nicer 🙂

includes/advanced-settings/class-accessibility-statement-page.php
laurelfulford: Good call! I went through and updated all of these so they're less redundant -- thank you!

includes/advanced-settings/class-accessibility-statement-page.php
laurelfulford: Done, plus the API ones and page creation one! 😄

miguelpeixe (approved):
Thanks, Laurel! Looks great and continues to work without issues 😄

laurelfulford:
Awesome! 🎉 Thanks for all the testing, @miguelpeixe!

thomasguillot:
Thank you @laurelfulford ❤️

PR #4028, opened 2025-06-10
feat: add manual option to mark subscriptions as giftable

laurelfulford:
### All Submissions:

This PR adds a 'Allow to be gifted' checkbox to products when:

* They are a subscription or variable subscription product
* When WooCommerce Subscriptions Gifting is installed and enabled, and
* When 'Limit Subscription' is set to 'Do not limit'. 

This checkbox is disabled by default, so all subscription products will not be considered 'giftable' by default.

This allows us to set up some more nuanced subscription gifting that works with our own subscription limitations (one per customer), and that won't clash with existing products. 

See NPPM-385.

2. Make sure you have Woo Subscriptions and Woo Subscriptions Gifting installed and activated on your site. 
3. Create a new product. Note the default checkbox options:

CleanShot 2025-06-10 at 11 22 49

4. Switch the product type to 'Subscription' or 'Variable Subscription' and comfirm you now have an 'Allow to be gifted' checkbox in the mix. It should be unchecked by default:

CleanShot 2025-06-10 at 11 28 31

5. Navigate to the 'Advanced' tab, and change 'Limit subscription' away from 'Do not limit'. Confirm that the 'Allow to be gifted' checkbox goes away. Switch it back to 'Do not limit'.
6. Make sure our show/hide behaviour works nicely: try checking the checkbox, switching the product back to a simple product, switching back to a subscription product, and confirm that it "remembers" its state while making the initial edits.
7. Fully set up and publish two subscription products, one with 'Allow to be gifted' checked, and one with 'Allow to be gifted' unchecked (so if you've checked it and change it to a simple product then back, it remains checked).
8. Run through a checkout on the front end with each product; confirm you get the 'This purchase is a gift' checkbox only on the product where it's allowed:

CleanShot 2025-06-10 at 11 51 49

9. View the individual product pages for each of the products; confirm that the 'This purchase is a gift' checkbox only appears when 'Allow to be gifted' is enabled on the product: 

CleanShot 2025-06-10 at 11 57 34

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions-gifting.php
+++ b/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions-gifting.php
@@ -19,6 +19,7 @@ class WooCommerce_Subscriptions_Gifting {
 	public static function init() {
 		\add_filter( 'wcsg_new_recipient_account_details_fields', [ __CLASS__, 'new_recipient_fields' ] );
 		\add_filter( 'wcsg_require_shipping_address_for_virtual_products', '__return_false' );
+		\add_filter( 'wcsg_is_giftable_product', [ __CLASS__, 'is_giftable_product' ], 10, 2 );
 		\add_filter( 'default_option_woocommerce_subscriptions_gifting_gifting_checkbox_text', [ __CLASS__, 'default_gifting_checkbox_text' ] );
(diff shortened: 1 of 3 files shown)


laurelfulford:
Switching this one back to 'Draft' mode while I dig into why it's not working for variable subscriptions.

PR #4029, opened 2025-06-10
feat(my-account): subscription payment notice

miguelpeixe:
### All Submissions:

This PR depends on #4016 and https://github.com/Automattic/newspack-blocks/pull/2141.

Implements a notice on every page of the "My Account" dashboard for readers with pending subscriptions or recurring donations.

If we can obtain a checkout URL from the subscription's last order, the link will trigger a modal checkout. Otherwise, we'll redirect to the subscription page and rely on the page action buttons to resolve the pending status.

image

### Donation 

1. Checkout this branch and https://github.com/Automattic/newspack-blocks/pull/2141
2. As a reader, purchase a recurring donation
3. As an admin, edit the recurring donation (WooCommerce -> Subscriptions) and create a pending renewal order:

image

4. Back to your reader session, navigate different pages of the My Account dashboard and confirm you get the notice:

image

5. Click the link and confirm that the modal checkout starts
6. Complete the checkout and confirm that you get redirected to the subscription page

### Subscriptions

1. As a reader, purchase a subscription product that is not a donation
2. Repeat steps 3 and 4 above and confirm you get the following notice:

image

3. Repeat steps 5 and 6 above

### Other information:


rbcorrales (approved):
I tested the changes with both donations and subscriptions, and the messages are displaying as expected. I also created multiple pending renewal orders for different items simultaneously, and each notice correctly rendered the appropriate link. I'm just double-checking to confirm that this displaying multiple notices is the expected behavior when there is more than one item pending attention.

image

thomasguillot:
Would it be possible to add more detail to the notice, indicating which subscription (name) or recurring donation (amount and frequency) it refers to?

I assume multiple notices are an edge case, but if they do appear, could we reduce the gap between them to 8px?

miguelpeixe:
The text and margin between notices have been updated:

image

As discussed internally, I'm also introducing this notice to our Newspack UI snackbar notices so this important message is not hidden behind the My Account dashboard:

image

This required a significant change in how we handle snackbar notices, including notice dismissal, and the provision of an API.

Dismissing a notice sends an AJAX request that can be picked up via the `newspack_ui_notice_dismissed` action hook, allowing the notice creator to handle it. In this case, dismissing the subscription payment notice keeps it from showing for 24 hours.

### Testing the frontend notice

1. As a reader, purchase a subscription and a recurring donation
2. As an admin, create a pending renewal order for both subscriptions
3. Visit the homepage as the reader and confirm that both notices render
4. Close one of the notices and refresh the page
5. Confirm only the notice you didn't close renders
6. Click the link in the notice and confirm you can go through the entire flow and once it ends, the page refreshes and no notice renders
7. Edit [this line](https://github.com/Automattic/newspack-plugin/blob/7ee768f08e7312f5151c98c671e0cc76332fc0a4/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php#L18) to `const NOTICE_INTERVAL = 10; // 10 seconds.` so you can see the dismissal expire and you can view the closed notice again
8. Refresh the page and confirm you only see the notice you've closed on step 4

thomasguillot:
@miguelpeixe I added a `closeSmall` icon because `close` seemed slightly too big, wdyt?

miguelpeixe:
Much better! Thank you

rbcorrales (approved):
I tested the changes locally, and they function as expected.

However, I noticed that after logging in, the notifications do not appear right away. They only show up after the next page refresh. I just wanted to bring this to your attention, in case it isn't the expected behavior.

miguelpeixe:
Thanks, @rbcorrales!

The notices are created on page load for an authenticated session. It won't render immediately after logging in via the auth modal because that doesn't necessarily trigger a page reload.

miguelpeixe:
@rbcorrales I just pushed a small change so these notices are restricted to My Account v1 and above. Could you double-check that the notices don't render without `define( 'NEWSPACK_MY_ACCOUNT_VERSION', '1.0.0' );`?

rbcorrales (approved):
I compared it with the previous commit (`git reset --hard HEAD^`) and it's working as expected based on the feature flag.

PR #4042, opened 2025-06-20
Add sticky post support to lite site

claudiulodro:
### All Submissions:

Note: This is a hotfix, but doesn't necessarily have to be if you disagree :)

A publisher generally needs the lite site mode for urgent breaking news (wildfires, etc.). In this situation, they're going to want/need to put the relevant breaking news story prominently at the top of the lite site. Currently it just displays the last X posts chronologically, so there isn't a way to feature the "Here's the latest coverage about this event!" post. This PR adds support for sticky posts to lite sites. That seems like a straightforward solution.

1. Enable the Lite Site feature. Make some posts sticky.
2. They should be featured prominently at the top of the lite version:

Screenshot 2025-06-20 at 11 02 23 AM

4. They should NOT affect content loop block or anything else that already has other ways of featuring content prominently.

### Other information:

### DIFF
--- a/includes/lite-site/templates/archive.php
+++ b/includes/lite-site/templates/archive.php
@@ -34,16 +34,16 @@ namespace Newspack;
 		$query_args['category__in'] = $categories;
 	}
 
-	$recent_posts = get_posts( $query_args );
-
 	// Render sticky posts prominently at the top.
 	$sticky_post_ids_option = get_option( 'sticky_posts' );
 	$sticky_post_ids        = [];
 	if ( ! empty( $sticky_post_ids_option ) ) {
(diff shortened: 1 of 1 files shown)


leogermani:
Thanks @claudiulodro !

I've pushed a change to respect the total number of posts to what's defined in the settings. And also to avoid template duplication..

Let's loop @thomasguillot in to see how we could highlight sticky posts... the `h3` tag is looking weird imo

thomasguillot:
I’m fine with using an H3. It’s a reasonable way to distinguish sticky posts from regular ones without adding CSS or increasing page load 👍

leogermani (approved):

PR #4044, opened 2025-06-23
feat(memberships): prevent membership editing if linked to a subscrip…

adekbadek:
### All Submissions:

The relationship between a WooCommerce Membership and Subscription can be complicated for some users. A customer might ask for their membership to be _cancelled_ – because they don't want to pay anymore – so the admin cancels the membership. However, the subscription is still running, and now the customer is paying and not getting membership benefits. 

This PR disables editing of the Status and start and end dates of a membership, if it's connected to a subscription:

image

1. Create a subscription-product-tied membership plan and create a user subscription (buy the product)
2. Visit the membership in WP Admin, observe the status and start and end date fields are disabled
3. Unlink the subscription (e.g. remove the `_subscription_id` meta field)
4. Reload the membership view, observe the values are editable
5. Grant such membership manually to a user, observe the values are editable

### Other information:


leogermani:
It works! But I can still Cancel the Membership via the quick links in the Memberships list

adekbadek:
@leogermani – I've added handling of post row actions and bulk edit, please have another look!

leogermani (changes requested):
It works.

Not a big fan of the approach for the bulk edit. Having the membership switching statuses twice might trigger events we wouldn't like... but if there's no way around it, let's at least add a Note to the memberhip explaining what happened

adekbadek:
@leogermani – I'm not a fan of this solution either, but could not come up with anything better in a reasonable time. This is quite an edge case, though – bulk-editing membership status. Should not be done anyway, generally. I've added code to add a note to membership.

leogermani (approved):

PR #4046, opened 2025-06-25
feat: exploring remote data blocks

leogermani:
Remote data blocks exploration

### DIFF
--- a/includes/foundation-events/class-foundation-events.php
+++ b/includes/foundation-events/class-foundation-events.php
@@ -1,0 +1,203 @@
+<?php
+/**
+ * Foundation events remote data blocks
+ *
+ * @package Newspack
+ */
+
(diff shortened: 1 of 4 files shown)


includes/foundation-events/class-foundation-events.php
```diff
@@ -0,0 +1,208 @@
...
+			}
+		}
+
+		return $input_variables;
+	}
+
+	public static function get_data_source() {
+		return [
+			'display_name'    => 'Foundation Events',
+			'endpoint'        => NEWSPACK_FOUNDATION_TEST_EVENTS_API_BASE_URL,
```
claudiulodro, line 94: Note: The site throws fatal errors if this constant is missing. It's unclear what exactly it should be, but I can work it out :)

PR #4049, opened 2025-06-26
feat(collections): support multiple CTAs in collections meta

rbcorrales:
### All Submissions:

This PR enhances the Collections feature to support multiple Call-to-Action (CTA) buttons with improved functionality:
- Instead of a single upload or link field. Each CTA can now be either an external link or a PDF file upload with customizable labels.
- All the CTAs will be recorded in a single post meta field as an array.
- CTAs can be reordered using drag-and-drop with visual feedback indicators.
- Improved the backend sanitizations and validation.
- Other smaller refactors for better code organization.

Closes NPPD-507.

1. Enable the `NEWSPACK_COLLECTIONS_ENABLED` feature flag in `wp-config.php` and the Collections module in the Newspack settings in wp-admin.
2. Create/edit a Collection post
    - Navigate to Collections > Add New or edit an existing collection.
    - Verify the "Call-to-Action" section appears in the meta panel:
image
2. Test the CTA management
    - Click "Add CTA" to create new call-to-action buttons.
    - Test both "External Link" and "File Upload" types.
    - Verify you can add/remove multiple CTAs.
image
    - Test drag-and-drop reordering by grabbing the drag handle icon in the top-right corner: 
image
3. Test validations and error handling
    - Enter invalid URLs in CTA link fields and verify that error messages appear.
    - Confirm that the global post save button becomes disabled while there are validation errors in all the URL fields at the CTA and top level.
    - Test that errors are properly maintained during drag-and-drop reordering.
    - Verify validation works independently for each field.
4. Test file attachment features
    - Switch a CTA type to "File Upload" and upload a PDF.
    - Verify the file displays with a PDF icon and download link.
    - Test the "Remove" button to clear uploaded files.
    - Confirm only PDF files are accepted for upload.
5. Test Backend Integration
    - Save a collection with multiple CTAs and verify that the data persists correctly.
    - Check that invalid CTAs (malformed URLs, non-existent attachments) are filtered out during sanitization.
    - Query the post meta via WP-CLI and verify that it keeps the correct array structure:
```php
wp post meta get <post_id> newspack_collection_ctas
array (
  0 =>
  array (
    'type' => 'link',
    'label' => 'View Digital',
    'url' => 'https://example.com/file',
  ),
  1 =>
  array (
    'type' => 'attachment',
    'label' => 'Download PDF',
    'id' => 123,
  ),
)
```

### Other information:


leogermani (commented):
> Confirm that the global post save button becomes disabled while there are validation errors in all the URL fields at the CTA and top level.

Could not confirm that. I see the error, but I clicked save draft and it saved. I clicked on the error and the error disappeared... and I'm still able to save draft or publish (imo it's ok if we just alert the error but we don't need to block saving)

Other things I noticed:

* I see that it only accepts PDFs, but there nothing in the UI indicating that. And if I try to upload a different format, there's no error. The file is not applied but there's nothing telling me why. Feels buggy

* When I upload a file, the file name overrides the label of the CTA

rbcorrales:
@leogermani thanks for checking this. I made some updates. See my replies below:

> Could not confirm that. I see the error, but I clicked save draft and it saved. I clicked on the error and the error disappeared... and I'm still able to save draft or publish (imo it's ok if we just alert the error but we don't need to block saving)

Interesting, the blocking was done using the editor store functionality, but it seems that at some point I broke that. Anyway, I'm removing this functionality as per your suggestion. The URLs are still sanitized via `esc_url_raw()` in the backend.

> I see that it only accepts PDFs, but there nothing in the UI indicating that. And if I try to upload a different format, there's no error. The file is not applied but there's nothing telling me why. Feels buggy

I'm using the `allowedTypes` prop of the [MediaUpload](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/media-upload/README.md) component to restrict it to PDF files, but there doesn't seem to be a way to prevent the selection of a recently uploaded non-PDF file. I added an error message below the file upload button for those cases, which should go away once an actual PDF is selected:
image
And also renamed the button to "Upload PDF" to be more explicit.

> When I upload a file, the file name overrides the label of the CTA

This was to get some starting value in the field, but now thinking more about it, there are probably not many cases where you would benefit from pre-populating the CTA label with the file name, even when empty. I removed this feature.

leogermani:
The URL validation seems to be linked across the CTAs..

When I enter a invalid URL, I see the warning. If I add a new CTA before fixing it, I see the warning in the new CTA, and it will only go away when I save the post - but then the warning is also removed from the CTA that still has the invalid URL

rbcorrales:
@leogermani I tried adding CTAs when there are errors, but every error seems to act independently. I'm not able to replicate that behavior. 🤔 
Am I doing the same steps as you in this video?

https://github.com/user-attachments/assets/5dddb4c5-3de3-4226-bfa4-9ed1b3e12754

leogermani:
Idk, it happened once

leogermani (approved):

PR #4064, opened 2025-07-02
fix(log): ensure `user_email` is always a string

miguelpeixe:
### All Submissions:

If the user email is not in `$data` and the user is not logged in, the value of `$email` will be `false` (from `wp_get_current_user()->user_email`). The `user_email` property must always be a string; otherwise, the log does not reach the server.

1. While on `release`, make sure you have newspack-manager active and Jetpack connected
2. Run `wp eval "Newspack\Logger::newspack_log( 'test', 'test' );"`
3. Confirm the entry never reaches the server
4. Checkout this branch and run the same command
5. Confirm the entry reaches the server

### Other information:

### DIFF
--- a/includes/class-logger.php
+++ b/includes/class-logger.php
@@ -116,7 +116,7 @@ class Logger {
 			[
 				'type'       => $type,
 				'data'       => $data,
-				'user_email' => $email ? $email : '', // always cast to string.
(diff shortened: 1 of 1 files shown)


leogermani (approved):
`(string) $email` would be simpler, but that works too

PR #4065, opened 2025-07-03
fix: ensure my-account is working w/out woocommerce

adekbadek:
### All Submissions:

Some code does not account for WooCommerce being inactive. 

1. On `trunk`
2. Set `NEWSPACK_MY_ACCOUNT_VERSION` to `1.0.0` to get the new My Account UI
3. Deactivate WooCommerce
4. Observe a hard crash on homepage
5. Switch to this branch, observe the homepage loading

### Other information:

### DIFF
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -196,7 +196,7 @@ class ESP_Sync extends Sync {
 	 */
 	public static function get_contact_data( $user_id ) {
 		if ( ! class_exists( '\WC_Customer' ) ) {
-			return;
+			return new \WP_Error( 'newspack_esp_sync_contact', __( 'WC_Customer class unavailable.', 'newspack-plugin' ) );
 		}
 		$user = \get_userdata( $user_id );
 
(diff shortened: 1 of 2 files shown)


naxoc (commented):
Looks good!

Do we only need `my-account-v1.js` if Woo is active?

I like the early return, I just want to make sure we are not returning before adding stuff that doesn't rely on Woo?

includes/reader-activation/sync/class-esp-sync.php
```diff
@@ -195,6 +195,9 @@ public static function scheduled_sync( $user_id, $context ) {
 	 * @return array|\WP_Error The contact data or WP_Error.
 	 */
 	public static function get_contact_data( $user_id ) {
+		if ( ! class_exists( '\WC_Customer' ) ) {
+			return;
+		}
```
miguelpeixe: Per the method docblock, it's best to return a WP_Error here

adekbadek: Good call, updated!

adekbadek:
@naxoc 

> Do we only need my-account-v1.js if Woo is active?

> I like the early return, I just want to make sure we are not returning before adding stuff that doesn't rely on Woo?

You're right – `myAccountUrl` is only used in `my-account-v1.js`, so the solution _could_ be more granular. However, this whole class is about the "My Account" page, which is a WC thing. So it makes sense, for simplicity, to bail early if WC is not available.

naxoc (approved):

miguelpeixe (approved):

PR #4081, opened 2025-07-15
feat: Indesign post export actions

Takshil-Kunadia:
### All Submissions:

Closes https://app.asana.com/1/26890605006346/project/1210562189110460/task/1210765993787106?focus=true .

Implements Milestone 2 of the Adobe InDesign Exporter feature by adding a new bulk action to the Posts admin screen: “Export as Adobe InDesign”. It also Adds a row action for each post: “Export as Adobe InDesign”.

On action triggers, the existing formatter class (from Milestone 1) is used to generate the .txt content in InDesign Tagged Text format.

1. Go to post editor screen.
2. Hover over a post and click `Export as Adobe Indesign`.
3. Check the downloaded file and verify the output.
4. Now select multiple posts > click on Bulk actions > select `Export as Adobe Indesign` option.
5. Click `Apply`.
6. Extract the zip and verify the contents of the files and its output.

### Other information:


includes/optional-modules/class-indesign-exporter.php
```diff
@@ -57,22 +57,255 @@ public static function is_feature_enabled() {
...
+	 */
+	public static function get_supported_post_types() {
+		$supported_post_types = [ 'post' ];
+
+		/**
+		 * Filters the post types that support InDesign export.
+		 *
+		 * @param array $supported_post_types Array of post type names that support InDesign export.
+		 */
+		return apply_filters( 'newspack_indesign_export_supported_post_types', $supported_post_types );
```
claudiulodro, line 76: I like that this filter is available. Good idea. You'll need to use `get_supported_post_types()` when you're setting things up in the `init()` method though. 

Currently you are setting up the actions on 'bulk_actions-edit-**post**', '**post**_row_actions', etc. so they are only ever displayed on the 'post' post type.

includes/optional-modules/class-indesign-exporter.php
```diff
@@ -57,22 +57,255 @@ public static function is_feature_enabled() {
...
+	}
 
-		$converter = new InDesign_Converter();
-		$content = $converter->convert_post( $test_post );
+	/**
+	 * Download multiple files as ZIP.
+	 *
+	 * @param array $files Array of file data.
+	 */
+	private static function download_zip_file( $files ) {
```
claudiulodro, line 250: This is a great solution for downloading multiple files at once. Good idea!

claudiulodro (approved):
Thanks! This looks solid and works great in my testing! I just left one thing that should be fixed if you want the filter to support other post types to be useful. That's not a blocker though, since it only affects some possible future integration.

includes/optional-modules/class-indesign-exporter.php
Takshil-Kunadia: Nice catch! It was only showing up as row action for cpts until now.
I've updated the init() method so the bulk actions now also register correctly for all supported post types.
thanks 🙇 .

Takshil-Kunadia:
Made the fix ✔️ . Merging this PR.

PR #4084, opened 2025-07-15
feat(ads): parent ad unit inventory

miguelpeixe:
### All Submissions:

Wizard changes required to implement https://github.com/Automattic/newspack-ads/pull/967

Further details and instructions at https://github.com/Automattic/newspack-ads/pull/967

### Other information:

### DIFF
--- a/src/wizards/advertising/views/ad-units/index.js
+++ b/src/wizards/advertising/views/ad-units/index.js
@@ -22,7 +22,7 @@ const CREATE_AD_ID_PARAM = 'create';
 /**
  * Advertising management screen.
  */
-const AdUnits = ( { adUnits, onDelete, wizardApiFetch, updateWithAPI, service, serviceData, fetchAdvertisingData } ) => {
+const AdUnits = ( { adUnits, parentAdUnits, onDelete, wizardApiFetch, updateWithAPI, service, serviceData, fetchAdvertisingData } ) => {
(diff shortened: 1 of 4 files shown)


src/wizards/advertising/views/ad-units/index.js
```diff
@@ -70,15 +78,31 @@ const AdUnits = ( { adUnits, onDelete, wizardApiFetch, updateWithAPI, service, s
...
+					<SelectControl
+						label={ __( 'Connected GAM network code', 'newspack-plugin' ) }
+						value={ networkCode }
+						options={ serviceData.available_networks.map( network => ( {
+							label: `${ network.name } (${ network.code })`,
+							value: network.code,
+						} ) ) }
+						onChange={ updateGAMNetworkCode }
+					/>
+					{ parentAdUnits.length > 0 && (
```
rbcorrales, line 105: We could maybe show a message if there are no parent units for user feedback, so users understand their options better.

rbcorrales (approved):
Working as expected.

src/wizards/advertising/views/ad-units/index.js
miguelpeixe: I'd say this is a niche feature and only publishers with GAM already structured in that way will benefit from it. I don't think we need to advertise the option in the UI. But I'll defer to @kmwilkerson on that.

PR #4093, opened 2025-07-17
Fix/scripts add hooks in Lite Site

leogermani:
### All Submissions:

Lite site: Fixes handling of script tags and add two new hooks

1. Add a custom HTML block to a post and some simple JS to it like `<script>alert(1)</script>`
2. Visit the post on the Lite site version of it
3. on `trunk` you'll see that `alert(1)` shows up as plain text
4. on this branch the script is gone
5. Check the new hooks in the code and confirm they look good

### Other information:

### DIFF
--- a/includes/lite-site/class-lite-site.php
+++ b/includes/lite-site/class-lite-site.php
@@ -440,10 +440,10 @@ class Lite_Site {
 		$content = preg_replace( '/<!--(.|\s)*?-->/', '', $content );
 
 		// First remove figures and their contents (including images and captions).
-		$content = preg_replace( '/<figure.*?>.*?<\/figure>/s', '', $content );
+		$content = preg_replace( '/<figure.*?>.*?<\/figure>/is', '', $content );
 
 		// Remove script tags.
-		$content = preg_replace( '/<script.*?>.*?<\/script>/s', '', $content );
+		$content = preg_replace( '/<script.*?>.*?<\/script>/is', '', $content );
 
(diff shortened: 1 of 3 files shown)


includes/lite-site/class-lite-site.php
```diff
@@ -442,6 +442,9 @@ public static function clean_content( $content ) {
 		// First remove figures and their contents (including images and captions).
 		$content = preg_replace( '/<figure.*?>.*?<\/figure>/s', '', $content );
 
+		// Remove script tags.
+		$content = preg_replace( '/<script.*?>.*?<\/script>/s', '', $content );
```
rbcorrales: It works, but is case sensitive, so this will still be printed:
```js
<SCRIPT>alert(1)</SCRIPT>
```

We should add the `i` modifier to the regex to prevent this:
```suggestion
		$content = preg_replace( '/<script.*?>.*?<\/script>/is', '', $content );
```

rbcorrales (changes requested):

includes/lite-site/class-lite-site.php
leogermani: Thanks. Fixed that in the `figure` tag a few lines above as well in https://github.com/Automattic/newspack-plugin/pull/4093/commits/b30b90bfc57e1f9ddea86b85a3f10c1e07ea032e

rbcorrales (approved):

PR #4095, opened 2025-07-18
hotfix(google-oauth): race conditions and cross-site issues

adekbadek:
### All Submissions:

Google OAuth login was failing to refresh the page after successful authentication on Chrome 138 with macOS Sonoma due to Cross-Origin-Opener-Policy (COOP) restrictions blocking access to `window.closed`.

#### Solution
Replaced the unreliable `window.closed` polling with a robust `postMessage` API for cross-origin communication between the OAuth popup and parent window.

#### Key Changes
- **Fixed COOP compatibility**: Replaced `window.closed` checks with `postMessage` communication
- **Improved reliability**: OAuth callback now uses `postMessage` to signal success to parent window
- **Enhanced debugging**: Added centralized debug logging system accessible via `window.newspackReaderActivation.debugLog`
- **Better error handling**: Graceful fallbacks for cross-origin communication failures

#### Files Modified
- `src/reader-activation-auth/google-oauth.js` - Updated to use postMessage listeners
- `includes/oauth/class-google-login.php` - OAuth callback sends postMessage to parent window
- `src/memberships-gate/gate.js` - Added support for `reader_logged_in` action detection
- `src/reader-activation/index.js` - Added centralized debug logging function

1. Ensure Google OAuth is configured in Newspack settings
1. Set up a membership gate on a test post/page
1. Navigate to a gated article while logged out
1. Run: `localStorage.setItem('newspack-google-oauth-debug', 'true')`
1. Click "Sign in with Google" button
1. Complete Google OAuth flow in popup window
1. **Expected**: Page should automatically refresh and show ungated content
1. **Expected**: No console errors related to `window.closed`
1. **Expected**: Console shows detailed logs of the OAuth process:
   - `[Google OAuth] Login button clicked`
   - `[Google OAuth] Received postMessage`
   - `[Gate] reader_logged_in action detected`
   - `[Gate] Reloading page!`

### Other information:

### DIFF
--- a/src/memberships-gate/gate.js
+++ b/src/memberships-gate/gate.js
@@ -3,20 +3,9 @@
  * Internal dependencies
  */
 import './gate.scss';
+import { debugLog } from '../reader-activation/utils';
 
 const EVENT_NAME = 'np_gate_interaction';
-
-/**
- * Debug logging function that uses the centralized logging from newspackReaderActivation.
- *
- * @param {string} level Log level ('log' or 'error')
- * @param {...any} args  Arguments to pass to console
- */
-const debugLog = ( level, ...args ) => {
-	if ( window.newspackReaderActivation?.debugLog ) {
-		window.newspackReaderActivation.debugLog( level, ...args );
-	}
-};
 
 /**
  * Specify a function to execute when the DOM is fully loaded.
--- a/src/memberships-gate/gate.js
+++ b/src/memberships-gate/gate.js
@@ -7,6 +7,18 @@ import './gate.scss';
 const EVENT_NAME = 'np_gate_interaction';
 
 /**
+ * Debug logging function that uses the centralized logging from newspackReaderActivation.
+ *
+ * @param {string} level Log level ('log' or 'error')
+ * @param {...any} args  Arguments to pass to console
+ */
+const debugLog = ( level, ...args ) => {
+	if ( window.newspackReaderActivation?.debugLog ) {
+		window.newspackReaderActivation.debugLog( level, ...args );
+	}
+};
+
+/**
  * Specify a function to execute when the DOM is fully loaded.
  *
  * @see https://github.com/WordPress/gutenberg/blob/trunk/packages/dom-ready/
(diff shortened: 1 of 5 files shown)


src/reader-activation/index.js
```diff
@@ -336,6 +336,21 @@ export function getAuthStrategy() {
...
+ * @param {...any} args  Arguments to pass to console.
+ */
+// eslint-disable-next-line no-console
+export function debugLog( level = 'log', ...args ) {
+	if ( localStorage.getItem( 'newspack-google-oauth-debug' ) === 'true' ) {
+		const method = level === 'error' ? 'error' : 'log';
+		// eslint-disable-next-line no-console
+		console[ method ]( ...args );
+	}
+}
```
miguelpeixe: We can avoid code repetition by placing this in `src/reader-activation/utils.js` and importing.

We should also use a more generic name here; `newspack-google-oauth-debug` sounds too specific for this function.

src/memberships-gate/gate.js
```diff
@@ -64,11 +86,22 @@ function initReloadHandler() {
...
 			// triggered another overlay – to be executed (setTimeout hack).
 			setTimeout( () => {
-				if ( ! ras.overlays.get().length && reload ) {
+				const overlays = ras.overlays.get();
+				const activities = window?.newspackReaderActivation?.getActivities();
+				debugLog( 'log', '[Gate] Checking reload conditions - overlays:', overlays, 'reload:', reload );
+				debugLog( 'log', '[Gate] Current activities:', activities );
+				if ( ! overlays.length && reload ) {
+					debugLog( 'log', '[Gate] Reloading page!' );
 					window.location.reload();
```
miguelpeixe, line 94: Non-blocking, but I think this is unnecessary logging and pollutes the codebase. Why are the activities relevant here?  "Reloading page!" is not going to be there for more than a millisecond.

src/reader-activation-auth/google-oauth.js
```diff
@@ -14,38 +26,73 @@ domReady( function () {
...
+				if ( event.origin === window.location.origin && event.data === 'google-oauth-success' ) {
+					debugLog( 'log', '[Google OAuth] Valid success message received, calling checkLoginStatus' );
+					cleanup();
+					checkLoginStatus( metadata );
+				}
+			};
+			window.addEventListener( 'message', messageListener );
+
+			// Keep the legacy event listener for backwards compatibility.
 			window.addEventListener( 'google-oauth-success', () => {
```
miguelpeixe: What would still be dispatching this event?

miguelpeixe (changes requested):
Tests great and good idea with the frontend logger!

src/memberships-gate/gate.js
adekbadek: Not if you preserve console log, which might be useful for debugging why the page did not reload when it should have 
image

src/reader-activation-auth/google-oauth.js
adekbadek: Yeah, that's overeager. It was meant as a super-safe backwards compatibility, but there's no need.

src/reader-activation/index.js
adekbadek: Good idea, added in 2e9f5ae5d103e82cbac1b4fed7436430523a19fc

src/memberships-gate/gate.js
miguelpeixe: Is there any scenario in which `window.location.reload()` does not reload the page? Also unclear to me why "activities" are a relevant resource for debugging the gate interaction.

miguelpeixe (approved):
Thank you for revising! Approving since the logging discussion is not a blocker.

src/memberships-gate/gate.js
adekbadek: > Is there any scenario in which window.location.reload() does not reload the page?

No idea, but the implementation might change in the future ¯\_(ツ)_/¯ I just like to have the logging follow the complete user path, so you can just look at the console for the whole story. But it's not a strong opinion. 

> Also unclear to me why "activities" are a relevant resource for debugging the gate interaction.

This data is used in determining the conditions for the `reload` variable (line 61, "When closing an overlay, check if the last activity was a checkout, registration, or login."). So might be useful to see why `reload` is what it is. But there's already `[Gate] Valid action detected …`. Anyway, not a strong opinion too.

PR #4100, opened 2025-07-21
feat: add collections to block filters (NPPD-717) 

laurelfulford:
### All Submissions:

This PR adds an option to filter posts by Collection in the Content Loop & Content Carousel blocks. It's a more reasonable version of https://github.com/Automattic/newspack-blocks/pull/2168 😅 

1. If not already, enable Collections under WP Admin > Newspack > Settings on your test site.
2. Make sure you have some Collections set up on your test site, and posts populating each of them.
3. Create a page and add a Content Loop block. In the Content panel of the right sidebar, confirm you have a Collections filter along with other options. Collections and Exclude Collections should appear above Collection Sections and Exclude Collection Sections, respectively. (I had to do something a little hacky with the taxonomies order to get this to work, otherwise Collections was showing up after Collection Sections, and after Sponsors on my test site).

image

4. Type in one of your collection names; confirm you get some autocomplete behaviour.
5. Pick the Collection; confirm your block now only shows that collection.
6. Publish your page and confirm the correct stories appear on the front-end.
7. Set up a post that's assigned to a Collection, and a Collection Section within that Collection.
8. In a Content Loop block, assign both the Collection and Collection Section; confirm you get the correct post.
9. Publish your page and confirm the correct stories appear on the front-end.
10. Repeat steps 3-9 with the Content Carousel block.
11. Disable the Collections module under WP Admin > Newspack > Settings.
12. Confirm that the Collections query option is no longer available under the Content Loop and Carousel block settings (ditto Collection Section, but that should happen automagically). Any blocks set to return posts based on Collection should be empty.

### Other information:

### DIFF
--- a/includes/collections/class-collection-taxonomy.php
+++ b/includes/collections/class-collection-taxonomy.php
@@ -228,19 +228,16 @@ class Collection_Taxonomy {
 			'label' => __( 'Collections', 'newspack-plugin' ),
 		];
 
-		// Don't add a duplicate if this taxonomy is already in the list.
-		foreach ( $custom_taxonomies as $tax ) {
-			if ( $tax['slug'] === $collections_taxonomy['slug'] ) {
-				return $custom_taxonomies;
-			}
-		}
-
-		// Capture where Collection Sections is in the sidebar.
 		$point_of_insertion = null;
+		// Loop through the taxonomies; confirm Collections doesn't already exist, and grab the index of Collection Sections if it's there.
 		foreach ( $custom_taxonomies as $index => $tax ) {
-			if ( isset( $tax['slug'] ) && $tax['slug'] === 'newspack_collection_section' ) {
-				$point_of_insertion = $index;
-				break;
+			if ( isset( $tax['slug'] ) ) {
+				if ( $tax['slug'] === $collections_taxonomy['slug'] ) {
+					return $custom_taxonomies;
+				}
+				if ( $tax['slug'] === Collection_Section_Taxonomy::get_taxonomy() ) {
+					$point_of_insertion = $index;
+				}
 			}
 		}
 
--- a/includes/collections/class-collection-taxonomy.php
+++ b/includes/collections/class-collection-taxonomy.php
@@ -64,6 +64,7 @@ class Collection_Taxonomy {
 	 */
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_taxonomy' ] );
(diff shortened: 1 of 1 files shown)


includes/collections/class-collection-taxonomy.php
```diff
@@ -214,4 +215,42 @@ public static function find_term_by_name( $term_name, $meta_conditions = [] ) {
...
+		foreach ( $custom_taxonomies as $tax ) {
+			if ( $tax['slug'] === $collections_taxonomy['slug'] ) {
+				return $custom_taxonomies;
+			}
+		}
+
+		// Capture where Collection Sections is in the sidebar.
+		$point_of_insertion = null;
+		foreach ( $custom_taxonomies as $index => $tax ) {
+			if ( isset( $tax['slug'] ) && $tax['slug'] === 'newspack_collection_section' ) {
```
rbcorrales: You can use `Collection_Section_Taxonomy::get_taxonomy()` here instead of hardcoding it.

includes/collections/class-collection-taxonomy.php
```diff
@@ -214,4 +215,42 @@ public static function find_term_by_name( $term_name, $meta_conditions = [] ) {
...
+				$point_of_insertion = $index;
+				break;
+			}
+		}
```
rbcorrales: You can avoid double-looping the same array and merge the logic:

```suggestion
		$point_of_insertion = null;
		foreach ( $custom_taxonomies as $index => $tax ) {
			if ( isset( $tax['slug'] ) ) {
				if ( $tax['slug'] === $collections_taxonomy['slug'] ) {
					return $custom_taxonomies;
				}
				if ( $tax['slug'] === Collection_Section_Taxonomy::get_taxonomy() ) {
					$point_of_insertion = $index;
				}
			}
		}
```

rbcorrales (changes requested):
Works as expected, but a few minor improvement suggestions.

includes/collections/class-collection-taxonomy.php
laurelfulford: Fixed! 🙂

includes/collections/class-collection-taxonomy.php
laurelfulford: Ah that's much nicer - updated!

laurelfulford:
Thanks @rbcorrales - both are nice improvements! This should be ready for your eyes again.

rbcorrales (approved):
Thanks for adding this functionality to Collections and addressing the comments!

laurelfulford:
Thanks for the review, @rbcorrales! 🎉

PR #4101, opened 2025-07-21
Fix CLI module activation logic

claudiulodro:
### All Submissions:

Fixes a tiny bug that I ran into today. The logic of the success/error message when activating an optional module via the CLI is not correct. It will throw an "Error: Cannot activate module – it's already active." error after successfully activating a module. This PR fixes that issue and makes a small readability improvement to the deactivate command. 

This does NOT need to go out as a hotfix since the command actually works fine and only we ever see its output.

1. Before applying this PR, have the InDesign module deactivated and run `wp newspack optional-module activate indesign-export`. Observe it will activate the module **and** say "Error: Cannot activate module – it's already active.".
2. Apply this PR. 
3. Via CLI, activate an inactive module and observe it says that it did so successfully. 
4. Try and activate it again and observe it won't let you. 
5. Deactivate the module. Observe it does so successfully. 
6. Deactivate it again, and observe it won't let you.

### Other information:

### DIFF
--- a/includes/cli/class-optional-modules.php
+++ b/includes/cli/class-optional-modules.php
@@ -78,12 +78,15 @@ class Optional_Modules {
 			\WP_CLI::error( sprintf( 'Module is not available. These are the available modules: %s', implode( ', ', $available_modules ) ) );
 		}
 
-		$settings = \Newspack\Optional_Modules::activate_optional_module( $module_name );
+		if ( \Newspack\Optional_Modules::is_optional_module_active( $module_name ) ) {
+			\WP_CLI::error( "Cannot activate module – it's already active." );
+		}
 
-		if ( empty( $settings[ \Newspack\Optional_Modules::MODULE_ENABLED_PREFIX . $module_name ] ) ) {
+		$settings = \Newspack\Optional_Modules::activate_optional_module( $module_name );
+		if ( ! empty( $settings[ \Newspack\Optional_Modules::MODULE_ENABLED_PREFIX . $module_name ] ) ) {
 			\WP_CLI::success( 'Module activated.' );
 		} else {
-			\WP_CLI::error( "Cannot activate module – it's already active" );
+			\WP_CLI::error( 'Failed to activate module.' );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


includes/cli/class-optional-modules.php
```diff
@@ -78,12 +78,15 @@ public static function cmd_activate_module( array $pos_args, array $assoc_args )
 			\WP_CLI::error( sprintf( 'Module is not available. These are the available modules: %s', implode( ', ', $available_modules ) ) );
 		}
 
-		$settings = \Newspack\Optional_Modules::activate_optional_module( $module_name );
+		if ( \Newspack\Optional_Modules::is_optional_module_active( $module_name ) ) {
+			\WP_CLI::error( "Cannot activate module – it's already active." );
```
adekbadek: Suggestion: 

```suggestion
			\WP_CLI::warning( "Cannot activate module – it's already active." );
			return;
```

includes/cli/class-optional-modules.php
```diff
@@ -98,13 +101,11 @@ public static function cmd_activate_module( array $pos_args, array $assoc_args )
 	public static function cmd_deactivate_module( array $pos_args, array $assoc_args ) {
 		$module_name = $pos_args[0];
 
-		$enabled_modules = array_filter( \Newspack\Optional_Modules::get_settings() );
-
-		if ( empty( $enabled_modules[ \Newspack\Optional_Modules::MODULE_ENABLED_PREFIX . $module_name ] ) ) {
+		if ( ! \Newspack\Optional_Modules::is_optional_module_active( $module_name ) ) {
 			\WP_CLI::error( 'Module is not active. Cannot deactivate it' );
```
adekbadek:
```suggestion
			\WP_CLI::warning( 'Module is not active. Cannot deactivate it' );
			return;
```

adekbadek (approved):
Left a suggestion to use warnings instead of errors, so it's clear that nothing wrong happened.

includes/cli/class-optional-modules.php
claudiulodro: Makes sense!

PR #4106, opened 2025-07-23
feat(rss): Filtering via custom taxonomies

Takshil-Kunadia:
### All Submissions:

This adds support for filtering RSS feed content by registered custom taxonomy terms. Each custom taxonomy now appears in the feed editor's content settings UI metabox with fields to include or exclude specific terms. When terms are selected:
- The Include field ensures only posts with those terms appear in the feed.
- The Exclude field ensures posts with those terms are excluded from the feed.

Closes https://app.asana.com/1/26890605006346/project/1210562189110460/task/1210861378590738?focus=true .

1. Register a few custom taxonomies and create some test terms in them
2. Create some posts and assign to these terms.
3. Go to RSS Feeds > New Feed.
4. Select terms in "Include only posts with <custom taxonomy>" fields.
5. Save the field and open the preview link.
6. Verify that posts with only these custom taxonomy terms appear in the feed.

### Other information:

### DIFF
--- a/includes/optional-modules/class-rss.php
+++ b/includes/optional-modules/class-rss.php
@@ -73,6 +73,7 @@ class RSS {
 			'category_include'          => [],
 			'category_exclude'          => [],
 			'tag_include'               => [],
+			'category_tag_relation'     => 'AND',
 			'use_image_tags'            => false,
 			'use_media_tags'            => false,
 			'use_updated_tags'          => false,
@@ -388,6 +389,18 @@ class RSS {
 						<?php foreach ( $tags as $tag ) : ?>
 							<option value="<?php echo esc_attr( $tag->term_id ); ?>" <?php selected( in_array( $tag->term_id, $settings['tag_include'] ) ); ?>><?php echo esc_html( $tag->name ); ?></option>
 						<?php endforeach; ?>
+					</select>
+				</td>
+			</tr>
+			<tr>
+				<th>
+					<?php esc_html_e( 'Category and tag filter relationship:', 'newspack-plugin' ); ?>
+					<p class="description"><?php echo esc_html_x( 'When both categories and tags are selected above, should posts match ALL conditions (AND) or ANY condition (OR)?', 'help text for category/tag relation', 'newspack-plugin' ); ?></p>
+				</th>
+				<td>
+					<select name="category_tag_relation">
+						<option value="AND" <?php selected( $settings['category_tag_relation'], 'AND' ); ?> ><?php esc_html_e( 'AND - Posts must match both category AND tag filters', 'newspack-plugin' ); ?></option>
+						<option value="OR" <?php selected( $settings['category_tag_relation'], 'OR' ); ?> ><?php esc_html_e( 'OR - Posts can match either category OR tag filters', 'newspack-plugin' ); ?></option>
 					</select>
 				</td>
 			</tr>
@@ -736,6 +749,11 @@ class RSS {
 			}
 		}
 
+		$category_tag_relation = filter_input( INPUT_POST, 'category_tag_relation', FILTER_SANITIZE_SPECIAL_CHARS );
+		if ( in_array( $category_tag_relation, [ 'AND', 'OR' ], true ) ) {
+			$settings['category_tag_relation'] = $category_tag_relation;
+		}
+
 		// Process Republication Tracker options only if the plugin is active.
 		if ( self::is_republication_tracker_plugin_active() ) {
 			$republication_tracker             = filter_input( INPUT_POST, 'republication_tracker', FILTER_SANITIZE_NUMBER_INT );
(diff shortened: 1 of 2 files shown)


Takshil-Kunadia:
Hi @leogermani @adekbadek I have merged `feat/rss-tags` branch into this and updated the logic. This can be merged once https://github.com/Automattic/newspack-plugin/pull/4107 is merged.

leogermani:
> Hi @leogermani @adekbadek I have merged `feat/rss-tags` branch into this and updated the logic. This can be merged once https://github.com/Automattic/newspack-plugin/pull/4107 is merged. 

Thanks! I have one more ask... The current component loads all the terms in the dom to provide that UI. This is a problem especially for tags, because some sites have thousands of tags and this can break the browser.

Ideally this UI would be a true auto complete, loading terms via ajax. At least for tags or other non hierarchical taxonomies.

Takshil-Kunadia:
Thanks for the suggestion @leogermani ! I’ve added AJAX search for categories, tags, and custom taxonomies, so it now behaves as a true autocomplete. happy to tweak further if needed.

leogermani:
ok, I started tweaking a few things in this PR and it turned out to be a major change.

1 - For the sake of simplicity, I removed the "exclude" option from the custom taxonomies and tags... It's hard wrap the head around the relationships between the queries and the exclusions... also, I anticipate a request where people would want a relationship OR between taxonomies, but that the exclusions were always respected, which is not possible

2- On the same topic, @adekbadek left a comment in the code, keeping the `category__not_in` separate because it should always be applied. I wrote a bunch of tests and noticed that this was not true. the "NOT IN" rule would follow the same `relation` rule of the tax_query, even if declared outside the tax_query... so for simplicity, we are always using tax_query now

3 - Reorganized the UI, creating a section for the taxonomies option, changed the label of the "relationship" option and things are more clear

4- Wrote a bunch of tests to make sure things are in place

5 - As a bonus, added a new option, for each taxonomy, to choose the relationship between the terms (all terms in the tax, or any term in the tax)

here's the final UI:

image

includes/optional-modules/class-rss.php
```diff
@@ -1141,6 +1339,67 @@ private static function is_republication_tracker_plugin_active() {
...
+	 * Get custom taxonomies registered for posts (excluding built-in taxonomies).
+	 *
+	 * @return array Array of custom taxonomy names registered for posts.
+	 */
+	private static function get_custom_taxonomies_for_posts() {
+		$custom_taxonomies = get_taxonomies(
+			[
+				'object_type' => [ 'post' ],
+				'public'      => true,
+				'_builtin'    => false,
```
claudiulodro: I don't know if I'm misunderstanding how this works, but I don't see [the Brand taxonomy from Newspack Multibranded](https://github.com/Automattic/newspack-multibranded-site/blob/ac2076d9cad38353bb67ba72a19d43d67d826016/includes/class-taxonomy.php#L88) nor [the Sponsors taxonomy from Newspack Sponsors](https://github.com/Automattic/newspack-sponsors/blob/438f04eca64a6c54bdacdd8321d14e2624e431f0/includes/class-core.php#L316) when editing an RSS feed. Both are registered as public as far as I can tell.

includes/optional-modules/class-rss.php
```diff
@@ -466,9 +585,31 @@ public static function render_content_settings_metabox( $feed_post ) {
...
 			jQuery( document ).ready( function() {
-				jQuery( '#category_include' ).select2();
-				jQuery( '#category_exclude' ).select2();
-				jQuery( '#tag_include' ).select2();
+				jQuery( '.newspack-ajax-taxonomy-select' ).select2( {
+					ajax: {
+						url: ajaxurl,
+						dataType: 'json',
+						type: 'POST',
+						delay: 250,
```
claudiulodro: Minor nitpick: I think maybe we want a bigger delay here, something like 2 seconds? I don't think the average person is typing more than 4 characters a second, so the delay isn't throttling a lot of ajax calls. It's generating an ajax request on most characters in my testing.

Screenshot 2025-07-25 at 3 27 35 PM

claudiulodro (approved):
This looks solid! Works well in my testing on a site with 50k tags, and I love that there's a ton of unit tests. I just left a couple non-blocking questions. Thanks!

includes/optional-modules/class-rss.php
leogermani: Good catch! That query was retrieving only taxonomies that applied ONLY to posts, and not any other post types, like pages.

fixed in https://github.com/Automattic/newspack-plugin/pull/4106/commits/bd84694213dbdeeb396e7c50fb710d19b3f4a8a3

includes/optional-modules/class-rss.php
leogermani: 2 seconds felt appropriate, let me know what you think

https://github.com/Automattic/newspack-plugin/pull/4106/commits/8525d7068f7ab922d2cdbe9ba9fcb66b35b16e4f

includes/optional-modules/class-rss.php
claudiulodro: Screenshot 2025-07-28 at 7 00 39 AM
Looks good!

includes/optional-modules/class-rss.php
claudiulodro: Feels good to me!

claudiulodro (approved):
Looks solid with the latest revisions!

PR #4107, opened 2025-07-23
feat(rss): category and tag inclusion: AND or OR condition

adekbadek:
### All Submissions:

In RSS Feeds settings, it will now be possible control the mixing of category and tag inclusions. Before, if both were set, an `AND` condition would be implicit. With this change, the user has the option to set this condition to either `AND` or `OR`.

Closes NPPM-2116

1. Edit or create an RSS Feed
2. Add category and tag inclusions, observe posts have to have both the categories and tags to be included
3. Change the "Category and tag filter relationship:" setting to `OR` and observe that posts with either the categories or tags are included

### Other information:

### DIFF
--- a/includes/optional-modules/class-rss.php
+++ b/includes/optional-modules/class-rss.php
@@ -73,6 +73,7 @@ class RSS {
 			'category_include'          => [],
 			'category_exclude'          => [],
 			'tag_include'               => [],
+			'category_tag_relation'     => 'AND',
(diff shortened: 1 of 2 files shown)


leogermani:
we're also adding support for custom taxonomies, so this will need to be adapted. I'll let @Takshil-Kunadia know

leogermani (approved):

PR #4108, opened 2025-07-23
feat(collections): add global setting for showing categories

rbcorrales:
### All Submissions:

The posts in the single pages template are rendered by reusing the `newspack-blocks/homepage-articles`. However, the arguments passed are fixed, and updating them requires adding code to the site. This PR adds a new "Show category" toggle to the global Collections settings that, when enabled, will append the `showCategory => true` attribute to the `render_block()` call.

Moreover, the way this change is structured allows future extensibility via the settings UI or by updating the option directly in the database, without modifying the `render_article()` logic.

The global Collections settings are stored in an array, and we're adding a nested `articles_block_attrs` array, whose values will get merged with the default block attributes:
```json
wp option get newspack_collections_settings --format=json
{
  "post_indicator_style": "card",
  "posts_per_page": 24,
  ...
  ...
  "articles_block_attrs": {
    "showCategory": true
  }
}
```

So, for example, if we want to modify the default behavior and add the `showDate` attribute, we can manipulate the option directly by overwriting the `articles_block_attrs` array:
```bash
wp option patch update newspack_collections_settings articles_block_attrs '{"showCategory":true,"showDate":true}' --format=json
```

Or extracting the values and modifying the specific nested key:
```bash
SETTINGS=$(wp option get newspack_collections_settings --format=json | jq -c '.articles_block_attrs.showDate = true') && wp option update newspack_collections_settings "$SETTINGS" --format=json
```

Closes NPPD-728.

1. Navigate to Newspack Settings > Collections and enable the module.
2. In the "Collection Single" section, enable the "Show category" toggle.
3. Go to a collection singular page and verify that the post category appears at the top of each article.
4. Turn off the toggle and verify that the categories disappear.
5. Test setting other [block attributes](https://github.com/Automattic/newspack-blocks/blob/4000d858b6931137d883e06b2b6c1200a420dfa7/src/blocks/homepage-articles/block.json#L5-L258) by directly modifying the `newspack_collections_settings` option via the methods described above and verify that the changes get reflected in the frontend.

### Other information:

### DIFF
--- a/includes/collections/class-settings.php
+++ b/includes/collections/class-settings.php
@@ -90,6 +90,12 @@ class Settings {
 				'default'           => false,
 				'sanitize_callback' => 'rest_sanitize_boolean',
 			],
+			// Collection Single section.
+			'articles_block_attrs'  => [
+				'required'          => false,
+				'default'           => [],
+				'sanitize_callback' => [ self::class, 'sanitize_articles_block_attrs' ],
+			],
 			// Collection Posts section.
 			'post_indicator_style'  => [
 				'required'          => false,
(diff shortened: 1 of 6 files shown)


jorostoyanov (approved):
Looking good @rbcorrales 💯 

Thanks!

PR #4130, opened 2025-08-06
feat: add the collections category to the Content Loop filters (NPPD-729)

laurelfulford:
### All Submissions:

Similar to https://github.com/Automattic/newspack-plugin/pull/4100, this PR adds the Collection Categories taxonomy as an option to filter Collections by. 

1. If not already, enable Collections under WP Admin > Newspack > Settings on your test site.
2. Make sure you have some Collections set up on your test site, and posts populating each of them.
3. Assign a couple different Collection Categories to your Collections.
4. Create a page and add a Content Loop block. In the Content panel of the right sidebar, confirm you have a Collection Categories filter along with other options. Collection Categories and Exclude Collection Categories should appear below Collection Sections and Exclude Collection Sections, respectively:

CleanShot 2025-08-05 at 17 38 53

5. Under Post Types, select Collections and deselect other post types. Confirm that the block is pulling in your Collections.
6. Under Collection Categories, try filtering the Collections in the block by one or two of your categories. Confirm the blocks are correct in the editor and on the front-end.
7. Repeat the above for the Exclude Collection Categories field, and confirm it removes Collections that are associated with the excluded category.
8. Repeat steps 4-7 with the Content Carousel block.
9. Disable the Collections module under WP Admin > Newspack > Settings.
10. Confirm that the Collection Categories query option is no longer available under the Content Loop and Carousel block settings. Any blocks set to return posts based on Collection should be empty.

### Other information:

### DIFF
--- a/includes/collections/class-collection-category-taxonomy.php
+++ b/includes/collections/class-collection-category-taxonomy.php
@@ -223,7 +223,7 @@ class Collection_Category_Taxonomy {
 		];
 
 		$point_of_insertion = null;
-		// Loop through the taxonomies; confirm Collections doesn't already exist, and grab the index of Collection Sections if it's there.
+		// Loop through the taxonomies; confirm Collection Categories doesn't already exist, and grab the index of Collection Sections if it's there.
 		foreach ( $custom_taxonomies as $index => $tax ) {
 			if ( isset( $tax['slug'] ) ) {
 				if ( $tax['slug'] === $collections_category_taxonomy['slug'] ) {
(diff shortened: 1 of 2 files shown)


rbcorrales (changes requested):
Were you able to test the filtering by categories? It's not working for me.

The thing is that the section and collection taxonomies are associated with the post itself, while the category is a taxonomy for the collection, so I guess some special handling is needed on the content loop side.

laurelfulford:
Thanks @rbcorrales! I think I need to rethink this one from a UX perspective -- the intention was to filter the Collection CPT itself by Collection Category for stuff like CTAs ("See all Summer Guides!" then a list of covers for each matching Collection) but that goes counter to how this block works otherwise in a big way. There's no other case where we filter anything but posts with the taxonomy filters, how you expected it to work makes a lot more sense than how it actually does 😕 

I'm going to close this one and go back to the drawing board.

PR #4134, opened 2025-08-08
fix: pattern categories count and add filter

leogermani:
### All Submissions:

The "count" column in the Pattern Categories screen was incorrectly linking to the the `post` Post type list, making it impossible to list patterns of a certain category. This PR

* Replaces the Count column with one with the right link
* Add a filter by category in the Patterns list

Refs NPPM-2159

1. Visit the Pattern categories page and check that the link in the counts are correct
2. Test the new filter in the Patterns screen

### Other information:

### DIFF
--- a/includes/class-patches.php
+++ b/includes/class-patches.php
@@ -191,7 +191,7 @@ class Patches {
 			]
 		);
 
-		if ( empty( $terms ) ) {
+		if ( empty( $terms ) || is_wp_error( $terms ) ) {
 			return;
 		}
 
--- a/includes/class-patches.php
+++ b/includes/class-patches.php
@@ -24,6 +24,10 @@ class Patches {
 		add_action( 'manage_edit-wp_block_columns', [ __CLASS__, 'add_custom_columns' ] );
 		add_action( 'manage_edit-wp_block_sortable_columns', [ __CLASS__, 'add_sortable_columns' ] );
 		add_action( 'manage_wp_block_posts_custom_column', [ __CLASS__, 'custom_column_content' ], 10, 2 );
+		add_action( 'restrict_manage_posts', [ __CLASS__, 'add_pattern_category_filter' ] );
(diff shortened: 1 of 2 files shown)


includes/class-patches.php
```diff
@@ -168,6 +172,73 @@ public static function custom_column_content( $column, $post_id ) {
...
+
+		$taxonomy = 'wp_pattern_category';
+		$terms    = get_terms(
+			[
+				'taxonomy'   => $taxonomy,
+				'hide_empty' => true,
+			]
+		);
+
+		if ( empty( $terms ) ) {
```
rbcorrales: Consider validating for errors here, as that's a possible return value for [get_terms()](https://developer.wordpress.org/reference/functions/get_terms/):
```php
if ( empty( $terms ) || is_wp_error( $terms ) ) {
```

includes/class-patches.php
```diff
@@ -168,6 +172,73 @@ public static function custom_column_content( $column, $post_id ) {
...
+		?>
+		<select name="<?php echo esc_attr( $taxonomy ); ?>">
+			<option value=""><?php esc_html_e( 'All Pattern Categories', 'newspack-plugin' ); ?></option>
+			<?php foreach ( $terms as $term ) : ?>
+				<option value="<?php echo esc_attr( $term->slug ); ?>" <?php selected( $current_term, $term->slug ); ?>>
+					<?php echo esc_html( $term->name ); ?>
+				</option>
+			<?php endforeach; ?>
+		</select>
+		<?php
```
rbcorrales: Consider adding a label for this select for screen readers. The date filter next to it has this, for example:
```html
<label for="filter-by-date" class="screen-reader-text">Filter by date</label>
```

rbcorrales (approved):
Approving, but please consider these comments.

includes/class-patches.php
leogermani: good point!

PR #4140, opened 2025-08-13
feat: add flag to disable sitekit check

leogermani:
### All Submissions:

This does two things

* Add a constant to allow us to disable this check on sites
* Do not run the check if that Manager method does not exist (and not only if it exists and return false)

1. Check the code

### Other information:


adekbadek (commented):
> Do not run the check if that Manager method does not exist 

What's the reason behind this change? This should not be of any difference to use, and a self-hosted site might benefit from having this check.

adekbadek (approved):

PR #4151, opened 2025-08-20
feat: add guardrails to distributed bylines

leogermani:
### All Submissions:

Refs NPPM-2176

This adds some guardrails for bylines distribution.

When replacing IDs in the shortcode, we used to leave IDs that were not informed in the mapping untouched. The problem with that is if it happen and there is a user in the site with that ID, when the plugin renders the byline, it will fetch the wrong author name from the DB.

This PR will remove the shortcode from the byline in case the ID doesn't map to an existing user in the site.

Take a look at the tests as they are pretty extensive.

This is a little hard to test because it handles a broken payload that should never happen. But you can mimic the broken payload by manually making changes to the `newspack_network_distributed_post_meta` and making it return an empty array or an array with different user emails.

Once you've done so, distribute the post and confirm that the end result in the destination post is a byline with only names, without the shortcodes.

### Other information:

### DIFF
--- a/includes/bylines/class-bylines.php
+++ b/includes/bylines/class-bylines.php
@@ -360,6 +360,7 @@ class Bylines {
 
 	/**
 	 * After an incoming post is inserted, update the IDs with the IDs in the target site, based on the mapping that was sent.
+	 * If an author ID is not found in the mapping, remove the entire author shortcode and keep only the author name.
 	 *
 	 * @param int   $post_id   The post ID.
 	 * @param bool  $is_linked Whether the post is linked.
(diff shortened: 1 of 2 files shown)


includes/bylines/class-bylines.php
```diff
@@ -371,22 +372,42 @@ public static function newspack_network_incoming_post_inserted( $post_id, $is_li
...
-			$local_user = get_user_by( 'email', $author_email );
-			if ( ! $local_user ) {
-				continue;
-			}
-			$byline = str_replace( 'id=' . $author_id, 'id=' . $local_user->ID, $byline );
-		}
+		// Process all author shortcodes in one pass.
+		$byline = preg_replace_callback(
+			'/\[Author id=(\d+)\](.*?)\[\/Author\]/',
+			function( $matches ) use ( $mapping ) {
```
miguelpeixe, line 390: Non-blocking: maybe too specific, but this looks like a nice utility method for the class. `transform_byline_author( $string, $callable( $author_string, $author_id, $author_name ) );`

miguelpeixe (approved):

PR #4164, opened 2025-09-01
feat: subscription tier modal

miguelpeixe:
### All Submissions:

Introduces subscription tiers form and modal for "Grouped Products" using the Newspack UI.

## Grouped Product

image

## Variable Subscription

image

Further details and instructions at https://github.com/Automattic/newspack-blocks/pull/2201

### Other information:


includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -63,5 +64,22 @@ public static function is_enabled() {
...
+
+	/**
+	 * Get the label for a frequency.
+	 *
+	 * @param string $frequency Frequency.
+	 *
+	 * @return string
+	 */
+	public static function get_frequency_label( $frequency ) {
+		$frequencies = [
```
dkoo, line 76: Should we make this filterable? I can foresee publishers wanting to add their own frequency labels like "Semi-Annual" or "Quarterly" etc.

dkoo (approved):
Looking good! Non-blocking, but `get_tiers_by_frequency` could be a good subject for unit testing.

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
miguelpeixe: Not sure if a filter here is the right way to support this. With the current approach, custom intervals will be rendered inside the saved frequency "unit":

image

The quarterly option should either live inside its own tab or stay in the Monthly tab. IMO, changing the tab to "Quarterly" via a filter would be worse for the existing "Monthly" option.

Should we add the interval as a frequency argument to create these common billing frequencies? Only Quarterly and Bi-Weekly come to mind. Not sure what Semi-Annual means. cc @thomasguillot

miguelpeixe:
> get_tiers_by_frequency could be a good subject for unit testing.

It requires so much Woo mocking that I have little confidence in its efficacy...

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
miguelpeixe: Added support for Bi-Weekly, Quarterly, and Semi-Annually common intervals in 4715a517665b91d715d4e3b13b7ba189cc111104

image

The frequency now comes as `{period}_{interval}`, so the new `newspack_subscriptions_frequency_label` filter can be used to handle uncommon frequencies. E.g. `month_2` = Bi-Monthly? 🤷

dkoo: I think this works for now—it's an edge case for sure, and frequencies/intervals beyond what you included here are probably quite rare. Thanks for the changes!

dkoo (approved):
> It requires so much Woo mocking that I have little confidence in its efficacy...

Good point!

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
thomasguillot: I'm a bit late here but this tabbed navigation doesn't look good now. Is there a way if it's more than 1 row, we change to a select?

miguelpeixe: I hope this is not a common offering! 😄
I can tackle it in #4177

miguelpeixe: Done in https://github.com/Automattic/newspack-plugin/pull/4177#issuecomment-3303802734

PR #4172, opened 2025-09-08
test(collections): add unit tests for collections block

rbcorrales:
### All Submissions:

This PR adds unit tests for the new Collections block and other classes changed.

Closes NPPD-797.

1. Run the tests via `n test-php`

### Other information:

### DIFF
--- a/tests/unit-tests/collections/class-test-collection-meta.php
+++ b/tests/unit-tests/collections/class-test-collection-meta.php
@@ -196,4 +196,70 @@ class Test_Collection_Meta extends WP_UnitTestCase {
 		$this->assertEquals( [], Collection_Meta::sanitize_ctas( 123 ), 'Should return empty array for integer input' );
 		$this->assertEquals( [], Collection_Meta::sanitize_ctas( null ), 'Should return empty array for null input' );
 	}
+
+	/**
+	 * Test register_rest_fields method.
+	 *
+	 * @covers \Newspack\Collections\Collection_Meta::register_rest_fields
+	 */
+	public function test_register_rest_fields() {
+		// Initialize REST API.
+		Collection_Meta::register_rest_fields();
+
+		$post_type   = Post_Type::get_post_type();
+		$rest_fields = $GLOBALS['wp_rest_additional_fields'][ $post_type ] ?? [];
+
+		$this->assertArrayHasKey( 'ctas', $rest_fields, 'CTA field should be registered.' );
+		$this->assertEquals( [ Collection_Meta::class, 'get_collection_ctas_for_rest' ], $rest_fields['ctas']['get_callback'], 'Should have correct get callback.' );
+	}
+
+	/**
+	 * Test get_collection_ctas_for_rest method.
+	 *
+	 * @covers \Newspack\Collections\Collection_Meta::get_collection_ctas_for_rest
+	 */
+	public function test_get_collection_ctas_for_rest() {
+		$collection_id = $this->create_test_collection();
+
+		// Mock CTAs data.
+		$ctas_data = [
+			[
+				'type'  => 'link',
+				'label' => 'Subscribe',
+				'url'   => 'https://example.com/subscribe',
+			],
+			[
+				'type'  => 'link',
+				'label' => 'Order',
+				'url'   => 'https://example.com/order',
+			],
+		];
+
+		Collection_Meta::set( $collection_id, 'ctas', $ctas_data );
+
+		$post_data = [ 'id' => $collection_id ];
+		$result    = Collection_Meta::get_collection_ctas_for_rest( $post_data );
+
+		$this->assertIsArray( $result, 'Result should be an array.' );
+		$this->assertCount( 2, $result, 'Should return 2 CTAs.' );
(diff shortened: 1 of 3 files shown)


tests/unit-tests/collections/class-test-collections-block.php
```diff
@@ -0,0 +1,561 @@
...
+		];
+
+		ob_start();
+		Collections_Block::render_collection_ctas( $collection, $attributes );
+		$output = ob_get_clean();
+
+		$this->assertStringContainsString( 'wp-block-newspack-collections__ctas', $output, 'Should contain CTAs wrapper' );
+		$this->assertStringContainsString( 'Click Here', $output, 'Should contain Click Here CTA' );
+		$this->assertStringContainsString( 'Download', $output, 'Should contain Download CTA' );
+	}
```
dkoo: I'm not sure what this one is testing? It sounds like it should be applying some filter as part of the test, but it looks like it's just testing that CTAs are rendered when present in block attributes?

tests/unit-tests/collections/class-test-collections-block.php
```diff
@@ -0,0 +1,561 @@
...
+			'columns'             => 0,
+			'numberOfCTAs'        => -1,
+			'selectedCollections' => [ 'invalid', '123', $collection_id ],
+		];
+
+		$output = $this->render_collections_block( $attributes );
+
+		// Should still render successfully due to sanitization.
+		$this->assertStringContainsString( 'wp-block-newspack-collections', $output, 'Should contain block wrapper despite invalid attributes' );
+		$this->assertStringContainsString( $post_title, $output, 'Should contain collection title' );
```
dkoo, line 504: Could we also test that the attributes with invalid values revert to default values?

tests/unit-tests/collections/class-test-collection-meta.php
```diff
@@ -196,4 +196,70 @@ public function test_sanitize_ctas_invalid_data() {
...
+				'type'  => 'link',
+				'label' => 'Subscribe',
+				'url'   => 'https://example.com/subscribe',
+			],
+			[
+				'type'  => 'link',
+				'label' => 'Order',
+				'url'   => 'https://example.com/order',
+			],
+		];
```
dkoo, line 236: Nit: is it worth validating/testing for invalid URLs? I see that the editor JS validates input before saving as meta, but should we also validate on the back-end in case of imported or altered data in the DB?

dkoo (approved):
Pretty comprehensive! I saw a couple of tests that could either be misnamed or don't go far enough to test what they say they test IMO. Feel free to merge after addressing those items.

tests/unit-tests/collections/class-test-collections-block.php
rbcorrales: Good observation. Test improved and renamed in [bdf776b](https://github.com/Automattic/newspack-plugin/pull/4172/commits/bdf776b355b4ca647a4ee8b2f3ec9bfeee0282dc). The idea is now to actually test the block toggles for the hierarchical Subscription and Order CTAs, which are separate from the others.

tests/unit-tests/collections/class-test-collections-block.php
rbcorrales: I realized we're not setting the defaults after sanitization, but a different value.
For example, the default for `numberOfItems` is 4, but in the validations [here](https://github.com/Automattic/newspack-plugin/blob/test/collections-block/src/blocks/collections/class-collections-block.php#L96-L102) if I pass an `'invalid'` value, it will be set to `1` instead of `4`.

It would be a good idea to set it to the defaults if it's invalid. Since the code structure that handles that has already changed with subsequent requests, I'll address this in a follow-up PR.

tests/unit-tests/collections/class-test-collection-meta.php
rbcorrales: Thanks for noting this! It led me to notice that we're not actually escaping the CTA values when exposed through the REST endpoint, since we're passing back the return from `Query_Helper::get_ctas()` directly:
https://github.com/Automattic/newspack-plugin/blob/test/collections-block/includes/collections/class-collection-meta.php#L248

Meanwhile, when rendering the markup via PHP, the values are actually escaped:
https://github.com/Automattic/newspack-plugin/blob/test/collections-block/includes/collections/class-template-helper.php#L321-L323

I will also address this in a follow-up PR.

tests/unit-tests/collections/class-test-collections-block.php
dkoo: Agreed, I think in case of invalid values it should revert to defaults instead of trying to interpret the invalid value.

tests/unit-tests/collections/class-test-collections-block.php
dkoo: Thanks for the clarification and test improvements!

dkoo (approved):
Approved! Noting that we'll tackle a couple of items in a separate PR:

- Handle invalid attribute values by falling back to defaults
- Ensure that CTA values (and all block attribute values) are escaped when exposed via REST

PR #4175, opened 2025-09-09
fix: register with empty name fields

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPM-2198/name-fields-of-checkout-modal-not-saving-in-wp-account

This PR fixes an issue where we were incorrectly setting a readers first name to the default nicename generated from the readers email on registration. As a result, Woo would not update the first name during checkout since Woo skips this step in favor of any name data that is already set.

Screenshot 2025-09-09 at 13 57 32

Screenshot 2025-09-09 at 13 57 17

1. Register as a new reader on the site (with Audience Management active and onboarded)
2. As admin, verify the new reader has an empty first and last name set via the User edit page in WP Admin
3. Again as the reader, go through modal checkout for any product or donation
4. Again as admin, verify the new reader name fields have been populated with the name fields entered during checkout

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -2255,7 +2255,7 @@ final class Reader_Activation {
 				]
 			);
 
-			// Temporarily remove our canonize_user_data callback from the woocommercer hook.
+			// Unhook from WooCommerce as it's already been canonized above.
 			remove_filter( 'woocommerce_new_customer_data', [ __CLASS__, 'canonize_user_data' ], 10, 1 );
 			if ( function_exists( '\wc_create_new_customer' ) ) {
 				/**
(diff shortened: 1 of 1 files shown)


includes/reader-activation/class-reader-activation.php
```diff
@@ -2255,6 +2255,8 @@ public static function register_reader( $email, $display_name = '', $authenticat
 				]
 			);
 
+			// Temporarily remove our canonize_user_data callback from the woocommercer hook.
```
miguelpeixe:
```suggestion
			// Unhook from WooCommerce as it's already been canonized above
```

miguelpeixe (approved):
Fixes the issue! Small nit suggestion for the comment

PR #4176, opened 2025-09-10
feat(content-gate): add countdown block

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPM-2204/content-restriction-block-shows-countdown-before-you-hit-the-wall-you

This PR adds a new content gate countdown block that displays the remaining free metered views for a post or page:

Screenshot 2025-09-15 at 19 45 53

1. First create some memberships and have some readers that are members and non-members
2. Create some content gates for these memberships, making sure each has metering active
3. Add the content gate countdown block to several of the restricted posts/pages in which the block has been added
4. Test that the block
  - appears on the frontend when within the metered view limit
  - countsdown as the reader views more relevant content
  - appears on the frontend for posts that have already been viewed when the limit is reached
  - does not appear on the frontend for newly viewed posts whenever the limit is reached (gate should be displayed instead)
  - does not appear for members who have access
  - Be sure to also test the for logged in members, logged in non-members, and logged out readers

### Other information:

### DIFF
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -70,7 +70,7 @@ final class Blocks {
 			$script_data['content_gate_data'] = [
 				'anonymous_metered_views' => Metering::get_total_metered_views( false ),
 				'loggedin_metered_views'  => Metering::get_total_metered_views( true ),
-				'remaining_metered_views' => Metering::get_remaining_metered_views(),
+				'metered_views'           => Metering::get_metered_views(),
 				'metering_period'         => Metering::get_metering_period(),
 			];
 		}
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -7,11 +7,11 @@
 
 namespace Newspack;
 
-use Newspack\Memberships;
-
 defined( 'ABSPATH' ) || exit;
 
 use Newspack\Optional_Modules\Collections;
+use Newspack\Memberships;
+use Newspack\Memberships\Metering;
 
 /**
  * Newspack Blocks Class.
@@ -52,23 +52,32 @@ final class Blocks {
 			NEWSPACK_PLUGIN_VERSION,
 			true
 		);
+		$script_data = [
+			'has_newsletters'         => class_exists( 'Newspack_Newsletters_Subscription' ),
+			'has_reader_activation'   => Reader_Activation::is_enabled(),
+			'newsletters_url'         => Wizards::get_wizard( 'newsletters' )->newsletters_settings_url(),
+			'has_google_oauth'        => Google_OAuth::is_oauth_configured(),
+			'google_logo_svg'         => \Newspack\Newspack_UI_Icons::get_svg( 'google' ),
+			'reader_activation_terms' => Reader_Activation::get_setting( 'terms_text' ),
+			'reader_activation_url'   => Reader_Activation::get_setting( 'terms_url' ),
+			'has_recaptcha'           => Recaptcha::can_use_captcha(),
+			'recaptcha_url'           => admin_url( 'admin.php?page=newspack-settings' ),
+			'corrections_enabled'     => wp_is_block_theme() && class_exists( 'Newspack\Corrections' ),
+			'collections_enabled'     => Collections::is_module_active(),
+			'has_content_gate'        => Memberships::has_gate(),
+		];
+		if ( Memberships::is_active() ) {
+			$script_data['content_gate_data'] = [
+				'anonymous_metered_views' => Metering::get_total_metered_views( false ),
+				'loggedin_metered_views'  => Metering::get_total_metered_views( true ),
+				'remaining_metered_views' => Metering::get_remaining_metered_views(),
+				'metering_period'         => Metering::get_metering_period(),
+			];
+		}
 		\wp_localize_script(
 			'newspack-blocks',
 			'newspack_blocks',
-			[
-				'has_newsletters'         => class_exists( 'Newspack_Newsletters_Subscription' ),
-				'has_reader_activation'   => Reader_Activation::is_enabled(),
-				'newsletters_url'         => Wizards::get_wizard( 'newsletters' )->newsletters_settings_url(),
-				'has_google_oauth'        => Google_OAuth::is_oauth_configured(),
-				'google_logo_svg'         => \Newspack\Newspack_UI_Icons::get_svg( 'google' ),
-				'reader_activation_terms' => Reader_Activation::get_setting( 'terms_text' ),
-				'reader_activation_url'   => Reader_Activation::get_setting( 'terms_url' ),
-				'has_recaptcha'           => Recaptcha::can_use_captcha(),
-				'recaptcha_url'           => admin_url( 'admin.php?page=newspack-settings' ),
-				'corrections_enabled'     => wp_is_block_theme() && class_exists( 'Newspack\Corrections' ),
-				'collections_enabled'     => Collections::is_module_active(),
-				'has_content_gate'        => Memberships::is_active() && Memberships::has_gate(),
-			]
+			$script_data
 		);
 		\wp_enqueue_style(
 			'newspack-blocks',
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -7,6 +7,8 @@
 
 namespace Newspack;
 
+use Newspack\Memberships;
+
 defined( 'ABSPATH' ) || exit;
 
 use Newspack\Optional_Modules\Collections;
@@ -20,6 +22,7 @@ final class Blocks {
 	 */
 	public static function init() {
 		require_once NEWSPACK_ABSPATH . 'src/blocks/reader-registration/index.php';
+		require_once NEWSPACK_ABSPATH . 'src/blocks/content-gate-countdown/class-content-gate-countdown-block.php';
 
 		if ( wp_is_block_theme() && class_exists( 'Newspack\Corrections' ) ) {
 			require_once NEWSPACK_ABSPATH . 'src/blocks/correction-box/class-correction-box-block.php';
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -67,6 +67,7 @@ final class Blocks {
 				'recaptcha_url'           => admin_url( 'admin.php?page=newspack-settings' ),
 				'corrections_enabled'     => wp_is_block_theme() && class_exists( 'Newspack\Corrections' ),
 				'collections_enabled'     => Collections::is_module_active(),
+				'has_content_gate'        => Memberships::is_active() && Memberships::has_gate(),
 			]
 		);
 		\wp_enqueue_style(
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -64,9 +64,9 @@ final class Blocks {
 			'recaptcha_url'           => admin_url( 'admin.php?page=newspack-settings' ),
 			'corrections_enabled'     => wp_is_block_theme() && class_exists( 'Newspack\Corrections' ),
 			'collections_enabled'     => Collections::is_module_active(),
-			'has_content_gate'        => Memberships::has_gate(),
+			'has_memberships'         => Memberships::is_active(),
 		];
-		if ( Memberships::is_active() ) {
+		if ( $script_data['has_memberships'] ) {
 			$script_data['content_gate_data'] = [
 				'anonymous_metered_views' => Metering::get_total_metered_views( false ),
 				'loggedin_metered_views'  => Metering::get_total_metered_views( true ),
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -691,7 +691,7 @@ class Memberships {
 			/**
 			 * Filter the list of blocks to exclude from the excerpt.
 			 *
-			 * @param array $excluded_blocks Array of block names to exclude. i.e. [ 'core/image', 'newspack/content-gate-countdown' ].
+			 * @param array $excluded_blocks Array of blocks to exclude. i.e. [ 'core/image', 'newspack/content-gate-countdown' ].
 			 *
 			 * @return array
 			 */
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -686,6 +686,29 @@ class Memberships {
 			$content = explode( '<!--more-->', $content )[0];
 		} else {
 			$count = (int) get_post_meta( $gate_post_id, 'visible_paragraphs', true );
+			// Remove all spaces.
+			$content = preg_replace( '/\s+/', ' ', $content );
+			/**
+			 * Filter the list of blocks to exclude from the excerpt.
+			 *
+			 * @param array $excluded_blocks Array of block names to exclude. i.e. [ 'core/image', 'newspack/content-gate-countdown' ].
+			 *
+			 * @return array
+			 */
+			$excluded_blocks = apply_filters( 'newspack_memberships_excerpt_excluded_blocks', [ 'newspack/content-gate-countdown' ] );
+			// Remove unwanted blocks from the content.
+			foreach ( $excluded_blocks as $block ) {
+				[ $category, $name ] = explode( '/', $block );
+				if ( ! $category || ! $name ) {
+					continue;
+				}
+				if ( 'core' === $category ) {
+					$regex = $name;
+				} else {
+					$regex = "$category\/$name";
+				}
+				$content = preg_replace( "/<!-- wp:$regex {?.*?}? -->.*?<!-- \/wp:$regex -->/s", '', $content );
+			}
 			// Split into paragraphs.
 			$content = explode( '</p>', $content );
 			// Extract the first $x paragraphs only.
--- a/src/blocks/content-gate-countdown/block.json
+++ b/src/blocks/content-gate-countdown/block.json
@@ -10,6 +10,7 @@
     }
   },
   "supports": {
+		"align": [ "wide", "full" ],
     "html": false,
     "color": {
       "link": true,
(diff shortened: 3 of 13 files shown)


src/blocks/content-gate-countdown/edit.jsx
```diff
@@ -0,0 +1,56 @@
...
+										align: 'center',
+										style: { typography: { fontWeight: '700' } },
+									},
+								],
+								[
+									'newspack-blocks/checkout-button',
+									{
+										text: __( 'Subscribe now', 'newspack-plugin' ),
+										align: 'center',
+										layout: { type: 'flex', justifyContent: 'center' },
```
chickenn00dle: Note, the checkout button does not appear center aligned when adding the block in the editor. This is a bug and is being fixed here: https://github.com/Automattic/newspack-blocks/pull/2207

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -686,6 +686,29 @@ public static function get_restricted_post_excerpt( $post ) {
...
+				if ( ! $category || ! $name ) {
+					continue;
+				}
+				if ( 'core' === $category ) {
+					$regex = $name;
+				} else {
+					$regex = "$category\/$name";
+				}
+				$content = preg_replace( "/<!-- wp:$regex {?.*?}? -->.*?<!-- \/wp:$regex -->/s", '', $content );
+			}
```
chickenn00dle, line 711: If we don't filter this, any inner paragraph blocks within the content gate block will appear in the excerpt despite us explicitly returning an empty string in the render callback.

src/blocks/content-gate-countdown/edit.jsx
```diff
@@ -0,0 +1,108 @@
...
+						align: 'center',
+						content: __( 'Get unlimited access.', 'newspack-plugin' ),
+						style: { typography: { fontWeight: '700' } },
+					},
+				],
+				[
+					'newspack-blocks/checkout-button',
+					{
+						text: __( 'Subscribe now', 'newspack-plugin' ),
+						align: 'center',
```
chickenn00dle, line 36: Note, alignment is broken in the editor right now and will be fixed in https://github.com/Automattic/newspack-blocks/pull/2207

src/blocks/content-gate-countdown/view.js
```diff
@@ -0,0 +1,39 @@
...
+		return;
+	}
+	const { count, gate_id } = newspack_metering_settings;
+	window.newspackRAS = window.newspackRAS || [];
+	window.newspackRAS.push( ras => {
+		const { authenticated } = ras?.getReader() || { authenticated: false };
+		if ( authenticated ) {
+			return;
+		}
+		const storeKey = 'metering-' + gate_id || 0;
```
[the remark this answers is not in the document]
chickenn00dle: Fixed in https://github.com/Automattic/newspack-plugin/pull/4176/commits/d116d5625454735dbf6c024e91b43626c93b6705

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -686,6 +686,29 @@ public static function get_restricted_post_excerpt( $post ) {
...
+			 *
+			 * @return array
+			 */
+			$excluded_blocks = apply_filters( 'newspack_memberships_excerpt_excluded_blocks', [ 'newspack/content-gate-countdown' ] );
+			// Remove unwanted blocks from the content.
+			foreach ( $excluded_blocks as $block ) {
+				[ $category, $name ] = explode( '/', $block );
+				if ( ! $category || ! $name ) {
+					continue;
+				}
```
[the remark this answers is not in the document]
chickenn00dle: Going to ignore since we are declaring the blocknames above ourselves and the filter gives examples that include category and name for each block.

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
```diff
@@ -0,0 +1,115 @@
...
+		);
+		$block_content = "<div $block_wrapper_attributes>
+			<div class='newspack-content-gate-countdown__content'>
+				<div class='newspack-content-gate-countdown__text'>
+					<span class='newspack-content-gate-countdown__countdown'>$countdown</span>
+					<p>$text</p>
+				</div>
+				$content
+			</div>
+		</div>";
```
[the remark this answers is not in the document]
chickenn00dle: Fixed in https://github.com/Automattic/newspack-plugin/pull/4176/commits/d116d5625454735dbf6c024e91b43626c93b6705

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -686,6 +686,29 @@ public static function get_restricted_post_excerpt( $post ) {
...
+				if ( 'core' === $category ) {
+					$regex = $name;
+				} else {
+					$regex = "$category\/$name";
+				}
+				$content = preg_replace( "/<!-- wp:$regex {?.*?}? -->.*?<!-- \/wp:$regex -->/s", '', $content );
```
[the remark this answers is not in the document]
chickenn00dle: Again, ignoring because we are declaring the blocks above and these are not from user input.

src/blocks/content-gate-countdown/edit.jsx
```diff
@@ -0,0 +1,108 @@
...
+	if ( ! totalViews ) {
+		return (
+			<Placeholder
+				icon={ caution }
+				label={ __(
+					'The content gate countdown block will only display in restricted content when metering is enabled.',
+					'newspack-plugin'
+				) }
+				className="no-metering"
+			/>
```
miguelpeixe: This should also have `blockProps`, otherwise the element won't interact well with the editor.

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
```diff
@@ -0,0 +1,115 @@
...
+ * Content Gate Countdown Block
+ *
+ * @package Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+use Newspack;
+use Newspack\Memberships;
+use Newspack\Memberships\Metering;
```
miguelpeixe: Can we namespace this to `Newspack` instead?

includes/class-blocks.php
```diff
@@ -10,6 +10,8 @@
 defined( 'ABSPATH' ) || exit;
 
 use Newspack\Optional_Modules\Collections;
+use Newspack\Memberships;
+use Newspack\Memberships\Metering;
```
miguelpeixe, line 14: This is superfluous as we're already in the `Newspack` namespace.

We can keep `Newspack\Memberships\Metering`, but I don't think it's bad to use `Memberships\Metering` in the code.

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param array  $attributes The block attributes.
+	 * @param string $content    The block content.
+	 *
+	 * @return string The block HTML.
+	 */
+	public static function render_block( array $attributes, string $content ) {
+		if ( ! Metering::is_metering() ) {
+			return '';
+		}
```
miguelpeixe: I think we also need to bail on `! Memberships::is_post_restricted()`, otherwise the block will render for readers with access.

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
```diff
@@ -0,0 +1,115 @@
...
+		$total_views = Metering::get_total_metered_views( \is_user_logged_in() );
+		if ( false === $total_views ) {
+			return '';
+		}
+		$remaining_views = Metering::get_remaining_metered_views( get_current_user_id() );
+		$countdown       = sprintf(
+			/* translators: 1: remaining metered views, 2: total metered views. */
+			__( '%1$d/%2$d', 'newspack-plugin' ),
+			max( 0, $total_views - $remaining_views ),
```
miguelpeixe: The amount was already calculated [here](https://github.com/Automattic/newspack-plugin/blob/d116d5625454735dbf6c024e91b43626c93b6705/includes/plugins/wc-memberships/class-metering.php#L343). It's going full circle and returning the number of articles read 😄 

Speaking of which, even though I'm aware that the block is called countdown, reading 4/5 on my first visit was a bit odd. Not sure if it's because of the way it's phrased, maybe it's just me. @thomasguillot can you confirm whether it should be the number of articles read or remaining?

src/blocks/content-gate-countdown/block.json
```diff
@@ -0,0 +1,38 @@
...
+  "apiVersion": 3,
+  "name": "newspack/content-gate-countdown",
+  "category": "newspack",
+  "attributes": {
+    "text": {
+      "type": "string",
+      "default": ""
+    }
+  },
+  "supports": {
```
miguelpeixe, line 12: Looking at the design, it seems that the block should also support `align`, so it can also render wide and full width.

src/blocks/content-gate-countdown/edit.jsx
```diff
@@ -0,0 +1,108 @@
...
+				[
+					'core/paragraph',
+					{
+						align: 'center',
+						content: __( 'Get unlimited access.', 'newspack-plugin' ),
+						style: { typography: { fontWeight: '700' } },
+					},
+				],
+				[
+					'newspack-blocks/checkout-button',
```
miguelpeixe, line 33: WDYT of adding `"backgroundColor":"primary"` to the template as well?

src/blocks/content-gate-countdown/edit.jsx
```diff
@@ -0,0 +1,108 @@
...
+		attributes.text
+			? attributes.text
+			: sprintf(
+					/* translators: %s is the metered period, e.g. "month" or "week". */
+					__( 'free articles this %s', 'newspack-plugin' ),
+					meteringPeriod
+			  )
+	);
+	// Admin is always logged in, so if no loggedin metered views are set, use the anonymous views instead.
+	const totalViews = loggedinViews > 0 ? loggedinViews : anonymousViews;
```
miguelpeixe, line 60: This is a bit confusing. Isn't this about whether the post gate has metering configured?

Something like:

```suggestion
	const hasMetering = loggedinViews || anonymousViews;
```

Still, this does not take into account whether metering is active (the `metering` meta)

miguelpeixe (changes requested):
It's working mostly well! I encountered a few issues and suggestions that I commented inline.

src/blocks/content-gate-countdown/block.json
chickenn00dle: Good catch! Added in https://github.com/Automattic/newspack-plugin/pull/4176/commits/c9ed5ab758841e481deb5b7db6d801b28c3244b2

src/blocks/content-gate-countdown/edit.jsx
chickenn00dle: Added in https://github.com/Automattic/newspack-plugin/pull/4176/commits/c9ed5ab758841e481deb5b7db6d801b28c3244b2

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
chickenn00dle: Good call. Changed in https://github.com/Automattic/newspack-plugin/pull/4176/commits/c9ed5ab758841e481deb5b7db6d801b28c3244b2

includes/class-blocks.php
chickenn00dle: It feels cleaner to use just `Metering` in the code imo, so I'll keep it as is. Happy to change it if you feel strongly about it though!

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
chickenn00dle: Ah. Good catch! I had removed this part of the check while working on styles and forgot to re-add it before submitting for review. Fixed in https://github.com/Automattic/newspack-plugin/pull/4176/commits/c9ed5ab758841e481deb5b7db6d801b28c3244b2

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
chickenn00dle: > The amount was already calculated [here](https://github.com/Automattic/newspack-plugin/blob/d116d5625454735dbf6c024e91b43626c93b6705/includes/plugins/wc-memberships/class-metering.php#L343). It's going full circle and returning the number of articles read 😄

Ha. Yeah. I initially had this counting down using the remaining count. Then at the last second I thought it made more sense to show the number of articles read and made this change. I've gone back to remaining count here in https://github.com/Automattic/newspack-plugin/pull/4176/commits/c9ed5ab758841e481deb5b7db6d801b28c3244b2, but I'd like to get confirmation from @thomasguillot as well to know what this number should be.

src/blocks/content-gate-countdown/edit.jsx
chickenn00dle: Sure thing! This requires setting textColor to secondary to avoid contrast issues. Added in https://github.com/Automattic/newspack-plugin/pull/4176/commits/c9ed5ab758841e481deb5b7db6d801b28c3244b2

src/blocks/content-gate-countdown/edit.jsx
chickenn00dle: The problem this is solving is in the editor, the user (admin) is always logged in, so if the post is metered for anonymous readers and not registered readers, then the views will always be 0 in the editor.

I named this variable totalViews because below we use it when displaying the countdown: https://github.com/Automattic/newspack-plugin/pull/4176/files/d116d5625454735dbf6c024e91b43626c93b6705#diff-da1ae48178cfb1fb24cf269739fc3d32517200fb0105cefdba1155dba8c45f9eR98

As for using the `||` operator instead of the ternary, I thought the ternary made it a bit clearer that we are falling back to anonymous views if loggedin views was 0. But both will have the same result, so happy to change it if you prefer! 

> Still, this does not take into account whether metering is active (the metering meta)

Yeah. I decided not to check the meta here because having both loggedinViews and anonymousViews be 0 is effectively the same as not having metering active.

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
thomasguillot: It's meant to be 4 articles read out of 5. So the first visit should be 1/5

thomasguillot: So you have read 1 article. And you have 5 free.

thomasguillot: And not read 4 articles

chickenn00dle: Thanks @thomasguillot! Updated this to count up in https://github.com/Automattic/newspack-plugin/pull/4176/commits/68df86c5557ebf15a21fb20e40299ead3a9a3972

src/blocks/content-gate-countdown/view.js
```diff
@@ -0,0 +1,39 @@
...
+		if ( ! countdownEl ) {
+			return;
+		}
+		// Replace countdown for anonymous users.
+		const countdown = sprintf(
+			/* translators: 1: current number of metered views, 2: total metered views. */ __( '%1$d/%2$d', 'newspack-plugin' ),
+			content.length,
+			count
+		);
+		countdownEl.textContent = countdown;
```
miguelpeixe, line 37: When accessing an article for the first time with anonymous metering, I'm getting 0 reads:

Image

We may need to listen to the [`metering_restricted` activity dispatch](https://github.com/Automattic/newspack-plugin/blob/4e1c26226035af51e41752e1188d8f58140b7ff4/src/memberships-gate/metering.js#L80) for an accurate number here

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
```diff
@@ -0,0 +1,116 @@
...
+	 *
+	 * @return void
+	 */
+	public static function enqueue_scripts() {
+		if ( ! Memberships::is_active() || ! is_singular() ) {
+			return;
+		}
+		wp_enqueue_style(
+			'newspack-content-gate-countdown-block',
+			\Newspack\Newspack::plugin_url() . '/dist/content-gate-countdown-block.css',
```
miguelpeixe: Not sure why, but I'm getting 404 for this asset.

miguelpeixe (changes requested):
Thank you for the revisions! It's working great, but I found a couple of issues:

src/blocks/content-gate-countdown/view.js
chickenn00dle: Strange, on my end the metering script is always run before the countdown script, and the value is correctly updated:
Screenshot 2025-09-24 at 16 55 59

I checked in multiple browsers to confirm this isn't some browser loading order quirk. In any case, I added the metering script handle as a dependency when enqueuing the countdown script to ensure this is always the case in https://github.com/Automattic/newspack-plugin/pull/4176/commits/4202b80a14a6ece38f4baa1b59fea8ed4702e168

src/blocks/content-gate-countdown/class-content-gate-countdown-block.php
chickenn00dle: Looks like the styles for this are being bundled into commons by webpack. Not sure if this is what you are seeing. On my end, there is no `content-gate-countdown-block.css` being fetched. Instead the styles live in `commons.css`

miguelpeixe: It's getting added to the markup and throwing 404:

image

chickenn00dle: Ah sorry! Seeing it now. I guess we don't need to [enqueue the styles here](https://github.com/Automattic/newspack-plugin/pull/4176/commits/492e9316ad2c2c9619e697076a82a795616b613f) 🤷‍♂️

miguelpeixe (approved):

PR #4178, opened 2025-09-11
feat(collections): replace archive grid with collections block

rbcorrales:
### All Submissions:

This commit refactors the Collections archive template to use the Collections block for grid rendering instead of hardcoded HTML loops. The archive template now delegates grid rendering to the Collections block through a new `Template_Helper::render_collections_grid()` method and uses similar styling.

- The Collections block was enhanced to handle pre-queried collections using the existing `selectedCollections` block param. It's now able to handle mixed arrays of `WP_Post` objects or IDs.
- Duplicate grid styles were removed from the archive stylesheet since they're now handled by the block's existing styles.
- The block asset loading logic was also updated to properly enqueue styles on Collections archive pages. 
Closes NPPD-882.

1. Navigate to the Collections archive page and verify the grid layout displays correctly with proper responsive behavior across desktop, tablet, and mobile breakpoints.
2. Test the archive filtering functionality (year and category filters) to ensure the Collections block properly renders filtered results and maintains grid styling consistency.
3. Enable pagination on the archive page with many collections and verify that pagination works correctly while maintaining proper grid layout and block functionality.

### Other information:


dkoo (changes requested):
Codewise and functionally looks good, so approving with some nitpicks about the archive filter UI. I'd argue that this is a great candidate for an AJAX-based UI instead, but I understand if that adds too much scope at this point.

1. I'm not a fan of the way the page reloads on select `change`. 
I think we might want to add a button to "apply" the selected filters so you can at least more easily choose options in both year + category filters in one go. Maybe also a "reset" button if any filters are active. Maybe the filter JS can also ensure that the selected value always matches the active filters. cc @thomas in case you have thoughts. My reasons:
  - If I want to filter by both year and category, it requires two separate page refreshes. 
  - Likewise if I want to reset both filters.
  - If I hit "back" or "forward" in the browser, the dropdowns might be preset to what I last selected—that is, not visually matching the active filters. Example: I selected the "Special Edition" option. After those results loaded, I selected "Collection Category". After the next refresh, I hit "back" to go back to "Special Editions". The page correctly showed the "Special Edition" collections, but the dropdown had "Collection Category" selected already, which felt confusing. Also, if I want to actually filter by "Collection Category" again at this point I need to select a different option in the dropdown (triggering a page refresh) before I can select it again.

Screenshot 2025-09-11 at 1 54 30 PM

2. If you happen to select a combination of filters that has 0 results, the filter UI disappears so you can't fix it without going back to the prior page:

Screenshot 2025-09-11 at 1 42 48 PM

3. (non-blocking) At first I wasn't sure why I wasn't seeing the "2025" year option after I had selected 2024 from the list—I thought the filters should show all available options, not just options from the current results. In this case if I wanted to go back to viewing 2025 collections, I'd need to set the filter to "All" and then "2025" again (requiring 2 page refreshes), which seems unintuitive to me. Not sure we can solve this easily without more extensive refactoring but it's worth mentioning.

rbcorrales:
@dkoo Thanks for the thoughtful review! Yeah, I agree there's definitely room to improve the filter UI. A couple of clarifications on scope and constraints:
- This PR doesn't change the filters behavior or design itself (same for the other open PRs), but we can spin up a follow-up issue to capture these ideas and explore them properly.
- For this first iteration of Collections, one of the premises was to lean on existing WordPress functionailty as much as possible. So in this case, we relied on the existign server-side query/pagination rather than introducing custom AJAX handling, which is why the filters currently trigger a page reload. But having a client-side rendered version sounds like a good upgrade.
- The "apply" button is a good point. One detail is that the Collection Category (or “Publication”) filter only appears if there's more than one category. Some customers already using this feature only have a single category, so adding a required "apply" action in those cases might feel like unnecessary extra friction.

From a pattern perspective, it's worth stepping back to decide whether we want these filters to act as immediate controls (change on select) or as part of a grouped "filter form" (with apply/reset). That's probably a bigger design choice that affects the UX consistency, so I think it makes sense to track it separately and revisit with input from @thomasguillot.

rbcorrales:
Re-requesting review mainly because you said you were approving, but on a comment requesting changes 😄

dkoo:
> Re-requesting review mainly because you said you were approving, but on a comment requesting changes 😄

You can consider it approved on my end—my approval is for the code quality and functionality, which look 👍 to me. My feedback on the filter UI is mainly subjective, and I think we should defer to @thomasguillot. Let's wait until he can weigh in before merging.

thomasguillot:
@rbcorrales let's remove the button 👍 

Maybe we can document how a publisher can simply add a Button block and copy/paste the archive URL?

thomasguillot:
> From a pattern perspective, it's worth stepping back to decide whether we want these filters to act as immediate controls (change on select) or as part of a grouped "filter form" (with apply/reset). That's probably a bigger design choice that affects the UX consistency, so I think it makes sense to track it separately and revisit with input from @thomasguillot.

I agree. Let's ship this and then we'll revisit.

thomasguillot:
✅ Design approved (with the caveat of removing the "See all" button)

rbcorrales:
@dkoo Sorry, what I meant is that you said it's approved and manually added the `[Status] Approved` tag, but you just need to actually approve the PR on GitHub, because you submitted your comments through "request changes", so merge is still blocked:
image

@thomasguillot I think your last three comments are related to the conversation in #4166, since the conversation here was about a possible "Apply" button on the filters. In any case, I'll remove the "See all" button from the block. Thanks!

dkoo (approved):
Sorry!

PR #4184, opened 2025-09-12
Only defer newspack-plugin scripts if Audience Management is not enabled

claudiulodro:
### All Submissions:

Closes NPPM-2105

The Perfmatters "Delay" feature waits X seconds or until the user does an interaction to load the delayed scripts. On desktop, moving the mouse counts as an interaction, but on mobile, the first interaction would be when someone touches the screen.

So, on mobile, if the first things someone does is click the Sign In icon soon after the page renders, the scripts will start loading in at that time, and their click on the Sign In doesn't do anything useful. We can solve this by not delaying the newspack-plugin scripts if Audience Management is enabled. This has tradeoffs though, in the form of slightly worse performance. If you don't think this is an acceptable tradeoff, LMK and we can figure something out (maybe breaking off the Audience Management scripts into their own file)

1. Before applying this patch, have Audience Management on a site and Perfmatters with the default script handling.
2. Open an incognito window and turn on the mobile responsive simulation. Load the site and immediately after the page renders, click the Sign In icon. Observe nothing happens except a `#` is appended to the current URL.

Screenshot 2025-09-12 at 2 41 13 PM

3. Apply this patch. Try the same thing. Observe the Sign In modal appears.

### Other information:

### DIFF
--- a/includes/plugins/class-perfmatters.php
+++ b/includes/plugins/class-perfmatters.php
@@ -27,9 +27,8 @@ class Perfmatters {
 	 * Scripts to delay spec.
 	 */
 	private static function scripts_to_delay() {
-		return [
+		$scripts_to_delay = [
 			// Newspack.
-			'newspack-plugin',
 			'newspack-popups',
 			'newspack-blocks',
 			'newspack-newsletters',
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
The tradeoff is fine with me!

PR #4190, opened 2025-09-16
fix(bylines): display inline name for missing users

miguelpeixe:
### All Submissions:

Custom bylines are not aware of account deletion, so the `[Author id=00]` shortcode may refer to a non-existing author. In those cases, the inline name gets replaced with an empty string. This PR changes that behavior so the inline name is preserved.

1. While on release, create a post with multiple authors and a custom byline
2. Delete the user for one of the authors
3. Visit the post and confirm the byline is missing the author name
4. Checkout this branch, refresh the page, and confirm the byline preserves the author name but without a link

### Other information:

### DIFF
--- a/includes/bylines/class-bylines.php
+++ b/includes/bylines/class-bylines.php
@@ -206,6 +206,11 @@ class Bylines {
 			function( $matches ) {
 				$author_id = $matches[1];
 
+				$author = get_user_by( 'id', $author_id );
+				if ( ! $author ) {
+					return $matches[2];
+				}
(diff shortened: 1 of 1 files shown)


leogermani:
Funny to see this PR because that was the intended behavior since the beginning and that's the only reason the name is there, to serve as a callback... did it never work?

leogermani (approved):

leogermani:
There's a chance I've broke it on a recent refactor

PR #4191, opened 2025-09-16
feat(audience): primary subscription product

miguelpeixe:
### All Submissions:

Introduces a new URL that triggers subscription purchase or upgrade. To enable the URL, the site must have an assigned primary subscription product, which should be configured in the "Audience > Subscriptions" admin page.

Products eligible for selection are "Grouped Products" or "Variable Subscriptions". In the context of grouped products, only linked subscriptions or variable subscriptions will be used.

image

Purchasing:

image

Upgrading:

image

This is currently a hidden feature, but the query argument can also be modified to determine the modal title and submit button:

```
https://example.com/?upgrade-subscription=Upgrade%20your%20membership
```

The custom label will be applied for purchases and subscription switching:

image

1. As an admin, navigate to Audience -> Subscriptions
2. Confirm the page renders the new "Primary Subscription Tier Product" card
3. Open the select options and confirm all Grouped Products and Variable Subscriptions are listed (except donation products)
4. Select a product and refresh the page
5. Confirm the selection persists
6. Copy the URL from the notice and paste it into a new session
7. Confirm the modal renders prompting a purchase
8. Complete the purchase 
9. In the same session, visit the URL again
10. Confirm the modal renders prompting an upgrade
11. Complete the upgrade

### Other information:


includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -456,5 +606,53 @@ public static function order_button_text( $text ) {
...
+					|| $line_item['variation_id'] === $product_id
+					|| ( method_exists( $product, 'get_children' ) && in_array( $line_item['product_id'], $product->get_children() ) ) // In case it's a grouped product.
+				) {
+					$item = $line_item;
+					break;
+				}
+			}
+			if ( $item ) {
+				$switch_data = [
+					'item_id'      => $item['id'],
```
dkoo: I was having trouble with getting the checkout modal to open from the "Change subscription" version of the modal until I changed it to this:

```suggestion
					'item_id'      => $item->get_id(),
```

includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -456,5 +606,53 @@ public static function order_button_text( $text ) {
...
+		$switch_data       = null;
+
+		if ( $user_subscription ) {
+			$product_id = $product->get_id();
+			$item       = null;
+			foreach ( $user_subscription->get_items() as $line_item ) {
+				if (
+					$line_item['product_id'] === $product_id
+					|| $line_item['variation_id'] === $product_id
+					|| ( method_exists( $product, 'get_children' ) && in_array( $line_item['product_id'], $product->get_children() ) ) // In case it's a grouped product.
```
dkoo: Nit:
```suggestion
					|| ( method_exists( $product, 'get_children' ) && in_array( $line_item['product_id'], $product->get_children(), true ) ) // In case it's a grouped product.
```

dkoo (changes requested):
I ran into a couple of blockers that I was able to fix with some code changes:

1. At testing step 7, I was having trouble getting the checkout modal to open from the tiers modal while I was on a page with `?upgrade-subscription=1` appended to the URL. I traced the issue to `subscription-tiers-form.js`. Looks like if there's already a query param in the current page, the URL constructed for the checkout modal will contain two `?` characters and be invalid. Changing [these lines](https://github.com/Automattic/newspack-plugin/blob/e1ba3b6a30953b4d0f25d427a9da1dbb91498fc5/src/reader-activation/subscription-tiers-form.js#L60-L64) to the following fixed it for me:

```diff
-                               const params = new URLSearchParams( formData );
                                modal?.setAttribute( 'data-state', 'closed' );
+                               const url = new URL( form.action );
+                               for ( const param of formData.entries() ) {
+                                       url.searchParams.append( param[ 0 ], param[ 1 ] );
+                               }
                                window.newspackOpenModalCheckout( {
-                                       url: form.action + '?' + params.toString(),
+                                       url: url.toString(),
```

2. At testing step 10, I again had trouble getting the checkout modal to load. It looks like the URL was being constructed with `item_id=0` because you can't access a `WC_Order_Item_Product`'s `id` directly by array key—you need to use its `$item->get_id()` method. See suggestion below.

### Non-blocking questions

1. I noticed this layering effect when I visited a page with Campaigns prompts. Should the tiers modal hide Campaigns prompts and other modals/overlays when it's active, like the auth and checkout modals?

Screenshot 2025-10-02 at 11 50 45 AM

2. In the wizard UI for setting a Primary Subscription Tier Product, once you've set a primary product there's no way to unset it without selecting a different product. Should we include a "clear" button in case that's ever needed?

3. Also in the wizard UI, it might be helpful for publishers to include some description copy here describing what the Primary Subscription Tier Product's purpose is. Something like "Select a grouped or variable subscription product to allow readers to change their active subscriptions amongst all of its child products and variations."

4. A nice-to-have: Once you've selected a primary product it would be nice to have a link to that product's edit page in WP admin.

5. Very nitpicky, but I wanted the "Share the following URL to trigger the subscription upgrade: <URL>" message to show the URL as a link so I could command-click and copy it.

miguelpeixe:
> Looks like if there's already a query param in the current page, the URL constructed for the checkout modal will contain two ? characters and be invalid. Changing [these lines](https://github.com/Automattic/newspack-plugin/blob/e1ba3b6a30953b4d0f25d427a9da1dbb91498fc5/src/reader-activation/subscription-tiers-form.js#L60-L64) to the following fixed it for me:

Thank you for catching that! Fixed in 965cfafe337b1506f55b4217596d788fa0bb4c12

> It looks like the URL was being constructed with item_id=0 because you can't access a WC_Order_Item_Product's id directly by array key—you need to use its $item->get_id() method. See suggestion below.

I have no idea how it was working before, but I also needed this change while testing now.

> Should the tiers modal hide Campaigns prompts and other modals/overlays when it's active, like the auth and checkout modals?

Good call, added in e46af19228207f1c086ea88d8c83df87d18262f5. I opted for disabling all prompts because this is a campaign on its own. I don't think the reader should hit an overlay campaign after closing this modal. Wdyt?

> Also in the wizard UI, it might be helpful for publishers to include some description copy here describing what the Primary Subscription Tier Product's purpose is. Something like "Select a grouped or variable subscription product to allow readers to change their active subscriptions amongst all of its child products and variations."

Changed it to "Select a grouped or variable subscription product to allow readers to change their active subscriptions amongst all of its **linked** products and variations." so it reflects the grouped product configuration terminology.

> A nice-to-have: Once you've selected a primary product it would be nice to have a link to that product's edit page in WP admin.

Nice! Added in 6ae77b2fbb4d60b35538b78453787491e0aa6fe3

image


> Very nitpicky, but I wanted the "Share the following URL to trigger the subscription upgrade: " message to show the URL as a link so I could command-click and copy it.

Good call! Added in 72a427b9dbe0e97082a0390676606b4b9e7a7a17

dkoo (approved):
> I opted for disabling all prompts because this is a campaign on its own. I don't think the reader should hit an overlay campaign after closing this modal. Wdyt?

I'm fine with this approach, as long as it's okay that any inline prompts will also be hidden. Do you anticipate most publishers will share this link with the param appended the homepage, or something else like My Account?

Not sure if you saw my earlier comment, but since it's non-blocking I'll approve:

> In the wizard UI for setting a Primary Subscription Tier Product, once you've set a primary product there's no way to unset it without selecting a different product. Should we include a "clear" button in case that's ever needed?

dkoo (changes requested):
@miguelpeixe I missed something before approving—with the latest changes I see the modal checkout opening in a new browser tab instead of inside the modal. Is something off with the appending to `searchParams`?

EDIT: looks like if there's not already a modal checkout feature on the page, `window.newspackOpenModalCheckout` is undefined. This was probably obscured before because one of my Campaigns prompts had a Donate block (and therefore was rendering a checkout modal).

miguelpeixe:
> looks like if there's not already a modal checkout feature on the page, window.newspackOpenModalCheckout is undefined. This was probably obscured before because one of my Campaigns prompts had a Donate block (and therefore was rendering a checkout modal).

Whoops! I never caught that because my homepage has Checkout Button blocks in it. Fixed in 0e99b7aa45980ff068e8e091249f84dae82e49f7

> In the wizard UI for setting a Primary Subscription Tier Product, once you've set a primary product there's no way to unset it without selecting a different product. Should we include a "clear" button in case that's ever needed?

Sorry, I missed this one. Good call, added in 13359a48d2612a9e17b4f14a9ce969137013f12d

image

dkoo (approved):
Looking good now, thanks!

miguelpeixe:
Thank you for the thorough review and great suggestions! 🙇

PR #4192, opened 2025-09-17
hotfix: change sitekit check cadence

leogermani:
### All Submissions:

This is generating too many alerts for disconnected sites. No need to check that often.

1. On `trunk` run `wp cron event list` and confirm the `newspack_googlesitekit_disconnection_logger` event is scheduled hourly
2. Checkout this branch and visit the admin
3. Check the event again and see that it's now daily

### Other information:


rbcorrales (approved):
Approved. The migration logic should work correctly, though you could simplify by always clearing/rescheduling instead of using a DB option since `wp_clear_scheduled_hook()` is lightweight and this approach would be self-healing without needing persistent state to avoid bloating the DB.

PR #4200, opened 2025-09-23
feat(my-account): edit donation amount

miguelpeixe:
### All Submissions:

NPPD-102

Add "name your price" support to the subscription tiers modal and implement it for updating the donation amount (subscription switching).

### Initial state

image

When opening the modal, submission is disabled until the value or frequency changes.

###  Changing frequency

image

When changing the frequency, the suggested amount will be calculated from the current value, unless it's less than the product suggestion. In that case, it'll use the product value.

### Checkout

image

1. Make sure you have `define( 'NEWSPACK_MY_ACCOUNT_VERSION', '1.0.0' );` in your `wp-config.php`
2. As a new reader, use the Donate block to become a recurring donor
3. Navigate to My Account and click "My Subscription"
4. Click the "Change price" button and confirm that the modal opens in its initial state
5. Double the value, submit, and confirm that the checkout modal opens
6. Complete the checkout flow and confirm that the page refreshes and your subscription reflects the new value
7. Click "Change price" again and select a different frequency
8. Confirm the input value is the same, but according to the frequency
9. Change the value to the minimum value possible (probably $5)
10. Go through the checkout flow
11. In the refreshed page, click "Change price" again and change the frequency
12. Confirm the input value is the donation suggested amount

### Other information:


thomasguillot:
image

Can we make some changes, please?

The Frequency field needs a label, and the button group should span the full width. We could introduce a new class for this component, such as `is-block` (like Core does for ToggleGroupControl) , to achieve the full-width layout. This treatment differs from the “price table” frequency, since here we’re editing settings, so it should look more like a standard input.

Speaking of inputs, I’d like to add a small currency element inside the input, perhaps with a `<span class="currency">` inside for the currency symbol (and use `:has(.currency)` to style this input)

Finally, the submit button label should clearly display the new price and frequency.

Edit: Let's simplify the "Amount / frequency" label to just "Amount" since we get the full "Amount / frequency" as part of the submit button

image

miguelpeixe:
That makes sense, and I really like this new NYP amount input!

miguelpeixe:
@thomasguillot 065bb4330c761bd667035b8baafbe0b9fdf4be97 implements the new UI and a new currency input to the newspack ui:

image

image

In the `?ui-demo`:

image

thomasguillot:
Thank you @miguelpeixe, it looks much better 🙌 I pushed 2 small tweaks

miguelpeixe:
@thomasguillot I no longer get the control to render full-width. Is it intentional?

image

thomasguillot:
> I no longer get the control to render full-width. Is it intentional?

Nope 🫢

miguelpeixe:
@thomasguillot 6a6ab63c886ae92bf2dc6770a4e129c62a330768 breaks the modal interaction for 3+ frequencies non-NYP products. Changing the select option is no longer updating the options because the "segmented control" relies on the tab markup:

image

Just pushed 7f82324e07f3a4d5f9ba1f90e411f1977f8c0f10 to address that while preserving the new markup.

thomasguillot:
Oops, thank you @miguelpeixe

miguelpeixe:
Thank you for the style changes!

includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -114,7 +114,12 @@ public static function print_switch_subscription_modal() {
 				continue;
 			}
 			$product = wc_get_product( reset( $parent_products ) );
-			self::render_modal( $product, null, null, $switch_data );
+			$label = __( 'Change subscription', 'newspack-plugin' );
+			if ( Donations::is_donation_product( $product->get_id() ) ) {
+				$title = __( 'Edit donation', 'newspack-plugin' );
+				$label = __( 'Confirm donation', 'newspack-plugin' );
```
dkoo, line 135: Non-blocking: is it possible to change the label of this button too, based on the same logic?

Screenshot 2025-10-16 at 2 23 25 PM

includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -356,6 +390,48 @@ private static function render_product_card( $product, $show_variation_attribute
...
+			$base_product   = wc_get_product( $switch_subscription['item']['product_id'] );
+			$base_frequency = $base_product->get_meta( '_subscription_period' );
+			$base_interval  = $base_product->get_meta( '_subscription_period_interval' );
+			$base_amount    = $switch_subscription['item']['line_total'] / $base_interval / $divisor[ $base_frequency ];
+			if ( $current ) {
+				$value = $base_amount * $divisor[ $frequency ];
+			} else {
+				$value = max( ceil( $base_amount * $divisor[ $frequency ] * $interval ), $value );
+			}
+		}
```
dkoo, line 436: Non-blocking: this is nitpicky of me, so feel free to ignore!

This strikes me as a very "engineery" way to calculate the new value, especially when transitioning to/from a monthly value. For example, if I have a monthly donation of $50 and I want to update to an annual donation, my human logic would expect that the annual equivalent of $50 / month is 50 × 12 = **$600**. But this calculates it as 50 / 30 * 365 = **$608.30** which seems like a pretty random value to me without knowing the logic behind this calculation. If there's an easy way to express the calculations based on number of x in a y, this might make more intuitive sense. e.g.:

- 7 days in a week
- 30 days in a month (only when transitioning between days and months)
- 365 days in a year
- 4 weeks in a month
- 52 weeks in a year
- 12 months in a year

Otherwise, I'm not sure it's worth the extra logic to calculate the new value—I'd argue that we should just show the suggested amount for the new frequency.

dkoo (approved):
I'm approving because this works as described with no issues! I have a couple of non-blocking suggestions below.

includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
miguelpeixe: Good call! Updated in 64daf9cec9dfad2e8b5f2feb86c0bdaaffd848e8

includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
miguelpeixe: That makes sense. Normalizing to daily rates was a quick solution to the issue, but mapping out frequency-to-frequency multipliers is more appropriate. Changed in 9de06b50522841f5300989f7efa90f44a6301ee8

dkoo: Thanks for this extra logic! The calculations now make more intuitive sense to me.

dkoo (changes requested):
@miguelpeixe all is looking good, except I noticed something I didn't catch before. If you have a post-checkout newsletter signup modal, the page refresh after changing your NYP amount gets triggered before you can interact with the signup modal. I suggest we use a strategy here similar to the one we have for content gating to determine when to refresh: https://github.com/Automattic/newspack-plugin/blob/feat/my-account-edit-donation-amount/src/memberships-gate/gate.js#L50

miguelpeixe:
Whoops, good catch! This is covered for the [other My Account modal checkouts](https://github.com/Automattic/newspack-plugin/blob/772e81b59695568a192e66e9298c8eb685f3b90e/src/my-account/v1/utils.js#L44-L62), but this modal is handled differently. Fixed in 772e81b59695568a192e66e9298c8eb685f3b90e

dkoo (approved):
Nice, working well now!

PR #4203, opened 2025-09-25
feat(collections): add archive link in settings page

rbcorrales:
### All Submissions:

Adds a discrete external link icon next to the "Collections Settings" title that links to the Collections archive page. The link uses the custom slug if configured. This makes it easier for users to quickly navigate from the settings page to view their live Collections archive.

image

Closes NPPD-903.

1. Navigate to Newspack Settings → Collections.
2. Enable the Collections module if not already enabled.
3. Verify the external link icon appears next to "Collections Settings" title.
4. Click the icon to confirm it opens the Collections archive page in a new tab.
5. Test with custom slug: enable custom naming, set a custom slug, save settings, and verify the link uses the custom slug
6. Test accessibility: hover over the icon to see the tooltip, and verify screen reader compatibility

### Other information:

### DIFF
--- a/src/wizards/newspack/views/settings/collections/index.tsx
+++ b/src/wizards/newspack/views/settings/collections/index.tsx
@@ -3,8 +3,10 @@
  */
 
 import { __ } from '@wordpress/i18n';
-import { useState, useEffect } from '@wordpress/element';
-import { ToggleControl } from '@wordpress/components';
+import { useState, useEffect, useMemo } from '@wordpress/element';
+import { ToggleControl, Tooltip, Icon } from '@wordpress/components';
+import { external } from '@wordpress/icons';
+import { cleanForSlug } from '@wordpress/url';
 import WizardSection from '../../../../wizards-section';
 import WizardsActionCard from '../../../../wizards-action-card';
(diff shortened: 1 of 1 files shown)


dkoo (approved):

thomasguillot:
I have to say, the external link feels a bit out of place. It’s like we’re opening the Collections Settings in a new tab, when it would make more sense for it to be part of the Collections Archive section:

Screenshot 2025-09-30 at 07 56 07@2x

rbcorrales:
@thomasguillot the initial idea of putting it in the title was to make it as prominent as possible when you land on the settings for the first time, but I'm also okay with setting it within the archive section. The only thing is that the sample layout isn't possible, as the `WizardSection` component only accepts a string in the `description` field.

One option would be to render the link as a chlid element, but will appear in the next line:
image

Otherwise, it would require modifying the `WizardSection` and `SectionHeader` components to be able to accept React nodes. I see the SectionHeader component already has some logic for calling functions in the description, but the type declaration says strings only.

Let me know if you think it's worth making these changes to the components, or rendering it as a child element, or keeping it in the current place.

thomasguillot:
@rbcorrales I think the WizardSection/SectionHeader needs to accept React nodes. This can be useful to be able to add links etc... to the description.

PR #4207, opened 2025-09-26
feat: content gifting

miguelpeixe:
### All Submissions:

NPPD-722

Allow subscribers to gift articles for 24 hours, up to a configured limit:

image

Gifting an article creates a user meta entry with a key to unlock it. This key is valid for 24 hours and is shared via a query argument formatted as such: `{user_id}|{key}`, e.g.: `?content_key=79|uM5cAOZZy73AOiWGG13inWc9blf7RoS5`

### Jetpack sharing module

image

#### Styles

| | |
| --- | --- |
| Icon + Text | image |
| Icon only | image |
| Text only | image |
| Official buttons | image |

When enabling the gifting feature, the "Newspack Gift Article" sharing service is automatically added to the enabled sharing services. The sharing service is also added to the defaults, in case nothing is configured in the database.

### Without Jetpack sharing module

image

## Gifting modal

image

### Limit exceeded

image

## Viewing a gifted article

When viewing a gifted article, you get a configurable CTA. If no URL is set, it will look for the primary subscription product to trigger a modal checkout.

### Light

image

### Dark

image

### Default settings

1. Make sure you have:
   1. the Jetpack sharing module enabled
   2. metering disabled
   3. a primary subscription product (Audience -> Subscriptions)
2. Navigate to Audience -> Content Gate and enable Content Gifting
3. As a member, navigate to a free article (not locked behind a gate)
4. Confirm the button is not present, as this is not gated content
5. Navigate to a restricted article and confirm the button renders next to other sharing services
6. Click the button and confirm the modal renders with correct information and the "Copy link" button adds the string to the clipboard
7. Paste the link in an anonymous session
8. Confirm you get the full article with the default CTA. Clicking "Subscribe now" starts the modal checkout for the primary subscription product

### Custom Jetpack styles

1. Go to the Jetpack sharing module settings
2. Test all the different style options and confirm they render as expected in the subscriber session

### Without Jetpack sharing

1. Deactivate the Jetpack sharing module
2. Visit a restricted article as a member and confirm the "Gift this article" buttons render as expected

### Custom CTA settings

1. Edit the Content Gifting settings to use a different style, text, and URL
2. Visit the gifted article URL and confirm that the changes are applied and the button links to the URL

### Metering

Metering is already a form of gifting, so you can't gift metered content.

1. Enable metering
2. Confirm the "Gift this Article" button no longer renders

### Gifting limits

1. Disable metering again
2. Change the limit of gifting to 1
3. As a member, visit another gifted article
4. Click the "Gift this Article" button and confirm that the link is disabled

### Other information:


thomasguillot:
@miguelpeixe I added a few things:

* Ability to edit the "Subscribe now" label
* Direct preview of the banner (including Newspack UI in wp-admin)
* Reorganised the settings with headings

miguelpeixe:
Thank you! It looks great and tested well. I've updated the PR screenshot to match the new UI.

src/content-gate/content-gifting.scss
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.sd-social-icon .sd-content ul li[class*="share-"].share-newspack-gift-article a.sd-button {
+	background: var(--newspack-theme-color-primary-against-white, var(--newspack-ui-color-primary-60, #{colors.$primary-600})) !important;
+	color: var(--newspack-theme-color-against-primary, var(--newspack-ui-color-neutral-0, #{colors.$neutral-000})) !important;
```
dkoo, line 21: With the "icon only" style enabled in Jetpack settings, colors aren't quite matching on the front-end vs. what I see in the Jetpack settings admin page.

Jetpack settings:
Screenshot 2025-11-13 at 1 54 47 PM

Front-end:
Screenshot 2025-11-13 at 1 54 42 PM

dkoo: Also, when I select "Official buttons" the "gift this article" button disappears from the preview:

Screenshot 2025-11-13 at 1 58 04 PM

I do still see it on the front-end, though:
Screenshot 2025-11-13 at 1 58 26 PM

src/wizards/audience/views/setup/content-gating.js
```diff
@@ -76,6 +84,124 @@ export default withWizardScreen( () => {
...
+								label={ __( 'Subscribe button label', 'newspack-plugin' ) }
+								help={ __( 'Text displayed on the subscribe button in the banner.', 'newspack-plugin' ) }
+								value={ config.content_gifting.button_label }
+								onChange={ value => setConfig( { ...config, content_gifting: { ...config.content_gifting, button_label: value } } ) }
+								__next40pxDefaultSize
+							/>
+							<TextControl
+								label={ __( 'Subscribe button URL', 'newspack-plugin' ) }
+								help={ __(
+									'URL for the subscribe button in the banner. If not provided, the primary subscription tier product will be used with modal checkout.',
```
dkoo: Non-blocking: can we link "primary subscription tier product" to the relevant wizard page?

dkoo (changes requested):
@miguelpeixe This is working well and the UI looks good! I just ran into a few minor issues, outlined below and in inline comments:

1. When using "light" styles, the color of the CTA button doesn't match between the preview and the front-end. For reference, in my site the primary color is `#FF8200` (the orange color).

Preview:
Screenshot 2025-11-13 at 2 51 44 PM

Front-end:
Screenshot 2025-11-13 at 2 50 00 PM

3.  On the front-end, the banner is overlaid on top of the bottom-most edge of the viewport, which means it can obscure elements that are down there. I think we should try to add some bottom-padding to the `body` or a top-level container element equal to the height of the banner when it's rendered, so that it doesn't obscure anything.

Screenshot 2025-11-13 at 2 01 07 PM

4. It looks like I can use the same gift link to access an article multiple times in different anonymous browser sessions—so a single gift link once generated can be shared more or less infinitely. Is this intentional?

5. Maybe we should strip the `?content_key` URL param after it's been processed? That might also help control reuse of the same gift link a bit.

src/content-gate/content-gifting.scss
miguelpeixe: This was a tricky one. Jetpack hardcodes the handling of the official button styles. In 9bce65018a7f2c7a6f72b89d2f6c1d1f88f80511 I trick Jetpack into thinking this is a Print button so it doesn't remove its contents.

This is where it's handling the preview: https://github.com/Automattic/jetpack/blob/817c26ec371251dead8428dd43573f955feb9528/projects/plugins/jetpack/modules/sharedaddy/admin-sharing.js#L163-L188

src/wizards/audience/views/setup/content-gating.js
miguelpeixe: Good call! Added in 64f1284c3936d642af6f13dc2c0caf7638bd890c

image

miguelpeixe:
Thanks for the review!

> When using "light" styles, the color of the CTA button doesn't match between the preview and the front-end. For reference, in my site the primary color is #FF8200 (the orange color).

We were missing the theme CSS variables to match the Newspack UI experience in the admin area. Added in da2fbd5bb7ae8677ef61f451f745db61ec7a0764

> With the "icon only" style enabled in Jetpack settings, colors aren't quite matching on the front-end vs. what I see in the Jetpack settings admin page.

With the change mentioned above, they should now match. They'll match with the odd color, though... Maybe something about [this calculation](https://github.com/Automattic/newspack-theme/blob/b2dbc607fdcbb70952a186b71ac6b7b81d4b994d/newspack-theme/inc/color-patterns.php#L28) from theme?

> On the front-end, the banner is overlaid on top of the bottom-most edge of the viewport, which means it can obscure elements that are down there. I think we should try to add some bottom-padding to the body or a top-level container element equal to the height of the banner when it's rendered, so that it doesn't obscure anything.

Good call! Added a padding with dynamic offset detection to account for mobile and dynamic CTA content in 9317fc3ca7c7d0a202bd53b3941f95463233ea46

> It looks like I can use the same gift link to access an article multiple times in different anonymous browser sessions—so a single gift link once generated can be shared more or less infinitely. Is this intentional?
> 
> Maybe we should strip the ?content_key URL param after it's been processed? That might also help control reuse of the same gift link a bit.

Yes, anyone can use the link until the 24-hour expiration. I like the idea of stripping the query param. I implemented that with a cookie replacement, so the recipient can still visit the article for the granted time. 97f65186901d6b3c9c20c7c2ec0118732faadbab

The cookie starts with `wp_`, so batcache skips it.

src/content-gate/content-gifting.scss
dkoo: Gotcha, the hacky solution works!
Screenshot 2025-11-14 at 9 40 08 AM

dkoo (changes requested):
Thanks for the changes, @miguelpeixe! This is almost ready now, the only thing is I'm still seeing some color discrepancies between preview vs. front-end:

For the CTA in the banner:

| preview | front-end |
| ---- | ---- |
| Screenshot 2025-11-14 at 9 33 10 AM | Screenshot 2025-11-14 at 10 05 04 AM |

With Jetpack sharing + icon-only styles:

| preview | front-end |
| ---- | ---- |
| Screenshot 2025-11-14 at 10 08 07 AM | Screenshot 2025-11-14 at 9 43 10 AM |

Other than that, everything works as described!

miguelpeixe:
Thanks, @dkoo! I missed a couple of theme variables. Added in 13783c57e624dcc563f5b97d018fd95980c6bd4e

dkoo (approved):

PR #4216, opened 2025-09-30
feat(my-account): add delete payment method modal confirmation

rbcorrales:
### All Submissions:

Adds a confirmation flow before removing saved payment methods. Each delete action now opens a modal with the card brand/last four/expiry, and now all the dropdowns close once the modal appears (including address-related modals).

image

Closes NPPD-909.

1. Log in as a reader with multiple saved payment methods and visit My Account → Payment information.
2. Open the overflow menu for any deletable card and click "Delete payment method"
3. Confirm a modal appears with the corresponding brand/last four/expiry and that the dropdown (three-dot menu) automatically closes.
4. Click "Cancel" and verify that the card remains and that the menu items can be selected again.
5. Reopen the modal for the same card, click "Delete payment method" and make sure the card is removed after the page refresh.
6. Test setting up a different default payment method. The dropdown menu should be kept open, but with the options deactivated while the page refreshes (same behavior as before).

### Other information:

### DIFF
--- a/includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
+++ b/includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
@@ -40,6 +40,7 @@ class My_Account_UI_V1 {
 		\add_filter( 'default_option_woocommerce_myaccount_add_payment_method_endpoint', [ __CLASS__, 'add_payment_method_endpoint' ] );
 		\add_action( 'template_redirect', [ __CLASS__, 'redirect_payment_information_endpoint' ] );
 		\add_action( 'newspack_woocommerce_after_account_payment_methods', [ __CLASS__, 'add_payment_method_modal' ] );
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):

miguelpeixe (changes requested):
Sorry, accidentally approved.

After confirming deletion, the payment method doesn't get deleted, and I get the following error:

Image

This is not an issue in trunk (without the modal).

rbcorrales:
@miguelpeixe Thanks for flagging this! Seems that I broke it when converting the URL to a form action, and the nonce was being stripped. This is due to how forms with GET act as per the HTML spec (all the query parameters get removed from the action and replaced with whatever is set via hidden fields).

Addressed via [f6b7503](https://github.com/Automattic/newspack-plugin/pull/4216/commits/f6b75037fcba00fce613856da8842f2afb641fe9).

miguelpeixe (approved):
Thank you for revising!

Looks good and tested well

PR #4220, opened 2025-10-02
fix(my-account): new payment method style tweaks

miguelpeixe:
### All Submissions:

NPPD-485

Style improvements to the new payment method form:

| Before | After |
| --- | --- |
| image | image |

Add a new payment method via My Account -> Payment information and confirm it works as expected

### Other information:

### DIFF
--- a/includes/class-newspack-ui.php
+++ b/includes/class-newspack-ui.php
@@ -239,6 +239,7 @@ class Newspack_UI {
 										'aria-required' => true,
 										'value'         => true,
 										'disabled'      => true,
+										'checked'       => true,
 									],
 									'select'   => [
 										'name'             => true,
--- a/includes/plugins/woocommerce/class-woocommerce-connection.php
+++ b/includes/plugins/woocommerce/class-woocommerce-connection.php
@@ -63,7 +63,7 @@ class WooCommerce_Connection {
 		\add_action( 'woocommerce_payment_complete', [ __CLASS__, 'order_paid' ], 101 );
 		\add_action( 'woocommerce_after_checkout_validation', [ __CLASS__, 'rate_limit_checkout' ], 10, 2 );
 		\add_filter( 'woocommerce_add_payment_method_form_is_valid', [ __CLASS__, 'rate_limit_payment_methods' ] );
-		\add_action( 'wc_stripe_save_to_subs_checked', '__return_true' );
+		\add_filter( 'wc_stripe_save_to_subs_checked', '__return_true' );
 
 		\add_filter( 'page_template', [ __CLASS__, 'page_template' ] );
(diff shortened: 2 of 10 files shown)


thomasguillot:
@miguelpeixe thank you for this PR. These are details that bothered me for quite some time now. I went even further with the  only 1 payment method:



I wonder if we can do something similar when we're checking out

miguelpeixe:
Awesome! I like that better too.

If you look at NPPD-485 you'll see that the "Change Payment Method" had this approach as well, but at some point got boxed, not sure at which point.

Let's use this PR to fix the inconsistencies

miguelpeixe:
This is how it looks now, btw

image

dkoo (changes requested):
Styles look good!

Screenshot 2025-10-10 at 11 15 57 AM

I'm seeing a PHP warning on the Payment Information page:

```
PHP Warning:  Undefined array key "form_id" in /newspack-repos/newspack-plugin/includes/class-newspack-ui.php on line 219
```

Also, this might be for another PR, but is it possible to make both checkboxes at the bottom of the modal either checked by default, or even checked by default + visually hidden?

miguelpeixe:
Thank you for the review!

> I'm seeing a PHP warning on the Payment Information page:
> PHP Warning:  Undefined array key "form_id" in /newspack-repos/newspack-plugin/includes/class-newspack-ui.php on line 219

Good catch! This is in trunk too, from #4216. Fixed in 2edb7ce016752f9a25d3684f4e6721bfb332aa20

>  Also, this might be for another PR, but is it possible to make both checkboxes at the bottom of the modal either checked by default, or even checked by default + visually hidden?

I agree that the first checkbox shouldn't exist, and checking it doesn't really change anything in this form. Fixed in f0f8016eedeee7345023b23dcbad5c97d71dc9d4.

IMO, the second checkbox is relevant and shouldn't be checked by default.

dkoo:
> IMO, the second checkbox is relevant and shouldn't be checked by default.

I'm open to discussing this further, but it's how things work in v0 now. #2934 and #3643 made the checkbox checked by default and hid it visually (because publishers were receiving feedback from readers being confused by it and/or not checking it after adding a new payment method, resulting in failed renewals), so showing it now would be a change in behavior.

miguelpeixe:
Ah! Thank you for the additional context. I didn't realize that. I'll fix that then

miguelpeixe:
I also realized that what was done in #2934 no longer works. Maybe something from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3461

Investigating

miguelpeixe:
Took me a while... it turned out to be kses filtering out the `checked` attribute. Fixed in fe16aae39be0d99900793f49cf6abddeccbaa426

Should be good now!

dkoo (approved):

PR #4224, opened 2025-10-06
fix: keep prevent one and two col Collections block grid from being increased on smaller screens [NPPM-2285]

laurelfulford:
### All Submissions:

The Collections block drops down to three, then two columns as you shrink down the browser window. This can be a little strange with the one and two column layouts, though, since it increases their column counts. This PR makes the two column layout stay two columns on tablet-sized screens, and the one column layout stay one column on all screen sizes.

1. Add a couple Collections blocks to a page; make them one column and two column; if you add more, make them three and four column just to test for regressions. 
2. Shrink down the screen size; note that the one column switches to three, then two columns, and that the two column switches to three on tablet-sized screens. This isn't necessarily bad if there's more than one or two collections in the block (though it might be unexpected); it's very likely these lower column counts only show one or two covers, like in a CTA, in which case the column change adds extra space:

One column on tablet-sized screen:
CleanShot 2025-10-06 at 10 44 44

Two columns on tablet-sized screen:
CleanShot 2025-10-06 at 10 45 02

One column on mobile-sized screen:
CleanShot 2025-10-06 at 10 45 14

4. Confirm the one and two column layouts remain that way on all screensizes:

One and two column on a tablet-sized screen:

CleanShot 2025-10-06 at 10 48 40

One Column on a mobile sized screen:

CleanShot 2025-10-06 at 10 47 33

Two Columns on a mobile sized screen:
CleanShot 2025-10-06 at 10 47 49
5. Confirm the other column counts still work the same way. 

### Other information:

### DIFF
--- a/src/blocks/collections/styles/_layout-grid.scss
+++ b/src/blocks/collections/styles/_layout-grid.scss
@@ -9,12 +9,16 @@
 
 
 		// Mobile first: 2 columns.
-		grid-template-columns: repeat(2, minmax(0, 1fr));
+		&:not(.columns-1) {
+			grid-template-columns: repeat(2, minmax(0, 1fr));
+		}
 
 		// Tablet: 3 columns.
 		@media #{bp.$media-sm-up} {
-			grid-template-columns: repeat(3, minmax(0, 1fr));
-			gap: 16px;
+			&:not(.columns-1, .columns-2) {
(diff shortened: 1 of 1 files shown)


leogermani (approved):

laurelfulford:
Thanks @leogermani!

PR #4226, opened 2025-10-07
fix(memberships-for-teams): sync contact data for team members

dkoo:
### All Submissions:

Adds some Data Events handlers and listeners for the Memberships for Teams plugin so that contact data is synced for readers who might not have any active subscriptions but can have memberships via teams. Closes a loop mentioned in  #3325: 

> Note that a sync to ESP is not triggered when a user is granted a membership or is added to a team. This will be subject for another PR

Also fixes some template collisions and styles in My Account when v1 is enabled.

Also fixes the value of `NP_Woo Team` to populate for team owners, not just members.

Also fixes the value of `NP_Woo Team` if a reader is an owner or member of more than one team.

Closes NPPM-2280.

1. Check out this branch.
2. Enable Memberships for Teams plugin and set up a team membership subscription product with 3+ maximum members. Tie this to a membership plan that grants membership with an active subscription for the product.
3. As a new user, purchase a subscription for the team membership product. Confirm that you see a data event fire for a `team_created` action, and that this triggers a contact sync to the connected ESP with a `NP_Woo Team` metadata field containing the slug of the newly created team.
4. Go to My Account > Teams, view the team, then go to Add Member. Either copy the registration link or invite an existing reader account invite email.
5. In a new session, visit the registration link or accept the email invite for the team.
6. After registering or logging in as the invited team member, confirm that you see a data event fire for a `team_member_added` action added and that this triggers a contact sync to the connected ESP with a `NP_Woo Team` metadata field containing the slug of the team the member has been added to, plus a `NP_Membership Status` with value of `team member` since they don't have a direct subscription of their own.
7. In a new session, log in as the team member from step 6 and confirm that you see a data event fire for `reader_logged_in` with the same contact info as in step 6.
8. In another new session, register or log into a different reader account that has no subscriptions or orders, then use the registration link from step 4 to join the team. Confirm that this triggers an ESP contact sync with similar info as in steps 6-7, but for this already-logged-in reader.
9. As a WP admin, in WooCommerce > Memberships > Teams create a new team and designate the team owner from step 3 as the owner of this team (so that they're associated with multiple teams).
10. After saving the new team, confirm that the owner's contact info is synced to the ESP with `NP_Woo Status` containing slugs for both teams.

### Other information:


includes/plugins/class-teams-for-memberships.php
```diff
@@ -21,11 +22,134 @@ class Teams_For_Memberships {
...
+	}
+
+	/**
+	 * Sync reader data on login.
+	 *
+	 * @param int   $timestamp Timestamp of the event.
+	 * @param array $data      Data associated with the event.
+	 * @param int   $client_id ID of the client that triggered the event.
+	 */
+	public static function reader_logged_in( $timestamp, $data, $client_id ) {
```
leogermani, line 137: Starting to test this PR now.

Why we need another handler of reader_logged_in here? Wouldn't the listener and handler that exist in the main ESP sync class already handle it?

dkoo: No, [not if the user has 0 orders](https://github.com/Automattic/newspack-plugin/blob/396e607674c1017133561a77541a67dac5b651de/includes/data-events/connectors/class-esp-connector.php#L121-L123). That's why lower down on line 145, we bail if the user does have more than 0 orders—that way we won't queue two syncs on the `reader_logged_in` handler.

leogermani (commented):
The code looks good, I left a comment, but I can't confirm that this is working.

I see the events being fired but I either don't see the handlers being invoked or the sync happening.

I see the `team_created` action being fired but nothing happens.

When I add a member via invitation, I see the `member_added` event but no sync happens with the additional data.

There's a deprecation notice. Here's the log I see with an additional line I added to callback to make sure it was being called.

```
[30-Oct-2025 21:33:45 UTC] [974]  [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::handle]: Executing action handlers for "team_member_added".
[30-Oct-2025 21:33:45 UTC] Syncing team member ID 62 for team ID 799
[30-Oct-2025 21:33:45 UTC] PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/wp-includes/formatting.php on line 3549
```

Maybe there's something going on with that thing where we run only one sync at the end of the request, and merge all the data that should be synced? I have network installed and the `newspack_network_user_updated` is also being triggered...

leogermani:
> Maybe there's something going on with that thing where we run only one sync at the end of the request, and merge all the data that should be synced? I have network installed and the `newspack_network_user_updated` is also being triggered...

@dkoo nvm, I had forgottern to activate the "Woo Team" field in the sync settings. It's working fine.

The only thing I noticed is that if the "Woo Team" field is not active, the "Membership Status" is sent as an empty value for team members. When I activated "Woo Team" I also saw that the "Status" now shows "team member". Not sure if that's intended

dkoo:
> The only thing I noticed is that if the "Woo Team" field is not active, the "Membership Status" is sent as an empty value for team members. When I activated "Woo Team" I also saw that the "Status" now shows "team member". Not sure if that's intended

@leogermani good catch! a82663254 tweaks the order of things so that the `team member` value for `NP Membership_Status` is still sent even if the `woo_teams` field isn't enabled.

leogermani (approved):

PR #4227, opened 2025-10-08
fix(metering): get current user metered views

miguelpeixe:
### All Submissions:

There's a missing `$count` variable from #4176 that slipped by me during review. This PR is meant to address that, but while at it, I also propose some changes to this public method so it's more predictable and generic.

The `// For logged-in users, calculate the remaining views based on their metering data.` comment is also not accurate and represents a previous approach.

Same as #4176

### Other information:

### DIFF
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -71,7 +71,7 @@ final class Blocks {
 			$script_data['content_gate_data'] = [
 				'anonymous_metered_views' => Metering::get_total_metered_views( false ),
 				'loggedin_metered_views'  => Metering::get_total_metered_views( true ),
-				'metered_views'           => Metering::get_metered_views(),
+				'metered_views'           => Metering::get_current_user_metered_views(),
 				'metering_period'         => Metering::get_metering_period(),
 			];
 		}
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Works as expected without any PHP warnings!

Approving for functionality, but I noticed a couple of minor style issues while testing that we could choose to tackle either here or in a separate PR if needed. The default inner blocks of the Countdown blocks don't collapse gracefully at smaller viewports:

Screenshot 2025-10-10 at 4 58 58 PM

If we were to enable the "Allow to wrap to multiple lines" attribute on the top-level Row block, it would collapse like this, which I think looks better at smaller viewports:

Screenshot 2025-10-10 at 4 58 16 PM
Screenshot 2025-10-10 at 4 58 28 PM

Also, the horizontal spacing around the Article Count block looks a little too much on the front-end, compared to the editor. It's especially noticeable if you add another block in front of the count like this:

Screenshot 2025-10-10 at 4 42 21 PM

dkoo:
@miguelpeixe while testing this PR I discovered a bug in the "auto-refresh" feature for content gating. Fix is here: https://github.com/Automattic/newspack-plugin/pull/4230

miguelpeixe:
Thanks @dkoo! I'll merge this one as it doesn't touch the countdown block styles. Started #4239 to address your comments.

PR #4218, opened 2025-10-01
feat(woocommerce): enable block emails

chickenn00dle:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/NPPM-2045/evaluate-woo-block-based-email-builder-andor-update-transactional

This PR enables the new woo email block editor for publishers by default. This also adds a setting to our Newspack email settings since the Woo option is currently a burried in Woo advanced settings:

Screenshot 2025-10-06 at 15 07 36

Finally this adds a one time upgrade to any of the new email templates for publishers that enabled the feature before the text was updated by Woo to be more reasonable.

### How to test the changes in this Pull Request:

1. Before checkout out this branch, go to WooCommerce > Settings > Advanced > Features and enable the `Block Email Editor (alpha)` setting.
2. Now go to WooCommerce > Settings > Emails and choose one or two emails and edit with any custom content. Make note of which emails you edited!
3. Disable the setting from step 1
4. Checkout this branch and build
5. Confirm a new setting exists in Newspack > Settings > Emails as pictured above
6. Confirm the new Woo email editor is enabled (see step 2)
    
Screenshot 2025-10-06 at 15 12 19

7. View the content from the emails you edited in step 2, and confirm the changes are still present
8. Now, using the toggle from step 5, disable the new email editor 
9. Confirm the block editor is no longer available and the previous woo email settings are present.

Screenshot 2025-10-06 at 15 12 39

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #4072, opened 2025-07-07
feat: prepare colors npm package

thomasguillot:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This moves the colours from the `src` folder into a the `package` directory, so it's organised the same way we organised [the icons](https://github.com/Automattic/newspack-plugin/pull/3984). This PR also adds the necessary files to publish this as an npm package, `newspack-colors`, so it can be imported into blocks and other plugins.

### How to test the changes in this Pull Request:

1. Switch to this branch
2. Check if the colors are still working (dashboard, blocks)
3. try modify the colors and see if they are applied.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #4238, opened 2025-10-15
chore: move /src/components to /packages and fix shared module imports

dkoo:
### All Submissions:

When reviewing #4218 we realized that the NPM package for `newspack-components` is out of sync with the src files in this repo, resulting in possible conflicts between this repo's code and those using the components package. In investigating, I also saw that the shared modules imported by modules inside `src/components` are not being packaged correctly with distributable builds, particularly the colors module which was moved from the `src` directory to `packages/colors` in #4072. To fix this, this PR makes the following changes:

- Moves the `src/components` directory to `packages/components`, so that all NPM packages within this directory live within the same parent directory
- Updates paths of all `src/*` modules referencing `src/components` to `packages/components`
- Moves `src/shared/js/index.js` and `src/shared/js/button-props.js` to `packages/components/src`, as these files are only referenced by modules within the components directory.
- Updates the `copy-styles.js` script to also copy the `packages/colors` module to the dist build so that these imports will continue to work once published as a standalone package.
- Removes a negative margin declaration from the `ActionCard` component which is no longer needed but which is conflicting with a style declaration in `src/admin/style.scss`, resulting in an unpredictable cascade.

Once this is tested and merged, we will publish a new version of `newspack-components` to NPM.

1. Check out this branch and deactivate all plugins that use `newspack-components` as an NPM package. This includes: `newspack-ads`, `newspack-blocks`, `newspack-popups`, `newspack-listings`, `newspack-block-theme`, `newspack-newsletters`, `newspack-multibranded-site`, and `newspack-story-budget`. This is to avoid potential conflicts with assets in those plugins for the first part of testing.
2. From the root directory, run `npm ci --legacy-peer-deps`, `npm run watch` and `npm run build` and confirm that all complete successfully.
3. Visit the Components Demo page at `<domain>/wp-admin/admin.php?page=newspack-components-demo` and confirm that all components still look and behave as expected.
4. Browse through some Newspack wizard pages in WP admin and confirm that all still look and behave as expected.
5. Test that the dist build will work as a standalone package. From `/packages/components` of this repo, run `npm i --legacy-peer-deps` and then `npm run prepublishOnly` to build the dist files.
6. In the repo directory for one of the plugins you deactivated during step 1, find `node_modules/newspack-components` and delete everything inside. Copy all files except for `node_modules` from inside `newspack-plugin/packages/components` to the other plugin's `node_modules/newspack-components` directory. This simulates what will happen when the `newspack-components` NPM package gets updated and pulled into another repo that uses it as a dependency.
7. In the other plugin's directory, run `npm run watch` and `npm run build` and confirm that they complete successfully.
8. Test some of the features of this plugin to ensure that the components still look and behave as expected. For example, if testing with Newspack Blocks, add some Content Loop, Checkout Button, and Carousel blocks to a post and confirm they still work in the editor.

### Other information:


chickenn00dle (changes requested):
Works mostly as expected. I am running into some issues with some of the wizard pages. For example, the audience wizard seems to infinitely render. This happens even when all of our newspack extensions are active:

https://github.com/user-attachments/assets/346344b3-bb19-40eb-9efe-a97b4156ef1e

chickenn00dle (approved):
Hmm. https://github.com/Automattic/newspack-plugin/pull/4238#pullrequestreview-3345817416 seems to still be happening after going back to `trunk` so I think this might be something with my dev environment. Gonna go ahead and approve this assuming conflict will get resolved first.

dkoo:
Thanks, @chickenn00dle! I can't replicate the infinite loading, so if you figure out what's causing it, let me know!

PR #4240, opened 2025-10-16
fix: don't attach an existing Collections taxonomy if it's already attached to a Collection CPT

laurelfulford:
### All Submissions:

Right now when you create a new Collection CPT, it looks for a `newspack_collection_taxonomy` term that matches the same name to use to connect posts and the CPT.

However, there's a weird edge case where, if you name Collections the same as an existing CPT, it will "steal" its taxonomy term. This ultimately makes it impossible to assign the original Collection to posts, even if you rename the second one. 

This PR tweaks the behaviour to check if the term exists AND if it's already assigned to a CPT. If it is, a new term should be created. This may not be the best approach to fixing this, though -- I'm open to feedback!! 

Note: this doesn't fix Collections that are already in the detached state 😕 If this makes sense, I think we'll need to do that manually as it comes up. 

1. Create and publish a Collection called "October 2025".
2. Create and publish a second Collection also called "October 2025". 
3. Change the name of the second Collection to something else (like "September 2025")
4. Create a new post and try to assign the "October 2025" Collection - it won't be returned. 
5. Apply this PR, and repeat steps 1-4 with new Collection names; confirm that both the second original Collection name and its copycat can be applied to a new post. 

### Other information:

### DIFF
--- a/includes/collections/class-sync.php
+++ b/includes/collections/class-sync.php
@@ -231,6 +231,13 @@ class Sync {
 			if ( empty( $existing_term ) ) {
 				self::create_linked_term( $post );
 			} else {
+				// Check if existing term is already linked to a Collection.
+				$existing_term_id = $existing_term[0];
+				$linked_post_id = self::get_collection_linked_to_term( $existing_term_id );
+				if ( $linked_post_id ) {
+					self::create_linked_term( $post );
+					return;
+				}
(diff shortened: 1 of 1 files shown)


laurelfulford:
This ends up creating orphaned CPTs 😅 Closing and finding a different approach!

PR #4258, opened 2025-10-27
fix: add corrections on earlier the_content filter

dkoo:
### All Submissions:

Tweaks the `the_content` filter priority for adding corrections to as early as possible, so that a post's corrections are more likely to appear directly after the actual post content and not after any other additions that might be added via a `the_content` filter (such as Campaigns prompts).

Closes NPPM-2331.

1. On `release`, publish an inline Campaigns prompt that appears at 100% of the post content, then add a correction to a post.
2. Observe that on the front-end, the correction appears after the Campaigns prompt.
3. Check out this branch, refresh, and confirm that the correction now appears before the Campaigns prompt.

### Other information:

### DIFF
--- a/includes/corrections/class-corrections.php
+++ b/includes/corrections/class-corrections.php
@@ -46,7 +46,7 @@ class Corrections {
 	 */
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_post_type' ] );
-		add_filter( 'the_content', [ __CLASS__, 'output_corrections_on_post' ] );
+		add_filter( 'the_content', [ __CLASS__, 'output_corrections_on_post' ], 1 );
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
Looks good!

Image

Nonblocking nit: wdyt about making this priority slightly larger than 1 to allow for some flexibility with customization?

dkoo:
Thanks, @chickenn00dle! Looks like [Campaigns runs its filter at `1` already](https://github.com/Automattic/newspack-popups/blob/trunk/includes/class-newspack-popups-inserter.php#L91) so we'd have to update both. Let's stick with `1` for now and see if we need to tweak this further in the future.

PR #4265, opened 2025-10-28
feat(content-gate): content rules

miguelpeixe:
### All Submissions:

Implements the gate content rules and organizes content gate components.

1. Make sure you have the `NEWSPACK_CONTENT_GATES`  constant set to `true`
2. Navigate to Audience -> Content Gates
3. Create a new gate
4. Add multiple access rules and content rules (make sure to include taxonomy content rules)
5. Set values for the rules and metering
6. Click Save
7. Refresh the page confirm the options persisted
8. Reorder the gates using the arrows and drag (regression test)
9. Refresh the page and confirm the order persisted

### Other information:


src/wizards/audience/views/content-gates/content-rule-control.tsx
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Internal dependencies
+ */
+import ContentRuleControlTaxonomy from './content-rule-control-taxonomy';
+import { FormTokenField } from '../../../../../packages/components/src';
+
+const noop = () => {};
+
+export default function ContentRuleControl( { slug, value, onChange }: GateContentRuleControlProps ) {
+	const rule = window.newspackAudienceContentGates.available_content_rules[ slug ];
```
chickenn00dle: Do we need to protect against `available_content_rules` or `available_content_rules` being undefined?

```
window.newspackAudienceContentGates?.available_content_rules?.[ slug ];
```

dkoo:
@miguelpeixe @chickenn00dle 1e1884c updates the REST API endpoints so that we can use Wizard methods for the initial data fetch and adding/saving/deleting content gates. I saw some weirdness in the behavior when dragging and dropping for changing priority, but I'm not sure whether it's related to my changes.

miguelpeixe:
Thanks @dkoo!

Pushed 02a413770b8bc9ac920a2703e493e8c528d8c0e8 so the saved gate data bubbles up to the gate list. This prevents an old gate version from displaying, in case the card is closed and reopened.

dkoo:
@miguelpeixe since you and I both did a lot of work on this PR, should someone else review it?

miguelpeixe:
I think so... Maybe @rbcorrales could take a look?

rbcorrales:
I can't add a line comment on this one since it's not part of the introduced changes by this PR, but there's a hardcoded post ID here in the `wp_get_post_tags` function:
https://github.com/Automattic/newspack-plugin/blob/7379869e629b83ed031c38075d384bd492ec7ee3/includes/content-gate/class-content-restriction-control.php#L100

rbcorrales:
Not part of the PR, but the access here shouldn't be `$gate['id']` instead of `$gate->ID`?
https://github.com/Automattic/newspack-plugin/blob/7379869e629b83ed031c38075d384bd492ec7ee3/includes/content-gate/class-content-restriction-control.php#L106-L108

Since `get_gates()` (through `get_gate()`) returns an array of arrays, not objects:
https://github.com/Automattic/newspack-plugin/blob/7379869e629b83ed031c38075d384bd492ec7ee3/includes/content-gate/class-content-gate.php#L767-L776

Edit: The method doesn't seem to be using the new `content_rules` meta being saved by this PR (`get_post_content_rules()`). It reads post_types meta and taxonomy terms directly from the gate post instead. I assume this is intentional/WIP given the related TODO on line `480: // TODO: Content Gate content rules check.`

src/wizards/audience/views/content-gates/content-gates.tsx
```diff
@@ -6,74 +6,98 @@
...
-				data.sort( ( a, b ) => a.priority - b.priority );
-				setGates( data );
-			} )
-			.catch( error => console.error( error ) ) // eslint-disable-line no-console
-			.finally( () => setIsInFlight( false ) );
-	}, [] );
+		if ( wizardData?.error ) {
+			console.error( wizardData.error ); // eslint-disable-line no-console
+		}
+	}, [ wizardData ] );
```
rbcorrales: Missing `hasCompletedInitialFetch` in `useEffect` dependencies.

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { decodeEntities } from '@wordpress/html-entities';
+import { addQueryArgs } from '@wordpress/url';
+
+const debounce = ( func: ( search?: string ) => void, wait: number ) => {
+	let timeout: NodeJS.Timeout;
+	return ( search?: string ) => {
+		clearTimeout( timeout );
+		timeout = setTimeout( () => func( search ), wait );
+	};
+};
```
rbcorrales, line 22: Minor: The debounce function could use ReturnType<typeof setTimeout> instead of NodeJS.Timeout for better browser compatibility since NodeJS.Timeout isn't guaranteed to be in scope.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -21,6 +21,71 @@ public static function init() {
 		add_filter( 'newspack_is_post_restricted', [ __CLASS__, 'is_post_restricted' ], 10, 2 );
 	}
 
+	/**
+	 * Get the post types that can be restricted.
+	 */
+	public static function get_available_post_types() {
```
rbcorrales, line 27: Minor: There are two similar methods for getting available post types with slightly different return formats:
https://github.com/Automattic/newspack-plugin/blob/7379869e629b83ed031c38075d384bd492ec7ee3/includes/content-gate/class-content-gate.php#L278-L314

The duplication is a potential source of bugs in the future. Ideally, one shape should be kept and the other removed or adapted. Could these be consolidated into one method, or have one call the other with a format transformation?

src/content-gate/editor.js
```diff
@@ -244,6 +255,26 @@ function GateEdit() {
 					</Fragment>
 				) }
 			</PluginDocumentSettingPanel>
+			{ availablePostTypes.length > 0 && (
+				<PluginDocumentSettingPanel name="content-gate-post-types-panel" title={ __( 'Post Types', 'newspack-plugin' ) }>
+					<BaseControl id="content-gate-post-types" help={ __( 'Restrict all posts of the selected post types.', 'newspack-plugin' ) } />
+					{ availablePostTypes.map( ( { name, label } ) => (
```
rbcorrales: Not sure if this is intentional, but the editor now pulls post types from `Content_Restriction_Control::get_available_post_types()` (value/label), while the UI still destructures `{ name, label }`. That may mean the checkbox keys are undefined and selections might not persist. If we intend to use this format, should the UI read `value` instead?

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -220,8 +221,8 @@ public function sanitize_gate( $gate ) {
 			'title'         => sanitize_text_field( $gate['title'] ),
 			'description'   => sanitize_text_field( $gate['description'] ),
 			'metering'      => $this->sanitize_metering( $gate['metering'] ),
-			'access_rules'  => $this->sanitize_access_rules( $gate['access_rules'] ),
-			'content_rules' => $gate['content_rules'], // TODO: Sanitize content rules.
+			'access_rules'  => $this->sanitize_rules( $gate['access_rules'] ),
+			'content_rules' => $this->sanitize_rules( $gate['content_rules'], 'content' ),
 			'priority'      => intval( $gate['priority'] ),
 		];
 	}
```
rbcorrales, line 242: Core only runs sanitizers defined on individual args. A route-level `sanitize_callback` isn't picked up. `register_rest_route treats` non-numeric keys as route options, and `WP_REST_Request::sanitize_params()` only looks at `attributes['args'][param]['sanitize_callback']` (see [wp-includes/rest-api/class-wp-rest-
  request.php](https://github.com/WordPress/wordpress-develop/blob/6.8.3/src/wp-includes/rest-api/class-wp-rest-request.php#L840-L850)). In this file the callbacks sit at the route level, so `sanitize_gate()/sanitize_gates()` might never run. We should move the sanitizer into each arg.

rbcorrales (changes requested):
I may have further feedback as I continue reviewing this, but I'm sending what I've noticed so far so you can begin addressing those.

Also, some of the Copilot comments seem valid.

rbcorrales:
Another issue unrelated to this PR, but the property here is named `$rules`, not `$registered_rules`. This means the duplicate rule check never works.
https://github.com/Automattic/newspack-plugin/blob/7379869e629b83ed031c38075d384bd492ec7ee3/includes/content-gate/class-access-rules.php#L24

https://github.com/Automattic/newspack-plugin/blob/7379869e629b83ed031c38075d384bd492ec7ee3/includes/content-gate/class-access-rules.php#L53

includes/content-gate/class-content-gate.php
```diff
@@ -45,7 +45,7 @@ class Content_Gate {
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_post_type' ] );
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
-		add_action( 'admin_init', [ __CLASS__, 'redirect_cpt' ] );
+		// Commenting out to temporarily enable CPT UI for testing: `add_action( 'admin_init', [ __CLASS__, 'redirect_cpt' ] );` !
```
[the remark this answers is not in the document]
miguelpeixe: We don't need to comment this out anymore. Changed the strategy in 7efb85cc4c3f11fec3cbad122a90f21eaeaf1c0d so it redirects to the right place if the feature flag is enabled.

src/wizards/audience/views/content-gates/content-gate-settings.tsx
```diff
@@ -1,173 +1,71 @@
...
+			{
+				onSuccess( data ) {
+					onSave( data );
+				},
+				onError( error ) {
+					console.error( error ); // eslint-disable-line no-console
+				},
+			}
+		);
+	}, [ gate.id, accessRules, contentRules, metering ] );
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 6c26688176c9bbd0ef7dea5a8e9989c1f54cf9ce

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
```diff
@@ -0,0 +1,130 @@
...
+		apiFetch< { id: number; name: string }[] >( {
+			path: addQueryArgs( 'wp/v2/' + endpoint, {
+				include: value.join( ',' ),
+			} ),
+		} ).then( terms => {
+			setSavedItems(
+				terms.map( term => ( { value: term.id.toString(), label: decodeEntities( term.name ) || __( '(no name)', 'newspack-plugin' ) } ) )
+			);
+		} );
+	}, [] );
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 4dd730528ca7fc88ce668eef5080a2f284c0b2a6

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+	// Set initial suggestions.
+	useEffect( () => {
+		fetchSuggestions();
+	}, [ fetchSuggestions ] );
+
+	const debouncedFetchSuggestions = useCallback( debounce( fetchSuggestions, 100 ), [] );
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 53641d049da6a66f370a7c39c2e1be9b91c7b0e7

src/wizards/audience/views/content-gates/content-rule-control.tsx
```diff
@@ -0,0 +1,47 @@
...
+					onChange={ ( items: string[] ) => onChange( rule.options?.filter( o => items.includes( o.label ) ).map( o => o.value ) ?? [] ) }
+					suggestions={ rule.options.map( o => o.label ) }
+					__experimentalExpandOnFocus={ true }
+				/>
+				<CheckboxControl
+					label={ __( 'Exclusion rule', 'newspack-plugin' ) }
+					help={ __( 'Apply this rule to everything EXCEPT the items matching the above.', 'newspack-plugin' ) }
+					checked={ false }
+					onChange={ noop }
+				/>
```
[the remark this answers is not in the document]
miguelpeixe: We'll keep it

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -289,11 +291,33 @@ public function sanitize_access_rule( $access_rule ) {
...
+	 *
+	 * @return mixed|\WP_Error The sanitized content rule or error if invalid.
+	 */
+	public function sanitize_content_rule( $content_rule ) {
+		$rules = Content_Gate::get_content_rules();
+		if ( ! isset( $rules[ $content_rule['slug'] ] ) ) {
+			return new \WP_Error( 'invalid_content_rule_slug', __( 'Invalid content rule slug.', 'newspack-plugin' ), [ 'status' => 400 ] );
+		}
+
+		if ( ! empty( $content_rule['options'] ) ) {
```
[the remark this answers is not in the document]
miguelpeixe: Nope. `$content_rule['options']` is correct.

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -289,11 +291,33 @@ public function sanitize_access_rule( $access_rule ) {
...
+			if ( empty( array_intersect( $content_rule['value'], $rules[ $content_rule['slug'] ]['options'] ) ) ) {
```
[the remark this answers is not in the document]
miguelpeixe: There was another issue with the sanitization strategy, which was addressed along with this one in a4785cc791f490c4fcdec885f36bb76003f9abbe.

The `sanitize_callback` was being called as a `register_rest_route()` root argument, and should actually be per argument item.

117ddf59749b93ce151ffcd791bf9e164aa1c653 also simplifies the priority update payload. There's no need to send out and sanitize the entire gate if only the ID and priority are used in the endpoint callback.

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
```diff
@@ -0,0 +1,130 @@
...
+				path: addQueryArgs( 'wp/v2/' + endpoint, {
+					search,
+					per_page: 10,
+					_fields: 'id,name',
+				} ),
+			} ).then( terms => {
+				if ( ! terms || terms.length === 0 ) {
+					setSuggestions( [] );
+					return;
+				}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 17c36bb6e602d2275b8c88cdab10974f7dc7588e

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
```diff
@@ -0,0 +1,130 @@
...
+			return;
+		}
+		apiFetch< { id: number; name: string }[] >( {
+			path: addQueryArgs( 'wp/v2/' + endpoint, {
+				include: value.join( ',' ),
+			} ),
+		} ).then( terms => {
+			setSavedItems(
+				terms.map( term => ( { value: term.id.toString(), label: decodeEntities( term.name ) || __( '(no name)', 'newspack-plugin' ) } ) )
+			);
```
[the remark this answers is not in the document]
miguelpeixe: Same approach as above. Addressed in f3376f242148e72ae17067384165530d29b25717

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+			// Find items.
+			const foundItems = newTokens.map( t => {
+				if ( typeof t === 'string' ) {
+					const [ val ] = t.split( ':' );
+					return items.find( i => i.value === val );
+				}
+				return items.find( i => i.value === t.value );
+			} );
+			onChange( foundItems.filter( i => i !== undefined )?.map( i => i.value ) );
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 0bbd67d2ccf81453c0934654cd63849671e850e2

miguelpeixe:
Thanks for the thorough review, @rbcorrales! I've addressed the copilot review in the inline comments.

> I can't add a line comment on this one since it's not part of the introduced changes by this PR, but there's a hardcoded post ID here in the wp_get_post_tags function:

Fixed in 3bdbc09df4e8121da270064975618ce48d82bff4

> Not part of the PR, but the access here shouldn't be $gate['id'] instead of $gate->ID?
> 
> Edit: The method doesn't seem to be using the new content_rules meta being saved by this PR (get_post_content_rules()). It reads post_types meta and taxonomy terms directly from the gate post instead. I assume this is intentional/WIP given the related TODO on line 480: // TODO: Content Gate content rules check.

Correct! I've updated it to use the correct data structure to get rid of the warning, but this is still a work in progress. Also added an inline comment so we don't lose track of it. 6f4402f0d72b3395dd30b0636e73cfa1fb186dff

> Another issue unrelated to this PR, but the property here is named $rules, not $registered_rules. This means the duplicate rule check never works.

Fixed in 25d1b8baabee95e1052b24fc03881578890a1a32

src/wizards/audience/views/content-gates/content-gates.tsx
miguelpeixe: Fixed in 24f71ff4f57929ce33b25a9e1a815e6bab5528fc

src/wizards/audience/views/content-gates/content-rule-control-taxonomy.tsx
miguelpeixe: Updated  in 96450d020d9ec674c6a97d6a02c742fd0d134ce0

includes/content-gate/class-content-restriction-control.php
miguelpeixe: Not sure why we have the other one. Removed for now in 4e8a8314da369f3858b1883647440ec6b12150cc

src/content-gate/editor.js
miguelpeixe: This is likely going away soon, but I updated in 54c1aad9ea69f26c7444cdca7ae52b4bd92612c6 just so it's not broken.

includes/wizards/audience/class-audience-content-gates.php
miguelpeixe: Fixed in the other comment by copilot here https://github.com/Automattic/newspack-plugin/pull/4265#discussion_r2569241876

rbcorrales (changes requested):
I'm now getting this error when trying to save, pointing to a recently updated line:
```json
{
    "code": "qm_fatal",
    "message": "Uncaught Error: Cannot access offset of type string on string in /newspack-repos/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php on line 285",
    "data": {
        "message": "Uncaught Error: Cannot access offset of type string on string",
        "file": "/newspack-repos/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php",
        "line": 285,
        "trace": [
            {
                "file": "/newspack-repos/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php",
                "line": 270,
                "function": "sanitize_access_rule",
                "class": "Newspack\\Audience_Content_Gates",
                "type": "->"
            },
            {
                "file": "/newspack-repos/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php",
                "line": 237,
                "function": "sanitize_rules",
                "class": "Newspack\\Audience_Content_Gates",
                "type": "->"
            },
            {
                "file": "/var/www/html/wp-includes/rest-api/class-wp-rest-request.php",
                "line": 850,
                "function": "sanitize_gate",
                "class": "Newspack\\Audience_Content_Gates",
                "type": "->"
            },
...
```

The payload looks like this:
```json
{
    "gate": {
        "id": 1376,
        "status": "publish",
        "title": "Another Gate",
        "description": "",
        "metering": {
            "enabled": true,
            "anonymous_count": "0",
            "registered_count": "0",
            "period": "week"
        },
        "priority": 0,
        "access_rules": [
            "automattic.com",
            {
                "slug": "email_domain",
                "value": "automattic.com"
            }
        ],
        "content_rules": [
            [
                "newspack_collection"
            ],
            {
                "slug": "post_types",
                "value": [
                    "newspack_collection"
                ]
            }
        ]
    }
}
```

miguelpeixe:
Thanks, @rbcorrales!

The sanitization hooks being incorrect (not executed at all) allowed for its logic to have a lot of inconsistencies 😅 

Should be good after 3d8c23c68fb7e0457a1789b7fe43889b72dae541. It's best to test with a fresh gate, as the sanitizer stored an incorrect data structure.

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -144,32 +145,45 @@ public function register_api_endpoints() {
...
 				'methods'             => 'POST',
 				'callback'            => [ $this, 'update_gate' ],
 				'permission_callback' => [ $this, 'api_permissions_check' ],
 				'sanitize_callback'   => [ $this, 'sanitize_gate' ],
 				'args'                => [
 					'gate' => [
-						'type'       => 'object',
-						'properties' => [
+						'type'              => 'object',
+						'sanitize_callback' => [ $this, 'sanitize_gate' ],
```
rbcorrales, line 184: Line 181 still has `'sanitize_callback' => [ $this, 'sanitize_gate' ]` at the route level, which is redundant since the arg-level `sanitize_callback` at line 185 will actually be called. It's harmless but could be removed for cleanliness.

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -258,18 +257,23 @@ public function sanitize_metering( $metering ) {
...
 	 *
 	 * @return array The sanitized access rules.
 	 */
-	public function sanitize_access_rules( $access_rules ) {
-		return array_map( [ $this, 'sanitize_access_rule' ], $access_rules );
+	public function sanitize_rules( $rules, $type = 'access' ) {
+		$sanitized_rules = [];
+		foreach ( $rules as $rule ) {
+			$sanitized_rules[] = $type === 'access' ? $this->sanitize_access_rule( $rule ) : $this->sanitize_content_rule( $rule );
+		}
```
rbcorrales, line 275: If `sanitize_access_rule()` or `sanitize_content_rule()` returns a `WP_Error`, it just gets added to the array. The method should check for errors and return early:
```php
$sanitized = $type === 'access' ? $this->sanitize_access_rule( $rule ) : $this->sanitize_content_rule( $rule );
if ( is_wp_error( $sanitized ) ) {
    return $sanitized;
}
$sanitized_rules[] = $sanitized;
```

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -281,17 +285,51 @@ public function sanitize_access_rule( $access_rule ) {
...
+	 * @return mixed|\WP_Error The sanitized content rule or error if invalid.
+	 */
+	public function sanitize_content_rule( $content_rule ) {
+		$rules = Content_Gate::get_content_rules();
+		if ( ! isset( $rules[ $content_rule['slug'] ] ) ) {
+			return new \WP_Error( 'invalid_content_rule_slug', __( 'Invalid content rule slug.', 'newspack-plugin' ), [ 'status' => 400 ] );
+		}
+
+		$slug = sanitize_text_field( $content_rule['slug'] );
+		if ( ! empty( $content_rule['options'] ) && empty( array_intersect( $content_rule['value'], array_column( $rules[ $slug ]['options'], 'value' ) ) ) ) {
```
rbcorrales: I think there might be two issues here:
1. The condition checks `$content_rule['options']`, but the client never sends `options`, it's server-side metadata. 
2. `array_intersect` only errors when there's no overlap. If you submit `['post', 'invalid_type']`, the intersection with `['post', 'page']` returns `['post']` (not empty), so no error and `invalid_type` will get saved. I tested it in a console via fetch and was able to store that (likely because of issue 1 at this point). Should probably use `array_diff` to catch any invalid value:
```php
$allowed = array_column( $rules[ $slug ]['options'], 'value' );
$invalid = array_diff( $content_rule['value'], $allowed );
if ( ! empty( $invalid ) ) {
    return new \WP_Error( ... );
}
```

rbcorrales (changes requested):

includes/wizards/audience/class-audience-content-gates.php
miguelpeixe: Whoops, good catch. Removed in 02516d27176afda1a02c793f1b468580b9dd14e9

includes/wizards/audience/class-audience-content-gates.php
miguelpeixe: Fixed in 48bab6301a78502c80a3804152f022fe0653fe96 and c8d8ef93af1e8e0b510697077c69418a429e5043. We shouldn't return the error at all in this case, as it'd also be set as the property in the `sanitize_gate()` method.

Added a comment 96a48600081802bf074518b2d93d7f34703270a5 to implement proper error handling, as this is also missing in the frontend.

includes/wizards/audience/class-audience-content-gates.php
miguelpeixe: Thanks for catching that! Fixed the logic in 93f2ee43e790537306e33f27c8fbe4103186822e

rbcorrales (approved):
Thanks @miguelpeixe for addressing all the issues!

miguelpeixe:
Thank you for the thorough review!

PR #4268, opened 2025-10-29
fix: add error when Collection title already exists [NPPM-2315]

laurelfulford:
### All Submissions:

Because Collections uses a shadow taxonomy, there can be some weirdness under the hood when you name two Collections exactly the same name: publishing the first Collection creates the CPT and tag, and then publishing the second on creates the CPT, but "steals" the initial Collection's tag. This causes a weird state where the original CPT is "orphaned" without a tag, and cannot be assigned to posts. 

This PR adds a warning so you can't create new Collections with the same name as existing Collections.

2. Create a new Collection and click 'Publish'; confirm it's created as expected.
3. As a bonus, run `wp post meta get [Collection_CPT_ID] _newspack_collection_term_id` with your Collection's post ID in place of `[Collection_CPT_ID]`, and confirm you get an ID returned.
4. Create a second new Collection with the exact same name as the first; confirm that you get a warning, and the Collection CPT stays in draft mode:

CleanShot 2025-10-29 at 14 03 09

5. Rerun `wp post meta get [Collection_CPT_ID] _newspack_collection_term_id` with the ID of the Collection that failed to publish and confirm it does NOT return an ID.
6. Change the name of your second copycat Collection so it's unique and try to republish; confirm it works this time.
7. Repeat step 5 and confirm this collection now has a `_newspack_collection_term_id` associated with it.
8. Create a post and make sure you can assign both Collections to it.

### Other information:


includes/collections/class-post-type.php
```diff
@@ -186,4 +189,142 @@ public static function output_collection_meta_data_for_admin_scripts( $current_s
...
+	 * @param string $title     The title to check.
+	 * @param int    $exclude_id Post ID to exclude from the check.
+	 * @return bool True if a collection with this title exists, false otherwise.
+	 */
+	public static function title_exists( $title, $exclude_id = 0 ) {
+		$args = [
+			'post_type'      => self::get_post_type(),
+			'post_status'    => [ 'publish', 'private', 'future', 'pending', 'draft' ],
+			'posts_per_page' => -1,
+			'fields'         => 'ids',
```
miguelpeixe, line 204: Can we use the title as `s` argument here? This should also allow us to set a reasonable `post_per_page` so this query is less heavy.

includes/collections/class-post-type.php
```diff
@@ -186,4 +189,142 @@ public static function output_collection_meta_data_for_admin_scripts( $current_s
...
+
+	/**
+	 * Validate collection title to ensure it's unique.
+	 *
+	 * @param array $data    An array of slashed post data.
+	 * @param array $postarr An array of sanitized, but otherwise unmodified post data.
+	 * @param array $unsanitized_postarr An array of unsanitized post data.
+	 * @return array Modified post data.
+	 */
+	public static function validate_collection_title( $data, $postarr, $unsanitized_postarr ) {
```
miguelpeixe: I did not test this for this PR, but I know we used this strategy in other places. We can avoid handling a post meta if this is hooked into [`pre_post_update`](https://developer.wordpress.org/reference/hooks/pre_post_update/) and exited with `wp_die( $message );`.

The message should get rendered in the editor without additional UI or notice management.

miguelpeixe (changes requested):

includes/collections/class-post-type.php
laurelfulford: Oh that's MUCH nicer - thanks @miguelpeixe! Much cleaner, and now it's not creating a draft 🙌

includes/collections/class-post-type.php
laurelfulford: Ah TIL - thanks @miguelpeixe! I ended up setting the `posts_per_page` to 100 since that matches a couple other uses in the Newspack Plugin -- I'm flip flopping between whether that's still excessive, or a nice buffer "just in case". Let me know if that seems like a lot!

includes/collections/class-post-type.php
```diff
@@ -186,4 +187,67 @@ public static function output_collection_meta_data_for_admin_scripts( $current_s
...
+		$title = trim( $data['post_title'] );
+		if ( empty( $title ) ) {
+			return $data;
+		}
+
+		$exclude_id = ! empty( $post_id ) ? $post_id : 0;
+		$title_exists = self::title_exists( $title, $exclude_id );
+
+		if ( $title_exists ) {
+			wp_die( __( 'This collection was not published because a collection with the same title already exists. Please choose a different title and try again.', 'newspack-plugin' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
miguelpeixe:
```suggestion
			wp_die( __( 'This collection could not be saved because a collection with the same title already exists. Please choose a different title and try again.', 'newspack-plugin' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

NIT suggestion, as every save attempt fails, not just publishing.

miguelpeixe (approved):
Thank you for revising! Tests well and looks good.

Small non-blocking suggestion regarding the message:

includes/collections/class-post-type.php
laurelfulford: Ooo, good catch! This has been updated in [4ed68d0](https://github.com/Automattic/newspack-plugin/pull/4268/commits/4ed68d0c695d9ae936e59d6574cd3924d7c93ef9) 🙂

laurelfulford:
Thanks @miguelpeixe!

PR #4271, opened 2025-10-30
feat(subscription-tiers): introduce "Sign In" link

miguelpeixe:
### All Submissions:

NPPD-958

Implements a Sign In button in subscription tiers modals:

image

image

This allows the modal to be aware of the reader's existing subscriptions from the start, properly handling subscription switches.

1. Setup a primary subscription product in Audience -> Subscriptions
2. In a fresh session, go through the `?upgrade-subscription=1` link flow and purchase the subscription with a new email
3. Reset your session, go to the upgrade link again, and Sign In via the modal button
4. Confirm that after signing in, you get the upgrade modal again, but it allows you to upgrade instead of a new purchase
5. Edit a page and add Checkout Button blocks targeting a Grouped Product and a Variable Subscription
6. In a fresh session, go through each checkout button flow and confirm you can sign in and continue the flow

**Note: There's an ongoing discussion in NPPD-959 about re-purchasing a grouped product sibling. This is not covered in this hotfix PR and will be handled separately.**

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -835,6 +835,12 @@ class Subscriptions_Tiers {
 			return;
 		}
 
+		// If coming from a subscription switch link, the reader must be logged in.
+		// The authentication flow will be handled in the frontend.
+		if ( self::should_attempt_to_switch_subscription() && ! is_user_logged_in() ) {
+			return;
+		}
+
 		if ( ! class_exists( '\Newspack_Blocks\Modal_Checkout' ) ) {
 			return;
 		}
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -34,7 +34,7 @@ class Subscriptions_Tiers {
 		add_filter( 'option_woocommerce_subscriptions_order_button_text', [ __CLASS__, 'order_button_text' ], 9 );
 
 		// Primary product rendering.
-		add_action( 'wp_footer', [ __CLASS__, 'print_primary_product_modal' ] );
+		add_action( 'wp_footer', [ __CLASS__, 'print_product_modal' ] );
 		add_filter( 'newspack_popups_assess_has_disabled_popups', [ __CLASS__, 'disable_popups' ] );
 
 		// Unhook Upgrade/Downgrade switch direction text.
@@ -91,6 +91,20 @@ class Subscriptions_Tiers {
 		 * @param string $query_param The URL query parameter.
 		 */
 		return apply_filters( 'newspack_subscriptions_upgrade_subscription_query_param', 'upgrade-subscription' );
+	}
+
+	/**
+	 * Get the URL for triggering the tiers modal for a given product.
+	 *
+	 * @return string The URL for triggering the tiers modal.
+	 */
+	public static function get_tiers_modal_query_param() {
+		/**
+		 * Filters the URL query parameter that triggers the tiers modal.
+		 *
+		 * @param string $query_param The URL query parameter.
+		 */
+		return apply_filters( 'newspack_subscriptions_purchase_product_query_param', 'tiers-modal' );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -651,6 +665,13 @@ public static function render_form( $product = null, $title = null, $button_labe
 
 			<button type="submit" class="newspack-ui__button newspack-ui__button--primary newspack-ui__button--wide"><?php echo esc_html( $button_label ); ?></button>
 			<button type="button" class="newspack-ui__button newspack-ui__button--ghost newspack-ui__button--wide newspack-ui__modal__cancel"><?php _e( 'Cancel', 'newspack-plugin' ); ?></button>
+
+			<?php if ( ! is_user_logged_in() ) : ?>
```
thomasguillot, line 676: Can we move this between the Submit and the Cancel button. And have this button as a secondary button?

Screenshot 2025-10-31 at 14 00 04@2x

thomasguillot (changes requested):

thomasguillot:
Design 👍

dkoo (approved):

miguelpeixe:
Thanks for the review, @dkoo!

4576acb64366b1eb01d00f3e93cf6b629898db7a implements changes that were required to the flow, so readers coming from a subscription switch link must always be signed in. This is to prevent an already registered reader from attempting to anonymously upgrade, which would just purchase a new subscription.  There's more info in NPPD-958.

The link is still useful for regular purchases, so there's no harm in keeping it.

To test these changes:

1. Access the upgrade link anonymously
2. Confirm you're prompted to sign in or register
3. Confirm that each auth action has the proper modal title: "Sign in to upgrade" and "Register to upgrade"
4. Complete the auth flow using a subscriber account and confirm that the modal hangs in the pending state (greyed-out and buttons disabled) until refreshed
5. Confirm the subscription switch modal triggers on page load, and you can complete the switch

That also required some tweaks to the auth modal API, which I go over inline.

src/reader-activation/index.js
```diff
@@ -174,14 +172,7 @@ export function openAuthModal( config = {} ) {
...
 	}
-	if ( readerActivation._openAuthModal ) {
-		readerActivation._openAuthModal( config );
-	} else {
-		console.warn( 'Authentication modal not available' ); // eslint-disable-line no-console
-		if ( config.onError && typeof config.onError === 'function' ) {
-			config.onError();
-		}
-	}
+	_openAuthModal( config );
```
miguelpeixe, line 175: I'm importing the openAuthModal function directly so it's always available on RAS initialization. This is required to trigger the auth modal on RAS init without a domReady race condition.

src/reader-activation-auth/auth-form.js
```diff
@@ -450,9 +453,6 @@ window.newspackRAS.push( function ( readerActivation ) {
 									} else if ( status !== 200 && ! container.config?.closeOnSuccess ) {
 										form.style.opacity = 1;
 									}
-									submitButtons.forEach( button => {
-										button.disabled = false;
-									} );
```
miguelpeixe, line 455: Moved this so it doesn't trigger on every `finally()`. This is to not enable the button in a `skipSuccess: true, skipNewslettersSignup: true, closeOnSuccess: false` scenario and it properly hangs to be handled by the caller callback.

src/reader-activation-auth/auth-modal.js
```diff
@@ -10,6 +11,22 @@ export function getModalContainer() {
...
+ * @return {void}
+ */
+export function onAuthFormReady( callback ) {
+	// If the auth form is already ready, call the callback immediately.
+	if ( document._newspackReaderAuthFormReady ) {
+		callback();
+		return;
+	}
+	document.addEventListener( 'newspack-reader-auth-form-ready', callback );
+}
```
miguelpeixe, line 28: Allows the `openAuthModal()` to be called on RAS init.

dkoo: This is a useful addition!

dkoo (approved):
@miguelpeixe thanks for explaining the latest changes. It works as described 👍

PR #4276, opened 2025-10-31
fix: class to listen to 3rd party newsletters forms

leogermani:
### All Submissions:

Fixes a regression that made the integration with 3rd party forms break.

The `newspack-subscribe-form` is supposed to be a generic class that can be added to any 3rd party form. Once it's submitted, we consider the reader a newsletter subscriber. This is currently not working.

1. On `release`, add a third party form for example using gravity forms or plain HTML. Add the `newspack-subscribe-form` class to the form
2. As an anonymous reader, submit the form and confirm you are not flagged as newsletter subscribe
3. Checkout this branch and repeat the process. Confirm you are flagged.

### Other information:

### DIFF
--- a/src/reader-activation/index.js
+++ b/src/reader-activation/index.js
@@ -393,8 +393,10 @@ function attachAuthCookiesListener() {
  * Set the reader as newsletter subscriber once a newsletter form is submitted.
  */
 function attachNewsletterFormListener() {
-	const newspackForms = [ '.newspack-newsletters-subscribe', '.newspack-subscribe-form' ];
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Approving because code looks fine to me, but testing with GF, I think their forms attach their own submit event handlers which might prevent ours from executing on time or at all. Did we confirm that this fix solves the problem on a live publisher site?

leogermani:
This doesn't work. FIxing the same regression in #4312

PR #4279, opened 2025-11-03
fix: handle store.set on button click and "enter" keyup, too

dkoo:
### All Submissions:

Addendum to #4276 to ensure that our handler is called even if a third-party form has other event handlers attached to the form `submit` event.

Follow instructions in #4276.

### Other information:

### DIFF
--- a/src/reader-activation/index.js
+++ b/src/reader-activation/index.js
@@ -401,6 +401,32 @@ function attachNewsletterFormListener() {
 		if ( ! form ) {
 			return;
 		}
+
+		// Some forms may have JS form submit handlers which may prevent other form submit handlers from being called.
+		// We attach event handlers on submit button clicks and 'Enter' key presses to ensure our handler is called.
+		form.setAttribute( 'data-submitted', 'false' );
+		const submitter = ( evt, button ) => {
+			evt.preventDefault();
+			form.setAttribute( 'data-submitted', 'true' );
+			store.set( 'is_newsletter_subscriber', true );
+			form.requestSubmit( button );
+		};
+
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This works well in my testing! Running `newspackReaderActivation.store.get('is_newsletter_subscriber');` in the console after submitting a GF form with the CSS class returned `true` when submitting the form with the Submit button, and with the 'Enter' key.

src/reader-activation/index.js
```diff
@@ -401,6 +401,32 @@ function attachNewsletterFormListener() {
...
 		}
+
+		// Some forms may have JS form submit handlers which may prevent other form submit handlers from being called.
+		// We attach event handlers on submit button clicks and 'Enter' key presses to ensure our handler is called.
+		form.setAttribute( 'data-submitted', 'false' );
+		const submitter = ( evt, button ) => {
+			evt.preventDefault();
+			form.setAttribute( 'data-submitted', 'true' );
+			store.set( 'is_newsletter_subscriber', true );
+			form.requestSubmit( button );
```
miguelpeixe, line 412: If the 3rd party is not relying on the form submission to run its thing, aren't we potentially introducing issues by doing this? This feels overly intrusive to unknown forms.

Since we are already pretty optimistic here - no form validation to run our callback - I suggest we attach the same form submission callback to the submit button interaction and not trigger anything on the form.

leogermani: yeah, that's another argument that makes me thing we should probably do a dedicated GF integration using https://docs.gravityforms.com/gform-submission/

dkoo: I'm going to close this one, then—if we want to handle GF forms specially using their API, that should be a different approach than the one here.

PR #4286, opened 2025-11-05
feat(variation-modal): display variation description

miguelpeixe:
### All Submissions:

NPPM-2370

Variation attributes are no longer part of the variation modal, which can be problematic for anything outside of frequency-based variations.

This PR implements rendering the variation description in the variation and subscription tiers modals:

image

image

1. Create a variable subscription product in which all variations share the same frequency setup
2. Edit each variation to have descriptions
3. Add a Checkout Button targeting the new product and allowing readers to select the variation
4. Visit the page and start the flow
5. Confirm the description renders for each option

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -391,7 +391,8 @@ class Subscriptions_Tiers {
 			$price = $product->get_price_html();
 		}
 
-		$description = $product->get_description();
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This works as described! 

It seems possible based on the Linear discussion that we may want to hide this behind a toggle (I think it should be okay as is but there definitely could be a use case I'm not thinking of that will come up there!). I think it's okay to merge as is for now, but if you want to wait to see how that goes down I'm happy to take another look at changes here 🙂

miguelpeixe:
Thanks for the review, @laurelfulford!

As discussed internally, 17836d6ef6cf3d087b3089e068eafd4da4a7eafb adds the global opt-out constant

laurelfulford (approved):
Tested adding `define( 'NEWSPACK_DISABLE_SUBSCRIPTION_DESCRIPTION', true );` to wp-config.php, and it successfully hid the description from the variation picker! 🚢

PR #4287, opened 2025-11-05
hotfix: limit free trial purchases to 1 per user

dkoo:
### All Submissions:

Adds filters to limit free trial subscriptions to one per user. Also looks for existing subscriptions with `pending` and `pending-cancel` statuses when determining whether a product with a limit of 1 active subscription per user can be purchased, as these statuses can transition to `active` via user action and are also considered "active" by the Memberships plugin.

Note that these changes have been proposed as an upstream fix in the WooCommerce Subscriptions plugin. If it gets merged and deployed to the Subscriptions plugin, we can deprecate the changes in this plugin.

See also https://github.com/Automattic/newspack-blocks/pull/2245

Closes NPPM-2302.

Check out this PR and https://github.com/Automattic/newspack-blocks/pull/2245. Follow testing instructions in https://github.com/woocommerce/woocommerce-subscriptions/pull/5181.

After the above, also test the following Newspack-specific features:

1. Attach the product to a Checkout Button block and start the transaction from there. When the product purchase is limited because you're logged into a reader account with an active (or `on-hold` or `pending-cancel`) subscription, confirm that the custom error message from https://github.com/Automattic/newspack-blocks/issues/2202 is shown instead of Woo's default "<product> has been removed from your cart because it can no longer be purchased" message.
2. Start a new session as an anonymous reader. Start checkout from the Checkout Button block and enter the email address for the same test reader account you've been using to test. Confirm that you're unable to complete checkout and that the custom error message is shown:

Screenshot 2025-11-05 at 3 24 10 PM

3. As an admin, remove the setting to limit the product to 1 active subscription per user.
4. Log in as a reader who has a subscription of any status with the free product trial and start a transaction from the Checkout Button.
5. Confirm that you're able to complete the purchase at full price in the modal checkout, but that the product summary at the top of the modal and the transaction details before the final submit button both don't mention any free trial.

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
+++ b/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
@@ -18,6 +18,7 @@ class WooCommerce_Subscriptions {
 	 */
 	public static function init() {
 		add_action( 'plugins_loaded', [ __CLASS__, 'woocommerce_subscriptions_integration_init' ] );
+		add_filter( 'woocommerce_subscriptions_product_limited_for_user', [ __CLASS__, 'maybe_limit_subscription_product_for_user' ], 10, 3 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Works as described!

I tested using a variable subscription and was able to "abuse" different frequencies, as they are different products. I'm considering this non-blocking for this hotfix, but we should check whether cross-variant free trials is a common thing

dkoo:
True, I think eventually this should be fixed at a deeper level in the Woo plugin

PR #4288, opened 2025-11-05
fix(stripe-gateway): disable Link by default on new installs

dkoo:
### All Submissions:

This was already done in #3918, but it looks like the sub-option name for this in the Stripe Gateway plugin has changed somewhere along the way. This PR handles the new option name as well as the old.

Closes NPPM-2295.

Make sure you're using the [latest public release of the Stripe Gateway plugin](https://github.com/woocommerce/woocommerce-gateway-stripe/releases/latest), and follow testing instructions from #3918.

### Other information:

### DIFF
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -169,17 +169,12 @@ class WooCommerce_Gateway_Stripe {
 		}
 
 		// Disable Link by Stripe for new installs.
-		if (
-			is_array( $settings['upe_checkout_experience_accepted_payments'] ) &&
-			! empty( $settings['upe_checkout_experience_accepted_payments'] ) &&
-			in_array( 'link', $settings['upe_checkout_experience_accepted_payments'], true )
-		) {
-			$settings['upe_checkout_experience_accepted_payments'] = array_diff(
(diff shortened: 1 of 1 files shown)


laurelfulford (changes requested):
This one is weird - it works fine when you install WooCommerce/Stripe directly from the plugins screen, but it also works there on trunk 😕 If you install set up Stripe as part of the Audience Development/Donation wizard set up, Link is still selected on trunk and this branch.

I replicated this a couple time on fresh JN sites.

dkoo:
@laurelfulford you're right, and it took a bit of digging to figure out why that is. It looks like since the Stripe Gateway plugin changed their onboarding flow, Link is now enabled by default when you first connect the plugin to a Stripe account, not when you first install the plugin. So after connecting from this screen:

Screenshot 2025-11-06 at 4 42 22 PM

8ba072301 should fix this—now, after you connect to Stripe, the Link and Apple/Google Pay options should be disabled by default. You should also be able to manually enable them after this.

laurelfulford (approved):
This looks good to me, @dkoo! Confirmed that it now works when enabling Stripe from the Audience Management wizard, and also still from activating the plugin manually.

Just in case, I also tried installing trunk, saving Link as an option, and confirmed upgrading to this branch didn't affect it.

PR #4293, opened 2025-11-07
fix: move some Collection margin styles from the block to theme

laurelfulford:
### All Submissions:

Along with https://github.com/Automattic/newspack-theme/pull/2581, this PR makes some Collection block styles a bit more specific:

When on the archive page, the grid view of the blocks need a top margin of `32px`. And when on the single Collection view, the blocks need a bottom margin of `32px`. However, having these margins applied to the blocks themselves means we can't use tricks wrapping the block in a Stack block with other blocks to reduce the spacing between them (a trick we use a lot with Headings + other blocks).

These PRs move the general top and bottom margins from the block styles to the theme styles and use the template classes to make them more specific.

1. Make sure you have a few Collections set up on your site. 
2. Apply this PR and https://github.com/Automattic/newspack-theme/pull/2581.
3. Add two Collections Blocks to a page, one using the Grid layout and one using the List layout; publish. 
4. Possibly not the best way to test this, but on the front-end, use the element inspector and confirm there's a `margin: 32px 0` coming from the theme on each block, but no other top/bottom margin. You can confirm that by "turning off" the theme-based margin and seeing that the blocks don't have it anymore.
5. View the Collections archive landing page; inspect the grid of Collections and confirm that it has a top margin of `32px` coming from the theme and targeting the `.layout-grid` class only when inside the `.post-type-archive-newspack_collection` body class.
6. View an individual Collection and inspect the Collection at the top of the page; confirm that it has a bottom margin of `32px` coming from the theme only when it's inside the `.single-newspack_collection` class.

### Other information:

### DIFF
--- a/src/blocks/collections/styles/_layout-list.scss
+++ b/src/blocks/collections/styles/_layout-list.scss
@@ -175,6 +175,10 @@
 	}
 
 	.wp-block-newspack-collections.layout-list {
-		margin-bottom: 64px;
+		margin-bottom: 32px;
+
+		@media #{bp.$media-lg-up} {
+			margin-bottom: 64px;
+		}
 	}
 }
--- a/src/blocks/collections/styles/_layout-list.scss
+++ b/src/blocks/collections/styles/_layout-list.scss
@@ -3,8 +3,6 @@
 .wp-block-newspack-collections {
 	// Layout: List.
 	&.layout-list {
-		margin-bottom: 32px;
-
 		.wp-block-newspack-collections__item {
 			display: grid;
 			grid-template-columns: 1fr;
--- a/src/blocks/collections/styles/_layout-list.scss
+++ b/src/blocks/collections/styles/_layout-list.scss
@@ -167,3 +167,10 @@
 		}
 	}
 }
+
+/* Single Collection styles */
+.single-newspack_collection {
+	.wp-block-newspack-collections.layout-list {
+		margin-bottom: 32px;
+	}
+}
--- a/src/blocks/collections/styles/_layout-list.scss
+++ b/src/blocks/collections/styles/_layout-list.scss
@@ -170,7 +170,11 @@
 
 /* Single Collection styles */
 .single-newspack_collection {
+	&.has-sidebar .site-main {
+		display: block;
+	}
+
 	.wp-block-newspack-collections.layout-list {
-		margin-bottom: 32px;
+		margin-bottom: 64px;
 	}
 }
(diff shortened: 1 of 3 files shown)


rbcorrales (changes requested):
The recent collections grid section at the bottom of the single template has some negative margins (I know, sorry!), which is causing this:
image

Also, minor nitpick, but the top section of the archive template (the list layout) used to have this margin:
image

But it's now closer to the hr:
image

laurelfulford:
Thanks for the review @rbcorrales! 

I added a longer comment to this effect to the theme PR, but I think we may need to move the Collections template to the Classic theme as part of the Block theme work (since the template markup won't be the same) but there's no need to do that yet! 

I moved the theme styles over to the plugin, and tweaked the negative margin to match the previous appearance. I wasn't able to recreate the issue with the list view on the single archive landing page, though - the spacing looks the same to me unless I remove the styles from the theme PR 😅 

Hopefully things look right, but just let me know if either issue persists for you!

src/blocks/collections/styles/_layout-list.scss
```diff
@@ -3,8 +3,6 @@
 .wp-block-newspack-collections {
 	// Layout: List.
 	&.layout-list {
-		margin-bottom: 32px;
-
```
rbcorrales, line 7: What I meant is that the removal of the margin here is making the highlighted collection in the archive page closer to the next `<hr>` than it was before:
image
This could be solved by just adding the margin to the template CSS selectors.

rbcorrales (approved):
Thanks for addressing the issues and all the explanations!

The negative-margin issue is resolved for me now, but the margin in the highlighted collection of the archive page is still missing. This is more like a nitpick, so I'm approving the PR, but providing a better description of the issue with a screenshot in the comments, in case you want to address it 😄 

Thanks!

src/blocks/collections/styles/_layout-list.scss
laurelfulford: Ahhhh I (finally!) see what's happening!! 

There's a CSS class getting added to the body when the sidebar widget area is populated that's making the `.site-main` use `display: flex` -- when this happens, the margin on the main Collection and separator don't overlap so the gap looks the same as it did before. I have sidebar widgets on my test site - when I remove them, I get the second screenshot 😄 

Thank god, I could not figure out why our sites were different with this PR -- fixing!

laurelfulford: I ended up overriding the `display: flex` in the single collection template styles (to make it easier to make this consistent), and tweaking the margin so it matches trunk: 64px on bigger screens, 32px on smaller screens 🙂

PR #4305, opened 2025-11-12
feat(contrib-meter): add contribution meter block

rbcorrales:
### All Submissions:

Adds a new block for displaying fundraising progress toward a goal, with data sourced from WooCommerce donation orders.

image

**What it does**
- Tracks contributions (donations only) from a configurable start date.
- Shows goal amount, amount raised, and completion percentage. The visibility of these values is configurable.
- Available in linear (progress bar) and circular (ring) styles.
- Amount raised is updated via a cached REST API endpoint. Cache duration can be overridden via a DB option or a filter.

**Configuration**
- Set fundraising goal amount.
- Choose start date for tracking contributions.
- Toggle display of goal, amount raised, and percentage.
- Customize progress bar style, color, and thickness.

Closes NPPD-944.

1. Add a new "Contribution Meter" block to a post or page.
2. Set a goal amount and start date in the block inspector:
image

3. Toggle visibility of goal, amount raised, and percentage to confirm they show/hide correctly. 
4. Switch between linear and circular meter styles, and adjust thickness settings.
5. Set a custom progress bar color and verify it displays correctly.

image

6. Test that all the possible settings permutations match the designs: 

image

7. Create test donation orders via WooCommerce and verify the meter reflects the correct revenue totals.
8. Change the start date and confirm that only donations after that date are counted.
9. Check that the data caches properly (subsequent loads don't re-query orders) and refreshes after the configured cache duration.
10. Apply wide and full alignment to the block to ensure proper display.
11. Check that the meters look the same in the editor and the frontend.

### Other information:

### DIFF
--- a/includes/contribution-meter/class-contribution-meter.php
+++ b/includes/contribution-meter/class-contribution-meter.php
@@ -199,13 +199,14 @@ class Contribution_Meter {
 			$user_end_date_obj = new \DateTime( $end_date, wp_timezone() );
 			$end_date          = ( $user_end_date_obj < $yesterday ) ? $end_date : $yesterday->format( 'Y-m-d' );
 		} else {
-			// No user end date - apply the default max range.
+			// No user end date - apply the default max range from start date.
 			$max_end_date_range = get_option( self::MAX_END_DATE_RANGE_OPTION, self::DEFAULT_MAX_END_DATE_RANGE );
 
-			// Calculate maximum allowed end date from today + max range (e.g., today + 6 months).
-			$max_allowed_end_date = new \DateTime( $max_end_date_range, wp_timezone() );
+			// Calculate maximum allowed end date from start date + max range (e.g., start date + 6 months).
+			$max_allowed_end_date = new \DateTime( $start_date, wp_timezone() );
+			$max_allowed_end_date->modify( $max_end_date_range );
 
-			// Use the minimum of (yesterday, maximum allowed end date).
+			// Use the minimum between yesterday and the maximum allowed end date.
 			$end_date = ( $max_allowed_end_date < $yesterday ) ? $max_allowed_end_date->format( 'Y-m-d' ) : $yesterday->format( 'Y-m-d' );
 		}
 
--- a/includes/contribution-meter/class-contribution-meter.php
+++ b/includes/contribution-meter/class-contribution-meter.php
@@ -22,6 +22,17 @@ class Contribution_Meter {
 	const REST_ROUTE = '/contribution-meter';
 
 	/**
+	 * Default oldest allowed start date range (relative to today).
+	 * E.g., '-2 months' means start date cannot be earlier than 2 months ago.
+	 */
+	const DEFAULT_START_DATE_RANGE = '-2 months';
+
+	/**
+	 * Start date range option name.
+	 */
+	const START_DATE_RANGE_OPTION = 'newspack_contribution_meter_start_date_range';
+
+	/**
 	 * Default data collection end date range (relative to today).
 	 * E.g., '-1 day' means collect up to yesterday, excluding today.
 	 */
@@ -31,6 +42,16 @@ class Contribution_Meter {
 	 * End date range option name.
 	 */
 	const END_DATE_RANGE_OPTION = 'newspack_contribution_meter_end_date_range';
+
+	/**
+	 * Default maximum allowed end date range (relative to today).
+	 */
+	const DEFAULT_MAX_END_DATE_RANGE = '+6 months';
+
+	/**
+	 * Maximum end date range option name.
+	 */
+	const MAX_END_DATE_RANGE_OPTION = 'newspack_contribution_meter_max_end_date_range';
 
 	/**
 	 * Cache key prefix for contribution meter data.
@@ -63,6 +84,11 @@ class Contribution_Meter {
 						'type'              => 'string',
 						'validate_callback' => [ __CLASS__, 'validate_date' ],
 					],
+					'endDate'   => [
+						'required'          => false,
+						'type'              => 'string',
+						'validate_callback' => [ __CLASS__, 'validate_date' ],
+					],
 				],
 			]
 		);
@@ -75,6 +101,11 @@ class Contribution_Meter {
 	 * @return bool|\WP_Error True if valid, WP_Error if invalid.
 	 */
 	public static function validate_date( $date ) {
+		// Allow empty string for optional dates.
+		if ( empty( $date ) ) {
+			return true;
+		}
+
 		// Validate basic date format.
 		if ( ! preg_match( '/^\d{4}-\d{2}-\d{2}$/', $date ) ) {
 			return new \WP_Error( 'invalid_date', __( 'Invalid date format. Expected YYYY-MM-DD.', 'newspack-plugin' ) );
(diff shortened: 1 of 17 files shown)


includes/contribution-meter/class-contribution-meter.php
```diff
@@ -0,0 +1,255 @@
...
+			return new \WP_Error( 'invalid_date', __( 'Invalid date. Please provide a valid date.', 'newspack-plugin' ) );
+		}
+		return true;
+	}
+
+	/**
+	 * REST API callback to get contribution data.
+	 *
+	 * @param WP_REST_Request $request Request object.
+	 * @return WP_REST_Response|WP_Error Response object or error.
```
miguelpeixe: Nit: across this file my IDE is yelling at me that `Newspack\Contribution_Meter\WP_...` does not exist.

Can we either add the root namespace in the comments or switch to `use WP_...` at the top? No strong preference on which.

includes/contribution-meter/class-contribution-meter.php
```diff
@@ -0,0 +1,255 @@
...
+		return rest_ensure_response( $data );
+	}
+
+	/**
+	 * Get contribution data with caching.
+	 *
+	 * @param string $start_date Valid start date in YYYY-MM-DD format.
+	 * @return array|WP_Error Array of contribution data or WP_Error on failure.
+	 */
+	public static function get_contribution_data( $start_date ) {
```
miguelpeixe: Not having a limit on how far back the start date can go can be problematic, even with the 10-minute transient cache. WDYT we set a 2-month limit?

This should be set both here and in the block editor. If a "giving season" page is created and forgotten, it risks querying years' worth of data when visited.

thomasguillot:
@rbcorrales I modified the styles so it uses Newspack UI now

thomasguillot:
Screenshot 2025-11-13 at 15 04 29@2x
Screenshot 2025-11-13 at 15 04 23@2x

How come we're not using Core's Styles style?

miguelpeixe (changes requested):
Tested well and looks great! Just a couple of comments inline:

includes/contribution-meter/class-contribution-meter.php
rbcorrales: Yeah, that's a good idea @miguelpeixe. TBH, I'm still wondering whether we should even go with a manual DB query instead of using WC methods to reduce the possible overhead of these classes when there are hundreds (thousands?) of orders to go through.

Also, I arbitrarily set the default cache time to 10 minutes, but do you think we should increase it a bit more? Maybe half an hour?

miguelpeixe: I think even 24 hours and collecting up to the day prior is not bad.

A more advanced approach would be to store daily amounts so it's never re-querying fixed data, but that's probably overkill.

rbcorrales: Implemented the date restrictions here: [27e16fe](https://github.com/Automattic/newspack-plugin/pull/4305/commits/27e16fe707501588aef66249a24e194009189599)

~~It can be overridden by a `newspack_contribution_meter_max_date_range` (and consequentially, via a `pre_option_newspack_contribution_meter_max_date_range` filter).~~

Edit: Option name is now `newspack_contribution_meter_start_date_range`.

includes/contribution-meter/class-contribution-meter.php
rbcorrales: Thanks for noticing this. Fixed here: [35a6fe4](https://github.com/Automattic/newspack-plugin/pull/4305/commits/35a6fe42211f79950823234dc6b45e6c6802b0d2)
Let me know if I missed one.

includes/contribution-meter/class-contribution-meter.php
miguelpeixe: Thanks for revising!

We're still allowing a stored `start_date` to go unlimited for saved blocks, should we leave it?

thomasguillot:
Made some tweaks to the `Placeholder`

thomasguillot:
Approved design ahead of https://github.com/Automattic/newspack-plugin/pull/4305#issuecomment-3528228024

includes/contribution-meter/class-contribution-meter.php
rbcorrales: No, after pushing that change, I realized that when you save a block and the days pass, you'll get an error if the start day is more than 2 months ago relative to the current date under the current implementation. I need to find a way to avoid this. I'm also working on caching daily data, so I'll push something together with this.

There's also the case you mentioned about requesting old pages with that block after a few years. Do you think we should have like an end date for the goal? cc @thomasguillot

rbcorrales: Ok, updated the caching logic to store the totals up to yesterday on a rolling basis.
But also removed the start date validation from the backend, as this would cause problems when reopening pages with this block over time. Keeping it on the frontend only.
[a2cfee5](https://github.com/Automattic/newspack-plugin/pull/4305/commits/a2cfee54f698e63d29158cc54d7123ba5afd1ee4)

thomasguillot: > There's also the case you mentioned about requesting old pages with that block after a few years. Do you think we should have like an end date for the goal?

That's a good question. Maybe optional? (might be worth asking Katie's opinion on this one)
ToggleControl: "Add an end date", help text: "Enable this if the contribution meter should stop counting contributions after a specific date."
Date: "End date", help text: "Contributions up to and including this date are counted in the total amount raised."

miguelpeixe: I like that. Let's check with @kmwilkerson

rbcorrales:
> How come we're not using Core's Styles style?

@thomasguillot the challenge is that linear vs circular are structural differences (different components/markup), not just CSS variations. If using core styles, we'll need to detect which style is selected by checking the class names (`is-style-circular`, `is-style-linear`) in both the editor and server-side rendering, or render the two components and hide one, but this is inefficient, unless you know a different approach.

Let me give it a shot to see how it looks with class detections.

includes/contribution-meter/class-contribution-meter.php
rbcorrales: The issue is that if the approach is optional, it won't be effective for our "open this page from 5 years ago" scenario. The block would still attempt to query and sum the orders data from the start date all the way to that day, 5 years later. Unless we address this with a strict rule to limit the goal counting to a specific timeframe in the code. For example, we could establish a guideline stating that "we will count donations from the start date up to a maximum of 6 months later."

miguelpeixe: > For example, we could establish a guideline stating that "we will count donations from the start date up to a maximum of 6 months later."

I vote we do this for now so we're not blocked by this

kmwilkerson: Can we do both?
1. Add an optional “End date” setting in the block
2. But as a safety in case it's not input, enforce a backend maximum query window (I like 6 months)

rbcorrales:
@thomasguillot ICYMI, yesterday I pushed [8b3bce2](https://github.com/Automattic/newspack-plugin/pull/4305/commits/8b3bce2d563dfb49dc8c73eeebcd64bba3d09053) for using block styles and then [e365c1b](https://github.com/Automattic/newspack-plugin/pull/4305/commits/e365c1b5b71b89840e39c4dd1116f8bf29723490) for handling a "preview mode" that wouldn't perform a REST request for pulling contribution data. Otherwise, the previews would take a while to load every time.

miguelpeixe:
@rbcorrales let me know when this is ready for re-review

thomasguillot:
@rbcorrales Yes that's great, thank you very much

includes/contribution-meter/class-contribution-meter.php
rbcorrales: Implemented that here: [1d07f3c](https://github.com/Automattic/newspack-plugin/pull/4305/commits/1d07f3c7d3d2c5a89fd552fb4913cb69c216023f)

rbcorrales:
Hey @miguelpeixe I implemented the end date functionality, and this is ready for another review, please!

miguelpeixe (approved):
Thank you for the revisions! Looks good

PR #4309, opened 2025-11-14
fix(subscription-modal): limited subscriptions

miguelpeixe:
### All Submissions:

NPPD-959

The subscription tier modal is currently restricting the purchase of any subscription that the reader already owns. This PR changes that behavior to check if the subscription is limited.

The warning notice should still render if not limited.

| Limited | Not Limited |
| --- | --- |
| image | image |

The subscription limit will also be automatically applied in the context of a grouped product.

1. If you don't have one, create a variable subscription product without any purchase limit
2. Add this product a grouped product with another subscription product
3. Add a Checkout Button block to a page targeting the product (allow readers to select the variation) and the grouped product
4. As a reader, purchase the product
5. Click the checkout button to purchase the product again, and confirm that you are presented with a warning notice, but can move forward
6. Click the grouped product checkout button and confirm the behavior is the same
7. Edit the product to limit the purchase:

image

8. As the reader, start the checkout flow and confirm that you get the same warning, but now with a button to view your active subscription
9. Do the same with the grouped product checkout button and confirm the behavior is the same

### Other information:


dkoo (approved):
This is working! Maybe we should also align the "subscription limit" message [in the Blocks repo](https://github.com/Automattic/newspack-blocks/blob/trunk/includes/class-modal-checkout.php#L1511) with same message and link to view/manage the subscription? This message is the one shown when the Checkout Button block is tied directly to a limited variation or simple product.

Screenshot 2025-11-25 at 11 13 57 AM

miguelpeixe:
Good call! I'll set a PR up soon

Thanks for the review

PR #4312, opened 2025-11-18
feat: listen to gforms events for newsletters

leogermani:
### All Submissions:

Add a specific listener for Gravity Forms forms.

Apparently, our generic solution to use the `newspack-subscribe-form` css class used to work, but it hasn't been working for a while. There was a regression added a while ago, which we tried to fix in https://github.com/Automattic/newspack-plugin/pull/4276, but with no success. Showing that this has been broken for a long time.

This PR uses GForms JS api to add a specific listener.

1. Install and activate Gravity Forms
2. Create a new form
3. In the form settings, add the `newspack-subscribe-form` CSS class to the form
4. Add the form to a page
5. As an anonymous reader in a new session, visit that page
6. Submit the form
7. Inspect the Browser Local storage and confirm you now see a key `np_reader_1_is_newsletter_subscriber` with `true` as the value.

### Other information:

### DIFF
--- a/src/reader-activation/index.js
+++ b/src/reader-activation/index.js
@@ -1,4 +1,4 @@
-/* globals newspack_ras_config, newspack_reader_data */
+/* globals newspack_ras_config, newspack_reader_data, gform */
 
 window.newspack_ras_config = window.newspack_ras_config || {};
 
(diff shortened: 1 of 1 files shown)


src/reader-activation/index.js
```diff
@@ -397,9 +399,35 @@ function attachNewsletterFormListener() {
...
+
+	// First-party forms success event listener.
 	document.querySelectorAll( newspackForms.join( ',' ) ).forEach( el => attachHandler( el, 'newspack-newsletters-subscribe-success' ) );
+
+	// Gravity forms handlers.
+	document.addEventListener( 'gform/post_render', event => {
+		if ( gformIds.includes( event.detail.formId ) ) {
+			if ( ! window.gform ) {
+				return;
+			}
```
dkoo: Nit: this is a bit more terse and also guards against unhandled errors in case of nonexistent nested properties:

```suggestion
		if ( window.gform?.utils?.addAsyncFilter && gformIds.includes( event.detail?.formId ) ) {
```

dkoo (approved):
Works! Left a non-blocking suggestion, but looks good otherwise.

PR #4319, opened 2025-11-20
fix: repeat button block styles in collections for WP 6.9

laurelfulford:
### All Submissions:

WordPress 6.9 includes [some nice front-end performance improvements](https://make.wordpress.org/core/2025/11/18/wordpress-6-9-frontend-performance-field-guide/), including loading block styles on demand in classic themes like ours. 

Unfortunately, this is causing some weirdness where we rely on these styles loading - for example, in the Collections block we're counting on some styles from `wp-block-button__link`, but they're only loaded for the block when there's a button block on the page, and they're not loaded at all on the /collections archive pages. It's minor, but this leads to the button text not being centred.

This PR adds some duplicated styles to help fix. 

Edit to add: this seems to be the one spot where we're using this CSS class on a non `<button>` button and a non-WP Button block -- so hopefully this is the only case of it!

1. On a test site running the latest WP 6.9 RC, make sure you have some collections with CTAs.
2. View the /collections landing page on the front end - note the text alignment in your CTAs:

CleanShot 2025-11-19 at 17 18 23

4. Confirm your CTAs now look okay:

CleanShot 2025-11-19 at 17 17 32

### Other information:

### DIFF
--- a/src/blocks/collections/styles/_ctas.scss
+++ b/src/blocks/collections/styles/_ctas.scss
@@ -6,6 +6,9 @@
 		gap: 8px;
 
 		a {
+			align-content: center;
+			display: flex;
+			justify-content: center;
 			width: 100%;
 		}
 	}


rbcorrales (approved):
Thanks for fixing this! Seems that it was caused by my non-standard way of using that button class here 😅

laurelfulford:
Thanks for the review @rbcorrales!

> Seems that it was caused by my non-standard way of using that button class here 😅

We do this a LOT and it's the right way to use it, at least with how we've been styling things 🙂 It means these buttons will pick up other button-y styles from the theme, the Customizer options, and Custom CSS without us having to change anything specifically for them.

It's only an issue with this block because it's the one case we've added it on an `<a>` tag rather than a `<button>` (so far, anyway!). The default `<button>` styles coming from the browser makes it so this change in WP 6.9 doesn't change how they look.

PR #4320, opened 2025-11-20
fix(contrib-meter): ensure percentage follows bar progress when using linear style

thomasguillot:
### All Submissions:

I missed something in the original PR review. When the meter is in its linear style and only the percentage is shown, the percentage text always sits in the centre. It should follow the position of the filled bar instead. I hadn’t represented that properly in Figma/Vercel.

This PR fixes that, so the percentage now aligns with the progress of the bar.

cc @rbcorrales 

__Before:__

__After:__

1. Add multiple linear meter on a page, with different target values enable just the %.
2. Notice no matter what, the % is centered.
3. Switch to this branch and refresh – it should now be level with the progress bar
4. Test other settings (show goals etc...) to make sure I didn't break the other combinations

### Other information:

### DIFF
--- a/src/blocks/contribution-meter/components/LinearMeter.jsx
+++ b/src/blocks/contribution-meter/components/LinearMeter.jsx
@@ -35,16 +35,16 @@ const LinearMeter = ( { amountRaised, goal, percentage, showGoal, showAmountRais
 
 	// Determine text alignment class based on what's displayed.
 	let textAlignClass = '';
-	if ( showPercentage && ! showGoal && ! showAmountRaised ) {
-		textAlignClass = 'contribution-meter__text--center';
-	} else if ( ! showAmountRaised && showGoal && ! showPercentage ) {
+	if ( ( showPercentage && ! showGoal && ! showAmountRaised ) || ( ! showAmountRaised && showGoal && ! showPercentage ) ) {
 		textAlignClass = 'contribution-meter__text--right';
 	}
+
+	const textStyle = showPercentage && ! showGoal && ! showAmountRaised ? { width: progressStyle.width } : undefined;
 
 	return (
 		<div className="contribution-meter__linear">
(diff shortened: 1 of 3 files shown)


src/blocks/contribution-meter/components/LinearMeter.jsx
```diff
@@ -35,16 +35,16 @@ const LinearMeter = ( { amountRaised, goal, percentage, showGoal, showAmountRais
...
 	// Determine text alignment class based on what's displayed.
 	let textAlignClass = '';
-	if ( showPercentage && ! showGoal && ! showAmountRaised ) {
-		textAlignClass = 'contribution-meter__text--center';
-	} else if ( ! showAmountRaised && showGoal && ! showPercentage ) {
+	if ( ( showPercentage && ! showGoal && ! showAmountRaised ) || ( ! showAmountRaised && showGoal && ! showPercentage ) ) {
 		textAlignClass = 'contribution-meter__text--right';
 	}
 
+	const textStyle = showPercentage && ! showGoal && ! showAmountRaised ? { width: progressStyle.width } : undefined;
```
rbcorrales, line 42: You might need to add a `min-width` to the style to prevent the percentage from getting outside the container with widths close to 0%:
image
Another option would be to align it to the left without setting a width if the progress is < than a certain percent.

rbcorrales (changes requested):

src/blocks/contribution-meter/components/LinearMeter.jsx
thomasguillot: https://github.com/Automattic/newspack-plugin/pull/4320/commits/3ddb6f68fecbf71af9b37579e82199b15401cde9 should do the trick

rbcorrales (approved):

PR #4326, opened 2025-11-26
fix(my-account): safe content argument to skip sanitization

miguelpeixe:
### All Submissions:

NPPD-989

Modal content from Woo template tags may include markup from hooks. This markup can be script and style tags, which should be allowed.

This PR introduces a `content_is_safe` flag so the modal content renders without sanitization.

1. While on trunk, make sure you have My Account v1 enabled
2. Configure and enable reCAPTCHA (Newspack -> Settings -> Connections)
3. As a reader, add a new payment method (My Account -> Payment information)
4. Confirm the reCAPTCHA script tag is sanitized and its contents printed:

image

5. Checkout this branch, refresh the page, and confirm the script is preserved in the markup
 

### Other information:

### DIFF
--- a/includes/class-newspack-ui.php
+++ b/includes/class-newspack-ui.php
@@ -198,6 +198,7 @@ class Newspack_UI {
 	 *     @type string $id The modal ID.
 	 *     @type string $title The modal title.
 	 *     @type string $content The modal content HTML.
+	 *     @type bool   $content_is_safe Whether the content is already safe HTML.
 	 *     @type string $footer The modal footer HTML.
 	 *     @type string $form The form method to use. If given, modal content and action buttons will be wrapped in a form element.
 	 *     @type array $actions {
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works! Noting that the bug only happens with reCAPTCHA v3 enabled, not v2.

dkoo:
Also noting that this only happens because we're [directly injecting](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/class-recaptcha.php#L471) the reCAPTCHA v3 JS via the `woocommerce_add_payment_method_form_bottom` hook. If we were to change this approach to enqueue the code instead, would that sidestep the issue and let us continue sanitizing the output?

miguelpeixe:
Thanks for the review, @dkoo!

> If we were to change this approach to enqueue the code instead, would that sidestep the issue and let us continue sanitizing the output?

The Braintree payment gateway also adds markup. I don't think sanitizing this content is a good idea. Action hooks are commonly used by 3rd parties to print additional markup, and we shouldn't interfere with that.

As long as we control which content is safe for print, we should be good. Note that `woocommerce_account_edit_address()` and `woocommerce_account_add_payment_method()` don't usually go through a sanitizer.

PR #4333, opened 2025-12-01
feat(payment-notice): detect equivalent subscription

miguelpeixe:
### All Submissions:

NPPD-1006

The "update payment" notice should not render if another equivalent subscription is active. This PR implements this strategy by detecting active subscriptions within the variable subscription or grouped product.

1. Create 3 variable subscription products and a grouped product, grouping only 2 of the variable subscription products
2. Create 2 checkout button blocks, targeting the grouped product and the ungrouped variable subscription product
3. As a reader, subscribe via both buttons
4. As an admin, generate a pending renewal order for each subscription:

image

5. Back as the reader, navigate to My Account and confirm you get both notices
6. Go through the checkout button flows again and make sure to select a different subscription product for the grouped product flow and a different variant for the variable subscription flow
7. Now that you have an equivalent subscription for both types of product setup, go to the My Account page
8. Confirm both notices are gone

### Other information:

### DIFF
--- a/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
+++ b/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
@@ -110,7 +110,7 @@ class WooCommerce_Update_Payment_Notice {
 	 * @return string[] The notices.
 	 */
 	private static function get_notices() {
-		if ( ! function_exists( 'wcs_get_subscriptions' ) ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described! 

One thing I noticed not necessarily related to these changes, but the snackbar notices that show these notices outside of My Account are appearing inside the modal checkout—we should probably not be showing these inside the modal as they already appear in the "parent" page:

Image

miguelpeixe:
Thanks, @dkoo! Filed NPPD-1054 for this

PR #4329, opened 2025-11-27
fix: check for gift memberships on content gate and sync - NPPM-2420

leogermani:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes the logic that checks for active subscriptions and account for gifted subscriptions. For the purposes of this method, it should only return true if checking for the gitf recipient.

This method is used for content gate/access control checks and to fetch the data to sync to the ESP.

### How to test the changes in this Pull Request:

1. Set up a giftable subscription
2. Set up a membership plan that restricts content
3. Set up the Newspack Content gate
4. As an anonymous reader, buy a subscription as gift to someone else.
5. In another browser, claim the gift and log in with the gift recipient
6. Confirm that the gifter does not have access to restricted content
7. Confirm that the recipient is granted access to the restricted content
8. Make sure you have `define('NEWSPACK_LOG_LEVEL', 3);` on wp-config
9. As an admin, go to Users and find the gifter and the recipient
10. Click "Resync contact to ESP" and watch the logs
11. Confirm the recipient has the information about the subscription
12. Confirm the gifter does NOT have the info about the subscription

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #4334, opened 2025-12-01
Hotfix/active subscription checks

leogermani:
### All Submissions:

https://github.com/Automattic/newspack-plugin/pull/4329 as a hotifix. Plus removing `pending` as  a subscription status that's considered active. It's not an active status. (NPPM-2396)

Fixes the logic that checks for active subscriptions and account for gifted subscriptions. For the purposes of this method, it should only return true if checking for the gitf recipient.

This method is used for content gate/access control checks and to fetch the data to sync to the ESP.

1. Set up a giftable subscription
2. Set up a membership plan that restricts content
3. Set up the Newspack Content gate
4. As an anonymous reader, buy a subscription as gift to someone else.
5. In another browser, claim the gift and log in with the gift recipient
6. Confirm that the gifter does not have access to restricted content
7. Confirm that the recipient is granted access to the restricted content
8. Make sure you have `define('NEWSPACK_LOG_LEVEL', 3);` on wp-config
9. As an admin, go to Users and find the gifter and the recipient
10. Click "Resync contact to ESP" and watch the logs
11. Confirm the recipient has the information about the subscription
12. Confirm the gifter does NOT have the info about the subscription

### Other information:

### DIFF
--- a/includes/data-events/connectors/class-esp-connector.php
+++ b/includes/data-events/connectors/class-esp-connector.php
@@ -11,6 +11,7 @@ use Newspack\Data_Events;
 use Newspack\Reader_Activation;
 use Newspack\Reader_Activation\Sync;
 use Newspack_Newsletters_Contacts;
+use Newspack\WooCommerce_Connection;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -118,7 +119,7 @@ class ESP_Connector extends Reader_Activation\ESP_Sync {
 		$customer = new \WC_Customer( $data['user_id'] );
 
 		// If user is not a Woo customer, don't need to sync them.
-		if ( ! $customer->get_order_count() ) {
+		if ( ! $customer->get_order_count() && empty( WooCommerce_Connection::get_active_subscriptions_for_user( $data['user_id'] ) ) ) {
 			return;
 		}
 
--- a/includes/plugins/woocommerce/class-woocommerce-connection.php
+++ b/includes/plugins/woocommerce/class-woocommerce-connection.php
@@ -270,6 +270,8 @@ class WooCommerce_Connection {
 	 * Can optionally pass an array of product IDs. If given, only subscriptions
 	 * that have at least one of the given product IDs will be returned.
 	 *
+	 * If the subscription was a gift, it will return true only if the given user is the recipient.
+	 *
 	 * @param int   $user_id User ID.
 	 * @param array $product_ids Optional array of product IDs to filter by.
 	 *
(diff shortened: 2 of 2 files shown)


dkoo (approved):
Approving as the testing instructions work as described. However, I have a question and noticed a couple of possible regressions while testing this:

### QUESTION

[This PR](https://github.com/Automattic/newspack-plugin/pull/4287/files#diff-ae452f417e51fa1a6021e09104bbc718889cf86eae8b4f68ad7b719e502d443dR115) checks whether users should be limited to one active subscription if they have a subscription already with `pending` status—should this remain? I think yes because they can activate the pending sub by paying for it, right?

### REGRESSIONS

Probably not strictly related to this PR, but documenting here so we don't forget to investigate:

1. Logging in as a reader is not triggering an ESP sync as I would expect. 
2. After registering for a reader during a purchase, the `np_reader_without_password` is not being set, so when logging in for the first time I'm prompted for a password despite never having set one.

dkoo (changes requested):
Actually, upon further investigation I think what I described above as regressions IS related to this PR:

1. [This check for order count](https://github.com/Automattic/newspack-plugin/blob/hotfix/active-subscription-checks/includes/data-events/connectors/class-esp-connector.php#L121) is preventing the gift recipient's data from syncing upon login. The gift recipient has 0 orders as the order is owned by the gift giver.
2. The Woo Subscriptions Gifting extension used to provide a direct link to "claim" a gifted subscription that included an onboarding flow requiring the recipient to provide billing information and set a password before they could claim the gift. Therefore, we [assume](https://github.com/Automattic/newspack-plugin/blob/hotfix/active-subscription-checks/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions-gifting.php#L77) that a gift recipient will have set a password during this process. However, this seems to have changed at some point. Testing the flow now, the recipient only receives a link to log in at My Account in order to claim the gift, with no special onboarding flow. We may want to remove this handler to allow gift recipients to log in via OTP by default.

Screenshot 2025-12-01 at 5 23 47 PM

leogermani:
@dkoo I'm struggling to see how these things are related to this PR

# Syncing on logging in

This seems to be a deeper problem when we might need to revisit this condition soon.

This is also preventing team members from being synced on login.

I added a "fix" in https://github.com/Automattic/newspack-plugin/pull/4334/commits/6f9fe2c639017e4aef96f6303dcebd4bb1575a4a to make it sync gift recipients... but I think there's more to look here.

If you agree I can create a separate issue for that

# Setting a password

The email that the gift recipient gets always include the "set a password" link.

image

But even if I don't click that and try to follow the other links, I'll be asked for my email and then end up in this screen that asks for a password and gives me other login options:

image

Removing that check doesn't change anything in this behavior... It also doesn't change the "reset password" link in My Account.

What seems to be happening is that when Woo Subscriptions creates the user for the recipient, it creates via normal WP flows. It doesn't use Reader Activation's `register_reader` method, so it never [sets the flag](https://github.com/Automattic/newspack-plugin/blob/hotfix%2Factive-subscription-checks/includes/reader-activation/class-reader-activation.php#L2294) that the user doesn't have a password.

Having said all that, I agree that we can probably remove that filter. But I think we should look at this issue in a separate ticket as well and try to improve this flow. WDYT?

dkoo:
> Syncing on logging in
> 
> This seems to be a deeper problem when we might need to revisit this condition soon.
> 
> This is also preventing team members from being synced on login.
> 
> I added a "fix" in https://github.com/Automattic/newspack-plugin/commit/6f9fe2c639017e4aef96f6303dcebd4bb1575a4a to make it sync gift recipients... but I think there's more to look here.
> 
> If you agree I can create a separate issue for that

Teams handles its own sync logic in our [custom plugin class](https://github.com/Automattic/newspack-plugin/blob/hotfix/active-subscription-checks/includes/plugins/class-teams-for-memberships.php#L89-L96). Maybe we can follow a similar pattern for other exceptions, rather than handle all of the exceptions in the "main" handler.

I do like the change you made in this PR to check for active subscriptions since there are cases where an active subscription might not have any orders (such as when it's manually created, or imported via the Importer/Exporter plugin). I'm okay with doing any other changes we might need on this in a separate PR.

> What seems to be happening is that when Woo Subscriptions creates the user for the recipient, it creates via normal WP flows. It doesn't use Reader Activation's `register_reader` method, so it never [sets the flag](https://github.com/Automattic/newspack-plugin/blob/hotfix%2Factive-subscription-checks/includes/reader-activation/class-reader-activation.php#L2294) that the user doesn't have a password.

Aha, this is the deeper issue, you're right. Agree, let's handle in a separate PR

dkoo (approved):
Approving since we'll handle my feedback in separate PRs, if needed. This is good to go as long as this is okay:

> [This PR](https://github.com/Automattic/newspack-plugin/pull/4287/files#diff-ae452f417e51fa1a6021e09104bbc718889cf86eae8b4f68ad7b719e502d443dR115) checks whether users should be limited to one active subscription if they have a subscription already with pending status—should this remain? I think yes because they can activate the pending sub by paying for it, right?

PR #4338, opened 2025-12-02
fix(content-gates): resolve fatals and mismatched data types

dkoo:
See discussion in #4251. As [mentioned](https://github.com/Automattic/newspack-plugin/pull/4251#pullrequestreview-3532591528):

- Fixes several issues with the `sanitize_gate` method (UI has no expected `description` field, `sanitize_access_rule` converts an array of product IDs to strings via `sanitize_text_field`, etc)
- Changes the UI element for selecting subscription products to a `FormTokenField`: the UI only let you choose a single value, but the back-end expects an array of values. 
- Corrects the content rule slug when evaluating the `post_types` rule (note plural)
- Fixes a fatal error due to mismatched parameters from the `newspack_content_gate_restrict_post` filter hook in `class-content-gifting.php`

### DIFF
--- a/includes/wizards/audience/class-audience-content-gates.php
+++ b/includes/wizards/audience/class-audience-content-gates.php
@@ -232,8 +232,8 @@ class Audience_Content_Gates extends Wizard {
 	 */
 	public function sanitize_gate( $gate ) {
 		return [
-			'title'         => sanitize_text_field( $gate['title'] ),
-			'description'   => sanitize_text_field( $gate['description'] ),
+			'title'         => isset( $gate['title'] ) ? sanitize_text_field( $gate['title'] ) : __( 'Untitled Content Gate', 'newspack-plugin' ),
+			'description'   => isset( $gate['description'] ) ? sanitize_text_field( $gate['description'] ) : '',
 			'metering'      => $this->sanitize_metering( $gate['metering'] ),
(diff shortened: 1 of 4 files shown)


includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -232,8 +232,8 @@ public function register_api_endpoints() {
 	 */
 	public function sanitize_gate( $gate ) {
 		return [
-			'title'         => sanitize_text_field( $gate['title'] ),
-			'description'   => sanitize_text_field( $gate['description'] ),
+			'title'         => isset( $gate['title'] ) ? sanitize_text_field( $gate['title'] ) : __( 'Untitled Content Gate', 'newspack-plugin' ),
+			'description'   => isset( $gate['description'] ) ? sanitize_text_field( $gate['description'] ) : '',
```
miguelpeixe: The description is planned to be a dynamic string based on the rules, which is why it was removed on the parent PR. We should remove it from here as well.

miguelpeixe (approved):
Thank you for addressing these! Just left a minor comment.

includes/wizards/audience/class-audience-content-gates.php
dkoo: Gotcha, I couldn't remember what we decided on that. Removed in 2c8ef6eec

PR #4339, opened 2025-12-03
fix: hide block visibility from all blocks

laurelfulford:
### All Submissions:

This PR hides the new 'visibility' option from all Newspack sites. 

We're doing this for a couple reasons:
* We don't know how it will interact with other plugins/services used on our sites. For example, our Newsletter plugin ignores it on sent messages.
* Since the blocks are so wholly hidden (only available in the 'List' view after being hidden) it seems like it'd be easy for publishers to lose track of this kind of stuff

The Newsletter plugin has an identical snippet of code because it's available on WP.org and the feature is wholly broken in that plugin; I've opted to add it to this plugin as a more general thing. 

It may also make sense to move this in the classic theme if it's likely we'll use the 'visibility' feature in the block theme (for example, if we want to add hidden elements to template parts so they can be easily shown without having to spin up multiple parts/patterns just to add/remove one element from one of them).

Lastly, this change also means that blocks that have set not to be visible cannot be made visible again without editing the code -- you have to manually remove `{"metadata":{"blockVisibility":false}}` from the block markup.

1. On a site running WP 6.9, add a block to the editor and open the `...` menu in the Block toolbar; note you have a Hide option.

CleanShot 2025-12-03 at 10 07 17

2. Apply this PR and run `npm run build`
3. Refresh the editor.
4. Open the `...` menu of a block in the Block Toolbar and confirm you don't see the 'Hide' option.

### Other information:

### DIFF
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -46,6 +46,16 @@ final class Blocks {
 	public static function enqueue_block_editor_assets() {
 		Newspack::load_common_assets();
 
+		// Enqueue universal editor modifications.
+		\wp_enqueue_script(
+			'newspack-editor',
+			Newspack::plugin_url() . '/dist/editor.js',
+			[],
+			NEWSPACK_PLUGIN_VERSION,
+			true
+		);
+
+		// Blocks script depends on editor script to ensure filters run first.
 		\wp_enqueue_script(
 			'newspack-blocks',
 			Newspack::plugin_url() . '/dist/blocks.js',
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -47,10 +47,11 @@ final class Blocks {
 		Newspack::load_common_assets();
 
 		// Enqueue universal editor modifications.
+		$editor_asset = require_once dirname( NEWSPACK_PLUGIN_FILE ) . '/dist/editor.asset.php';
 		\wp_enqueue_script(
 			'newspack-editor',
 			Newspack::plugin_url() . '/dist/editor.js',
-			[],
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Works, but I agree that maybe the classic Theme is the better place for this code if we decide that the Block Theme _should_ support visibility.

dkoo (changes requested):
Oh wait, I just caught that the new `editor.js` is outputting an `asset.php` file with WP dependencies. We should include these dependencies in the `enqueue_script` array similar to how we do here: https://github.com/Automattic/newspack-plugin/blob/fix/remove-visibility-from-all-blocks/includes/content-gate/content-gifting/class-content-gifting.php#L451-L452

laurelfulford:
Thanks @dkoo!

>  I agree that maybe the classic Theme is the better place for this code if we decide that the Block Theme should support visibility.

It seems easier than messing with a theme check, anyway 🙂 I'm going to leave it here for now, and can take care of moving it if we do decide we'd like to use this in the Block Theme. I think the risk of confusion/conflict still stands, but that there's some possible use cases there that would outweigh that. 

> Oh wait, I just caught that the new editor.js is outputting an asset.php file with WP dependencies. 

Eep, good catch! This should be fixed in [60f1510](https://github.com/Automattic/newspack-plugin/pull/4339/commits/60f15105c47116ac8b716310e71c3d1d714e22c5) but let me know if it still looks funky!

dkoo (approved):

laurelfulford:
Thanks @dkoo!

PR #4342, opened 2025-12-04
fix: workaround issue and conflict with teams

leogermani:
### All Submissions:

The Teams for Memberships plugin has an action that saves the billing first and last name, and the display name, whenever a WC Customer is created.

The problem is that action is designed to be executed in some specific form in the admin, but it has no guardrails and it ends up being invoked also when we register a reader via RAS.

This is messing up the Display name and generating some conflicts with the Newsletter Subscription form block, because that block has a `last_name` field. Teams was then adding a metadata with only the billing last name to the user (but not the first).

As a result, when we sync the user to the ESP, and fetch the data from the customer [here](https://github.com/Automattic/newspack-plugin/blob/ac12cca41eca26a532e1f2dc2ea7c49258343566/includes/reader-activation/sync/class-woocommerce.php#L418), we end up with only the last name as the name, because firs_name returns empty.

This PR simply removes that action upon reader registration and fixes the issue.

On `trunk` or `release`

1. Activate Woo Memberships and Teams for Memberships
2. Activate RAS
3. Add a Newsletter Subscription block to a page, and enable the first and last name fields
4. As an anonymous reader, go to that page and submit the form
5. On the DB, observe that the user gets a `billing_last_name` metadata and that the display name is equal the login
6. On the connected ESP, confirm that the first name of the contact was replaced with the last name.

Checkout this branch and repeat the steps above:

* Confirm the `billing_last_name` meta is NOT created
* Confirm the display name looks correct
* Confirm the name on the ESP looks correct

### Other information:

### DIFF
--- a/includes/reader-activation/class-reader-activation.php
+++ b/includes/reader-activation/class-reader-activation.php
@@ -2268,6 +2268,13 @@ final class Reader_Activation {
 			// Unhook from WooCommerce as it's already been canonized above.
 			remove_filter( 'woocommerce_new_customer_data', [ __CLASS__, 'canonize_user_data' ], 10, 1 );
 			if ( function_exists( '\wc_create_new_customer' ) ) {
+
+				// Avoid an issue with Teams for Memberships plugin that messes up billing first and last name and display name.
+				// See SkyVerge\WooCommerce\Memberships\Teams\Frontend::save_team_member_name().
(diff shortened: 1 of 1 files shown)


rbcorrales (approved):
I was able to replicate the issue in the DB and confirm that the correct data was being saved in the user display name and that the `billing_last_name` meta was no longer being saved. However, I couldn't identify any last name sync issues with the ESP I was testing.

PR #4343, opened 2025-12-04
fix(my-account): set new payment method as default

miguelpeixe:
### All Submissions:

NPPD-985

Via the `add_filter( 'wc_stripe_save_to_subs_checked', '__return_true' )` we enforce a new Stripe payment method to be set to all active subscriptions.

This is not properly communicated, and the payment method is also not automatically set as **Default**, which adds to the confusion.

This PR adds a paragraph inside the Stripe gateway form and automatically sets the new method as the default.

image

1. Make sure you have Stripe as your payment gateway
2. As a reader with active subscriptions, visit "My Account -> Payment information"
3. Click to add a new payment method and confirm the new "The card will be automatically saved to your active subscriptions and set as your default payment method." paragraph
4. Add the new card and confirm it's set as default, and is set to your active subscriptions

### Other information:

### DIFF
--- a/includes/plugins/woocommerce/class-woocommerce-connection.php
+++ b/includes/plugins/woocommerce/class-woocommerce-connection.php
@@ -71,7 +71,7 @@ class WooCommerce_Connection {
 
 		// Automatically set newly added payment method as default and update subscriptions.
 		\add_action( 'woocommerce_new_payment_token', [ __CLASS__, 'set_payment_method_as_default' ], 10, 2 );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works! Should we also do the same when adding a new card via WooPayments?

miguelpeixe:
> Should we also do the same when adding a new card via WooPayments?

I looked for it, but WooPayments doesn't provide anything similar to the `wc_stripe_save_to_subs_checked` filter, which is the foundation for this behavior.

PR #4344, opened 2025-12-05
refactor(content-gating): separate Newspack & Memberships gate CPTs

dkoo:
### All Submissions:

Separates the post types used by the Woo Memberships integration and our in-development content gating features. This establishes a new `np_content_gate` post type to use for our new first-party features going forward, and renames the existing post type to "Membership Gate" in all admin-facing contexts to make it easier to distinguish between the two.

We decided to separate the data for these featuresets for a few reasons:

- To avoid unintended data pollution and changes when switching back and forth between implementations
- To allow the Memberships class to handle functionality that's only relevant for the Woo Memberships integration
- To lay the groundwork for allowing the gate posts in each implementation to expect and handle different post meta (e.g. Memberships gates don't need a priority, while Newspack gates do; Newspack gates don't need a separate meta value for anonymous vs. registered readers, while Memberships gates do, etc.)
- When both implementations are enabled, the Woo Memberships implementation will take precedence on the front-end, but the admin will allow you to configure both implementations. This will make it easier to manually migrate configuration between implementations while allowing existing Memberships-based configurations to continue working on the front-end.

For now, editor JS and some meta handling (such as metering, content gifting, and countdown banner settings) are still shared between the two implementations; these may be split up more in future.

1. Enable Woo Memberships AND add `define( 'NEWSPACK_CONTENT_GATES', true );` to wp-config.php
2. In WP admin, confirm that you can see both **Audience > Configuration > Content Gating** (Woo Memberships) AND **Audience > Content Gates** (Newspack) admin pages
3. In **Audience > Configuration > Content Gating**, confirm that all Content Gifting, Countdown Banner, and other global settings appear and behave as expected (no changes).
4. Edit the memberships gate from **Audience > Configuration > Content Gating**, and confirm that all post type labels in the editor refer to the post as a "Membership Gate" instead of "Content Gate".
5. Confirm that all settings in the editor still appear and behave as expected (no changes for Memberships gates).
6. As a reader, confirm that all content gating features still behave as expected and that the Woo Memberships gate is rendered when content is restricted.
7. As an admin, visit **Audience > Content Gates** and confirm that there are no content gates listed, even if you had created content gates that appear here on `trunk`.
8. Create a content gate and confirm that you can configure its content and access rules. Confirm that you can edit the content by clicking on the gate's title in the card's header component.
9. While editing the gate, add some blocks (try a Newspack Content Gating pattern). Confirm that all post type labels say "Content Gate". Confirm that the appropriate **Audience > Content Gating** WP menu item is highlighted. Confirm that all expected gate settings appear in the sidebar, except for this Memberships-specific panel:

Screenshot 2025-12-04 at 5 41 02 PM

10. Configure the gate post with different display and metering settings than the Memberships gate from step 4 and publish.
11. In new reader sessions, confirm that with Woo Memberships still active, you still see the gate and behavior from the memberships gate from step 4, but after deactivating Woo Memberships, you see the content gate and behavior from step 9.
12. Confirm that you can easily switch back and forth between implementations by activating or deactivating Woo Memberships while `NEWSPACK_CONTENT_GATES` is defined as `true`.

### Other information:


src/wizards/audience/views/setup/index.js
```diff
@@ -99,10 +99,11 @@ function AudienceWizard( { confirmAction, pluginRequirements, wizardApiFetch },
...
 		},
-		config.enabled && {
-			label: __( 'Content Gating', 'newspack-plugin' ),
-			path: '/content-gating',
-		},
+		config.enabled &&
+			config.has_memberships && {
+				label: __( 'Content Gating', 'newspack-plugin' ),
+				path: '/content-gating',
+			},
```
miguelpeixe, line 108: `has_memberships` is a localized value (`window.newspackAudience.has_memberships`), not part of this config payload.

The Content Gating tab is never rendering.

src/content-gate/editor/editor.js
```diff
@@ -36,6 +36,7 @@ const overlaySizes = [
 ];
 
 function GateEdit() {
+	const newspack_memberships_gate = window.newspack_memberships_gate || {};
```
miguelpeixe, line 39: Even though this file is currently shared between the two strategies, this will likely remain the Memberships approach, and something a lot simpler will handle the gate layout. WDYT of moving this to `memberships-gate-editor.js` for now?

includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -69,15 +74,193 @@ public static function is_active() {
...
+
+	/**
+	 * Get the URL for editing the custom gate.
+	 *
+	 * @param int|false $gate_id Gate ID.
+	 *
+	 * @return string
+	 */
+	public static function get_edit_gate_url( $gate_id = false ) {
+		$action = 'newspack_edit_content_gate';
```
miguelpeixe: Similar to the previous comment, can we change this to `newspack_edit_memberships_gate`?

includes/content-gate/class-content-gate.php
```diff
@@ -271,14 +275,22 @@ public static function register_meta() {
...
-							'type' => 'string',
+							'type'       => 'object',
+							'properties' => [
+								'slug'  => [
+									'type' => 'string',
+								],
+								'value' => [
+									'type' => 'mixed',
+								],
+							],
```
miguelpeixe: We're not editing `content_rules`, `access_rules`, or `gate_priority` in the block editor. Is there another reason to maintain `register_meta()` for such meta?

As we move to a different CPT for reg/hard walls, this should also split so style and other editor-available meta live under the layout CPT.

includes/content-gate/class-metering.php
```diff
@@ -86,18 +86,20 @@ public static function register_meta() {
...
+					$key,
+					[
+						'object_subtype' => $cpt,
+						'show_in_rest'   => true,
+						'type'           => $config['type'],
+						'default'        => $config['default'],
+						'single'         => true,
+					]
+				);
+			}
```
miguelpeixe, line 102: I'm also not sure about this. The new gate strategy requires a different schema for metering, and it's also not handled in the block editor.

miguelpeixe (changes requested):

src/wizards/audience/views/setup/index.js
dkoo: Whoops, you're right! 14753dbbb fixes this.

src/content-gate/editor/editor.js
dkoo: I agree, I didn't make this change in this PR since I started it before we decided to ultimately split the layouts into a new CPT. I think we should handle that change in a separate PR.

includes/content-gate/class-metering.php
dkoo: I agree, but I'd like to handle moving to our new data structure in a separate PR to not make this one quite so big. This PR in my mind is more about splitting out our own gate posts from the existing Memberships gate posts while not making too many changes to the underlying functionality just yet.

includes/plugins/wc-memberships/class-memberships.php
dkoo: Good idea, updated in 49f9dd131!

includes/content-gate/class-content-gate.php
dkoo: Agreed on not needing to register those meta fields since we're no longer exposing them in the block editor at all. Removed in df1a2dd13.

I think we can handle further splitting the meta structure for our new CPTs in a separate PR.

dkoo:
@miguelpeixe thanks for the feedback! I think I'd like to get this one merged without getting into splitting our own gates into gate vs. layout CPTs just yet. I might even want to merge the other two outstanding PRs before we start making those deeper changes too, just because many of these changes will be useful for that data restructure too.

miguelpeixe (approved):

PR #4345, opened 2025-12-05
feat(content-gating): update content gates UI

dkoo:
### All Submissions:

Tweaks for the Content Gates UI:

- When only one gate exists, show expanded by default and hide drag/drop/resorting controls
- Add badge to denote gate status
- Swap order of Content and Access Rules sections (content comes first, then access)
- Add controls to publish and unpublish gates
- Add controls to trash, restore, and permanently delete gates
- Control the expanded/collapsed state of gate card components when adding new gates (when adding a new gate, expand the new one and collapse all others)
- When adding new gates, increment priority value so the newest gate is always last in order
- Allow "enter" keypress to submit the "add content gate" modal
- Reorganize metering settings into a separate `Metering` component for parity with `AccessRules` and `ContentRules` components

1. Start with zero content gate posts: `wp post delete $(wp post list --post_type=np_content_gate --post_status=all --format=ids) --force`
2. Visit **Audience > Content Gates** and click "Add content gate". Enter a gate name and press the "enter" key to submit the form.
3. Confirm that a new blank gate is created with "draft" status, and that it displays as expanded without any drag/drop or resorting controls:

Screenshot 2025-12-05 at 12 03 18 PM

4. Confirm that the bottom of the card has "Delete", "Save Draft", and "Publish" buttons (but that the "Publish" button is disabled for now because there are no Content Rules or Access Rules yet)
5. Add some Content Rules and Access Rules and confirm that the "Publish" button becomes available. Tweak metering settings to confirm no regressions. For now, click "Save Draft" and confirm that your changes are saved and persist after a page refresh. (Note: at this time, the added rules aren't yet validated to ensure that they'll do anything on the front-end, so it's still possible to publish a gate that won't render due to missing rules configuration.)

Screenshot 2025-12-05 at 12 03 31 PM

6. Click "Publish" and confirm that the gate shows a "Published" status now. Confirm that the status persists after a refresh. Confirm that the buttons shown now are "Unpublish" and "Update".

Screenshot 2025-12-05 at 12 04 03 PM

7. Make some more changes and click "Update", and confirm that changes persist after a refresh. Click "Unpublish" and confirm that the gate goes back to "Draft" status, and that the draft status persists after a refresh.
8. Make sure the single gate component is expanded. Add another gate and confirm that the new draft gets added at the end of the gate list, and that both gate components now have drag/drop/resorting controls. Confirm that the existing gate component is automatically collapsed and that the new one is automatically expanded. Confirm that the drag/drop/resorting controls still work as expected.
9. Click "Delete" for one of the gate posts, and confirm that it gets updated to "Trash" status. Confirm that the status persists after a refresh.

Screenshot 2025-12-05 at 12 03 46 PM

10. As a reader, view restricted posts and confirm that you only ever see the published gate, never the trashed one even if it has more priority.
11. Back in the admin, test the "Restore" and "Permanently delete" buttons for the trashed gate and make sure they set the gate to "draft" status and permanently delete the gate post from the DB, respectively.

### Other information:


dkoo:
Tagging @thomasguillot for a design review since this implements some of the UI changes we talked about but never implemented during the meetup. Also, I thought about the idea of adding some filter controls to show and hide Trashed gates, but I think this PR is big enough for now. We can always implement more controls in another PR.

thomasguillot:
@dkoo I pushed some small changes: https://github.com/Automattic/newspack-plugin/pull/4345/commits/9224c6e922ae4148ac64e8119d6b7a9131f9c8bd

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -269,6 +270,7 @@ public function sanitize_gate( $gate ) {
 			'access_rules'  => $this->sanitize_rules( $gate['access_rules'] ),
 			'content_rules' => $this->sanitize_rules( $gate['content_rules'], 'content' ),
 			'priority'      => intval( $gate['priority'] ),
+			'status'        => sanitize_text_field( $gate['status'] ),
```
miguelpeixe: Should this be using `sanitize_status`?

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -394,6 +396,22 @@ public function sanitize_content_rule( $content_rule ) {
...
+	 * @param string $status Post status.
+	 *
+	 * @return string The sanitized post status.
+	 */
+	public function sanitize_status( $status ) {
+		$sanitized = sanitize_text_field( $status );
+		$valid = in_array( $sanitized, Content_Gate::get_post_statuses(), true );
+		if ( ! $valid ) {
+			return new \WP_Error( 'invalid_gate_post_status', __( 'Invalid gate post status.', 'newspack-plugin' ), [ 'status' => 400 ] );
+		}
```
miguelpeixe, line 412: The method only expects only `string` return.

To support returning error, this should also be handled in the `sanitize_gate()` method, otherwise the `status` will be populated with an error object.

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -434,7 +452,12 @@ public function delete_gate( $request ) {
 		if ( Content_Gate::GATE_CPT !== $gate->post_type ) {
 			return new \WP_Error( 'invalid_gate_type', __( 'Invalid gate type.', 'newspack-plugin' ), [ 'status' => 400 ] );
 		}
-		wp_delete_post( $id, true );
+		$force = $gate->post_status === 'trash';
+		if ( $force ) {
+			wp_delete_post( $id, $force );
+		} else {
+			wp_trash_post( $id );
+		}
```
miguelpeixe, line 461: Non-blocking, but that's the default behavior of `wp_delete_post( $id );` without the second argument.

miguelpeixe (changes requested):
Great changes! A few comments inline

includes/wizards/audience/class-audience-content-gates.php
dkoo: Yes, updated in 2c0d334b8

includes/wizards/audience/class-audience-content-gates.php
dkoo: Ah yes, 0cde63876 just avoids returning a `WP_Error` and falls back to the gate's existing status if the passed status is invalid, or `draft` if it can't get that.

includes/wizards/audience/class-audience-content-gates.php
dkoo: That's what I thought too, but doing `wp_delete_post( $id, false );` seems to permanently delete the post and skip trash regardless of what you pass for that second arg.

miguelpeixe (approved):
Thank you for the revisions!

PR #4351, opened 2025-12-09
fix: hide 'trashed' subscriptions from My Account

laurelfulford:
### All Submissions:

This PR hides trashed subscriptions from the My Account view. 

If you only have trashed subscriptions, it hides the Subscriptions menu all together.

See: NPPM-2442

1. Start on a test site with RAS set up
2. Set up a subscription product that can be purchased multiple times by the same reader (just to make testing easier), and add a Checkout button block 
3. In an incognito window, run through a purchase of your subscription with the credit card `4000 0000 0000 0341` (which will result in a failed payment).
4. After the transaction fails, go to 'My Account', and note the trash sub:

CleanShot 2025-12-09 at 15 15 41

5. Apply this PR and run `npm run build`.
6. In the incognito window, refresh the page; confirm you don't have any subscriptions listed, and there isn't a Subscription menu item (so the typical reader wouldn't get to this screen).

CleanShot 2025-12-09 at 15 16 17

7. Go back and try to buy the subscription again, this time using a successful test card.
8. Go back to My Account, and confirm you have a single My Subscription; when you click on it it should show your one subscription details:

CleanShot 2025-12-09 at 15 27 23

9. Purchase a second subscription and confirm you get the Subscriptions menu item with a list as the top level in My Account, and that only non-trashed subs are shown (you can also mess with the subscriptions, cancelling them, etc.). All of them should show, unless they are trashed:

CleanShot 2025-12-09 at 15 28 51

10. Lastly, as the admin manually trash all but one subscription and make sure you're back to the view in step 8 (if you just refresh the /subscriptions page as the user, you'll still see the list with one item, but the navigation should say My Subscription and just go to the remaining non-trashed one).

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
+++ b/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
@@ -20,6 +20,7 @@ class WooCommerce_Subscriptions {
 		add_action( 'plugins_loaded', [ __CLASS__, 'woocommerce_subscriptions_integration_init' ] );
 		add_filter( 'woocommerce_subscriptions_product_limited_for_user', [ __CLASS__, 'maybe_limit_subscription_product_for_user' ], 10, 3 );
 		add_filter( 'woocommerce_subscriptions_product_trial_length', [ __CLASS__, 'limit_free_trials_to_one_per_user' ], 10, 2 );
+		add_filter( 'wcs_get_users_subscriptions', [ __CLASS__, 'filter_subscriptions_for_account_page' ], 10, 1 );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


miguelpeixe (changes requested):
This tested well, but I think we can go for a simpler approach. The user subscription list and "Subscriptions/My Subscription" button rely on `wcs_get_users_subscriptions()`, which supports a hook.

Something like this should be enough:

```php
add_filter(
	'wcs_get_users_subscriptions',
	function( $subscriptions ) {
		if ( function_exists( 'is_account_page' ) && is_account_page() ) {
			$subscriptions = array_filter(
				$subscriptions,
				function( $subscription ) {
					return ! $subscription->has_status( 'trash' );
				}
			);
		}
		return $subscriptions;
	},
	10
);
```

laurelfulford:
Oh, that's so much nicer - thanks @miguelpeixe! 🙌 

I left a couple pieces of the original code in:
* I noticed when you just had My Subscription in the menu it didn't highlight when you clicked on it, so that's been fixed
* I tweaked the styles of buttons when they're in notices to just look like regular links - they were displaying full size before and sticking out of the bottom. I only saw this kind of message + button when I visited `/my-account/subscriptions/` when no subscriptions were actually showing - a reader's not going to get to that but it was bugging me 😆

miguelpeixe (approved):
Thank you for revising! 🙌

PR #4366, opened 2025-12-16
fix: select specific product for content gifting and countdown banners

dkoo:
### All Submissions:

Tweaks the UI for the Content Gifting and Countdown banners to make the behavior of the CTA buttons more explicit and configurable. Instead of a single text field that behaves differently if left empty, there's now a toggle option to choose whether the button should start a modal checkout (and if so, which product), or link to a URL (and if so, what URL).

Screenshot 2025-12-16 at 3 55 33 PM

Also moves the enqueuing of the `content-banner` front-end assets to the main `Content_Gate` class. Since both featuresets share the same front-end JS and CSS, this makes it clearer when the assets should be enqueued. On `trunk`, because of the way each class attempts to enqueue the same assets but on different conditions, I was running into some edge cases where I would expect the `content-banner` JS and CSS assets to be enqueued, but they weren't due to conflicting or missing asset dependencies. In particular, the `content-banner` assets weren't always enqueued on posts where the countdown meter should appear, if the post wasn't also giftable. The refactor should fix those edge cases.

Closes NPPM-2204.

1. Check out this branch. Make sure to select a grouped or variable product as your Primary Subscription Product in **Audience > Subscriptions**.
2. Visit **Audience > Configuration > Content Gating**. Enable both Content Gifting and Countdown Banner options if not already enabled and confirm that you see the new UI as depicted above.
3. Confirm that both settings cards default to "Product" and that the product selection defaults to the primary subscription product set in step 1.
4. On the front-end, smoke test both content gifting and countdown banners and confirm that their buttons start modal checkouts with the selected product. 
5. Change the product selections for both and confirm on the front-end that the modal checkout flows now use the newly selected products.
6. Change the "Action" settings to "Landing Page" and save settings with an empty "Subscribe button URL" field. Confirm that both the previews and front-end banners show no button with an empty URL field.
7. Enter URLs for each field and save. On the front-end, confirm that the buttons now link to the configured URLs.

### Other information:


dkoo:
Tagging @thomasguillot for a design review!

thomasguillot:
@dkoo small design update here → https://github.com/Automattic/newspack-plugin/pull/4366/commits/27e229a38676d0164c7b6c07dcd57a7fe2c01e41

thomasguillot:
I had a weird bug when the toggle wasn't enabled. Updating the if/else check fixed it: https://github.com/Automattic/newspack-plugin/pull/4366/commits/f6c9dab210f73e6a3227b4c4bb5def228e2a75c4

bug

dkoo:
Ah, thanks for finding/fixing that bug, @thomasguillot!

includes/content-gate/class-access-rules.php
```diff
@@ -213,10 +213,34 @@ public static function get_subscription_products_options() {
...
+								return true;
+							}
+						}
+						return false;
+					}
+					return true;
+				}
+			)
+		);
+
```
miguelpeixe: This seems unrelated to the PR, and I'm not sure I agree with the change... I'd prefer mapping the rule to subscriptions directly rather than proxied via a group.

Groups are more unstable, as you can rearrange and delete them as needed.

miguelpeixe (changes requested):
Tested well!

includes/content-gate/class-access-rules.php
dkoo: Ah, my thinking here was that we might want to be able to map access rules to the primary subscription tier product, so this allows selecting group products with the same conditions as that feature. In that scenario, you think it would be safer to require admins to explicitly select every product that might also be in the primary subscription (grouped) product?

miguelpeixe: It differs because the primary subscription product is an offer, to choose between "Basic" or "Pro". This will likely translate to different gates.

You might also modify that grouped product for new offerings and switch opportunities, which may require you to redefine your access rules.

dkoo: 7829c9f restores the method for access rules, and creates a new `Content_Gate::get_purchasable_product_options()` method which fetches eligible products via `Subscriptions_Tiers::get_tier_eligible_products()` and formats them as an options array.

miguelpeixe: `get_tier_eligible_products()` returns only variable subscriptions or grouped products. No regular subscriptions. Do you think this will be an issue?

miguelpeixe (approved):
Looks good!

PR #4368, opened 2025-12-17
fix(content-gating): stop running `restrict_post` after gate is rendered

dkoo:
### All Submissions:

Fixes some minor bugs in the rendering of content gates:

- Avoids repeat runs of `restrict_post()` if the gate has been rendered for the current post already. This avoids potential rendering errors resulting from passing the post content and gate through `the_content` filters multiple times.
- Apply `the_content` filters instead of `newspack_gate_content` inside `get_restricted_post_excerpt()`, to avoid adding unintended `<br />` elements and `nbsp;` characters to the gate content markup.

Also ensures that the number inputs for anonymous and registered metering views must be positive numbers, and adds a warning hint in the content gate settings component that explains that setting both anonymous and registered metering views to `0` is effectively the same as having metering disabled:

Screenshot 2025-12-17 at 11 30 24 AM

1. On `trunk`, set `define( 'NEWSPACK_CONTENT_GATES', true );` in wp-config.php and deactivate WooCommerce Memberships.
2. In **Audience > Content Gates**, publish a single content gate for Post Type: Posts with metering enabled. Confirm that you can't set anonymous or registered views to values below 0, and that if you set both anonymous and registered views to `0`, you see the warning as in the above screenshot.
3. Save the gate with some non-zero number of anonymous and registered views. In the gate's content, insert a pattern that contains login/register buttons and a Checkout Button (I used the "Paywall with One Tier and Metering" block pattern), and set the gate style to "Inline".
4. On the front-end, view posts until you see the content gate. Observe that some elements have extra whitespace:

Screenshot 2025-12-17 at 12 22 19 PM

5. In a post that will be gated, insert some Page Break blocks into the content.
6. On the front-end, view the post with the page breaks and observe that the inline gate renders with malformed markup:
 
Screenshot 2025-12-17 at 12 24 58 PM

7. Check out this branch and view gated posts both with and without page breaks. Confirm that the inline gate renders as expected with no visual issues, and that all interactive elements such as buttons/links still function as expected.

Screenshot 2025-12-17 at 12 29 24 PM

8. Change the gate to "Overlay" style and view the gated posts again. Confirm that the gate renders as an overlay with no visual or functional issues.

### Other information:

### DIFF
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -508,7 +508,9 @@ class Content_Gate {
 
 		// Wrap gate in a div for styling.
 		$gate = '<div class="newspack-content-gate__gate newspack-content-gate__inline-gate">' . $gate . '</div>';
-		return $gate;
+
+		// Strip line breaks and tabs.
+		return preg_replace( '/\s+/S', ' ', $gate );
 	}
 
 	/**
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -517,7 +517,7 @@ class Content_Gate {
 	 * @return string
 	 */
 	public static function get_inline_gate_html() {
-		$gate_content = preg_replace( '/\s+/', ' ', self::get_inline_gate_content() );
+		$gate_content = preg_replace( '/\s+/S', ' ', self::get_inline_gate_content() );
 		return apply_filters( 'newspack_gate_content', $gate_content );
 	}
 
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -90,6 +90,9 @@ class Content_Gate {
 	 * @param \WP_Query $query Query object.
 	 */
 	public static function restrict_post( $post, $query ) {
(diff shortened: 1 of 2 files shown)


includes/content-gate/class-content-gate.php
```diff
@@ -194,12 +192,27 @@ public static function do_blocks( $content ) {
...
 	}
 
+	/**
+	 * _restore_wpautop_hook filter, but for the newspack_gate_content filter instead of the_content
+	 *
+	 * @param string $content Content.
+	 * @return string
+	 */
+	public static function restore_wpautop_hook( $content ) {
+		$current_priority = has_filter( 'newspack_gate_content', '_restore_wpautop_hook' );
```
miguelpeixe: This should also be `[ __CLASS__, 'restore_wpautop_hook' ]`

miguelpeixe (changes requested):

includes/content-gate/class-content-gate.php
dkoo: Whoops, fixed in f75b3c58c

miguelpeixe (approved):

PR #4371, opened 2025-12-20
feat(content-gating): add group subscription product options

dkoo:
### All Submissions:

Adds Group Subscription settings to product and subscription admin pages.

If a subscription is created with a product that has Group Subscription settings, it will inherit the product's settings. The settings can be manually overridden on a per-subscription basis when editing or creating the subscription.

This PR only handles adding the settings meta and UI to products and subscriptions, as well as the inheritance/override behavior. Handling logic for the settings will come in another PR.

Closes NPPD-1037.

1. Check out this branch
2. Create or edit a simple subscription product, a variable subscription product, and a non-subscription product.
3. Edit each product and confirm that the non-subscription product has no new settings.
4. Edit the simple subscription product and confirm that the new settings appear in the Product Data > General panel. By default, the "Group subscription" checkbox should be unchecked, and the "Group subscription member limit" number input should be hidden.

Screenshot 2026-01-08 at 4 55 09 PM

5. Confirm that checking the "Group subscription" checkbox unhides the "Group subscription member limit" number input, and unchecking it hides it again.
6. Leave the checkbox checked, add a number to the limit field, and save. Confirm that the values persist after the page reloads.
7. Edit the variable subscription and confirm that the new settings appear for each variation in the Product Data > Variations panel.
8. Confirm that the default states and the hiding/unhiding behavior work for each separate variation.
9. Update with different values for each variation, save, and confirm that all values persist for each variation after the page reloads.
10. As a reader, purchase subscriptions with the simple and variable subscription products. Make sure to purchase a subscription for each variation of the latter. (You can also test with preexisting subscriptions of a product to which you've added Group Subscription settings.)
11. In WP admin, edit each purchased subscription and confirm that there's a new Group Subscription settings metabox in the sidebar, with values matching the purchased product (also confirm that the product name and link here are correct):

Screenshot 2026-01-08 at 5 03 24 PM

12. Check and uncheck the "Group subscription enabled" checkbox and confirm that it shows and hides the "Group subscription member limit" field.
13. Update the values of both checkbox and limit field and save. Confirm that the changes persist after the page refreshes.
14. In WP admin, manually create a new subscription without any product attached, and update Group Subscription settings. Confirm that the values persist after the subscription is created.
15. Smoke test the "Auto-complete orders" custom product option feature and confirm there are no regressions with that (this PR uses the existing `WooCommerce_Products` class for that feature to add the new Group Subscription settings to product pages).

### Other information:


includes/plugins/woocommerce-subscriptions/class-group-subscriptions.php
```diff
@@ -0,0 +1,258 @@
+<?php
+/**
+ * Newspack Group Subscriptions.
+ *
+ * @package Newspack
+ */
```
miguelpeixe, line 6: Non-blocking, but I see this feature as a Woo add-on. Although it's being built primarily to interact with content gating, it fits best under `includes/plugins/woocommerce-subscriptions`.

includes/plugins/woocommerce-subscriptions/class-group-subscriptions.php
```diff
@@ -0,0 +1,258 @@
...
+			return $custom_options;
+		}
+		$custom_options['newspack_group_subscription_enabled'] = [
+			'id'            => '_newspack_group_subscription_enabled',
+			'label'         => __( 'Group subscription', 'newspack-plugin' ),
+			'description'   => __( 'Enable group subscriptions for this product.', 'newspack-plugin' ),
+			'default'       => self::DEFAULT_SETTINGS['enabled'],
+			'product_types' => [ 'subscription', 'subscription_variation' ],
+			'type'          => 'boolean',
+		];
```
miguelpeixe, line 56: Creating a new product never renders this option, only after saving the draft:

Image

miguelpeixe (changes requested):
Tested mostly well, and the approach looks good!

includes/plugins/woocommerce-subscriptions/class-group-subscriptions.php
dkoo: Good catch, I forgot to test this! 😬 da94aa917 should fix this for new products.

includes/plugins/woocommerce-subscriptions/class-group-subscriptions.php
dkoo: Good call. 543271529 moves the class file to `includes/plugins/woocommerce-subscriptions`.

miguelpeixe (approved):
Thank you for revising!

PR #4387, opened 2026-01-09
fix(memberships): handle expired team subscription resubscribes

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPM-2392/newsroom-nz-reactivated-subscriptions-not-being-reapplied-to-the-the

This PR addresses an issue where Teams for WooCommerce Memberships doesn't handle expired subscriptions when resubscribing, however our on-hold duration feature enforces subscriptions that fail renewal to expired. As a result, if a subscription goes through the on-hold duration, the subscriber is unable to resubscribe and get re-assigned to the previous team since we've set the status of the subscription to expired.

We fix this by replicating the resubscribe logic in teams but for expired subscriptions only.

1. As admin, set up a team membership, a team membership (subscription) product, and make sure the two are associated with each other
2. As a reader, subscribe to the team membership product and verify the teams menu becomes available in my account
3. As admin, verify the subscription, the new team, and membership
4. Expire the subscription manually via the Subscription edit page
5. Verify the team is no longer manageable in WP Admin
6. As the reader, resubscribe to the subscription via my account > Subscriptions
7. Verify the team menu becomes accessible again
8. As admin, verify the previous team is now manageable again and is still associated with the reader account

### Other information:

### DIFF
--- a/includes/plugins/class-teams-for-memberships.php
+++ b/includes/plugins/class-teams-for-memberships.php
@@ -288,13 +288,13 @@ class Teams_For_Memberships {
 						$user_membership->set_end_date( $new_subscription->get_date( 'end' ) );
 						// set the membership's subscription ID.
 						$subscription_membership = new \WC_Memberships_Integration_Subscriptions_User_Membership( $user_membership->post );
-						$subscription_membership->set_subscription_id( $subscription->get_id() );
+						$subscription_membership->set_subscription_id( $new_subscription->get_id() );
 						// bail if not associated with an order.
 						if ( ! $order instanceof \WC_Order ) {
 							continue;
 						}
 						$note     = '';
-						$order_id = $order->get_id();
+						$order_id = $resubscribe_order->get_id();
 						$subscription_membership->set_order_id( $order_id );
 						if ( $product instanceof \WC_Product ) {
 							$subscription_membership->set_product_id( $product->get_id() );
--- a/includes/plugins/class-teams-for-memberships.php
+++ b/includes/plugins/class-teams-for-memberships.php
@@ -274,15 +274,11 @@ class Teams_For_Memberships {
 		if ( $old_subscription && in_array( $old_subscription->get_status(), [ 'expired' ] ) ) {
 			$team_subscriptions = new \SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions();
 			$existing_teams     = $team_subscriptions->get_teams_from_subscription( $old_subscription_id );
-
 			if ( ! empty( $existing_teams ) ) {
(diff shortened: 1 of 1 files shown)


includes/plugins/class-teams-for-memberships.php
```diff
@@ -244,6 +245,70 @@ public static function enable_members_area_for_team_members( $disabled_wc_menu_i
...
+		if ( ! method_exists( '\WC_Memberships_Integration_Subscriptions_User_Membership', 'set_subscription_id' ) ) {
+			return;
+		}
+
+		$new_order_id        = $resubscribe_order->get_id();
+		$new_subscription_id = $new_subscription->get_id();
+		$old_subscription_id = $new_subscription_id > 0 ? $new_subscription->get_meta( '_subscription_resubscribe' ) : 0;
+		$old_subscription    = $old_subscription_id > 0 ? wcs_get_subscription( $old_subscription_id ) : null;
+
+		if ( $old_subscription && in_array( $old_subscription->get_status(), [ 'expired' ] ) ) {
```
dkoo, line 274: Do we also want to do this for subscriptions with a `cancelled` status, in case the team owner resubscribes for a previously cancelled subscription?

dkoo (approved):
The approach looks solid and tests well! Approving, but I also had a question below about possibly doing this for cancelled subscriptions as well? It would make sense to me to do the same for subscribers who repurchase previously cancelled subscriptions.

includes/plugins/class-teams-for-memberships.php
chickenn00dle: Good question! Actually Teams already handles the `cancelled` status (as well as `pending-cancel`). It treats expired subscriptions as if they are no longer valid, which is what we need to bypass as a result of our plugin forcing all failed renewals to the expired status.

chickenn00dle:
Thanks for the review @dkoo! 

@laurelfulford, this got a thumbs up, but I know you were testing it on a clone site. Should I wait for you to finish testing before merging this one?

includes/plugins/class-teams-for-memberships.php
dkoo: Nice, thanks for verifying!

laurelfulford:
Thanks for checking @chickenn00dle! I didn't end up testing on the clone because I was a little leery about getting the site to the point I could test with it without causing other issues (hooking up Stripe in test mode and turning off Stop Emails). I installed the custom plugins the site had locally instead, and didn't spot any issues!

chickenn00dle:
Thanks @laurelfulford!

PR #4393, opened 2026-01-13
feat: enable authors filter - NPPM-2518

leogermani:
### All Submissions:

In https://github.com/Automattic/newspack-plugin/pull/1985 we introduced a filter by Authors to the Posts page.

But we soon had to disable it for sites using CoAuthors plus in https://github.com/Automattic/newspack-plugin/pull/2050.

The reason was that this would break the Posts screen for sites with too many authors. And that's because the way CAP fetches the list of authors, having to combine WP Users and Guest Authors (posts in the DB...)

We have been slowly deprecating Guest Authors in Newspack, and migrating sites to the new Guest Contributors user role, so it seems a good idea to re-enable this filter for all sites, but keeping it looking only for standard WP Users. 

This means that it will not display CAP's Guest Authors, but that's fine. The goal is that sites not using Guest Authors will still be able to use this filter.

1. Test that the filter works as expected
2. Confirm that filtering the author by clicking in the author name under the "Authors" column in the list also results in the correct item being selected in the filter
3. Stress test it in large sites -> I have tested this in a site with 4000+ authors in a database of more than 100,000 WP users and it worked fine.

### Other information:

### DIFF
--- a/includes/author-filter/class-author-filter.php
+++ b/includes/author-filter/class-author-filter.php
@@ -72,31 +72,24 @@ class Author_Filter {
 			return;
 		}
 
-		// Disable this for sites with co authors plus enabled until we fix its issue with large sites.
-		if ( self::is_coauthors_plus_enabled( $post_type ) ) {
-			return;
-		}
-
 		$type_object = get_post_type_object( $post_type );
 
 		if ( ! $type_object instanceof WP_Post_Type ) {
 			return;
 		}
 
+		$value_field = 'user_nicename';
+		$select_name = 'author_name';
+		$capability  = Guest_Contributor_Role::ASSIGNABLE_TO_POSTS_CAPABILITY_NAME; // This includes both Guest Contributors and regular Authors.
+		$users       = self::get_options_from_users( $capability );
+
 		if ( self::is_coauthors_plus_enabled( $post_type ) ) {
 			// When using Coauthors plugin, dont show the filter if user is browsing the "Mine" tab. It can lead to inconsistent UI.
 			if ( ! empty( $_GET['author'] ) && get_current_user_id() === (int) $_GET['author'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 				return;
 			}
-			$value_field = 'user_login';
-			$select_name = 'author_name';
-			$users       = self::get_options_from_coauthors();
-		} else {
-			$value_field = 'ID';
-			$select_name = 'author';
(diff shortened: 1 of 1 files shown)


includes/author-filter/class-author-filter.php
```diff
@@ -72,32 +72,25 @@ public static function author_filters_admin( $post_type ) {
...
-
 		$type_object = get_post_type_object( $post_type );
 
 		if ( ! $type_object instanceof WP_Post_Type ) {
 			return;
 		}
 
+		$value_field = 'user_nicename';
+		$select_name = 'author_name';
+		$capability  = Guest_Contributor_Role::ASSIGNABLE_TO_POSTS_CAPABILITY_NAME; // This includes both Guest Contributors and regular Authors.
```
miguelpeixe, line 83: This is restricting the dropdown to only render guest contributors. All other users are not appearing in the list (admins, editors, authors, contributors)

miguelpeixe (changes requested):

includes/author-filter/class-author-filter.php
leogermani: I don't see that.

All these roles also have this capability. See https://github.com/Automattic/newspack-plugin/blob/fa0113a33227edba01ddab62d3b4cfd964f1f8a8/includes/plugins/co-authors-plus/class-guest-contributor-role.php#L251

miguelpeixe: This is my dropdown in trunk:

Captura de Tela 2026-01-13 às 12 47 26

This branch:

Captura de Tela 2026-01-13 às 12 48 55

miguelpeixe: Running `wp option delete newspack_coauthors_plus_settings_version` to reset the setup fixed it. Not sure why my env had it borked.

image

miguelpeixe (approved):

includes/author-filter/class-author-filter.php
miguelpeixe: The issue is probably because I executed `wp role reset` at some point after the `newspack_coauthors_plus_settings_version` was set. This will remove the custom cap.

The [populate_roles()](https://github.com/WordPress/wordpress-develop/blob/35313faf1e455d8e63c7711048215b9f786b4a0b/src/wp-admin/includes/schema.php#L714-L743) function has no hook, so I think our best option is hooking to `update_option_{$role_key}` to avoid that issue.

leogermani: hmm that's a good idea

PR #4398, opened 2026-01-15
fix(content-gate): change 'paragraph' to 'block' in settings text

wil-gerken:
### All Submissions:

Updated terminology in the content gate settings panel from "paragraph" to "block" to better align with WordPress block terminology.

Since the Content Gate can be applied to any block type, not just paragraphs, this change makes the UI text more accurate and less confusing for users.

`NPPM-2492`

1. In wp-admin, go here: Audience > Configuration > Content Gating > Content Gate 
2. Click on Configure 
3. View the Settings in the sidebar to review the copy changes, which are a swap of "paragraph" for "block" on the label: "Default Block Count", and its help: "Number of blocks that readers can see above the content gate.",  plus the help on the More tag checkbox: "Override the default block count on pages where a “More” block has been placed." 

### Other information:

### DIFF
--- a/src/content-gate/editor/editor.js
+++ b/src/content-gate/editor/editor.js
@@ -190,16 +190,16 @@ function GateEdit() {
 					type="number"
 					min="0"
 					value={ meta.visible_paragraphs }
-					label={ __( 'Default paragraph count', 'newspack-plugin' ) }
+					label={ __( 'Default block count', 'newspack-plugin' ) }
 					onChange={ value => editPost( { meta: { visible_paragraphs: value } } ) }
-					help={ __( 'Number of paragraphs that readers can see above the content gate.', 'newspack-plugin' ) }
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
I'm not sure we should make this change. It counts only paragraphs when calculating the visible content for gated articles:

https://github.com/Automattic/newspack-plugin/blob/c8428a6d8bab7cdb17b04d13d93f9e218a8b0dea/includes/content-gate/trait-content-gate-layout.php#L163-L167

Changing the terminology can be misleading.

leogermani:
Removing the Needs review label until we know whether we want this change or not. I suggest surfacing this discussion in the related Linear issue

wil-gerken:
We've decided not to make this change at this time. The understanding is that the wording is actually accurate as it was.

PR #4399, opened 2026-01-15
feat(content-gate): create layout and improve asset management

miguelpeixe:
### All Submissions:

This PR implements a few things:

 - Allow the creation of gate layout posts for each gate mode (registration and custom access)
 - Move the layout block editor asset enqueueing strategy to the trait
 - Isolate the Memberships block editor asset so its own class enqueues it
 - Isolate the Metering block editor asset so its own class also enqueues it

1. Make sure you have Woo Memberships enabled and multiple plans configured
2. Edit the Memberships gate (Audience -> Configuration -> Content Gate)
3. Confirm every panel continues to work as expected
   1. Edit gates for differents plan and save
   2. Customize layout settings
   3. Customize metering settings
5. Create/edit a regular content gate (Audience -> Content Gates)
6. Enable both Registration and Paid access
7. Click "Edit layout" for each mode
8. Confirm each creates a draft with layout panels
9. Edit both layouts and confirm the values persist
10. Back in the "Content Gates" page, confirm that clicking "Edit layout" redirects to the created posts (and not new ones)

### Other information:

### DIFF
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -484,7 +484,7 @@ class Content_Gate {
 		if ( $gate_mode ) {
 			$url = add_query_arg( 'gate_mode', $gate_mode, $url );
 		}
-		return str_replace( site_url(), '', $url );
+		return \wp_make_link_relative( $url );
 	}
 
 	/**
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -49,8 +49,8 @@ class Content_Gate {
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_post_type' ] );
 		add_action( 'admin_init', [ __CLASS__, 'redirect_cpt' ] );
+		add_action( 'admin_init', [ __CLASS__, 'handle_edit_gate_layout' ] );
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
-		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
 		add_action( 'wp_footer', [ __CLASS__, 'render_overlay_gate' ], 1 );
 		add_filter( 'newspack_popups_assess_has_disabled_popups', [ __CLASS__, 'disable_popups' ] );
 		add_filter( 'newspack_reader_activity_article_view', [ __CLASS__, 'suppress_article_view_activity' ], 100 );
@@ -325,36 +325,6 @@ class Content_Gate {
 	}
 
 	/**
-	 * Enqueue block editor assets.
-	 */
-	public static function enqueue_block_editor_assets() {
-		if ( ! in_array( get_post_type(), self::get_gate_post_types(), true ) ) {
-			return;
-		}
-		\wp_enqueue_script(
-			'newspack-content-gate',
-			Newspack::plugin_url() . '/dist/content-gate-editor.js',
-			[],
-			filemtime( dirname( NEWSPACK_PLUGIN_FILE ) . '/dist/content-gate-editor.js' ),
-			true
-		);
-		\wp_localize_script(
-			'newspack-content-gate',
-			'newspack_content_gate',
-			[
-				'has_campaigns' => class_exists( 'Newspack_Popups' ),
-			]
-		);
-
-		\wp_enqueue_style(
-			'newspack-content-gate',
-			Newspack::plugin_url() . '/dist/content-gate-editor.css',
-			[],
-			filemtime( dirname( NEWSPACK_PLUGIN_FILE ) . '/dist/content-gate-editor.css' )
-		);
-	}
-
-	/**
 	 * Get the post ID of the custom gate.
 	 *
 	 * @param int $post_id Post ID to find gate for.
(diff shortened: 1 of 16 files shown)


includes/content-gate/class-metering.php
```diff
@@ -53,6 +54,17 @@ public static function restrict_post( $restrict ) {
 		return $restrict;
 	}
 
+	/**
+	 * Block editor assets for metering settings.
+	 */
+	public static function enqueue_block_editor_assets() {
+		if ( ! in_array( get_post_type(), Content_Gate::get_gate_post_types(), true ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
miguelpeixe: > but layout posts registered via the trait will still trigger this enqueue since they're registered as separate post types

This is intended. The memberships content gate is registered via the trait and has metering settings.

includes/content-gate/class-content-gate.php
```diff
@@ -497,6 +467,85 @@ public static function create_gate( $title = '', $post_type = self::GATE_CPT ) {
...
+		} else {
+			\wp_die( esc_html( __( 'Invalid gate mode.', 'newspack' ) ) );
+		}
+
+		$gate_layout = get_post( $gate_layout_id );
+		if ( $gate_layout ) {
+			if ( 'trash' === get_post_status( $gate_layout_id ) ) {
+				\wp_untrash_post( $gate_layout_id );
+			}
+			\wp_safe_redirect( \admin_url( 'post.php?post=' . $gate_layout_id . '&action=edit' ) );
```
dkoo: Non-blocking nit:

```suggestion
			\wp_safe_redirect( \get_edit_post_link( $gate_layout_id, 'edit' ) );
```

includes/content-gate/class-content-gate.php
```diff
@@ -497,6 +467,85 @@ public static function create_gate( $title = '', $post_type = self::GATE_CPT ) {
...
+			exit;
+		} else {
+			$gate_layout_id = self::create_gate( $gate_layout_default_title, self::GATE_LAYOUT_CPT );
+			if ( is_wp_error( $gate_layout_id ) ) {
+				\wp_die( esc_html( $gate_layout_id->get_error_message() ) );
+			}
+			$gate[ $gate_mode ]['gate_layout_id'] = $gate_layout_id;
+			self::update_gate_settings( $gate_id, $gate );
+			\wp_safe_redirect( \admin_url( 'post.php?post=' . $gate_layout_id . '&action=edit' ) );
```
dkoo: Non-blocking nit:

```suggestion
			\wp_safe_redirect( \get_edit_post_link( $gate_layout_id, 'edit' ) );
```

includes/content-gate/class-content-gate.php
```diff
@@ -497,6 +467,85 @@ public static function create_gate( $title = '', $post_type = self::GATE_CPT ) {
...
+
+		$gate_layout = get_post( $gate_layout_id );
+		if ( $gate_layout ) {
+			if ( 'trash' === get_post_status( $gate_layout_id ) ) {
+				\wp_untrash_post( $gate_layout_id );
+			}
+			\wp_safe_redirect( \admin_url( 'post.php?post=' . $gate_layout_id . '&action=edit' ) );
+			exit;
+		} else {
+			$gate_layout_id = self::create_gate( $gate_layout_default_title, self::GATE_LAYOUT_CPT );
```
dkoo: I think we should isolate two different methods for creating gate parent posts vs gate layout posts for the following reasons:

- The gate parent post doesn't need/shouldn't get any post content—this is relevant only for gate layout posts
- The gate layout doesn't need/shouldn't get a `gate_priority` meta value—this is relevant only for gate parent posts

dkoo (changes requested):
The reorganization of assets looks good! It makes sense to separate out the metering and Memberships-specific options into separate scripts for editing memberships gates.

I ran into some minor issues with new gate layouts being created when I didn't expect:

- I missed two instances of changing `gate_id` to `gate_layout_id` in the last PR, which affects this PR. On lines [412](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-layout-assets/includes/wizards/audience/class-audience-content-gates.php#L412) and [428](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-layout-assets/includes/wizards/audience/class-audience-content-gates.php#L428) of `includes/wizards/audience/class-audience-content-gates.php`, `gate_id` (in both key and value) needs to be `gate_layout_id`, otherwise when saving/updating a gate, the key will be dropped.

- Even after fixing the above, I notice that the `update_registration_settings` and `update_custom_access_settings` methods will update the gate layout IDs as passed in from the browser in addition to the other settings. I'm not sure this should be the case since the layout ID is calculated when clicking the "edit layout" buttons and the `update_` methods could be called without the correct calculated layout IDs. Perhaps before updating the post meta for `registration` or `custom_access`, we should check for an existing `gate_layout_id` and preserve it if it exists? For example, try the following:

  1. Create a new content gate, or edit a content gate that doesn't yet have any gate layouts for Registration or Paid Access
  2. Open the "Edit layout" button links in a new tab while keeping the **Audience > Content Gates** admin page open in the existing tab: the gate layout posts will be created because the gate post in the DB doesn't have any gate layouts yet.
  3. In the **Audience > Content Gates** admin page tab, click the "Update" button to save the gate post: the gate layout post IDs will be overwritten as `0` by the stale gate config object, so the next time you click "Edit layout" a new gate layout post will be created again.

- Non-blocking: I'm not sure gate layout posts need an editable post title, since it's never shown to readers. Are you thinking it's useful for admins to be able to tell which layout they're currently editing?

More inline:

miguelpeixe:
Thank you for the review, @dkoo!

> I missed two instances of changing gate_id to gate_layout_id in the last PR, which affects this PR. On lines [412](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-layout-assets/includes/wizards/audience/class-audience-content-gates.php#L412) and [428](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-layout-assets/includes/wizards/audience/class-audience-content-gates.php#L428) of includes/wizards/audience/class-audience-content-gates.php, gate_id (in both key and value) needs to be gate_layout_id, otherwise when saving/updating a gate, the key will be dropped.

Whoops! Good catch, fixed in 65a098cfb0a25df6b0e80bd6e97c99773c2d0f55

> Even after fixing the above, I notice that the update_registration_settings and update_custom_access_settings methods will update the gate layout IDs as passed in from the browser in addition to the other settings.

Good point. We should be extra safe here to avoid settings changes from overriding layout IDs. 904543b1d217ac007c5c96d2d7efe766aaf7661e implements this change, so the frontend doesn't include layout ID in the payload, and the API can safely update the settings without it.

> Non-blocking: I'm not sure gate layout posts need an editable post title, since it's never shown to readers. Are you thinking it's useful for admins to be able to tell which layout they're currently editing?

Agreed, we need to keep the title for our existing gates, but 76953c12b86f19ffd42830ed3c1239c20e1873ef removes it for the layout CPT.

This renders "No title" in the editor UI, which I don't really like, but it's better than a useless gigantic H1 in the editor

image

image

I looked a bit at how to customize the references above, but yet to find a solid solution. Using `editPost( { title: 'Something' } );` works, but leaves the editor in a permanent dirty state since the title prop can't be saved. I'm not a fan, so we can maybe tackle this separately later.

includes/content-gate/class-content-gate.php
miguelpeixe: Updated in 76953c12b86f19ffd42830ed3c1239c20e1873ef

I'm keeping the post content in the gate creation for now because of our current memberships gate.

dkoo (approved):
I see what you mean about the "No title" being displayed in the editor. I'm okay with reverting that change and keeping the title support for gate layout posts for now. If anything we could just use CSS to hide the title field like we sometimes do for newsletter CPTs, but it's a very minor thing.

Re-tested the issue with the `gate_layout_id` keys getting deleted upon updating, and it looks good now! Once the gate layout is created for a gate it's preserved even if you update the gate without that key in the passed config. 👍

miguelpeixe:
Thanks, @dkoo! I liked your idea better. ee0ef4c73b3643973ebe551d1880775307b13871 restores the title support and hides it with CSS. The CSS selector restricts to the layout CPT, so it should remain unchanged for the existing memberships gate.

dkoo (approved):
Nice! No editable title but I see the title in the editor breadcrumb and sidebar. 👍 

Image

PR #4405, opened 2026-01-17
feat(byline): add byline block with custom byline and CAP support

rbcorrales:
### All Submissions:

Adds a Byline block for block themes that displays post author attribution.

The block supports three authorship systems in priority order:
1. Newspack Custom Bylines
2. CoAuthors Plus (including guest contributors)
3. Default WordPress author

Block settings include prefix text and author archive linking. When custom bylines are active, these settings are hidden since custom bylines control their own formatting.

Only available for block themes. Classic themes continue using the `pre_newspack_posted_by` filter.

See the [README](https://github.com/Automattic/newspack-plugin/blob/feat/byline-block/src/blocks/byline/README.md) for additional info.

Related PR: https://github.com/Automattic/newspack-block-theme/pull/399

Closes NPPD-1122.

_**Note:** There's a known mismatch between the `newspack/avatar` and `newspack/byline` blocks when custom bylines are active. The avatar block currently shows the post author or CAP authors, but isn't aware of custom bylines. This means the avatar may display a different author than the byline text. This will be addressed in a follow-up PR by adding custom byline support to the avatar block._

1. Activate the [block theme](https://github.com/Automattic/newspack-block-theme).
2. Create a new post and add the new "Byline" block.
3. Verify it displays the post author name with "By " prefix and with a link to the author archive.
5. In the block settings test:
     - Toggling "Link to author archive" off.
     - Changing the prefix.

**Custom Bylines:**

6. Open the Bylines sidebar panel and enable custom bylines.
7. Add one or more authors to the custom byline.
8. Verify the byline block displays the custom byline content.
9. Verify prefix and link settings are hidden in the block inspector.
10. Disable custom bylines. Block should fall back to the default author.

**CoAuthors Plus (requires [plugin](https://github.com/Automattic/co-authors-plus)):**

11. Disable custom bylines if active.
12. In the "Authors" sidebar panel, add a second WP user as a co-author.
13. Verify byline displays both authors with "and" conjunction (e.g., "By Jane and John")
14. Add a third co-author. Verify the Oxford comma formatting (e.g., "By Jane, John, and Bob")
15. Create a guest contributor in CAP and assign to the post.
16. Verify that the guest also displays correctly and links to their archive page.

**Edge cases:**

17. Create a post with no author (author ID = 0). Block should render empty.
18. Insert block in a Query Loop. Verify it uses each post's author, not the current post.

**Unit tests:**

19. Run JS tests and PHP tests and verify that all pass.

### Other information:

### DIFF
--- a/src/blocks/byline/README.md
+++ b/src/blocks/byline/README.md
@@ -1,0 +1,62 @@
+# Byline Block
+
+A Gutenberg block that displays post author attribution with support for multiple authorship systems.
+
+## Overview
+
+The byline block automatically detects and displays the appropriate author information based on what's available for the post. It follows a priority-based approach to determine which byline source to use.
+
+## Byline priority
+
+The block checks for author information in the following order:
+
+1. **Newspack Custom Bylines** - If the custom byline feature is enabled for the post (`_newspack_byline_active` meta), the block displays the custom byline content. Custom bylines support free-text editing with embedded author links using shortcode syntax.
+
+2. **CoAuthors Plus** - If CoAuthors Plus is active and the post has authors assigned through CAP, the block displays all co-authors with proper formatting (e.g., "Author1, Author2, and Author3"). This works with both WordPress users and CAP Guest Authors.
+
+3. **Default WordPress Author** - Falls back to the standard WordPress post author when neither custom bylines nor CoAuthors Plus authors are available.
+
+## Block attributes
+
+| Attribute | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `prefix` | string | `"By "` | Text displayed before author names. Hidden in inspector when custom byline is active. |
+| `linkToAuthorArchive` | boolean | `true` | Whether to link author names to their archive pages. Hidden in inspector when custom byline is active. |
+
+When a custom byline is enabled, both settings are hidden in the block inspector since custom bylines control their own prefix and link behavior.
+
+## Editor behavior
+
+The block reads from the CoAuthors Plus data store (`cap/authors`) for real-time updates in the editor. When authors are added or removed via the Authors panel in the sidebar, the byline block updates immediately without requiring a page refresh.
+
+## Availability
+
+This block is only available for block themes. For classic themes, bylines are handled through the `pre_newspack_posted_by` filter in the Newspack theme.
+
+## Usage with block theme patterns
+
+The block theme post-meta patterns are configured to use this block when available:
+
+```html
+<!-- wp:newspack/byline {"prefix":"By "} /-->
+```
+
+For patterns that include avatars, use the `newspack/avatar` block as a sibling:
+
+```html
+<!-- wp:group {"layout":{"type":"flex"}} -->
+<div class="wp-block-group">
+  <!-- wp:newspack/avatar {"size":48} /-->
+  <!-- wp:newspack/byline /-->
+</div>
+<!-- /wp:group -->
+```
+
+## Developer notes
+
+For CAP authors, the block applies the `coauthors_posts_link` filter for each author link, matching CAP's own `coauthors_posts_links_single()` behavior.
+
+## Related
+
+- **Newspack Avatar Block** (`newspack/avatar`) - Displays author avatar(s) with support for CoAuthors Plus
+- **Newspack Bylines Feature** - The underlying custom bylines system in Newspack (`includes/class-bylines.php`)
--- a/src/blocks/byline/README.md
+++ b/src/blocks/byline/README.md
@@ -31,7 +31,7 @@ The block reads from the CoAuthors Plus data store (`cap/authors`) for real-time
 
 ## Availability
 
-This block is only available for block themes. For classic themes, bylines are handled through the `pre_newspack_posted_by` filter in the Newspack theme.
+This block replaces the Post Author block and is only available in block themes. For Newspack's classic theme, custom bylines are handled through the `pre_newspack_posted_by` filter, and CoAuthors Plus support is added through custom code.
 
 ## Usage with block theme patterns
 
--- a/src/blocks/byline/README.md
+++ b/src/blocks/byline/README.md
@@ -20,7 +20,7 @@ The block checks for author information in the following order:
 
 | Attribute | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `prefix` | string | `"By "` | Text displayed before author names. Hidden in inspector when custom byline is active. |
+| `prefix` | string | `"By"` | Text displayed before author names. A space is added automatically after the prefix. Hidden in inspector when custom byline is active. |
 | `linkToAuthorArchive` | boolean | `true` | Whether to link author names to their archive pages. Hidden in inspector when custom byline is active. |
 
 When a custom byline is enabled, both settings are hidden in the block inspector since custom bylines control their own prefix and link behavior.
@@ -38,7 +38,7 @@ This block is only available for block themes. For classic themes, bylines are h
 The block theme post-meta patterns are configured to use this block when available:
 
 ```html
-<!-- wp:newspack/byline {"prefix":"By "} /-->
+<!-- wp:newspack/byline {"prefix":"By"} /-->
 ```
 
 For patterns that include avatars, use the `newspack/avatar` block as a sibling:
--- a/src/blocks/byline/class-byline-block.php
+++ b/src/blocks/byline/class-byline-block.php
@@ -139,7 +139,7 @@ final class Byline_Block {
 		return sprintf(
 			'<div %1$s><span class="byline">%2$s%3$s</span></div>',
 			$wrapper_attributes,
-			esc_html( $prefix ),
+			! empty( $prefix ) ? esc_html( $prefix ) . ' ' : '',
 			$byline_content
 		);
 	}
@@ -230,7 +230,7 @@ final class Byline_Block {
 		return sprintf(
 			'<div %1$s><span class="byline">%2$s%3$s</span></div>',
 			$wrapper_attributes,
-			esc_html( $prefix ),
+			! empty( $prefix ) ? esc_html( $prefix ) . ' ' : '',
 			$author_html
 		);
 	}
@@ -238,7 +238,7 @@ final class Byline_Block {
 	/**
 	 * Get the translated prefix.
 	 *
-	 * The block.json default "By " is not translatable. This method checks if
+	 * The block.json default "By" is not translatable. This method checks if
 	 * the prefix matches the English default and returns the translated version.
 	 * If the user has set a custom prefix, it's returned as-is.
 	 *
@@ -248,8 +248,8 @@ final class Byline_Block {
 	 */
 	private static function get_translated_prefix( string $prefix ) {
 		// If prefix is empty or matches the English default, use translated version.
-		if ( empty( $prefix ) || 'By ' === $prefix ) {
-			return __( 'By ', 'newspack-plugin' );
+		if ( empty( $prefix ) || 'By' === $prefix ) {
+			return __( 'By', 'newspack-plugin' );
 		}
 		return $prefix;
 	}
--- a/src/blocks/byline/class-byline-block.php
+++ b/src/blocks/byline/class-byline-block.php
@@ -108,7 +108,7 @@ final class Byline_Block {
 	 * @return string The block HTML.
 	 */
 	private static function render_coauthors( array $attributes, array $coauthors ) {
-		$prefix             = $attributes['prefix'] ?? __( 'By ', 'newspack-plugin' );
+		$prefix             = self::get_translated_prefix( $attributes['prefix'] ?? '' );
 		$link_to_archive    = $attributes['linkToAuthorArchive'] ?? true;
 		$wrapper_attributes = get_block_wrapper_attributes( [ 'class' => 'wp-block-newspack-byline' ] );
 
@@ -121,14 +121,7 @@ final class Byline_Block {
 			}
 
 			if ( $link_to_archive ) {
-				// Use get_author_posts_url with user_nicename - CAP hooks the author_link filter
-				// via CoAuthors_Guest_Authors::filter_author_link() to construct proper URLs.
-				$author_url     = get_author_posts_url( $coauthor->ID, $coauthor->user_nicename );
-				$author_links[] = sprintf(
-					'<span class="author vcard"><a class="url fn n" href="%1$s">%2$s</a></span>',
-					esc_url( $author_url ),
-					esc_html( $display_name )
-				);
+				$author_links[] = self::get_coauthor_link( $coauthor, $display_name );
 			} else {
 				$author_links[] = sprintf(
 					'<span class="author vcard"><span class="fn n">%1$s</span></span>',
@@ -152,6 +145,53 @@ final class Byline_Block {
 	}
 
 	/**
+	 * Get the author link HTML for a CoAuthors Plus author.
+	 *
+	 * Uses get_author_posts_url() which CAP hooks via the 'author_link' filter
+	 * (CoAuthors_Guest_Authors::filter_author_link) to construct proper URLs for
+	 * both WordPress users and CAP guest authors.
+	 *
+	 * Also applies the 'coauthors_posts_link' filter for compatibility with CAP's
+	 * own coauthors_posts_links_single() function.
+	 *
+	 * @param object $coauthor     The coauthor object.
+	 * @param string $display_name The display name to show.
+	 *
+	 * @return string Author link HTML.
+	 */
+	private static function get_coauthor_link( $coauthor, string $display_name ) {
+		$args = [
+			'before_html' => '',
+			'href'        => get_author_posts_url( $coauthor->ID, $coauthor->user_nicename ),
+			'rel'         => 'author',
+			'class'       => 'url fn n',
+			'text'        => $display_name,
+			'after_html'  => '',
+		];
+
+		/**
+		 * Filter the author link arguments.
+		 *
+		 * This filter is provided by CoAuthors Plus and allows modification of
+		 * author link attributes. We apply it for full CAP compatibility.
+		 *
+		 * @param array  $args     Link arguments: href, rel, class, text, before_html, after_html.
+		 * @param object $coauthor The coauthor object.
+		 */
+		$args = apply_filters( 'coauthors_posts_link', $args, $coauthor );
+
+		return sprintf(
+			'%1$s<span class="author vcard"><a class="%2$s" href="%3$s" rel="%4$s">%5$s</a></span>%6$s',
+			$args['before_html'],
+			esc_attr( $args['class'] ),
+			esc_url( $args['href'] ),
+			esc_attr( $args['rel'] ),
+			esc_html( $args['text'] ),
+			$args['after_html']
+		);
+	}
+
+	/**
 	 * Render default WordPress author.
 	 *
 	 * @param array $attributes The block attributes.
@@ -160,7 +200,7 @@ final class Byline_Block {
 	 * @return string The block HTML.
 	 */
 	private static function render_default_author( array $attributes, int $post_id ) {
-		$prefix             = $attributes['prefix'] ?? __( 'By ', 'newspack-plugin' );
+		$prefix             = self::get_translated_prefix( $attributes['prefix'] ?? '' );
 		$link_to_archive    = $attributes['linkToAuthorArchive'] ?? true;
 		$wrapper_attributes = get_block_wrapper_attributes( [ 'class' => 'wp-block-newspack-byline' ] );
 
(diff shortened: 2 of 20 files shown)


tests/unit-tests/data-events.php
```diff
@@ -98,8 +98,14 @@ public function test_execute_queued_dispatches() {
 
 		$this->assertIsArray( $hook_request );
 		$this->assertIsArray( $hook_queued_dispatches );
-		$this->assertEquals( $action_name, $hook_queued_dispatches[0]['action_name'] );
-		$this->assertEquals( $data, $hook_queued_dispatches[0]['data'] );
+
+		// Find our test action (other events may be queued from other tests).
+		$matching = array_filter(
+			$hook_queued_dispatches,
+			fn( $d ) => $d['action_name'] === $action_name
```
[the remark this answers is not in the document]
rbcorrales: Thanks Copilot, but we currently support 8+, and you should too.

src/blocks/byline/class-byline-block.php
```diff
@@ -0,0 +1,281 @@
...
+		$args = apply_filters( 'coauthors_posts_link', $args, $coauthor );
+
+		return sprintf(
+			'%1$s<span class="author vcard"><a class="%2$s" href="%3$s" rel="%4$s">%5$s</a></span>%6$s',
+			$args['before_html'],
+			esc_attr( $args['class'] ),
+			esc_url( $args['href'] ),
+			esc_attr( $args['rel'] ),
+			esc_html( $args['text'] ),
+			$args['after_html']
```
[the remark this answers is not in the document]
rbcorrales: Addressed here: [1b50047](https://github.com/Automattic/newspack-plugin/pull/4405/commits/1b5004726c3a5399f326eb48126fbf4be9f234da).

src/blocks/byline/utils.js
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	// Add remaining text after last match.
+	if ( lastIndex < bylineContent.length ) {
+		const textAfter = bylineContent.slice( lastIndex );
+		elements.push( decodeHtmlEntities( textAfter ) );
+	}
+
+	return elements;
+}
```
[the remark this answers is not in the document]
rbcorrales: It's already checked, but adding the guard is harmless: [4f40717](https://github.com/Automattic/newspack-plugin/pull/4405/commits/4f40717b149cff58ff66bf1cda9b147044948429).

tests/unit-tests/bylines/class-test-byline-block.php
```diff
@@ -0,0 +1,378 @@
...
+				'post_status' => 'publish',
+			]
+		);
+
+		Bylines::register_post_meta();
+
+		if ( ! \WP_Block_Type_Registry::get_instance()->is_registered( 'newspack/byline' ) ) {
+			Byline_Block::register_block();
+		}
+	}
```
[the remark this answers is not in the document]
rbcorrales: I think this is a standard test pattern. `WP_UnitTestCase` handles cleanup via DB transaction rollback after each test. Using `setUpBeforeClass()` would add complexity without meaningful benefit. The full suite runs in 0.2s

tests/unit-tests/bylines/class-test-byline-block.php
```diff
@@ -0,0 +1,378 @@
...
+	public function tear_down(): void {
+		delete_post_meta( self::$post_id, Bylines::META_KEY_ACTIVE );
+		delete_post_meta( self::$post_id, Bylines::META_KEY_BYLINE );
+
+		if ( \WP_Block_Type_Registry::get_instance()->is_registered( 'newspack/byline' ) ) {
+			unregister_block_type( 'newspack/byline' );
+		}
+
+		parent::tear_down();
+	}
```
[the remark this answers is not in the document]
rbcorrales: `WP_UnitTestCase` automatically rolls back database transactions after each test. Manual cleanup isn't needed

src/blocks/byline/utils.js
```diff
@@ -0,0 +1,122 @@
...
+		const name = author.display_name || author.name;
+
+		return createElement(
+			'span',
+			{ key: `author-wrapper-${ author.id || partIndex }`, className: 'author vcard' },
+			linkToArchive
+				? createElement( 'a', { href: '#author-link', onClick: e => e.preventDefault(), className: 'url fn n' }, name )
+				: createElement( 'span', { className: 'fn n' }, name )
+		);
+	} );
```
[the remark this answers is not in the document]
rbcorrales: `Intl.ListFormat` has 95%+ global browser support ([see caniuse.com](https://caniuse.com/?search=Intl.ListFormat)). The block editor itself requires modern browsers. Adding a large fallback for this use case seems like overkill.

src/blocks/byline/utils.js
```diff
@@ -0,0 +1,122 @@
...
+	const listFormatter = new Intl.ListFormat( locale, { style: 'long', type: 'conjunction' } );
+
+	// Create placeholder strings to get the formatted parts.
+	const placeholders = authors.map( ( _, i ) => `__AUTHOR_${ i }__` );
+	const formattedParts = listFormatter.formatToParts( placeholders );
+
+	// Build React elements from the formatted parts.
+	return formattedParts.map( ( part, partIndex ) => {
+		if ( part.type === 'literal' ) {
+			return part.value;
```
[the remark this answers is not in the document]
rbcorrales: Mixing strings and React elements in an array does not cause key warnings. Keys are only required for React elements, not primitives, and the elements here already have them.

src/blocks/byline/utils.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Decode HTML entities in a string.
+ *
+ * @param {string} text Text with HTML entities.
+ * @return {string} Decoded text.
+ */
+export function decodeHtmlEntities( text ) {
+	const textarea = document.createElement( 'textarea' );
+	textarea.innerHTML = text;
+	return textarea.value;
```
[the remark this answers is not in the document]
rbcorrales: I tried DOMParser, but it normalizes whitespace per HTML parsing rules, breaking strings like " and " (leading space gets trimmed). The textarea approach is the appropriate pattern. Setting `innerHTML` on a textarea that isn't attached to the DOM doesn't execute scripts and preserves whitespace. Added a guard clause for empty input and a comment documenting why DOMParser isn't suitable here. [d7a5da0](https://github.com/Automattic/newspack-plugin/pull/4405/commits/d7a5da0472581b7f87971729eab0a0d6d8648214)

thomasguillot:
LGTM 👍

src/blocks/byline/edit.jsx
```diff
@@ -0,0 +1,249 @@
...
+								<span className="fn n">{ defaultAuthor.name }</span>
+							) }
+						</span>
+					</span>
+				</div>
+			</>
+		);
+	}
+
+	// Fallback - no author found.
```
laurelfulford: It looks like this is the version of the block that's used in the Site Editor because it doesn't have a post ID. The placeholder text is fine, but it'd be nice to have the Setting panel there all the time (either in the fallback version or a Site-Editor specific version) 🙂 I think that's where we'll use this block the most.

src/blocks/byline/class-byline-block.php
```diff
@@ -0,0 +1,281 @@
...
+	 * If the user has set a custom prefix, it's returned as-is.
+	 *
+	 * @param string $prefix The prefix attribute value.
+	 *
+	 * @return string The translated or custom prefix.
+	 */
+	private static function get_translated_prefix( string $prefix ) {
+		// If prefix is empty or matches the English default, use translated version.
+		if ( empty( $prefix ) || 'By ' === $prefix ) {
+			return __( 'By ', 'newspack-plugin' );
```
laurelfulford: Normally [we try to avoid leading/trailing strings in translations](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#best-practices-for-writing-strings), just because it means something can break if the space isn't used in the translation, too. 

In this specific case, including the space in the prefix means publishers will always have to remember to do that when they edit it.

What do you think about adding the space when the byline is all put together, like [here](https://github.com/Automattic/newspack-plugin/pull/4405/changes#diff-9fe582f920cdadbc48deb6be16e595272e147552b291687d6e985b4faede9e39R140)?

src/blocks/byline/class-byline-block.php
```diff
@@ -0,0 +1,281 @@
...
+	 * Render custom byline.
+	 *
+	 * @param array  $attributes    The block attributes.
+	 * @param string $custom_byline The custom byline HTML from Bylines class.
+	 *
+	 * @return string The block HTML.
+	 */
+	private static function render_custom_byline( array $attributes, string $custom_byline ) {
+		// Custom bylines already include their prefix text, so we ignore the prefix attribute.
+		$wrapper_attributes = get_block_wrapper_attributes( [ 'class' => 'wp-block-newspack-byline' ] );
```
laurelfulford: I ran into this accidentally while testing adding a Custom CSS class to the block: this `[ 'class' => 'wp-block-newspack-byline' ]` that's added to the wrapper attributes a couple times makes it so the class is added twice when you add a custom CSS class. The block should add the class automagically, so you should be able to get rid of all of these!

src/blocks/byline/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+When a custom byline is enabled, both settings are hidden in the block inspector since custom bylines control their own prefix and link behavior.
+
+## Editor behavior
+
+The block reads from the CoAuthors Plus data store (`cap/authors`) for real-time updates in the editor. When authors are added or removed via the Authors panel in the sidebar, the byline block updates immediately without requiring a page refresh.
+
+## Availability
+
+This block is only available for block themes. For classic themes, bylines are handled through the `pre_newspack_posted_by` filter in the Newspack theme.
```
laurelfulford:
```suggestion
This block is intended to be used instead of the Post Author block in block themes. For Newspack's classic theme, custom bylines are handled through the `pre_newspack_posted_by` filter, and CoAuthors Plus are added through custom code.
```

Nit: I'm not sure I made this bit any clearer, but I removed some misleading bits: you can still technically insert this block in the content of a classic theme, and CoAuthors don't use `pre_newspack_posted_by`.

src/blocks/byline/edit.jsx
```diff
@@ -0,0 +1,249 @@
...
+						<TextControl
+							__nextHasNoMarginBottom
+							label={ __( 'Prefix', 'newspack-plugin' ) }
+							help={ __( 'Text displayed before the author name(s).', 'newspack-plugin' ) }
+							value={ attributes.prefix }
+							onChange={ prefix => setAttributes( { prefix } ) }
+						/>
+						<ToggleControl
+							__nextHasNoMarginBottom
+							label={ __( 'Link to author archive', 'newspack-plugin' ) }
```
laurelfulford: It's a bit of an odd case, but when you use an actual author in the Custom Byline, it is linked and you can't turn it off like you can when the byline isn't customized. 

(Feel free to ignore this one - it's one of those things that's bugging my brain so I had to write it down, but I'm not sure we've even had publishers ask to not link authors to archive pages so it's probably worth leaving as is until someone asks for it. If someone does need to do it, they can also just not use the tags!).

src/blocks/byline/class-byline-block.php
```diff
@@ -0,0 +1,281 @@
...
+			return '';
+		}
+
+		if ( 1 === count( $author_links ) ) {
+			return $author_links[0];
+		}
+
+		// Use wp_sprintf_l for localized list formatting.
+		// The %l placeholder formats the array as a proper list with commas and "and".
+		return wp_sprintf_l( '%l', $author_links );
```
laurelfulford: TIL about `wp_sprintf_l()`!

I'm a little worried about the use of an Oxford comma in a byline - even publishers who've battled us for AP date formatting have never complained about how multi-author bylines are formatted. It will also make the bylines in the theme inconsistent with the ones used in the Content Loop and Content Carousel blocks. 

It looks like we should be able to filter the `between_last_two` string to remove it: https://developer.wordpress.org/reference/functions/wp_sprintf_l/

laurelfulford (changes requested):
Ah this is so cool, @rbcorrales! I took the block for a test drive and hammered it a bit. It works really well overall! I found some odds and ends but I think they're all pretty minor. Most are in the code, except this random edge case:

When trying to test 17 with CoAuthor Plus enabled, I found if you add a Query Loop to a post (without changing the post's author at all), then add the Byline block to the Query Loop, it'll add an author to the post for each story:

Image

Just let me know if you have any questions about any of these, and feel free to bump some to a future PR!

src/blocks/byline/class-byline-block.php
rbcorrales: Good idea! Moved the space to assembly time so it's not an issue for translations: [2609a2e](https://github.com/Automattic/newspack-plugin/pull/4405/commits/2609a2e9ef488cfc8658310f4982311e885a0dad).

src/blocks/byline/README.md
rbcorrales: > you can still technically insert this block in the content of a classic theme

This is partially true, because the block PHP inclusion is currently wrapped in a `wp_is_block_theme()` check:

```php
if ( wp_is_block_theme() ) {
    require_once NEWSPACK_ABSPATH . 'src/blocks/avatar/class-avatar-block.php';
    require_once NEWSPACK_ABSPATH . 'src/blocks/byline/class-byline-block.php';
}
```

So, while reviewing this, I noticed the block was still appearing in the editor for classic themes (the JS registration wasn't gated), even though the block wouldn't be rendered on the frontend (same for the `newspack/avatar` block).

I updated that too to make it consistent, so now the block won't show up in the inserter at all for classic themes. Although it would show the `Your site doesn't include support for the "newspack/byline" block...` type of warning in the editor if you previously added the block and switched back to a classic theme.

Also updated the README to mix your wording and it now look like this:
> This block replaces the Post Author block and is only available in block themes. For Newspack's classic theme, custom bylines are handled through the `pre_newspack_posted_by` filter, and CoAuthors Plus support is added through custom code.

WDYT?

[bc29165](https://github.com/Automattic/newspack-plugin/pull/4405/commits/bc291655a1c0acacaea94a5caea13b8ac29263eb)

src/blocks/byline/edit.jsx
rbcorrales: Added the settings panel to the fallback view: [07e2f04](https://github.com/Automattic/newspack-plugin/pull/4405/commits/07e2f0483e2f48375d4314e3ae28c770013c873d).

rbcorrales: Also refactored the `edit.jsx` file into multiple files for readability ([d60851b](https://github.com/Automattic/newspack-plugin/pull/4405/commits/d60851b2081f722cc28ec493002d7cdf8d9efb48)) and added unit tests for those too ([dec45da](https://github.com/Automattic/newspack-plugin/pull/4405/commits/dec45daecfe5e5bcdee82ff800050e604b2e0e65)).

src/blocks/byline/class-byline-block.php
rbcorrales: Thanks for flagging this! Fixed: [8b0a842](https://github.com/Automattic/newspack-plugin/pull/4405/commits/8b0a8429eba6737a36f1e845f398cae69a9f1f80).

src/blocks/byline/class-byline-block.php
rbcorrales: Yeah, I learned about `wp_sprintf_l()` I think last year 😄

Looked at how the content loop renders it (`newspack_blocks_format_byline`) without the Oxford comma, and switched our byline block approach from `wp_sprintf_l` and `Intl.ListFormat` to a simplified manual formatting to match, to reduce the overhead, since adding a filter would affect every call on the site to `wp_sprintf_l` under the hood (unless we get hacky with filters and flags).

Both PHP and JS now use `_x( ' and ', 'post author separator', 'newspack-plugin' )` for i18n consistency.

Changes here: [40496ca](https://github.com/Automattic/newspack-plugin/pull/4405/commits/40496ca0b4e753c51c42672951f98b0fa22b0e21).

src/blocks/byline/edit.jsx
rbcorrales: Yeah, we already hide the link toggle when a custom byline is active, so users won't see a broken control. Since it's a limitation of the Bylines system itself, and users can work around it by not using author tags, I think we can leave it for now. Happy to revisit if publishers request it.

rbcorrales:
@laurelfulford, thanks for the thorough review! I addressed most of the feedback items, but still need to take a deeper look at the main one from your screenshot.

src/blocks/byline/edit.jsx
laurelfulford: I'm good with this - marking as resolved! 🙂

src/blocks/byline/edit.jsx
laurelfulford: Thanks @rbcorrales! 

Both the controls work well from the Site Editor to change the settings, and splitting up the files was smart! 

The only other feedback I have on this one is that it felt a little strange not to get the same visual feedback for toggling on/off links and changing the byline prefix. What do you think about wrapping the "No author" text in a link when applicable, and show/update the prefix: helpful, or would it just make this placeholder state weird?

src/blocks/byline/class-byline-block.php
laurelfulford: Looks good - thanks!

src/blocks/byline/class-byline-block.php
laurelfulford: Looks good!

src/blocks/byline/README.md
laurelfulford: > This is partially true, because the block PHP inclusion is currently wrapped in a wp_is_block_theme() check...

You can tell I'm just testing the editor and not publishing anything 😆 I've confirmed you can't add this block to the editor of classic theme anymore!

> Also updated the README to mix your wording and it now look like this

Ah that's much better! 

Both of these changes look good!

src/blocks/byline/class-byline-block.php
laurelfulford: Thanks @rbcorrales! This looks good in my testing, and good call on going a non-filter route. 🙂 

Marking this one as resolved!

rbcorrales:
@laurelfulford Okay, I was able to reproduce the duplication bug. A Byline block inside a Query Loop caused authors to accumulate incorrectly, corrupting both the loop bylines and the main post's byline.

The root cause was that CAP's author data store is designed for a single-post editing context. It ignores the `postId` parameter and returns all authors it knows about, which causes accumulation when multiple Byline blocks exist (one per post in the loop).

So, I implemented a hybrid approach in `useCoAuthors`:
- The post being currently edited keeps using CAP's JS store (to preserve real-time updates when changing authors)
- Posts inside a Query Loop now use `newspack_author_info` from the REST API

So now Query Loop posts display their correct CAP authors in the editor, and the post being edited still gets real-time updates.

Changes: [82f4af2](https://github.com/Automattic/newspack-plugin/pull/4405/commits/82f4af24dc5de5153e63ffb6d266a6f5f4c792ea) and [785fa9e](https://github.com/Automattic/newspack-plugin/pull/4405/commits/785fa9e4e151a1cc1c2f550a2261d3f099bd2f3e).

src/blocks/byline/edit.jsx
rbcorrales: Good point @laurelfulford . 'No author' is a placeholder that users shouldn't see normally, other than when editing the site.

Adding the prefix and link would be technically consistent but might make the placeholder feel like real content and potentially be confusing (like "is 'No author' an actual author on my site?").
I kind of see valid arguments for both directions.

What about using something different from "No author", like something that looks like an actual placeholder? Like [Author]. Then we could render it as "By [Author]" with a link if that option is enabled.

But happy to add it as you suggested if you feel strongly though!

laurelfulford: @rbcorrales I agree that changing the placeholder text will help! I was going to say something like `by Author Placeholder` but I prefer `[Author]` -- it's a little clearer that it's not what's going to show for real 🙂

laurelfulford:
> Okay, I was able to reproduce the duplication bug.

Nice! 🙌 I tried a combination of with and without CoAuthors Plus, with and without some Custom Bylines in the mix, and editing a post's authors after adding a Query Block to the content... I think this all looks good! 

I think it's just the placeholder tweak and that was it for my feedback here! 🎉

src/blocks/byline/edit.jsx
rbcorrales: Changed the placeholder to `[Author]` and added the prefix/link settings so you get visual feedback when toggling them in the Site Editor. Also removed the custom underline styles so the theme controls link styling between the editor and the frontend consistently (it was rendering link underlines only on hover in the site editor).

[aad9f89](https://github.com/Automattic/newspack-plugin/pull/4405/commits/aad9f89f6e7fd405b56d79d98432f7f3001a30b3)

laurelfulford: That's perfect! 🙌 

And thanks for catching & fixing the link style difference!

laurelfulford (approved):
🚢 This is looking awesome, @rbcorrales! Thanks for addressing all my nitpicking!

PR #4409, opened 2026-01-20
feat: Add My Account button block

laurelfulford:
### All Submissions:

This PR, along with https://github.com/Automattic/newspack-block-theme/pull/400 and https://github.com/Automattic/newspack-theme/pull/2618 adds a My Account button block to the Newspack Plugin, adds the button to just one pattern (for now) in the Block Theme (for testing), and moves some CSS from the Newspack Plugin to the classic theme to avoid clashing. 

I have some open questions/possible weird cruft to still deal with:

* For the block itself, it's a pretty basic button-style block with an editable label and an icon. You can edit most things you can edit with a regular button. 
* It does _not_ have unique state for mobile - I was wondering if that should be something we do based on the screen breakpoint, or something we do in the block settings (eg. have a My Account button block style that shows only the icon that can be used on mobile?) cc @thomasguillot 
* Some of the logic for the block -- specifically the trigger that opens the sign-in modal and the JavaScript that updates the button label based on whether you're a reader -- still lives in the RAS code outside of the block. I think this is okay but I might be wrong!
* Rather than hiding the block when you're logged in as an admin/not a reader, I gave it a 'disabled' state. I find it kind of confusing that the button isn't visible to admins in the classic theme, and I think that's how we end up with it looking so weird in some publisher headers: they're always logged in so they don't see it. 
* I've only added the block to the default header pattern in the theme to start, and added a check for RAS before it gets inserted. Once I nail down the mobile bit, I can add it to the rest!

Some of these (like the mobile state, and maybe adding an 'Outline' style) can also be handled in a separate PR if there isn't a for-sure direction to follow.

See NPPD-1123.

1. Start with a test site running RAS.
2. Apply this PR, https://github.com/Automattic/newspack-block-theme/pull/400 and https://github.com/Automattic/newspack-theme/pull/2618.
3. Confirm the default header has the My Account button in the header, alongside the Donate Block (note: if you've edited the Header Desktop template part, you'll need to reset it)
4. In an incognito window, click the button and confirm it triggers the My Account login window; once you sign up/sign in, the button should switch to say My Account.
5. In your regular window (as an admin) the button should still be visible but slightly greyed out:

CleanShot 2026-01-20 at 16 28 15

6. Open the Site Editor, and edit the Header (Desktop) template part.
7. When you click the My Account button, you should have an option to switch between Signed In and Signed Out states. 

CleanShot 2026-01-20 at 16 28 52

8. In each state, try editing the button text to change it to something else.
9. Try changing some other block options: colours, font size, border radius, padding...
10. Save and publish your changes.
11. Confirm they display on the front end, and that your labels both appear for your incognito reader:

CleanShot 2026-01-20 at 16 33 26@2x

12. Test these PRs on a site running the Classic theme. It should look/act the same -- I mostly wanted to move the styles so we wouldn't have to be overwriting too much classic-theme specific stuff in the block.
13. Lastly, if possible test these PRs on a site without RAS and confirm that the My Account button isn't inserted.

### Other information:


thomasguillot:
The button should not disappear or be disabled. It should then link to My account.

Screenshot 2026-01-21 at 12 25 51@2x

(ignore the change of style in the screenshot above)

laurelfulford:
Thanks @thomasguillot!

> The button should not... be disabled. It should then link to My account.

The logic in the RAS code only switches the button to the signed in state when you're a registered reader and not when you're logged in. I can do this differently in the block and just have it show the signed in label for anyone who's logged in, but I didn't know the history of that choice re: making it reader specific. Do you?

thomasguillot:
> The logic in the RAS code only switches the button to the signed in state when you're a registered reader and not when you're logged in

Isn't it the same thing? 😕

thomasguillot:
@laurelfulford I pushed some changes. Basically instead of having yet another custom button block, I think we should leverage Core's button block markup (note: I haven't looked at the editor and it might be broken)

laurelfulford:
Thanks @thomasguillot! 

> Basically instead of having yet another custom button block, I think we should leverage Core's button block markup

That's a good call! I'll take a look at the editor... 

> Isn't it the same thing? 😕

Not quite - it distinguishes between someone who's been registered through reader registration vs. other WordPress roles (like editor or administrator). 

Right now the button's hidden if you're logged into the other roles (unless you're in the Customizer) so what it shows to the publisher is kind of moot in the classic theme. We could just make it visible and treat all logged in users the same if that makes more sense to you? I can ask the wider team if this would cause issues for some reason -- I don't remember the decision making behind this.

laurelfulford:
@thomasguillot So far it sounds like the main concern was creating kind of a dead end for admins/editors since they don't need to use /my-account.

What would be your preference for how it behaves for logged in publishers: look/act the same as for readers (say "My Account" and link to /my-account) or do something else that (somehow) indicates it's for readers. I'm thinking like how the Newspack Subscription block doesn't let you subscribe as a admin/editor 🤔

thomasguillot:
> Look/act the same as for readers (say "My Account" and link to /my-account)

This 👍

laurelfulford:
Thanks @thomasguillot! I've made the following changes:

* I tweaked the editor markup so it matches the front-end structurally and visually
* I removed the 'disabled' styles from the block
* I made it so the button block always goes to /my-account when the user is logged in, including editorial roles

One last thing from me: what do you think we should do for mobile? I was thinking either make the button only show the icon on small screens, or have an icon-only style as an option that can be added to the mobile header. Or maybe a third option - do you have any preferences?

src/blocks/my-account-button/class-my-account-button-block.php
```diff
@@ -0,0 +1,183 @@
...
+
+		$labels = [
+			'signedin'  => $attrs['signedInLabel'],
+			'signedout' => $attrs['signedOutLabel'],
+		];
+
+		$extra_classes = [
+			'wp-element-button',
+			'wp-block-button__link',
+		];
```
rbcorrales, line 129: The `$extra_classes` array is missing the `newspack-reader__account-link` class. This class is required for the JS handler in `reader-activation-auth/index.js` to find the button and update its label when the reader's authentication state changes.

Without this class, the `handleReaderChanges()` function won't be able to dynamically switch the label from "Sign in" to "My Account" when a reader logs in (without a full page reload).

Suggested fix:
```suggestion
		$extra_classes = [
			'wp-element-button',
			'wp-block-button__link',
			'newspack-reader__account-link',
		];
```

src/blocks/my-account-button/class-my-account-button-block.php
```diff
@@ -0,0 +1,183 @@
...
+			[
+				'class' => implode( ' ', $extra_classes ),
+				'href'  => \esc_url_raw( $href ),
+			]
+		);
+
+		$link = '<div class="' . \esc_attr( implode( ' ', $wrapper_div_classes ) ) . '">';
+		$link .= '<div class="wp-block-button">';
+		$link .= '<a ' . $wrapper_attributes . ' data-labels="' . \esc_attr( \wp_json_encode( $labels ) ) . '" ' . $should_modal_trigger . '>';
+		$link .= '<span class="wp-block-newspack-my-account-button__icon">';
```
rbcorrales: The icon span is missing `aria-hidden="true"` for accessibility. The editor version in `edit.js` correctly includes this attribute, but the server-rendered output doesn't.

Suggested fix:
```suggestion
		$link .= '<span class="wp-block-newspack-my-account-button__icon" aria-hidden="true">';
```

This would help screen readers skip the decorative icon and only announce the button label text.

tests/unit-tests/class-test-my-account-button-block.php
rbcorrales, line 1: The test coverage could be expanded a bit. Currently, there's only one test case.
A few ideas:
- Signed-in user rendering: Verifying that the button links to `/my-account` and doesn't have the modal trigger attribute.
- RAS disabled: Verifying the block returns an empty string.
- Empty label: Verify the block returns empty when the label is blank.
- Missing WooCommerce account URL: Verifying signed-in users don't see a broken button

src/reader-activation-auth/index.js
rbcorrales, line 111: Minor note: The `handleReaderChanges()` function updates the label for both auth states, but only sets the href for signed-in:

```php
if ( reader?.authenticated ) {
    link.setAttribute( 'href', newspack_ras_config.account_url );
}
// No else branch to reset href to "#" for signed-out
```

This isn't a practical bug though. The current My Account logout flow requires navigation to `/wp-login.php?action=logout` which triggers a full page reload, so PHP always renders the correct signed-out state fresh. There's no code path where `setAuthenticated(false)` is called client-side.

Just flagging for awareness. No action needed unless we plan to support client-side logout in the future.

rbcorrales (changes requested):

src/blocks/my-account-button/class-my-account-button-block.php
laurelfulford: Thanks @rbcorrales! Not sure how I missed that it wasn't updating immediately. Fixed in [28483ea](https://github.com/Automattic/newspack-plugin/pull/4409/commits/28483eac6b95cccc7d47ddd6411e0dc717f312bd).

src/blocks/my-account-button/class-my-account-button-block.php
laurelfulford: D'oh 🤦‍♀️ Fixed in [6ad8233](https://github.com/Automattic/newspack-plugin/pull/4409/commits/6ad8233f20e511c39eba9975dbba9a90e1438550) -- thanks for catching that!

tests/unit-tests/class-test-my-account-button-block.php
laurelfulford: Thanks @rbcorrales! 

I added most of the suggested tests in [6cd316e](https://github.com/Automattic/newspack-plugin/pull/4409/commits/6cd316e92ebc8b561983320aae6068ea11df7d21), [44b18d1](https://github.com/Automattic/newspack-plugin/pull/4409/commits/44b18d11cac24c3126c497fc124c684863edc4c7), and [dfeb6de](https://github.com/Automattic/newspack-plugin/pull/4409/commits/dfeb6de3cef2df2450d1e104a59d1942e0654e38).

I don't have context for the 'why', but it looks like [we're always setting RAS as 'enabled' for test environments](https://github.com/Automattic/newspack-plugin/blob/c58fd3211ebf0894b87866529b37e93159f869e5/includes/reader-activation/class-reader-activation.php#L950-L968), so I've left the check for RAS being disabled/enabled out for now.

src/reader-activation-auth/index.js
laurelfulford: That's a good point, @rbcorrales! I think it's safe to leave since we're trying to keep sessions logged in, I think we're unlikely to add a front-end log out button. But it's a good thing to keep in mind in case that changes 🙂

rbcorrales (approved):
Thanks for addressing all the comments. The button is looking great!

laurelfulford:
Thanks for the review, @rbcorrales! 🙌

PR #4420, opened 2026-01-23
fix(content-gate): persist restricted content

miguelpeixe:
### All Submissions:

Filters applied to `the_content` may restore the original content to a restricted post. This ensures the restricted content persists.

1. Without Woo Memberships, configure a content gate (Audience -> Content Gates) with inline layout
2. While in trunk, visit a restricted post and confirm the gate never renders, and the original post is displayed
3. Checkout this branch, run `wp cache flush`, and refresh the page
4. Confirm the content is restricted, and the gate now renders in the expected position
5. Change the gate layout to overlay and confirm it renders as expected and the content is also restricted

### Other information:

### DIFF
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -165,7 +165,7 @@ class Content_Gate {
 
 		$content = self::get_restricted_post_excerpt( $post );
 
-		$post->post_content   = $content . self::get_inline_gate_content();
+		$post->post_content   = $content . self::get_inline_gate_html();
 		$post->post_excerpt   = $content;
 		$post->comment_status = 'closed';
 		$post->comment_count  = 0;
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -44,6 +44,13 @@ class Content_Gate {
 	public static $valid_gate_post_statuses = [ 'publish', 'draft', 'pending', 'future', 'private', 'trash' ];
 
 	/**
+	 * Restricted content per post ID.
+	 *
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
This isn't quite working for me. Caching the content does seem to work well, but it also causes inline gate layout post content to get rendered without content filtering, meaning the raw block markup is rendered instead of the transformed HTML.

Separately, not due to anything in this PR, if both Registration and Paid Access layouts are enabled, visiting a restricted post will cause the Paid Access layout to always appear and never the Registration layout. This might be intentional, but I'm not sure it should be the case.

#4421 proposes some fixes which may address both of these issues, but let me know if it warrants further discussion.

miguelpeixe:
Thanks for the review, @dkoo!

I could confirm the issue with markup, but I'm opting to use the existing `get_inline_gate_html()` method to preserve the original method and avoid applying the filter twice. 289c6da6a7112e4e57e7df2f17645b83dc04db97

Regarding the layout order, I agree!

dkoo:
That's a better solution, I forgot about that other method! I closed #4421 since you addressed the rendering issue with 289c6da—do you want to reorder the gate layout logic in this PR, too?

miguelpeixe:
Thanks, @dkoo!

Since it's unrelated to this issue, we can address the ordering separately if that's ok.

dkoo (approved):

dkoo:
> Since it's unrelated to this issue, we can address the ordering separately if that's ok.

@miguelpeixe I opened #4426 to implement this fix without the extra `newspack_gate_content` filter.

PR #4427, opened 2026-01-26
fix(my-account): "order again" checkout redirect

miguelpeixe:
### All Submissions:

NPPM-2561

Since https://github.com/Automattic/newspack-blocks/pull/2141, the `newspackOpenModalCheckout()` function allows the incoming URL to set the modal destination. This is to support different checkout endpoints, like when changing the payment method.

This breaks the modal for the "Order again" button because it redirects to the cart page, which is not supported in the modal.

This PR fixes the issue by introducing a `redirectToResponseUrl` bool to the button registration, allowing the "Order again" button to not set the redirect URL. In this case, it'll resort to the default checkout URL.

1. As a reader, do a one-time donation
2. Visit the order page and click "Order again"
3. Confirm you get an error:

image

4. Checkout this branch, refresh the order page, and click the "Order again" button
5. Confirm the modal renders without error,s and you are able to complete the order

### Other information:

### DIFF
--- a/src/my-account/v1/modal-checkout.js
+++ b/src/my-account/v1/modal-checkout.js
@@ -55,10 +55,17 @@ domReady( () => {
 	 */
 	const orderAgain = document.querySelectorAll( 'p.order-again a' );
 	orderAgain.forEach( button => {
-		registerModalCheckoutButton( button, null, 'order_again', data => {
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works!

PR #4431, opened 2026-01-27
fix: add fallback selector for the content gate in block theme

laurelfulford:
### All Submissions:

Both styles of content gate seem to work well out of the box with the block theme. So far, the only thing I've noticed is missing is some kind of fallback for the `entry` selector.

The Post Content block (part of the Query Block) [also uses `.entry-content` as a CSS class](https://github.com/WordPress/gutenberg/blob/ba0dc9a7c83078609de5b6e51f10263d8ba11ae8/packages/block-library/src/post-content/index.php#L63), so the main selector is available.

The `#content` ID that's used as a fallback isn't part of the block theme. Instead of trying to add it, I added a second fallback using `#wp--skip-link--target`, which is added to the `<main>` element by [some JS in WordPress](https://github.com/WordPress/wordpress-develop/blob/0b50baa5d164bedd17d3ae9e2eafc6286d6255b6/src/wp-includes/theme-templates.php#L172-L210), at roughly the same level as `#content`. 

See NPPD-1126

1. Apply this PR & run `npm run build`.
2. Test on a site running our block theme.
3. Set up a Content Gate (either using Memberships or Newspack gates), and set the gate to be an overlay.
4. Test in an incognito window, and confirm the gate is triggered.
5. Edit line 246 in the src/content-gate/gate.js file to modify the initial `let entry` definition, to make sure it doesn't match any CSS classes, so the fallback will need to be used (eg. `let entry = document.querySelector( '.entry-content-not-a-real-class' );`
6. Rerun `npm run build`, and retest the gate in the incognito window. The skip link ID is fairly high so the gate will trigger before scolling, but this is also the case with the Classic theme and `#content` (it contains the post header, whereas `.entry-content` only contains the post body).

### Other information:

### DIFF
--- a/src/content-gate/gate.js
+++ b/src/content-gate/gate.js
@@ -245,7 +245,7 @@ function handleFormSubmission( evt, gate ) {
 function initOverlay( gate ) {
 	let entry = document.querySelector( '.entry-content' );
 	if ( ! entry ) {
-		entry = document.querySelector( '#content' );
+		entry = document.querySelector( '#content, #wp--skip-link--target' );
 	}
 	gate.style.removeProperty( 'display' );
 	let seen = false;
--- a/src/content-gate/gate.js
+++ b/src/content-gate/gate.js
@@ -245,7 +245,7 @@ function handleFormSubmission( evt, gate ) {
 function initOverlay( gate ) {
 	let entry = document.querySelector( '.entry-content' );
 	if ( ! entry ) {
-		entry = document.querySelector( '#content, #wp--skip-link--target' );
+		entry = document.querySelector( '#content' ) || document.querySelector( 'main' );
 	}
 	gate.style.removeProperty( 'display' );
 	let seen = false;


rbcorrales (changes requested):
Hey @laurelfulford! Nice small fix. One thing worth considering:

`#wp--skip-link--target` is added to `<main>` by a WordPress core JS script, not by server-rendered HTML. Since `gate.js` loads with `async`, there's a chance the gate runs before that ID exists in the DOM, making the fallback a no-op.

A simpler alternative that avoids the timing issue could be querying `main` directly, since that's the element WP adds the ID to anyway:

```js
if ( ! entry ) {
    entry = document.querySelector( '#content' ) || document.querySelector( 'main' );
}
```

WDYT?

laurelfulford:
That's a better approach -- thanks @rbcorrales! 🙂 Updated in [84ddbf2](https://github.com/Automattic/newspack-plugin/pull/4431/commits/84ddbf2b5e224e785331a7ac808e77620262850a)!

rbcorrales (approved):
Thanks for addressing! Tested on the block theme with an overlay gate and it works correctly :thumbsup:

PR #4433, opened 2026-01-29
feat: integrations barebones

leogermani:
### All Submissions:

This PR introduces the basic foundations of the integraions.

As simple integration interface, with just one method to push contact data.

An integrations controller, to register, enable, disable integrations.

Modifies the Sync class to loop through integrations instead of just pushing directly to the ESP.

I kept it small to make it easier for review. On the next PR we'll rename some classes and move more things around, but still without changing the resulting behavior of anything.

1. Smoke test the sync process
2. enable RAS and perform actions that will trigger a sync to the ESP
3. Confirm it works just the same as before

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -92,6 +92,8 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-woocommerce.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-esp-sync.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-esp-sync-admin.php';
+		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-integrations.php';
+		\Newspack\Reader_Activation\Integrations::init();
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-utils.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-data-events.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-webhooks.php';
--- a/includes/reader-activation/sync/class-esp-sync.php
+++ b/includes/reader-activation/sync/class-esp-sync.php
@@ -142,7 +142,7 @@ class ESP_Sync extends Sync {
 		foreach ( $integrations as $integration ) {
 			// TODO: We know there's only one integration for now and we expect result to be wp_error or true. We'll refactor this to do a try catch.
 			// Not changing it now because of the retry scheduled by Newspack\WooCommerce_My_Account::sync_email_change_with_esp.
-			$result = $integration->push_contact_data( $contact, $context );
+			$result = $integration->push_contact_data( $contact, $context, $existing_contact );
 		}
 
 		return \is_wp_error( $result ) ? $result : true;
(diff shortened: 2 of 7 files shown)


includes/class-newspack.php
jason10lee, line 1: Had to add `include_once` lines for `includes/reader-activation/integrations/class-integration.php` and `includes/reader-activation/integrations/class-esp.php` to get this to work on my test instance.  Other than that, subscribe/unsubscribe synced just fine with Mailchimp.

jason10lee (changes requested):
Other than includes, LGTM.  This looks like a real solid, simple interface for future integrations.

includes/class-newspack.php
leogermani: These are handled by composer's autoloader, you just need to refresh it when you checkout the branch.

`n build` takes care of it. but you can simply `composer install`, or even the more specific `composer dump-autoload`

leogermani: The Integration include is actually there

* https://github.com/Automattic/newspack-plugin/pull/4433/changes#diff-9553626017fe483a31f15d343b20154aee3274c697997b72bd0e35dad79b7071R37

includes/reader-activation/sync/class-esp-sync.php
```diff
@@ -135,7 +136,14 @@ public static function sync( $contact, $context = '', $existing_contact = null )
...
+
+		$integrations = Integrations::get_active_integrations();
+
+		foreach ( $integrations as $integration ) {
+			// TODO: We know there's only one integration for now and we expect result to be wp_error or true. We'll refactor this to do a try catch.
+			// Not changing it now because of the retry scheduled by Newspack\WooCommerce_My_Account::sync_email_change_with_esp.
+			$result = $integration->push_contact_data( $contact, $context );
+		}
 
 		return \is_wp_error( $result ) ? $result : true;
```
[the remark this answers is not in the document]
leogermani: there is one integration that's harcoded and will always be there in this transition period. This will still be further refactored. No need to worry about it at this point

includes/reader-activation/sync/class-esp-sync.php
```diff
@@ -135,7 +136,14 @@ public static function sync( $contact, $context = '', $existing_contact = null )
...
 		$contact = Sync\Metadata::normalize_contact_data( $contact );
-		$result  = \Newspack_Newsletters_Contacts::upsert( $contact, $master_list_id, $context, $existing_contact );
+
+		$integrations = Integrations::get_active_integrations();
+
+		foreach ( $integrations as $integration ) {
+			// TODO: We know there's only one integration for now and we expect result to be wp_error or true. We'll refactor this to do a try catch.
+			// Not changing it now because of the retry scheduled by Newspack\WooCommerce_My_Account::sync_email_change_with_esp.
+			$result = $integration->push_contact_data( $contact, $context );
+		}
```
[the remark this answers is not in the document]
leogermani: Ok, let's keep it for now and we can revisit this later... this param is confusing but it seems to be used by the change email logic somehow...

reintroduced in https://github.com/Automattic/newspack-plugin/pull/4433/commits/b903409e12055f6871e4c8e1a48259ef719d1cd6

includes/class-newspack.php
jason10lee: Thanks; I'd done an `n npm run build` instead.  All resolved here.

jason10lee (approved):

miguelpeixe (approved):

PR #4434, opened 2026-01-29
fix(collections): hide indicator for unpublished collections

rbcorrales:
### All Submissions:

When a published post is assigned to a draft or scheduled Collection, the frontend no longer displays the collection indicator ("This article appears in..." text or "Browse [CollectionName]" card). Previously, the indicator would render regardless of the Collection's publish status, resulting in broken links and premature disclosure of upcoming Collections.

Closes NPPM-2558.

1. The Collections module should be enabled (Newspack > Settings > Collections).
2. Create a new Collection and leave it as a **draft**.
3. Create and publish a regular post, then assign it to the draft Collection via the Collection Settings panel in the post editor.
4. View the published post on the frontend. Verify **no** collection indicator appears (neither the "This article appears in..." text nor the "Browse" card variant).
5. Now publish the Collection.
6. Reload the post on the frontend. Verify the collection indicator **does** appear with a working link.
7. Create a second Collection (published) and assign the same post to both collections. Verify only the published collection appears in the indicator.
8. PHP tests should pass: `phpunit --filter Test_Content_Inserter`.

### Other information:

### DIFF
--- a/includes/collections/class-content-inserter.php
+++ b/includes/collections/class-content-inserter.php
@@ -64,14 +64,9 @@ class Content_Inserter {
 			return;
 		}
 
-		$collections = Query_Helper::get_post_collections( $post );
-		if ( empty( $collections ) ) {
-			return;
-		}
-
 		// Only show indicators for published collections.
 		$collections = array_filter(
-			$collections,
+			Query_Helper::get_post_collections( $post ),
 			function ( $collection_id ) {
 				return 'publish' === get_post_status( $collection_id );
 			}
--- a/includes/collections/class-content-inserter.php
+++ b/includes/collections/class-content-inserter.php
@@ -69,6 +69,17 @@ class Content_Inserter {
 			return;
 		}
 
+		// Only show indicators for published collections.
+		$collections = array_filter(
+			$collections,
+			function ( $collection_id ) {
+				return 'publish' === get_post_status( $collection_id );
+			}
+		);
+		if ( empty( $collections ) ) {
+			return;
+		}
+
 		// Sort by post date descending.
 		usort(
 			$collections,
--- a/tests/unit-tests/collections/class-test-content-inserter.php
+++ b/tests/unit-tests/collections/class-test-content-inserter.php
@@ -319,12 +319,12 @@ class Test_Content_Inserter extends \WP_UnitTestCase {
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


tests/unit-tests/collections/class-test-content-inserter.php
```diff
@@ -278,6 +278,88 @@ public function test_insert_after_nth_block() {
...
 
+	/**
+	 * Test check_if_post_is_in_collection excludes draft collections.
+	 *
+	 * @covers \Newspack\Collections\Content_Inserter::check_if_post_is_in_collection
+	 */
+	public function test_draft_collection_excludes_indicator() {
+		$filter_name     = 'the_content';
+		$filter_function = [ Content_Inserter::class, 'maybe_insert_collection_indicators' ];
+
```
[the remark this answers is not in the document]
rbcorrales: Added in [b93ac44](https://github.com/Automattic/newspack-plugin/pull/4434/commits/b93ac4456) — the mixed-status test now also covers a scheduled collection.

includes/collections/class-content-inserter.php
```diff
@@ -69,6 +69,17 @@ public static function check_if_post_is_in_collection() {
...
 		}
 
+		// Only show indicators for published collections.
+		$collections = array_filter(
+			$collections,
+			function ( $collection_id ) {
+				return 'publish' === get_post_status( $collection_id );
+			}
+		);
+		if ( empty( $collections ) ) {
```
chickenn00dle, line 74: Rather than repeat the same check twice, we can filter the array before line 68. If `Query_Helper::get_post_collections( $post )` returns an empty array, `array_filter` will do nothing and also return an empty array.

chickenn00dle (changes requested):
Works as expected!

I left one small nitpick then this should be good to go @rbcorrales

includes/collections/class-content-inserter.php
rbcorrales: Good call, simplified in [38b85fb](https://github.com/Automattic/newspack-plugin/pull/4434/commits/38b85fbd2). I was initially thinking about avoiding the extra function call when not needed, but the overhead of calling array_filter on an empty array is negligible.

chickenn00dle (approved):
Thanks for the quick turnaround. 🚢

PR #4442, opened 2026-02-02
feat(content-gate): support group subscriptions access rule

miguelpeixe:
### All Submissions:

NPPD-1120

Depends on #4435.

Implements support for group subscription members to content gating access rules.

I originally considered adding this logic to the `WooCommerce_Connection::get_active_subscriptions_for_user()` method, but other parts of Newspack make use of this method that may not be compatible with group subscriptions.

One of the uses is the ESP sync, for which we need to decide how to treat a group subscription member.

This PR also inaugurates unit tests for content gate access rules, which required a few tweaks to our WC mocks. 

1. Make sure new unit tests are accurate and passing
2. Create a content gate with "Paid Access" using a subscription product access rule
3. Create a new subscription owned by a reader, enable "Group Subscription" and add another reader to it
4. In a new session, confirm that the gate restriction is applied to the targeted article
5. Sign in as the subscription owner and confirm the article is no longer gated
6. Sign in as the subscription member and confirm the article is no longer gated

### Other information:


includes/content-gate/class-access-rules.php
```diff
@@ -276,13 +276,40 @@ public static function get_subscription_products_options() {
...
+				return true;
+			}
+			// Check if the subscription has any of the required products.
+			foreach ( $product_ids as $product_id ) {
+				if ( $subscription->has_product( $product_id ) ) {
+					return true;
+				}
+			}
+		}
+
```
leogermani, line 314: Let's add a filter here. We'll need it to add support for Network. We'll check if they have an equivalent subscription in another site

miguelpeixe: Added in 8ce62777e4d248ff42d4d97c269447c565820526

dkoo: On a similar/related note, there's also a [filter](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-group-subscription-access-rule/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php#L226) for a user's group subscriptions, so Network can hook into that to look for group subscription memberships across the network.

dkoo (approved):
Works and tests well!

PR #4443, opened 2026-02-02
feat[my-account-block]: add option to toggle off icon/text, and update tests

laurelfulford:
### All Submissions:

This PR adds an option to hide either the icon or label on the My Account Button block (but not both at the same time).

As part of that, I changed how the button acts if you don't give it a text label (instead of not showing anything, it will fallback to the default labels) -- when it's hidden in the settings, the label is still used as screen reader text so we don't want it to be empty even if it's not visible.

I also updated the tests to check for the default (vs. making sure empty strings were empty), and tweaked the behaviour to make sure logged in admins (not just readers) would see the My Account labels - I thought this was working before but I was seeing Sign In in trunk. 

See NPPD-1183.

1. Apply this PR and run `npm run build`. As a bonus you can also test this alongside https://github.com/Automattic/newspack-block-theme/pull/406, but it's not required. 
2. If using the default header patterns, the button should already be in the header. If not, add the My Account Button block to the header.
3. Click on the button and confirm you have a Display panel under the Settings tab, with an option to Display label and Display icon. Both should be enabled by default:

CleanShot 2026-02-02 at 10 48 31@2x

4. Try turning of each option and confirm that the button updates appropriately (hiding the text or the icon). Confirm that you can't turn off both options at once -- if you try to turn off say, the icon, the label should switch to visible if it isn't already, and vis versa.
5. Try changing the text labels on each state of the button (Signed in, Signed out), turn off the Show label toggle, and view on the front end.
6. Confirm you only see the icons; when you inspect the icon, you should see your label text wrapped in a `screen-reader-text` span.
7. Try signing in/signing up in an incognito window, and confirm hidden text is right (that you can see the Signed out text in the markup when you're not logged in, and Signed in text when you are).
8. Try editing the header again and switching the settings so you can see the text, but not the icon; repeat step 7. 
9. Try editing the text label of each button and leave it blank; confirm you can still see the placeholder in the editor.
10. Publish your changes and confirm you see the default text on the front-end (Sign In and My Account). 
11. Run the tests just to make sure they come back okay.

### Other information:


src/blocks/my-account-button/class-my-account-button-block.php
```diff
@@ -96,10 +96,19 @@ public static function render_block( $attrs ) {
...
+		$show_icon        = isset( $attrs['showIcon'] ) ? (bool) $attrs['showIcon'] : true;
+
+		if ( ! $show_label && ! $show_icon ) {
+			$show_label = true;
+		}
+
 		if ( '' === trim( (string) $label ) ) {
-			return '';
+			$label = $is_signed_in ? $default_attrs['signedInLabel'] : $default_attrs['signedOutLabel'];
 		}
```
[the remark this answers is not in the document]
laurelfulford: [9d9a68d](https://github.com/Automattic/newspack-plugin/pull/4443/commits/9d9a68dd62d1f57df0c6ff7defa52c9aa0101a1e)

tests/unit-tests/class-test-my-account-button-block.php
```diff
@@ -118,7 +126,8 @@ public function test_render_block_empty_label() {
 			'<!-- wp:newspack/my-account-button {"signedInLabel":"","signedOutLabel":"Sign in"} /-->'
 		);
 
-		$this->assertSame( '', trim( $signed_in_output ) );
+		$this->assertNotEmpty( $signed_in_output );
+		$this->assertStringContainsString( '&quot;signedin&quot;:&quot;My Account&quot;', $signed_in_output );
 	}
```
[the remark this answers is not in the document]
laurelfulford: [84a9ef8](https://github.com/Automattic/newspack-plugin/pull/4443/commits/84a9ef810953876f65c61fe97c4371785b97b321)

thomasguillot:
Sorry to be a pain @laurelfulford. After giving it some more thought, I’m not keen on the toggles or how they behave. Would it be possible to switch to a `ToggleGroupControl` with these 3 options?

* Default (help text: Displays the icon with text.)
* Icon only (ht: Displays the icon without text.)
* Text only (ht: Displays the text without an icon.)

The help text would update dynamically based on the selection.

Also, I’d prefer to keep a single tab rather than splitting things between Settings and Styles.

thomasguillot:
Here's an update, using styles: https://github.com/Automattic/newspack-plugin/pull/4443/changes/d315db9d57cfc77391b29967cd794ca74be37980

image

laurelfulford:
@thomasguillot Looks good to me! 🙂 Now we just need to see what an uninvolved third-party has to say!

src/blocks/my-account-button/edit.js
```diff
@@ -24,16 +24,23 @@ import {
...
-function MyAccountButtonEdit( { attributes, setAttributes } ) {
-	const { signedInLabel, signedOutLabel, style, className: customClassName } = attributes;
+function MyAccountButtonEdit( { attributes, setAttributes, className: wrapperClassName } ) {
+	const { signedInLabel, signedOutLabel, style, className: blockClassName } = attributes;
 	const borderProps = useBorderProps( attributes );
 	const colorProps = useColorProps( attributes );
 	const spacingProps = useSpacingProps( attributes );
+
+	// Block style comes from the Styles panel (wrapper) or saved className.
+	const blockWrapperProps = useBlockProps();
```
rbcorrales: Calling `useBlockProps()` twice in the same component is not a supported pattern. Each call generates its own ref and block binding, so the editor will bind internal state (selection, focus, drag) to whichever ref was registered last. The first call here is never attached to a DOM node, which can cause subtle bugs.

The `is-style-*` class is already available on `attributes.className` (already serialized by WP). You can technically derive the style variant directly from `blockClassName` without this extra call:

```js
const isIconOnly = ( blockClassName || '' ).includes( 'is-style-icon-only' );
const isTextOnly = ( blockClassName || '' ).includes( 'is-style-text-only' );
```

src/blocks/my-account-button/class-my-account-button-block.php
```diff
@@ -153,20 +166,26 @@ public static function render_block( $attrs ) {
...
-		$wrapper_attributes = \get_block_wrapper_attributes(
-			[
-				'class' => implode( ' ', $extra_classes ),
-				'href'  => \esc_url_raw( $href ),
-			]
-		);
+		$wrapper_attribute_args = [
+			'class'             => implode( ' ', $extra_classes ),
+			'href'              => \esc_url_raw( $href ),
+			'data-wp-logged-in' => $is_signed_in ? '1' : '0',
```
rbcorrales: The `data-wp-*` prefix is used by WordPress Interactivity API, and custom plugins should avoid it to prevent future collisions. Can we rename it to something different, like `data-newspack-logged-in`?

Also, this attribute explicitly states the user's authentication state in the DOM, though it's already inferable from the button's href (hashtag vs anything else). Maybe this is not really needed.

src/blocks/my-account-button/class-my-account-button-block.php
```diff
@@ -93,13 +93,26 @@ public static function render_block( $attrs ) {
...
-		if ( '' === trim( (string) $label ) ) {
-			return '';
+		$is_signed_in     = \is_user_logged_in();
+		$signed_in_label  = '' === trim( (string) $attrs['signedInLabel'] ) ? $default_attrs['signedInLabel'] : $attrs['signedInLabel'];
+		$signed_out_label = '' === trim( (string) $attrs['signedOutLabel'] ) ? $default_attrs['signedOutLabel'] : $attrs['signedOutLabel'];
+		$label            = $is_signed_in ? $signed_in_label : $signed_out_label;
+
+		/** Display mode from block style class in className (default = icon + text). */
+		$wrapper_class = (string) $attrs['className'];
+		if ( \strpos( $wrapper_class, 'is-style-icon-only' ) !== false ) {
```
rbcorrales: A hypothetical additional class like `is-style-icon-only-alt` would also match. A more robust approach:

```php
$classes = explode( ' ', (string) $attrs['className'] );
if ( in_array( 'is-style-icon-only', $classes, true ) ) {
```

src/blocks/my-account-button/edit.js
```diff
@@ -51,52 +58,62 @@ function MyAccountButtonEdit( { attributes, setAttributes } ) {
...
-						isPressed={ ! isSignedOutPreview }
-						onClick={ () => setPreviewState( 'signedin' ) }
-						style={ { paddingLeft: '12px', paddingRight: '12px' } }
-					>
-						{ __( 'Signed in', 'newspack-plugin' ) }
-					</ToolbarButton>
-				</ToolbarGroup>
-			</BlockControls>
-			<div className={ classnames( 'wp-block-buttons', customClassName ) }>
+			{ isLabelVisible && (
```
rbcorrales, line 75: When icon-only style is selected, `isLabelVisible` is `false`, so the "Signed out"/"Signed in" toolbar toggle disappears entirely. The user cannot edit the screen reader label text for the signed-in state without switching styles first, and then back.
I'm not sure if there's a solution for this, since it needs to render the text in order to be modified, or if this is even an issue 🤔.

src/blocks/my-account-button/class-my-account-button-block.php
```diff
@@ -93,13 +93,26 @@ public static function render_block( $attrs ) {
...
-		$is_signed_in = \is_user_logged_in();
-		$label        = $is_signed_in ? $attrs['signedInLabel'] : $attrs['signedOutLabel'];
-		if ( '' === trim( (string) $label ) ) {
-			return '';
+		$is_signed_in     = \is_user_logged_in();
+		$signed_in_label  = '' === trim( (string) $attrs['signedInLabel'] ) ? $default_attrs['signedInLabel'] : $attrs['signedInLabel'];
+		$signed_out_label = '' === trim( (string) $attrs['signedOutLabel'] ) ? $default_attrs['signedOutLabel'] : $attrs['signedOutLabel'];
+		$label            = $is_signed_in ? $signed_in_label : $signed_out_label;
+
+		/** Display mode from block style class in className (default = icon + text). */
```
rbcorrales: Nit: WPCS recommends `//` for inline comments. `/** */` style is reserved for docblocks.

src/blocks/my-account-button/class-my-account-button-block.php
rbcorrales: Not on this PR, but `$default_block_class` on line 148 is assigned but never used.

src/blocks/my-account-button/edit.js
```diff
@@ -24,16 +24,23 @@ import {
...
+	const { signedInLabel, signedOutLabel, style, className: blockClassName } = attributes;
 	const borderProps = useBorderProps( attributes );
 	const colorProps = useColorProps( attributes );
 	const spacingProps = useSpacingProps( attributes );
+
+	// Block style comes from the Styles panel (wrapper) or saved className.
+	const blockWrapperProps = useBlockProps();
+	const className = blockWrapperProps?.className ?? wrapperClassName ?? blockClassName ?? '';
+	const isIconOnly = className.includes( 'is-style-icon-only' );
+	const isTextOnly = className.includes( 'is-style-text-only' );
```
rbcorrales: Similar comment about class names matching.

rbcorrales (changes requested):
Nice approach using block styles instead of custom toggles! Left a few inline comments, the main ones being the double `useBlockProps()` call in the editor and the `data-wp-*` namespace on the logged-in attribute. The rest is mostly minor.

src/blocks/my-account-button/edit.js
thomasguillot: Yeah but I think it's OK since it uses the default aria-labels

src/blocks/my-account-button/class-my-account-button-block.php
thomasguillot: https://github.com/Automattic/newspack-plugin/pull/4443/commits/9fed074727322a14d9e8440c9431537bcd116843

src/blocks/my-account-button/class-my-account-button-block.php
thomasguillot: Removed as part of https://github.com/Automattic/newspack-plugin/pull/4443/commits/9fed074727322a14d9e8440c9431537bcd116843

src/blocks/my-account-button/edit.js
thomasguillot: https://github.com/Automattic/newspack-plugin/pull/4443/commits/05812420815307af49d213a4773694e79da3f681

src/blocks/my-account-button/edit.js
thomasguillot: https://github.com/Automattic/newspack-plugin/pull/4443/commits/05812420815307af49d213a4773694e79da3f681

thomasguillot:
@laurelfulford I tried to fix some of the things @rbcorrales mentioned. I haven’t touched the PHP, if you’d like to take care of that.

src/blocks/my-account-button/class-my-account-button-block.php
laurelfulford: Thanks @rbcorrales! I updated the attribute name in [81ee30d](https://github.com/Automattic/newspack-plugin/pull/4443/commits/81ee30dae7ae3a8c3959733a479026e7fdc18ad0), but left it -- it's a little yucky, but the My Account Button block behaves a little different than the My Account link that's inserted into the theme (the latter isn't visible for logged in non-readers, and when it is in the Customizer, it says 'Sign In'). The attribute seemed like the nicest way to switch it in just the one spot.

src/blocks/my-account-button/class-my-account-button-block.php
laurelfulford: Good call! Fixed in [a49f0e0](https://github.com/Automattic/newspack-plugin/pull/4443/commits/a49f0e01e6800c916676d3bde4477d1cdd08993a).

src/blocks/my-account-button/edit.js
laurelfulford: It is a little funky, but I think it's probably okay as is for now -- or at least the kind of thing that should be covered in the documentation (how to edit the hidden label), and if we get a lot of feedback about friction, we can revisit 🙂

laurelfulford: Readme has been updated in [cf68793](https://github.com/Automattic/newspack-plugin/pull/4443/commits/cf6879355928f85588121d6066c2ca2b95d808ff)!

laurelfulford:
Thanks for the review, @rbcorrales, and for the fixes, @thomasguillot! 

I think I got the rest, but let me know if anything still seems funky. I'm going to update the Docs in Linear to also reflect the little "gotcha" around editing the label when hidden.

src/blocks/my-account-button/edit.js
thomasguillot: I'd say keep it as it is for now. If we get feedback about this, we can always add 2 textcontrols to the sidebar

rbcorrales (approved):
Thanks for addressing all the feedback @laurelfulford @thomasguillot! Everything looks good, but one thing that wasn't addressed: the class matching in `edit.js` still uses `String.includes()`, which could match partial class names (e.g. `is-style-icon-only-alt`). The PHP side was correctly updated to `explode` + `in_array`, but the JS side wasn't. The equivalent fix:

```js
const classes = ( blockClassName || '' ).split( ' ' );
const isIconOnly = classes.includes( 'is-style-icon-only' );
const isTextOnly = classes.includes( 'is-style-text-only' );
```

Approving, but would be great to get that last item in before merging.

laurelfulford:
Thanks @rbcorrales! 

> Everything looks good, but one thing that wasn't addressed: the class matching in `edit.js` still uses `String.includes()`...

🤦‍♀️ This was entirely my bad, and is fixed in [b47c2a2](https://github.com/Automattic/newspack-plugin/pull/4443/commits/b47c2a2c9cd71bc76922534ed696e4bbf0c2ea6a) -- thank you for catching that!

PR #4448, opened 2026-02-02
feat(author-profile): avatar and social blocks for nested author profile

rbcorrales:
### All Submissions:

This PR provides the blocks that render inside the Author Profile as nested content and updates the Avatar block:

**Avatar block enhancements:**
- Consumes `newspack-blocks/author` context from the parent Author Profile block to render a single author's avatar in nested mode.
- Shows the site's default gravatar when no custom avatar is set.
- Shows a placeholder (diagonal line) for text-only custom bylines that have no author reference.
- Debounces image fetching when dragging the size slider to avoid excessive network requests.
- Renames "Stacked" style to "Overlapped" for clarity.

**New social blocks:**
- **Author Social Links** (`newspack/author-profile-social`): Container block that renders the author's social media links and email. Reads author data from the parent Author Profile block context.
- **Author Social Link** (`newspack/author-social-link`): Individual social link with SVG icon. Supports icon size customization via parent context.
- Both blocks use a shared `socialButton` SCSS mixin that provides consistent circular icon backgrounds across all block themes, using theme color tokens (`base-2`, `base-3`, `contrast`).
- Includes a PHP fallback SVG icon map for when rendering a block created in the classic theme in the block theme.
- These blocks are in their base version but can be further extended to provide greater flexibility for adding new social networks.

**Other:**
- Enriched `newspack_author_info` REST field with avatar URL, social links, and other metadata to reduce editor API calls. Overrides the `newspack-newsletters` field declaration to avoid having that repo as a dependency.
- Shared `useCoAuthors` hook updated to support single-author mode for nested context.

Closes NPPD-1147.

Related PRs:
- https://github.com/Automattic/newspack-blocks/pull/2286
- https://github.com/Automattic/newspack-block-theme/pull/412

Full end-to-end testing steps are in the [newspack-blocks PR](https://github.com/Automattic/newspack-blocks/pull/2286). The following are additional tests specific to this PR:

**Test standalone avatar (no regression):**
1. Insert an Avatar block outside of an Author Profile block.
2. Verify it still renders the post author's avatar(s) as before.
3. With Co-Authors Plus, verify that multiple author avatars display correctly.
4. Verify the "Overlapped" style (previously "Stacked") still works.
5. Test with a Custom Byline that has no `[Author]` shortcode. The avatar should show a placeholder with a diagonal line.
6. Test the block behavior inside a Query Loop.

**Test social block styling independently:**

1. Verify that the icon size slider in the Author Social Links sidebar works.
2. Verify that the fallback SVG icons render for different services.

### Other information:


src/blocks/author-profile-social/block.json
```diff
@@ -0,0 +1,31 @@
...
+	"title": "Author Social Links",
+	"category": "newspack",
+	"ancestor": [ "newspack-blocks/author-profile" ],
+	"description": "Display the author's social media links and email.",
+	"usesContext": [ "newspack-blocks/author" ],
+	"providesContext": {
+		"newspack-blocks/iconSize": "iconSize"
+	},
+	"textdomain": "newspack",
+	"attributes": {
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [55f3cad](https://github.com/Automattic/newspack-plugin/pull/4448/commits/55f3cad67).

src/blocks/author-social-link/block.json
```diff
@@ -0,0 +1,21 @@
...
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "newspack/author-social-link",
+	"title": "Author Social Link",
+	"category": "newspack",
+	"parent": [ "newspack/author-profile-social" ],
+	"description": "Display a single social media icon for the author.",
+	"usesContext": [ "newspack-blocks/author", "newspack-blocks/iconSize" ],
+	"textdomain": "newspack",
+	"attributes": {
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [55f3cad](https://github.com/Automattic/newspack-plugin/pull/4448/commits/55f3cad67).

src/blocks/author-profile-social/index.js
```diff
@@ -0,0 +1,27 @@
...
+const { name } = metadata;
+
+export { name };
+
+export const settings = {
+	...metadata,
+	title: __( 'Author Social Links', 'newspack' ),
+	icon: share,
+	edit,
+	save: () => <InnerBlocks.Content />,
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [55f3cad](https://github.com/Automattic/newspack-plugin/pull/4448/commits/55f3cad67).

src/blocks/author-social-link/index.js
```diff
@@ -0,0 +1,24 @@
...
+const { name } = metadata;
+
+export { name };
+
+export const settings = {
+	...metadata,
+	title: __( 'Author Social Link', 'newspack' ),
+	icon: share,
+	edit,
+	save: () => null,
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [55f3cad](https://github.com/Automattic/newspack-plugin/pull/4448/commits/55f3cad67).

src/blocks/author-social-link/class-author-social-link-block.php
```diff
@@ -0,0 +1,153 @@
...
+
+		$icon_size = $block->context['newspack-blocks/iconSize'] ?? 24;
+		$svg       = self::get_social_service_svg( $author, $service );
+
+		$output  = '<li class="wp-block-newspack-author-social-link">';
+		$output .= sprintf( '<a href="%s">', esc_url( $url ) );
+
+		if ( $svg ) {
+			$output .= sprintf(
+				'<span style="width: %dpx; height: %dpx;">%s</span>',
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [e51626b](https://github.com/Automattic/newspack-plugin/pull/4448/commits/e51626b4a).

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -0,0 +1,305 @@
...
+			if ( $svg ) {
+				$output .= sprintf(
+					'<span style="width: %dpx; height: %dpx;">%s</span>',
+					absint( $icon_size ),
+					absint( $icon_size ),
+					Newspack_Blocks::sanitize_svg( $svg )
+				);
+			} else {
+				$output .= sprintf( '<span class="service-name">%s</span>', esc_html( $service ) );
+			}
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [e51626b](https://github.com/Automattic/newspack-plugin/pull/4448/commits/e51626b4a).

src/blocks/author-social-link/edit.jsx
```diff
@@ -0,0 +1,152 @@
...
+					</p>
+				</PanelBody>
+			</InspectorControls>
+			<li { ...blockProps }>
+				<a href={ url } onClick={ e => e.preventDefault() }>
+					{ svg ? (
+						<span dangerouslySetInnerHTML={ { __html: svg } } style={ { width: iconSize, height: iconSize } } />
+					) : (
+						<span className="service-name">{ service }</span>
+					) }
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [e51626b](https://github.com/Automattic/newspack-plugin/pull/4448/commits/e51626b4a).

src/blocks/author-social-link/edit.jsx
```diff
@@ -0,0 +1,152 @@
...
+					<p style={ { wordBreak: 'break-all', fontSize: '12px', color: '#757575' } }>
+						<ExternalLink href={ url }>{ url }</ExternalLink>
+					</p>
+				</PanelBody>
+			</InspectorControls>
+			<li { ...blockProps }>
+				<a href={ url } onClick={ e => e.preventDefault() }>
+					{ svg ? (
+						<span dangerouslySetInnerHTML={ { __html: svg } } style={ { width: iconSize, height: iconSize } } />
+					) : (
```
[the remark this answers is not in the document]
rbcorrales: Added `aria-hidden` to the SVG span in [e51626b](https://github.com/Automattic/newspack-plugin/pull/4448/commits/e51626b4a), and sanitized REST API SVGs using `safeHTML` from `@wordpress/dom` in [9eef684](https://github.com/Automattic/newspack-plugin/pull/4448/commits/9eef6849c).

src/blocks/author-profile-social/edit.js
```diff
@@ -0,0 +1,180 @@
...
+			'--icon-size': `${ iconSize }px`,
+		},
+	} );
+
+	// Pass iconSize to author context for child blocks to read.
+	// We use a mutable property on the author object as a simple way to pass it down.
+	if ( author ) {
+		author._iconSize = iconSize;
+	}
+
```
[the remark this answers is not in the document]
rbcorrales: Fixed in [2ff58a3](https://github.com/Automattic/newspack-plugin/pull/4448/commits/2ff58a37d).

src/blocks/author-social-link/edit.js
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @param {Object} props            Block props.
+ * @param {Object} props.attributes Block attributes.
+ * @return {JSX.Element|null} The edit component.
+ */
+export default function AuthorSocialLinkEdit( { attributes } ) {
+	const AuthorContext = getSharedAuthorContext();
+	const author = useContext( AuthorContext );
+	const { service } = attributes;
+	const iconSize = author?._iconSize || 24;
```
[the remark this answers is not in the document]
rbcorrales: Children now read `iconSize` from the block context prop instead. Fixed in [2ff58a3](https://github.com/Automattic/newspack-plugin/pull/4448/commits/2ff58a37d).

src/blocks/author-profile-social/edit.jsx
```diff
@@ -0,0 +1,180 @@
...
+		replaceInnerBlocks( clientId, blocks, false );
+	};
+
+	const addMissingLinks = () => {
+		const { createBlock } = wp.blocks;
+		const newBlocks = missingServices.map( service => createBlock( 'newspack/author-social-link', { service } ) );
+		const editor = wp.data.select( 'core/block-editor' );
+		const existingBlocks = editor.getBlocks( clientId );
+		replaceInnerBlocks( clientId, [ ...existingBlocks, ...newBlocks ], false );
+	};
```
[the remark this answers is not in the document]
rbcorrales: Imported `createBlock` from `@wordpress/blocks` and exposed `innerBlocks` from `useSelect` to avoid needing standalone `select`. Fixed in [a9470ac](https://github.com/Automattic/newspack-plugin/pull/4448/commits/a9470ace9).

src/blocks/avatar/class-avatar-block.php
```diff
@@ -135,6 +143,81 @@ class="<?php echo esc_attr( $class ); ?>"
...
+			$avatar_url = get_avatar_url( $author['id'], [ 'size' => $image_size * 2 ] );
+		} else {
+			$avatar_url = '';
+		}
+
+		if ( empty( $avatar_url ) ) {
+			return '';
+		}
+
+		$author_name = esc_attr( $author['name'] ?? '' );
```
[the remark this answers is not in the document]
rbcorrales: Removed `esc_attr()` from assignment, keeping the escaping only at output. Fixed the nested render path in [2a2d86f](https://github.com/Automattic/newspack-plugin/pull/4448/commits/2a2d86fa0) and the standalone render path in [4b68b8a](https://github.com/Automattic/newspack-plugin/pull/4448/commits/4b68b8a96).

src/shared/hooks/use-coauthors.js
```diff
@@ -12,11 +12,74 @@ import apiFetch from '@wordpress/api-fetch';
...
+	}
+	inflightRequests[ nicename ] = apiFetch( {
+		path: `${ COAUTHORS_ENDPOINT }/${ encodeURIComponent( nicename ) }`,
+	} )
+		.then( result => {
+			if ( result?.avatar_urls ) {
+				guestAvatarCache[ nicename ] = result.avatar_urls;
+			}
+		} )
+		.catch( () => {} ) // Silently skip failed fetches.
```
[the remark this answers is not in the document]
rbcorrales: Now caches `false` as a sentinel for missing avatars and failed fetches, using `in` operator for the cache check. Fixed in [b72ea99](https://github.com/Automattic/newspack-plugin/pull/4448/commits/b72ea99fc).

src/blocks/avatar/style.scss
```diff
@@ -3,8 +3,8 @@
 	display: flex;
 	gap: 0.5rem;
 
-	&.is-style-stacked,
-	&[class*="is-style-stacked"] {
+	&.is-style-overlapped,
+	&[class*="is-style-overlapped"] {
```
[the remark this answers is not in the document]
rbcorrales: Not applicable. This is a new block with no existing saved content using the old style name.

src/blocks/avatar/class-avatar-block.php
```diff
@@ -36,8 +36,8 @@ public static function register_block_styles() {
 			register_block_style(
 				'newspack/avatar',
 				[
-					'name'  => 'stacked',
-					'label' => __( 'Stacked', 'newspack-plugin' ),
+					'name'  => 'overlapped',
+					'label' => __( 'Overlapped', 'newspack-plugin' ),
 				]
 			);
```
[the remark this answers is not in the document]
rbcorrales: Not applicable. This is a new block with no existing saved content using the old style name.

src/shared/scss/_mixins.scss
```diff
@@ -10,3 +10,48 @@
 	position: absolute !important;
 	word-wrap: normal !important; /* Many screen reader and browser combinations announce broken words as they would appear visually. */
 }
+
+// Circular social button style
+// Used by: author-profile-social, author-social-link blocks
+@mixin socialButton {
+	background: var(--wp--preset--color--base-2);
```
laurelfulford: Edge case: consider adding fallbacks for these hex values, in case the theme's different or a publisher has deleted the default colours 😅

includes/plugins/co-authors-plus/class-author-rest-fields.php
```diff
@@ -0,0 +1,121 @@
...
+ */
+class Author_Rest_Fields {
+
+	/**
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		// Priority 11 overrides the simpler registration of the same field
+		// by newspack-newsletters (priority 10), adding enriched avatar and guest data.
+		add_action( 'rest_api_init', [ __CLASS__, 'register_fields' ], 11 );
```
laurelfulford, line 26: Minor nit, but should we only be adding this file/these fields for block themes since the byline/avatar blocks are only available there?

(And something I should've added to the Newspack Blocks review: was this new version going to be opt in/opt out for the Block Theme, so it was only needed for more extreme layout needs?)

src/blocks/avatar/edit.jsx
```diff
@@ -120,14 +142,78 @@ const AvatarWrapper = ( { avatar, size, attributes, placeholder = false } ) => {
...
+	// Hooks must be called unconditionally per React rules.
+	const defaultAvatarUrl = useDefaultAvatar();
 	const { postId, postType } = context;
 	const avatar = useUserAvatar( { userId: attributes?.userId, postId, postType } );
 	const allAuthors = usePostAuthors( { postId, postType } );
 	const { bylineActive, bylineContent } = useCustomByline( postId, postType );
-	const blockProps = useBlockProps();
 
-	// Text-only custom byline (no [Author] shortcodes) — show placeholder.
+	// Memoize author ID extraction to avoid running regex on every render.
```
laurelfulford, line 47: Small typo:
```suggestion
	// Memorize author ID extraction to avoid running regex on every render.
```

src/blocks/avatar/class-avatar-block.php
```diff
@@ -135,6 +143,81 @@ class="<?php echo esc_attr( $class ); ?>"
...
+					<img
+						src="<?php echo esc_url( $avatar_url ); ?>"
+						class="<?php echo esc_attr( $class ); ?>"
+						alt="<?php echo esc_attr( $author_name ); ?>"
+						width="<?php echo esc_attr( $image_size ); ?>"
+						height="<?php echo esc_attr( $image_size ); ?>"
+						style="<?php echo esc_attr( $border_attributes['style'] ?? '' ); ?>"
+					/>
+				<?php endif; ?>
+			</div>
```
laurelfulford: This is more personal preference than anything, but what do you think about wrapping just the link markup in the link_to_author check so the image doesn't need to be duplicated?

```suggestion
			<div class="newspack-avatar-wrapper <?php echo esc_attr( $duotone_class ); ?>">
				<?php if ( $link_to_author && ! empty( $author_url ) ) : ?>
					<a href="<?php echo esc_url( $author_url ); ?>" class="wp-block-newspack-avatar__link">
				<?php endif; ?>
						<img
							src="<?php echo esc_url( $avatar_url ); ?>"
							class="<?php echo esc_attr( $class ); ?>"
							alt="<?php echo esc_attr( $author_name ); ?>"
							width="<?php echo esc_attr( $image_size ); ?>"
							height="<?php echo esc_attr( $image_size ); ?>"
							style="<?php echo esc_attr( $border_attributes['style'] ?? '' ); ?>"
						/>
				<?php if ( $link_to_author && ! empty( $author_url ) ) : ?>
					</a>
				<?php endif; ?>
			</div>
```

src/blocks/author-social-link/social-icons.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * WordPress dependencies
+ */
+import { safeHTML } from '@wordpress/dom';
+
+/**
+ * Fallback SVG icon map for social services.
+ * Used when Newspack_SVG_Icons (from newspack-theme) is not available (e.g., block theme).
```
laurelfulford: This overlaps with some feedback I left in the Newspack Blocks PR, but what do you think about making one central source for these SVGs in the Newspack Plugin, and using them for both the old and new version of the Author Profile block? That way, these wouldn't be "fallbacks", they'd just be the one set used for Author Profiles. 

Also: is it possible to have one set of SVG code that's used by both the PHP and JS, just to simplify this code? It's unlikely to change unless we add more fields ourselves, but it could make maintenance easier down the road!

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -0,0 +1,306 @@
...
...ath d="M8.9 16.1L9 6.6c0-.1.1-.1.1-.1.1 0 .1 0 .1.1l.1 9.5c0 .1 0 .1-.1.1s-.2 0-.3-.1zM.4 13.5c0 .1-.1.1-.2.1s-.2-.1-.2-.1L0 11.9v1.5c0 .1.1.2.2.2s.2-.1.2-.2v-1.5c0 .2 0 1.2 0 1.6zm1-1.1c-.1 0-.1 0-.2.1L1 11.1v2.4l.2 1.3c0 .1.1.1.2.1s.1 0 .2-.1l.2-1.3v-2.5c0-.1-.1-.1-.2-.1zm.8-1c-.1 0-.2.1-.2.2L1.9 9.5l.1 4v.1c.1.1.1.1.2.1s.2-.1.2-.2l.1-4.1-.2-2.2v.2c-.1.1-.1 0-.1 0zm.9-.3c-.2 0-.2.1-.2.2l-.2 2.5.2 3.8c0 .1.1.2.2.2.2 0 .2-.1.2-.2l.2-3.8-.2-2.5c0-.1-.1-.2-.2-.2zm1 .2c-.2 0-.2.1-.3.2l-.1 2.3.1 3.8c0 .2.1.2.3.2.1 0 .2-.1.2-.2l.2-3.8-.2-2.3c0-.1-.1-.2-.2-.2zm.8-1.8c-.2 0-.3.1-.3.3L4.5 12l.1 3.8c0 .2.1.3.3.3.1 0 .2-.1.3-.3l.2-3.8-.2-2.5c0-.2-.1-.3-.3-.3zM6 8.8c-.2 0-.3.1-.3.3l-.2 2.9.2 3.7c0 .2.1.3.3.3.2 0 .3-.1.3-.3l.2-3.7-.2-2.9c0-.2-.2-.3-.3-.3zm.8-.4c-.2 0-.3.2-.3.3l-.2 3.3.2 3.7c0 .2.2.3.3.3.2 0 .3-.1.4-.3l.1-3.7-.1-3.3c0-.2-.2-.3-.4-.3zm.8-.1c-.2 0-.4.2-.4.4L7.1 12l.1 3.6c0 .2.2.4.4.4s.4-.2.4-.4l.1-3.6-.1-3.3c0-.2-.2-.4-.4-.4zm.9-.4c-.3 0-.4.2-.4.4l-.1 3.7.1 3.5c0 .2.2.4.4.4.3 0 .4-.2.4-.4l.2-3.5-.2-3.7c-.1-.2-.2-.4-.4-.4zm3.4-1.1c-.2 0-.3 0-.5.1-.2-2.7-2.4-4.8-5.2-4.8-.7 0-1.4.1-2 .4-.3.1-.4.2-.4.5v12.3c0 .3.2.5.4.5h7.7c1.9 0 3.4-1.5 3.4-3.4s-1.5-3.5-3.4-3.5v-.1z"></path></svg>',
```
laurelfulford: Both the SoundCloud and Wikipedia icons are a little funky when loaded from the Newspack Plugin:

Image

src/blocks/author-profile-social/edit.jsx
```diff
@@ -0,0 +1,171 @@
...
+					<RangeControl
+						label={ __( 'Icon Size', 'newspack-plugin' ) }
+						value={ iconSize }
+						onChange={ value => setAttributes( { iconSize: value } ) }
+						min={ 16 }
+						max={ 48 }
+					/>
+					{ missingServices.length > 0 && (
+						<Button variant="secondary" onClick={ addMissingLinks }>
+							{ __( 'Add missing links', 'newspack-plugin' ) }
```
laurelfulford, line 128: This button works great to fix it, but the mismatch between the front-end and editor after making profile edits is a little strange 😅 I've noticed some things -- like adding/removing a phone number -- are picked up right away in the editor, but not the social links, even after exiting the editor. The front end stays up to date, though.

Is there any way the links can be fetched again when the editor loads to keep things synced up?

src/blocks/author-social-link/edit.js
```diff
@@ -0,0 +1,153 @@
...
+			instagram: 'Instagram',
+			linkedin: 'LinkedIn',
+			youtube: 'YouTube',
+			bluesky: 'Bluesky',
+			pinterest: 'Pinterest',
+			myspace: 'Myspace',
+			soundcloud: 'SoundCloud',
+			tumblr: 'Tumblr',
+			wikipedia: 'Wikipedia',
+			website: __( 'Website', 'newspack-plugin' ),
```
laurelfulford: The website isn't being output, but it doesn't look like we use it anywhere normally (in trunk for the Profile block, or in the Classic theme -- at least not yet!), so it can probably be removed for now.

laurelfulford (changes requested):
Hi @rbcorrales! These blocks seem to be working well overall! I found some odds and ends and added them to (I think) the right PRs as feedback.

Given the size of this PR in particular, I think it might be good to get a second set of eyes on it -- I'm not confident in my assessment of some of the changes, and I'm sure I've missed a couple silly things! 😅 

----

Most of the social icon information is strictly opt-in: a user doesn't have to add values and can control whether it shows on the front end by not populating it. The one exception is the email address, which all regular WordPress users must have.

What do you think about adding some way to hide the email from the block? The current theme has a site-wide setting to hide it from where it's generated by the theme template, and the original block had a per-block toggle. 

----

Similar-ish to the above, when you're editing the Author Social Links block, it lets you remove the inner blocks (like delete one of the Social Media icons), but that doesn't persist when you publish the changes from a post/page or the Site Editor.

It might be best to block removing these blocks? I think that'll be the least confusing way to handle it (at least for now!). 

----

This isn't new in this PR but just capturing it here - I can move it to Linear if it's a pain to fix: 

When you use the "Overlapped" style and pair it with a different border radius than the default, the "border" that separates the two images is still round: 

Image

----

Also not required for this PR, but down the road what do you think about adding some style options for the Author Social Links block -- specifically the ability to change the icon colours/remove the background?

src/blocks/avatar/edit.jsx
rbcorrales: I think this might be a misunderstanding\! "Memoize" is actually a CS term for caching the result of a function call so it doesn't need to be recomputed: https://en.wikipedia.org/wiki/Memoization.

rbcorrales:
@laurelfulford good catches! I'm still going through everything, but addressing these top-level comments here (as I can't reply to them individually). These two are related:

>  What do you think about adding some way to hide the email from the block? The current theme has a site-wide setting to hide it from where it's generated by the theme template, and the original block had a per-block toggle.

> Similar-ish to the above, when you're editing the Author Social Links block, it lets you remove the inner blocks (like delete one of the Social Media icons), but that doesn't persist when you publish the changes from a post/page or the Site Editor.

The issue is that the PHP render callback for `render_social_with_inner_blocks` was re-adding any author services not present in the saved inner blocks as a "fallback." So if you deleted a social link in the editor and saved, PHP would see it as "missing" and render it again on the frontend. Same thing for email, as `get_author_available_services()` unconditionally included it, so there was no way to hide it in InnerBlocks mode.

This was introduced [here](https://github.com/Automattic/newspack-blocks/pull/2286/changes/d3cd8f8d008e562c6b4593c75144536a872f82ea#diff-2fadb2b9c7672752a5fcd5fcf4488d00af6dc325cc7ba69330b51578c43e14dfR124-R152) when the individual author-social-link inner block was added (before migrating them from the blocks repo to here). The original flat render didn't have this problem.

I removed the fallback loop and the now-orphaned `get_author_available_services()` method. The render callback now only renders inner blocks that are actually saved in post content. The editor already has "Reset links" and "Add missing links" buttons for when new services become available, so nothing is lost.

Code removed here: [455ea09](https://github.com/Automattic/newspack-plugin/pull/4448/commits/455ea095eb0be03ae5125b44c0590fde46e9a6da).

With this fix, there's no need for a separate toggle to hide email (or any other service) because the inner blocks themselves are the source of truth. You can just remove the blocks you don't want from that particular block instance within a post or a template.

> When you use the "Overlapped" style and pair it with a different border radius than the default, the "border" that separates the two images is still round

Yeah, I noticed this too. It's pre-existing from when the block was initially created. The overlapped style uses a `radial-gradient` mask for the separator, which is always circular regardless of the `border-radius` value. I can try fix it here since I'm already touching the avatar styles. Should be a matter of switching from `radial-gradient` to a mask that respects the block's border-radius.

> Also not required for this PR, but down the road what do you think about adding some style options for the Author Social Links block -- specifically the ability to change the icon colours/remove the background?

Yup, there are multiple things to improve in this block, like easily adding additional networks, for example. If you have a clear idea of what should be included, can you create a subtask for this in the related Linear issue, including the details, so we don't lose track?

src/shared/scss/_mixins.scss
rbcorrales: Added hex fallbacks from the block theme's theme.json defaults: 6fc7714.

includes/plugins/co-authors-plus/class-author-rest-fields.php
rbcorrales: This registration intentionally runs at priority 11 to override the simpler version in `newspack-newsletters` (priority 10), which only returns `id`, `display_name`, and `author_link`. This version adds `user_nicename`, `is_guest`, and `avatar_urls` on top of those, so it's a superset. I'd rather keep it available for all themes, so the plan is to eventually remove the one from newsletters and keep this as the single source.

Also, since the field is `context: ['edit']` only, it doesn't add anything to frontend API responses. And gating it on theme type could break (though kind of unlikely) if someone switches themes while someone else is editing.

> (And something I should've added to the Newspack Blocks review: was this new version going to be opt in/opt out for the Block Theme, so it was only needed for more extreme layout needs?)

The nested version (`layoutVersion: 2`) would now be the default for block themes, not opt-in. The idea is that if you're on a block theme, the Author Profile should work like any other block in the Site Editor (composable with InnerBlocks, editable via block bindings, and customizable through block patterns). Using the flat version on a block theme would mean losing all of that for no real benefit.

The flat version (`layoutVersion: 1`) stays as the default for classic themes, where the Site Editor isn't available and InnerBlocks don't make sense.

Blocks created with one version don't get automatically upgraded to the other. A `layoutVersion: 1` block created on a classic theme will keep working as-is if the site switches to a block theme (and vice versa). So there's no risk (🤞) of breaking existing content on theme switches.

I had a feature flag for this early on (`NEWSPACK_AUTHOR_PROFILE_NESTED_BLOCKS`), but removed it because `wp_is_block_theme()` is already the natural gate. I feel like adding a separate toggle on top of that would be redundant, unless you really want to keep the ability to add the old version of the block in the new theme.

**Important note:** All of the above is simply my rationale for some of these decisions, and none of it is set in stone. If you feel strongly about one of them, please let me know 😄.

src/blocks/author-profile-social/edit.jsx
rbcorrales: See my other [longer comment](https://github.com/Automattic/newspack-plugin/pull/4448#issuecomment-3916799218). I think the frontend/editor inconsistency was addressed here [455ea09](https://github.com/Automattic/newspack-plugin/pull/4448/commits/455ea095eb0be03ae5125b44c0590fde46e9a6da), but let me know if I'm still missing anything.

Edit: Sorry, I re-read everything, and it seems that you're talking about actually editing the user data? Are you doing it on a separate tab? If so, I wouldn't expect the editor to automatically keep track of all data changes in real-time, but I would expect a reload to render the updated info (unless I'm still misunderstanding the issue).

In any case, could you please check whether the above-mentioned changes resolve this? If not, could you share the steps to replicate the issue?

src/blocks/avatar/edit.jsx
laurelfulford: Ha! TIL 😄 Wikipedia knew I was going to make that mistake:

CleanShot 2026-02-17 at 14 11 17

src/blocks/avatar/class-avatar-block.php
rbcorrales: Good point. Extracted both render paths into a shared `render_avatar_image()` helper that outputs the wrapper `div`, conditional link, and `` tag. Both `render_block()` (standalone/multi-author) and `render_single_author_avatar()` (nested/single-author) now call it instead of duplicating the markup. The helper outputs directly since both callers are already inside `ob_start()`.

Updated here: [3d1fa9a](https://github.com/Automattic/newspack-plugin/pull/4448/commits/3d1fa9a0513b62a04e186d8f710b0f0080ebc850).

src/blocks/author-social-link/edit.js
rbcorrales: Removed it from the JS icon/label maps and the PHP SVG map. Can always add it back if there's a use case. [2373555](https://github.com/Automattic/newspack-plugin/pull/4448/commits/23735557a2d15102ceda55d01f76ff92327f3350)

src/blocks/author-profile-social/class-author-profile-social-block.php
rbcorrales: Fixed in [bf523ce](https://github.com/Automattic/newspack-plugin/pull/4448/commits/bf523ce46b18210f9259e1c0511a7f11b3b7ab83). Updated SoundCloud, Wikipedia, and MySpace SVGs (in both PHP and JS maps) to match the theme versions. MySpace had some rendering artifacts with the original paths.

Kept the plugin versions for Twitter (theme uses the bird, plugin uses the current X logo) and Tumblr (plugin version matches the current logo).

src/blocks/author-social-link/social-icons.js
rbcorrales: Done. The plugin is now the canonical source for social SVGs. The REST controller in `newspack-blocks` checks the plugin first and falls back to the classic theme's `Newspack_SVG_Icons` only if the plugin class is not available.

Also renamed `get_fallback_social_svg` to `get_social_svg` since it is no longer a fallback.

- Plugin: [9339795](https://github.com/Automattic/newspack-plugin/pull/4448/commits/9339795595444d5c5a120fb32a43b35f8fac755a)
- Blocks: [5ed084d](https://github.com/Automattic/newspack-blocks/pull/2286/commits/5ed084d2b8ccac6473d94c6fba352b5b8e2c1947)

As for sharing a single set between PHP and JS: the JS map is used in the editor as a client-side fallback when the REST response does not include SVGs (e.g. author data not yet loaded). Since PHP runs server-side and JS runs client-side, they need separate copies of the data. We could generate the JS map from PHP at build time, but that adds complexity for marginal benefit.

rbcorrales: Okay, I reconsidered this and made a couple of changes to simplify/dedupe things:

1. Created a `Social_Icons` utility class in newspack-plugin (`includes/class-social-icons.php`) with the SVG map as a PHP class constant. Both the block render callbacks and the `newspack-blocks` REST controller use `Social_Icons::get_svg()`. The `Newspack_SVG_Icons` (theme) fallback in the REST controller was removed, so the plugin is the only source for both old and new Author Profile blocks.

2. The JS file no longer carries its own copy of the SVGs. It reads them from the REST API author data, which is populated by PHP. So there's a single set of SVG markup defined once in PHP, shared everywhere.

One tradeoff worth noting: the SVG strings are still repeated per author in the REST JSON response when multiple authors share the same social networks. This was already the case before these changes and could be refactored later for a minimal optimization on the editor side if needed.

- Plugin: [`4e3a5ff`](https://github.com/Automattic/newspack-plugin/pull/4448/commits/4e3a5ffc2)
- Blocks: [`07d1a22`](https://github.com/Automattic/newspack-blocks/pull/2286/commits/07d1a220)

Let me know if this approach makes sense!

includes/plugins/co-authors-plus/class-author-rest-fields.php
laurelfulford: > I'd rather keep it available for all themes, so the plan is to eventually remove the one from newsletters and keep this as the single source.

Makes sense -- I'm good with that!

> I feel like adding a separate toggle on top of that would be redundant, unless you really want to keep the ability to add the old version of the block in the new theme.

I picked up the idea that this would be opt in/opt out from the [comments in Linear](https://linear.app/a8c/issue/NPPD-1147/author-bio-block#comment-c3f24014), but I'm double-guessing it. 

IIRC the idea behind the flag was that the block's UI wouldn't be any more complicated than a site needed it to be, so if a publisher was okay with the default options, we wouldn't expose anything beyond that. I can't find that in writing, though, which makes me wonder if that was my personal rational 😆 

This might be a good one to loop @thomasguillot in on just to make sure I'm just misremembering, or get thoughts at our next call. I can also add a comment to [this test task](https://linear.app/a8c/issue/NPPD-1240/author-profile-block-tam-testing) to get TAM feedback while they're all looking at it. Let me know if you'd like me to do that!

I think it's fine to leave as is until we get feedback beyond me 🙂

src/blocks/avatar/class-avatar-block.php
laurelfulford: Nice, looks good!

src/blocks/author-profile-social/edit.jsx
laurelfulford: ~~**Edited to add:** please see comment below - good chance this is just me and working fine!~~

**Edited edited to add:** I'm still seeing issues with the block in the Site Editor but not in the content, so please ignore that bit!

I was talking about editing user data, yes! When I tested this originally, these were the steps:

1. Set up an Author Profile block that shows a specific user and publishing it. 
2. Editing that user's information to add social links.
3. View on the front-end -- the Author Profile block displays the changes.
4. View in the editor, try refreshing, opening and closing it -- the Author Profile block does not display the changes until the "Add new links" button is clicked.

Now when I try to retest it, there's consistency across the editor and front-end, but neither seem to be picking up the changes when you add social accounts:

1. Repeat the above steps (adding a Author Profile block with a specific user, publishing the page).
2. Also add the Author Profile in contextual mode to the Single Post template.
3. Edit the user's social links and add some links.
4. View on the front-end; the social links are not updated.
5. View in the editor; the social links aren't updated on their own, but updating them by clicking the 'Add New' button and publishing the page; that also carries to the front-end.
6. View a story by the user and view the profile in the footer - note that the icons are not updated there either.

When you remove social links in all of these cases, they update automatically, it's just when you add links that it doesn't. 

Just let me know if that doesn't make sense, or doesn't line up with what you're seeing! I can try to set it up on a JN site in that second case to rule out any environmental weirdness on my part!

rbcorrales:
@laurelfulford thanks for the very detailed review!

I fixed the overlapped style issues in [7b82e97](https://github.com/Automattic/newspack-plugin/pull/4448/commits/7b82e973838089e7280cf22bc42bba42313d4148). The SVG mask now uses a `<rect rx>` that adapts to the `border-radius`, so the separator cutout matches the avatar shape. The CSS fallback preserves the circular default. The editor preview also updates live via `getOverlapMask()`.

I think that was the last one pending from the feedback on this particular PR. Let me know if you have any questions or comments.

src/blocks/author-social-link/edit.js
laurelfulford: Thanks @rbcorrales!

src/shared/scss/_mixins.scss
laurelfulford: Awesome, thanks!

src/blocks/author-profile-social/edit.jsx
laurelfulford: Well, now when I just rested this to test something else, it worked right away ಠ_ಠ
 
I'm going to do more testing but please ignore this for now, it's very likely it's something local to me!

includes/plugins/co-authors-plus/class-author-rest-fields.php
rbcorrales: > so if a publisher was okay with the default options, we wouldn't expose anything beyond that. I can't find that in writing, though, which makes me wonder if that was my personal rational 😆

Yeah, I think we talked about this as a possibility (and I incorporated the flag in my first attempt).

Technically, you could still create instances of the flat version of the block in the block theme by just removing `"layoutVersion":2` from the markup in the code or by not adding that to patterns, templates, etc.
What I'm trying to avoid is having that many switches and alternate behaviors based on multiple conditions that we have to maintain over time, themes, and across sites.

> so if a publisher was okay with the default options, we wouldn't expose anything beyond that

Also, there's still a missing piece to this that is in the works, which is the pattern selector (I had to stash my progress on that to address the feedback from the current PRs, but with all the changes that would be kind of a mess to merge 😆). That could also simplify the flow and provide additional layouts for publishers, which could, in turn, reduce the likelihood that they want to further reorder/customize things.

src/blocks/author-profile-social/edit.jsx
laurelfulford: Ok, in repeated testing the block seems to work when adding and removing social links in the content (both in the editor and on the front-end).

In the Site Editor, it seems to work for some icons but not all, if that makes sense? Like adding/removing Facebook updates in real time, but some of the other services (like Instagram) refuse to show. This is even after removing the block in the editor, and re-adding it:

CleanShot 2026-02-18 at 14 23 27

Let me know if it'd help if I set this up on a JN site and retest if you're not seeing any of this!

src/blocks/author-profile-social/class-author-profile-social-block.php
laurelfulford: Thanks @rbcorrales! Those two icons look good in the post/page editor now.

> Kept the plugin versions for Twitter (theme uses the bird, plugin uses the current X logo) and Tumblr (plugin version matches the current logo).

This makes sense! The Classic theme uses x.com for X and kept the bird for Twitter (from when the branding was transitioning); both that and the Tumblr icon should probably be updated there at some point 🙂

src/blocks/author-social-link/social-icons.js
laurelfulford: Thanks @rbcorrales! This looks good overall and work well on the front-end and in the post editor! 

The only issue I'm seeing is in the Site Editor placeholders, I think just because the SVGs aren't included there. I thought they were getting  distorted somehow, but the black squiggles are services names all smooshed:

CleanShot 2026-02-18 at 14 29 07

src/blocks/author-profile-social/edit.jsx
rbcorrales: 🤦 I broke it when I moved the icons to the shared mapping and removed the JS mapping.
The nested mode (v2) placeholder was missing SVG data for social icons, causing them to render as text labels in the Site Editor. Fixed by adding a REST endpoint (`GET /newspack/v1/social-icons`) to `Social_Icons` that returns the SVG map, and fetching it from the placeholder logic only when in template context.

The result is cached at the module level, so only one request is made regardless of how many block instances are on the page.

- newspack-plugin: [9858470](https://github.com/Automattic/newspack-plugin/pull/4448/commits/98584700f62f42586a18925da5fbba8dd4d048d6)
- newspack-blocks: [a33ee4c](https://github.com/Automattic/newspack-blocks/pull/2286/commits/a33ee4cff3176eafc4c7aa2f1520a0bfcff6d58f)

I still need to look into the add/removal issue you mentioned.

Edit: BTW, this is _one_ way to solve the icons issue in the Site Editor. The other option would be to add it to the page as a localized JS variable, or duplicate the icons map 🙂. Let me know if any of those sound more appealing to you.

src/blocks/author-social-link/social-icons.js
rbcorrales: Sorry, I fixed it here, but I think I replied to the wrong comment 😅 
https://github.com/Automattic/newspack-plugin/pull/4448#discussion_r2824892484

Edit: I just realized that I indeed replied to the correct comment, but you ninja-edited it 😆

laurelfulford: Sorry that was my bad! I commented in a different spot originally, then realized it was something else entirely (I thought the squiggles were messed up SVGs at first!)

laurelfulford: Also this looks good! I still get a flash of the letters but it resolves. I think that may be my machine -- if it's noticeable to anyone else we could give it a negative text-indent in the CSS but if you're not seeing that that's probably overkill!

includes/plugins/co-authors-plus/class-author-rest-fields.php
laurelfulford: That's fair! I'm good with shelving this until we hear from others, it doesn't need to be a blocker here 🙂 I'll add it as a comment to the TAM feedback issue just to make sure they realize it's an open question (with the complexity of having to maintain both as a possible factor).

thomasguillot:
Current:
Screenshot 2026-02-19 at 17 52 47@2x

The border radius on the mask is too small. The radius, should be selected radius + the whatever the gap is.

Expected:
Screenshot 2026-02-19 at 17 56 40@2x

includes/plugins/co-authors-plus/class-author-rest-fields.php
laurelfulford: Question for TAMs asked [here](https://linear.app/a8c/issue/NPPD-1240/author-profile-block-tam-testing#comment-f75441a2); I'm going to mark this as 'resolved' for this PR, and I can spin up a separate issue if it turns out the flag is the preferred approach.

laurelfulford:
Thanks @rbcorrales! I'll defer to @thomasguillot's feedback on the overlap, he caught something I did not 🙂 

The only other bit outstanding for me was the end-of-story Author Profile (like when it's added to the Site Editor) still seems [limited to what social icons it shows](https://github.com/Automattic/newspack-plugin/pull/4448#discussion_r2824892484) -- it seems to just be the ones used in the placeholders. Otherwise I think that's it for this one, thanks for all your hard work and apologies for all the nitpicks! 😅

rbcorrales:
Hey @thomasguillot, I wanted to clarify what you meant here. Looking at the screenshots you shared, the expected result actually shows a smaller border radius on the mask (tighter corners on the cutout), not a larger one. Did you mean that the mask radius should account for the gap by subtracting it rather than adding?

That's the approach I went with in [bb3b3f0](https://github.com/Automattic/newspack-plugin/pull/4448/commits/bb3b3f04322580d3fa6f685b95f7c00e6093aedd): `mask_rx = avatar_rx - gap`.
The core update for the calculation is this line (and its PHP equivalent), the rest is mostly refactoring:
```js
rx = Math.round( Math.max( 0, Math.min( SVG_SIZE / 2, rx - OVERLAP_GAP ) ) * 100 ) / 100;
```

The gap is the visible separator between the overlapping avatars (7.5% of the avatar size on the SVG scale), which comes from the difference between the cutout width (25%) and the overlap amount (17.5%).

The idea is that the gap itself contributes to the perceived roundness of the separator corner, so the mask curve needs to be tighter to compensate. The total visual transition (mask curve + gap) then matches the avatar's border radius.

A couple of caveats worth noting with this approach:

1. **Small radii:** when the border radius is smaller than the gap offset, the mask radius clamps to 0 (sharp corners). For example on a 128px avatar, any border radius below ~10px produces a square cornered mask. This seems acceptable since those avatars are nearly square anyway.
2. **Large radii approaching circular:** the avatar's visual roundness maxes out at 50% border radius (perfect circle), but the mask hasn't caught up yet at that point. The mask continues getting rounder until the border radius reaches roughly 50% + gap. For example, on a 100px avatar the avatar is already [a perfect circle](https://www.youtube.com/watch?v=xTgKRCXybSM) at 50px, but the mask doesn't become fully circular until ~58px. In practice this means avatars with border radius close to 50% will have a slightly less round mask. For exact 50% we skip the SVG mask entirely and use a radial gradient.
3. **Sync requirement:** this change introduces a shared overlap fraction value (0.175) that needs to stay in sync across three files: the SCSS (where it sets the negative margin), the JS (editor side mask generation), and the PHP (frontend mask generation). Previously, this was only defined in CSS.

Let me know if the current result matches what you had in mind or if it needs further tweaking. But unless you feel strongly about this approach, I'd lean towards keeping the original radius without the gap offset. The visual difference is subtle, and it avoids both the maintenance cost of keeping three files in sync and the mismatch near the clamped extremes where the mask and avatar roundness fall out of step.

thomasguillot:
@rbcorrales it still looks a bit odd but I think it's ok. In my opinion, if you're not using round avatars, you're better off using the default style and not the overlapped one anyway.

src/blocks/author-profile-social/edit.jsx
rbcorrales: @laurelfulford I added the rest of the social icons to the placeholder in this commit to the block repo:
[d0f7095](https://github.com/Automattic/newspack-blocks/pull/2286/commits/d0f7095445cbe34a6be931176ad58235989e09f8)

Let me know if that fixes things for you!

rbcorrales: @laurelfulford okay, one (hopefully) last change to the icons. In the post editor, when a nested author profile block was added in contextual mode, the social block auto-populated inner blocks based only on the first preview author's configured services. This means if the first author has 5 networks, all other authors' previews were limited to those same 5, even if they have more networks configured.

The social block now fetches the full list of supported services from the `/newspack/v1/social-icons` endpoint we created before and uses it for auto-populating inner blocks. This ensures inner blocks exist for every possible service regardless of which author is previewed first. Individual author-social-link blocks already return null for services the author doesn't have, so extra inner blocks are invisible.

[d2a93eb](https://github.com/Automattic/newspack-plugin/pull/4448/commits/d2a93eb2582118cc90bfa67fc03a023624d3d9d1)

thomasguillot:
padding/margin were not working on the front-end this should be fixed: https://github.com/Automattic/newspack-plugin/pull/4448/commits/fa917a2e726cb86eef0ee82d00e3b6aaae7e0efc

Also, I moved the service label as block title https://github.com/Automattic/newspack-plugin/pull/4448/commits/ceb63f25758eec92e5fbaf1ba3dff609eb0e4218

rbcorrales:
@thomasguillot I committed 620319a1b, which fixes a duplicate `wp-block-newspack-author-profile-social` class on the frontend wrapper introduced in fa917a2e7. `get_block_wrapper_attributes()` already generates the block class from the registration when `WP_Block_Supports::$block_to_render` is set, so passing it manually in the class argument doubled it.

rbcorrales:
@laurelfulford I updated the overlap mask math to use your approach from the [minimize-math](https://github.com/Automattic/newspack-plugin/commit/2eb8dde59b3fdd6198420e43466079046aff3dc7) branch as it looked and worked better!

I didn't merge it directly because there were conflicts with earlier changes on the same files, so I replicated the logic, did some CSS cleanup, named the geometry values as local variables (so they have a context and look less magical 😄), and updated comments. Let me know if this looks good.

[0f439e7](https://github.com/Automattic/newspack-plugin/pull/4448/commits/0f439e746953d02123f78d752990b408f38b9803)

thomasguillot:
I noticed an issue with the duotone filter: When it's as part of the Author Profile block, the filter isn't applied on the front-end. (Note: As a standalone block, the Avatar block correctly displays the filter)

Editor:
Screenshot 2026-02-23 at 16 31 22@2x

Front-end:
Screenshot 2026-02-23 at 16 31 29@2x

Also, I think we should allow for Spacer and Separator blocks.

thomasguillot:
@rbcorrales In https://github.com/Automattic/newspack-plugin/pull/4448/commits/301184c148b7db22016f5c29ae539569800c5ef5 I replaced the icon size's RangeControl with a SelectControl.

This way it's the same behaviour as Core's Social Icons block.

To differentiate Normal from Default, I'm rounding the numbers to 2 so 23→24, 24→24.

src/blocks/author-profile-social/edit.jsx
laurelfulford: @rbcorrales This looks good! 
* I can get any/all social icons showing up when the block's added to the Site Editor, regardless of author order (the author with the most links is second, so order isn't influencing it)
* I can do the same in the post/page editor and on the front end when added there regardless of who's first/previewed first
* Deleting/resetting (to bring back deleted icons) and reordering all seem to work well

Thanks!

laurelfulford:
> I updated the overlap mask math to use your approach from the [minimize-math](https://github.com/Automattic/newspack-plugin/commit/2eb8dde59b3fdd6198420e43466079046aff3dc7) branch... I didn't merge it directly because there were conflicts with earlier changes on the same files, so I replicated the logic, did some CSS cleanup, named the geometry values as local variables

Pretty sure this means it's changed enough that I can review it - looks good! 🙂 

The social links also look good! That was it for stuff I found.

> I noticed an issue with the duotone filter: When it's as part of the Author Profile block, the filter isn't applied on the front-end. (Note: As a standalone block, the Avatar block correctly displays the filter)...

I can recreate this issue @thomasguillot found but it doesn't seem to be exactly the same: it looks like the issue's happening for him in the post/page editor, but it works fine there for me with the Author Profile block, I only see the problem when I add it via the Site Editor 😕 

I do think we're at the point where it'd be nice to get this bigger PR merged and knock out any remaining issues in smaller PRs. Let me know what you think - if that sounds okay, I can mark this as approved and move the duotone feedback to Linear.

rbcorrales:
Created these two linear issues:
- NPPD-1269 - Duotone filter bug
- NPPD-1270 - Spacer/Separator support

laurelfulford (approved):
Thanks @rbcorrales! 🚢

PR #4450, opened 2026-02-03
fix: skip redirect from login modal

jason10lee:
### All Submissions:

Login flow for existing users from Reader Registration block happens in a modal.  And end of successful login, reader has the option to either hit the 'Continue' button or the 'X' to exit the flow. 'Continue' redirects to a signup page; 'X' closes the modal and leaves the user on the page they started the flow.

This PR detects whether the form is being shown in a modal; if so, it skips the redirect step so that 'Continue' will close the modal just like the 'X' button.

Note that the block may appear on a standalone page (like `/my-account/` when not logged in).  In those cases, the existing behavior--the redirect--is left unchanged so that the reader exits at a useful location.

Addresses [NPPM-2542: Bug: Reader Registration block redirects off the current page after using the Magic Link OTP](https://linear.app/a8c/issue/NPPM-2542/bug-reader-registration-block-redirects-off-the-current-page-after).

#### Original modal flow as documented in the Linear issue
1. Pull `fix/modal-skip-login-redirect` topic branch.
2. Add the Reader Registration block to any page, such as the Home Page
4. Create a reader account on the site (if one doesn't exist).
5. Either log out or open a fresh browser session.
6. Navigate to the Reader Registration block and select "Sign into an existing account".
7. Enter the email address of the existing reader account and select "Continue".
8. Enter the OTP and hit continue.
9. On the success page, select 'Continue'. Verify it closes the modal and leaves you where you started.

#### Standalone page test
1. As 1-3 above; make sure you're logged out before proceeding.
2. Navigate to a page that will show a standalone login form, like `/my-account/`.
3. Complete the login flow.
4. On success, select 'Continue'.  Verify it redirects you off the login page.

### Other information:

### DIFF
--- a/src/reader-activation-auth/index.js
+++ b/src/reader-activation-auth/index.js
@@ -47,6 +47,14 @@ window.newspackRAS.push( readerActivation => {
 			ev.preventDefault();
 			const modalTrigger = ev.target;
 			let callback, redirect;
+
+			// If reader sees this form as a modal, it's likely they'll want
+			// to dismiss the modal and stay on the underlying page when
+			// sign-in flow is complete.
+			// Detect a modal container in ancestors and flag so we can skip
+			// the redirect later.
+			const isInsideOverlay = modalTrigger.closest( '.newspack-popup, .newspack-lightbox, [class*="lightbox"], [class*="popup"]' );
+
 			if ( ev.target.getAttribute( 'data-redirect' ) ) {
 				redirect = ev.target.getAttribute( 'data-redirect' );
 			} else {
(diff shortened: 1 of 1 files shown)


jason10lee:
The PR as is should meet the question as asked.

However, I noticed that either 'X'-ing out or hitting 'Continue' without redirecting leaves links to 'My Account' (in the success state of the Reader Registration block or other existing login links) in a bit of a broken state.  It redirects to `/my-account/` which then leaves the user at a sign-in prompt instead of moving them along to the `/my-account/edit-account/` page they'd usually see when logged in.

I suspect other things that check for login will also be broken until the page is refreshed.

On that note, maybe an alternative might be to redirect to the current page to refresh the login state rather than dropping the redirect altogether?  This'd probably need to be added to the 'X', too.
Screenshot 2026-02-03 at 09 08 42
Screenshot 2026-02-03 at 09 09 20

jason10lee:
Back to draft while I review the new state of things as of #4452.

jason10lee:
Tested #4452, and it looks like it handles this problem!  Even if not, the new flow obsoletes this changeset.  Closing this PR out.

PR #4462, opened 2026-02-06
feat(newspack-components): add divider component

thomasguillot:
### All Submissions:

This PR adds a new component: `Divider`. I added a readme to explain how to use this in the future.

1. Switch to this branch
2. Check if Dashboard still works
3. Check Components Demo.

### Other information:


dkoo (changes requested):
The `100vw` width doesn't work when it's used inside a container-type element:

Image

We could add an exception for when it's inside a `newspack-card` element with borders, but I'm not sure that would catch all instances like this. IMO `full-width` shouldn't be the default alignment.

Also, React recommends using function-style components over class-style components, so I refactored this to a function-style component in 37f38972c.

thomasguillot:
@dkoo PR updated with better defaults https://github.com/Automattic/newspack-plugin/pull/4462/commits/98932a8fcef9a63264b8361a00ddf16577ec0f70

dkoo (approved):

PR #4474, opened 2026-02-11
feat(content-gating): new UI for adding/editing content gates

dkoo:
### All Submissions:

Implements part of the new i3 designs for content gating admin UI. Does not implement i3 designs for the main content gates "list" view—this PR only covers new designs for the empty state, and the new Add/Edit view.

In this PR:

- New `__experimentalCoreCard` and `__experimentalCoreProps` for the Newspack `Card` component to use the [core Card component](https://wordpress.github.io/gutenberg/?path=/docs/components-card--docs) (see components demo for examples)
- New `CardSettingsGroup` component which uses the experimental core Card component as an alternative to `ActionCard` components
- New `Badge` component which renders a standalone Newspack-styled badge element
- New icons for `content`, `currency`, `settings`, and `paywall` in the Icons package
- New props for the `SectionHeader` component for badges, icons, and font style/alignment
- New props for the `Wizard` component to allow for a "fixed" header style, as well as getting/setting header breadcrumb text and action buttons/menus via Redux store
- `Wizard` component now passes shared props, section-specific props, and Router props to each section it renders as a child component
- New `edit` view for creating new/editing existing content gates
- "Delete" button now skips trash and permanently deletes content gates
- `create` REST API route now allows for setting all available gate config upon post creation (previously it only allowed setting the title upon creation)
- Updated "Content Gates" admin menu name and related wizard API routes to "Access Control"
- Does NOT implement the new "Snackbar"-style notices in i3 designs—these will come in a later PR along with updated designs for the Content Gates list view.

Note that the Content Gates list view will technically still render usable UI if one or more content gates exist, but this view may look broken as it's currently in an in-between state in terms of design.

Related to NPPD-1146.

1. Check out this branch with `define( 'NEWSPACK_CONTENT_GATES', true );` set in wp-config.php.
2. Delete all existing content gates: `wp post delete $(wp post list --post_type=np_content_gate --format=ids) --force`
3. Visit **Audience > Access Control** (note new menu name and URL structure)
4. Note the new "empty" state for the `ContentGates` view component. We'll call this the "Onboarding" view:

Screenshot 2026-02-10 at 5 41 24 PM

5. Click on the button for "Restrict all posts". Confirm that you're redirected to an "Add new content gate" view with the "Restrict all posts" option pre-selected. Also confirm that the header breadcrumb changes to "Audience Management / Access Control / Add new" and a disabled "Save" button appears on the right side of the header. We'll call this the "Add new" view:

Screenshot 2026-02-10 at 5 42 23 PM

6. Click on the left-caret next to the "Add new content gate" title and confirm that it navigates you back to the Onboarding view. Confirm that the header breadcrumb reverts to "Audience Management / Access Control" and the "Save" button disappears.
7. From the Onboarding view, click on "Choose specific content" and confirm that it navigates you to the "Add new" view with the "Choose specific content" panel preselected and open:

Screenshot 2026-02-10 at 5 43 38 PM

8. Change the "Choose specific content" setting, switch back and forth between "Choose specific content" and "Restrict all posts", and confirm that the settings controls work as expected.
9. Enter a gate title, choose some content rules, enable Registration and/or Paid Access, and tweak all settings, confirming that all controls work as expected.
10. Once you have a title, content rules, and either Registration or Paid Access enabled (or both), confirm that you can click the "Save" button in the fixed header and that it persists all configured settings. Confirm that after the loading animation finishes, the view changes slightly: the header breadcrumb should now say "Audience Management / Access Control / Edit", there should be a dropdown menu icon next to the "Save" button, and the title `TextControl` field is no longer shown. We'll call this the "Edit" view:

Screenshot 2026-02-10 at 5 50 37 PM

11. Refresh the page and confirm that your changes persist after the browser reload.
12. Open the dropdown menu in the header and confirm you see the following options:

Screenshot 2026-02-10 at 5 51 33 PM

13. Test the "Rename" item: it should show the Title `TextControl` field again and allow you to edit the gate title. (You must click "Save" to persist title changes.)
14. Test the "Activate" item: it should save the gate with an updated `publish` status (shown as an "Active" badge next to the title).
15. The "Activate" menu item is now replaced by a "Deactivate" item. Confirm that it reverts the gate to `draft`/"Inactive" status.
16. Test the "Delete" item: after a confirmation dialog, it should delete the post (skipping trash) and navigate you back to the empty Onboarding view.
17. Test what happens when you pass an invalid gate ID in the URL: point your browser to `domain.com/wp-admin/admin.php?page=newspack-audience-access-control#/edit/1234` where `1234` is a non-existent post ID. Confirm that you get redirected to an "Add new" page with an error message.

### Other information:


miguelpeixe:
### Code review

Found 3 issues:

1. **Wrong layout URL in Custom Access component** — In `custom-access.tsx`, the "Edit Layout" button calls `getEditGateLayoutUrl( gateId, 'registration' )` instead of passing `'custom_access'` as the gate mode. This means clicking "Edit Layout" in the Custom Access section opens the Registration layout editor instead of the Custom Access layout editor. The Registration component at `registration.tsx:39` correctly uses `'registration'`, but Custom Access should use its own mode.

https://github.com/Automattic/newspack-plugin/blob/00ba6d34e73a1f274a015dc399c794f52fb9bba6/src/wizards/audience/views/content-gates/edit/custom-access.tsx#L51-L53

2. **`create_gate()` signature change breaks existing callers** — The method signature was changed from `create_gate( $title = '' )` to `create_gate( $gate )`, where `$gate` is now expected to be an array. However, the WC Memberships integration still calls it with a string title at `class-memberships.php:149` and `class-memberships.php:285`, and 15 unit tests in `content-gates.php` also pass string titles. These callers will cause fatal errors (array access on a string).

https://github.com/Automattic/newspack-plugin/blob/00ba6d34e73a1f274a015dc399c794f52fb9bba6/includes/content-gate/class-content-gate.php#L497-L499

3. **Save button allows publishing gates with Paid Access enabled but no access rules** — The disabled condition only checks `customAccess.active` but does not verify that `customAccess.access_rules` contains at least one rule. A user can enable Paid Access, configure no rules, and publish a gate in a broken state where paid access is enforced but no rules define who has access.

https://github.com/Automattic/newspack-plugin/blob/00ba6d34e73a1f274a015dc399c794f52fb9bba6/src/wizards/audience/views/content-gates/edit/index.tsx#L165-L167

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/components/src/badge/index.tsx
```diff
@@ -0,0 +1,24 @@
+/**
```
thomasguillot, line 1: Can we use WordPress's [Badge](https://wordpress.github.io/gutenberg/?path=/docs/components-badge--docs) instead (with some custom css/class to hide the badge icon if needed)

includes/collections/class-collection-category-taxonomy.php
```diff
@@ -218,8 +218,9 @@ public static function save_term_meta( $term_id ) {
 	 */
 	public static function add_support_for_content_gates( $taxonomies ) {
 		$taxonomies[] = [
-			'slug'  => self::get_taxonomy(),
-			'label' => __( 'Collection Categories', 'newspack-plugin' ),
+			'slug'        => self::get_taxonomy(),
+			'label'       => __( 'Collection Categories', 'newspack-plugin' ),
```
thomasguillot: Could we use sentence case for all the labels, please?

packages/components/src/card/core-card.js
```diff
@@ -0,0 +1,87 @@
...
+						<div className="newspack-card--core__icon">
+							<Icon icon={ icon } height={ isSmall ? 24 : 48 } width={ isSmall ? 24 : 48 } />
+						</div>
+					) }
+					{ header && <div className="newspack-card--core__header-content">{ header }</div> }
+					{ actionType === 'chevron' && <Icon className="newspack-card--core__action" icon={ chevronRight } height={ 24 } width={ 24 } /> }
+					{ actionType === 'toggle' && <ToggleControl className="newspack-card--core__action" checked={ isActive } /> }
+				</CardHeader>
+			) }
+			{ children && <div className="newspack-card--core__body">{ children }</div> }
```
thomasguillot, line 89: How come we're not using `CardBody` here?

packages/icons/src/currency.js
```diff
@@ -0,0 +1,17 @@
+/**
```
thomasguillot: This is `@wordpress/icons`'s `currencyDollar`

packages/icons/src/content.js
```diff
@@ -0,0 +1,12 @@
+/**
```
thomasguillot: This is `@wordpress/icons`'s `postList`

packages/icons/src/settings.js
```diff
@@ -0,0 +1,16 @@
+/**
```
thomasguillot: This is `@wordpress/icons`'s `settings`

src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
```diff
@@ -1,28 +1,24 @@
...
 		return null;
 	}
-	if ( rule.is_boolean ) {
-		return <CheckboxControl label={ rule.name } checked={ true } onChange={ noop } disabled help={ rule.description } />;
-	}
 	if ( rule.options && rule.options.length > 0 ) {
 		return (
 			<FormTokenField
-				label={ rule.name }
+				label={ '' }
```
thomasguillot, line 21: How come we're not using `hideLabelFromVision`?

src/wizards/audience/views/content-gates/edit/content-rule-control-taxonomy.tsx
```diff
@@ -130,7 +130,7 @@ export default function ContentRuleControlTaxonomy( { slug, value, onChange }: G
 
 	return (
 		<FormTokenField
-			label={ rule.name }
+			label={ '' }
```
thomasguillot, line 133: Same https://github.com/Automattic/newspack-plugin/pull/4474/changes#r2791829030

thomasguillot (changes requested):
Thanks @dkoo. I’ll test this thoroughly today. In the meantime, I’ve left a few comments and suggestions.

src/wizards/index.tsx
```diff
@@ -43,7 +43,7 @@ const components: Record< string, any > = {
 		label: __( 'Audience Campaigns', 'newspack-plugin' ),
 		component: lazy( () => import( /* webpackChunkName: "audience-wizards" */ './audience/views/campaigns' ) ),
 	},
-	'newspack-audience-content-gates': {
+	'newspack-audience-access-control': {
 		label: __( 'Audience Content Gates', 'newspack-plugin' ),
 		component: lazy( () => import( /* webpackChunkName: "audience-wizards" */ './audience/views/content-gates' ) ),
```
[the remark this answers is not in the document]
dkoo: Fixed in a338ced

src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+		}
+		setGate( matchedGate );
+		setTitle( matchedGate.title );
+		setContentRules( matchedGate.content_rules );
+		setRegistration( matchedGate.registration );
+		setCustomAccess( matchedGate.custom_access );
+		setStatus( matchedGate.status );
+		setContentType( getContentTypeFromRules( matchedGate.content_rules ) );
+		setHeaderSection( isNew ? __( 'Add new', 'newspack-plugin' ) : __( 'Edit', 'newspack-plugin' ) );
+	}, [ gates, id, isNew ] );
```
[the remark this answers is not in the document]
dkoo: a338ced5c

src/wizards/audience/views/content-gates/utils.tsx
```diff
@@ -21,3 +26,11 @@ export function getEditGateLayoutUrl( gateId: number, gateMode: string ) {
...
 	return url;
 }
+
+export const getGateStatus = ( status: GateStatus ) => {
+	return status === 'publish' ? __( 'Active', 'newspack-plugin' ) : __( 'Inactive', 'newspack-plugin' );
+};
+
+export const getGateStatusBadgeLevel = ( status: GateStatus ) => {
+	return status === 'publish' ? 'success' : 'info';
+};
```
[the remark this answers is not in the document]
dkoo: This is intentional—the design calls for simplifying statuses to simply "active" vs. "inactive".

includes/content-gate/class-content-gate.php
```diff
@@ -490,19 +490,19 @@ public static function is_post_restricted( $post_id = null ) {
...
 		$all_gates = self::get_gates();
 		$gate_id   = \wp_insert_post(
 			[
-				'post_title'   => $title,
+				'post_title'   => $gate['title'],
 				'post_type'    => $post_type,
-				'post_status'  => 'draft',
-				'post_content' => self::get_default_gate_content(),
+				'post_status'  => $gate['status'] ?? 'draft',
+				'post_content' => '',
```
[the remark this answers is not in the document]
dkoo: dfc389a84 fixes all instances of `create_gate` that were still passing the title string only.

src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+	useEffect( () => {
+		setContentRules( contentType === 'all' ? DEFAULT_GATE.content_rules : contentRules );
+	}, [ contentType ] );
+
+	// Update gate status.
+	useEffect( () => {
+		if ( status !== gate.status ) {
+			handleSave();
+		}
+	}, [ status ] );
```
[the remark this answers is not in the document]
dkoo: Fixed in e2d1fc134

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -246,15 +246,85 @@ public function register_api_endpoints() {
...
+									'type'       => 'object',
+									'properties' => [
+										'slug'      => [ 'type' => 'string' ],
+										'value'     => [ 'type' => 'mixed' ],
+										'exclusion' => [ 'type' => 'boolean' ],
+									],
+								],
+							],
+							'registration'  => [
+								'type'       => 'object',
```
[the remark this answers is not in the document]
dkoo: Fixed in 700b90b99

src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+				type: 'more',
+				label: __( 'Delete', 'newspack-plugin' ),
+				action: () => handleDelete( gate.id ),
+				disabled: isFetching,
+				destructive: true,
+			} );
+		}
+		return _actions;
+	}, [ contentRules, customAccess, isFetching, isRenaming, registration, title ] );
+
```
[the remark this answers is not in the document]
dkoo: Fixed in f7787d0ae

includes/content-gate/class-access-rules.php
```diff
@@ -86,20 +86,20 @@ public static function get_registered_rules() {
 	public static function register_default_rules() {
 		$rules = [
 			'subscription' => [
-				'name'        => 'Has Active Subscription',
-				'description' => 'The user must be logged into a reader account and have an active subscription with one of the selected products.',
+				'name'        => 'Active Subscription',
+				'description' => 'Requires an active subscription to selected products.',
 				'options'     => [ __CLASS__, 'get_subscription_products_options' ],
 				'callback'    => [ __CLASS__, 'has_active_subscription' ],
 			],
```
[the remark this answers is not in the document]
dkoo: Fixed in 143c3592e

src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+				data: { gate: _gate },
+			},
+			{
+				onSuccess( data: Gate ) {
+					updateGatesData( gates.map( g => ( g.id === data.id ? data : g ) ) );
+					setIsRenaming( false );
+				},
+			}
+		);
+	}, [ gate, contentRules, registration, customAccess, status, title ] );
```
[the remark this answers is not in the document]
dkoo: Tweaked the state handling in b440dbaf1. `gates` can now be `null` before the wizard data is fetched/returned via REST API so we can wait until this resolves before comparing the component state with gates' data. This lets us detect if the post ID in the edit URL is invalid and reset the state, if needed. 

All `handle*` functions are now wrapped in `useCallback` so that we can be sure they represent the current gate data when called from inside `useEffect` hooks.

dkoo:
@miguelpeixe fixes for the following issues:

>     1. **Wrong layout URL in Custom Access component** — In `custom-access.tsx`, the "Edit Layout" button calls `getEditGateLayoutUrl( gateId, 'registration' )` instead of passing `'custom_access'` as the gate mode. This means clicking "Edit Layout" in the Custom Access section opens the Registration layout editor instead of the Custom Access layout editor. The Registration component at `registration.tsx:39` correctly uses `'registration'`, but Custom Access should use its own mode.
> 
> 
> https://github.com/Automattic/newspack-plugin/blob/00ba6d34e73a1f274a015dc399c794f52fb9bba6/src/wizards/audience/views/content-gates/edit/custom-access.tsx#L51-L53

Good catch! Fixed in 144166dd2.

>     2. **`create_gate()` signature change breaks existing callers** — The method signature was changed from `create_gate( $title = '' )` to `create_gate( $gate )`, where `$gate` is now expected to be an array. However, the WC Memberships integration still calls it with a string title at `class-memberships.php:149` and `class-memberships.php:285`, and 15 unit tests in `content-gates.php` also pass string titles. These callers will cause fatal errors (array access on a string).

This was also [caught by Copilot](https://github.com/Automattic/newspack-plugin/pull/4474#discussion_r2791008952). dfc389a fixes all lingering instances of `create_gate` that were still passing the old title string as an argument.

> https://github.com/Automattic/newspack-plugin/blob/00ba6d34e73a1f274a015dc399c794f52fb9bba6/includes/content-gate/class-content-gate.php#L497-L499
> 
>     3. **Save button allows publishing gates with Paid Access enabled but no access rules** — The disabled condition only checks `customAccess.active` but does not verify that `customAccess.access_rules` contains at least one rule. A user can enable Paid Access, configure no rules, and publish a gate in a broken state where paid access is enforced but no rules define who has access.
> 
> 
> https://github.com/Automattic/newspack-plugin/blob/00ba6d34e73a1f274a015dc399c794f52fb9bba6/src/wizards/audience/views/content-gates/edit/index.tsx#L165-L167

I'm not so sure we need to be so prescriptive, but e0be779ec does disable the "Save" button if only Paid Access is enabled with no rules enabled or no rules with valid values.

packages/icons/src/currency.js
dkoo: Didn't realize! 7f0d3ebc0 removes the icon from packages in favor of the one from `@wordpress/icons`.

packages/icons/src/content.js
dkoo: Didn't realize! 7f0d3ebc0 removes the icon from packages in favor of the one from `@wordpress/icons`.

packages/icons/src/settings.js
dkoo: Didn't realize! 7f0d3ebc0 removes the icon from packages in favor of the one from `@wordpress/icons`.

src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
dkoo: `FormTokenField` doesn't support this prop. 😞 [reference](https://wordpress.github.io/gutenberg/?path=/docs/components-formtokenfield--docs)

src/wizards/audience/views/content-gates/edit/content-rule-control-taxonomy.tsx
dkoo: `FormTokenField` doesn't support this prop. 😞 [reference](https://wordpress.github.io/gutenberg/?path=/docs/components-formtokenfield--docs)

packages/components/src/card/core-card.js
dkoo: This is so that we can pass multiple stacked `CardBody`, `CardDivider`, `CardMedia` etc. in a single `children` prop. I toyed with trying to detect when `children` didn't have those components as the top-level elements and then wrap them in a `CardBody` only then, but it felt like creating a whole new unnecessary API when we can just do something like:

```
<Card __experimentalCoreCard>
	<CardMedia>Full bleed child</CardMedia>
	<CardBody>First child</CardBody>
	<CardDivider />
	<CardBody>Second child</CardBody>
</Card>
```

The `newspack-card--core__body` wrapper is so that if you do pass children without those core `Card*` child components, we can at least replicate the padding behavior of a `CardBody`.

includes/collections/class-collection-category-taxonomy.php
dkoo: Fixed in fb3188680

packages/components/src/badge/index.tsx
dkoo: According to this component's [documentation](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/badge), it's locked inside the package's private APIs and not to be used outside of code shipped as part of WP core.

dkoo:
~PHP linting and testing are still failing possibly due to a cached testing environment in GitHub actions. They're passing locally after a862507.~ Looks like they're passing now 👍

src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
thomasguillot: 🤯
Do you know if we can use `aria-label`?

packages/components/src/card/core-card.js
thomasguillot: Got it 👍

packages/components/src/badge/index.tsx
thomasguillot: Oh... ok. What a shame 😭

thomasguillot:
@dkoo I have made some changes to the empty state's grid and margins/gaps (https://github.com/Automattic/newspack-plugin/pull/4474/commits/0e44425fba540262bcd82d8117a602e7560e024a)

I also decided to change the Access control icon. I wasn't too keen on the wall + dollar iconography. I switched to a door + lock. https://github.com/Automattic/newspack-plugin/pull/4474/commits/c65c483f2c69b4ba0fd111d26d79d8b317602a10

update

thomasguillot:
@dkoo Design-wise, we're good to go

thomasguillot (approved):

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -246,15 +246,85 @@ public function register_api_endpoints() {
 				'methods'             => 'POST',
 				'callback'            => [ $this, 'create_gate' ],
 				'args'                => [
-					'title' => [
-						'type'     => 'string',
-						'required' => true,
-						'messages' => [
-							'required' => __( 'Title is required.', 'newspack-plugin' ),
+					'gate' => [
```
miguelpeixe, line 329: Can we move these args to a variable so `create_gate` and `update_gate` share the same?

miguelpeixe (changes requested):
This looks great! Left a minor request inline.

I'm concerned about the approach for the new header design. My understanding is that this will become the new header across wizards. In that case, the SectionHeader component doesn't feel like the right component for this.

What do you think about leveraging [SlotFills](https://developer.wordpress.org/block-editor/reference-guides/slotfills/) and/or React hooks in the existing wizard root header?

includes/wizards/audience/class-audience-content-gates.php
dkoo: Great idea, moved to a static variable in 3bff28bb6!

dkoo:
> I'm concerned about the approach for the new header design. My understanding is that this will become the new header across wizards. In that case, the SectionHeader component doesn't feel like the right component for this.
> 
> What do you think about leveraging [SlotFills](https://developer.wordpress.org/block-editor/reference-guides/slotfills/) and/or React hooks in the existing wizard root header?

@miguelpeixe I'm not sure about using SlotFills for this since it would require each section component to define the components used to render inside the slots.

bf685c711 refactors things so that the wizard Redux store renders header data that can be set by dispatch from inside the section components, which lets us pass actions and header name/title data and render them in a controlled way. WDYT?

Also open to changing this approach in a separate PR if it needs deeper rethinking.

miguelpeixe (approved):
Thanks, @dkoo!

I agree SlotFills are not the solution here. I like the approach using Redux dispatch. I'm thinking about how we can wrap the dispatch hook and data lifecycle in a simpler react hook, but this can be a follow-up PR.

PR #4484, opened 2026-02-12
fix: add check for my account before switching error notice

laurelfulford:
### All Submissions:

Right now there's an issue in trunk where the 'Billing Information' screen is being skipped in the modal checkout.

This appears to be something to do with swapping the error.php -- at least so sez AI: 

> It’s because notices/error.php doesn’t output any HTML. It calls Newspack_UI::add_notice() (snackbar) and returns nothing. On normal My Account pages that’s fine, but during modal checkout the first‑step validation relies on Woo’s notice HTML to come back in the AJAX response (update_order_review). When the error template outputs nothing, the response has no messages, so the JS thinks validation succeeded and skips to the next screen.

This PR adds a check to only swap out the error.php when using My Account -- there might be a better way to do this so I'm open to suggestions!

1. Set up your site to have a Checkout button block, and disable requiring signing in when making a purchase.
2. In an incognito window, run through the checkout; note you get bumped immediately to the second step without being asked for billing info:

CleanShot 2026-02-12 at 10 37 20@2x

3. Apply this PR.
4. Repeat step 2 and make sure you get asked for your Billing details:

CleanShot 2026-02-12 at 10 36 45@2x

### Other information:

### DIFF
--- a/includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
+++ b/includes/plugins/woocommerce/my-account/class-my-account-ui-v1.php
@@ -184,7 +184,8 @@ class My_Account_UI_V1 {
 				return __DIR__ . '/templates/v1/order-again.php';
 			case 'notices/error.php':
 				// Only override error notices on My Account pages to avoid breaking checkout validation.
-				if ( function_exists( 'is_account_page' ) && \is_account_page() ) {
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):

laurelfulford:
Thanks @chickenn00dle! 🙌

PR #4494, opened 2026-02-17
Add dynamic copyright date block

rbcorrales:
### All Submissions:

Adds a new `newspack/copyright-date` dynamic block that renders the current year with configurable prefix and suffix text. The year is derived from the site's timezone (via `wp_date` in PHP and `dateI18n` in JS), so it stays accurate regardless of the visitor's location.

The block ships with sensible defaults (© prefix, empty suffix) and disables rich text formatting on both fields in favor of inheriting the block's typography styles.

Closes NPPD-1186.

Related PR: https://github.com/Automattic/newspack-block-theme/pull/414

1. Insert the "Copyright Date" block in the block editor.
2. Verify the default rendering shows `© <current year>` in both the editor and frontend.
3. Edit the prefix (e.g., change to "Copyright") and add a suffix (e.g., "My Publication"). Confirm both render correctly on the frontend as separate `<span>` elements with spaces between parts.
4. Clear the prefix entirely and verify only the year renders (no empty span in the markup).
5. Clear the suffix and verify no suffix span appears.
6. Confirm the year matches the site's configured timezone (Settings > General > Timezone), not the browser's local time.
7. Try entering HTML in the prefix/suffix fields and verify it is escaped in the frontend output.

### Other information:

### DIFF
--- a/src/blocks/copyright-date/class-copyright-date-block.php
+++ b/src/blocks/copyright-date/class-copyright-date-block.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * Copyright Date Block.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Blocks\CopyrightDate;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Copyright_Date_Block Class
+ */
+final class Copyright_Date_Block {
+
+	/**
+	 * Block name.
+	 */
+	public const BLOCK_NAME = 'newspack/copyright-date';
+
+	/**
+	 * Initializes the block.
+	 *
+	 * @return void
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'register_block' ] );
+	}
+
+	/**
+	 * Register the copyright date block.
+	 *
+	 * @return void
+	 */
+	public static function register_block() {
+		register_block_type_from_metadata(
+			__DIR__ . '/block.json',
+			[
+				'render_callback' => [ __CLASS__, 'render_block' ],
+			]
+		);
+	}
+
+	/**
+	 * Block render callback.
+	 *
+	 * @param array $attributes The block attributes.
+	 *
+	 * @return string The block HTML.
+	 */
+	public static function render_block( array $attributes ) {
+		$prefix      = $attributes['prefix'] ?? '';
+		$suffix      = $attributes['suffix'] ?? '';
+		$year        = wp_date( 'Y' );
+		$block_class = wp_get_block_default_classname( self::BLOCK_NAME );
+
+		$parts = [];
+
+		if ( '' !== $prefix ) {
+			$parts[] = sprintf(
+				'<span class="%s__prefix">%s</span>',
+				$block_class,
+				esc_html( $prefix )
+			);
+		}
+
+		$parts[] = sprintf(
+			'<span class="%s__year">%s</span>',
+			$block_class,
+			esc_html( $year )
+		);
+
+		if ( '' !== $suffix ) {
+			$parts[] = sprintf(
+				'<span class="%s__suffix">%s</span>',
+				$block_class,
+				esc_html( $suffix )
+			);
+		}
+
+		$wrapper_attributes = get_block_wrapper_attributes();
+
+		return sprintf(
+			'<div %1$s>%2$s</div>',
+			$wrapper_attributes,
+			implode( ' ', $parts )
+		);
+	}
+}
+
+Copyright_Date_Block::init();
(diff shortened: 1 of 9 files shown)


tests/unit-tests/class-test-copyright-date-block.php
```diff
@@ -0,0 +1,195 @@
...
+			[
+				'prefix' => "\u{00a9}",
+				'suffix' => 'Acme Inc',
+			]
+		);
+
+		$prefix_pos = strpos( $output, $this->block_class . '__prefix' );
+		$year_pos   = strpos( $output, $this->block_class . '__year' );
+		$suffix_pos = strpos( $output, $this->block_class . '__suffix' );
+
```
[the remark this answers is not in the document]
rbcorrales: The test would still fail either way since the `assertStringContainsString` calls catch missing spans. The `strpos` false coercion could only make an `assertGreaterThan` pass misleadingly, not the whole test. That said, reordered the assertions so the containment checks run first, making the failure point clearer if a span is ever missing: [de0b668](https://github.com/Automattic/newspack-plugin/pull/4494/commits/de0b668b91e39f826ba204f726701b4b7784a9a8).

src/blocks/copyright-date/edit.jsx
```diff
@@ -0,0 +1,50 @@
...
+	return (
+		<div { ...blockProps }>
+			<RichText
+				className={ `${ blockClass }__prefix` }
+				tagName="span"
+				placeholder={ __( 'Prefix…', 'newspack-plugin' ) }
+				value={ prefix }
+				onChange={ value => setAttributes( { prefix: value } ) }
+				allowedFormats={ [] }
+			/>{ ' ' }
```
laurelfulford:
```suggestion
			/>
```
Super-specific nit, but we may want to move the spaces into the placeholders/into the default text to make the layout a hair more flexible - I'm specifically thinking about adding a date range, like: &copy; 2020-2026:

Image

laurelfulford (changes requested):
This is great, @rbcorrales! I have one nit about spacing, but that's it -- it works well, the markup acts as advertised!

Also what do you think about supporting the link option in the block toolbar? I do like how you're using the Site Title in the patterns -- I think that's a better way to go for that -- but for cases when a site has the parent company/another organization name in the copyright it might be nice to be able to link that to whatever that org is within the block.

src/blocks/copyright-date/edit.jsx
rbcorrales: Moved the space into the prefix default value in `block.json`. The block theme patterns use the default attributes, so they pick up the change automatically.
[c9c9aaa](https://github.com/Automattic/newspack-plugin/pull/4494/commits/c9c9aaa129642e20487b69b198fcaf3446bb6e92)

rbcorrales:
@laurelfulford Thanks for the review! Added the ability to set up links for both the prefix and the suffix: [d477048](https://github.com/Automattic/newspack-plugin/pull/4494/commits/d4770484b7f74f4f7b065ede780c6fb368867f01).

src/blocks/copyright-date/class-copyright-date-block.php
```diff
@@ -61,7 +61,7 @@ public static function render_block( array $attributes ) {
 			$parts[] = sprintf(
 				'<span class="%s__prefix">%s</span>',
 				$block_class,
-				esc_html( $prefix )
+				wp_kses_post( $prefix )
```
laurelfulford: Non-blocking nit: this & the suffix sanitization could probably be stripped down to just run `wp_kses()` with allowed HTML - looks like all the possible attributes are:
```
		$allowed_html = array(
			'a' => array(
				'href' => true,
				'class' => true,
				'target' => true,
				'rel' => true,
			)
		);

		if ( '' !== $prefix ) {
			$inner .= sprintf(
				'<span class="%s__prefix">%s</span>',
				$block_class,
				// wp_kses_post( $prefix )
				wp_kses( $prefix, $allowed_html ),
			);
		}
```

laurelfulford (approved):
Thanks @rbcorrales! The changes look good and the links work well! 

I added a non-blocking nit (just to switch the sanitization for the prefix/suffix to something with fewer checks) but that's minor!

src/blocks/copyright-date/class-copyright-date-block.php
rbcorrales: Makes sense. Added here: [6152fd0](https://github.com/Automattic/newspack-plugin/pull/4494/commits/6152fd067922db5fd453fbb4cdc9a62f5d013d59).

PR #4496, opened 2026-02-18
Integrations/pull contact data

leogermani:
### All Submissions:

Implements the pull contact data strategy.

It triggers an async pull every five minutes when the user is active in the site.

If the user is returning after some time away:
* It will trigger an immediate async pull if they were around in the last 24h
* it will trigger a synchronous pull if last time they were around was over 24h ago

The idea is that we want to get fresh data on their first request to be able to properly segment them.

But, we time cap this - the sync request should not hold the page load for too long.

The goal here is to make sure the user is properly segmented on the first page load when they are back after some time away.

TODOs:
* We might need to change how we store selected fields when we have more than just keys there (data types, etc)
* Need to surface pull errors. Wonder if we can combine those with other errors in the dedicated view we are going to build. How to store them?

1. Open WP Shell and set a selected field for the ESP integration. Ex:

```
$i = new Newspack\Reader_Activation\Integrations\ESP();
$i->set_selected_fields( [ 'HAS_SOMETHING' ] );
```

2. Add a value to this metadata in a contact directly in your ESP dashboard
3. Visit the site logged in as that reader. Look for something like this in the logs:

```
[NEWSPACK][Newspack\Reader_Activation\Integrations\Contact_Pull::pull_single_integration]: Pulled data from esp: {"HAS_SOMETHING":"yes"}
[NEWSPACK][Newspack\Reader_Activation\Integrations\Contact_Pull::pull_sync]: Loopback pull succeeded for esp.
```

4. Wait 5 minutes - or edit the code to make the interval shorter - and watch it again in the logs
5. See the actions piling up with: `wp action-scheduler action list --hook=newspack_pull_integration_contact_data`
6. Verify that the metadata value was stored as user meta:

```
newspack_reader_data_item_HAS_SOMETHING                         | yes
```

7. Clear the np_integrations_last_pull metadata for the user and add a sleep in the integration callback (or lower the timeout limit).
8. Confirm you see the error in the log and that an action is scheduled right away (all should be printed in the logs)

### Other information:


includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -0,0 +1,234 @@
...
+		foreach ( $integrations as $id => $integration ) {
+			unset( $remaining_integrations[ $id ] );
+
+			$elapsed = microtime( true ) - $start;
+			if ( $elapsed >= self::PULL_TIME_LIMIT ) {
+				Logger::log( 'Pull routine time limit reached. Scheduling remaining integrations async.' );
+				// Re-add the current integration that didn't get a chance to run.
+				$remaining_integrations = [ $id => $integration ] + $remaining_integrations;
+				break;
+			}
```
miguelpeixe: This won't protect us from a slow request. If any integration pull exceeds our limit, we'll only be aware of it in the next loop and bail the next integration from making it worse.

An alternative solution is to run this behind an ajax action from the backend, similar to how we currently run data events, but instead of fire-and-forget with 0 timeout, we set it to the allowed limit and process the response to trigger the async flow. Even if it succeeds after the configured request timeout, it's a win because we get the data on the next page load and secure a retry.

includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @var int
+	 */
+	const PULL_SYNC_THRESHOLD = 86400;
+
+	/**
+	 * Max seconds for the entire synchronous pull routine.
+	 *
+	 * @var int
+	 */
+	const PULL_TIME_LIMIT = 5;
```
miguelpeixe: This is a lot of seconds to let a reader hanging. I suggest something around 700ms or less

includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	/**
+	 * Pull data from a single integration and store selected fields.
+	 *
+	 * @param int                                     $user_id     WordPress user ID.
+	 * @param \Newspack\Reader_Activation\Integration $integration The integration instance.
+	 * @param int                                     $timeout     Max seconds for this call.
+	 */
+	public static function pull_single_integration( $user_id, $integration, $timeout = 30 ) {
```
miguelpeixe:
```suggestion
	protected static function pull( $user_id, $integration, $timeout = 30 ) {
```

Non-blocking, but it doesn't seem like we need to expose this method, and the name looks redundant

miguelpeixe (changes requested):
I still haven't gone through testing everything, but I wanted to respond to the sync pull strategy, since it's an important aspect of this.

includes/reader-activation/integrations/class-contact-pull.php
leogermani: ah of course. yes!

includes/reader-activation/integrations/class-contact-pull.php
leogermani: It's invoked in tests... and I can see this being useful for CLI, or for manually triggering pulls

includes/reader-activation/integrations/class-contact-pull.php
leogermani: I tried 1 but we have to consider that there's the overhead of loading WP... At least in my local environment, not even 2 seconds was enough. I'll leave it at 2 for now.

See my top level comment about yet having to validate if this is gonna be useful or not. Then we can revisit

includes/reader-activation/integrations/class-contact-pull.php
leogermani: Not sure how I completely missed it... We've been using this aproach for async requests so much that I forgot that this is actually synchronous by default.

leogermani:
@miguelpeixe thanks for pointing me to the obvious!

Looks good now, but there's still an open question we might be able to answer later in the project:

If we update the Reader data at that point, will the data be available in time to be used by the content gate checks, and the popups segmentation?

In a quick test I did, the data added is not available in the `the_post` hook yet, we probably need to reload all usermeta for that user, because the update happened in another request. Need to look a bit deeper into this.

Also, I don't see the data being hydrated into the browser local storage... which I think is expected (and I suspect I saw happening the other day...)

miguelpeixe:
> In a quick test I did, the data added is not available in the the_post hook yet, we probably need to reload all usermeta for that user, because the update happened in another request. Need to look a bit deeper into this.

Because it executed in a separate PHP process, it's likely that the data was already fetched from the object cache, so it's already in memory and need to be flushed/refetched. Maybe `wp_cache_flush_runtime()`?

includes/reader-activation/integrations/class-contact-pull.php
miguelpeixe: Can we place this behind a filter? Our local envs are notoriously not the fastest.

leogermani: yes, good call

leogermani: done

leogermani:
> > In a quick test I did, the data added is not available in the the_post hook yet, we probably need to reload all usermeta for that user, because the update happened in another request. Need to look a bit deeper into this.
> 
> Because it executed in a separate PHP process, it's likely that the data was already fetched from the object cache, so it's already in memory and need to be flushed/refetched. Maybe `wp_cache_flush_runtime()`?

We can look into it later when we start using this data for segmentation and access control

includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -0,0 +1,323 @@
...
+				Logger::log( 'Pull error from ' . $integration->get_id() . ': ' . $data->get_error_message() );
+				return $data;
+			}
+
+			$selected_keys = array_flip( $selected_fields );
+			$data          = array_intersect_key( $data, $selected_keys );
+			Logger::log( 'Pulled data from ' . $integration->get_id() . ': ' . wp_json_encode( $data ) );
+
+			foreach ( $data as $key => $value ) {
+				\Newspack\Reader_Data::update_item( $user_id, $key, $value );
```
miguelpeixe:
```suggestion
				\Newspack\Reader_Data::update_item( $user_id, $key, wp_json_encode( $value ) );
```

Step 6 of the test instructions was failing for me because the reader data item value must be a JSON-encoded string.

includes/reader-activation/integrations/class-esp.php
```diff
@@ -104,6 +105,36 @@ public function push_contact_data( $contact, $context = '', $existing_contact =
...
+		if ( ! $this->can_sync() ) {
+			return new \WP_Error( 'missing_dependency', __( 'ESP Integration is not fully configured.', 'newspack-plugin' ) );
+		}
+
+		$user = get_userdata( $user_id );
+		if ( ! $user ) {
+			return new \WP_Error( 'user_not_found', __( 'User not found.', 'newspack-plugin' ) );
+		}
+
+		$contact_data = Newspack_Newsletters_Subscription::get_contact_data( $user->user_email, true );
```
miguelpeixe, line 126: Could use a `class_exists` here to avoid a fatal error. This could live under `can_sync()` for broader coverage.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -15,6 +15,13 @@
  * This class should be extended by specific integration implementations.
  */
 abstract class Integration {
+	/**
+	 * Option name prefix for storing selected fields per integration.
+	 *
+	 * @var string
+	 */
+	const OPTION_PREFIX = 'np_integration_selected_fields_';
```
miguelpeixe: We usually go for `newspack_` as the prefix. The database limit is 191 characters, it should hold fine.

includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -0,0 +1,323 @@
...
+	 * @var string
+	 */
+	const NONCE_ACTION = 'newspack_pull_integration_nonce';
+
+	/**
+	 * User meta key for last pull timestamp.
+	 *
+	 * @var string
+	 */
+	const LAST_PULL_META = 'np_integrations_last_pull';
```
miguelpeixe: Same here, `newspack_` as the prefix

includes/reader-activation/integrations/class-esp.php
```diff
@@ -104,6 +105,36 @@ public function push_contact_data( $contact, $context = '', $existing_contact =
...
+	 * Pull contact data from the ESP for a given user.
+	 *
+	 * @param int $user_id WordPress user ID.
+	 *
+	 * @return array|\WP_Error Associative array of field_key => value pairs on success, WP_Error on failure.
+	 */
+	public function pull_contact_data( $user_id ) {
+		if ( ! $this->can_sync() ) {
+			return new \WP_Error( 'missing_dependency', __( 'ESP Integration is not fully configured.', 'newspack-plugin' ) );
+		}
```
miguelpeixe: We can benefit from better error handling by setting `$return_errors = true` in `can_sync()` here.

miguelpeixe (changes requested):

includes/reader-activation/integrations/class-contact-pull.php
leogermani: aha!

includes/reader-activation/integrations/class-esp.php
leogermani: it is behind can_sync already

includes/reader-activation/integrations/class-esp.php
leogermani: good call

leogermani: https://github.com/Automattic/newspack-plugin/pull/4496/commits/0ed1444f709658dc4225310e590b01d52d696bd8

includes/reader-activation/integrations/class-contact-pull.php
leogermani: https://github.com/Automattic/newspack-plugin/pull/4496/commits/3e232a5314206f6b3d1988171225cc3ee421b700

includes/reader-activation/integrations/class-integration.php
leogermani: https://github.com/Automattic/newspack-plugin/pull/4496/commits/3e232a5314206f6b3d1988171225cc3ee421b700

includes/reader-activation/integrations/class-esp.php
miguelpeixe: Ugh, morning brain. Sorry about that

miguelpeixe (changes requested):
Need to update the tests

miguelpeixe (approved):

PR #4500, opened 2026-02-19
fix(my-account): don't show subscription payment notice in modal checkout

dkoo:
### All Submissions:

Fixes a minor error notice rendering bug found while testing #4498. Also fixes a PHP warning, and makes the subscription payment snackbar notice "urgent" so it requires manual dismissal instead of auto-hiding after 5 seconds.

1. If #4498 hasn't been merged to trunk yet, follow steps 1-5 of testing instructions for that PR to trigger the broken modal checkout window. If it has been merged, then you can force an error by temporarily undoing the changes in that PR and following steps 1-5 to trigger the broken modal checkout window.
2. Observe that the broken modal checkout renders with both a snackbar notice and an error message:

Screenshot 2026-02-19 at 4 48 48 PM

3. Also observe a PHP notice in debug.log: `PHP Notice:  Only variables should be passed by reference in /newspack-repos/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php on line 142`
4. Check out this branch and trigger the broken modal window again.
5. This time, confirm that the modal appears without the snackbar notice. Also confirm no PHP notice.

Screenshot 2026-02-19 at 4 46 20 PM

6. After clicking "Go back" to close the broken checkout modal, confirm that the subscription payment message appears as a snackbar in My Account pages and does not auto-hide (should close only after you click the × button in the notice). 

Screenshot 2026-02-19 at 4 57 37 PM

### Other information:

### DIFF
--- a/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
+++ b/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
@@ -145,6 +145,9 @@ class WooCommerce_Update_Payment_Notice {
 			}
 
 			$line_items = $subscription->get_items();
+			if ( empty( $line_items ) ) {
+				continue;
+			}
 			$line_item  = reset( $line_items );
 			$product    = wc_get_product( $line_item->get_product_id() );
 			// If the product has a parent, use the parent product.
--- a/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
+++ b/includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
@@ -85,6 +85,11 @@ class WooCommerce_Update_Payment_Notice {
 			return;
 		}
 
+		// Don't show in modal checkout frame.
+		if ( method_exists( 'Newspack_Blocks\Modal_Checkout', 'is_modal_checkout' ) && \Newspack_Blocks\Modal_Checkout::is_modal_checkout() ) {
+			return;
+		}
+
 		$notices = self::get_notices();
 		if ( empty( $notices ) ) {
 			return;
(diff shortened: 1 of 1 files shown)


includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
```diff
@@ -189,6 +195,18 @@ function( $key, $value ) {
...
+				'newspack_ui_notice_is_urgent',
+				function( $is_urgent, $printed_notice ) use ( $notice ) {
+					if ( $printed_notice === $notice ) {
+						$is_urgent = true;
+					}
+					return $is_urgent;
+				},
+				10,
+				2
+			);
```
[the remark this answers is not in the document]
dkoo: Unfortunately not the case—the filter is necessary to ensure the snackbars are rendered with `'autohide' => false`.

includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
```diff
@@ -189,6 +198,18 @@ function( $key, $value ) {
...
```
miguelpeixe, line 212: Can we move this filter outside the loop and `use ( $notices )` instead?

miguelpeixe (changes requested):
Tests well!

includes/plugins/woocommerce/class-woocommerce-update-payment-notice.php
dkoo: This was the only way I could figure out how to match the exact message in the filter callback, since notices don't have unique IDs to reference. We don't want to make all notices urgent, just the ones we're adding for this scenario.

miguelpeixe (approved):

PR #4503, opened 2026-02-20
feat(sync): add batch queue for contact syncing

miguelpeixe:
### All Submissions:

Introduces a new `Contact_Sync_Batch` class that provides batch handling for contact data syncing via ActionScheduler. Instead of scheduling one AS action per user (admin) or running a synchronous loop (CLI), bulk sync operations are now chunked and processed asynchronously with progress tracking.

- **Admin bulk action updated** to use the batch queue instead of per-user AS scheduling
- **CLI sync updated** from a synchronous loop to collect → enqueue → poll with a WP-CLI progress bar
- Progress records stored in `wp_options` with automatic 24h cleanup
- Batch cancellation only affects the target batch's pending actions

1. Enable ESP sync admin tools: `define( 'NEWSPACK_ESP_SYNC_ADMIN', true );` and ensure at least one ESP integration is connected and enabled.
2. Go to Users, select multiple users, choose "Resync to the ESP" from Bulk Actions, and apply. Verify the success notice says "X contacts scheduled for resync" and the monitoring link points to `newspack_contact_sync_batch` actions.
3. Run `wp newspack esp sync --batch-size=5` and verify a progress bar displays and completes.
4. Run `wp newspack esp sync --dry-run` to verify it counts contacts without syncing.
5. Run `wp newspack esp sync --user-ids=1,2,3` to verify specific user sync via the batch queue.

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -103,6 +103,7 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-woocommerce.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync-admin.php';
+		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync-batch.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-integrations.php';
 		\Newspack\Reader_Activation\Integrations::init();
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-utils.php';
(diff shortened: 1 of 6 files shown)


leogermani (commented):
It worked well on admin, but no so much on CLI.

on CLI, when I ran the command it just stayed there with no action until I navigated to the site for it to trigger the scheduled actions... (and then it was a pain to cancel them)

I think it doesn't make sense to use this approach for CLI where there's no timeout issue.

Anywhere else you're planning to use this? What triggered it?

miguelpeixe:
> on CLI, when I ran the command it just stayed there with no action until I navigated to the site for it to trigger the scheduled actions... (and then it was a pain to cancel them)

Yeah, it'll hang until AS gets triggered, which will take longer than just doing it synchronously.

> I think it doesn't make sense to use this approach for CLI where there's no timeout issue.

Placing this behind our AS strategy is ideal to handle errors and rate limiting, which can happen under batch runs. I guess since the implementation of the retry directly in the contact sync, this is less problematic.

PR #4509, opened 2026-02-24
feat(author-profile-social): add support for colors, block spacing, brand style

thomasguillot:
### All Submissions:

This PR adds a bunch of enhancements to the block:

* New style: "Brand" that inherits the brand's colours
* If default style selected: User can control Icon and icon background colours (I used Core's Color panel instead of custom one because of ordering. I renamed "Text" and "Background" to "Icon color" and "Icon background" with some JS)
* We now support block spacing
* I removed the padding setting

We are extracting the gaps and colours to apply them to the actual link

https://github.com/user-attachments/assets/a51170eb-001e-4980-9d28-dedf4778f026

1. Add an author profile block
2. Edit the social links
3. Switch to this branch and refresh
4. You should see more settings in the sidebar.
5. Test the colours
6. The spacings (gap and margin)
7. Test brand vs default

### Other information:


src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -202,30 +202,117 @@ private static function get_block_wrapper_attributes( WP_Block $block, array $at
...
 
 	/**
-	 * Build wrapper style from block attributes.style (spacing, color, border, etc.) and block-specific vars.
-	 * Uses the style engine so presets (e.g. var:preset|spacing|20) are converted to CSS.
+	 * Convert a preset token (var:preset|type|slug) to a CSS variable reference.
+	 *
+	 * @param string $value Raw value, e.g. "var:preset|spacing|20" or "#fff".
+	 * @return string CSS value, e.g. "var(--wp--preset--spacing--20)" or "#fff".
+	 */
+	private static function preset_to_css( string $value ): string {
```
laurelfulford, line 254: Looking at Core's Social Links block, it uses the `<ul>` as the outer wrapper, rather than [an extra `<div>`](https://github.com/Automattic/newspack-plugin/blob/343c3ae32af4064d377eadd73f9e111905ffc4e7/src/blocks/author-profile-social/class-author-profile-social-block.php#L113) that the author profile social block uses. 

I _think_ we can rejig it a bit so the `<ul>` is the outermost container, then the block gap stuff should just work without this workaround (cc @rbcorrales in case I am horribly wrong)! This looks like it works, but it'd be nice to take advantage of the baked in way to do it if we can 🙂

src/blocks/author-profile-social/edit.jsx
```diff
@@ -40,16 +89,75 @@ const fetchAllServiceKeys = () => {
...
+	const gapVars = resolveBlockGap( styleAttr?.spacing?.blockGap );
+
+	const blockProps = stripColorFromBlockProps( useBlockProps( { className: 'wp-block-newspack-author-profile-social' } ) );
+	blockProps.style = {
+		...blockProps.style,
+		'--icon-size': `${ roundIconSize( iconSizeValue ) }px`,
+		...gapVars,
+		...( ! isBrand && iconColor && { '--icon-color': iconColor } ),
+		...( ! isBrand && iconBackground && { '--icon-background': iconBackground } ),
+	};
```
laurelfulford: Nit: this just tweaks the order a bit so blockProps isn't defined and then immediately mutated (it skips that by saving the original blockProps value in a separate variable). It's strictly a readability tweak, either works the same:

```suggestion
const rawProps = stripColorFromBlockProps( useBlockProps( { className:
  'wp-block-newspack-author-profile-social' } ) );
  const blockProps = {
      ...rawProps,
      style: {
          ...rawProps.style,
          '--icon-size': `${ roundIconSize( iconSizeValue ) }px`,
          ...gapVars,
          ...( ! isBrand && iconColor && { '--icon-color': iconColor } ),
          ...( ! isBrand && iconBackground && { '--icon-background': iconBackground } ),
      },
  };
```

src/blocks/author-profile-social/edit.jsx
```diff
@@ -40,16 +89,75 @@ const fetchAllServiceKeys = () => {
...
+	const { iconSize, style: styleAttr, textColor, backgroundColor, className } = attributes;
 	const hasPopulated = useRef( false );
 	const [ allServiceKeys, setAllServiceKeys ] = useState( null ); // null = loading
 
-	const blockProps = useBlockProps( {
-		className: 'wp-block-newspack-author-profile-social',
-		style: {
-			'--icon-size': `${ roundIconSize( iconSize ) }px`,
-		},
+	const isBrand = ( className || '' ).includes( 'is-style-brand' );
```
laurelfulford: Nit: `.includes()` can match a CSS class name with `is-style-brand` part of it of it, not just a 1:1 match.  

I think that's really unlikely to come up as an issue in this case, but just in case tweaking it would make it only match `is-style-brand` as a whole class name:

```suggestion
	const isBrand = ( className || '' ).split( ' ' ).includes( 'is-style-brand' );
```

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -202,30 +202,117 @@ private static function get_block_wrapper_attributes( WP_Block $block, array $at
...
-		$parts = [];
-		$style = $attributes['style'] ?? null;
-		if ( ! empty( $style ) && is_array( $style ) ) {
-			$styles = wp_style_engine_get_styles(
-				$style,
-				[ 'context' => 'block-supports' ]
-			);
-			if ( ! empty( $styles['css'] ) ) {
-				$parts[] = $styles['css'];
+		$parts    = [];
```
laurelfulford, line 291: The removal of the `$style` code here appears to have broken the margin styles on the front-end:

```suggestion
		$parts    = [];
		$style = $attributes['style'] ?? null;
		
		...
		
		if ( ! empty( $style ) && is_array( $style ) ) {
			$styles = wp_style_engine_get_styles(
				$style,
				[ 'context' => 'block-supports' ]
			);
			if ( ! empty( $styles['css'] ) ) {
				$parts[] = $styles['css'];
			}
		}
```

src/blocks/author-profile-social/edit.jsx
```diff
@@ -40,16 +89,75 @@ const fetchAllServiceKeys = () => {
...
+	const iconSizeValue = typeof iconSize === 'number' ? iconSize : parseInt( iconSize ?? 24, 10 ) || 24;
+	const iconColor = resolveColor( textColor, styleAttr?.color?.text );
+	const iconBackground = resolveColor( backgroundColor, styleAttr?.color?.background );
+
+	// Hide color panel when "Brand" is active; rename labels when "Default".
+	useEffect( () => {
+		const inspector = document.querySelector( '.block-editor-block-inspector' );
+		if ( ! inspector ) {
+			return;
+		}
```
laurelfulford: Right now the 'hide colour' useEffect runs on every render, which works visually but it means it fires when unrelated changes happen, too. 

I tried to fix this by adding `isBrand` as a dependency (so it only fired when you switched between Default and Brand) but that didn't totally fix it (if the brand was already selected when you opened the editor, the Color panel would still be visible). 

Moving the observer to a parent of `.block-editor-block-inspector`, `.interface-complementary-area` seems to help, and means we don't have to watch `document.body`.

This change is split over two suggestions; just let me know if it doesn't work as expected for you, though!

```suggestion
		const sidebar = document.querySelector( '.interface-complementary-area' );
		if ( ! sidebar ) {
			return;
		}
```

src/blocks/author-profile-social/edit.jsx
```diff
@@ -40,16 +89,75 @@ const fetchAllServiceKeys = () => {
...
+				updateColorPanel( document.body );
+			}
+		} );
+		bodyObserver.observe( document.body, { childList: true, subtree: true } );
+
+		return () => {
+			inspectorObserver.disconnect();
+			bodyObserver.disconnect();
+		};
 	} );
```
laurelfulford: This is part 2 of the useEffect changes:

```suggestion
		updateColorPanel( sidebar );

		const observer = new MutationObserver( () => updateColorPanel( sidebar ) );
		observer.observe( sidebar, { childList: true, subtree: true } );

		return () => observer.disconnect();
	}, [ isBrand ] );
```

laurelfulford (dismissed):
This is looking really cool, @thomasguillot! 

I have a couple suggestions and some not-necessary nits. Just let me know if anything doesn't make sense, or doesn't line up with how it seems to work for you!

src/blocks/author-profile-social/edit.jsx
thomasguillot: Fixed: https://github.com/Automattic/newspack-plugin/pull/4509/commits/5d10a0f325313927595a73ea2c06d50b9df8b39e

src/blocks/author-profile-social/edit.jsx
thomasguillot: Fixed: https://github.com/Automattic/newspack-plugin/pull/4509/commits/5d10a0f325313927595a73ea2c06d50b9df8b39e

src/blocks/author-profile-social/class-author-profile-social-block.php
thomasguillot: One issue with this approach is that it would leave us with two separate versions of the block that use different structures: the legacy version built with `<div><ul>`, and the InnerBlocks version that uses only `<ul>`.

rbcorrales: I think @laurelfulford is right here. Since the `InnerBlocks` path (v2) is brand new and no site is using it yet, now is the best time to get the wrapper structure right.

The two render paths (`render_social_flat` and `render_social_with_inner_blocks`) are already structurally different, and old blocks will keep using the flat path based on whether `$block->inner_blocks` is populated, not based on the active theme. So even when sites eventually migrate to the block theme, existing blocks will continue rendering through the flat path. The two structures won't conflict on the same block instance.

The CSS might need a small adjustment to handle both wrapper shapes, but I think it might be worth it to use the native block gap support in v2 instead of the var workaround.

rbcorrales:
Removing the review label since there are still changes requested.

I think Thomas might be AFK for a few days, so @laurelfulford feel free to incorporate those markup changes you suggested, since you might be more familiar with that than me 😅 

Otherwise, I can take a stab at that at some point later next week.

laurelfulford:
Thanks @rbcorrales! Taking a swing at updating that now - I think @thomasguillot tackled everything else 🙂

laurelfulford:
I think I got the block markup sorted in [acfc2ea](https://github.com/Automattic/newspack-plugin/pull/4509/commits/acfc2eaf10b4e80b2776396e89cf93fd3a703bd6) -- the colour/icon blocks still had to work the original way, but the margins and gap should now be coming from WordPress magic!

rbcorrales (approved):
@laurelfulford Thanks for making these improvements. I found two issues while testing with how the block behaves in the editor:

**1. Block Spacing control wasn't working for me in the editor**

The block uses `<InnerBlocks>` as a child of the `<ul>` wrapper, but WordPress injects layout classes (`is-layout-flex`, scoped container classes) only through `useInnerBlocksProps`. Without those classes on the wrapper, the scoped gap CSS generated by the Block Spacing control has no matching selector, so changing the slider has no visible effect in the editor. The frontend is fine because WP handles layout classes server-side.

**2. Default icon gap was too wide**

With native WP layout, the block inherits the theme's global `blockGap` (2rem in newspack-block-theme), which is way too much space between small circular icons. The old code defaulted to `--wp--preset--spacing--20` (0.5rem).

Image

Core's `social-links` block gets its tight default from a `blockGap` entry in the block theme's `theme.json`, but our block doesn't have one, and we shouldn't require every theme to add it.

Image

---

So, I pushed a fix in b100210 with two changes:

1. Switched from `<InnerBlocks>` to `useInnerBlocksProps`, same pattern Core's `social-links` uses. This causes WP to inject the layout classes into the wrapper `<ul>`, so the Block Spacing control works in the editor. Also removed the now-redundant `display: flex` rules from `editor.scss` since `is-layout-flex` handles that.

2. Added a `wp_theme_json_data_blocks` filter that sets a default `blockGap` of `spacing-20` for this block. This sits at the "blocks" layer in the theme.json cascade (default < blocks < theme < user), so themes can still override it, and user block spacing overrides work correctly. I went with this approach over CSS because theme.json-level styles get the right specificity and cascade position automatically.

https://github.com/user-attachments/assets/afe09d32-ebfe-40a4-8332-617f0a2a85de

One thing to be aware of is that the Block Spacing slider shows empty/leftmost on a freshly inserted block, even though the gap is visually correct (8px). This is expected WP behavior for theme.json-level defaults. The slider only reflects explicitly saved block attributes, not inherited defaults.

Unrelated, but I noticed Core's `social-links` block has a similar issue: once you change the Block Spacing from the default, the gap renders without any spacing in the editor regardless of the value:

https://github.com/user-attachments/assets/e1992eeb-ee34-4f27-aec6-f6abbfe65e8e

I'm going to approve the PR since it's working fine with our collective changes on my end 😄, but feel free to let me know if there's something that needs updating.

laurelfulford:
Thanks @rbcorrales! I took the latest for a spin and retested the sizing/spacing, everything looks good to me! Thank you for taking this one to the finish line, I think it's good to go! 🙂

rbcorrales:
I'll go ahead and merge it so we can get this closed. Thanks!

PR #4513, opened 2026-02-24
Fix Avatar block duotone filter for custom colors

rbcorrales:
### All Submissions:

The Avatar block's duotone filter was not applied on the frontend when using custom (non-preset) colors. Preset duotone values (e.g. "Magenta and yellow") worked, but picking custom colors from the color picker had no effect.

The root cause was the `selectors.filter.duotone` value in `block.json`. It was set to `.newspack-avatar-wrapper img`, but WordPress core's duotone system scopes CSS by prepending the filter class as a compound selector (no space): `.wp-duotone-xxx.newspack-avatar-wrapper img`. Since the filter class is added to the outer block wrapper and `.newspack-avatar-wrapper` is an inner element, the compound selector never matched.

Preset duotone happened to work because a manual `newspack_get_duotone_class_name()` method copied the class to the inner wrapper. But that method only handled preset values (`var:preset|duotone|slug`), not custom color arrays.

The fix updates the selector to `.wp-block-newspack-avatar .newspack-avatar-wrapper img`, which correctly scopes to the block wrapper where core adds the filter class. This makes the manual duotone class propagation redundant, so it has been removed from both PHP and JS.

Closes NPPD-1269.

1. Add an Avatar block to a post. Apply a custom duotone filter (pick two custom colors, not a preset). Verify the filter shows in the editor preview and on the frontend.
2. Add an Avatar block to a template in the Site Editor (e.g. Single Posts). Apply a custom duotone filter. Save, then view a post using that template. Verify the filter renders on the frontend.
3. Add an Author Profile block (nested/v2 mode) to a post with a specific author. Set a custom duotone on the inner Avatar block. Verify it renders in the editor and on the frontend.
4. Edit a template in the Site Editor and add a contextual Author Profile block. Set a custom duotone on the inner Avatar block. Save and verify the filter renders on the frontend for all authors.
5. In any of the above scenarios, also test with a preset duotone (e.g. "Magenta and yellow") to confirm no regression.

### Other information:

### DIFF
--- a/src/blocks/avatar/avatar-wrapper.jsx
+++ b/src/blocks/avatar/avatar-wrapper.jsx
@@ -6,11 +6,8 @@ import clsx from 'clsx';
 /**
  * WordPress dependencies
  */
-import {
-	useBlockProps,
-	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
-	__experimentalUseBorderProps as useBorderProps,
-} from '@wordpress/block-editor';
+// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
+import { __experimentalUseBorderProps as useBorderProps } from '@wordpress/block-editor';
 import { __ } from '@wordpress/i18n';
 import { useEffect, useState } from '@wordpress/element';
 import { addQueryArgs, removeQueryArgs } from '@wordpress/url';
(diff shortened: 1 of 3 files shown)


thomasguillot:
It works! 🎉 Thank you @rbcorrales!

miguelpeixe (approved):

rbcorrales:
Thanks for the review @thomasguillot and @miguelpeixe!

PR #4517, opened 2026-02-25
IP access cache booster endpoint

leogermani:
## Summary

- Adds a `/institutional-access` endpoint that checks the reader's IP via the `newspack_content_gate_check_ip` filter
- If the IP is valid, sets a `wp_nocache_ip` cookie (24h TTL) so subsequent requests bypass cache and allow server-side IP checking
- Redirects the reader to the homepage after processing
- Flushes rewrite rules once on first deployment

## Test plan

- [ ] Visit `/institutional-access` on the site and confirm you are redirected to the homepage
- [ ] Confirm that by default no `wp_nocache_ip` cookie is set (the filter returns `false`)
- [ ] To test the valid IP flow, temporarily add the following snippet to a plugin or `functions.php`:
  ```php
  add_filter( 'newspack_content_gate_check_ip', '__return_true' );
  ```
- [ ] Visit `/institutional-access` again and confirm the `wp_nocache_ip` cookie is now set with a 24h expiration
- [ ] Remove the snippet (or change to `__return_false`) and visit the endpoint again to confirm the cookie is **not** set on new requests

### DIFF
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -84,6 +84,7 @@ class Content_Gate {
 		include __DIR__ . '/class-metering.php';
 		include __DIR__ . '/class-metering-countdown.php';
 		include __DIR__ . '/content-gifting/class-content-gifting.php';
+		include __DIR__ . '/class-ip-access-rule.php';
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works as described!

miguelpeixe (approved):

PR #4539, opened 2026-03-03
fix(block-theme): ensure content gate contents inserted in time for styles to be rendered

laurelfulford:
### All Submissions:

This PR is an alternative to https://github.com/Automattic/newspack-plugin/pull/4449 -- rather than looping through the blocks in the content gate to build the styles, this PR moves where the content gate is inserted so the styles "just work". 

Certain block styles are generated as needed by WordPress, and they're not automatically available. By the time the overlay gate is injected at `wp_footer`, it's "too late" for its blocks to be included in those inline styles. This means certain styles -- like block alignments and spacing -- don't work in the gates.

This PR moves where the gate is injected, to `get_footer()` for the classic theme, and inserted at the post-content block for the Block Theme. The latter necessitated a bit of CSS tweaks to make sure the gate still looked right. 

Closes NPPD-1181

1. Start with the Classic Theme.
2. Set up a content gate (using Memberships or the new Newspack approach).
3. In the Content gate, insert the Registration Card (compact) pattern (the second one in the list of Content Gate block patterns.
4. Publish.
5. In an incognito window, trigger the gate on a post - note the layout is a little messed up (the button and text should be on opposite sides, and the spacing is wrong between the two lines of text):

CleanShot 2026-03-03 at 15 44 59

6. Apply this PR and run npm run build.
7. Repeat step 5 and confirm the blocks look closer to this:

CleanShot 2026-03-03 at 15 46 11

8. Switch to the Block Theme.
9. Repeat step 7 and confirm things still look okay:

CleanShot 2026-03-03 at 15 46 53

### Other information:

### DIFF
--- a/includes/content-gate/class-content-gate.php
+++ b/includes/content-gate/class-content-gate.php
@@ -73,11 +73,7 @@ class Content_Gate {
 		add_action( 'admin_init', [ __CLASS__, 'handle_edit_gate_layout' ] );
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
-		if ( function_exists( 'wp_is_block_theme' ) && wp_is_block_theme() ) {
-			add_filter( 'render_block', [ __CLASS__, 'inject_overlay_gate_after_post_content_block' ], 10, 2 );
-		} else {
-			add_action( 'get_footer', [ __CLASS__, 'render_overlay_gate' ], 1 );
-		}
+		add_action( 'after_setup_theme', [ __CLASS__, 'register_overlay_gate_hooks' ] );
 		add_action( 'before_delete_post', [ __CLASS__, 'delete_gate_layouts' ], 10, 2 );
 		add_filter( 'newspack_popups_assess_has_disabled_popups', [ __CLASS__, 'disable_popups' ] );
 		add_filter( 'newspack_reader_activity_article_view', [ __CLASS__, 'suppress_article_view_activity' ], 100 );
@@ -870,6 +866,20 @@ class Content_Gate {
 	}
 
 	/**
+	 * Register overlay gate hooks after the theme has been set up.
+	 *
+	 * Deferred to after_setup_theme so that wp_is_block_theme() can be called safely,
+	 * after theme directories have been registered.
+	 */
+	public static function register_overlay_gate_hooks() {
+		if ( self::is_block_theme() ) {
+			add_filter( 'render_block', [ __CLASS__, 'inject_overlay_gate_after_post_content_block' ], 10, 2 );
+		} else {
+			add_action( 'get_footer', [ __CLASS__, 'render_overlay_gate' ], 1 );
+		}
+	}
+
+	/**
 	 * Inject overlay gate markup right after the post content block.
 	 *
 	 * Used for block themes where there aren't hooks to use in time to get do_blocks() to run.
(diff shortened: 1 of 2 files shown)


includes/content-gate/class-content-gate.php
```diff
@@ -66,7 +73,11 @@ public static function init() {
 		add_action( 'admin_init', [ __CLASS__, 'handle_edit_gate_layout' ] );
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
-		add_action( 'wp_footer', [ __CLASS__, 'render_overlay_gate' ], 1 );
+		if ( function_exists( 'wp_is_block_theme' ) && wp_is_block_theme() ) {
```
rbcorrales: Suggestion: `wp_is_block_theme()` is called at plugin load time here. This works fine in standard installs because `register_theme_directory()` runs before plugins load in `wp-settings.php`, but WordPress 6.8 added a `_doing_it_wrong` guard for calling it before theme directories are registered.

I think the main scenario where that could be a problem is if the plugin were ever loaded as an MU-plugin (which loads earlier), but that's not how Newspack works today.

Still, it might be worth deferring the hook registration to `after_setup_theme` just for robustness:


```suggestion
add_action( 'after_setup_theme', function() {
    if ( self::is_block_theme() ) {
        add_filter( 'render_block', [ __CLASS__, 'inject_overlay_gate_after_post_content_block' ], 10, 2 );
    } else {
        add_action( 'get_footer', [ __CLASS__, 'render_overlay_gate' ], 1 );
    }
} );
```
(the above is an example within the code context, but a named function would be even better than an anonymous one).

BTW the `Content_Gate_Layout` trait already has a `is_block_theme()` helper at line 157 that does the same check. Could reuse that for DRY.

includes/content-gate/class-content-gate.php
```diff
@@ -850,14 +861,41 @@ public static function render_overlay_gate() {
...
+
+		if ( $injected || self::$overlay_gate_output || ! is_singular() ) {
+			return $block_content;
+		}
+
+		if ( 'core/post-content' !== ( $block['blockName'] ?? '' ) ) {
+			return $block_content;
+		}
+
+		$injected = true;
```
rbcorrales, line 905: Nit: The static `$injected` and s`elf::$overlay_gate_output` serve overlapping but distinct purposes: `$injected` prevents re-entry even if `render_overlay_gate()` bails out early (e.g., gate style is not "overlay"), while `$overlay_gate_output` is only set when HTML is actually rendered. This dual-guard is a valid defensive pattern, but a brief inline comment explaining why both are needed would help future maintainers.

rbcorrales (approved):
Clean fix, nice approach splitting the injection point per theme type instead of trying to work around the timing issue. Tested on both classic and block themes with the Registration Card (compact) pattern, and the layout looks correct on both.

Left a couple of minor suggestions (non-blocking).

includes/content-gate/class-content-gate.php
laurelfulford: Oooo good call! Added in [8a7599f](https://github.com/Automattic/newspack-plugin/pull/4539/commits/8a7599fe77f85c60bfccdca2e659d9303a632377).

includes/content-gate/class-content-gate.php
laurelfulford: > Still, it might be worth deferring the hook registration to after_setup_theme just for robustness:

Agreed! It doesn't hurt anything and could avoid a weird issue down the road in non-standard setups 🙂 Also I think I added that `is_block_theme()` check in a parallel PR and completely forgot about it - thanks for the reminder! 😄 Added in [8a7599f](https://github.com/Automattic/newspack-plugin/pull/4539/commits/8a7599fe77f85c60bfccdca2e659d9303a632377)

laurelfulford:
Thanks @rbcorrales!! 🎉 Fixes applied, gonna merge this thing!

PR #4540, opened 2026-03-04
feat(content-gate): default block patterns

miguelpeixe:
### All Submissions:

NPPD-1191

Introduces a more advanced handling of the content gate default layouts based on the gate configuration:

1. Registration Access Layout:
   1. Without paid access, uses the `registration-wall` pattern
   2. With paid Access (metering), uses `pay-wall-one-tier-metering` pattern
2. Paid Access Layout uses `pay-wall-one-tier` pattern

1. Create 2 new gates, one with registration access alone and another with both registration and paid access
2. Edit the gates and confirm they reflect the description above

### Other information:

### DIFF
--- a/includes/content-gate/block-patterns/pay-wall-one-tier-metering.php
+++ b/includes/content-gate/block-patterns/pay-wall-one-tier-metering.php
@@ -10,23 +10,23 @@
 <hr class="wp-block-separator has-alpha-channel-opacity is-style-dots"/>
 <!-- /wp:separator -->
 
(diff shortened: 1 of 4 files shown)


dkoo (changes requested):
The new defaults work great for Registration-only and Paid Access-only gates!

I still think we need to reconcile a scenario that could confuse readers. If you create a gate that has BOTH Registration and Paid Access enabled with metering, this is the default layout created for Registration:

Image

If you also require verification in Registration settings, as a reader once you use the "Register for free" button (which launches the reg/auth modal without the new verification flow) the post will refresh but will still be gated due to the lack of verification. There's no indication to the reader about needing to verify at this piont.

Image

Maybe we need to add a modified version of the `registration-wall` that also includes a Checkout Button stacked underneath for subscribing immediately? Or somehow replace the `#register_modal` button with a Registration block. If we're already planning to add the same verification flow to the Register modal flow, then you can ignore this feedback.

dkoo (approved):
As discussed, we'll handle creating new patterns in a separate PR

miguelpeixe:
Thanks, @dkoo!

PR #4549, opened 2026-03-05
feat: make it easier to do "Block until consent given" setups in Complianz and improve blocking

claudiulodro:
### All Submissions:

Closes NEWS-1499.

This PR adds a new Privacy screen in Newspack > Settings if Complianz is active. This screen lets a user:
- Add a "block until consent given" setup to their site regardless of what geographic area they've set Complianz up for
- Optionally block ads until consent is given

1. Install and activate Complianz. Go through the setup wizard and select US and none of the state-specific options. This will give you a permissive setup with a default "optout" for trackers and ads (they will load by default) 

Screenshot 2026-03-19 at 2 40 08 PM

3. Have Site Kit, Google Ads, Meta Pixel, Twitter Pixel, and any other standard third-party trackers active on your site.
4. With network inspector open, visit the site in an incognito window. Observe that scripts are loaded from stats.wp.com and parse.ly.

Screenshot 2026-03-05 at 3 39 45 PM
Screenshot 2026-03-19 at 3 33 08 PM

5. Toggle on the feature to block third party trackers until after consent is given

Screenshot 2026-03-19 at 3 34 11 PM

6. Visit the site again. Observe that the trackers are not loaded in network inspector.

Screenshot 2026-03-19 at 3 34 50 PM

7. Also toggle on the feature to block ads until after consent is given. Visit the site again. Observe that trackers AND ads are not loaded in network inspector (and observe visually for ads).

Screenshot 2026-03-19 at 3 35 14 PM

8. Accept the notice. Observe that everything then loads in.

Screenshot 2026-03-19 at 3 35 22 PM

### Other information:


includes/plugins/class-complianz.php
```diff
@@ -0,0 +1,89 @@
...
+		// Format is 'domain' => 'category'.
+		// Category is one of 'statistics', 'marketing', or 'functional'.
+		// 'functional' doesn't make sense here though because those shouldn't be blocked.
+		$scripts_to_block = [
+			'googletagmanager.com' => 'statistics',
+			'doubleclick.net'      => 'marketing',
+			'parsely.com'          => 'statistics',
+		];
+
+		$script_pattern = '/<script[^>]*? src=[\'\"]([^>]*?)[\'\"][^>]*?>/is';
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/b76140252cefa0f2327d9916019c22b1bfe75e5f

includes/tracking/class-pixel.php
```diff
@@ -148,7 +148,7 @@ public function create_js_snippet( $payload ) {
 		if ( $this->is_amp() || ! $this->is_configured() ) {
 			return;
 		}
-		echo str_replace( '__PIXEL_ID__', $this->get_pixel_id(), $payload ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo apply_filters( 'newspack_pixel_script_markup', str_replace( '__PIXEL_ID__', $this->get_pixel_id(), $payload ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
leogermani: doc block please

claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/3e99d86821a7b8def7ff559c5b0dac36fbe6f529

includes/plugins/class-complianz.php
```diff
@@ -0,0 +1,147 @@
...
+	 * @param string $markup Pixel markup.
+	 * @return string Modified $markup.
+	 */
+	public static function pixel_handling_for_complianz( $markup ) {
+		$privacy_settings = Privacy_Section::get_settings();
+		if ( $privacy_settings['block_before_consent'] && self::is_complianz_with_cookie_blocker_active() ) {
+			$markup = str_ireplace( '<script', '<script type="text/plain" data-category="marketing"', $markup );
+		}
+		return $markup;
+	}
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/d7ee552133f52e33a73469366e695128998816c2

includes/plugins/class-complianz.php
```diff
@@ -0,0 +1,147 @@
...
+		$privacy_settings = Privacy_Section::get_settings();
+		if ( ! $privacy_settings['block_before_consent'] && ! $privacy_settings['block_ads_before_consent'] ) {
+			return $output;
+		}
+
+		$scripts_to_block = [];
+		if ( $privacy_settings['block_before_consent'] ) {
+			$scripts_to_block = array_merge( $scripts_to_block, $trackers );
+		}
+		if ( $privacy_settings['block_ads_before_consent'] ) {
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/d7ee552133f52e33a73469366e695128998816c2

includes/wizards/newspack/class-privacy-section.php
```diff
@@ -0,0 +1,99 @@
...
+	 * @return \WP_REST_Response
+	 */
+	public function api_update( $request ) {
+		if ( isset( $request['block_ads_before_consent'] ) ) {
+			update_option( self::OPTION_PREFIX . 'block_ads_before_consent', (bool) $request['block_ads_before_consent'] );
+		}
+		if ( isset( $request['block_before_consent'] ) ) {
+			update_option( self::OPTION_PREFIX . 'block_before_consent', (bool) $request['block_before_consent'] );
+		}
+		return rest_ensure_response( self::get_settings() );
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/d7ee552133f52e33a73469366e695128998816c2

src/wizards/newspack/views/settings/privacy/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+					) }
+				/>
+				{ data.block_before_consent && (
+					<ActionCard
+						isMedium
+						disabled={ isFetching }
+						toggleChecked={ data.block_ads_before_consent }
+						title={ __( 'Block ad scripts before consent', 'newspack-plugin' ) }
+						toggleOnChange={ ( block_ads_before_consent: boolean ) => setData( { ...data, block_ads_before_consent } ) }
+						description={ __(
```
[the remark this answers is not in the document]
claudiulodro: I'm ok with this since it's handled by the other changes in https://github.com/Automattic/newspack-plugin/pull/4549/commits/d7ee552133f52e33a73469366e695128998816c2. I don't necessarily want to blow out the ad option every time someone toggles this off

claudiulodro:
Updated title and description to better reflect changes to this feature since initial PR.

leogermani (approved):
pre approving but there are some conflicts to resolve. 

Works fine for me.

One thing I miss in the PR, and I think we should add to the class docs, or a README or something is the "why". Why and what are we doing here?

* Are we extending complianz to also block Newspack specific features that wouldn't be blocked otherwise, like Ads and pixel trackers?
* Why do we need an option to force a specific behavior of Complianz? Or the behavior is not quite what we want?

includes/plugins/class-complianz.php
```diff
@@ -0,0 +1,161 @@
...
+ * Main class.
+ */
+class Complianz {
+	/**
+	 * Initialize hooks and filters.
+	 */
+	public static function init() {
+		add_filter( 'cmplz_cookie_blocker_output', [ __CLASS__, 'extra_third_party_script_blocking' ] );
+		add_filter( 'newspack_pixel_script_markup', [ __CLASS__, 'pixel_handling_for_complianz' ] );
+		add_filter( 'cmplz_option_enable_cookie_blocker', [ __CLASS__, 'block_before_consent' ] );
```
adekbadek, line 30: **Suggestion:** The `newspack_pixel_script_markup` filter fires on every pixel render even when Complianz is not installed, triggering unnecessary `get_option()` calls inside the callback chain. The `cmplz_*` filters are harmless without Complianz (they never fire), but this one always does.

Consider guarding it:
```php
if ( self::is_complianz_active() ) {
    add_filter( 'newspack_pixel_script_markup', [ __CLASS__, 'pixel_handling_for_complianz' ] );
}
```
Or add an early return at the top of `pixel_handling_for_complianz`:
```php
if ( ! self::is_complianz_active() ) {
    return $markup;
}
```

includes/plugins/class-complianz.php
```diff
@@ -0,0 +1,161 @@
...
+
+	/**
+	 * In Cookie Blocker mode, also block pixels until consent is given.
+	 *
+	 * @param string $markup Pixel markup.
+	 * @return string Modified $markup.
+	 */
+	public static function pixel_handling_for_complianz( $markup ) {
+		if ( self::is_complianz_with_cookie_blocker_active() && self::should_block_trackers_before_consent() ) {
+			$markup = str_ireplace( '<script', '<script type="text/plain" data-category="marketing"', $markup );
```
adekbadek, line 123: **Suggestion:** `str_ireplace( '<script', ... )` replaces *every* occurrence of `<script` in the payload. Current pixel payloads (Meta, Twitter) have only one `<script` tag so this works, but if a pixel ever contains multiple tags (e.g. an inline fallback), all would get `type="text/plain"` injected.

A safer approach would limit to the first occurrence:
```php
$markup = preg_replace( '/<script\b/i', '<script type="text/plain" data-category="marketing"', $markup, 1 );
```

includes/wizards/newspack/class-privacy-section.php
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * API: update privacy settings.
+	 *
+	 * @param \WP_REST_Request $request Full details about the request.
+	 * @return \WP_REST_Response
+	 */
+	public function api_update( $request ) {
+		if ( isset( $request['block_ads_before_consent'] ) ) {
+			update_option( self::OPTION_PREFIX . 'block_ads_before_consent', (bool) $request['block_ads_before_consent'] );
```
adekbadek: **Suggestion:** The API allows setting `block_ads_before_consent` to `true` even when `block_before_consent` is `false` (the UI hides the toggle, but a direct API call could do it). The runtime logic in `should_block_ads_before_consent()` checks both, so it's not a functional bug, but consider resetting the child when the parent is turned off:
```php
if ( isset( $request['block_before_consent'] ) ) {
    update_option( self::OPTION_PREFIX . 'block_before_consent', (bool) $request['block_before_consent'] );
    if ( ! (bool) $request['block_before_consent'] ) {
        update_option( self::OPTION_PREFIX . 'block_ads_before_consent', false );
    }
}
```

tests/unit-tests/plugins/complianz.php
```diff
@@ -0,0 +1,209 @@
...
+	 * When block_before_consent is on AND Complianz cookie blocker is active,
+	 * pixel <script> tags get type="text/plain" data-category="marketing".
+	 */
+	public function test_pixel_blocked_when_settings_and_complianz_active() {
+		update_option( Privacy_Section::OPTION_PREFIX . 'block_before_consent', true );
+
+		// Simulate Complianz being active with cookie blocker enabled.
+		if ( ! function_exists( 'cmplz_can_run_cookie_blocker' ) ) {
+			// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound 
+			function cmplz_can_run_cookie_blocker() { // phpcs:ignore Squiz.Commenting.FunctionComment.Missing
```
adekbadek: **Suggestion:** This `cmplz_can_run_cookie_blocker` mock is defined inside a `function_exists` guard, but once defined it persists for the entire PHPUnit process. If test execution order changes and `test_pixel_blocked_when_settings_and_complianz_active` runs before `test_pixel_unchanged_when_complianz_cookie_blocker_not_active`, the latter would fail because the function would already exist and return `true`.

Consider `@runInSeparateProcess` on this test, or restructuring the mock to use a global flag that controls the return value.

src/wizards/newspack/views/settings/privacy/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+function Privacy() {
+	const { wizardApiFetch, isFetching } = useWizardApiFetch( 'newspack-settings/privacy' );
+
+	const [ data, setData ] = useState< PrivacyData >( {
+		block_ads_before_consent: false,
+		block_before_consent: false,
+	} );
+
+	useEffect( get, [] );
```
adekbadek: Nit: `useEffect( get, [] )` works due to function hoisting but reads a bit unusually. `useEffect( () => get(), [] )` is more idiomatic.

adekbadek (commented):
Tested locally with Complianz 7.4.5 – UI works as expected, settings persist, toggles interact correctly. Verified all four Complianz hooks (`cmplz_cookie_blocker_output`, `cmplz_option_enable_cookie_blocker`, `cmplz_consenttype`, `cmplz_can_run_cookie_blocker`) against the Complianz source – all are used correctly with the right arguments, return values, and data attribute conventions (`data-cmplz-src`, `type="text/plain"`, `data-category`). 

A few suggestions inline.

includes/plugins/class-complianz.php
claudiulodro: In that situation we would want all of them to get replaced, no? It would defeat the purpose if the main pixel is paused until after consent is given but the fallback fires instantly

tests/unit-tests/plugins/complianz.php
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/42123e8d0e0681840be24b242953e3d03f38b18d

includes/plugins/class-complianz.php
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/ee012ee6d72654b9c89fd13ffd935493bc5f7b7b

includes/wizards/newspack/class-privacy-section.php
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/258154075c56f6f624cecbecd6eb6ac670ce4fc7

src/wizards/newspack/views/settings/privacy/index.tsx
claudiulodro: https://github.com/Automattic/newspack-plugin/pull/4549/commits/258154075c56f6f624cecbecd6eb6ac670ce4fc7

includes/plugins/class-complianz.php
claudiulodro: "Complianz::init() is called at file-include time (the bare Complianz::init(); at the bottom of the file). If Newspack's includes() runs before Complianz has defined cmplz_can_run_cookie_blocker, the   
  is_complianz_active() check returns false and the filter is never registered — even though Complianz is installed and active.
                                                                                                                                                                                                           
  The other three cmplz_* filters are safe because they fire lazily (only when Complianz processes output, by which time all plugins are loaded). But this one is registered conditionally at load time.   
  
  In practice the directory name complianz-gdpr sorts before newspack-plugin so this won't bite most installs, but it's a fragile assumption. The simplest fix is to unconditionally register the filter   
  and move the is_complianz_active() check inside the callback — consistent with how the other three filters work."

Note for myself

claudiulodro: Looking at it further, I don't think it does trigger "unnecessary get_option() calls inside the callback chain" if Complianz is not active. `self::is_complianz_with_cookie_blocker_active()` first checks to see if Complianz is active.

I've reverted in https://github.com/Automattic/newspack-plugin/pull/4549/commits/6768b3783fb0fd8cb89f25383e21ec8fa7876bc0 to avoid the race condition it introduced

claudiulodro:
I've addressed all feedback as well as some new feedback my agents raised. Thanks for reviewing it, Leo and Adam!

includes/plugins/class-complianz.php
adekbadek: Ah yeah, we can assume all are pixel tags then.

adekbadek (approved):
👌

PR #4553, opened 2026-03-07
feat(content-gating): show group subscriptions to members in My Account

dkoo:
### All Submissions:

Allows group subscription members to see a limited read-only view of subscriptions they're a member of. Closes NPPD-1340.

1. As a group manager or admin, invite another user as a member to a group subscription.
2. Log into My Account as the member user.
3. Confirm that the member user can see the group subscription in the Subscriptions list view with a "Group" label and no payment method info (only if they have other subscriptions already—if not, there will just be a "Subscription" link in the menu that links to the subscription).

Screenshot 2026-03-06 at 5 32 56 PM

4. View the subscription. Confirm that the subscription has a "Group" label in the header, and that you can see some limited details about the subscription, including the subscription owner, but not anything about payment method, related orders, or any actions.

Screenshot 2026-03-06 at 5 34 16 PM

### Other information:


includes/plugins/woocommerce/my-account/templates/v1/subscription-details.php
```diff
@@ -69,21 +75,37 @@
...
 			</tr>
 		<?php endif; ?>
-		<?php do_action( 'wcs_subscription_details_table_before_payment_method', $subscription ); ?>
+		<?php
+		if ( $is_group_member_subscription ) :
+			$owner = get_user_by( 'id', $subscription->get_user_id() );
+			?>
+		<tr>
+			<td><?php esc_html_e( 'Subscription owner', 'newspack-plugin' ); ?></td>
+			<td>
```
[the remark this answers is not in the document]
dkoo: Added in b5c86e865

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
```diff
@@ -243,6 +257,67 @@ public static function handle_cancel_invite() {
...
+	 */
+	public static function grant_group_member_view_order_cap( $caps, $cap, $user_id, $args ) {
+		if ( 'view_order' !== $cap || ! function_exists( 'is_account_page' ) || ! \is_account_page() ) {
+			return $caps;
+		}
+		$order_id     = isset( $args[0] ) ? absint( $args[0] ) : 0;
+		$subscription = WooCommerce_Subscriptions::sanitize_subscription( $order_id );
+		if ( ! $subscription ) {
+			return $caps;
+		}
```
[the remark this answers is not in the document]
dkoo: Fixed in ca88ed634

thomasguillot:
Is it possible to move "You are a member of this group subscription. It is managed by the subscription owner above." to underneath the product name. And use a `notice` (default). And just "You are a member of this group subscription. It is managed by {NAME}."

dkoo:
@thomasguillot 21ad73d30 moves that message to a Notice element just below the header (and cleaned up the display of the owner name + email link too):

Screenshot 2026-03-10 at 10 45 49 AM

thomasguillot:
Thank you @dkoo

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
```diff
@@ -139,13 +143,23 @@ public static function render_group_subscription_members_template() {
...
+		if ( ! function_exists( 'is_account_page' ) || ! \is_account_page() || ! Group_Subscription::is_group_subscription( $subscription ) ) {
 			return $actions;
 		}
-		$actions['manage_members'] = [
-			'url'  => self::get_manage_members_url( $subscription ),
-			'name' => __( 'Manage members', 'newspack-plugin' ),
-		];
+
+		// Non-manager group members get a view-only experience: no actions.
+		if ( $subscription->get_customer_id() !== $user_id && Group_Subscription::user_is_member( $user_id, $subscription ) ) {
```
miguelpeixe: I think we should use `Group_Subscription::user_is_manager()` to follow the same pattern and support the `newspack_group_subscription_managers` filter.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
```diff
@@ -139,13 +143,23 @@ public static function render_group_subscription_members_template() {
...
+			return [];
+		}
+
+		// Managers (subscription owners) get a "Manage members" action.
+		if ( $subscription->get_customer_id() === $user_id && Group_Subscription::user_is_manager( $user_id, $subscription ) ) {
```
miguelpeixe: Why enforce both checks here? I thought `user_is_manager()` should suffice.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
```diff
@@ -255,6 +269,67 @@ public static function handle_cancel_invite() {
...
+	public static function grant_group_member_view_order_cap( $caps, $cap, $user_id, $args ) {
+		if ( 'view_order' !== $cap || ! function_exists( 'is_account_page' ) || ! \is_account_page() ) {
+			return $caps;
+		}
+		$order_id     = isset( $args[0] ) ? absint( $args[0] ) : 0;
+		$subscription = WooCommerce_Subscriptions::sanitize_subscription( $order_id );
+		if ( ! $subscription || $subscription->has_status( 'trash' ) ) {
+			return $caps;
+		}
+		if ( Group_Subscription::user_is_member( $user_id, $subscription ) ) {
```
miguelpeixe, line 327: Because `user_is_member` returns `true` for managers as well, should we be concerned about this downgrading their access?

includes/plugins/woocommerce/my-account/templates/v1/my-subscriptions.php
```diff
@@ -30,7 +32,13 @@
 	</thead>
 
 	<tbody>
-		<?php foreach ( $subscriptions as $subscription_id => $subscription ) : ?>
+		<?php
+		foreach ( $subscriptions as $subscription_id => $subscription ) :
+			$is_group_member_subscription = class_exists( 'Newspack\\Group_Subscription' )
```
miguelpeixe: NIT, but I don't think we need a `class_exists()` check here.

miguelpeixe (changes requested):
Works well! Left some comments inline

includes/plugins/woocommerce/my-account/templates/v1/my-subscriptions.php
dkoo: Removed all of these checks in 3c88b5968

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
dkoo: Since we also have `user_is_manager`, I was starting to get confused about the use of `user_is_member`, so 799f0700e changes the latter method to return `true` only for members and not for managers. I think this makes the various checks for member vs. manager role throughout My Account templates much simpler, and sidesteps the concern about accidentally downgrading capabilities.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
dkoo: True, removed in 799f0700e

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
dkoo: Since 799f0700e simplifies the `user_is_member` method to return true for members only (and not managers), this check can now be much simpler as `if ( Group_Subscription::user_is_member( $user_id, $subscription ) )`.

dkoo:
@miguelpeixe thanks for the feedback! I've pushed some changes that should simplify the logic a bit.

miguelpeixe (approved):
Thank you for the revisions! I feel responsible for the `user_is_member()` mess because I believe it was my suggestion to have it include managers, sorry about that 😅 

Tested well and looks a lot cleaner with the refactored `user_is_member()`.

Non-blocking question: Should we be concerned about members having access to the monetary values?

dkoo:
Thanks for the re-test, @miguelpeixe!

> Thank you for the revisions! I feel responsible for the `user_is_member()` mess because I believe it was my suggestion to have it include managers, sorry about that 😅

No blame at all, it made sense at the time since we were just building out the abstract API without a clear sense yet of exactly how we'd be using it. I think now that we have both `user_is_manager` and `user_is_member`, that lets us be more precise with our checks and intentionally include or not include both types of users.

> Non-blocking question: Should we be concerned about members having access to the monetary values?

I'm not sure about this...I think we should kick this question up the chain, and if we need to hide prices for group members, handle it in a separate PR.

PR #4565, opened 2026-03-13
Feat/contact metadata refactor

leogermani:
### All Submissions:

This PR creates the abstraction to implement a new metadata schema, and applies to abstraction to the legacy schema.

Next step here is to actually begin implementing the new schema.

This is a regression test of the old schema. Sync should work exactly as it works now.

1. Set mailchimp as your ESP
2. Enable sync
3. Register as a reader and confirm it syncs, confirm the mailchimp status and status_if_new fields are synced.

Now perform a bunch of action with this reader in a way that will trigger all these events. Check that

1. Sync takes the fields that you selected in the options
2. Sync has all these contexts described below.

* User log in - context: "RAS Reader login"
* Any completed order (except renewals) - buy something. context "RAS Order completed"
* Change of status in subscription - manually change the status or trigger a renewal, you will see 
Subscription renewal (should schedule a sync for 2 minutes later). You will see first a context like "RAS Woo  Subscription updated. Status changed from active to on-hold; RAS Woo Subscription updated. Status changed from on-hold to active" and then after 2 minutes a "RAS Woo subscription 2921 for novolemedjdi@leo.com renewed."
* New network reader (register in another site in the network and wait for the reader to be added to the site) - context: "RAS Newspack Network: User propagated from another site in the network. Propagated from http://super.local to https://leogermani.sa.ngrok.io."
* Reader deleted - it should sync and delete from the ESP in case that option is selected.
* Change the list of selected newsletters in my account (syncs only the newsletters_selection field) - Note. I didn't migrate it to the new way of doing things because this trigger is likely to be deprecated. But it should still work. - context: "Updating newsletter_selection field after a change in the subscription lists."

### Other information:


includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -137,13 +137,12 @@ public static function sync( $contact, $context = '', $existing_contact = null )
...
 		$contact = \apply_filters( 'newspack_esp_sync_contact', $contact, $context );
-		$contact = Sync\Metadata::normalize_contact_data( $contact );
-
 		$integrations = Integrations::get_active_integrations();
 		$errors       = [];
 
 		foreach ( $integrations as $integration_id => $integration ) {
-			$result = $integration->push_contact_data( $contact, $context, $existing_contact );
+			$integration_contact = $integration->prepare_contact( $contact );
+			$result              = $integration->push_contact_data( $integration_contact, $context, $existing_contact );
```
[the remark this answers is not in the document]
leogermani: normalize_contact_data method is called from within the Legacy\Basic Contact Metadata class

includes/reader-activation/integrations/class-integration.php
```diff
@@ -416,6 +416,40 @@ public function get_metadata_prefix() {
...
+		$keys_map       = Sync\Metadata::get_keys();
+		$prepared       = [];
+
+		foreach ( $contact['metadata'] as $raw_key => $value ) {
+			if ( isset( $keys_map[ $raw_key ] ) && in_array( $keys_map[ $raw_key ], $enabled_fields, true ) ) {
+				$prepared[ $prefix . $keys_map[ $raw_key ] ] = $value;
+			}
+		}
+
+		$contact['metadata'] = $prepared;
```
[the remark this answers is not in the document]
leogermani: additional values are still being able to be added via the `newspack_esp_sync_contact`. The mailchimp status field is still working.

in https://github.com/Automattic/newspack-plugin/pull/4565/commits/0ed59c9d7c982b74053dbd1631e7758e21a6810f I fixed the hook to add more meta keys via filters.

includes/reader-activation/sync/class-metadata.php
```diff
@@ -59,11 +93,7 @@ protected static function get_version() {
 	 * @return array List of fields.
 	 */
 	public static function get_keys() {
-		if ( 'legacy' === self::get_version() ) {
-			return Legacy_Metadata::get_keys();
-		}
-
-		return self::get_all_fields();
+		return self::get_all_fields( true );
```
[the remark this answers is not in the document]
leogermani: done in https://github.com/Automattic/newspack-plugin/commit/0ed59c9d7c982b74053dbd1631e7758e21a6810f

includes/reader-activation/sync/class-metadata.php
```diff
@@ -35,12 +35,46 @@ class Metadata {
...
+				'Legacy_Payment',
+			];
+		} else {
+			$classes = [
+				'Profile',
+				'Subscription',
+				'Donation',
+				'Content_Gate',
+				'Financial',
+			];
```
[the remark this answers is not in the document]
leogermani: these are still going to be implemented

includes/reader-activation/sync/class-contact-sync.php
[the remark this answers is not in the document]
leogermani: fixed in https://github.com/Automattic/newspack-plugin/pull/4565/commits/2942fb93d686ba0df0f57a92c4af3949da4f0bbe

chickenn00dle:
Checklist for my reference:

### Registration & Login
  - [x] Register a new reader via the RAS modal. Verify contact is created in ESP with correct status fields, basic metadata (Account, Registration Date), and empty payment fields. Context: "RAS Reader registration"
  - [x] Log in as a reader with no purchase history. Verify no sync is triggered.
  - [x] Log in as a reader with purchase history. Verify sync triggers with all selected metadata fields. Context: "RAS Reader login"

### Orders & Donations
  - [x] Complete a one-time donation. Verify payment metadata is populated (last payment amount/date, membership status, payment page, UTMs, total_paid). Context: "RAS Order completed"
  - [x] Complete a non-donation product purchase. Verify subscription/product metadata reflects the purchase. Context: "RAS Order completed"

### Subscriptions
  - [x] Purchase a recurring donation (monthly or yearly). Verify subscription metadata: start date, billing cycle, recurring amount, next payment date, product name. Context: "RAS Order completed"
  - [x] Trigger a subscription renewal. Verify two syncs: first with status change context ("Status changed from active to on-hold; ...on-hold to active"), then after ~2 minutes with renewal context ("RAS Woo subscription {ID} for {email} renewed."). Verify renewal does NOT
  trigger "RAS Order completed".
  - [x] Manually change a subscription status (e.g., active → on-hold or cancelled). Verify sync triggers with correct status change context and updated membership status metadata.
  - [x] Cancel a subscription. Verify membership status updates to cancelled/expired equivalent and cancellation reason is synced if provided.

### Network (if multi-site)
  - [x] Register a reader on another site in the Newspack Network. Verify sync on the current site with context: "RAS Newspack Network: User propagated from another site in the network. Propagated from {source} to {target}."

### Reader Deletion
  - [x] With "delete from ESP" enabled: delete a reader and verify the contact is removed from the ESP.
  - [x] With "delete from ESP" disabled: delete a reader and verify the contact remains in the ESP.

### Newsletter Selection
  - [ ] Change newsletter list selections in My Account. Verify only the newsletter_selection field is synced. Context: "Updating newsletter_selection field after a change in the subscription lists."

### Field Selection & Prefix
  - [x] Deselect some metadata fields in sync settings, trigger a sync, and verify only selected fields appear in ESP contact metadata.
  - [x] Verify all synced fields use the configured prefix (default NP_), with no double-prefixing (e.g., no NP_NP_Account).

includes/data-events/connectors/class-contact-sync-connector.php
```diff
@@ -158,18 +152,18 @@ public static function order_completed( $timestamp, $data, $client_id ) {
 	 * @param int   $client_id ID of the client that triggered the event.
 	 */
 	public static function subscription_updated( $timestamp, $data, $client_id ) {
-		if ( empty( $data['status_before'] ) || empty( $data['status_after'] ) || empty( $data['subscription_id'] ) ) {
+		if ( empty( $data['status_before'] ) || empty( $data['status_after'] ) || empty( $data['subscription_id'] ) || ! function_exists( 'wc_get_order' ) ) {
 			return;
 		}
 
-		$contact = Sync_WooCommerce::get_contact_from_order( $data['subscription_id'] );
+		$order = \wc_get_order( $data['subscription_id'] );
```
chickenn00dle: Should this method be using `wcs_get_subscription`?

includes/reader-activation/integrations/class-integration.php
```diff
@@ -416,6 +416,50 @@ public function get_metadata_prefix() {
...
+		}
+
+		$enabled_fields = $this->get_enabled_outgoing_fields();
+		$prefix         = $this->get_metadata_prefix();
+		$keys_map       = Sync\Metadata::get_keys();
+		$prepared       = [];
+
+		foreach ( $contact['metadata'] as $key => $value ) {
+			// If the key is already prefixed, keep it as-is if its field is enabled.
+			if ( ! empty( $prefix ) && 0 === strpos( $key, $prefix ) ) {
```
chickenn00dle: Nit: I don't think `$prefix` can be an empty string.

includes/reader-activation/sync/contact-metadata/class-donation.php
```diff
@@ -0,0 +1,58 @@
...
+	public static function get_fields() {
+		return [
+			'Donor_Status'                   => 'Donor_Status',
+			'Active_Donation_Count'          => 'Active_Donation_Count',
+			'Current_Donation_Start_Date'    => 'Current_Donation_Start_Date',
+			'Current_Donation_End_Date'      => 'Current_Donation_End_Date',
+			'Current_Donation_Cycle'         => 'Current_Donation_Cycle',
+			'Current_Recurring_Donation'     => 'Current_Recurring_Donation',
+			'Next_Donation_Date'             => 'Next_Donation_Date',
+			'_Current_Donation_Product_Name' => '_Current_Donation_Product_Name',
```
chickenn00dle: What is the reason for the `_` here?

I see there is a double underscore in the schema sheet, but have we confirmed this is intentional?

includes/reader-activation/sync/class-contact-metadata.php
```diff
@@ -0,0 +1,121 @@
...
+		if ( $user_customer_or_order instanceof \WC_Order ) {
+			$this->order = $user_customer_or_order;
+			$user_id     = $this->order->get_customer_id();
+		} elseif ( $user_customer_or_order instanceof \WC_Customer ) {
+			$this->customer = $user_customer_or_order;
+			$user_id        = $this->customer->get_id();
+		} elseif ( $user_customer_or_order instanceof \WP_User ) {
+			$this->user = $user_customer_or_order;
+			$user_id    = $this->user->ID;
+		} else {
```
chickenn00dle, line 53: Nit: This feels a bit convoluted. Can this be simplified at all?

For example, can we move the order assignment to the specific child classes where its relevant.

It seems to me only customer and user need to be assigned here, and for both, you only need a user_id.

includes/reader-activation/sync/class-contact-metadata.php
```diff
@@ -0,0 +1,121 @@
...
+	 * @return string
+	 */
+	public function get_email() {
+		if ( $this->customer ) {
+			return $this->customer->get_email();
+		}
+		if ( $this->user ) {
+			return $this->user->user_email;
+		}
+		if ( $this->order ) {
```
chickenn00dle: This condition should never be reached right? If there is an order, there is a customer and WC_Customer will definitely exist for this to be set in the contructor.

includes/reader-activation/sync/class-contact-metadata.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * Get the full name for the contact from the WC_Customer billing name.
+	 *
+	 * @return string
+	 */
+	public function get_full_name() {
+		if ( $this->customer ) {
+			return trim( $this->customer->get_billing_first_name() . ' ' . $this->customer->get_billing_last_name() );
+		}
+		return '';
```
chickenn00dle, line 123: No fallback from `WP_User`?

chickenn00dle (changes requested):
Nice work @leogermani!

I tested this and pretty much everything is working as expected. There was just one issue with subscribing to a newsletter via My Account. For some reason, contact metadata is null when handling the `newsletter_subscribed` data event, which is triggering an error with `array_merge`. 

```
[18-Mar-2026 21:53:17 UTC] PHP Warning:  Undefined array key "metadata" in /newspack-repos/newspack-plugin/includes/data-events/connectors/class-contact-sync-connector.php on line 292
[18-Mar-2026 21:53:17 UTC] [5368] [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::log][ERROR]: array_merge(): Argument #1 must be of type array, null given
[18-Mar-2026 21:53:17 UTC] [5368] [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::log]: Scheduled retry 1/5 for handler on "newsletter_subscribed" in 30s. Error: array_merge(): Argument #1 must be of type array, null given
```

All subsequent retries fail with the same error.

I also left a handful of mostly non-blocking comments.

includes/reader-activation/sync/class-contact-metadata.php
chickenn00dle: I'm realizing now the order is probably accounting for anonymous readers. Still, it seems unnecessary to declare an order at the base Metadata class since it won't be relevant to all Metadata.

includes/reader-activation/sync/contact-metadata/class-donation.php
leogermani: typo

includes/reader-activation/sync/class-contact-metadata.php
leogermani: Just because I'm mimicking the current behavior we have [here](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/reader-activation/sync/class-woocommerce.php#L416-L418).

We should fallback to the user, yes, but I didn't want to introduce changes to the flow just yet.

includes/reader-activation/sync/class-contact-metadata.php
leogermani: That's a good call. I just wanted to make sure we could construct the class with any of those inputs, but I agree all the others can be fetched on demand. No need to initialize them all right away.

includes/data-events/connectors/class-contact-sync-connector.php
leogermani: It works just the same

chickenn00dle: That's fair. There is a difference. `wc_get_order` will work when passed an id that is for an order (and not a subscription), while `wcs_get_subscription` will not. But I guess in the context of our sync we can be sure we are always providing a valid subscription id here.

leogermani:
> Nice work @leogermani!
> 
> I tested this and pretty much everything is working as expected. There was just one issue with subscribing to a newsletter via My Account. For some reason, contact metadata is null when handling the `newsletter_subscribed` data event, which is triggering an error with `array_merge`.
> 
> ```
> [18-Mar-2026 21:53:17 UTC] PHP Warning:  Undefined array key "metadata" in /newspack-repos/newspack-plugin/includes/data-events/connectors/class-contact-sync-connector.php on line 292
> [18-Mar-2026 21:53:17 UTC] [5368] [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::log][ERROR]: array_merge(): Argument #1 must be of type array, null given
> [18-Mar-2026 21:53:17 UTC] [5368] [NEWSPACK-DATA-EVENTS][Newspack\Data_Events::log]: Scheduled retry 1/5 for handler on "newsletter_subscribed" in 30s. Error: array_merge(): Argument #1 must be of type array, null given
> ```
> 
> All subsequent retries fail with the same error.
> 
> I also left a handful of mostly non-blocking comments.

I looked into this and it doesnt seem to be related to anything we changed in this PR. 

This same callback is tied to two different events, while [newsletter_subscribed includes a $contact](https://github.com/Automattic/newspack-plugin/blob/a5333075242a84adbb8da8148f3ceba253ec483f/includes/data-events/listeners.php#L132), [newsletters_updated don't](https://github.com/Automattic/newspack-plugin/blob/a5333075242a84adbb8da8148f3ceba253ec483f/includes/data-events/listeners.php#L152-L157).

I'm happy to push a fix on a separate PR.

includes/reader-activation/sync/class-contact-metadata.php
leogermani: I'm looking at this again... and I think it looks fine.

The idea here is that you can instantiate this class from different objects (user id, customer, order, etc), and you will always get a user at the end.

So you see that the if only sets the object that was used to instantiate the class and the user. So we make sure that all classes will have a user.

> Still, it seems unnecessary to declare an order at the base Metadata class since it won't be relevant to all Metadata.

We are not setting an order when it's not necessary. We are only setting the order property when an order was used to instantiate the class. And this will happen when the trigger for the sync is something around the order, for example order_complete. Then we can just instantiate the class with the order to get the user with their metadata.

The if down below that sets the WC_Customer may be superfluous, but it doesn't hurt.. and we can revisit this when we actually implement the metadata fields and see how this looks in practice.

includes/reader-activation/sync/contact-metadata/class-donation.php
leogermani: https://github.com/Automattic/newspack-plugin/pull/4565/commits/636f12a051886533dfae01502af1acee31864c92 also removed the underscores from the human readable names

includes/data-events/connectors/class-contact-sync-connector.php
leogermani: but yes, it's better to use the more specific method: https://github.com/Automattic/newspack-plugin/pull/4565/commits/6b2acbaddfb18158336d97e7f9ffd21ed7d98286

For a moment I thought that this was there already and I was just moving code around.. but this is new.

includes/reader-activation/integrations/class-integration.php
leogermani: I don't see guardrails for that.. we should probably add it to the get_metadata_prefix... if what's saved is empty it should return the default...

includes/reader-activation/sync/class-contact-metadata.php
leogermani: yes. https://github.com/Automattic/newspack-plugin/pull/4565/commits/4413ddcb6be0cd41b96773585e391539bc5104df

includes/reader-activation/integrations/class-integration.php
leogermani: https://github.com/Automattic/newspack-plugin/pull/4565/commits/0a5727de79c7f84868ea42116ce98fb1c1358c78

leogermani:
@chickenn00dle thanks for the review and the thorough testing. This is ready for another round.

includes/reader-activation/sync/class-contact-metadata.php
chickenn00dle: > we can revisit this when we actually implement the metadata fields and see how this looks in practice.

My initial thought was the base metadata class should never even be concerned with the concept of an order, even if its set conditionally. It seems to me that the class can be instantiated with something like `$user_or_user_id` and be able to set customer or user accordingly while still providing the core email/name functionality. That said, I don't feel to strongly about this so this sounds good to me!

chickenn00dle (approved):
Thanks for addressing my feedback @leogermani!

I popped in to get this reviewed so you aren't blocked while I'm AFK.

> I'm happy to push a fix on a separate PR.

Sure. Or we can get a new linear tasked opened if it isn't a critical blocker.

Other than this, the changes look good. Gonna approve assuming you'll get the phpunit tests passing.

PR #4577, opened 2026-03-17
feat(access-control): premium newsletters UI

dkoo:
### All Submissions:

Implements admin UI for managing premium newsletters. This UI appears in its own separate admin page away from Access Control, as it's technically a different type of access control compared to restricting site content. However, on the back-end premium newsletter data and restriction logic are managed exactly as site content restrictions, with the same content gate posts.

This PR introduces a new meta boolean flag `is_newsletter` which can be applied to any gate post. If a gate post has this flag, it's treated as a premium newsletter gate and is not shown or managed via the UI in the Access Control wizard. Instead, it's shown and managed via the new Premium Newsletters wizard. Because of this, newsletter gate posts are also not subject to priority. All `get_gate()` methods fetch non-newsletter gates only unless explicitly passed an argument to fetch newsletter gates instead. Aside from these differences, premium newsletter gates are the same CPT, behave exactly the same way, and have the same data schema as non-newsletter gates.

Closes NPPD-1368.

1. Check out this branch. Make sure your test site is connected to an ESP and that you've enabled some subscription lists in **Newsletters > Settings**.
2. Publish some content gates via **Audience > Access control** and confirm that all still works and behaves as expected in this wizard UI.
3. Confirm there's a new **Newsletters > Premium newsletters** menu item. Visit this and confirm that it's properly highlighted and its parent menu is expanded while on the page:

Screenshot 2026-03-18 at 3 35 45 PM

4. Confirm that you see an onboarding screen similar to the empty state at the Access control wizard, but with messaging tailored to premium newsletters. Click "Restrict all lists" and confirm it navigates you to an edit screen with the "Restrict all lists" option preselected.
5. Click on the back button in the header and click on "Choose specific lists" instead. Confirm that it navigates you to the edit screen with this option preselected instead. Confirm that there's no Include/Exclude option for newsletters.

Screenshot 2026-03-18 at 3 36 31 PM

6. Click on the autocomplete field and confirm that it shows all* subscription lists as suggestions. Confirm that you can search for lists by name and select one or more.
7. Confirm that you can configure Registered and/or Paid Access rules as with regular content gates, but without any metering options. **Note:** Does it even make sense to offer Registered Access options here?
8. Set some access rules such as "active subscription" and save.
9. Confirm that you're navigated back to the list view, which should look the same as the Access Control list view but show only premium newsletter gates.

Screenshot 2026-03-18 at 3 36 46 PM

10. Click on the settings icon button next to the header and confirm that it opens a settings modal with a single option (so far), enabled by default. Note that the option itself doesn't do anything yet, that'll come in a future PR.

Screenshot 2026-03-18 at 3 37 22 PM

11. Toggle on/off the option and save, confirm that it works as expected.
12. Test editing the existing premium newsletter gate, adding a new one, activating/deactivating/deleting gates, etc.
13. Leave some premium newsletter gates active and navigate back to **Audience > Access control**.
14. Confirm that you don't see any of the premium newsletter gates in this page, and that all content gate functionality such as creating/editing/deleting gates and updating priorities still work as expected.

* The autocomplete field shows ALL available lists, not just enabled—this is so that premium newsletters can be configured while the list isn't yet enabled, so there's no potential gap period where a premium list is visible to readers before it's ready

### Other information:


includes/wizards/newsletters/class-premium-newsletters.php
```diff
@@ -0,0 +1,371 @@
...
+		);
+
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/priority',
+			[
+				'methods'             => 'POST',
+				'callback'            => [ $this, 'update_gate_priorities' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+				'args'                => [
```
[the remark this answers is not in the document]
dkoo: This route isn't needed for premium newsletters. fe8c1f1 removes it.

thomasguillot:
## Icons update

Pushed a commit that updates the `newspack-icons` package as part of this PR:

**Removed:**
- `lockDoor` — replaced by more specific icons (see below)

**Added:**
- `contentLocked` — a document with a padlock overlay, used as the header icon for the access control onboarding screen
- `emailPremium` — an email envelope with a currency/dollar badge, used as the header icon for the premium newsletters onboarding screen

**Usage changes:**
- The `ContentGatesOnboarding` header icon now uses `emailPremium` when `isNewsletter=true`, and `contentLocked` otherwise
- The "Restrict all" card in the content gate editor (`edit/index.tsx`) now uses the WordPress `envelope` icon when `isNewsletter=true`, and `postList` otherwise

**Version:** `newspack-icons` bumped from `1.0.7` → `1.0.8`

> ⚠️ Once this PR is merged, a new version of `newspack-icons` will need to be published to npm so that other repos consuming the package can pick up the new icons and the removal of `lockDoor`.

thomasguillot:
Screenshot 2026-03-19 at 17 15 54@2x

Is it possible to remove the toggle?

Screenshot 2026-03-19 at 17 16 34@2x

So we get more something like this.

Screenshot 2026-03-19 at 17 17 11@2x

Bonus point if we can have this "focused" by default so it doesn't look like just an empty input with no instructions

<del>Edit: see https://github.com/Automattic/newspack-plugin/pull/4577#issuecomment-4091942309</del>

Edit 2: let's go with this

thomasguillot:
<del>I'm so sorry @dkoo. Looking back at the screenshots I don't like the FormTokenField for this case.
Would it be possible to use toggles instead?</del>
Screenshot 2026-03-19 at 17 22 22@2x

Edit: we're sticking with FormTokenField

thomasguillot:
Something I've only just noticed. There's an extra separator added at the end of the list
Screenshot 2026-03-19 at 17 31 24@2x

thomasguillot:
Font/colours don't quite match Access control
Screenshot 2026-03-19 at 17 35 08@2x
Screenshot 2026-03-19 at 17 35 48@2x

thomasguillot:
Can we use a more/dropdown for the advanced settings?
Screenshot 2026-03-19 at 17 40 16@2x

dkoo:
> Looking back at the screenshots I don't like the FormTokenField for this case.
Would it be possible to use toggles instead?

@thomasguillot wouldn't this be a concern if a publisher has many available lists? I've seen some publishers with literally hundreds of Mailchimp groups and tags. I think the `FormTokenField` is the best UI component available for cases like this, right?

thomasguillot:
> wouldn't this be a concern if a publisher has many available lists? I've seen some publishers with literally hundreds of Mailchimp groups and tags. I think the FormTokenField is the best UI component available for cases like this, right?

Let's go ahead with the FormTokenField for now (and I will think/research of a potential alternative)

See https://github.com/Automattic/newspack-plugin/pull/4577#issuecomment-4091896556 for feedback on the FormTokenField

dkoo:
@thomasguillot this should be ready for another design review!

> Is it possible to remove the toggle?
> Bonus point if we can have this "focused" by default so it doesn't look like just an empty input with no instructions

e1f4170 removes the toggle for premium newsletters content rules only (since there's only one). It also adds a placeholder prop to the FormTokenField so it appears like this when empty:

Screenshot 2026-03-19 at 4 21 35 PM

> Font/colours don't quite match Access control
> Something I've only just noticed. There's an extra separator added at the end of the list

Fixed in e1f4170.

> Can we use a more/dropdown for the advanced settings?

Done in b5f9b0d:

Screenshot 2026-03-19 at 4 48 56 PM

thomasguillot:
Thank you for the changes @dkoo. I made 2 small tweaks:

* https://github.com/Automattic/newspack-plugin/pull/4577/commits/53f5be92477b18f369c4b01d1d29a043e78c7902 Removes the margin-top from the `FormTokenField`
* https://github.com/Automattic/newspack-plugin/pull/4577/commits/eb51ef75f14f0412e21251d9f81e32c320fb6a78 Changed the menu label to simply "Premium". I noticed it was over 2 lines when it was "Premium newsletters" (unselected)

src/wizards/newsletters/views/premium-newsletters/advanced-settings.tsx
```diff
@@ -0,0 +1,101 @@
...
+	return (
+		showModal && (
+			<Modal onClose={ closeModal } size="medium" title={ __( 'Advanced settings', 'newspack-plugin' ) } onRequestClose={ closeModal }>
+				<VStack>
+					<ToggleControl
+						label={ __( 'Auto signup', 'newspack-plugin' ) }
+						help={ __( 'Automatically sign up users when they meet access requirements for premium newsletters.', 'newspack-plugin' ) }
+						checked={ config?.auto_signup || true }
+						onChange={ value => setConfig( { ...config, auto_signup: value } ) }
+					/>
```
miguelpeixe, line 82: In my testing this toggle is stuck enabled.

src/wizards/newsletters/views/premium-newsletters/advanced-settings.tsx
```diff
@@ -0,0 +1,101 @@
...
+					updateWizardSettings( {
+						slug: PREMIUM_NEWSLETTERS_WIZARD_SLUG,
+						path: [ 'config' ],
+						value: _config,
+					} );
+					addNotice( {
+						message: __( 'Settings updated.', 'newspack-plugin' ),
+						type: 'success',
+						id: 'premium-newsletters-advanced-settings-updated',
+						actions: [ { label: __( 'Undo', 'newspack-plugin' ), onClick: () => updateConfig.current?.( oldConfig ) } ],
```
miguelpeixe, line 60: `oldConfig` was updated in line 37 with the latest config. The Undo action will not work as intended.

miguelpeixe (changes requested):
Looks good and tested mostly well, a few small issues commented inline.

A non-blocking suggestion, moving the API logic away from the wizard class is a good move, but the content gate main class is now 1700+ lines. Maybe we can introduce a `Content_Gate_API` class for this?

src/wizards/newsletters/views/premium-newsletters/advanced-settings.tsx
dkoo: Ah, you're right. The fallback clause was making it impossible to set the option to a falsy value. 593c1ca4d should fix this.

dkoo: 593c1ca4d also changes the option value saved to the options table to a boolean-style integer (`0` or `1`) instead of a true boolean. For some reason I was having trouble updating the nonexistent option to `false` as opposed to `0`.

src/wizards/newsletters/views/premium-newsletters/advanced-settings.tsx
dkoo: Fixed in 593c1ca4d!

dkoo:
@miguelpeixe thanks for the feedback!

> A non-blocking suggestion, moving the API logic away from the wizard class is a good move, but the content gate main class is now 1700+ lines. Maybe we can introduce a `Content_Gate_API` class for this?

Great idea! ab866fa25 creates two new subclasses for `Content_Rules` and `Content_Gate_API` methods and properties, reducing the `Content_Gate` class file by about 400 lines.

Also pushed fixes for your inline comments and responded inline.

miguelpeixe (changes requested):
Thanks, @dkoo!

I caught a few more things that I think might be affecting the main gate editor as well, apologies for not finding it sooner.

When switching from "specific lists" to "restrict all" then back I'd expect the selected lists to persist. I don't think switching those options should clear the input:

https://github.com/user-attachments/assets/b20b968a-bf44-4c2f-9e36-f26bc827bd7c

While testing, I had a network hiccup that rendered an error notice in the editor, but it was flickering, as if it was stuck in an infinite re-render loop. Unfortunately, I was not able to reproduce it again, but while trying to, I added `return new \WP_Error( 'test_error', 'Test error );` to line 365 of `includes/wizards/newsletters/class-premium-newsletters.php` and even though the API response returns the error on gate updates, nothing shows up in the UI. We should have error handling there.

dkoo:
> While testing, I had a network hiccup that rendered an error notice in the editor, but it was flickering, as if it was stuck in an infinite re-render loop. Unfortunately, I was not able to reproduce it again, but while trying to, I added `return new \WP_Error( 'test_error', 'Test error );` to line 365 of `includes/wizards/newsletters/class-premium-newsletters.php` and even though the API response returns the error on gate updates, nothing shows up in the UI. We should have error handling there.

@miguelpeixe good catches! I had thought the `Wizard` component would handle setting error messages at a higher component level but failed to thoroughly test that. I think something was getting tripped up in an infinite render loop between the wizard store's error data getting passed to local component state. When I `console.log( errorMessage )` in the `edit/index.tsx` component, I see it getting set and then immediately nulled out for some reason.

Instead of trying to read error messages from the wizard store and rendering these as `<Notice />` components inside each view component, cea7ea4 and fb4dec2 dispatch error messages to be rendered as `<Snackbar />` components as rendered by the higher-level Wizard component itself, sidestepping the infinite render loop. As a plus I think the visual presentation of errors is now more consistent with success messaging throughout all views:

Screenshot 2026-03-24 at 10 07 27 AM

Returning a `WP_Error` from any REST endpoint callback should now display that error in the same way in all Access Control and Premium Newsletter views.

dkoo:
Whoops, forgot to respond to this:

> When switching from "specific lists" to "restrict all" then back I'd expect the selected lists to persist. I don't think switching those options should clear the input

67fe930 should fix this. Switching between "Restrict all posts/lists" and "Choose specific content/lists" should no longer cause the saved tokens to be lost in state.

miguelpeixe (approved):
Thank you for the revisions!

PR #4580, opened 2026-03-18
feat(image-upload): simplify component to use basecontrol; remove info prop

thomasguillot:
### All Submissions:

This PR simplifies the `ImageUpload`. Instead of custom label and help text we are leveraging Core's `BaseControl`. This means visually it looks closer to Core's (especially if they end up making tweaks to the label or help text).

I removed the `info` prop. It wasn't being used and I'd rather we use the help text anyway.

I've also update the buttons to use the `tertiary` variant instead of `isLink`.

__Before:__

__After:__

1. Navigation to a page that has image uploads (e.g. Newspack > Settings > Theme and brand)
2. Check how it looks and works now
3. Switch to this branch and refresh page
4. Test if everything still work (upload / remove / replace)

### Other information:

### DIFF
--- a/packages/components/src/image-upload/index.js
+++ b/packages/components/src/image-upload/index.js
@@ -21,8 +21,6 @@ import './style.scss';
 import classnames from 'classnames';
 
 class ImageUpload extends Component {
-	static instanceCounter = 0;
-
 	/**
 	 * Constructor.
 	 */
@@ -31,9 +29,6 @@ class ImageUpload extends Component {
 		this.state = {
 			frame: false,
 		};
-
-		ImageUpload.instanceCounter += 1;
-		this.baseControlId = this.props.id || `newspack-image-upload-${ ImageUpload.instanceCounter }`;
 	}
 
 	/**
@@ -85,7 +80,8 @@ class ImageUpload extends Component {
 			{ 'newspack-image-upload__image--covering': isCovering }
 		);
 		return (
-			<BaseControl className={ classnames( 'newspack-image-upload', className ) } help={ help } id={ this.baseControlId } label={ label }>
+			<BaseControl __nextHasNoMarginBottom className={ classnames( 'newspack-image-upload', className ) } help={ help }>
+				{ label && <BaseControl.VisualLabel>{ label }</BaseControl.VisualLabel> }
 				<div className={ classes } style={ style }>
 					{ image?.url ? (
 						<>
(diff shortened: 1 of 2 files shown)


packages/components/src/image-upload/style.scss
```diff
@@ -3,40 +3,35 @@
...
+				background: wp-colors.$white !important;
+				border-radius: wp-vars.$radius-small !important;
+				padding: wp-vars.$grid-unit-05 wp-vars.$grid-unit-20 !important;
 
 				&:active,
 				&:focus,
 				&:hover {
-					background: white !important;
+					background: wp-colors.$white !important;
 				}
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 8f87f016. Changed `.newspack-button` to `.components-button` since the Newspack `Button` wraps `@wordpress/components` Button which renders with class `components-button`.

packages/components/src/image-upload/index.js
```diff
@@ -72,41 +78,34 @@ class ImageUpload extends Component {
...
 		);
 		return (
-			<div className={ classnames( 'newspack-image-upload', className ) }>
-				<div className="newspack-image-upload__header">
-					{ /* eslint-disable-next-line jsx-a11y/label-has-for, jsx-a11y/label-has-associated-control */ }
-					{ label && <label className="newspack-image-upload__label">{ label }</label> }
-					{ info && <InfoButton text={ info } /> }
-				</div>
-				<div className={ classes } style={ { ...style } }>
+			<BaseControl className={ classnames( 'newspack-image-upload', className ) } help={ help } id={ this.baseControlId } label={ label }>
```
rbcorrales: The `id` prop on `BaseControl` makes it render a `<label htmlFor={id}>`, but when an image is present, no child element has that id (it's only on the Upload button in the empty state). So the label points to nothing.

Image

vs:

Image

Since this wraps composite content (buttons + image preview) rather than a single form input, I think [`BaseControl.VisualLabel`](https://developer.wordpress.org/block-editor/reference-guides/components/base-control/#basecontrol-visuallabel) would be a better fit here. That way the label is visible but doesn't try to associate with a specific element via htmlFor.

rbcorrales: Also, this same line is missing `__nextHasNoMarginBottom` on `BaseControl`. Without that you'll get a deprecation warning, and the default bottom margin from BaseControl will stack with the existing `$grid-unit-40 0` margin from the SCSS.

This is what I see in the console:

Image

rbcorrales (approved):
It works as expected (and looks nicer), but I left a couple of comments.
I'm approving the PR, but please consider tackling those before merging.

packages/components/src/image-upload/index.js
thomasguillot: Addressed in 26b68ac7. Replaced `id` + `label` props on `BaseControl` with `BaseControl.VisualLabel` so the label is visible but no longer tries to associate with a specific element via `htmlFor`. Also cleaned up the now-unused `baseControlId` instance property.

thomasguillot: Addressed in 26b68ac7. Added `__nextHasNoMarginBottom` to `BaseControl`.

PR #4581, opened 2026-03-18
feat(color-picker): simplify component to use basecontrol

thomasguillot:
### All Submissions:

Similar to https://github.com/Automattic/newspack-plugin/pull/4580 we need to stop having "hard coded" labels and start using `BaseControl` for these.

Aslo, I added support for a help text.

1. Go somewhere we have a color picker (e.g. Newspack > Settings > Theme and brand)
2. Notice the label looks off compared with the other settings
3. Switch to this branch
4. Test if the color picker still works and notice now the labels are in caps

### Other information:

### DIFF
--- a/packages/components/src/color-picker/index.js
+++ b/packages/components/src/color-picker/index.js
@@ -1,7 +1,8 @@
 /**
  * WordPress dependencies.
  */
-import { ColorPicker as ColorPickerComponent } from '@wordpress/components';
+import { BaseControl, ColorPicker as ColorPickerComponent } from '@wordpress/components';
+import { useInstanceId } from '@wordpress/compose';
 import { useState, useRef } from '@wordpress/element';
 
 /**
@@ -26,26 +27,26 @@ const { InteractiveDiv } = utils;
  *
  * @param {Object}             props             - Component props.
  * @param {JSX.Element|string} props.label       - Label for the color picker.
+ * @param {JSX.Element|string} props.help        - Help text for the color picker.
  * @param {string}             [props.color]     - Default color.
  * @param {Function}           props.onChange    - Function to call when the color changes.
  * @param {string}             [props.className] - Additional class name.
  * @return {JSX.Element} ColorPicker component.
  */
-const ColorPicker = ( { label, color = '#fff', onChange, className } ) => {
+const ColorPicker = ( { label, help, color = '#ffffff', onChange, className } ) => {
 	const [ isExpanded, setIsExpanded ] = useState( false );
 	const ref = useRef();
+	const id = useInstanceId( ColorPicker, 'newspack-color-picker' );
 	const colordColor = colord( color );
 	hooks.useOnClickOutside( ref, () => setIsExpanded( false ) );
 	return (
-		<div className={ classnames( 'newspack-color-picker', className ) }>
-			<div className="newspack-color-picker__label">{ label }</div>
-
+		<BaseControl className={ classnames( 'newspack-color-picker', className ) } id={ id } label={ label } help={ help }>
 			<InteractiveDiv
 				className={ 'newspack-color-picker__expander' }
 				onClick={ () => setIsExpanded( ! isExpanded ) }
 				style={ {
 					backgroundColor: color,
-					color: colordColor.contrast() > colordColor.contrast( '#000' ) ? '#fff' : '#000',
+					color: colordColor.contrast() > colordColor.contrast( '#000000' ) ? '#ffffff' : '#000000',
 				} }
 			>
 				{ color }
@@ -54,7 +55,7 @@ const ColorPicker = ( { label, color = '#fff', onChange, className } ) => {
 			<div className="newspack-color-picker__main" ref={ ref }>
 				{ isExpanded && <ColorPickerComponent color={ color } onChange={ hex => onChange( hex ) } enableAlpha={ false } /> }
 			</div>
-		</div>
+		</BaseControl>
 	);
 };
 
--- a/packages/components/src/color-picker/index.js
+++ b/packages/components/src/color-picker/index.js
@@ -42,6 +42,8 @@ const ColorPicker = ( { label, help, color = '#ffffff', onChange, className } ) 
 	return (
 		<BaseControl className={ classnames( 'newspack-color-picker', className ) } id={ id } label={ label } help={ help }>
 			<InteractiveDiv
+				id={ id }
+				aria-expanded={ isExpanded }
 				className={ 'newspack-color-picker__expander' }
 				onClick={ () => setIsExpanded( ! isExpanded ) }
 				style={ {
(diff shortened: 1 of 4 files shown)


packages/components/src/color-picker/index.js
```diff
@@ -26,26 +27,26 @@ const { InteractiveDiv } = utils;
...
 	const colordColor = colord( color );
 	hooks.useOnClickOutside( ref, () => setIsExpanded( false ) );
 	return (
-		<div className={ classnames( 'newspack-color-picker', className ) }>
-			<div className="newspack-color-picker__label">{ label }</div>
-
+		<BaseControl className={ classnames( 'newspack-color-picker', className ) } id={ id } label={ label } help={ help }>
 			<InteractiveDiv
 				className={ 'newspack-color-picker__expander' }
 				onClick={ () => setIsExpanded( ! isExpanded ) }
```
thomasguillot, line 52: **Blocker — broken label association (WCAG 1.3.1 / 4.1.2)**

The `id` from `useInstanceId` is passed to `BaseControl`, which renders `<label for="{id}">`. But no child element has that `id`, so the label references nothing — screen readers can't associate the label with the control, and clicking the label won't focus anything.

Fix: pass `id` to `InteractiveDiv` and add `aria-expanded` so assistive technology knows the picker is open/closed.

```suggestion
		<InteractiveDiv
			id={ id }
			aria-expanded={ isExpanded }
			className={ 'newspack-color-picker__expander' }
			onClick={ () => setIsExpanded( ! isExpanded ) }
```

packages/components/src/color-picker/index.js
```diff
@@ -54,7 +55,7 @@ const ColorPicker = ( { label, color = '#fff', onChange, className } ) => {
 			<div className="newspack-color-picker__main" ref={ ref }>
 				{ isExpanded && <ColorPickerComponent color={ color } onChange={ hex => onChange( hex ) } enableAlpha={ false } /> }
```
thomasguillot: Nit: the arrow wrapper passes `hex` straight through with no transformation — can be simplified.

```suggestion
			{ isExpanded && <ColorPickerComponent color={ color } onChange={ onChange } enableAlpha={ false } /> }
```

packages/components/src/color-picker/style.scss
```diff
@@ -2,31 +2,29 @@
...
+@use "~@wordpress/base-styles/variables" as wp-vars;
+
 .newspack-color-picker {
 	position: relative;
-
-	&__label {
-		margin-bottom: 8px;
-	}
-
 	&__expander {
```
thomasguillot, line 10: Heads up: the border opacity changed from `0.2` (original `rgb(0 0 0 / 20%)`) to `0.54` (`rgba($black, 0.54)`). That's a significant visual jump — the swatch border will be noticeably darker. Is this intentional?

packages/components/src/color-picker/style.scss
```diff
@@ -2,31 +2,29 @@
...
 		}
 	}
 
 	&__main {
-		background: white;
-		box-shadow: 0 0 8px 4px rgba(black, 0.08);
-		margin: 12px 0;
+		background: wp-colors.$white;
+		box-shadow: wp-vars.$elevation-small;
+		margin: wp-vars.$grid-unit-10 0 0;
```
thomasguillot, line 27: Nit: `padding` changed from `12px` → `$grid-unit-10` (which resolves to `8px`), and the `__main` margin changed from `12px 0` (top + bottom) → `$grid-unit-10 0 0` (top only, smaller). Both may be intentional alignment with the WP grid, but neither is mentioned in the PR description — worth a quick visual check.

thomasguillot (commented):
## Code review

Good refactor — using `BaseControl` + `useInstanceId` is the right pattern and the SCSS token migration is a nice improvement. One accessibility blocker to fix before merge, plus a few minor notes.

**CI**: All checks pass ✓
**Design system**: No violations — the `packages/components/src/color-picker/` directory is excluded from design review (raw hex by design).

packages/components/src/color-picker/index.js
thomasguillot: Addressed in 210992ed.

packages/components/src/color-picker/index.js
thomasguillot: Addressed in 210992ed.

thomasguillot:
Addressed in c2433f8f — replaced `id`/`label` on `BaseControl` with `BaseControl.VisualLabel` (rendered as a `<span>`) carrying its own `labelId`, and wired the expander via `aria-labelledby`. This keeps `InteractiveDiv` intact while properly associating the label for assistive technology.

packages/components/src/color-picker/index.js
```diff
@@ -26,35 +27,40 @@ const { InteractiveDiv } = utils;
...
 	const ref = useRef();
+	const id = useInstanceId( ColorPicker, 'newspack-color-picker' );
+	const labelId = `${ id }-label`;
 	const colordColor = colord( color );
 	hooks.useOnClickOutside( ref, () => setIsExpanded( false ) );
 	return (
-		<div className={ classnames( 'newspack-color-picker', className ) }>
-			<div className="newspack-color-picker__label">{ label }</div>
-
+		<BaseControl className={ classnames( 'newspack-color-picker', className ) } help={ help }>
```
rbcorrales: Same thing about `__nextHasNoMarginBottom` mentioned [here](https://github.com/Automattic/newspack-plugin/pull/4580#pullrequestreview-3983544896).

rbcorrales (approved):
Tested locally, and it looks good. Left one minor note, and there are a couple of unresolved Copilot threads worth checking.

The VisualLabel/aria-labelledby thread looks addressed by c2433f8, just needs resolving.

The aria-describedby one is a valid point, though. When `help` is passed, `BaseControl` renders it in a paragraph with an id based on the component id plus `__help`. But the `InteractiveDiv` doesn't have an `aria-describedby` pointing to that paragraph, so screen readers won't announce the help text. Wiring that up would fix it

packages/components/src/color-picker/index.js
```diff
@@ -26,35 +27,40 @@ const { InteractiveDiv } = utils;
...
+			<BaseControl.VisualLabel id={ labelId }>{ label }</BaseControl.VisualLabel>
 			<InteractiveDiv
+				id={ id }
+				aria-labelledby={ labelId }
+				aria-expanded={ isExpanded }
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 0cef7a75.

packages/components/src/color-picker/index.js
```diff
@@ -26,35 +27,40 @@ const { InteractiveDiv } = utils;
...
  * @param {Function}           props.onChange    - Function to call when the color changes.
  * @param {string}             [props.className] - Additional class name.
  * @return {JSX.Element} ColorPicker component.
  */
-const ColorPicker = ( { label, color = '#fff', onChange, className } ) => {
+const ColorPicker = ( { label, help, color = '#ffffff', onChange, className } ) => {
 	const [ isExpanded, setIsExpanded ] = useState( false );
 	const ref = useRef();
+	const id = useInstanceId( ColorPicker, 'newspack-color-picker' );
+	const labelId = `${ id }-label`;
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 0cef7a75.

packages/components/src/color-picker/index.js
thomasguillot: Addressed in 0cef7a75.

PR #4589, opened 2026-03-20
feat(access-control): filter available lists by content restrictions

dkoo:
### All Submissions:

Requires https://github.com/Automattic/newspack-newsletters/pull/2053. Apply access control restrictions to available newsletter subscription lists, and auto-signup users depending on the `newspack_premium_newsletters_auto_signup` option value.

Also removes Registered Access options for premium newsletter gates, as these make less sense for restricting newsletter lists.

To avoid rapid roundtrip requests to the ESP and the potential race conditions that might cause, implements debouncing logic for access checks via an access check queue. When a data event occurs that might trigger a change in a user's access to premium newsletters, their user ID is added to a queue to check their access 10 minutes in the future. Future checks are scheduled via ActionScheduler if available, using WP Cron as a backup if not.

If a scheduled check already exists due to a prior action and isn't scheduled to happen within one minute, the user ID is added to the queue for that check.

If there's no scheduled check or the next check is happening imminently, a new check is scheduled for 10 minutes after the imminent check.

When the scheduled check occurs, the callback reads the queue of user IDs and performs access checks for all at once. Each user is added to or removed from premium newsletter lists based on this access check.

In this way, we avoid trying to update users' lists in real time and only perform the updates once, well after all transitional states have passed. This only applies to the auto-signup/removal mechanism. Access to lists is reflected in real time on the site, so that users can still see and sign up for premium lists manually in the gap period if desired.

Data events that trigger an access check:

- Successful payment for a subscription: fires for new, switch, and renewal payments. 
- Failed payment for a subscription: only for renewal payments.
- Subscription status changes
- Reader verification
- Updates to reader data in the DB

This PR doesn't attempt to reconcile the scenario where a user may have previously unsubscribed from a premium list they have access to; in this PR, the user will be readded automatically if they have access at the time of the check. I'll try to implement a fix for this in a future PR as this one is getting pretty big already.

Closes NPPD-1369.

For easier testing, you may want to temporarily reduce the `Premium_Newsletters::DEFAULT_DELAY` and `Premium_Newsletters::FUTURE_EVENT_THRESHOLD` constants for faster turnaround. I tested successfully with these set to 15 seconds and 10 seconds, respectively.

#### Testing access checks with Active Subscription requirements:

1. Choose a couple of newsletter lists to be restricted. Make sure the restricted lists are shown in Reader Registration and Newsletter Subscription blocks. Also enable these lists for the post-signup newsletter modal in **Audience > Configuration**. 
2. In **Newsletters > Premium** make sure Auto-signup is enabled in Advanced Settings. Set up and activate at least one premium newsletter gate targeting these lists at **Newsletters > Premium**. Try setting up a couple with different lists and access requirements! For example:

Screenshot 2026-03-27 at 11 25 14 AM

3. As an anonymous reader, confirm that you can't see the restricted lists in Reader Registration and Newsletter Subscription blocks, even though they're configured there.
4. Register a reader account via the Sign In modal flow and confirm that you still don't see the restricted lists in the post-registration signup modal. Also confirm that the restricted lists are not shown in My Account > Newsletters.
5. Purchase a subscription product that's required by a premium newsletter gate. After completing the purchase, confirm that you DO see the lists restricted by that gate in the post-checkout signup modal and in My Account > Newsletters.
6. Wait until the `Premium_Newsletters::DEFAULT_DELAY` elapses (or manually trigger the pending ActionScheduler event).
7. In your connected ESP, confirm that the user was added to the restricted lists automatically.
8. As an admin, update the subscription's status to a non-active status.
9. Wait until the delay elapses again. Confirm that the user was removed from the restricted lists automatically.
10. Reactivate the subscription and wait until they're re-added to the restricted lists.
11. As the reader, log into My Account and update your default payment method to a number that will be declined (hint: use the [Stripe test number](https://docs.stripe.com/testing#declined-payments) that will "decline after attaching"). Make sure the active subscription is using this card number.
12. Trigger an early renewal for the subscription that will fail.
13. After the access check delay elapses, confirm that the user was automatically removed from the restricted lists.
14. As the reader, visit My Account and pay for the failed subscription order.
15. After the access check delay elapses, confirm that the user was automatically re-added to the restricted lists.

#### Testing access checks with email domain whitelist

1. Enable an email domain whitelist for a premium newsletter gate.
2. As a reader, register an account with that email domain. Verify via My Account.
3. After the delay elapses, confirm that the reader was automatically added to the restricted lists.
4. In My Account > Newsletters, confirm that the restricted lists are shown.

~#### Testing access with reader data updates~

We decided to remove this for now. Leaving the testing steps here for posterity.

~Note: this wasn't always working reliably for me, so it might need some more work. 😕 ~

1. Enable a reader data key/value pair for a premium newsletter gate. e.g. `foo=bar`
2. As a reader, register an account. Verify via My Account.
3. Via `wp shell`, update the reader data for the user: `Newspack\Reader_Data::update_item( $user_id, 'foo', 'bar' )`
4. [Note: this wasn't always working reliably for me, so it might need some more work. 😕] After the delay elapses, confirm that the reader was automatically added to the restricted lists.
5. In My Account > Newsletters, confirm that the restricted lists are shown.

#### Testing access with institutional access

Note: I didn't implement any auto-signup or removal features for institutional access, since it feels kind of invasive to do so on something like an IP check.

1. Set up an institution with IP whitelist and add this institution to a premium newsletter gate.
2. As a reader, visit the `/institutional-access` endpoint to trigger the IP check.
3. Browse around the site and confirm that the restricted lists are visible wherever shown.

### Other information:


includes/content-gate/class-content-restriction-control.php
```diff
@@ -191,8 +203,17 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null
...
+		 * so this lets us get restriction status for a specific user ID.
+		 *
+		 * @param int $user_id Current user ID.
+		 */
+		$user_id = apply_filters( 'newspack_content_restriction_control_user_id', $user_id ?? get_current_user_id() );
+
 		// Don't restrict this post for users who can edit it.
-		if ( ! empty( $post_id ) && current_user_can( 'edit_post', $post_id ) ) {
+		if ( ! empty( $post_id ) && user_can( $user_id, 'edit_post', $post_id ) ) {
 			return false;
```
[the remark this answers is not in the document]
dkoo: Good catch. 578dbfa caches the `$user_id` being checked and adds it to the post/gate map keys so checks for different user IDs shouldn't collide.

includes/content-gate/class-premium-newsletters.php
```diff
@@ -0,0 +1,384 @@
...
+	 * FUTURE_EVENT_THRESHOLD seconds in the future, the user ID simply rides along
+	 * with that event and no new event is created. If the queue is stale (older than
+	 * MAX_QUEUE_AGE), the delay is overridden to 0 so the event fires on the earliest
+	 * available cron/AS tick.
+	 *
+	 * @param int $user_id The ID of the user to schedule the access check for.
+	 *
+	 * @return void
+	 */
+	private static function schedule_access_check( $user_id ) {
```
miguelpeixe: I'm still going to do a more thorough review, but I'd like to comment on this approach first.

I find this scheduling strategy a bit over-engineered. We can have a recurring cron/AS action always running to pick up queued user IDs.

Each pushed user ID would have a timestamp (maybe just a map `[ $user_id => $timestamp ]`) that gets refreshed anytime it's pushed to the queue again. The recurring action fetches all user IDs with a timestamp older than the determined delay and processes their access check.

includes/data-events/listeners.php
```diff
@@ -11,6 +11,7 @@
 use Newspack\Reader_Activation;
 use Newspack\Donations;
 use Newspack\Memberships;
+use Newspack\Reader_Data;
```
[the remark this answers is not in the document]
dkoo: 12da01e actually makes use of the class, so this is now outdated.

includes/content-gate/class-premium-newsletters.php
```diff
@@ -0,0 +1,379 @@
...
+	 */
+	public static function get_restricted_lists() {
+		if ( ! empty( self::$restricted_lists ) ) {
+			return self::$restricted_lists;
+		}
+		$gates = Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish', true );
+		if ( empty( $gates ) ) {
+			return;
+		}
+		$restricted_lists = [];
```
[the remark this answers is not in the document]
dkoo: Fixed in fbb2f1a00

includes/content-gate/class-content-restriction-control.php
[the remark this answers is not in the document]
dkoo: That fix didn't work. I played with trying to secure this better with a nonce, but I don't think this is worth the effort since the same issue already exists in `trunk`.

tests/unit-tests/content-gate/class-premium-newsletters.php
```diff
@@ -0,0 +1,839 @@
...
+	 */
+	private function get_scheduled_hook_timestamps(): array {
+		$timestamps = [];
+
+		// WP cron.
+		foreach ( _get_cron_array() as $timestamp => $cron ) {
+			if ( isset( $cron[ Premium_Newsletters::SCHEDULED_HOOK ] ) ) {
+				$timestamps[] = (int) $timestamp;
+			}
+		}
```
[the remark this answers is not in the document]
dkoo: Fixed in b2eac08

includes/content-gate/class-premium-newsletters.php
```diff
@@ -0,0 +1,379 @@
...
+	private static function get_public_id( $list_id ) {
+		if ( ! class_exists( 'Newspack\Newsletters\Subscription_List' ) ) {
+			return null;
+		}
+		$list = new Subscription_List( $list_id );
+		if ( ! $list ) {
+			return null;
+		}
+		return $list->get_public_id();
+	}
```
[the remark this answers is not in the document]
dkoo: Fixed in b7ff63f67

includes/reader-activation/class-reader-activation.php
```diff
@@ -1704,8 +1704,21 @@ public static function render_auth_modal() {
...
+				'newspack_content_restriction_control_user_id',
+				function() use ( $user_id ) {
+					return $user_id->ID;
+				}
+			);
+		}
 		ob_start();
-		self::render_newsletters_signup_modal( $request['email_address'] );
+		self::render_newsletters_signup_modal( $email );
 		$html = trim( ob_get_clean() );
```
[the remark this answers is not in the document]
dkoo: This is intentional so that the request to fetch the user's lists in the post-checkout/post-registration modals accurately reflect that user, but open to other ideas

includes/content-gate/class-premium-newsletters.php
dkoo: e5f33a9 refactors the event strategy to a simpler recurring AS/cron event once per 10 minutes (or hourly if cron). The user ID queue is now a simple flat array of user IDs, and the extra complexity around checking for stale queues and future single events has been removed.

includes/reader-activation/class-reader-activation.php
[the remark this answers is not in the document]
dkoo: Actually, I think we can safely remove the filter after rendering the newsletter form since the user should be properly authenticated after this happens. Removed in 96a5891

includes/reader-activation/class-reader-activation.php
```diff
@@ -1704,8 +1704,21 @@ public static function render_auth_modal() {
...
+		// If the email address is associated with a user, use that user's ID for evaluating content restrictions.
+		$user_id = get_user_by( 'email', $email );
+		if ( $user_id ) {
+			$email = $user_id->user_email;
+			add_filter(
+				'newspack_content_restriction_control_user_id',
+				function() use ( $user_id ) {
+					return $user_id->ID;
+				}
+			);
```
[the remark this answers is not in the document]
dkoo: This is true, but I think the security risk is low since the only information you can glean about another user from this REST API is what subscription lists they have access to. I'm open to doing more to fix if we think it's a concern, though. Maybe we can generate a nonce using the email address used to register a new account in the prior request and cross-check that on this REST API callback.

includes/content-gate/class-premium-newsletters.php
```diff
@@ -0,0 +1,359 @@
...
+	 * @return void
+	 */
+	public static function register_access_check_as_event() {
+		if ( false === wp_cache_get( 'newspack_premium_newsletters_recurring_action_scheduled' ) ) {
+			if ( ! as_has_scheduled_action( self::SCHEDULED_HOOK ) ) {
+				as_schedule_recurring_action( time(), self::DEFAULT_DELAY, self::SCHEDULED_HOOK, [], 'newspack' );
+			}
+			wp_cache_set( 'newspack_premium_newsletters_recurring_action_scheduled', true, self::DEFAULT_DELAY );
+		}
+	}
```
miguelpeixe: With the default 30-day retention, this will take 4320 rows in AS. What do you think about using just cron for this? A custom interval could be set with `cron_schedules`, but IMO, every hour should be fine.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -229,14 +260,14 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null
...
+				if ( ! empty( $access_rules ) && ! Access_Rules::evaluate_rules( $access_rules, $user_id ) ) {
 					$is_restricted  = true;
 					$gate_layout_id = $gate['custom_access']['gate_layout_id'] ?? $gate['id'];
 				}
 			}
 
 			if ( $is_restricted && $gate_layout_id ) {
-				self::$post_gate_id_map[ $post_id ] = $gate['id'];
+				self::$post_gate_id_map[ $post_id . '_' . self::$user_id ] = $gate['id'];
 				self::$post_gate_layout_id_map[ $post_id ] = $gate_layout_id;
```
miguelpeixe: It's safer to also use the `$user_id` for the `$post_gate_layout_id_map`.

includes/content-gate/class-premium-newsletters.php
```diff
@@ -0,0 +1,359 @@
...
+
+	/**
+	 * Maybe add or remove the user from restricted lists based on their access status.
+	 *
+	 * @param int   $timestamp Timestamp of the event.
+	 * @param array $data      Data associated with the event.
+	 * @param int   $client_id ID of the client that triggered the event.
+	 */
+	public static function maybe_add_or_remove_lists( $timestamp, $data, $client_id ) {
```
miguelpeixe: Suggestion for clarity:

```suggestion
	public static function maybe_enqueue_access_check( $timestamp, $data, $client_id ) {
```

includes/data-events/listeners.php
```diff
@@ -111,6 +112,39 @@ function( $user ) {
...
+ * suppress the re-dispatch to avoid recursive chains.
+ */
+Data_Events::register_listener(
+	'newspack_reader_data_updated',
+	'reader_data_updated',
+	function( $user_id, $key, $value ) {
+		// Don't dispatch for system-managed read-only keys.
+		if ( in_array( $key, Reader_Data::get_read_only_keys(), true ) ) {
+			return;
+		}
```
miguelpeixe: We shouldn't encourage the use of non-read-only keys for premium newsletters, as these can be edited by the user.

Even though the docblock states that other handlers cover read-only keys, this is not OOB for new keys that we or integrations may implement.

This listener will also enqueue every authenticated reader on every load because of non-read-only keys like `pageviews` getting synced. Maybe we should skip this trigger entirely. Instead, provide and document the API to enqueue the access check for a user.

includes/data-events/listeners.php
```diff
@@ -418,6 +452,69 @@ function( $product_id ) {
...
 
+/**
+ * When a WooCommerce Subscription successfully completes a payment.
+ * This hook is fired for initial orders, renewals, and successful switch orders.
+ *
+ * See: https://woocommerce.com/document/subscriptions/develop/action-reference/#woocommerce_subscription_payment_complete
+ */
+Data_Events::register_listener(
+	'woocommerce_subscription_payment_complete',
+	'subscription_payment_complete',
```
miguelpeixe: Can we rely on existing data events for this? We've been talking about reducing the number of data events we register and leveraging more of their properties: p1774641013508989/1774558789.778009-slack-C07LB7B14GZ

E.g., if we have a `subscription_changed` event, trigger the access check based on the status change.

includes/reader-activation/class-reader-activation.php
```diff
@@ -1704,9 +1726,24 @@ public static function render_auth_modal() {
...
 		ob_start();
-		self::render_newsletters_signup_modal( $request['email_address'] );
+		self::render_newsletters_signup_modal( $email );
 		$html = trim( ob_get_clean() );
+
+		// Reset the current reader user ID so it doesn't affect other operations in this session.
+		if ( $user && $user->ID !== get_current_user_id() ) {
+			self::$current_reader_user_id = 0;
+			remove_filter( 'newspack_content_restriction_control_user_id', [ self::class, 'get_user_id_for_content_restriction' ] );
+		}
```
miguelpeixe, line 1746: This looks a bit convoluted... Is it not possible to set an `X-WP-Nonce` header when fetching this data from `refreshNewslettersSignupModal()` and rely on the current user?

miguelpeixe (changes requested):

includes/reader-activation/class-reader-activation.php
dkoo: The challenge is that when this REST API request is fired, it happens right before the reader account is registered and authenticated, so the current user isn't available yet in this execution.

miguelpeixe: I'm confused... How does it set `$current_reader_user_id` then?

includes/content-gate/class-premium-newsletters.php
dkoo: I was a bit worried that an hour delay might be a bit much for the auto-signup feature. I don't see much benefit to using AS over regular cron, though, since it's now a simple recurring event instead of single events based on triggers.

includes/data-events/listeners.php
dkoo: Originally I didn't want to fire the checks on most status changes because of the transitional status changes that happen during automatic renewals, so these hooks were a more reliable indicator of a successful or failed renewal. But since we're more or less debouncing the checks with the scheduled jobs, I suppose this is no longer as much of a concern.

includes/content-gate/class-premium-newsletters.php
dkoo: Updated in 8baf36423

includes/content-gate/class-content-restriction-control.php
dkoo: Updated in 8baf36423

includes/content-gate/class-premium-newsletters.php
dkoo: dbbcdae91 removes the use of ActionScheduler and relies on an hourly cron job only.

includes/data-events/listeners.php
dkoo: f38fe57cc removes this data event listener and handler. Access checks are now effectively only scheduled based on subscription status changes and upon reader account verification.

includes/data-events/listeners.php
dkoo: f38fe57cc removes these new data event listeners and handlers. Access checks are now effectively only scheduled based on subscription status changes and upon reader account verification.

includes/reader-activation/class-reader-activation.php
dkoo: I think we can continue discussing a solution for this, but it's best tackled in a separate PR since it might touch other flows outside of premium newsletters.

miguelpeixe (approved):
Tests well! Thank you for the revisions.

I'm working on improving the session hydration handling for the post-checkout premium newsletter offering in #4618.

I'm still able to see restricted newsletters in the My Account -> Newsletters page, but let's tackle it in a separate PR.

miguelpeixe:
Hmm... I can actually see restricted newsletters in the newsletter sign-up modal and on the My Account page, regardless of my access.

I had them unchecked in the "Present newsletter signup after checkout and registration" while first testing, and toggled them on after purchasing the subscription. That's why I missed it.

includes/reader-activation/class-reader-activation.php
dkoo: Forgot to respond to this, sorry!

> I'm confused... How does it set `$current_reader_user_id` then?

The `api_render_newsletters_signup_form` callback can run in an unauthenticated OR an authenticated session:

- If an anonymous reader registers for a new account or completes a Woo purchase, it will fire in an unauthenticated session as the reader account is created on the server
- If the reader is already logged in and completes a Woo purchase, it will fire in the authenticated session

In the first scenario, `get_current_user_id()` will return `0`, so we can compare that to the `$user` we fetch via the passed email address. In this case, we assume that the passed email address is the brand-new user account that was just created and cache that user ID for the sole purpose of rendering the newsletter signup modal.

In the second scenario, `get_current_user_id()` will return the correct user ID, so no further action is needed.

dkoo:
@miguelpeixe is your Newsletters plugin running the [required branch](https://github.com/Automattic/newspack-newsletters/pull/2053)?

miguelpeixe:
Not this time around 🤦 
Sorry about that

PR #4594, opened 2026-03-24
feat(reader-data): add engagement fields

chickenn00dle:
### All Submissions:

Adds five new reader data fields to the Reader Data library for tracking reader engagement behavior:

- **`first_visit_date`** — Timestamp of the reader's first visit (set once, never overwritten)
- **`last_active`** — Timestamp of the reader's most recent page load (updated every page load)
- **`articles_read`** — Count of unique articles viewed (derived from `article_view` activities)
- **`favorite_categories`** — Top 5 category IDs by view frequency (derived from `article_view` activity data)
- **`paywall_hits`** — Count of paywall encounters

All fields are client-authoritative and server-persisted — computed in JS, stored in localStorage, and synced to server via the existing store sync mechanism for logged-in users.

Closes https://linear.app/a8c/issue/NPPD-1387/create-new-fields-on-reader-data-library .

1. Open DevTools and view **Local Storage**, filter by `np_reader_1_`
2. Clear localStorage, reload the page
3. Verify `np_reader_1_first_visit_date` and `np_reader_1_last_active` appear with timestamp values
4. Reload again — verify `first_visit_date` is unchanged but `last_active` has a newer timestamp
5. Navigate to an article — verify `np_reader_1_articles_read` appears with value `1`
6. Navigate to a different article (ideally in a different category) — verify `articles_read` is now `2`
7. Check `np_reader_1_favorite_categories` — should contain an array of category IDs
8. **Paywall test**: Enable a paywall, and as an anonymous visitor, verify `np_reader_1_paywall_hits` increments
9. **Sync test**: Log in as a reader, verify fields sync to server via Network tab (POST requests to `/newspack/v1/reader-data`)

### Other information:

### DIFF
--- a/includes/content-gate/class-metering.php
+++ b/includes/content-gate/class-metering.php
@@ -434,13 +434,6 @@ class Metering {
 		 */
 		self::$logged_in_metering_cache[ $post_id ] = apply_filters( 'newspack_content_gate_is_logged_in_metering_allowed', $allowed, $post_id );
 
-		// Increment paywall_hits in reader data when the reader is blocked.
-		if ( ! self::$logged_in_metering_cache[ $post_id ] ) {
-			$current_hits = Reader_Data::get_data( \get_current_user_id(), 'paywall_hits' );
-			$current_hits = $current_hits ? (int) json_decode( $current_hits ) : 0;
-			Reader_Data::update_item( \get_current_user_id(), 'paywall_hits', $current_hits + 1 );
-		}
-
 		return self::$logged_in_metering_cache[ $post_id ];
 	}
 
--- a/includes/content-gate/class-metering.php
+++ b/includes/content-gate/class-metering.php
@@ -426,13 +426,6 @@ class Metering {
 		// Allowed if the content has been accessed or the metering limit has not been reached.
 		$allowed = $accessed_content || ! $limited;
 
-		// Increment paywall_hits in reader data when the reader is blocked.
-		if ( ! $allowed ) {
-			$current_hits = Reader_Data::get_data( \get_current_user_id(), 'paywall_hits' );
-			$current_hits = $current_hits ? (int) json_decode( $current_hits ) : 0;
-			Reader_Data::update_item( \get_current_user_id(), 'paywall_hits', $current_hits + 1 );
-		}
-
 		/**
 		 * Filters whether to allow content rendering through metering for logged in user.
 		 *
@@ -440,6 +433,13 @@ class Metering {
 		 * @param int  $post_id                       Post ID.
 		 */
 		self::$logged_in_metering_cache[ $post_id ] = apply_filters( 'newspack_content_gate_is_logged_in_metering_allowed', $allowed, $post_id );
+
+		// Increment paywall_hits in reader data when the reader is blocked.
+		if ( ! self::$logged_in_metering_cache[ $post_id ] ) {
+			$current_hits = Reader_Data::get_data( \get_current_user_id(), 'paywall_hits' );
+			$current_hits = $current_hits ? (int) json_decode( $current_hits ) : 0;
+			Reader_Data::update_item( \get_current_user_id(), 'paywall_hits', $current_hits + 1 );
+		}
 
 		return self::$logged_in_metering_cache[ $post_id ];
 	}
--- a/includes/content-gate/class-metering.php
+++ b/includes/content-gate/class-metering.php
@@ -426,6 +426,13 @@ class Metering {
 		// Allowed if the content has been accessed or the metering limit has not been reached.
 		$allowed = $accessed_content || ! $limited;
 
+		// Increment paywall_hits in reader data when the reader is blocked.
+		if ( ! $allowed ) {
+			$current_hits = Reader_Data::get_data( \get_current_user_id(), 'paywall_hits' );
+			$current_hits = $current_hits ? (int) json_decode( $current_hits ) : 0;
+			Reader_Data::update_item( \get_current_user_id(), 'paywall_hits', $current_hits + 1 );
+		}
+
 		/**
 		 * Filters whether to allow content rendering through metering for logged in user.
 		 *
--- a/src/content-gate/gate.js
+++ b/src/content-gate/gate.js
@@ -160,15 +160,18 @@ function getGateEventPayload( payload, gate ) {
 /**
  * Handle when the gate is seen.
  *
- * @param {HTMLElement} gate The gate element.
+ * @param {HTMLElement} gate            The gate element.
+ * @param {boolean}     shouldRecordHit Whether to record a hit in RAS for this seen event. Defaults to false.
  */
-function handleSeen( gate ) {
-	// paywall_hits - Number of times reader has reached a paywall.
-	window.newspackRAS = window.newspackRAS || [];
-	window.newspackRAS.push( function ( ras ) {
-		const currentHits = ras.store.get( 'paywall_hits' ) || 0;
-		ras.store.set( 'paywall_hits', currentHits + 1 );
-	} );
+function handleSeen( gate, shouldRecordHit = false ) {
+	if ( shouldRecordHit ) {
+		// paywall_hits - Number of times reader has reached a paywall.
+		window.newspackRAS = window.newspackRAS || [];
+		window.newspackRAS.push( function ( ras ) {
+			const currentHits = ras.store.get( 'paywall_hits' ) || 0;
+			ras.store.set( 'paywall_hits', currentHits + 1 );
+		} );
+	}
 
 	if ( 'function' !== typeof window.gtag ) {
 		return;
@@ -321,12 +324,16 @@ domReady( function () {
 	} else {
 		window.addEventListener( 'resize', handleFloatingElements );
 		handleFloatingElements();
+		let seen = false;
 		// Seen event for inline gate.
 		const detectSeen = () => {
 			const delta = ( gate?.getBoundingClientRect().top || 0 ) - window.innerHeight / 2;
 			if ( delta < 0 ) {
-				handleSeen( gate );
-				document.removeEventListener( 'scroll', detectSeen );
+				handleSeen( gate, ! seen );
+				if ( 'function' === typeof window.gtag ) {
+					document.removeEventListener( 'scroll', detectSeen );
+				}
+				seen = true;
 			}
 		};
 		document.addEventListener( 'scroll', detectSeen );
--- a/src/content-gate/gate.js
+++ b/src/content-gate/gate.js
@@ -163,6 +163,13 @@ function getGateEventPayload( payload, gate ) {
  * @param {HTMLElement} gate The gate element.
  */
 function handleSeen( gate ) {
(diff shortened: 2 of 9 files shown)


includes/content-gate/class-metering.php
```diff
@@ -426,6 +426,13 @@ public static function is_logged_in_metering_allowed( $post_id = null ) {
 		// Allowed if the content has been accessed or the metering limit has not been reached.
 		$allowed = $accessed_content || ! $limited;
 
+		// Increment paywall_hits in reader data when the reader is blocked.
+		if ( ! $allowed ) {
```
chickenn00dle: This needs to be incremented here since logged in metered paywall hits are done on the backend.

src/content-gate/metering.js
```diff
@@ -94,6 +94,8 @@ function meter( ras ) {
 	if ( settings.count <= data.content.length && ! data.content.includes( settings.post_id ) ) {
 		lockContent( ras );
 		ras.dispatchActivity( 'metering_restricted', { post_id: settings.post_id, metering: data } );
+		const currentHits = ras.store.get( 'paywall_hits' ) || 0;
```
chickenn00dle: I'm calculating paywall hits in metering logic because the metadata schema specifically mentions this should only update for metered paywalls. But I do want to confirm whether this is correct and we don't want to increment this for non-metered paywalls.

includes/content-gate/class-metering.php
```diff
@@ -426,6 +426,13 @@ public static function is_logged_in_metering_allowed( $post_id = null ) {
...
+			$current_hits = Reader_Data::get_data( \get_current_user_id(), 'paywall_hits' );
+			$current_hits = $current_hits ? (int) json_decode( $current_hits ) : 0;
+			Reader_Data::update_item( \get_current_user_id(), 'paywall_hits', $current_hits + 1 );
+		}
```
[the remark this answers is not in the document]
chickenn00dle: Moved this after the filter in https://github.com/Automattic/newspack-plugin/pull/4594/commits/c3db9f42213b00263aa0e9e63aac680f6784f07d

src/reader-activation/engagement.js
```diff
@@ -0,0 +1,14 @@
...
+ */
+export default function setupEngagement( ras ) {
+	// first_visit_date — set once, never overwritten.
+	if ( ! ras.store.get( 'first_visit_date' ) ) {
+		ras.store.set( 'first_visit_date', Date.now() );
+	}
+
+	// last_active — update on every page load.
+	ras.store.set( 'last_active', Date.now() );
+}
```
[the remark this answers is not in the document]
chickenn00dle: Added unit tests in https://github.com/Automattic/newspack-plugin/pull/4594/commits/ab17b6a419b33b2a40d97cdf465a1bd3d66fe99f

src/reader-activation/article-view.js
```diff
@@ -36,5 +36,24 @@ export default function setupArticleViewsAggregates( ras ) {
...
+			const cats = activity.data?.categories || [];
+			for ( const cat of cats ) {
+				catCounts[ cat ] = ( catCounts[ cat ] || 0 ) + 1;
+			}
+		}
+		const topCategories = Object.entries( catCounts )
+			.sort( ( a, b ) => b[ 1 ] - a[ 1 ] )
+			.slice( 0, 5 )
+			.map( ( [ id ] ) => Number( id ) );
+		ras.store.set( 'favorite_categories', topCategories );
```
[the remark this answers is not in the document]
chickenn00dle: Added unit tests in https://github.com/Automattic/newspack-plugin/pull/4594/commits/ab17b6a419b33b2a40d97cdf465a1bd3d66fe99f

src/reader-activation/mocks/ras.js
```diff
@@ -0,0 +1,72 @@
...
+			handlers[ event ] = callback;
+		} ),
+		getActivities: jest.fn( () => activities ),
+		getUniqueActivitiesBy: jest.fn( () => {
+			const seen = {};
+			return activities.filter( a => {
+				if ( seen[ a.data.post_id ] ) {
+					return false;
+				}
+				seen[ a.data.post_id ] = true;
```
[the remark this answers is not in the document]
chickenn00dle: I'll leave this as is since only two tests are using this. We can iterate on this as we create more tests for reader data library

src/reader-activation/mocks/ras.js
```diff
@@ -0,0 +1,72 @@
...
+				handlers[ event ]( { detail } );
+			}
+		},
+		/**
+		 * Reset all state between tests.
+		 */
+		reset() {
+			for ( const key in storeData ) {
+				delete storeData[ key ];
+			}
```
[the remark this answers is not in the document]
chickenn00dle: Added in https://github.com/Automattic/newspack-plugin/pull/4594/commits/21c456eef54e695c68163367c031be1800ae59e8

src/content-gate/gate.js
```diff
@@ -318,10 +325,8 @@ domReady( function () {
 		const detectSeen = () => {
 			const delta = ( gate?.getBoundingClientRect().top || 0 ) - window.innerHeight / 2;
 			if ( delta < 0 ) {
-				if ( 'function' === typeof window.gtag ) {
-					handleSeen( gate );
-					document.removeEventListener( 'scroll', detectSeen );
-				}
+				handleSeen( gate );
+				document.removeEventListener( 'scroll', detectSeen );
```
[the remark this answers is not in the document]
chickenn00dle: Ah. Good call. Re-added check for gtag in https://github.com/Automattic/newspack-plugin/pull/4594/commits/21c456eef54e695c68163367c031be1800ae59e8

src/reader-activation/article-view.test.js
```diff
@@ -0,0 +1,67 @@
...
+	beforeEach( () => {
+		mock = createMockRAS();
+		setupArticleViewsAggregates( mock.ras );
+	} );
+
+	afterEach( () => {
+		mock.reset();
+	} );
+
+	function simulateArticleView( data, timestamp = Date.now() ) {
```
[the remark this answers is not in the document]
chickenn00dle: Gonna leave this one as is since only two tests use this. We can iterate as more tests are added.

src/content-gate/gate.js
```diff
@@ -314,14 +324,16 @@ domReady( function () {
 	} else {
 		window.addEventListener( 'resize', handleFloatingElements );
 		handleFloatingElements();
+		let seen = false;
 		// Seen event for inline gate.
 		const detectSeen = () => {
 			const delta = ( gate?.getBoundingClientRect().top || 0 ) - window.innerHeight / 2;
 			if ( delta < 0 ) {
+				handleSeen( gate, ! seen );
```
chickenn00dle, line 332: Just noting I initially removed the typeof window.gtag check below to ensure this only triggers once, but it seems this check is to avoid a race condition and ensure the listener isn't removed when gtag hasn't been added to window before this runs. So instead, I am adding a `seen` flag to determine when to update paywall hits.

src/reader-activation/article-view.js
```diff
@@ -36,5 +36,24 @@ export default function setupArticleViewsAggregates( ras ) {
...
+		for ( const activity of allActivities ) {
+			const cats = activity.data?.categories || [];
+			for ( const cat of cats ) {
+				catCounts[ cat ] = ( catCounts[ cat ] || 0 ) + 1;
+			}
+		}
+		const topCategories = Object.entries( catCounts )
+			.sort( ( a, b ) => b[ 1 ] - a[ 1 ] )
+			.slice( 0, 5 )
+			.map( ( [ id ] ) => Number( id ) );
```
chickenn00dle, line 57: The spec calls for a list of category strings, but since we only have ids here, I am storing ids which can be converted to string representation in the backend. But let me know if we prefer to store as strings here instead.

src/reader-activation/engagement.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Set up general reader engagement fields.
+ *
+ * @param {Object} ras Reader Activation object.
+ */
+export default function setupEngagement( ras ) {
+	// first_visit_date — Date of the reader's very first visit to the site, regardless of whether or not they registered.
+	if ( ! ras.store.get( 'first_visit_date' ) ) {
+		ras.store.set( 'first_visit_date', Date.now() );
+	}
```
miguelpeixe: The store sync strategy will overwrite this on every new authenticated session. The fresh session will generate a new `first_visit_date` and upon authentication, the new value will replace the one in the db:

https://github.com/Automattic/newspack-plugin/blob/bb4813cf15185cabb045bc90be9a816f05b63c84/src/reader-activation/store.js#L263-L266

I don't think we ever got a consensus on whether the stored value takes precedence over the new value from the reader session. There's a case for both...

src/reader-activation/article-view.js
```diff
@@ -36,5 +36,24 @@ export default function setupArticleViewsAggregates( ras ) {
...
+		ras.store.set( 'favorite_categories', topCategories );
```
miguelpeixe, line 58: Can we replace the existing `favorite_categories` matching function on newspack-popups to use this, so we have a single source of truth?

miguelpeixe (changes requested):

src/reader-activation/engagement.js
chickenn00dle: Good catch @miguelpeixe!

I added a commit to ensure we always defer to the oldest date for `first_visit_date` here: https://github.com/Automattic/newspack-plugin/pull/4594/commits/4a7f352a67b38e787686b2727841fe0e6266c8e1

I suspect `first_visit_date` might be the only key which requires the stored value take precedence as the other fields should be updated with the latest activity right? I'm happy to revisit this if this isn't the case though.

src/reader-activation/article-view.js
chickenn00dle: Sure thing! I'll work on this in this PR: https://github.com/Automattic/newspack-popups/pull/1542

miguelpeixe: Looks like we're losing [part of the matching logic](https://github.com/Automattic/newspack-popups/pull/1542/changes#diff-8cd87b1ff5acb20a08d0614b3aae6606d72c93efd456edcaf0101c37c8fad6adL28-L33), to have at least 2 categories or 2 posts within the same category.

Do you think we can move that from the matching function to the list generation itself? In reality, the reader shouldn't really have a _favorite_ category if they just read a single article that happens to match a category.

src/reader-activation/engagement.js
miguelpeixe: I think this works well for this case. I'm working on a session hydration feature in #4618 that will require some tweaks to support this new requirement. I can adapt once this is merged.

src/reader-activation/article-view.js
chickenn00dle: Good call. Added a filter for this in https://github.com/Automattic/newspack-plugin/pull/4594/changes/68296d0e3ca8c2a46c8e0824d5376c7f69cbcec9

miguelpeixe (approved):
Thank you for the revisions!

PR #4597, opened 2026-03-24
feat(integrations): add subscription and donation metadata

chickenn00dle:
## Summary

- Implements `get_metadata()` for the `Subscription` contact metadata class with 13 get methods covering subscriber status, active count, dates, billing cycle, payments, product names, coupon codes, and cancellation reasons — all filtered to non-donation subscriptions only.
- Implements `get_metadata()` for the `Donation` class by extending `Subscription`, inverting the filter to donation-only, and adding donor status labels (Monthly/Yearly/Ex-* Donor), one-time donation fallbacks, and previous donation amount tracking.

Closes NPPD-1388

## Test plan

1. Enabled the following in `wp-config.php`:
```
define( 'NEWSPACK_LOG_LEVEL', 3 );
define( `NEWSPACK_SYNC_METADATA_VERSION`, '1.0' );
define( 'NEWSPACK_INTEGRATIONS_SETTINGS_ENABLED', true );
```
2. Navigate to Audience > Integrations > ESP and confirm the donation and subscription fields listed in the linear task are available in the outgoing metadata section
3. Check them all and save, confirming they persist on reload
4. As a new reader, register to the site
5. Check the logs and verify the contact is synced with the relevant metadata in a default empty state (note you can also just check the connected ESP, however you may run into issues with merge field limits depending on the ESP):
Screenshot 2026-03-25 at 16 47 37
6. As the same reader, purchase a one-time donation
7. Check the logs for another sync entry, this time with the donation status, product name, amount, and date populated
8. As the same reader, purchase a recurring donation
9. Check the logs for another sync entry, this time with all relevant donation fields populated (only end date and previous product/amount should be empty)
10. As the reader, purchase a non-donation subscription product
11. Check the logs for another sync entry, this time with all the relevant subscription fields populated (only end date, cancellation reason, and previous product fields should be empty)
12. Now cancel the subscription (either as the reader or manually as the admin)
13. Check the logs for another sync entry, this time the subscription status, count, end date, and reason should be populated
14. Finally, cancel the recurring donation
15. Check the logs for another sync entry, this time the donation status, count, and end date should be populated

### DIFF
--- a/includes/reader-activation/sync/class-contact-metadata.php
+++ b/includes/reader-activation/sync/class-contact-metadata.php
@@ -14,9 +14,9 @@ defined( 'ABSPATH' ) || exit;
  */
 abstract class Contact_Metadata {
 	/**
-	 * The date format to use for all date fields, which is MM/DD/YYYY.
+	 * The date format to use for all date fields, which is YYYY-MM-DD HH:MM:SS.
 	 */
-	const DATE_FORMAT = 'm/d/Y';
+	const DATE_FORMAT = 'Y-m-d H:i:s';
 
 	/**
 	 * The WP_User object.
--- a/includes/reader-activation/sync/class-contact-metadata.php
+++ b/includes/reader-activation/sync/class-contact-metadata.php
@@ -13,6 +13,10 @@ defined( 'ABSPATH' ) || exit;
  * Reader Activation Class.
  */
 abstract class Contact_Metadata {
+	/**
+	 * The date format to use for all date fields, which is MM/DD/YYYY.
+	 */
+	const DATE_FORMAT = 'm/d/Y';
 
 	/**
 	 * The WP_User object.
@@ -121,5 +125,22 @@ abstract class Contact_Metadata {
 			return trim( $this->customer->get_billing_first_name() . ' ' . $this->customer->get_billing_last_name() );
 		}
 		return '';
+	}
+
+	/**
+	 * Format a date string to MM/DD/YYYY.
+	 *
+	 * @param string $date_string Date string from WooCommerce.
+	 * @return string Formatted date or empty string.
+	 */
+	protected function format_date( $date_string ) {
+		if ( empty( $date_string ) || '0' === $date_string ) {
+			return '';
+		}
+		$timestamp = strtotime( $date_string );
+		if ( ! $timestamp ) {
+			return '';
+		}
+		return gmdate( self::DATE_FORMAT, $timestamp );
 	}
 }
--- a/includes/reader-activation/sync/class-contact-sync.php
+++ b/includes/reader-activation/sync/class-contact-sync.php
@@ -132,6 +132,9 @@ class Contact_Sync extends Sync {
 			}
 		}
 
+		Logger::log( sprintf( 'Syncing contact %s for context "%s".', $contact['email'] ?? 'unknown', $context ) );
+		Logger::log( $contact );
+
 		return self::push_to_integrations( $contact, $context, $existing_contact );
 	}
 
--- a/includes/reader-activation/sync/class-contact-sync.php
+++ b/includes/reader-activation/sync/class-contact-sync.php
@@ -11,6 +11,7 @@ use Newspack\Reader_Activation;
 use Newspack\Reader_Activation\Integrations;
 use Newspack\Data_Events;
 use Newspack\Logger;
+use Newspack\Reader_Activation\Sync\Metadata;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -132,8 +133,11 @@ class Contact_Sync extends Sync {
 			}
 		}
 
-		Logger::log( sprintf( 'Syncing contact %s for context "%s".', $contact['email'] ?? 'unknown', $context ) );
-		Logger::log( $contact );
+		// Added logging here to more easily monitor integration sync data. Can be removed once integrations are released.
+		if ( 'legacy' !== Metadata::get_version() ) {
+			Logger::log( sprintf( 'Syncing contact %s for context "%s".', $contact['email'] ?? 'unknown', $context ) );
+			Logger::log( $contact );
+		}
 
 		return self::push_to_integrations( $contact, $context, $existing_contact );
 	}
--- a/includes/reader-activation/sync/contact-metadata/class-donation.php
+++ b/includes/reader-activation/sync/contact-metadata/class-donation.php
@@ -7,23 +7,32 @@
 
 namespace Newspack\Reader_Activation\Sync\Contact_Metadata;
 
-use Newspack\Reader_Activation\Sync\Contact_Metadata;
+use Newspack\Donations;
+use Newspack\WooCommerce_Connection;
 
 defined( 'ABSPATH' ) || exit;
 
 /**
  * Donation metadata class.
+ *
+ * Extends Subscription to reuse subscription data access helpers,
+ * with the filter inverted to only consider donation subscriptions.
  */
-class Donation extends Contact_Metadata {
+class Donation extends Subscription {
 
 	/**
-	 * Whether or not the metadata fields of this class are available to be synced.
+	 * Cache for the one-time donation order.
 	 *
-	 * @return boolean
+	 * @var \WC_Order|null
 	 */
-	public static function is_available() {
-		return true;
-	}
+	private $one_time_donation_order_cache;
+
+	/**
+	 * Whether the one-time donation order has been resolved.
+	 *
+	 * @var bool
+	 */
+	private $one_time_donation_order_resolved = false;
 
 	/**
 	 * The name of the metadata class, used as a section name for the fields handled by this class when syncing and in the UI for selecting which fields to sync.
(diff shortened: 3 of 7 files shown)


includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -132,6 +132,9 @@ public static function sync( $contact, $context = '', $existing_contact = null )
 			}
 		}
 
+		Logger::log( sprintf( 'Syncing contact %s for context "%s".', $contact['email'] ?? 'unknown', $context ) );
+		Logger::log( $contact );
```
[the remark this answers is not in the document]
chickenn00dle: This is already being logged in the legacy implementation when normalizing contact data. I'm gonna keep this here, but I did add a check for non-legacy before logging.

includes/reader-activation/sync/contact-metadata/class-subscription.php
```diff
@@ -63,6 +86,370 @@ public static function get_fields() {
...
+		if ( ! empty( $active ) ) {
+			$this->current_subscription_cache = reset( $active );
+			return $this->current_subscription_cache;
+		}
+
+		foreach ( $this->get_user_subscriptions() as $subscription ) {
+			if ( $subscription->has_status( WooCommerce_Connection::FORMER_SUBSCRIBER_STATUSES ) ) {
+				$this->current_subscription_cache = $subscription;
+				return $this->current_subscription_cache;
+			}
```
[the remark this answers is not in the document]
chickenn00dle: `wcs_get_users_subscriptions` does return an associative array keyed by subscription id and our `get_user_subscriptions` method currently handles this. Also, `wcs_get_users_subscriptions` already sorts subscriptions based on id from highest to lowest, which means the latest subscriptions will appear first in the array.

includes/reader-activation/sync/contact-metadata/class-donation.php
```diff
@@ -62,6 +71,220 @@ public static function get_fields() {
...
+			]
+		);
+
+		foreach ( $orders as $order ) {
+			if ( Donations::is_donation_order( $order ) ) {
+				$this->one_time_donation_order_cache = $order;
+				return $this->one_time_donation_order_cache;
+			}
+		}
+
```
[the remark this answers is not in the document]
chickenn00dle: Good call. Updated in https://github.com/Automattic/newspack-plugin/pull/4597/commits/cf97977eeb1b8dcfea679b3c988fda77d4068fca

tests/mocks/wc-mocks.php
```diff
@@ -239,11 +290,28 @@ public function get_billing_period() {
...
+			return false;
 		}
-		return false;
+		if ( ! empty( $exclude_statuses ) ) {
+			foreach ( array_reverse( $this->orders ) as $order ) {
+				if ( ! $order->has_status( $exclude_statuses ) ) {
+					return $order;
+				}
+			}
+			return false;
```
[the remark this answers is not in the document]
chickenn00dle: Fixed in https://github.com/Automattic/newspack-plugin/pull/4597/commits/cf97977eeb1b8dcfea679b3c988fda77d4068fca

tests/mocks/wc-mocks.php
```diff
@@ -239,11 +290,28 @@ public function get_billing_period() {
...
+		}
+		return end( $this->orders );
```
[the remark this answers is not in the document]
chickenn00dle: Fixed in https://github.com/Automattic/newspack-plugin/pull/4597/commits/cf97977eeb1b8dcfea679b3c988fda77d4068fca

includes/reader-activation/sync/contact-metadata/class-donation.php
```diff
@@ -62,6 +71,224 @@ public static function get_fields() {
...
+			);
+
+			foreach ( $orders as $order ) {
+				if ( Donations::is_donation_order( $order ) ) {
+					$this->one_time_donation_order_cache = $order;
+					return $this->one_time_donation_order_cache;
+				}
+			}
+		} while ( ! empty( $orders ) );
+
```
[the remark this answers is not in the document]
chickenn00dle: I updated the mock to "handle" pagination instead: https://github.com/Automattic/newspack-plugin/pull/4597/commits/e4e5722823da381cdec95feec4683be8a6775225

includes/reader-activation/sync/contact-metadata/class-subscription.php
```diff
@@ -63,6 +86,370 @@ public static function get_fields() {
...
+	protected function get_current_subscription() {
+		if ( $this->current_subscription_resolved ) {
+			return $this->current_subscription_cache;
+		}
+
+		$this->current_subscription_resolved = true;
+
+		$active = $this->get_active_subscriptions();
+		if ( ! empty( $active ) ) {
+			$this->current_subscription_cache = reset( $active );
```
leogermani: you are just getting the first review in the array. Will it be the most recent? How are they ordered?

includes/reader-activation/sync/contact-metadata/class-subscription.php
```diff
@@ -63,6 +86,370 @@ public static function get_fields() {
...
+			return $this->current_subscription_cache;
+		}
+
+		foreach ( $this->get_user_subscriptions() as $subscription ) {
```
leogermani: same question about order.

includes/reader-activation/sync/contact-metadata/class-subscription.php
leogermani: Looking at the current methods, it seems that they do the same thing... but worth checking if this is true.

Also, I think there's a requirement to favor non-gift subscriptions over gifts, so it's worth adding this rule as well

includes/reader-activation/sync/class-contact-metadata.php
```diff
@@ -13,6 +13,10 @@
  * Reader Activation Class.
  */
 abstract class Contact_Metadata {
+	/**
+	 * The date format to use for all date fields, which is MM/DD/YYYY.
+	 */
+	const DATE_FORMAT = 'm/d/Y';
```
leogermani: The current default format is 'Y-m-d H:i:s'. Any reason to use a different one? https://github.com/Automattic/newspack-plugin/blob/18a1495d6a2cd5655a270ffd68eba7f4dfb75fb4/includes/reader-activation/sync/class-metadata.php#L20

leogermani (changes requested):
Beautiful!

Works well and has a clean implementation! 

The only blocker comment is that it's missing the rule to prioritize non-gift subscriptions over gifts in the current subscription.

I also left a few minor comments.

And I have one kind of a NIT, let me know what you think.

I think that instead of Donations extending Subscriptions, it would be better if both classes extended a common shared class. Only because the relationship would be more explicit and it's less likely that someone in the future makes a change to the Subscription class without realizing it also affects Donations...

includes/reader-activation/sync/contact-metadata/class-subscription.php
chickenn00dle: > you are just getting the first review in the array. Will it be the most recent? How are they ordered?

Yes. It should be the most recent. Subscriptions are populated here: https://github.com/Automattic/newspack-plugin/blob/e4e5722823da381cdec95feec4683be8a6775225/includes/reader-activation/sync/contact-metadata/class-subscription.php#L140-L145

The `wcs_get_users_subscriptions` method returns an array of subscriptions [in descending order by ID](https://github.com/woocommerce/woocommerce-subscriptions/blob/6c147710e1188a79f21428e8b618f7c35d4740df/includes/core/data-stores/class-wcs-customer-store-cpt.php#L55). We should be respecting this order when populating our user subscriptions cache and so the first entry should be the subscription with the largest ID (the most recent).

That said, if we feel this is too fragile, I'm happy to implement our own sorting logic.

includes/reader-activation/sync/contact-metadata/class-subscription.php
chickenn00dle: See https://github.com/Automattic/newspack-plugin/pull/4597#discussion_r3029076275

includes/reader-activation/sync/contact-metadata/class-subscription.php
chickenn00dle: Good catch about non-gift preference! Added a commit to account for this here: https://github.com/Automattic/newspack-plugin/pull/4597/commits/df3220bdbe029bdac4248c8e2320f67ab7c72fce

includes/reader-activation/sync/class-contact-metadata.php
chickenn00dle: Good question. No reason in particular. This was an arbitrary decision made by Claude 😝. Updated to use the current default in https://github.com/Automattic/newspack-plugin/pull/4597/commits/df3220bdbe029bdac4248c8e2320f67ab7c72fce

chickenn00dle:
Thanks for your review @leogermani! I addressed most of your feedback and this is ready for another round of review.

> I think that instead of Donations extending Subscriptions, it would be better if both classes extended a common shared class. Only because the relationship would be more explicit and it's less likely that someone in the future makes a change to the Subscription class without realizing it also affects Donations...

My thinking is donations overlap subscriptions in functionality so much so that if there needs to be a change to the subscription class in the future, its very likely it will also need to be made to the donation class as well. There are also references to the donation class in the `is_relevant_subscription` method used throughout the class although admittedly the file is a bit long so this could be missed. I've left it as is for this round of changes, but if you feel strongly about this, I can abstract some of the functionality into a common shared one.

includes/reader-activation/sync/class-contact-metadata.php
chickenn00dle: Oh wait. Sorry. This was not arbitrary. Looks like [the metadata spec sheet](https://docs.google.com/spreadsheets/d/1uVJaDgk5GRBPmIOx3vipZtSQp3RhCemjfwYEGkfoicc/edit?gid=870733192#gid=870733192) defines the 'm/d/Y' format. 

Is this a hard requirement? If so I'll revert this format change to what I had initially.

leogermani:
> Thanks for your review @leogermani! I addressed most of your feedback and this is ready for another round of review.
> 
> > I think that instead of Donations extending Subscriptions, it would be better if both classes extended a common shared class. Only because the relationship would be more explicit and it's less likely that someone in the future makes a change to the Subscription class without realizing it also affects Donations...
> 
> My thinking is donations overlap subscriptions in functionality so much so that if there needs to be a change to the subscription class in the future, its very likely it will also need to be made to the donation class as well. There are also references to the donation class in the `is_relevant_subscription` method used throughout the class although admittedly the file is a bit long so this could be missed. I've left it as is for this round of changes, but if you feel strongly about this, I can abstract some of the functionality into a common shared one.

Makes sense

leogermani (approved):
Thanks for the changes.

Approving for now, we can check with Katie and others about the date format later

PR #4603, opened 2026-03-27
feat(handoff): add URL-based handoff with customizable banner text

thomasguillot:
### All Submissions:

### Context:

The handoff feature appears to have been broken on `trunk` for some time — the return banner doesn't seem to render at all when testing via the Components Demo. On top of that, the old implementation had become quite dated: the JS banner was embedding the **old Newspack logo** as a hardcoded inline SVG in CSS. This PR fixes the breakage, removes the logo entirely, and modernises the banner design and the component's API.

Extends the Handoff component and ActionCard to support handing off to any arbitrary WordPress admin URL, not just Newspack-managed plugins. Previously, handoff was only possible when a managed plugin slug was available.

Key changes:
- New `url` prop on `Handoff` component for URL-based navigation (no plugin required)
- New `handoffUrl` prop on `ActionCard` as a convenience wrapper
- New `bannerText` and `bannerButtonText` props on both components for customizing the return banner message
- New `POST /newspack/v1/handoff` REST endpoint that stores the destination URL and banner options; validates that the URL is same-site to prevent open-redirect attacks
- New server-side rendered handoff banner (`insert_handoff_banner_static`) for Newspack wizard pages where the JS banner can't render, hooked via a new `newspack_before_wizard_content` action
- Handoff banner redesigned: solid primary-color background, white text, uses WP base-styles variables — old Newspack logo removed
- The `needs_handoff_return_ui()` gate no longer requires setup to be incomplete — the banner now shows whenever a handoff is active
- `clear_handoff_url()` extended to also clear on non-Newspack screens when a destination page was registered, preventing the banner from lingering on unrelated admin pages
- `register_handoff_for_plugin()` now resets `NEWSPACK_HANDOFF_DESTINATION_PAGE` to prevent stale state from a prior URL-based handoff affecting subsequent plugin-based handoffs
- PHPUnit tests added for the new REST endpoint (missing param, external URL rejection, success, unauthorized access)

1. On **trunk** (before this PR), navigate to **Newspack > Components Demo**, scroll to Handoff Buttons, and click any handoff button — confirm the return banner does **not** appear (verifying the pre-existing breakage).
2. Switch to this branch and rebuild. Navigate to **Newspack > Components Demo** and scroll to the Handoff Buttons section.
3. Click "Go to Dashboard" — you should be redirected to the Newspack Dashboard with a blue return banner at the top reading "Return to Components Demo once you're done."
4. Verify the banner shows a "Dismiss" button and a "Back to Components Demo" button. Click Back — you should return to Components Demo.
5. Repeat step 3, click "Dismiss" — the banner should disappear without navigating away.
6. Scroll to the Action Cards section and find the "Handoff with URL" card. Click "Go to Dashboard" and verify the same banner behavior as above.
7. Navigate to a Newspack wizard page (e.g., Settings). Trigger a handoff programmatically or via the demo and verify the static PHP-rendered banner appears at the top of the wizard page.
8. Test an existing plugin-based handoff (e.g., Jetpack card) — confirm it still works and now accepts custom `bannerText`/`bannerButtonText` when passed.
9. After a handoff to a non-Newspack page, navigate to an unrelated admin page (e.g., Posts) — the banner should clear and no longer be visible.
10. After a handoff, navigate back to a Newspack page — the banner should also be gone.
11. Verify the banner is correctly offset on pages with the WordPress admin sidebar (no overflow or misalignment).
12. On a WooCommerce admin page, trigger a handoff and verify the banner adjusts for the WooCommerce header offset.

### Other information:

### DIFF
--- a/includes/api/class-plugins-controller.php
+++ b/includes/api/class-plugins-controller.php
@@ -156,6 +156,31 @@ class Plugins_Controller extends WP_REST_Controller {
 					'methods'             => 'POST',
 					'callback'            => [ $this, 'handoff_to_url' ],
 					'permission_callback' => [ $this, 'handoff_item_permissions_check' ],
+					'args'                => [
+						'destinationUrl'    => [
+							'required'          => true,
+							'type'              => 'string',
+							'sanitize_callback' => 'esc_url_raw',
+							'validate_callback' => [ $this, 'validate_same_site_url' ],
+						],
+						'handoffReturnUrl'  => [
+							'type'              => 'string',
+							'sanitize_callback' => 'esc_url_raw',
+							'validate_callback' => [ $this, 'validate_same_site_url' ],
+						],
+						'bannerText'        => [
+							'type'              => 'string',
+							'sanitize_callback' => 'sanitize_text_field',
+						],
+						'bannerButtonText'  => [
+							'type'              => 'string',
+							'sanitize_callback' => 'sanitize_text_field',
+						],
+						'showOnBlockEditor' => [
+							'type'    => 'boolean',
+							'default' => false,
+						],
+					],
 				],
 			]
 		);
@@ -339,15 +364,6 @@ class Plugins_Controller extends WP_REST_Controller {
 			return new \WP_Error( 'newspack_handoff_missing_url', __( 'destinationUrl is required.', 'newspack-plugin' ), [ 'status' => 400 ] );
 		}
 
-		// Reject external URLs to prevent open-redirect attacks.
-		$parsed_destination = wp_parse_url( $destination_url );
-		if ( ! empty( $parsed_destination['host'] ) ) {
-			$site_host = wp_parse_url( admin_url(), PHP_URL_HOST );
-			if ( $parsed_destination['host'] !== $site_host ) {
-				return new \WP_Error( 'newspack_handoff_invalid_url', __( 'destinationUrl must be a same-site URL.', 'newspack-plugin' ), [ 'status' => 400 ] );
-			}
-		}
-
 		$handoff_return_url   = $request->get_param( 'handoffReturnUrl' );
 		$show_on_block_editor = $request->get_param( 'showOnBlockEditor' );
 		$banner_text          = (string) $request->get_param( 'bannerText' );
@@ -355,10 +371,10 @@ class Plugins_Controller extends WP_REST_Controller {
 
 		update_option( NEWSPACK_HANDOFF, 'url' );
 		update_option( NEWSPACK_HANDOFF_SHOW_ON_BLOCK_EDITOR, (bool) $show_on_block_editor );
-		update_option( NEWSPACK_HANDOFF_BANNER_TEXT, sanitize_text_field( $banner_text ) );
-		update_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, sanitize_text_field( $banner_button_text ) );
+		update_option( NEWSPACK_HANDOFF_BANNER_TEXT, $banner_text );
+		update_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, $banner_button_text );
 		if ( ! empty( $handoff_return_url ) ) {
-			update_option( NEWSPACK_HANDOFF_RETURN_URL, esc_url( $handoff_return_url ) );
+			update_option( NEWSPACK_HANDOFF_RETURN_URL, $handoff_return_url );
 		}
 
 		$parsed_url = wp_parse_url( $destination_url );
@@ -369,7 +385,42 @@ class Plugins_Controller extends WP_REST_Controller {
 			}
 		}
 
-		return rest_ensure_response( [ 'HandoffLink' => esc_url_raw( $destination_url ) ] );
+		return rest_ensure_response( [ 'HandoffLink' => $destination_url ] );
+	}
+
+	/**
+	 * Validate that a URL is empty, relative, or a same-site http(s) URL.
+	 *
+	 * Runs before sanitize_callback, so the raw input is checked (e.g. `javascript:` is
+	 * caught here before `esc_url_raw` would silently strip the scheme).
+	 *
+	 * @param mixed            $value   Value being validated.
+	 * @param \WP_REST_Request $request Request object.
+	 * @param string           $param   Parameter name.
+	 * @return true|WP_Error True if valid, WP_Error otherwise.
+	 */
+	public function validate_same_site_url( $value, $request, $param ) {
+		if ( empty( $value ) ) {
+			return true;
+		}
+		$parsed = wp_parse_url( $value );
+		if ( false === $parsed ) {
+			// translators: %s is the parameter name.
+			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s must be a valid URL.', 'newspack-plugin' ), $param ), [ 'status' => 400 ] );
+		}
+		$scheme = isset( $parsed['scheme'] ) ? strtolower( $parsed['scheme'] ) : '';
+		if ( $scheme && ! in_array( $scheme, [ 'http', 'https' ], true ) ) {
+			// translators: %s is the parameter name.
+			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s must be an http(s) URL.', 'newspack-plugin' ), $param ), [ 'status' => 400 ] );
+		}
+		if ( ! empty( $parsed['host'] ) ) {
+			$site_host = wp_parse_url( admin_url(), PHP_URL_HOST );
+			if ( $parsed['host'] !== $site_host ) {
+				// translators: %s is the parameter name.
+				return new WP_Error( 'rest_invalid_param', sprintf( __( '%s must be a same-site URL.', 'newspack-plugin' ), $param ), [ 'status' => 400 ] );
+			}
+		}
+		return true;
 	}
 
 	/**
--- a/includes/api/class-plugins-controller.php
+++ b/includes/api/class-plugins-controller.php
@@ -339,6 +339,15 @@ class Plugins_Controller extends WP_REST_Controller {
 			return new \WP_Error( 'newspack_handoff_missing_url', __( 'destinationUrl is required.', 'newspack-plugin' ), [ 'status' => 400 ] );
 		}
 
+		// Reject external URLs to prevent open-redirect attacks.
+		$parsed_destination = wp_parse_url( $destination_url );
+		if ( ! empty( $parsed_destination['host'] ) ) {
+			$site_host = wp_parse_url( admin_url(), PHP_URL_HOST );
+			if ( $parsed_destination['host'] !== $site_host ) {
+				return new \WP_Error( 'newspack_handoff_invalid_url', __( 'destinationUrl must be a same-site URL.', 'newspack-plugin' ), [ 'status' => 400 ] );
+			}
+		}
+
 		$handoff_return_url   = $request->get_param( 'handoffReturnUrl' );
 		$show_on_block_editor = $request->get_param( 'showOnBlockEditor' );
 		$banner_text          = (string) $request->get_param( 'bannerText' );
--- a/includes/api/class-plugins-controller.php
+++ b/includes/api/class-plugins-controller.php
@@ -359,11 +359,7 @@ class Plugins_Controller extends WP_REST_Controller {
 	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
 	 */
 	public function handoff_to_url( $request ) {
-		$destination_url = $request->get_param( 'destinationUrl' );
-		if ( empty( $destination_url ) ) {
-			return new \WP_Error( 'newspack_handoff_missing_url', __( 'destinationUrl is required.', 'newspack-plugin' ), [ 'status' => 400 ] );
-		}
-
+		$destination_url      = $request->get_param( 'destinationUrl' );
 		$handoff_return_url   = $request->get_param( 'handoffReturnUrl' );
 		$show_on_block_editor = $request->get_param( 'showOnBlockEditor' );
 		$banner_text          = (string) $request->get_param( 'bannerText' );
--- a/includes/api/class-plugins-controller.php
+++ b/includes/api/class-plugins-controller.php
@@ -147,6 +147,19 @@ class Plugins_Controller extends WP_REST_Controller {
 			]
 		);
 
+		// Register newspack/v1/handoff endpoint for URL-based handoff.
+		register_rest_route(
+			$this->namespace,
+			'/handoff',
+			[
+				[
+					'methods'             => 'POST',
+					'callback'            => [ $this, 'handoff_to_url' ],
+					'permission_callback' => [ $this, 'handoff_item_permissions_check' ],
+				],
+			]
+		);
+
 		// Register newspack/v1/plugins/some-plugin/handoff endpoint.
 		register_rest_route(
 			$this->namespace,
@@ -315,6 +328,42 @@ class Plugins_Controller extends WP_REST_Controller {
 	}
 
 	/**
+	 * Handoff to an arbitrary admin URL.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function handoff_to_url( $request ) {
+		$destination_url = $request->get_param( 'destinationUrl' );
+		if ( empty( $destination_url ) ) {
+			return new \WP_Error( 'newspack_handoff_missing_url', __( 'destinationUrl is required.', 'newspack-plugin' ), [ 'status' => 400 ] );
+		}
+
+		$handoff_return_url   = $request->get_param( 'handoffReturnUrl' );
+		$show_on_block_editor = $request->get_param( 'showOnBlockEditor' );
+		$banner_text          = (string) $request->get_param( 'bannerText' );
+		$banner_button_text   = (string) $request->get_param( 'bannerButtonText' );
+
+		update_option( NEWSPACK_HANDOFF, 'url' );
+		update_option( NEWSPACK_HANDOFF_SHOW_ON_BLOCK_EDITOR, (bool) $show_on_block_editor );
+		update_option( NEWSPACK_HANDOFF_BANNER_TEXT, sanitize_text_field( $banner_text ) );
+		update_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, sanitize_text_field( $banner_button_text ) );
+		if ( ! empty( $handoff_return_url ) ) {
+			update_option( NEWSPACK_HANDOFF_RETURN_URL, esc_url( $handoff_return_url ) );
+		}
+
+		$parsed_url = wp_parse_url( $destination_url );
+		if ( ! empty( $parsed_url['query'] ) ) {
+			wp_parse_str( $parsed_url['query'], $query_params );
+			if ( ! empty( $query_params['page'] ) ) {
+				update_option( NEWSPACK_HANDOFF_DESTINATION_PAGE, sanitize_text_field( $query_params['page'] ) );
+			}
+		}
+
+		return rest_ensure_response( [ 'HandoffLink' => esc_url_raw( $destination_url ) ] );
+	}
+
+	/**
 	 * Handoff to a managed plugin.
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
@@ -329,7 +378,9 @@ class Plugins_Controller extends WP_REST_Controller {
 		}
 
 		$show_on_block_editor = $request->get_param( 'showOnBlockEditor' );
-		Handoff_Banner::register_handoff_for_plugin( $slug, (bool) $show_on_block_editor );
+		$banner_text          = (string) $request->get_param( 'bannerText' );
+		$banner_button_text   = (string) $request->get_param( 'bannerButtonText' );
+		Handoff_Banner::register_handoff_for_plugin( $slug, (bool) $show_on_block_editor, $banner_text, $banner_button_text );
 		$managed_plugins = Plugin_Manager::get_managed_plugins();
 
 		$response           = $managed_plugins[ $slug ];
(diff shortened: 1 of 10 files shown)


includes/class-handoff-banner.php
```diff
@@ -80,18 +147,24 @@ public function enqueue_styles() {
...
+			[ 'wp-components' ],
 			NEWSPACK_PLUGIN_VERSION
 		);
 		wp_enqueue_style( $handle );
 
+		$screen = get_current_screen();
+		if ( $screen && stristr( $screen->id, 'newspack' ) ) {
+			return;
+		}
+
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 9b68782.

includes/class-handoff-banner.php
```diff
@@ -139,8 +212,16 @@ public static function needs_block_editor_handoff_return_ui() {
...
+			// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			if ( $destination_page && isset( $_GET['page'] ) && sanitize_text_field( wp_unslash( $_GET['page'] ) ) === $destination_page ) {
+				return;
+			}
 			update_option( NEWSPACK_HANDOFF, null );
 			update_option( NEWSPACK_HANDOFF_SHOW_ON_BLOCK_EDITOR, false );
+			update_option( NEWSPACK_HANDOFF_BANNER_TEXT, '' );
+			update_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, '' );
+			update_option( NEWSPACK_HANDOFF_DESTINATION_PAGE, '' );
 		}
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 9b68782.

packages/components/src/handoff/index.js
```diff
@@ -92,30 +113,47 @@ class Handoff extends Component {
...
 			<Fragment>
-				{ Name && 'active' === Status && (
+				{ url && (
+					<Button
+						className={ classes }
+						isSecondary={ ! otherProps.isPrimary && ! otherProps.isTertiary && ! otherProps.isLink }
+						{ ...otherProps }
+						onClick={ () => ( useModal ? this.setState( { showModal: true } ) : goTo() ) }
+					>
+						{ children }
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 9b68782.

includes/api/class-plugins-controller.php
```diff
@@ -147,6 +147,19 @@ public function register_routes() {
...
+			$this->namespace,
+			'/handoff',
+			[
+				[
+					'methods'             => 'POST',
+					'callback'            => [ $this, 'handoff_to_url' ],
+					'permission_callback' => [ $this, 'handoff_item_permissions_check' ],
+				],
+			]
+		);
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 9b68782.

includes/class-handoff-banner.php
```diff
@@ -102,11 +175,15 @@ public function enqueue_styles() {
...
+	 * @param  string  $banner_text Custom banner body text.
+	 * @param  string  $banner_button_text Custom banner button text.
 	 * @return void
 	 */
-	public static function register_handoff_for_plugin( $plugin, $show_on_block_editor = false ) {
+	public static function register_handoff_for_plugin( $plugin, $show_on_block_editor = false, $banner_text = '', $banner_button_text = '' ) {
 		update_option( NEWSPACK_HANDOFF, $plugin );
 		update_option( NEWSPACK_HANDOFF_SHOW_ON_BLOCK_EDITOR, (bool) $show_on_block_editor );
+		update_option( NEWSPACK_HANDOFF_BANNER_TEXT, sanitize_text_field( $banner_text ) );
+		update_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, sanitize_text_field( $banner_button_text ) );
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 9b68782.

includes/api/class-plugins-controller.php
```diff
@@ -314,6 +327,42 @@ public function uninstall_item( $request ) {
...
+		$parsed_url = wp_parse_url( $destination_url );
+		if ( ! empty( $parsed_url['query'] ) ) {
+			wp_parse_str( $parsed_url['query'], $query_params );
+			if ( ! empty( $query_params['page'] ) ) {
+				update_option( NEWSPACK_HANDOFF_DESTINATION_PAGE, sanitize_text_field( $query_params['page'] ) );
+			}
+		}
+
+		return rest_ensure_response( [ 'HandoffLink' => esc_url_raw( $destination_url ) ] );
+	}
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 9b68782.

includes/api/class-plugins-controller.php
```diff
@@ -314,6 +327,51 @@ public function uninstall_item( $request ) {
...
+	 */
+	public function handoff_to_url( $request ) {
+		$destination_url = $request->get_param( 'destinationUrl' );
+		if ( empty( $destination_url ) ) {
+			return new \WP_Error( 'newspack_handoff_missing_url', __( 'destinationUrl is required.', 'newspack-plugin' ), [ 'status' => 400 ] );
+		}
+
+		// Reject external URLs to prevent open-redirect attacks.
+		$parsed_destination = wp_parse_url( $destination_url );
+		if ( ! empty( $parsed_destination['host'] ) ) {
```
adekbadek: **Blocker: same-site check accepts non-http(s) schemes.**

`wp_parse_url( 'javascript:alert(1)' )` returns no `host`, so this guard is skipped entirely. The input is accidentally neutralized later by `esc_url_raw()` stripping the unsafe scheme (returns `''`), but that's latent — a refactor could easily regress, and the response still returns 200 with the options already written.

Suggested fix: explicitly allowlist `http`/`https` (or empty scheme with a path starting with `/wp-admin/`):

```php
$scheme = isset( $parsed_destination['scheme'] ) ? strtolower( $parsed_destination['scheme'] ) : '';
if ( $scheme && ! in_array( $scheme, [ 'http', 'https' ], true ) ) {
    return new \WP_Error( 'newspack_handoff_invalid_url', __( 'destinationUrl must be an http(s) URL.', 'newspack-plugin' ), [ 'status' => 400 ] );
}
```

includes/api/class-plugins-controller.php
```diff
@@ -314,6 +327,51 @@ public function uninstall_item( $request ) {
...
+		$show_on_block_editor = $request->get_param( 'showOnBlockEditor' );
+		$banner_text          = (string) $request->get_param( 'bannerText' );
+		$banner_button_text   = (string) $request->get_param( 'bannerButtonText' );
+
+		update_option( NEWSPACK_HANDOFF, 'url' );
+		update_option( NEWSPACK_HANDOFF_SHOW_ON_BLOCK_EDITOR, (bool) $show_on_block_editor );
+		update_option( NEWSPACK_HANDOFF_BANNER_TEXT, sanitize_text_field( $banner_text ) );
+		update_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, sanitize_text_field( $banner_button_text ) );
+		if ( ! empty( $handoff_return_url ) ) {
+			update_option( NEWSPACK_HANDOFF_RETURN_URL, esc_url( $handoff_return_url ) );
```
adekbadek: **Blocker: `handoffReturnUrl` is not validated as same-site.**

Unlike `destinationUrl`, the return URL is only `esc_url()`'d before being stored. A caller could point the "Back to X" button off-site. Please apply the same same-site check (or `wp_validate_redirect()`) here.

Also minor: use `esc_url_raw()` rather than `esc_url()` when writing to the DB — `esc_url()` is for output contexts and encodes HTML entities.

includes/api/class-plugins-controller.php
```diff
@@ -147,6 +147,19 @@ public function register_routes() {
 			]
 		);
 
+		// Register newspack/v1/handoff endpoint for URL-based handoff.
+		register_rest_route(
+			$this->namespace,
+			'/handoff',
+			[
+				[
+					'methods'             => 'POST',
```
adekbadek, line 156: **Suggestion:** the new `/handoff` route declares no `args` schema. Idiomatic WordPress is to use `args` with `sanitize_callback` / `validate_callback` for `destinationUrl`, `handoffReturnUrl`, `bannerText`, `bannerButtonText`, and `showOnBlockEditor`. That's the natural place to enforce the same-site check from the two comments above and would let you drop the manual validation from the callback.

src/wizards/handoff-banner/index.js
```diff
@@ -41,10 +41,43 @@ const HandoffBanner = ( {
...
+		}
+	}
+
+	const { primary_button_url: primaryButtonURL, banner_text: bodyText, banner_button_text: primaryButtonText } = el.dataset;
+	render(
+		createElement( HandoffBanner, {
+			primaryButtonURL,
+			...( bodyText && { bodyText } ),
+			...( primaryButtonText && { primaryButtonText } ),
+		} ),
```
adekbadek, line 82: **Blocker: this `render()` clobbers the server-rendered static banner on wizard pages.**

The PHP `insert_handoff_banner_static()` on `newspack_before_wizard_content` emits a `<div id="newspack-handoff-banner">` containing full HTML with the *custom* `bannerText` / `bannerButtonText`. But `enqueue_styles()` also enqueues this script on wizard pages, so it runs, finds the same element, reads `el.dataset.banner_text` (undefined — the static markup has no `data-*` attributes), and `render()`s a fresh `<HandoffBanner />` into the element — **wiping the server output and replacing it with the React component's default fallback strings** (`Return to Newspack after completing configuration` / `Back to Newspack`). The custom banner text never reaches the user, and `primaryButtonURL` is also undefined so the Back button doesn't actually navigate.

A few options:

1. Skip enqueuing `handoff-banner.js` on wizard pages (preferred — the static banner is self-contained).
2. Give the static banner a distinct id (`#newspack-handoff-banner-static`) so the two don't collide.
3. Bail out of this script if `el` already has non-empty children.

I'd also add a test (or just a manual pass) that asserts the custom `bannerText` actually renders on a wizard page after a handoff.

includes/class-handoff-banner.php
```diff
@@ -132,15 +207,52 @@ public static function needs_block_editor_handoff_return_ui() {
...
-			update_option( NEWSPACK_HANDOFF, null );
-			update_option( NEWSPACK_HANDOFF_SHOW_ON_BLOCK_EDITOR, false );
+			$this->clear_all_handoff_options();
+			return;
+		}
+
+		// For URL-based handoffs with a known destination page, also clear on any
+		// non-Newspack page that isn't the destination, to prevent the banner from
+		// persisting across unrelated admin screens.
+		if ( $destination_page ) {
```
adekbadek: **Blocker: this wipes handoff state before the user reaches the destination.**

After `wp-login.php` submits, WordPress redirects to `/wp-admin/` (screen id `dashboard`, no `page` query param). With a URL-based handoff queued up, `$destination_page` is set and `$current_page` is empty, so this branch fires and calls `clear_all_handoff_options()`. Any handoff registered before a re-auth is gone — the banner never renders.

The same thing happens on any intermediate redirect between the REST call and the destination page (e.g. a first-run wizard that bounces through another slug), since those pages also won't match `$destination_page`.

The issue is that this branch assumes "non-Newspack page means user has finished with the handoff," but that's not true before the destination has been reached for the first time. Two ways out:

1. Track a `visited` flag: set it in the early-return at line 242 (when we're on the destination), and only enter this "clear on unrelated page" branch if visited is true.
2. Clear only when we're on the `NEWSPACK_HANDOFF_RETURN_URL` page (i.e. the user actually used the Back button), not on any random admin page.

Option 1 is probably the simpler minimal fix.

includes/class-handoff-banner.php
```diff
@@ -60,9 +131,11 @@ public function insert_block_editor_handoff_banner() {
 			true
 		);
 
-		$script_info = [
-			'text'       => __( 'Return to Newspack after completing configuration', 'newspack' ),
-			'buttonText' => __( 'Back to Newspack', 'newspack' ),
+		$banner_text        = get_option( NEWSPACK_HANDOFF_BANNER_TEXT, '' );
+		$banner_button_text = get_option( NEWSPACK_HANDOFF_BANNER_BUTTON_TEXT, '' );
+		$script_info        = [
+			'text'       => $banner_text ? $banner_text : __( 'Return to Newspack after completing configuration', 'newspack' ),
```
adekbadek: **Suggestion:** these two `__()` calls still use the `'newspack'` text domain, which isn't a registered domain for this plugin (the static banner just above uses `'newspack-plugin'` correctly). These strings will never be translated. Since you're touching these lines, worth fixing to `'newspack-plugin'`.

includes/class-handoff-banner.php
```diff
@@ -132,15 +207,52 @@ public static function needs_block_editor_handoff_return_ui() {
 	}
 
 	/**
-	 * If the current admin page is part of the Newspack dashboard, clear the handoff URL. This ensures the handoff banner won't be shown on Newspack admin pages.
+	 * Clear all handoff-related options.
+	 *
+	 * @return void
+	 */
+	private function clear_all_handoff_options() {
+		update_option( NEWSPACK_HANDOFF, null );
```
adekbadek: **Suggestion:** `update_option( NEWSPACK_HANDOFF, null )` stores an empty string rather than removing the row, which is subtly fragile (depends on cache-layer null handling). Prefer `delete_option( NEWSPACK_HANDOFF )` for a clean "no handoff" state. Minor, but since `needs_handoff_return_ui()` does a truthy check, the intent is "unset," so `delete_option()` matches the semantics.

tests/unit-tests/api-plugins-controller.php
```diff
@@ -96,6 +96,60 @@ public function test_get_plugins_authorized() {
...
+		$request  = new WP_REST_Request( 'POST', $this->api_namespace . '/handoff' );
+		$response = $this->server->dispatch( $request );
+		$this->assertEquals( 400, $response->get_status() );
+		$this->assertEquals( 'newspack_handoff_missing_url', $response->get_data()['code'] );
+	}
+
+	/**
+	 * Test handoff to URL rejects external URLs.
+	 */
+	public function test_handoff_to_url_rejects_external_url() {
```
adekbadek, line 112: **Suggestion: add a `javascript:` rejection case here.**

Given the blocker on `class-plugins-controller.php:344`, this is the right test to assert the fix. Something like:

```php
public function test_handoff_to_url_rejects_javascript_scheme() {
    wp_set_current_user( $this->administrator );
    $request = new WP_REST_Request( 'POST', $this->api_namespace . '/handoff' );
    $request->set_param( 'destinationUrl', 'javascript:alert(1)' );
    $response = $this->server->dispatch( $request );
    $this->assertEquals( 400, $response->get_status() );
}
```

Also worth adding: protocol-relative rejection (`//evil.example/x`), and a `handoffReturnUrl` same-site case once that validation is added.

adekbadek (changes requested):
Thanks @thomasguillot — the redesign and the fix for the broken-on-trunk banner are great additions. A few things before merge: two security items in the URL validation, and two correctness bugs in the banner lifecycle that I believe prevent the URL-based handoff from showing the custom text or the right return link in the real flow. Details inline.

includes/api/class-plugins-controller.php
thomasguillot: Addressed in 6c33b5a — validation moved to the route's `args` schema with an explicit http(s) allowlist in `validate_same_site_url()`.

includes/api/class-plugins-controller.php
thomasguillot: Addressed in 6c33b5a — `handoffReturnUrl` now runs through the same `validate_same_site_url` validator and uses `esc_url_raw` as its `sanitize_callback`.

includes/api/class-plugins-controller.php
thomasguillot: Addressed in 6c33b5a — route now declares an `args` schema for all five params with `sanitize_callback` / `validate_callback`, and the callback was slimmed down accordingly.

src/wizards/handoff-banner/index.js
thomasguillot: Addressed in 7615d71 — went with option 1 (skip the JS enqueue on Newspack screens). The static banner is self-contained and all Newspack admin pages go through `Wizard::render_wizard()`, so the `newspack_before_wizard_content` hook always fires. CSS is still enqueued so the static markup remains styled.

includes/class-handoff-banner.php
thomasguillot: Addressed in eb07b29 — went with option 2 (clear only when the user reaches the return URL).

The reasoning: a handoff banner is a "you're on a detour, here's the way back" UI. The user is sent to a destination to make configuration changes, which often involves navigating sub-pages of the destination plugin (e.g. WooCommerce → Settings → Payments → Stripe → Webhooks). Option 1 (visited flag + clear on navigation away) would clear the banner the moment the user leaves the exact destination page, even though they're still mid-task. Option 2 keeps the banner sticky throughout the detour and only clears when the user explicitly returns via the stored return URL.

This also aligns URL-based handoffs with how plugin handoffs already work in practice — plugin handoffs use `stristr($screen->id, 'newspack')` as the "back on Newspack turf" clear trigger, which is the same spirit as "clear on return URL".

Login redirects and intermediate bounces are now transparently survived. If the user abandons the task entirely, they can dismiss the banner manually.

As a side-effect cleanup: `clear_all_handoff_options()` now also deletes `NEWSPACK_HANDOFF_RETURN_URL`, which was never being cleared before — a latent issue that would have been more visible under option 2.

includes/class-handoff-banner.php
thomasguillot: Addressed in 3619230 — changed both to `'newspack-plugin'`.

includes/class-handoff-banner.php
thomasguillot: Addressed in 3619230 — `clear_all_handoff_options()` now uses `delete_option()` for every handoff-related option, including `NEWSPACK_HANDOFF_RETURN_URL` which was previously never cleared.

tests/unit-tests/api-plugins-controller.php
thomasguillot: Addressed in f411fca — added tests for `javascript:` scheme, protocol-relative URL, and external `handoffReturnUrl`. Existing tests updated to account for the new `args`-schema error codes.

includes/api/class-plugins-controller.php
```diff
@@ -314,6 +352,77 @@ public function uninstall_item( $request ) {
...
 
+	/**
+	 * Handoff to an arbitrary admin URL.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function handoff_to_url( $request ) {
+		$destination_url = $request->get_param( 'destinationUrl' );
+		if ( empty( $destination_url ) ) {
```
adekbadek: Nit: dead code now that `destinationUrl` is declared `required => true` with a `validate_callback` in the args schema. The REST layer rejects the call before this runs. Worth removing (or adding a comment) to avoid future confusion about which layer does the validation.

includes/class-handoff-banner.php
```diff
@@ -38,7 +43,73 @@ public function insert_handoff_banner() {
...
+		}
+		?>
+		<div id="newspack-handoff-banner">
+			<div class="newspack-handoff-banner">
+				<div class="newspack-handoff-banner__text"><?php echo esc_html( $banner_text ); ?></div>
+				<div class="newspack-handoff-banner__buttons">
+					<button
+						type="button"
+						class="components-button is-tertiary is-small"
+						onclick="this.closest('#newspack-handoff-banner').remove()"
```
adekbadek: Nit: inline `onclick="this.closest(...).remove()"` trips strict CSP policies. You already have an inline `<script>` block a few lines below doing the margin/width math – binding the dismiss handler there (`document.querySelector('#newspack-handoff-banner [data-dismiss]').addEventListener('click', …)`) would keep everything CSP-friendly.

packages/components/src/handoff/index.js
```diff
@@ -92,30 +113,47 @@ class Handoff extends Component {
...
 			<Fragment>
-				{ Name && 'active' === Status && (
+				{ url && (
+					<Button
+						className={ classes }
+						isSecondary={ ! otherProps.isPrimary && ! otherProps.isTertiary && ! otherProps.isLink }
+						{ ...otherProps }
+						onClick={ () => ( useModal ? this.setState( { showModal: true } ) : goTo() ) }
+					>
+						{ children ? children : primaryButton }
```
adekbadek, line 138: Edge case: if a caller passes `url` + `useModal` without `children`, the modal opens with empty strings from `textForPlugin({})` (no plugin info in the URL path). Either disallow `useModal` with `url`, or require `children` / an explicit button label when `url` is set. Not currently exercised by any caller, but worth documenting or guarding.

src/wizards/handoff-banner/index.js
```diff
@@ -41,10 +41,43 @@ const HandoffBanner = ( {
...
+			}
+			return false;
+		};
+		if ( ! applyWooCommerceOffset() ) {
+			const observer = new MutationObserver( () => {
+				if ( applyWooCommerceOffset() ) {
+					observer.disconnect();
+				}
+			} );
+			observer.observe( wpbody, { attributes: true, attributeFilter: [ 'style' ] } );
```
adekbadek, line 72: Potential leak: if the WooCommerce header never appears (which is the common case on non-Woo admin pages), the `MutationObserver` is never disconnected and keeps listening for the life of the page. Adding a timeout fallback would bound it:

```js
const timeoutId = setTimeout( () => observer.disconnect(), 5000 );
const observer = new MutationObserver( () => {
	if ( applyWooCommerceOffset() ) {
		clearTimeout( timeoutId );
		observer.disconnect();
	}
} );
```

adekbadek (approved):
Tested end-to-end. All prior-round feedback is resolved and browser testing confirms the new behavior works cleanly:

- ✅ URL handoff renders custom banner text on destination (Newspack Dashboard)
- ✅ Back button returns to origin
- ✅ Dismiss removes the banner
- ✅ Banner stays sticky across navigation to unrelated admin pages (Posts)
- ✅ Banner auto-clears when the user lands on the stored return URL
- ✅ No PHP warnings/notices during the run

Approving. Leaving a few non-blocking nits inline – feel free to address or defer.

includes/api/class-plugins-controller.php
thomasguillot: Addressed in 7a47a4d.

includes/class-handoff-banner.php
thomasguillot: Addressed in 7a47a4d.

packages/components/src/handoff/index.js
thomasguillot: Addressed in 7a47a4d.

src/wizards/handoff-banner/index.js
thomasguillot: Addressed in 7a47a4d.

PR #4604, opened 2026-03-27
feat: reader activation segments

miguelpeixe:
### All Submissions:

Introduces `segments` to the `newspackReaderActivation` API. Segments can be registered and have the match set for publication purposes. The authority on segmentation and matching logic remains `newspack-popups`.

This can also serve as the initial step and foundation when moving campaign logic from `newspack-popups` to here, which has been a long-standing conversation.

Closes NPPD-1391.

1. Checkout this branch and https://github.com/Automattic/newspack-popups/pull/1539
2. Make sure you have the standard audience management setup for campaigns
3. Visit the frontend and open the console
4. Verify `window.newspackReaderActivation.segments.getAll()` returns the registered segments with name, criteria, and priority fields
5. Verify `window.newspackReaderActivation.segments.getMatch()` returns the matched segment object (or null if no match)
6. Test event subscription: run `window.newspackReaderActivation.on('segment', e => console.log(e.detail))`, then reload the page. The event should fire with segmentId, segment, and all in the event detail object

### Other information:


src/reader-activation/segments.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Get the matched segment.
+ *
+ * @return {Object|null} Matched segment object with id, or null.
+ */
+function getMatch() {
+	if ( ! matchedSegment || ! allSegments[ matchedSegment ] ) {
+		return matchedSegment ? { id: matchedSegment } : null;
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in b5e12f740.

src/reader-activation/segments.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @return {Object|null} Matched segment object or null.
+ */
+function setMatch( segmentId = null ) {
+	if ( segmentId === matchedSegment ) {
+		return getMatch();
+	}
+	matchedSegment = segmentId;
+	const segment = matchedSegment ? allSegments[ matchedSegment ] || null : null;
+	emit( EVENTS.segment, { segmentId: matchedSegment, segment, all: { ...allSegments } } );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in b5e12f740.

src/reader-activation/segments.test.js
```diff
@@ -0,0 +1,51 @@
...
+		const detail = callback.mock.calls[ 0 ][ 0 ].detail;
+		expect( detail.segmentId ).toBe( '42' );
+		expect( detail.segment.name ).toBe( 'Loyal Readers' );
+	} );
+	it( 'should not re-emit when setting same match', () => {
+		const callback = jest.fn();
+		on( 'segment', callback );
+		segments.setMatch( '42' );
+		expect( callback ).not.toHaveBeenCalled();
+	} );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in b5e12f740.

src/reader-activation/segments.test.js
```diff
@@ -0,0 +1,51 @@
...
+	it( 'should return matched segment via getMatch', () => {
+		const match = segments.getMatch();
+		expect( match.id ).toBe( '42' );
+		expect( match.name ).toBe( 'Loyal Readers' );
+		expect( match.priority ).toBe( 0 );
+	} );
+	it( 'should clear match and emit event', () => {
+		const callback = jest.fn();
+		on( 'segment', callback );
+		segments.setMatch( null );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in b5e12f740.

src/reader-activation/segments.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Reset module state. For testing only.
+ */
+function reset() {
+	allSegments = {};
+	matchedSegment = null;
+}
+
+export default {
```
adekbadek, line 69: nit: `reset()` ships as part of the public API (`window.newspackReaderActivation.segments.reset()`) but is documented as "For testing only." The `overlays` module doesn't expose an equivalent. Worth a conscious decision – either accept it as public or keep it internal to tests.

src/reader-activation/segments.test.js
```diff
@@ -0,0 +1,66 @@
...
+		expect( all[ '42' ].name ).toBe( 'Loyal Readers' );
+		expect( all[ '43' ].name ).toBe( 'New Visitors' );
+	} );
+	it( 'should set match and emit segment event', () => {
+		const callback = jest.fn();
+		on( 'segment', callback );
+		segments.register( sampleSegments );
+		segments.setMatch( '42' );
+		expect( callback ).toHaveBeenCalled();
+		const detail = callback.mock.calls[ 0 ][ 0 ].detail;
```
adekbadek, line 31: nit: The PR description calls out `all` as part of the event detail contract, but the test doesn't assert on it. A quick `expect( detail.all ).toEqual( expect.objectContaining( { '42': expect.any( Object ) } ) )` would lock that in.

src/reader-activation/segments.js
```diff
@@ -0,0 +1,71 @@
...
+ * Set the matched segment for the current reader.
+ *
+ * @param {string|null} segmentId Segment ID or null to clear.
+ *
+ * @return {Object|null} Matched segment object or null.
+ */
+function setMatch( segmentId = null ) {
+	const normalizedId = segmentId !== null && segmentId !== undefined ? String( segmentId ) : null;
+	if ( normalizedId === matchedSegment ) {
+		return getMatch();
```
adekbadek, line 32: nit: If `setMatch('999')` is called before the segment is registered, the internal state becomes `'999'` but `getMatch()` returns `null`. A later `register({ '999': {...} })` would silently make `getMatch()` start returning a value without re-emitting the event. Unlikely to happen in practice since popups always registers first – just flagging for awareness.

adekbadek (approved):
Clean, well-scoped addition. The `segments` sub-module follows the existing `overlays` pattern nicely, tests cover the core flows, and the cross-repo contract is backward-compatible. All "How to test" steps verified in an isolated environment.

A few minor notes inline – nothing blocking.

src/reader-activation/segments.js
miguelpeixe: Addressed in c6e3971b3 — removed `reset()` from the default export so it's no longer part of the public API on `window.newspackReaderActivation.segments`. It's still available as a named export for tests.

src/reader-activation/segments.test.js
miguelpeixe: Addressed in c6e3971b3 — added `expect( detail.all ).toEqual( expect.objectContaining( { 42: expect.any( Object ) } ) )` to lock in the contract.

src/reader-activation/segments.js
miguelpeixe: Addressed in c6e3971b3 — `register()` now re-emits the segment event if there's a pending match that resolves. Added a test for this edge case.

PR #4606, opened 2026-03-30
fix(memberships): force non-empty message to allow wc_memberships_notice_html to fire

dkoo:
### All Submissions:

Closes NPPM-2707.

When all WC Memberships message templates are set to empty strings (which is the case when Newspack content gates replace them), the `get_message_html()` method in WC Memberships skips `get_notice_html()` entirely. This means the `wc_memberships_notice_html` filter never fires, so Newspack's content gate is never rendered. The result is that restricted content is truncated but no gate is shown to the reader.

This fix adds a `wc_memberships_restricted_message` filter that returns a non-breaking space (`&nbsp;`) when the message is empty and a Newspack gate is configured. This ensures `get_notice_html()` is called, which fires `wc_memberships_notice_html`, and the gate renders correctly. The placeholder is fully replaced by Newspack's gate HTML in `wc_memberships_notice_html()`.

The filter has three guard clauses to limit its scope:
- Only intervenes when the message is empty (leaves publisher-configured messages alone)
- Only applies to `content_` message codes (not product discounts etc.)
- Only activates when a Newspack gate is configured

1. Ensure WC Memberships 1.28.1 is installed and active with a membership plan that restricts all posts.
2. Set the WC Memberships message templates to empty strings (this happens automatically when Newspack content gates are configured, or set manually via `update_option('wc_memberships_messages', ['content_restricted_message' => '', ...])`)
3. Configure a Newspack content gate (Audience > Access Control > create a gate restricting all posts with registered access).
4. Log in as a subscriber/reader who does NOT have an active membership in the restricting plan.
5. Visit a restricted post.

**Expected on `release` (bug):** Content is truncated but no gate appears below it.
**Expected on this branch (fix):** Content is truncated and the memberships gate is rendered below.

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-memberships.php
+++ b/includes/plugins/wc-memberships/class-memberships.php
@@ -48,6 +48,7 @@ class Memberships {
 
 		// WC Memberships hooks.
 		add_filter( 'wc_memberships_notice_html', [ __CLASS__, 'wc_memberships_notice_html' ], 100, 4 );
+		add_filter( 'wc_memberships_restricted_message', [ __CLASS__, 'wc_memberships_restricted_message' ], 100, 4 );
 		add_filter( 'wc_memberships_restricted_content_excerpt', [ __CLASS__, 'wc_memberships_excerpt' ], 100, 3 );
 		add_filter( 'wc_memberships_message_excerpt_apply_the_content_filter', '__return_false' );
 		add_filter( 'wc_memberships_admin_screen_ids', [ __CLASS__, 'admin_screens' ] );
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Tested locally against WC Memberships 1.28.1. Confirmed the bug on `release` (content truncated, no gate) and verified the fix on this branch (gate renders correctly).

The approach is sound – the three guard clauses (empty message only, content restriction codes only, Newspack gate present) keep the scope tight. The `&nbsp;` placeholder is fully replaced by Newspack's gate HTML in `wc_memberships_notice_html()`, so it never leaks to the frontend.

One minor note: on product pages, the `is_product()` check in `wc_memberships_notice_html` would return the notice unchanged, but the `str_contains($message_code, 'content_')` guard effectively prevents this since product restriction codes don't contain `content_`.

PR #4613, opened 2026-03-30
feat(access-control): advanced settings for RSS content restriction

dkoo:
### All Submissions:

Adds an Advanced Settings modal in the Access Control wizard page to contain options that don't fit into any other UI views or components. Implements the following options:

- **Restrict content in feeds** - Truncate restricted content in RSS feeds. Enabled by default.
- ~**Hide Everlit in restricted content** - Prevents Everlit audio embed blocks from rendering in restricted content. Enabled by default.~ Note: we've removed this option as this should be handled by the Everlit plugin, not here.

Closes NPPD-1035.

1. In **Audience > Access Control**, confirm that the "Advanced settings" item in the dropdown menu next to the page title is no longer disabled:

Screenshot 2026-03-31 at 3 35 24 PM

2. Click it and confirm it shows the new modal:

Screenshot 2026-03-31 at 2 55 55 PM

3. Toggle off the option, save, refresh, and confirm that the saved settings persist.
4. Test the "undo" link in the snackbar after saving and confirm that it undoes the most recent save.
5. Leave the option toggled off and save.
6. Publish a content gate that applies to all posts with Registered and Paid Access enabled + metering.
7. Visit your RSS feed (`https://sitedomain/feed`) and confirm that all restricted articles show their full content.
8. Toggle on the option and save.
9. Repeat steps 7-8 and confirm that the feed shows truncated content for restricted posts unless logged into a user account that has full access

### Other information:


includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -162,6 +162,25 @@ public function register_api_endpoints() {
...
+					'advanced_settings' => [
+						'type'       => 'object',
+						'properties' => [
+							'restrict_feeds'   => [ 'type' => 'boolean' ],
+							'restrict_everlit' => [ 'type' => 'boolean' ],
+						],
+					],
+				],
+			]
+		);
```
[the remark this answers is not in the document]
dkoo: Updated in bf55bb730

includes/content-gate/class-content-gate-settings.php
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	/**
+	 * Update the advanced settings.
+	 *
+	 * @param array $settings The advanced settings.
+	 */
+	public static function update_settings( $settings ) {
+		update_option( self::OPTION_PREFIX . 'restrict_feeds', boolval( $settings['restrict_feeds'] ) ? 1 : 0, false );
+		update_option( self::OPTION_PREFIX . 'restrict_everlit', boolval( $settings['restrict_everlit'] ) ? 1 : 0, false );
```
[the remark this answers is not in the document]
dkoo: Updated in bf55bb730

includes/content-gate/class-content-gate-settings.php
```diff
@@ -0,0 +1,153 @@
...
+		$settings = [
+			'restrict_feeds' => get_option( self::OPTION_PREFIX . 'restrict_feeds', 1 ),
+		];
+
+		// Everlit.
+		$everlit = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'everlit' );
+		if ( $everlit->is_active() ) {
+			$settings['restrict_everlit'] = get_option( self::OPTION_PREFIX . 'restrict_everlit', 1 );
+		}
+		return $settings;
```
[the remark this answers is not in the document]
dkoo: Updated in bf55bb730

src/wizards/audience/views/content-gates/advanced-settings.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useWizardApiFetch } from '../../../hooks/use-wizard-api-fetch';
+import { AUDIENCE_CONTENT_GATES_WIZARD_SLUG } from './consts';
+
+const AdvancedSettings = ( { closeModal, showModal }: { closeModal: () => void; showModal: boolean } ) => {
+	const wizardData = useWizardData( AUDIENCE_CONTENT_GATES_WIZARD_SLUG ) as WizardData;
+	const { wizardApiFetch, isFetching, resetError, setError } = useWizardApiFetch( AUDIENCE_CONTENT_GATES_WIZARD_SLUG );
+	const { addNotice, resetNotices, updateWizardSettings } = useDispatch( WIZARD_STORE_NAMESPACE );
+	const [ config, setConfig ] = useState< AdvancedSettingsConfig >(
+		( wizardData?.config?.advanced_settings as AdvancedSettingsConfig ) || { auto_signup: true }
+	);
```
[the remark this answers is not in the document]
dkoo: Updated in bf55bb730

src/wizards/audience/views/content-gates/advanced-settings.tsx
```diff
@@ -0,0 +1,108 @@
...
+						/>
+					) }
+				</VStack>
+				<HStack justify="end">
+					<Button variant="tertiary" disabled={ isFetching } onClick={ closeModal }>
+						{ __( 'Cancel', 'newspack-plugin' ) }
+					</Button>
+					<Button
+						variant="primary"
+						disabled={ isFetching || JSON.stringify( wizardData?.config || {} ) === JSON.stringify( config ) }
```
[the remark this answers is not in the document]
dkoo: Updated in bf55bb730

thomasguillot:
@dkoo — Thomas has reviewed these changes and will be marking this PR as design approved.

Changes pushed in the latest commit:
- Moved modal action buttons inside the `VStack` for correct layout flow
- Switched Institutions menu item from `href: '#/institutions'` to `history.push('/institutions')` via the router proxy, for consistent SPA navigation
- Fixed `AdvancedSettingsConfig` type — `restrict_content_feeds` was a typo; renamed to `restrict_feeds` to match the PHP and component code

includes/content-gate/class-content-gate-settings.php
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Main class.
+ */
+class Content_Gate_Settings {
```
miguelpeixe: Not to be confused with the component at `src/wizards/audience/views/content-gates/content-gate-settings.tsx`

Non-blocking suggestion: we could make this class just about the RSS feed handling. It is already by implementing the feed restriction on top of the settings, just need to make it official with the docblocks and class name 😄

I can see a year from now, this class having 1000+ lines of small features and patches mixed together...

tests/unit-tests/content-gate/test-advanced-settings.php
```diff
@@ -0,0 +1,208 @@
...
+
+namespace Newspack\Tests\Content_Gate;
+
+use Newspack\Content_Gate;
+use Newspack\Content_Gate_Settings;
+
+/**
+ * Tests for RSS feed content restriction.
+ */
+class Test_Advanced_Settings extends \WP_UnitTestCase {
```
miguelpeixe, line 16: Also non-blocking, but I think we can have this either for RSS Feed restriction control alone or just additional tests within `tests/unit-tests/content-gate/content-gates.php`. "Advanced settings" is just a UI layer.

src/wizards/audience/views/content-gates/advanced-settings.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useWizardData } from '../../../../../packages/components/src/wizard/store/utils';
+import { WIZARD_STORE_NAMESPACE } from '../../../../../packages/components/src/wizard/store';
+import { useWizardApiFetch } from '../../../hooks/use-wizard-api-fetch';
+import { AUDIENCE_CONTENT_GATES_WIZARD_SLUG } from './consts';
+
+const AdvancedSettings = ( { closeModal, showModal }: { closeModal: () => void; showModal: boolean } ) => {
+	const wizardData = useWizardData( AUDIENCE_CONTENT_GATES_WIZARD_SLUG ) as WizardData;
+	const { wizardApiFetch, isFetching, resetError, setError } = useWizardApiFetch( AUDIENCE_CONTENT_GATES_WIZARD_SLUG );
+	const { addNotice, resetNotices, updateWizardSettings } = useDispatch( WIZARD_STORE_NAMESPACE );
+	const [ config, setConfig ] = useState< AdvancedSettingsConfig >( ( wizardData?.config?.advanced_settings as AdvancedSettingsConfig ) || {} );
```
miguelpeixe: The `AdvancedSettings` component is always mounted, so changing config, closing, and re-opening the modal will show stale data. I don't think we need to introduce a dirty state confirmation modal, but might be a good idea to reset on close here.

miguelpeixe (changes requested):
Tested well! Left a couple of non-blocking suggestions and a small issue.

includes/content-gate/class-content-gate-settings.php
dkoo: I see your point, but this class is kind of where I envision we'll add additional "global" settings beyond just this RSS option, hence why it originally also owned the Everlit option. What if we renamed it to `Content_Gate_Advanced_Settings` or `Content_Gate_Global_Settings`?

miguelpeixe: I think `Content_Gate_Advanced_Settings` is better.

src/wizards/audience/views/content-gates/advanced-settings.tsx
dkoo: The component always being mounted is intentional, as this is what allows the Undo button in the Snackbar component that appears after saving to function (otherwise the prior state in the ref is lost when the component unmounts). The internal `config` state is updated after a successful save in tandem with the wizard store, using the same data, so it's not likely to get out of sync. If you're worried about this being possibly too fragile we could do away with the "Undo" button so the component can be unmounted when the modal isn't visible. But so far I haven't run into any trouble with the component state and store data getting out of sync. 🤷

includes/content-gate/class-content-gate-settings.php
dkoo: 00d01c40a renames the class to `Content_Gate_Global_Settings` (and the test class too) to hopefully make the intention a bit clearer.

tests/unit-tests/content-gate/test-advanced-settings.php
dkoo: 00d01c40a renames this to `Test_Global_Settings` to match the renamed class it tests.

dkoo:
@miguelpeixe thanks for the feedback! Happy to discuss further if you feel the changes aren't enough

src/wizards/audience/views/content-gates/advanced-settings.tsx
miguelpeixe: I see... I had a different UX perception for this. When I open something in a modal, make changes, and close without clicking the Save button, I'd not expect that state to be preserved. I'd also not expect that closed modal state to be persisted if I click save somewhere else.

dkoo: That's a valid point. b748566e2 fixes that issue, so that the local state is reset upon closing the modal. Also applies the same fix to the Premium Newsletters advanced settings modal, which uses the same component pattern (maybe we could combine these into one reusable component in the future).

miguelpeixe: If I save the new value, opening the modal again will render the state from page load.

dkoo: I'm not able to replicate this, but does 8e5ed756b fix it for you?

miguelpeixe: Not really. It's stuck:

https://github.com/user-attachments/assets/3560a05d-e24a-4ad9-b346-20564bb84cde

miguelpeixe: Actually, it must've been object cache stuck again. Started working again after a manual purge of the option... It's working fine now.

miguelpeixe (approved):

includes/content-gate/class-content-gate-settings.php
dkoo: 73fc24431 renames it to `Content_Gate_Advanced_Settings` (sorry, hadn't seen the latest response)

miguelpeixe: No worries! Thanks for revising 😄

PR #4614, opened 2026-03-30
fix(indesign): exclude rich media blocks from export output

wil-gerken:
### All Submissions:

When a post contains a `core/file` block (PDF embed) or `core/embed` block (YouTube, etc.), the InDesign export was including raw HTML markup in the output — such as `<object>` tags, download link text, and bare embed URLs — instead of skipping those blocks entirely. This resulted in import errors and corrupted print output for publishers using the InDesign export tool.

Rich media blocks have no print equivalent and should be excluded from the export. This fix adds an explicit exclusion list in `process_blocks()` for `core/file`, `core/embed`, `core/video`, and `core/audio`.

These blocks are now skipped before serialization, ensuring that only printable content is exported while leaving surrounding content intact.

Closes `NPPM-2677`.

**Reproduce the problem (without this PR):**

1. Enable the InDesign export optional module: `wp newspack optional-modules enable indesign-export`
2. Create a post containing a paragraph, a PDF file block (`core/file`), and another paragraph after it.
3. From the Posts list, click **Export as Adobe InDesign** in the row actions for that post.
4. Open the downloaded `.txt` file — observe that `<object>` tags, the PDF filename, and "Download" text appear between the two paragraphs in the export output.

**Verify the fix (with this PR):**

5. Export the same post again.
6. Open the downloaded `.txt` file — the two paragraphs should appear cleanly with no `<object>` tag, PDF filename, or "Download" text between them.
7. Repeat with a post containing a YouTube embed block (`core/embed`) — the YouTube URL should not appear in the export.
8. Test a post with only standard content (paragraphs, headings, blockquotes, lists) — verify all content is present and unchanged.

### Other information:

### DIFF
--- a/includes/optional-modules/indesign-export/class-indesign-converter.php
+++ b/includes/optional-modules/indesign-export/class-indesign-converter.php
@@ -190,10 +190,13 @@ class InDesign_Converter {
 		// the InDesign output. Strip recursively so nested occurrences inside
 		// container blocks (core/group, core/columns, etc.) are also removed.
 		// Publishers can extend this list via the filter for custom block types.
-		$excluded_block_types = apply_filters(
+		// Normalize the filter result to an array of strings in case a callback
+		// returns a non-array or mixed-type value.
+		$excluded_block_types = (array) apply_filters(
 			'newspack_indesign_export_excluded_blocks',
 			self::EXCLUDED_BLOCK_TYPES
 		);
+		$excluded_block_types = array_values( array_filter( $excluded_block_types, 'is_string' ) );
 
 		$blocks  = $this->strip_excluded_blocks( parse_blocks( $content ), $excluded_block_types );
 		$content = '';
@@ -256,11 +259,12 @@ class InDesign_Converter {
 	/**
 	 * Check whether a block name should be excluded from export.
 	 *
-	 * Handles both the explicit exclusion list and legacy core-embed/* block
-	 * names used in content created before WordPress 5.6.
+	 * Legacy core-embed/* block names (pre-WP 5.6) follow the same exclusion
+	 * state as core/embed — if core/embed is in the filtered list, its legacy
+	 * variants are excluded too.
 	 *
 	 * @param string   $block_name           Block type name.
-	 * @param string[] $excluded_block_types Explicit list of excluded block types.
+	 * @param string[] $excluded_block_types Filtered list of excluded block types.
 	 *
 	 * @return bool True if the block should be excluded.
 	 */
@@ -272,8 +276,11 @@ class InDesign_Converter {
 		if ( in_array( $block_name, $excluded_block_types, true ) ) {
 			return true;
 		}
-		// Legacy embed block names (pre-WP 5.6) use the core-embed/* namespace.
-		if ( 0 === strpos( $block_name, 'core-embed/' ) ) {
+		// Legacy core-embed/* variants follow core/embed's exclusion state.
+		if (
+			in_array( 'core/embed', $excluded_block_types, true )
+			&& 0 === strpos( $block_name, 'core-embed/' )
+		) {
 			return true;
 		}
 		return false;
--- a/includes/optional-modules/indesign-export/class-indesign-converter.php
+++ b/includes/optional-modules/indesign-export/class-indesign-converter.php
@@ -172,9 +172,22 @@ class InDesign_Converter {
 	 * @return string Content with processed blocks.
 	 */
 	private function process_blocks( $content ) {
-		$blocks = parse_blocks( $content );
+		// Rich media blocks have no print equivalent. Exclude them entirely to
+		// prevent raw HTML (e.g. <object> tags, embed URLs) from leaking into
+		// the InDesign output.
+		$excluded_block_types = [
+			'core/file',
+			'core/embed',
+			'core/video',
+			'core/audio',
+		];
+
+		$blocks  = parse_blocks( $content );
 		$content = '';
 		foreach ( $blocks as $block ) {
+			if ( in_array( $block['blockName'], $excluded_block_types, true ) ) {
+				continue;
+			}
 			$tag = $this->get_block_tag( $block );
 			if ( ! empty( $tag ) ) {
 				$content .= $tag . $this->get_transformed_text( preg_replace( '/^<[^>]+>(.*)<\/[^>]+>$/s', '$1', trim( $block['innerHTML'] ) ) );
--- a/includes/optional-modules/indesign-export/class-indesign-converter.php
+++ b/includes/optional-modules/indesign-export/class-indesign-converter.php
@@ -15,6 +15,19 @@ defined( 'ABSPATH' ) || exit;
 class InDesign_Converter {
 
 	/**
+	 * Block types with no print equivalent, excluded from InDesign export by default.
+	 * Filterable via the newspack_indesign_export_excluded_blocks filter.
+	 *
+	 * @var string[]
+	 */
+	const EXCLUDED_BLOCK_TYPES = [
+		'core/file',
+		'core/embed',
+		'core/video',
+		'core/audio',
+	];
+
+	/**
 	 * Default InDesign styles configuration.
 	 *
 	 * @var array
@@ -176,12 +189,11 @@ class InDesign_Converter {
 		// prevent raw HTML (e.g. <object> tags, embed URLs) from leaking into
 		// the InDesign output. Strip recursively so nested occurrences inside
 		// container blocks (core/group, core/columns, etc.) are also removed.
-		$excluded_block_types = [
-			'core/file',
-			'core/embed',
-			'core/video',
-			'core/audio',
-		];
+		// Publishers can extend this list via the filter for custom block types.
+		$excluded_block_types = apply_filters(
+			'newspack_indesign_export_excluded_blocks',
+			self::EXCLUDED_BLOCK_TYPES
+		);
 
 		$blocks  = $this->strip_excluded_blocks( parse_blocks( $content ), $excluded_block_types );
 		$content = '';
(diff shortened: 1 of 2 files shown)


includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -172,9 +172,22 @@ private function process_post_content( $content, $options = [] ) {
...
+			'core/video',
+			'core/audio',
+		];
+
+		$blocks  = parse_blocks( $content );
 		$content = '';
 		foreach ( $blocks as $block ) {
+			if ( in_array( $block['blockName'], $excluded_block_types, true ) ) {
+				continue;
+			}
```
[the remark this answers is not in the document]
wil-gerken: Thanks and great call. Addressed in the following commit.

tests/unit-tests/indesign-exporter/indesign-exporter.php
```diff
@@ -278,6 +278,53 @@ public function test_convert_horizontal_rule() {
...
+		$converter = new InDesign_Converter();
+		$content   = $converter->convert_post( $post_id );
+
+		$this->assertStringContainsString( 'Before the embed.', $content );
+		$this->assertStringContainsString( 'After the embed.', $content );
+		$this->assertStringNotContainsString( 'youtube.com', $content );
+		$this->assertStringNotContainsString( 'abc123', $content );
+	}
+
 	/**
```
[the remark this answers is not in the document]
wil-gerken: Thanks and done! Both fail against the original code and pass with the recursive fix.

includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -172,7 +172,18 @@ private function process_post_content( $content, $options = [] ) {
...
-		$blocks = parse_blocks( $content );
+		// Rich media blocks have no print equivalent. Exclude them entirely to
+		// prevent raw HTML (e.g. <object> tags, embed URLs) from leaking into
+		// the InDesign output. Strip recursively so nested occurrences inside
+		// container blocks (core/group, core/columns, etc.) are also removed.
+		$excluded_block_types = [
+			'core/file',
+			'core/embed',
+			'core/video',
+			'core/audio',
```
adekbadek: Suggestion: expose this as a filter (e.g. `newspack_indesign_export_excluded_blocks`) so publishers with custom rich-media blocks can extend the list without patching. Also worth promoting to a `private const` since it's rebuilt on every `process_blocks()` call.

Minor related note: legacy `core-embed/*` block names (pre-WP 5.6 content) won't match `core/embed` – a filter sidesteps that, or a `strpos( $blockName, 'core-embed/' ) === 0` check would catch them. Low priority, depends on how old your content is.

includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -185,6 +196,47 @@ private function process_blocks( $content ) {
...
+			}
+			if ( ! empty( $block['innerBlocks'] ) ) {
+				$new_inner_blocks  = [];
+				$new_inner_content = [];
+				$inner_index       = 0;
+				foreach ( $block['innerContent'] as $chunk ) {
+					if ( is_string( $chunk ) ) {
+						$new_inner_content[] = $chunk;
+					} else {
+						$inner_block = $block['innerBlocks'][ $inner_index++ ];
```
adekbadek, line 244: Defensive nit: `$block['innerBlocks'][ $inner_index++ ]` assumes the `parse_blocks()` invariant that null placeholders in `innerContent` line up positionally with `innerBlocks` entries. That's safe in practice, but a `! isset(...)` guard would harden it against malformed input. Take it or leave it.

tests/unit-tests/indesign-exporter/indesign-exporter.php
```diff
@@ -278,6 +278,94 @@ public function test_convert_horizontal_rule() {
...
+			]
+		);
+
+		$converter = new InDesign_Converter();
+		$content   = $converter->convert_post( $post_id );
+
+		$this->assertStringContainsString( 'Before the group.', $content );
+		$this->assertStringContainsString( 'After the group.', $content );
+		$this->assertStringNotContainsString( 'youtube.com', $content );
+		$this->assertStringNotContainsString( 'abc123', $content );
```
adekbadek, line 366: Coverage gap worth filling: the implementation also strips `core/video` and `core/audio` and the docblock calls out `core/columns` as a supported container, but neither is exercised here. I verified at runtime that all three work, but suggest adding cases for:

1. `core/video` and/or `core/audio` exclusion (locks in the full list).
2. An embed nested in `core/columns` (different `innerContent` shape than `core/group`).
3. Two excluded siblings in a row inside a container, to exercise the `$inner_index` increment path.

adekbadek (approved):
Tested in an isolated env by running `InDesign_Converter::convert_post()` against six fixture posts: file/embed/video/audio at top level, file inside a `core/group`, embed inside `core/columns`, and a plain control. All exclusions worked as expected and the surrounding content was preserved in every case. No PHP notices in `debug.log`.

The fix is correct and the recursion handles nested containers as advertised. Approving – a few non-blocking suggestions inline. The main one is that the unit tests only cover `core/file`/`core/embed` nested in `core/group`, but the implementation (and docblock) supports `core/video`, `core/audio`, and `core/columns` too. I verified those work at runtime, but it would be nice to lock them in with tests.

includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -172,7 +185,17 @@ private function process_post_content( $content, $options = [] ) {
...
+		// prevent raw HTML (e.g. <object> tags, embed URLs) from leaking into
+		// the InDesign output. Strip recursively so nested occurrences inside
+		// container blocks (core/group, core/columns, etc.) are also removed.
+		// Publishers can extend this list via the filter for custom block types.
+		$excluded_block_types = apply_filters(
+			'newspack_indesign_export_excluded_blocks',
+			self::EXCLUDED_BLOCK_TYPES
+		);
+
+		$blocks  = $this->strip_excluded_blocks( parse_blocks( $content ), $excluded_block_types );
```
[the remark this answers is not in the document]
wil-gerken: Great catch, thanks. The filter result is now normalized to an array of strings: `(array)` cast handles non-array returns (null, string, etc.) gracefully, and `array_filter( ..., 'is_string' )` drops any non-string entries before the array is used. 

Added a test (`test_filter_returning_non_array_does_not_break_export`) that registers a filter returning `null` and confirms the export still works without a TypeError.

includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -185,6 +208,77 @@ private function process_blocks( $content ) {
...
+		// parse_blocks() returns null blockName for freeform/whitespace chunks.
+		if ( ! is_string( $block_name ) || '' === $block_name ) {
+			return false;
+		}
+		if ( in_array( $block_name, $excluded_block_types, true ) ) {
+			return true;
+		}
+		// Legacy embed block names (pre-WP 5.6) use the core-embed/* namespace.
+		if ( 0 === strpos( $block_name, 'core-embed/' ) ) {
+			return true;
```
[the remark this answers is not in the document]
wil-gerken: Good catch on the inconsistency. Went with the second option and tied the legacy check to the filter state: `core-embed/*` blocks are now excluded only when `core/embed` is in the filtered list. That way a publisher who removes `core/embed` from the filter also implicitly un-blocks legacy variants, which is almost certainly what they'd want since they're semantically the same block. 

Updated the docblock to describe the new behavior and added a test (`test_legacy_core_embed_follows_core_embed_filter`) that verifies this.

includes/optional-modules/indesign-export/class-indesign-converter.php
wil-gerken: Thanks! This is a great suggestion. 

* I've promoted the array to an `EXCLUDED_BLOCK_TYPES` class constant. I left it as unscoped `const` rather than `private const` so the default list remains accessible if needed, but I'm happy to change if you think it should be private.

* I added the `newspack_indesign_export_excluded_blocks` filter so publishers can extend the list for custom block types. The filter result is cast to an array and filtered to strings, so a misbehaving callback won't break the export.

* I added an `is_excluded_block()` helper which handles the explicit list and legacy `core-embed/*` names via a `strpos` prefix check. Legacy variants are tied to `core/embed`'s filter state, so if a publisher removes `core/embed` from the filter, the legacy variants are also un-blocked for consistency.

* Last, I added a null guard in that helper since `parse_blocks()` returns `blockName === null` for freeform chunks. I have a pre-push review command that caught that as a PHP 8.3 deprecation.

I don't think this fix is urgent, and the changes were extensive enough that it seems best to re-review. Can you please give it another look when you have a chance?

includes/optional-modules/indesign-export/class-indesign-converter.php
wil-gerken: Thanks! Taking it / added. I like safeguards like this much better as well.

tests/unit-tests/indesign-exporter/indesign-exporter.php
wil-gerken: Great suggestions. I added all three: 
* `core/video` and `core/audio` top-level exclusion
* an embed nested in `core/columns`
* two excluded siblings in a row to exercise the `$inner_index` increment path.

I also added a test for legacy `core-embed/youtube` blocks while I was in there, to lock in the prefix check from the other thread. 

All 27 InDesign tests pass.

includes/optional-modules/indesign-export/class-indesign-converter.php
adekbadek: Implementation looks great. Unscoped `const` is fine – keeps defaults accessible. Filter normalization with `(array)` + `array_filter( ..., 'is_string' )` is robust, and tying legacy `core-embed/*` to `core/embed`'s filter state is the right call. Good catch on the null blockName guard for freeform chunks. LGTM.

includes/optional-modules/indesign-export/class-indesign-converter.php
adekbadek: Thanks!

tests/unit-tests/indesign-exporter/indesign-exporter.php
adekbadek: All four cases covered, and the bonus legacy embed test is a nice add. LGTM.

adekbadek (approved):
Re-reviewed after the latest round of changes. All three of my prior comments are addressed cleanly and the new tests cover the regressions well. Ran the full InDesign test suite locally – 27 tests, 78 assertions, all green.

PR #4618, opened 2026-04-01
feat: session hydration

miguelpeixe:
### All Submissions:

This PR introduces a mechanism to bridge the gap between authentication and the next page load. After login, a short-lived transient binds the `newspack-cid` cookie to the authenticated user ID, and a new REST endpoint (`/reader/session`) returns a `wp_rest` nonce for subsequent authenticated API calls.

This session hydration strategy is used by the reader data store and the newsletter sign-up modal to offer premium newsletters to the signed-in user.

The data store will no longer require a page refresh to sync pending data items, and the newsletter modal refresh endpoint will rely on authentication, simplifying its logic.

How it works:

1. In the backend: any authentication (via Woo checkout registration or auth/registration modal), bind the session's `newspack-cid` to the `$user_id` for **2 minutes**.
2. In the frontend: detect the auth cookie and request the `/reader/session` endpoint to fetch a `wp_rest` nonce
3. Once hydrated (nonce fetched), a triggered event gets picked up by the data store and newsletter modal, so they run their logic

The `/reader/session` response payload has a filter so components can initialize their state without additional requests. The reader data store leverages that filter to populate stored reader data items for db -> localStorage hydration.

#### Simple authentication flow

1. In a fresh session, open devtools to inspect network requests
2. Sign in as an existing reader using the auth modal
3. Confirm that after authenticating, a request is made to `/reader/session` with the following response:

```json
{
    "nonce": "365493a533",
    "reader_data_items": {
        "is_donor": "false",
        "is_former_donor": "true",
        "pageviews": "{\"day\":{\"count\":22,\"start\":1775053485659},\"week\":{\"count\":22,\"start\":1775053485659},\"month\":{\"count\":22,\"start\":1775053485659}}",
        "is_newsletter_subscriber": "true",
        "newsletter_subscribed_lists": "[\"5\"]",
        "active_subscriptions": "[20157]",
        "article_view_per_week": "{\"1772247600000\":{\"3836\":true},\"1774062000000\":{\"4837\":true},\"1774666800000\":{\"4837\":true,\"8597\":true,\"20826\":true}}",
        "article_view_per_month": "{\"1772334000000\":{\"3836\":true},\"1775012400000\":{\"4837\":true,\"8597\":true,\"20826\":true}}",
    }
}
```

#### Reader data store hydration

1. In the same session, confirm that the reader data items from the response above are now hydrated to localStorage
2. Also confirm a successful request is made to `/reader-data` with the nonce as `X-WP-Nonce` header, syncing local items to the db

#### Premium newsletters

1. Set up premium newsletters attached to a subscription product
2. Make sure to check the premium lists in the "Present newsletter signup after checkout and registration" section of audience configuration
3. In a fresh session, open the console and run `newspackReaderActivation.openNewslettersSignupModal()`
4. Confirm the modal doesn't include the premium list(s) and close
5. Go through the modal checkout flow to purchase the subscription that unlocks the lists
6. Confirm that the post-checkout newsletters modal includes the list(s)

### Other information:


src/reader-activation/session.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+export function hydrateSession() {
+	if ( ! pending ) {
+		pending = fetchSession().then( data => {
+			if ( data?.nonce ) {
+				window.newspack_reader_data.nonce = data.nonce;
+				emit( EVENTS.session, data );
+			}
+			return data?.nonce || null;
+		} );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in bbec72d44 — always initialize `window.newspack_reader_data` unconditionally.

src/reader-activation/store.js
```diff
@@ -256,6 +258,29 @@ export default function Store() {
...
+		const items = detail?.reader_data_items || {};
+		newspack_reader_data.items = items;
+		if ( ! newspack_reader_data?.is_temporary ) {
+			const unsyncedKeys = _get( 'unsynced', true ) || [];
+			for ( const key of Object.keys( items ) ) {
+				if ( ! unsyncedKeys.includes( key ) ) {
+					_set( key, JSON.parse( items[ key ] ) );
+				}
+			}
+		}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in bbec72d44 — replaced `JSON.parse()` with the existing `decode()` helper in both hydration paths.

src/reader-activation/session.js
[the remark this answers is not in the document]
miguelpeixe: Addressed in bbec72d44 — `pending` now resets to `null` on failure so hydration can be retried.

tests/unit-tests/session-hydration.php
```diff
@@ -0,0 +1,180 @@
...
+		$_COOKIE[ NEWSPACK_CLIENT_ID_COOKIE_NAME ] = self::$test_cid; // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+
+		Session_Hydration::bind_cid( $user_id );
+
+		$stored_user_id = get_transient( 'newspack_cid_' . self::$test_cid );
+		$this->assertEquals( $user_id, $stored_user_id );
+
+		unset( $_COOKIE[ NEWSPACK_CLIENT_ID_COOKIE_NAME ] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+		wp_delete_user( $user_id );
+	}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 3dc4ce4ae — added `tear_down()` that cleans up cookies and transients after each test.

src/reader-activation-newsletters/newsletters-modal.js
```diff
@@ -11,39 +13,47 @@ export function getModalContainer() {
...
+	const headers = {};
+	if ( nonce ) {
+		headers[ 'X-WP-Nonce' ] = nonce;
+	}
+
+	const res = await fetch( '/wp-json/newspack/v1/reader-newsletter-signup-lists', {
+		credentials: 'same-origin',
+		headers,
+	} );
+	const { html } = await res.json();
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 3dc4ce4ae — added try/catch and `res.ok` check.

includes/reader-activation/class-session-hydration.php
```diff
@@ -0,0 +1,169 @@
...
+	 *
+	 * @param int $user_id The authenticated user's ID.
+	 */
+	public static function bind_cid( $user_id ) {
+		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+		$cid = isset( $_COOKIE[ NEWSPACK_CLIENT_ID_COOKIE_NAME ] ) ? sanitize_text_field( $_COOKIE[ NEWSPACK_CLIENT_ID_COOKIE_NAME ] ) : '';
+		if ( empty( $cid ) ) {
+			return;
+		}
+		set_transient( self::TRANSIENT_PREFIX . $cid, (int) $user_id, self::TRANSIENT_TTL );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 3dc4ce4ae — CID is now hashed with `md5()` before use in transient key.

includes/reader-activation/class-reader-activation.php
```diff
@@ -1814,8 +1772,8 @@ public static function render_newsletters_signup_modal( $email_address = '' ) {
 		if ( ! self::is_newsletters_signup_available() ) {
 			return;
 		}
-		if ( ! is_email( $email_address ) ) {
-			$email_address = self::get_logged_in_reader_email_address();
+		if ( ! is_email( $email_address ) && \is_user_logged_in() && self::is_user_reader( \wp_get_current_user() ) ) {
```
dkoo, line 1776: After purchasing a subscription as an anonymous reader without logging in, when the post-checkout newsletter modal appears and fires the API request to refresh the list, at this line I'm still seeing `is_user_logged_in()` return `false`. This causes the lists to always be fetched for an anonymous reader instead of the reader account that was created during checkout.

Here's a writeup from Claude about the potential cause + fix:

> The answer is about a cookie and timing gap between two systems.
> 
> ---
> 
> Here's the chain that breaks:
> 
> **1. `refreshNewslettersSignupModal` is called immediately on checkout completion**
> 
> In `newspack-blocks/src/modal-checkout/modal.js:197-198`, `onCheckoutComplete` fires and directly calls:
> ```js
> window.newspackReaderActivation.refreshNewslettersSignupModal( email );
> ```
> This fires immediately when the checkout iframe completes — before any hydration.
> 
> **2. `getApiNonce()` returns `null` at that moment**
> 
> After this PR, `refreshNewslettersSignupModal` no longer uses the passed `email` argument. Instead it calls `getApiNonce()` to get an auth nonce. But hydration (`hydrateSession()`) hasn't run yet, so the nonce is `null` and the request goes out without `X-WP-Nonce`.
> 
> **3. Why hasn't `hydrateSession()` run yet?**
> 
> `hydrateSession()` is triggered by `attachAuthCookiesListener` (in `index.js:374`), which polls for the `np_auth_reader` cookie:
> ```js
> const authCookie = getCookie( 'np_auth_reader' );
> if ( authCookie ) { hydrateSession(); }
> ```
> 
> `np_auth_reader` is set by `login_set_reader_cookie()` — which is hooked to `wp_login`. But WooCommerce checkout registration calls `wc_set_customer_auth_cookie()` directly, which calls `wp_set_auth_cookie` **without** going through `wp_signon()`, so the `wp_login` hook never fires. `np_auth_reader` is never set. The listener never triggers.
> 
> **4. The result**
> 
> `/reader-newsletter-signup-lists` is called without `X-WP-Nonce`. WordPress REST API won't authenticate a user via cookies alone — it requires a nonce for cookie-based auth. So `is_user_logged_in()` is `false`.
> 
> The transient created by `on_woo_customer_created` is correct and `/reader/session` *would* authenticate the user just fine — but the frontend never calls it because `np_auth_reader` is the missing trigger.
> 
> ---
> 
> **The fix** likely needs to happen in one (or both) of two places:
> 
> - In `Session_Hydration::on_woo_customer_created`, also call `Reader_Activation::set_auth_reader_cookie( get_userdata( $user_id ) )` so the `np_auth_reader` cookie gets set during checkout, giving the listener something to detect
> - Or in `newspack-blocks/modal.js`, use `hydrateSession()` directly on checkout completion and wait for the returned promise before calling `refreshNewslettersSignupModal`, rather than calling `refreshNewslettersSignupModal` immediately

tests/unit-tests/session-hydration.php
```diff
@@ -0,0 +1,190 @@
...
+	private static $test_cid = 'testcid12345';
+
+	/**
+	 * Clean up after each test.
+	 */
+	public function tear_down() {
+		parent::tear_down();
+		unset( $_COOKIE[ NEWSPACK_CLIENT_ID_COOKIE_NAME ] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+		unset( $_COOKIE[ LOGGED_IN_COOKIE ] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+		delete_transient( Session_Hydration::TRANSIENT_PREFIX . self::$test_cid );
```
dkoo: This should be MD5 hashed, right?

```suggestion
		delete_transient( Session_Hydration::get_transient_key( self::$test_cid ) );
```

includes/reader-activation/class-session-hydration.php
```diff
@@ -0,0 +1,188 @@
...
+		if ( false === $stored_user_id || (int) $stored_user_id !== \get_current_user_id() ) {
+			return new \WP_Error(
+				'newspack_session_invalid',
+				__( 'Invalid session.', 'newspack-plugin' ),
+				[ 'status' => 403 ]
+			);
+		}
+
+		// One-time use: delete the transient.
+		delete_transient( $transient_key );
```
dkoo: Highlighted by Claude:

> **Transient deleted before filter callbacks run**
> `class-session-hydration.php:165-183` — `delete_transient()` fires before `apply_filters('newspack_session_hydration_response')`. If a filter callback throws after deletion, the transient is burned but the nonce never reaches the client. Since the JS resets `pending = null` on catch (allowing retry), the user would be locked out until re-authentication. Consider deleting the transient after assembling the full response, or move deletion to a `shutdown` action.

includes/reader-activation/class-session-hydration.php
```diff
@@ -0,0 +1,188 @@
...
+			return;
+		}
+		set_transient( self::get_transient_key( $cid ), (int) $user_id, self::TRANSIENT_TTL );
+	}
+
+	/**
+	 * Register REST routes.
+	 */
+	public static function register_routes() {
+		\register_rest_route(
```
dkoo, line 105: Highlighted by Claude:

> **REST route not gated by `Reader_Activation::is_enabled()`**
> `class-session-hydration.php:35-39` — `Reader_Activation::register_routes()` is only hooked when `self::is_enabled()` is true (class-reader-activation.php:97-98). The new `/reader/session` endpoint registers unconditionally, creating an inconsistency and unnecessary attack surface when the feature is disabled.

src/reader-activation/session.js
```diff
@@ -0,0 +1,75 @@
...
+ * @return {Promise<Object|null>} The response data, or null on failure.
+ */
+async function fetchSession() {
+	const cid = getCookie( 'newspack-cid' );
+	if ( ! cid ) {
+		return null;
+	}
+
+	try {
+		const response = await fetch( '/wp-json/newspack/v1/reader/session', {
```
dkoo: (Non-blocking) highlighted by Claude:

> **Hardcoded REST URL breaks non-root installs**
> `session.js:65, newsletters-modal.js:35` — `/wp-json/newspack/v1/reader/session` is hardcoded. Breaks subdirectory installs or custom REST prefix. The backend already localizes `newspack_reader_data` for the script handle — the session URL should be localized alongside `api_url` rather than hardcoded. (Pre-existing issue in `newsletters-modal.js`, but new code repeats the pattern.)

dkoo (changes requested):
@miguelpeixe I like the overall approach here, and I think it provides everything we need to fix the long-standing issue of the newsletter signup modal showing lists for the "wrong" user after a new account registration. However, that testing step is not working for me currently. See more details inline.

More feedback from a Claude code review below.

includes/reader-activation/class-reader-activation.php
miguelpeixe: Thanks, @dkoo! Addressed in e38acabad2b3f5571cf253a4c17dae8a1dd4300f. The Woo customer creation now also creates the cookie used to detect authentication and hydrate the session.

tests/unit-tests/session-hydration.php
miguelpeixe: Yep! Got left behind after a few iterations. Addressed in e38acabad2b3f5571cf253a4c17dae8a1dd4300f

includes/reader-activation/class-session-hydration.php
miguelpeixe: Addressed in e38acabad2b3f5571cf253a4c17dae8a1dd4300f

includes/reader-activation/class-session-hydration.php
miguelpeixe: Addressed in e38acabad2b3f5571cf253a4c17dae8a1dd4300f

src/reader-activation/session.js
miguelpeixe: Addressed in e38acabad2b3f5571cf253a4c17dae8a1dd4300f

includes/reader-activation/class-reader-activation.php
dkoo: Thanks for the fix! It's working for me now.

dkoo (approved):

PR #4632, opened 2026-04-06
fix(campaigns): guard against null taxonomy entries in prompt descriptions

jason10lee:
### All Submissions:

Guards `promptDescription()` and `warningForPopup()` in `utils.js` against null/non-array taxonomy data (tags, categories, campaign_groups) that causes the Campaigns admin page to crash with `TypeError: Cannot read properties of null (reading 'name')`.

Uses `Array.isArray()` + `.filter(Boolean)` to handle both `false` values from PHP's `get_the_tags()`/`get_the_terms()` and null entries from stale object cache.

Fixes [NPPM-2729: Campaigns admin page crashes when tag data contains null entries](https://linear.app/a8c/issue/NPPM-2729/campaigns-admin-page-crashes-when-tag-data-contains-null-entries).

1. Navigate to **Audience > Campaigns** — the page should load normally.
2. Add a mu-plugin to inject null tags:
   ```php
   add_filter( 'get_the_tags', function( $terms ) {
       if ( is_array( $terms ) ) {
           $terms[] = null;
       }
       return $terms;
   } );
   ```
3. Reload the Campaigns page — it should still render without errors.
4. Check the browser console — no `TypeError` about null `.name`.

### Other information:

### DIFF
--- a/src/wizards/audience/views/campaigns/utils.js
+++ b/src/wizards/audience/views/campaigns/utils.js
@@ -118,17 +118,20 @@ export const getCardClassName = ( status, forceDisabled = false ) => {
 export const promptDescription = prompt => {
 	const { categories, tags, campaign_groups: campaigns, status } = prompt;
 	const descriptionMessages = [];
-	if ( campaigns.length > 0 ) {
-		const campaignsList = campaigns.map( ( { name } ) => name ).join( ', ' );
+	const validCampaigns = Array.isArray( campaigns ) ? campaigns.filter( Boolean ) : [];
+	const validCategories = Array.isArray( categories ) ? categories.filter( Boolean ) : [];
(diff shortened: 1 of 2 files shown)


jason10lee:
My only comment here as to the approach would be that maybe this fix happens too far down the tree:  we own the plugin that's setting these variables, so this normalization could be happening there.  Maybe the framing of the original Linear task determined the agent's choice of fix?

Still, the fix works, and could serve as part of a broader belt-and-suspenders approach with a corresponding tweak to newspack-popups.

jason10lee:
Spotted that `warningForPopup()` is probably similarly affected, copied the fix over to it, and then just refactored because that's just way too many repeats.  Added tests, too, because why not while we're in there?

So...the scope's creeped, but for a good cause.

PR #4641, opened 2026-04-07
fix(reader-data): make is_donor read-only only when platform has server-side tracking

jason10lee:
### Changes proposed:

The `is_donor` reader data key was made unconditionally read-only in #4532, which broke a [Campaigns feature](https://github.com/Automattic/newspack-popups/pull/1181) that sets `is_donor` client-side on post-transaction landing pages. This mainly affects sites using non-Newspack donation platforms (NRH, other) where there is no server-side path to set the flag.

This PR makes `is_donor` conditionally read-only based on whether the donation platform has server-side donor tracking:

- **WooCommerce sites:** `is_donor` remains read-only. The `donation_new` data event handles it server-side.
- **NRH / other sites:** `is_donor` is writable, allowing the existing client-side landing page flow to work for both anonymous and logged-in readers.

The change introduces `Donations::has_server_side_donor_tracking()`, a capability method that decouples the policy ("does this platform manage donor status server-side?") from the platform identifier. Future platforms with secure server-side handling add themselves to this one method.

Note: when `is_donor` is not read-only, any authenticated reader can set it via the REST API. This is an intentional trade-off documented in the code — `is_donor` is used for segmentation and analytics, not access control. Consumers that need to distinguish server-verified from client-asserted donor status should check `Donations::has_server_side_donor_tracking()`.

Fixes [NPPD-1431: Allow non-Woo integrations to write `is_donor`](https://linear.app/a8c/issue/NPPD-1431/allow-non-woo-integrations-to-write-is-donor).

### How to test:

1. With the donation platform set to **WooCommerce** (default):
   - Confirm `is_donor` is in the read-only keys list: in the browser console, check `newspack_reader_data.read_only_keys` includes `is_donor`.
   - As a logged-in reader, confirm that `window.newspackReaderActivation.store.set('is_donor', true)` throws a read-only error.

2. Switch the donation platform to **NRH** (Audience > Checkout & Payment):
   - Confirm `is_donor` is NOT in `newspack_reader_data.read_only_keys`.
   - As a logged-in reader, confirm that `window.newspackReaderActivation.store.set('is_donor', true)` succeeds without error.
   - Confirm the value persists in localStorage as `np_reader_1_is_donor`.

   > **Note:** If switching via WP-CLI (`wp option update newspack_reader_revenue_platform nrh`), run `wp cache flush` afterward to clear the autoloaded option cache. Switching via the admin UI flushes the cache automatically.  This bit yr. humble dev while he was trying to red/green test, so making a note of it!

### DIFF
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -672,7 +672,14 @@ class Donations {
 	 * @return bool
 	 */
 	public static function has_server_side_donor_tracking() {
-		return self::is_platform_wc();
+		/**
+		 * Allow third parties to declare server-side donor tracking,
+		 * which will cause is_donor to be enforced as read-only
+		 * at the public API boundary.
+		 *
+		 * @param bool $has_tracking Whether the platform has server-side tracking. Default to whether we're using Woo.
+		 */
+		return apply_filters( 'newspack_has_server_side_donor_tracking', self::is_platform_wc() );
 	}
 
 	/**
--- a/includes/class-donations.php
+++ b/includes/class-donations.php
@@ -662,6 +662,20 @@ class Donations {
 	}
 
 	/**
+	 * Whether the current donation platform has a secure server-side mechanism
+	 * to manage donor status (e.g., via the donation_new data event).
+	 *
+	 * Platforms with server-side tracking can enforce is_donor as read-only
+	 * at the public API boundary. Platforms without it (NRH, other) need
+	 * client-side write access for post-transaction landing page flows.
+	 *
+	 * @return bool
+	 */
+	public static function has_server_side_donor_tracking() {
+		return self::is_platform_wc();
+	}
+
+	/**
 	 * Handle submission of the donation form.
 	 */
 	public static function process_donation_request() {
--- a/tests/unit-tests/reader-data-read-only-keys.php
+++ b/tests/unit-tests/reader-data-read-only-keys.php
@@ -17,10 +17,21 @@ use Newspack\Reader_Data;
 class Newspack_Test_Reader_Data_Read_Only_Keys extends WP_UnitTestCase {
 
 	/**
-	 * Tear down after each test — reset platform to default.
+	 * Filter callback registered during test_filter_can_add_custom_key.
+	 *
+	 * @var callable|null
+	 */
+	private $custom_key_filter = null;
+
+	/**
+	 * Tear down after each test.
 	 */
 	public function tear_down() {
 		Donations::set_platform_slug( 'wc' );
+		if ( $this->custom_key_filter ) {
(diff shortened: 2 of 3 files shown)


tests/unit-tests/reader-data-read-only-keys.php
```diff
@@ -0,0 +1,142 @@
...
+		};
+		add_filter( 'newspack_reader_data_read_only_keys', $add_key );
+
+		self::assertContains(
+			'custom_key',
+			Reader_Data::get_read_only_keys(),
+			'Filter should be able to add custom read-only keys.'
+		);
+
+		remove_filter( 'newspack_reader_data_read_only_keys', $add_key );
```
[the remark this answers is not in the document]
jason10lee: We've already got `tear_down()` so using that.  Fixed in [eba769a](https://github.com/Automattic/newspack-plugin/pull/4641/commits/eba769ad6591f97d0a70e6ea923390d7806c99db).

includes/class-donations.php
```diff
@@ -661,6 +661,20 @@ public static function is_platform_other() {
...
+	 * to manage donor status (e.g., via the donation_new data event).
+	 *
+	 * Platforms with server-side tracking can enforce is_donor as read-only
+	 * at the public API boundary. Platforms without it (NRH, other) need
+	 * client-side write access for post-transaction landing page flows.
+	 *
+	 * @return bool
+	 */
+	public static function has_server_side_donor_tracking() {
+		return self::is_platform_wc();
```
dkoo: Suggestion: since we've already decoupled this check into a handy helper method, let's apply a filter for this value when returning. That will let future integrations we or third-party devs build declare server-side donor tracking if they have it.

dkoo (changes requested):
Code looks good and tests well! The `is_donor` flag is set on reader data on the front-end in an anonymous or logged-in session. If set during an anonymous session where I then register a new reader account, the flag is successfully synced and persisted to the reader's user meta.

I have one suggestion below which will set us up for future integrations, otherwise looks good to go.

includes/class-donations.php
jason10lee: Yeah, it's really just half of a useful thing right now, isn't it?  I'll throw a filter on shortly...

jason10lee: Addressed in [76e40fc](https://github.com/Automattic/newspack-plugin/pull/4641/commits/76e40fcd1f565e6d87b9c5cf8c0339ba234253a8) with companion tests in [9f1d694](https://github.com/Automattic/newspack-plugin/pull/4641/commits/9f1d6949f5c2790b173cf2be0a6556d0102f5aea).

dkoo (approved):
Thanks for the revisions!

PR #4647, opened 2026-04-08
feat: require reader to set name when commenting

miguelpeixe:
### All Submissions:

Readers may register with just their email. In this case, their name defaults to the email handle. This should be fine for internal purposes, but shouldn't be used as a public display name in comments.

This PR implements a new input in the comment form, requiring a reader to set their display name if it has not yet been set. Once set, the input should no longer render.

image

The input placement follows the same standard as when commenting anonymously.

Closes NPPM-2701

1. Sign in as a reader who already has a display name set
2. Visit an article and confirm there's no new input
3. In a fresh session, register as a new reader using the registration flow from the header "Sign In" button
4. Visit an article and confirm the new input renders as in the image above
5. Confirm you are not able to comment without setting a name first
6. Set a name, post the comment, and confirm that the filled name is set
7. Navigate to My Account, verify the account, and confirm the chosen name is saved in the user
8. Visit another article and confirm the input no longer renders

### Other information:

### DIFF
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -105,9 +105,11 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-woocommerce.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync-admin.php';
+		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-comment-display-name.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-action-scheduler.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-integrations.php';
 		\Newspack\Reader_Activation\Integrations::init();
+		\Newspack\Reader_Activation\Comment_Display_Name::init();
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-promoted-fields.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-session-hydration.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-utils.php';
--- a/includes/class-newspack.php
+++ b/includes/class-newspack.php
@@ -108,8 +108,6 @@ final class Newspack {
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-comment-display-name.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-action-scheduler.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-integrations.php';
-		\Newspack\Reader_Activation\Integrations::init();
-		\Newspack\Reader_Activation\Comment_Display_Name::init();
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-promoted-fields.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-session-hydration.php';
 		include_once NEWSPACK_ABSPATH . 'includes/data-events/class-utils.php';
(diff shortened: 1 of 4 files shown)


includes/class-newspack.php
```diff
@@ -105,9 +105,11 @@ private function includes() {
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-woocommerce.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/sync/class-contact-sync-admin.php';
+		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-comment-display-name.php';
 		include_once NEWSPACK_ABSPATH . 'includes/class-action-scheduler.php';
 		include_once NEWSPACK_ABSPATH . 'includes/reader-activation/class-integrations.php';
 		\Newspack\Reader_Activation\Integrations::init();
+		\Newspack\Reader_Activation\Comment_Display_Name::init();
```
dkoo: Nit: I'm not sure I like this new pattern of calling `Class_Name::init()` in the main `Newspack` class. Our usual/informal convention is to have class files call their own `init` methods, right?

dkoo (commented):
Tests well! I left one non-blocking comment inline, but otherwise looks good to me.

Other non-blocking feedback from Claude to consider below.

> ## Code Review
> 
> **Functional Testing** — All scenarios pass:
> 
> - ✅ Reader with generic display name sees the "Name *" field in the comment form
> - ✅ `required` attribute blocks empty submission (client-side); server-side `wp_die()` provides fallback
> - ✅ Server-side validation rejects empty names and email-derived names
> - ✅ Submitting with a valid name updates comment_author, display_name, first_name, and last_name
> - ✅ Field disappears on subsequent pages after name is set
> - ✅ Non-reader users and readers with custom display names don't see the field
> - ✅ PHPUnit: 10/10 pass, 18 assertions
> 
> **Code Quality** — Clean, well-scoped implementation:
> 
> - Good use of `preprocess_comment` to both validate and override `comment_author` in one pass — avoids the race where WP reads the stale profile name before `wp_insert_comment`.
> - Proper sanitization with `sanitize_text_field` / `wp_unslash` on `$_POST` data.
> - Validation covers both `generate_user_nicename()` and `strip_email_domain()` patterns.
> - `NEWSPACK_ALLOW_GENERIC_READER_DISPLAY_NAMES` constant escape hatch (from the existing `reader_has_generic_display_name`) is a nice touch for sites that don't want this behavior.
> 
> **Suggestions (non-blocking):**
> 
> 1. **Inline style on input** (`class-comment-display-name.php:116`): `style="display:block;width:100%"` — the Newspack theme's comment form already styles `input[type="text"]` to be full-width. Consider removing the inline style or using a CSS class to keep styling in the theme layer. Worth verifying against the block theme as well.
> 
> 2. **No error handling on `wp_update_user()`** (`class-comment-display-name.php:92`): If `wp_update_user()` returns a `WP_Error`, the comment is still posted with the new name in `comment_author` but the user profile isn't actually updated. On the next page load, the reader would be prompted again. A minor edge case, but worth noting.
> 
> 3. **Case-sensitive email-derived name check** (`class-comment-display-name.php:72-73`): The `strip_email_domain()` comparison is case-sensitive. A reader with email `Jane.Doe@example.com` whose generic display name is `jane-doe` (lowercased via `generate_user_nicename`) could submit `Jane.Doe` (matching `strip_email_domain` output) and be blocked, but submitting `jane.doe` with different casing than the raw email handle would pass. This is a very minor edge case since the intent is to prevent accidental reuse, not strict enforcement — just flagging for awareness.
> 
> Overall, this is a clean, focused PR with good test coverage. Nice work! 👍

includes/class-newspack.php
miguelpeixe: Oh wow, well spotted. That looks awful

miguelpeixe: Addressed in efdd6eb5183dac60a491d4e5e72cf8e986adce00

dkoo (approved):
~Tests well! I left one non-blocking inline comment below, otherwise looks good to me.~ ghost message leftover in the browser, sorry

PR #4652, opened 2026-04-10
feat(block-theme): add size options to the Overlay Menu

laurelfulford:
### All Submissions:

This PR adds a couple options to the Overlay Menu block:
1. It adds a toggle to open the Menu Overlay panel as a full screen size. When this is enabled, the direction (and new size option) is hidden.
2. It adds an option to change the menu panel's width

It also adds `"supports": { "lock": false }` to both the Overlay Button (trigger) and Menu Panel block.json files, so users can't toggle the lock UI on those children. The parent Overlay Menu block already uses `templateLock="all"` on its `InnerBlocks`, so combined with this change the trigger and panel can't be removed or moved.

Closes NPPD-1362

1. Edit a post/page/template part, and add the Overlay Menu Block.
2. Open the panel, and select it. In the right sidebar, confirm you now have an option to switch the menu to 'full width', and to change its widths -- by default, the full-width option should be off, the panel should come from the left, and should use the 'S' size:

CleanShot 2026-04-10 at 08 30 48

4. Try enabling 'Full screen'; confirm that the other two options go away. This is because the menu will fade in (a left-or-right entry would be a little jarring at that size), and doesn't need a size set. 

CleanShot 2026-04-10 at 08 31 42

5. Publish and test on front-end; ensure it works well.  
6. Edit again, toggle off "Full screen", and test out the size options. Try them in the editor and on the front end; combine them with the different direction options, and test on different screen sizes. Confirm everything looks good. 

### Other information:

### DIFF
--- a/src/blocks/overlay-menu/panel/block.json
+++ b/src/blocks/overlay-menu/panel/block.json
@@ -15,6 +15,7 @@
 	"attributes": {
 		"slideDirection": {
 			"type": "string",
+			"enum": [ "left", "right" ],
 			"default": "left"
 		},
 		"isFullScreen": {
@@ -23,6 +24,7 @@
 		},
 		"panelWidth": {
 			"type": "string",
+			"enum": [ "x-small", "small", "medium", "large", "x-large" ],
 			"default": "small"
 		},
 		"overlayColor": {
--- a/src/blocks/overlay-menu/panel/block.json
+++ b/src/blocks/overlay-menu/panel/block.json
@@ -39,7 +39,8 @@
 		}
 	},
 	"supports": {
-		"html": false
+		"html": false,
+		"lock": false
 	},
 	"editorStyle": "file:../../../../dist/blocks.css",
 	"style": "file:../../../../dist/blocks.css"
--- a/src/blocks/overlay-menu/panel/block.json
+++ b/src/blocks/overlay-menu/panel/block.json
@@ -17,6 +17,14 @@
 			"type": "string",
 			"default": "left"
 		},
+		"isFullScreen": {
+			"type": "boolean",
+			"default": false
+		},
+		"panelWidth": {
+			"type": "string",
+			"default": "small"
+		},
 		"overlayColor": {
 			"type": "string",
 			"default": ""
--- a/src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
+++ b/src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
@@ -55,12 +55,24 @@ final class Overlay_Menu_Panel_Block {
 		$overlay_color    = $attributes['overlayColor'] ?? '';
 		$panel_bg_color   = $attributes['panelBackgroundColor'] ?? '';
 		$panel_text_color = $attributes['panelTextColor'] ?? '';
+		$is_full_screen   = ! empty( $attributes['isFullScreen'] );
+		$panel_width      = $attributes['panelWidth'] ?? 'small';
 
 		$valid_directions = [ 'left', 'right' ];
 		if ( ! in_array( $direction, $valid_directions, true ) ) {
 			$direction = 'left';
 		}
-		$position_class = 'overlay-menu__panel--' . $direction;
+
+		$valid_widths = [ 'x-small', 'small', 'medium', 'large', 'x-large' ];
+		if ( ! in_array( $panel_width, $valid_widths, true ) ) {
+			$panel_width = 'small';
+		}
+
+		if ( $is_full_screen ) {
+			$panel_class = 'overlay-menu__panel is-layout-constrained overlay-menu__panel--full-screen';
+		} else {
+			$panel_class = 'overlay-menu__panel is-layout-constrained overlay-menu__panel--' . $direction . ' overlay-menu__panel--width--' . $panel_width;
+		}
 
 		$panel_styles = [];
 		if ( $panel_bg_color ) {
@@ -71,8 +83,8 @@ final class Overlay_Menu_Panel_Block {
 		}
 		$extra_attributes = [
 			'id'                 => 'newspack-overlay-panel-' . $instance_id,
-			'class'              => 'overlay-menu__panel is-layout-constrained ' . $position_class,
-			'data-direction'     => $direction,
+			'class'              => $panel_class,
+			'data-direction'     => $is_full_screen ? 'full-screen' : $direction,
 			'data-overlay-color' => $overlay_color,
 			'aria-hidden'        => 'true',
 			'inert'              => 'true',
--- a/src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
+++ b/src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
@@ -84,7 +84,6 @@ final class Overlay_Menu_Panel_Block {
 		$extra_attributes = [
 			'id'                 => 'newspack-overlay-panel-' . $instance_id,
 			'class'              => $panel_class,
-			'data-direction'     => $is_full_screen ? 'full-screen' : $direction,
 			'data-overlay-color' => $overlay_color,
 			'aria-hidden'        => 'true',
 			'inert'              => 'true',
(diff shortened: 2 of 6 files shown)


src/blocks/overlay-menu/style.scss
```diff
@@ -101,29 +102,78 @@
...
+
+	// ─── Open state (shared by frontend JS and block editor preview) ──────────
 
-// ─── Open state (shared by frontend JS and block editor preview) ──────────────
+	&--open {
+		opacity: 1;
+
+		&.overlay-menu__panel--left  { left: 0; }
+		&.overlay-menu__panel--right { right: 0; }
+	}
```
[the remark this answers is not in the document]
laurelfulford: Added in [a7140e8](https://github.com/Automattic/newspack-plugin/pull/4652/commits/a7140e83d5839dd34bed51ab9921d6b296a72015)!

laurelfulford:
Hey @thomasguillot! This could probably use your eyes just to make sure the sizing feels okay 🙂 Thank you!

thomasguillot:
@laurelfulford Added `aria-label`s to the Width `ToggleGroupControlOption`s so each size shows a tooltip on hover (XS → Extra small, S → Small, M → Medium, L → Large, XL → Extra large). Also refactored the options into a `PANEL_WIDTH_OPTIONS` array to cut down on repetition.

thomasguillot:
@laurelfulford Refactored the panel width CSS: extracted a `panel-width` Sass mixin so the five size rules are one line each, and switched the `min-width: 782px` media queries to use the shared `breakpoints.$tablet_width` variable.

thomasguillot:
@laurelfulford Added `"supports": { "lock": false }` to both the Overlay Button (trigger) and Menu Panel `block.json` files, so users can't toggle the lock UI on those children. The parent Overlay Menu block already uses `templateLock="all"` on its `InnerBlocks`, so combined with this change the trigger and panel can't be removed or moved.

src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
```diff
@@ -71,8 +83,8 @@ public static function render_block( array $attributes, string $content, \WP_Blo
 		}
 		$extra_attributes = [
 			'id'                 => 'newspack-overlay-panel-' . $instance_id,
-			'class'              => 'overlay-menu__panel is-layout-constrained ' . $position_class,
-			'data-direction'     => $direction,
+			'class'              => $panel_class,
+			'data-direction'     => $is_full_screen ? 'full-screen' : $direction,
```
adekbadek: `data-direction` is set to the literal `"full-screen"` here, but the attribute's enum is `left|right` and this value isn't consumed by `view.js` or any CSS selector. Either drop `data-direction` in the full-screen branch or treat the new value as intentional public API and document it. Right now it's a slightly dead string.

src/blocks/overlay-menu/panel/block.json
```diff
@@ -17,6 +17,14 @@
 			"type": "string",
 			"default": "left"
 		},
+		"isFullScreen": {
+			"type": "boolean",
```
adekbadek, line 22: Consider adding `"enum": ["x-small", "small", "medium", "large", "x-large"]` to `panelWidth` (and `["left", "right"]` to `slideDirection`). The PHP allowlist already prevents class-name injection, so this is purely defense-in-depth on the schema side, but it makes the contract explicit and gives editor-side validation for free.

src/blocks/overlay-menu/style.scss
```diff
@@ -1,3 +1,5 @@
+@use "../../newspack-ui/scss/variables/breakpoints";
```
adekbadek, line 1: Nit / out of scope: this is the only place in `src/blocks/` that pulls in `newspack-ui/scss/variables/breakpoints`. The relative path couples this block's stylesheet to the `newspack-ui` directory layout. If this dependency starts spreading, a shared `_breakpoints.scss` in `src/shared/scss/` (next to the existing `_mixins.scss`) would be a more durable home.

src/blocks/overlay-menu/style.scss
```diff
@@ -87,43 +89,78 @@
...
+			width: calc( var( --wp--preset--spacing--40, 1.5rem ) + var( #{$var}, #{$fallback} ) + var( --wp--preset--spacing--40, 1.5rem ) );
+		}
+	}
 
-	&.overlay-menu__panel--left   { left: 0; }
-	&.overlay-menu__panel--right  { right: 0; }
+	&--width--x-small { @include panel-width( --wp--custom--width--x-small, 300px ); }
+	&--width--small   { @include panel-width( --wp--custom--width--small, 410px ); }
+	&--width--medium  { @include panel-width( --wp--style--global--content-size, 632px ); }
+	&--width--large   { @include panel-width( --wp--custom--width--large, 964px ); }
```
adekbadek, line 163: Nit: the `medium` fallback maps to `--wp--style--global--content-size`, which is a different variable family than the others (`--wp--custom--width--*`). It's intentional (matches editor content width), but a one-line comment would prevent it reading like a typo.

adekbadek (approved):
Verified the new options end-to-end: defaults render as Full Screen off / direction Left / width S; the Full Screen toggle correctly hides direction + width; and the saved page on the frontend opens the panel from the configured side at the configured width. No PR-related PHP warnings showed up in the debug log.

Approving with a few small notes inline.

A couple of broader things that didn't fit on a line:

- The PR description doesn't mention the `lock: false` additions on the panel and trigger blocks. The commit message documents the intent, but it would be worth a one-line mention in the description so reviewers don't have to dig through commits.
- `render_block` now has enough branching (full-screen vs direction + width allowlist + invalid-value fallback) that a small PHPUnit test covering the class-string output for valid and invalid attribute combos would be a nice way to lock in the contract — especially the allowlist behavior, since that's the bit doing real defense.

src/blocks/overlay-menu/style.scss
laurelfulford: Thanks Adam! I'm going to make a Linear task to circle back to this one. It looks like Collections has its own set of breakpoints, and it would be nice to get a consolidated set we could use across all the front-end elements in the plugin.

src/blocks/overlay-menu/panel/block.json
laurelfulford: Thanks! This should be fixed in [368ca1c](https://github.com/Automattic/newspack-plugin/pull/4652/commits/368ca1c42528f717ae8630d6c2e206a83961b59c).

src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
laurelfulford: Eep, good call! Removed in [b2fc20f](https://github.com/Automattic/newspack-plugin/pull/4652/commits/b2fc20f8663b1e13b3148f3b7bc3efa34aadb650).

src/blocks/overlay-menu/style.scss
laurelfulford: Good call! Added a comment explaining the difference in [88f305c](https://github.com/Automattic/newspack-plugin/pull/4652/commits/88f305cf345df8f712c2e3bffa74ab293256e9bd).

laurelfulford:
Thanks @adekbadek!

> The PR description doesn't mention the `lock: false` additions on the panel and trigger blocks. The commit message documents the intent, but it would be worth a one-line mention in the description so reviewers don't have to dig through commits.

Good call! I've added this after the list of feature additions, since it's kind of its own change.

> `render_block` now has enough branching (full-screen vs direction + width allowlist + invalid-value fallback) that a small PHPUnit test covering the class-string output for valid and invalid attribute combos would be a nice way to lock in the contract — especially the allowlist behavior, since that's the bit doing real defense.

Tests have been ~~vibed~~ added in [8bf0d16](https://github.com/Automattic/newspack-plugin/pull/4652/commits/8bf0d16f05f318791781eaabac75fc5e5d2f3692).

Just let me know if anything else needs updating!

Edited to add: merging since this was approved, but we can iterate if needed.

PR #4660, opened 2026-04-14
fix(memberships): prevent duplicate teams on renewal + diagnostics CLI

adekbadek:

SkyVerge's WC Memberships for Teams dispatches team actions (create / renew / seat_change) on order completion purely from order-item meta set at initial checkout:

- `_wc_memberships_for_teams_team_id`
- `_wc_memberships_for_teams_team_renewal`
- `_wc_memberships_for_teams_team_seat_change`

WC Subscriptions copies those onto renewal order items. But if an admin manually replaces a subscription line item in wp-admin (delete + re-add), the new `woocommerce_order_itemmeta` row has no such meta. On the next renewal, Teams plugin falls through to `action=create` and **spawns a duplicate team for an already-linked subscription**, then moves the existing user_membership onto the new (orphaned, `_subscription_id`-less) team. Future renewals never reactivate the member. Same class of bug we patched reactively at Newsroom NZ in Nov 2023.

This PR adds two things:

**1. Prevention – `Teams_For_Memberships::restore_team_meta_on_renewal`**

A new filter callback on `wc_memberships_for_teams_determine_order_item_action` (Teams plugin's `src/Orders.php:131`). When the proposed action is \`create\` and the host order is a subscription renewal whose parent subscription already has a linked team for the same product, we:

- Restore `_wc_memberships_for_teams_team_id` + `_wc_memberships_for_teams_team_renewal` on the order item.
- Rewrite the action from `create` to `renew`.

Teams' own `process_team_creation_action` then reads the restored `_team_id`, calls `wc_memberships_for_teams_create_team( [ 'team_id' => ... ], 'renew' )`, and safely updates the existing team instead of creating a duplicate. Idempotent, and orthogonal to the existing `update_team_subscription_on_resubscribe` method (which handles the `expired → resubscribe` path, not this one).

Lookup helper reused: `SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions::get_teams_from_subscription()`.

**2. Detection + repair – `wp newspack teams-for-memberships diagnostics [--fix]`**

New reusable WP-CLI command (`includes/cli/class-teams-for-memberships-diagnostics.php`). Scans any Newspack site running Teams for the five related data pathologies:

1. Duplicate teams (same title + same `post_author`).
2. Teams missing `_subscription_id`.
3. user_memberships missing `_subscription_id` but whose team has one.
4. Team `_member_id` vs linked subscription `_customer_user` mismatch.
5. Subscription line items with missing or **stale** `_wc_memberships_for_teams_team_id` (stale = points at a team whose `_subscription_id` doesn't match the containing subscription – this is the state left after a duplicate is created and is the ticking bomb for the next renewal cycle).

Read-only by default; `--fix` repairs 1, 2, 3 and 5 (4 is intentionally manual – a human has to decide whether to transfer the team or the subscription). `--dry-run` is implied when `--fix` is absent. `--fix` requires `WP_CLI::confirm` unless `--yes` is passed.

**Unit tests** – 4 new tests:

\`\`\`
docker exec newspack_dev bash -c '/var/scripts/test-php.sh newspack-plugin --filter Test_Teams_For_Memberships'
\`\`\`

Expected: \`OK (4 tests, 5 assertions)\`.

**End-to-end repro in an isolated env** (also runs the filter against real Teams plugin + WC Subs):

1. \`n env create nppm2741 --worktree newspack-plugin:fix/NPPM-2741-teams-for-memberships --worktree newspack-network:trunk\`
2. \`n env up nppm2741\`
4. Run the repro script (creates a team-product subscription, strips the dispatch meta on the sub's line item, forces a renewal, and counts teams):
   - **With the fix registered:** total teams stays at 1; the renewal item ends up with \`_team_id\` + \`_team_renewal\` restored.
   - **With the filter callback removed from \`init()\`:** total teams goes from 1 → 2; Teams plugin spawns a duplicate. Confirms the bug reproduces and the fix is what's preventing it.

**CLI smoke test in the same env:**

\`\`\`
wp newspack teams-for-memberships diagnostics           # read-only
wp newspack teams-for-memberships diagnostics --fix --yes
\`\`\`

Expected on a clean env: \`No issues found.\`

### Other information:

- [x] New tests added (`tests/unit-tests/plugins/teams-for-memberships.php`, 4 tests)
- [x] Tests run locally
- [x] PHPCS clean on changed files
- [x] End-to-end reproduction verified in an isolated env

### DIFF
--- a/includes/cli/class-initializer.php
+++ b/includes/cli/class-initializer.php
@@ -29,6 +29,7 @@ class Initializer {
 		include_once NEWSPACK_ABSPATH . 'includes/cli/class-mailchimp.php';
 		include_once NEWSPACK_ABSPATH . 'includes/cli/class-optional-modules.php';
 		include_once NEWSPACK_ABSPATH . 'includes/cli/class-woocommerce-subscriptions.php';
+		include_once NEWSPACK_ABSPATH . 'includes/cli/class-teams-for-memberships-diagnostics.php';
 	}
 
 	/**
@@ -78,6 +79,11 @@ class Initializer {
 		WP_CLI::add_command( 'newspack migrate-co-authors-guest-authors', [ 'Newspack\CLI\Co_Authors_Plus', 'migrate_guest_authors' ] );
 		WP_CLI::add_command( 'newspack backfill-non-editing-contributors', [ 'Newspack\CLI\Co_Authors_Plus', 'backfill_non_editing_contributor' ] );
 		WP_CLI::add_command( 'newspack migrate-expired-subscriptions', [ 'Newspack\CLI\WooCommerce_Subscriptions', 'migrate_expired_subscriptions' ] );
+
+		WP_CLI::add_command(
+			'newspack teams-for-memberships diagnostics',
+			[ 'Newspack\CLI\Teams_For_Memberships_Diagnostics', 'diagnostics' ]
+		);
 
 		Optional_Modules::register_commands();
 	}
--- a/includes/cli/class-teams-for-memberships-diagnostics.php
+++ b/includes/cli/class-teams-for-memberships-diagnostics.php
@@ -1,0 +1,538 @@
+<?php
+/**
+ * WC Memberships for Teams diagnostics CLI command.
+ *
+ * Detects (and optionally repairs) data inconsistencies introduced by SkyVerge
+ * WC Memberships for Teams when a subscription's line items lose their team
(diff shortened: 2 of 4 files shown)


adekbadek:
Split into two stacked PRs for easier review: #PREVENTION (prevention fix) and #DETECTION (diagnostics CLI, stacked on top).

PR #4661, opened 2026-04-14
fix(memberships): prevent duplicate teams on stripped-meta renewals

adekbadek:

SkyVerge's WC Memberships for Teams dispatches team actions (create / renew / seat_change) on order completion purely from order-item meta set at initial checkout:

- `_wc_memberships_for_teams_team_id`
- `_wc_memberships_for_teams_team_renewal`
- `_wc_memberships_for_teams_team_seat_change`

WC Subscriptions copies those onto renewal order items. But if an admin manually replaces a subscription line item in wp-admin (delete + re-add), the new `woocommerce_order_itemmeta` row has no such meta. On the next renewal, Teams plugin falls through to `action=create` and **spawns a duplicate team for an already-linked subscription**, then moves the existing user_membership onto the new (orphaned, `_subscription_id`-less) team. Future renewals never reactivate the member. Same class of bug we patched reactively at Newsroom NZ in Nov 2023 (see `newspack-custom-code/plugins/newspack-newsroomnz-fix-teams/`).

**The fix — `Teams_For_Memberships::restore_team_meta_on_renewal`**

A new filter callback on `wc_memberships_for_teams_determine_order_item_action` (Teams plugin's `src/Orders.php:131`). When the proposed action is `create` and the host order is a subscription renewal whose parent subscription already has a linked team for the same product, we:

- Restore `_wc_memberships_for_teams_team_id` + `_wc_memberships_for_teams_team_renewal` on the order item.
- Rewrite the action from `create` to `renew`.

Teams' own `process_team_creation_action` then reads the restored `_team_id`, calls `wc_memberships_for_teams_create_team( [ 'team_id' => ... ], 'renew' )`, and safely updates the existing team instead of creating a duplicate. Idempotent, and orthogonal to the existing `update_team_subscription_on_resubscribe` method (which handles the `expired → resubscribe` path, not this one).

Lookup helper reused: `SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions::get_teams_from_subscription()`.

A follow-up PR (stacked on this branch) adds a detection + repair WP-CLI command for the existing data damage this prevention fix doesn't touch.

**Unit tests** – 4 new tests:

```
docker exec newspack_dev bash -c '/var/scripts/test-php.sh newspack-plugin --filter Test_Teams_For_Memberships'
```

Expected: `OK (4 tests, 5 assertions)`.

**End-to-end reproduction in an isolated env** (exercises the filter against real Teams plugin + WC Subs):

1. `n env create nppm2741 --worktree newspack-plugin:fix/NPPM-2741-prevent-duplicate-teams-on-renewal`
2. `n env up nppm2741`
4. Create a team-enabled subscription product (set `_wc_memberships_for_teams_has_team_membership = yes` and `_wc_memberships_for_teams_plan = <plan id>` on the product), a plan, a user; create an active subscription with one item; create a team via `wc_memberships_for_teams_create_team` with `plan_id`, `product_id`, `owner_id`, `seats`, then set `_subscription_id` meta on the team post.
5. Strip the dispatch meta from the subscription's line item (simulating the admin edit): `wc_delete_order_item_meta` for all three meta keys.
6. `$renewal = wcs_create_renewal_order( $subscription ); $renewal->update_status( 'completed' );`
7. Assert: team count is unchanged (still 1), the renewal order item now has `_team_id` + `_team_renewal` set (restored by this filter).

**Reverting the filter registration in `init()` reproduces the bug:** team count goes from 1 → 2 and the renewal item ends up with a new duplicate team id. With the filter registered: team count stays at 1.

### Other information:

- [x] New tests added (`tests/unit-tests/plugins/teams-for-memberships.php`, 4 tests)
- [x] Tests run locally (4 tests / 5 assertions / OK)
- [x] PHPCS clean on changed files
- [x] End-to-end reproduction verified in an isolated env

### DIFF
--- a/includes/plugins/class-teams-for-memberships.php
+++ b/includes/plugins/class-teams-for-memberships.php
@@ -29,6 +29,7 @@ class Teams_For_Memberships {
 		add_filter( 'newspack_esp_sync_contact', [ __CLASS__, 'handle_esp_sync_contact' ] );
 		add_filter( 'newspack_my_account_disabled_pages', [ __CLASS__, 'enable_members_area_for_team_members' ] );
 		add_action( 'woocommerce_checkout_subscription_created', [ __CLASS__, 'update_team_subscription_on_resubscribe' ], 21, 2 );
+		add_filter( 'wc_memberships_for_teams_determine_order_item_action', [ __CLASS__, 'restore_team_meta_on_renewal' ], 10, 2 );
 	}
 
 	/**
@@ -309,6 +310,69 @@ class Teams_For_Memberships {
 				}
 			}
 		}
+	}
+
+	/**
+	 * Restore stripped team order-item meta on renewal orders.
+	 *
+	 * SkyVerge's Teams plugin dispatches team actions from order item meta:
+	 * `_wc_memberships_for_teams_team_id`, `_wc_memberships_for_teams_team_renewal`,
+	 * `_wc_memberships_for_teams_team_seat_change`. WC Subscriptions copies these onto
+	 * renewal order items at generation time, but if an admin manually replaces a
+	 * subscription line item in wp-admin, the replacement row has no such meta. On the
+	 * next renewal, Teams falls through to the `create` branch and creates a duplicate
+	 * team for an already-linked subscription.
+	 *
+	 * This filter runs on `wc_memberships_for_teams_determine_order_item_action`: when
+	 * the proposed action is `create` and the order is a subscription renewal whose
+	 * parent subscription already has a linked team for the same product, we restore
+	 * the stripped meta on the item and force the action to `renew`.
+	 *
+	 * @see https://linear.app/a8c/issue/NPPM-2741
+	 *
+	 * @param string         $action The action determined by SkyVerge Teams (create|renew|seat_change).
+	 * @param \WC_Order_Item $item   The order item being processed.
+	 * @return string The possibly-rewritten action.
+	 */
+	public static function restore_team_meta_on_renewal( $action, $item ) {
+		if ( 'create' !== $action ) {
+			return $action;
+		}
+		if ( ! function_exists( 'wcs_order_contains_renewal' ) || ! function_exists( 'wcs_get_subscriptions_for_renewal_order' ) ) {
+			return $action;
+		}
+		if ( ! method_exists( '\SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions', 'get_teams_from_subscription' ) ) {
+			return $action;
+		}
+		if ( ! $item instanceof \WC_Order_Item_Product ) {
+			return $action;
+		}
+
+		$order = $item->get_order();
+		if ( ! $order instanceof \WC_Order || ! wcs_order_contains_renewal( $order ) ) {
+			return $action;
+		}
+
+		$subscriptions      = wcs_get_subscriptions_for_renewal_order( $order );
+		$team_subscriptions = new \SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions();
+		$item_product_id    = (int) $item->get_product_id();
+
+		foreach ( $subscriptions as $subscription ) {
+			$teams = $team_subscriptions->get_teams_from_subscription( $subscription->get_id() );
+			if ( empty( $teams ) ) {
+				continue;
+			}
+			foreach ( $teams as $team ) {
+				if ( (int) $team->get_product_id() !== $item_product_id ) {
+					continue;
+				}
+				wc_update_order_item_meta( $item->get_id(), '_wc_memberships_for_teams_team_id', $team->get_id() );
+				wc_update_order_item_meta( $item->get_id(), '_wc_memberships_for_teams_team_renewal', true );
+				return 'renew';
+			}
+		}
+
+		return $action;
 	}
 }
 
--- a/includes/plugins/class-teams-for-memberships.php
+++ b/includes/plugins/class-teams-for-memberships.php
@@ -355,7 +355,11 @@ class Teams_For_Memberships {
 
 		$subscriptions      = wcs_get_subscriptions_for_renewal_order( $order );
 		$team_subscriptions = new \SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions();
-		$item_product_id    = (int) $item->get_product_id();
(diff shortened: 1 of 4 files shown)


tests/unit-tests/plugins/teams-for-memberships.php
```diff
@@ -0,0 +1,70 @@
...
+	 */
+	public function test_restore_team_meta_on_renewal_ignores_non_product_items() {
+		// stdClass is not a WC_Order_Item_Product, so the instanceof check should short-circuit.
+		$this->assertSame(
+			'create',
+			Teams_For_Memberships::restore_team_meta_on_renewal( 'create', new stdClass() )
+		);
+	}
+
+	/**
```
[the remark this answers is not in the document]
adekbadek: Addressed in 2eca7c48. Removed this test (and the tautological `bails_without_subscriptions` test); replaced with a stub fixture at `tests/mocks/teams-for-memberships-mocks.php` that defines the minimal `wcs_*` helpers and a fake SkyVerge Subscriptions integration, so the new `test_restore_team_meta_on_renewal_restores_meta_and_returns_renew` can actually exercise the branch.

includes/plugins/class-teams-for-memberships.php
```diff
@@ -310,6 +311,69 @@ public static function update_team_subscription_on_resubscribe( $new_subscriptio
...
+				continue;
+			}
+			foreach ( $teams as $team ) {
+				if ( (int) $team->get_product_id() !== $item_product_id ) {
+					continue;
+				}
+				wc_update_order_item_meta( $item->get_id(), '_wc_memberships_for_teams_team_id', $team->get_id() );
+				wc_update_order_item_meta( $item->get_id(), '_wc_memberships_for_teams_team_renewal', true );
+				return 'renew';
+			}
```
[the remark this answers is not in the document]
adekbadek: Addressed in 2eca7c48. Added `test_restore_team_meta_on_renewal_restores_meta_and_returns_renew`: stubs the WC Subscriptions helpers and SkyVerge Teams Subscriptions integration via the new fixture in `tests/mocks/teams-for-memberships-mocks.php`, exercises the full branch, and asserts that (a) the action is rewritten from `create` to `renew` and (b) both `_wc_memberships_for_teams_team_id` and `_wc_memberships_for_teams_team_renewal` are written onto the order item. Also added an `ignores_items_for_unrelated_products` companion test that proves the same code path passes through cleanly for non-matching product ids. Test count is now 4 tests / 9 assertions.

includes/plugins/class-teams-for-memberships.php
```diff
@@ -310,6 +311,69 @@ public static function update_team_subscription_on_resubscribe( $new_subscriptio
...
+		$team_subscriptions = new \SkyVerge\WooCommerce\Memberships\Teams\Integrations\Subscriptions();
+		$item_product_id    = (int) $item->get_product_id();
+
+		foreach ( $subscriptions as $subscription ) {
+			$teams = $team_subscriptions->get_teams_from_subscription( $subscription->get_id() );
+			if ( empty( $teams ) ) {
+				continue;
+			}
+			foreach ( $teams as $team ) {
+				if ( (int) $team->get_product_id() !== $item_product_id ) {
```
dkoo: If the Teams-enabled product is a variable subscription product, then this comparison won't match. `$item->get_product_id()` returns the variation's parent product ID, while `$team->get_product_id()` returns the variation ID. When we're fetching the product ID, we may want to check for a parent ID and use that instead.

includes/plugins/class-teams-for-memberships.php
```diff
@@ -310,6 +311,69 @@ public static function update_team_subscription_on_resubscribe( $new_subscriptio
...
+					continue;
+				}
+				wc_update_order_item_meta( $item->get_id(), '_wc_memberships_for_teams_team_id', $team->get_id() );
+				wc_update_order_item_meta( $item->get_id(), '_wc_memberships_for_teams_team_renewal', true );
```
dkoo: The PR description mentions `_wc_memberships_for_teams_team_seat_change` but we're not resetting it here—should we?

dkoo: Also, should we be using `$item->update_meta_data()` instead of `wc_update_order_item_meta()`? My understanding is the former is preferred, e.g.:

```suggestion
				$item->update_meta_data( '_wc_memberships_for_teams_team_id', $team->get_id() );
				$item->update_meta_data( '_wc_memberships_for_teams_team_renewal', true );
				$item->save();
```

dkoo (changes requested):
Largely works with manual testing, but I'm not sure if this covers variations of a variable subscription product, too? In my manual testing it seemed to work for a subscription variation, but when I manually compared the results of `$item->get_product_id()` vs `$team->get_product_id()` using `wp shell` they differed, which makes me think the comparison logic will fail as written.

More details and other comments inline.

includes/plugins/class-teams-for-memberships.php
adekbadek: Addressed in f977debf. Now compares the team's product id against both `$item->get_product_id()` and `$item->get_variation_id()` via an `array_filter`-ed lookup array, so a team linked to a variation matches a renewal line item whose parent product id differs. Added `test_restore_team_meta_on_renewal_matches_variation_id` to cover the case.

includes/plugins/class-teams-for-memberships.php
adekbadek: Addressed in f977debf. Now calls `$item->delete_meta_data( '_wc_memberships_for_teams_team_seat_change' )` so the dispatcher treats the item unambiguously as a renewal rather than potentially picking up a stale flag. The happy-path test asserts that the meta is explicitly cleared.

adekbadek: Addressed in f977debf. Switched to `$item->update_meta_data()` + `$item->save()` (and `$item->delete_meta_data()` for the seat-change clear).

dkoo (approved):
@adekbadek Thanks for the changes! Didn't redo manual testing, but the refactors and new unit tests look solid to me. The extra check against both parent product ID and variation ID should capture all simple + variable subscription products, and using the `get_meta_data` + `save` methods should avoid potential race conditions inherent in the legacy `wc_update_order_item_meta` method.

PR #4662, opened 2026-04-14
feat(cli): teams-for-memberships diagnostics command

adekbadek:

**Stacked on #4661** (the prevention fix). Review #4661 first. This PR adds only the CLI class and its registration.

New reusable WP-CLI command for detecting (and optionally repairing) the data pathologies that result from the duplicate-team bug — both fresh occurrences and the accumulated damage from older incidents like the Newsroom NZ issue in Nov 2023.

```
wp newspack teams-for-memberships diagnostics [--fix] [--team-id=<id>] [--yes]
```

Five checks, read-only by default:

1. **Duplicate teams** (same title + same `post_author`).
2. **Teams missing `_subscription_id`**.
3. **user_memberships missing `_subscription_id`** but whose team has one.
4. **Team `_member_id` vs linked subscription `_customer_user` mismatch** — intentionally manual. A human has to decide whether to transfer the team or the subscription; auto-fixing either side is the path most likely to chain into further data loss. The prevention fix in #4661 makes the system resilient to the mismatch at renewal dispatch time anyway.
5. **Subscription line items with missing or stale `_wc_memberships_for_teams_team_id`** — "stale" = meta is present but points at a team whose `_subscription_id` doesn't match the containing subscription. This is the state left behind after a duplicate team has been created, and is the ticking bomb for the next renewal cycle.

`--fix` repairs 1, 2, 3 and 5:

- **Check 1 repair:** pick the oldest team that carries a `_subscription_id` as the canonical one. Move every team_member from the duplicate onto the original via `$original_team->add_member()`, relink any remaining orphan user_memberships to the original, copy over the newer `_order_id` if needed, then `wp_delete_post` the empty duplicate. Mirrors `newspack_fix_nz_teams` in `newspack-custom-code/plugins/newspack-newsroomnz-fix-teams/newspack-newsroomnz-fix-teams.php:97-189`.
- **Check 2 repair:** look for a single candidate `_subscription_id` via (a) any user_membership on the team that carries its own `_subscription_id` meta, or (b) an active subscription owned by the team owner that contains the team's product. If exactly one candidate, set it on the team; otherwise log and skip.
- **Check 3 repair:** copy `_subscription_id` from the team post onto the membership. Direct port of `newspack_link_nz_memberships` lines 191–255 in the same NZ plugin.
- **Check 5 repair:** for each flagged item, write `_wc_memberships_for_teams_team_id` (from the team linked to that subscription) and `_wc_memberships_for_teams_team_renewal=true`. Redundant with the prevention fix for fresh occurrences but useful for backfilling historical state.

`--fix` requires `WP_CLI::confirm` unless `--yes` is passed. `--dry-run` is implied when `--fix` is absent.

No auto-repair runs on check 4 (owner mismatch).

**Clean env smoke test:**

1. `n env create nppm2741 --worktree newspack-plugin:fix/NPPM-2741-teams-diagnostics-cli`
2. `n env up nppm2741`
4. `wp newspack teams-for-memberships diagnostics` → expect `Success: No issues found.`

**Seeded issues test:**

1. Create a team via `wc_memberships_for_teams_create_team` without `_subscription_id` → Check 2 should flag.
2. Create a user_membership with `_team_id` meta but no `_subscription_id` → Check 3 should flag.
3. Create two teams with the same title + same `post_author` → Check 1 should flag.
4. Seed a subscription whose line item's `_wc_memberships_for_teams_team_id` points at a team whose own `_subscription_id` is a different sub → Check 5 should flag "stale team id".
5. Run `wp newspack teams-for-memberships diagnostics --fix --yes`.
6. Re-run diagnostics → expect `No issues found` for checks 1, 2, 3, 5.

### Other information:

- [x] Stacked on #4661; review that PR first
- [x] PHPCS clean
- [x] Command is gated behind `class_exists( 'WC_Memberships_For_Teams_Loader' )` so it's a no-op on sites without Teams
- [x] Smoke-tested end-to-end in an isolated env (both clean and seeded cases)

### DIFF
--- a/includes/cli/class-initializer.php
+++ b/includes/cli/class-initializer.php
@@ -30,6 +30,7 @@ class Initializer {
 		include_once NEWSPACK_ABSPATH . 'includes/cli/class-optional-modules.php';
 		include_once NEWSPACK_ABSPATH . 'includes/cli/class-woocommerce-subscriptions.php';
 		include_once NEWSPACK_ABSPATH . 'includes/cli/class-ga4-dimensions.php';
+		include_once NEWSPACK_ABSPATH . 'includes/cli/class-teams-for-memberships-diagnostics.php';
 	}
 
 	/**
@@ -80,6 +81,11 @@ class Initializer {
 		WP_CLI::add_command( 'newspack backfill-non-editing-contributors', [ 'Newspack\CLI\Co_Authors_Plus', 'backfill_non_editing_contributor' ] );
 		WP_CLI::add_command( 'newspack migrate-expired-subscriptions', [ 'Newspack\CLI\WooCommerce_Subscriptions', 'migrate_expired_subscriptions' ] );
 		WP_CLI::add_command( 'newspack ga4-dimensions', 'Newspack\CLI\GA4_Dimensions' );
+
+		WP_CLI::add_command(
+			'newspack teams-for-memberships diagnostics',
+			[ 'Newspack\CLI\Teams_For_Memberships_Diagnostics', 'diagnostics' ]
+		);
 
 		Optional_Modules::register_commands();
 	}
--- a/includes/cli/class-initializer.php
+++ b/includes/cli/class-initializer.php
@@ -82,10 +82,14 @@ class Initializer {
 		WP_CLI::add_command( 'newspack migrate-expired-subscriptions', [ 'Newspack\CLI\WooCommerce_Subscriptions', 'migrate_expired_subscriptions' ] );
 		WP_CLI::add_command( 'newspack ga4-dimensions', 'Newspack\CLI\GA4_Dimensions' );
 
-		WP_CLI::add_command(
-			'newspack teams-for-memberships diagnostics',
-			[ 'Newspack\CLI\Teams_For_Memberships_Diagnostics', 'diagnostics' ]
-		);
+		// Only register the Teams for Memberships diagnostics command on sites where the
+		// SkyVerge plugin is active. No reason to surface it in `wp help` otherwise.
+		if ( class_exists( 'WC_Memberships_For_Teams_Loader' ) ) {
+			WP_CLI::add_command(
+				'newspack teams-for-memberships diagnostics',
+				[ 'Newspack\CLI\Teams_For_Memberships_Diagnostics', 'diagnostics' ]
+			);
+		}
 
 		Optional_Modules::register_commands();
 	}
--- a/includes/cli/class-teams-for-memberships-diagnostics.php
+++ b/includes/cli/class-teams-for-memberships-diagnostics.php
@@ -251,7 +251,6 @@ class Teams_For_Memberships_Diagnostics {
 		}
 
 		// Finally delete the now-empty duplicate.
-		$remaining = get_post_meta( $duplicate->ID, '_member_id', true );
 		$remaining_members = $duplicate_team->get_member_ids();
 		if ( empty( $remaining_members ) ) {
 			WP_CLI::line( sprintf( '    DELETE duplicate team #%d', $duplicate->ID ) );
@@ -315,11 +314,17 @@ class Teams_For_Memberships_Diagnostics {
 		}
 
 		// Candidate 2: an active subscription owned by the team owner that includes the team's product.
-		$member_id  = (int) get_post_meta( $team->ID, '_member_id', true );
-		$product_id = (int) get_post_meta( $team->ID, '_product_id', true );
+		// `wcs_get_users_subscriptions` returns every status, so filter to live subscriptions – cancelled or
+		// expired subs would give us a wrong link and either mis-repair or block fix via the "multiple candidates" branch.
+		$active_statuses = [ 'active', 'pending-cancel' ];
+		$member_id       = (int) get_post_meta( $team->ID, '_member_id', true );
+		$product_id      = (int) get_post_meta( $team->ID, '_product_id', true );
 		if ( $member_id && $product_id && function_exists( 'wcs_get_users_subscriptions' ) ) {
 			$user_subs = wcs_get_users_subscriptions( $member_id );
 			foreach ( $user_subs as $sub ) {
+				if ( ! in_array( $sub->get_status(), $active_statuses, true ) ) {
+					continue;
+				}
 				foreach ( $sub->get_items() as $sub_item ) {
 					if ( (int) $sub_item->get_product_id() === $product_id ) {
 						$candidate_sub_ids[ (int) $sub->get_id() ] = true;
@@ -350,12 +355,17 @@ class Teams_For_Memberships_Diagnostics {
 		WP_CLI::line( 'Check 3: memberships missing _subscription_id but recoverable via team' );
 		global $wpdb;
 
-		$base_sql = "SELECT p.ID AS membership_id, pm_team.meta_value AS team_id
+		// Pull `team._subscription_id` via a join in the same query so we don't issue per-row
+		// get_post_meta() calls for every flagged membership. The inner join on pm_team_sub
+		// also naturally filters out Check 2 cases (teams that have no _subscription_id either).
+		$base_sql = "SELECT p.ID AS membership_id, pm_team.meta_value AS team_id, pm_team_sub.meta_value AS team_sub_id
 			FROM $wpdb->posts p
 			JOIN $wpdb->postmeta pm_team ON pm_team.post_id = p.ID AND pm_team.meta_key = '_team_id'
 			LEFT JOIN $wpdb->postmeta pm_sub ON pm_sub.post_id = p.ID AND pm_sub.meta_key = '_subscription_id'
+			JOIN $wpdb->postmeta pm_team_sub ON pm_team_sub.post_id = pm_team.meta_value AND pm_team_sub.meta_key = '_subscription_id'
 			WHERE p.post_type = 'wc_user_membership'
-			AND ( pm_sub.meta_value IS NULL OR pm_sub.meta_value = '' )";
+			AND ( pm_sub.meta_value IS NULL OR pm_sub.meta_value = '' )
+			AND pm_team_sub.meta_value <> ''";
 
 		if ( self::$team_id ) {
 			$rows = $wpdb->get_results( // phpcs:ignore WordPress.DB.DirectDatabaseQuery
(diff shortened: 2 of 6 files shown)


includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,538 @@
...
+		$product_id = (int) get_post_meta( $team->ID, '_product_id', true );
+		if ( $member_id && $product_id && function_exists( 'wcs_get_users_subscriptions' ) ) {
+			$user_subs = wcs_get_users_subscriptions( $member_id );
+			foreach ( $user_subs as $sub ) {
+				foreach ( $sub->get_items() as $sub_item ) {
+					if ( (int) $sub_item->get_product_id() === $product_id ) {
+						$candidate_sub_ids[ (int) $sub->get_id() ] = true;
+					}
+				}
+			}
```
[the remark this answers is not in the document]
adekbadek: Addressed in da7f8617. Filter added to only consider `active` and `pending-cancel` subscriptions when scanning the team owner's subs as Candidate 2. Cancelled/expired subs are now skipped so they can't become a wrong-link repair target.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,538 @@
...
+		} else {
+			$rows = $wpdb->get_results( $base_sql ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery, WordPress.DB.PreparedSQL.NotPrepared
+		}
+
+		$issue_count = 0;
+		foreach ( $rows as $row ) {
+			$team_sub_id = get_post_meta( (int) $row->team_id, '_subscription_id', true );
+			if ( empty( $team_sub_id ) ) {
+				// The team has no subscription either. This is a Check 2 problem, not a Check 3 fix.
+				continue;
```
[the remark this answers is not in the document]
adekbadek: Addressed in da7f8617. Added an inner join on `postmeta pm_team_sub` in the Check 3 query so the team's `_subscription_id` comes back in the same result set. No more per-row `get_post_meta` calls, and the join naturally filters out Check 2 cases (teams missing `_subscription_id`), so the in-loop short-circuit is gone too.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,538 @@
...
+		// Bring over the newer _order_id if needed.
+		$dup_order_id      = get_post_meta( $duplicate->ID, '_order_id', true );
+		$original_order_id = get_post_meta( $original->ID, '_order_id', true );
+		if ( $dup_order_id && $dup_order_id !== $original_order_id ) {
+			WP_CLI::line( sprintf( '    COPY _order_id %d to original team #%d', $dup_order_id, $original->ID ) );
+			update_post_meta( $original->ID, '_order_id', $dup_order_id );
+		}
+
+		// Finally delete the now-empty duplicate.
+		$remaining = get_post_meta( $duplicate->ID, '_member_id', true );
```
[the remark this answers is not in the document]
adekbadek: Addressed in da7f8617. Dead line removed.

includes/cli/class-initializer.php
```diff
@@ -79,6 +80,11 @@ public static function register_comands() {
 		WP_CLI::add_command( 'newspack backfill-non-editing-contributors', [ 'Newspack\CLI\Co_Authors_Plus', 'backfill_non_editing_contributor' ] );
 		WP_CLI::add_command( 'newspack migrate-expired-subscriptions', [ 'Newspack\CLI\WooCommerce_Subscriptions', 'migrate_expired_subscriptions' ] );
 
+		WP_CLI::add_command(
+			'newspack teams-for-memberships diagnostics',
+			[ 'Newspack\CLI\Teams_For_Memberships_Diagnostics', 'diagnostics' ]
+		);
```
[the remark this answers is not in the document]
adekbadek: Addressed in da7f8617. Moved the `class_exists( 'WC_Memberships_For_Teams_Loader' )` gate up to the `WP_CLI::add_command` call in `class-initializer.php`, so the command is not registered at all on sites without the Teams plugin. The in-class `WP_CLI::error` stays as a defense-in-depth guard.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,538 @@
...
+	/**
+	 * Load all team posts (optionally scoped to --team-id). Cached per-run.
+	 *
+	 * @return \WP_Post[]
+	 */
+	private static function get_all_teams() {
+		static $cache = null;
+		if ( null !== $cache ) {
+			return $cache;
+		}
```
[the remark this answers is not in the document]
adekbadek: Addressed in da7f8617. Dropped the static cache in `get_all_teams()`; each call now hits the DB. Four teams queries instead of one is a trivial cost for a diagnostic tool and avoids the stale-after-fix problem entirely.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,542 @@
...
+					)
+				);
+				if ( self::$fix ) {
+					self::fix_duplicate_team( $original, $dup );
+				}
+			}
+		}
+
+		self::$counts['Duplicate teams'] = $issue_count;
+	}
```
chickenn00dle, line 247: If `--team-id=<id>` is supplied, this check becomes a no-op — `get_all_teams()` returns a single row, so `count($bucket) > 1` can never be true. The natural way to target a known-duplicated team via `--team-id` would silently report "OK: no duplicates". Could either widen the scope when `--team-id` is set (find all duplicates whose set *includes* the requested ID), or refuse the combination.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,542 @@
...
+		$orphaned_memberships = get_posts(
+			[
+				'post_type'      => 'wc_user_membership',
+				'post_status'    => 'any',
+				'meta_key'       => '_team_id', // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key
+				'meta_value'     => $duplicate->ID, // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+				'posts_per_page' => -1,
+				'fields'         => 'ids',
+			]
+		);
```
chickenn00dle: `posts_per_page => -1` here, and again at lines 298-307 inside `fix_team_missing_subscription_id`, will load every matching `wc_user_membership` in one query. The 298-307 path runs once per team in Check 2, so it's N × unbounded. Publishers that hit this bug tend to have inflated team counts. Could either batch with paging, or pre-build a `team_id => [membership_ids]` map via a single `$wpdb` query before the Check 2 loop.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,542 @@
...
+		$query_args = [
+			'post_type'      => 'wc_memberships_team',
+			'post_status'    => 'any',
+			'posts_per_page' => -1,
+		];
+		if ( self::$team_id ) {
+			$query_args['p'] = self::$team_id;
+		}
+		return get_posts( $query_args );
+	}
```
chickenn00dle, line 657: `get_all_teams()` is called once per check (5x). In read-only mode none of the checks mutate state, so the result can be cached on the class and only re-queried after a destructive check has run.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,542 @@
...
+		if ( self::$team_id ) {
+			WP_CLI::line( sprintf( 'Scoped to team #%d.', self::$team_id ) );
+		}
+		WP_CLI::line( '' );
+
+		self::check_duplicate_teams();
+		self::check_teams_missing_subscription_id();
+		self::check_memberships_missing_subscription_id();
+		self::check_team_owner_subscription_mismatch();
+		self::check_subscription_line_items_missing_team_meta();
```
chickenn00dle: Worth wrapping this pipeline in try/catch — if Check 2 throws, Checks 3–5 don't run and the summary doesn't print, so you lose visibility into the partial state.

includes/cli/class-teams-for-memberships-diagnostics.php
```diff
@@ -0,0 +1,542 @@
...
+		}
+
+		self::$fix     = isset( $assoc_args['fix'] );
+		self::$team_id = isset( $assoc_args['team-id'] ) ? (int) $assoc_args['team-id'] : null;
+		self::$counts  = [];
+
+		if ( self::$fix ) {
+			$skip_confirm = isset( $assoc_args['yes'] );
+			if ( ! $skip_confirm ) {
+				WP_CLI::confirm( 'Apply fixes to the database? This will modify team, membership, subscription and order-item records.' );
```
chickenn00dle: The confirm runs before the checks, so a `--fix` invocation that finds 0 issues still prompts. Could run all checks dry first, summarize, then prompt — or just skip the prompt when there's nothing to fix.

chickenn00dle (commented):
Tested locally and the command works as advertised — nice cleanup. A few non-blocking thoughts inline; feel free to take or leave.

chickenn00dle (approved):

includes/cli/class-teams-for-memberships-diagnostics.php
adekbadek: Addressed in 159a5f6. Moved the confirmation prompt to a just-in-time `ensure_fix_confirmed()` helper that fires from each fix path on first use. `--fix` on a clean site now exits with "No issues found." without prompting.

includes/cli/class-teams-for-memberships-diagnostics.php
adekbadek: Addressed in 159a5f6. Wrapped each check call in a try/catch loop in `diagnostics()`. A throwing check now logs a warning and the remaining checks plus the summary footer still run.

includes/cli/class-teams-for-memberships-diagnostics.php
adekbadek: Addressed in 159a5f6. Restored a static cache on `get_all_teams()` and added an explicit `invalidate_caches()` call from `fix_duplicate_team()` (the only destructive path that mutates the team set). The four read-only callers now share a single query.

includes/cli/class-teams-for-memberships-diagnostics.php
adekbadek: Addressed in 159a5f6. Replaced the N times unbounded `get_posts()` lookup in `fix_team_missing_subscription_id` with one shared `get_team_memberships_map()` SQL query (`postmeta JOIN postmeta`), pre-built once and reused. The relink loop in `fix_duplicate_team` uses the same map.

includes/cli/class-teams-for-memberships-diagnostics.php
adekbadek: Addressed in 159a5f6. Check 1 now detects `--team-id` and widens the query to `author + title` of the requested team, so the duplicate set including that ID is found. If the supplied ID is not a `wc_memberships_team` post the check skips with a clear message.

PR #4667, opened 2026-04-15
feat(access-control): human-readable group subscription names

dkoo:
## Summary

Adds an editable field to set human-readable names for Group subscriptions (which defaults to `<subscription owner>'s Group`), and adds admin UX enhancements for the WooCommerce > Subscriptions list view:

Subscription edit page:
1888d914-1a01-4bd0-aec7-5dee6bfbb1ee

Subscriptions list view:
159df88d-5683-47e6-a62e-6d49c94bfe50

- **Group name in Subscription column**: Group-enabled subscriptions show `<group name> (<x> of <y> members)` instead of the standard `#<id> for <name>` format, with `y` showing the member limit or "unlimited" if no limit is set.
- **Group subscription filter dropdown**: A new dropdown filter alongside the existing product/customer/payment method filters with options for "Group subscriptions" and "Non-group subscriptions". Captures subscriptions with group enabled directly and those inheriting group status from their (non-variable) product, while respecting subscription-level opt-outs. Supports both HPOS and legacy CPT modes.
- **Group name in search**: The group name meta is now searchable from the subscription search box, covering the HPOS orders table, the WCS data store search, and standard `WP_Query` searches.
- **Unit tests**: 4 new tests covering the group name setting (default name, custom override, update-and-read-back, empty fallback).

### Group filter precedence

The filter respects the same precedence as `get_subscription_settings()`:

1. Subscription meta `enabled = yes` → included (explicit opt-in)
2. Subscription meta `enabled = no` → excluded (explicit opt-out overrides product)
3. No subscription meta + product meta `enabled = yes` → included (inheritance)
4. No subscription meta + no product meta → excluded

Variable parent products are skipped in the inheritance path since variations have their own independent group settings.

### Performance

- The group subscription ID list is cached in a 5-minute transient (`newspack_group_subscription_ids`)
- Cache is invalidated when a subscription's or product's group settings change
- Queries use WCS API functions (`wcs_get_subscriptions`, `wcs_get_subscriptions_for_product`) which properly handle meta queries in both CPT and HPOS storage modes

### Prerequisites
- A Newspack site with WooCommerce Subscriptions active
- At least one subscription product with "Group subscription" enabled
- At least one active subscription with a custom group name set and members added
- At least one regular (non-group) subscription

### Subscription column
- [ ] Navigate to WooCommerce > Subscriptions
- [ ] Verify group-enabled subscriptions show the group name and member count (e.g. "Acme Newsroom (2 of 10 members)")
- [ ] Verify subscriptions with no member limit show "unlimited" (e.g. "Test Group (0 of unlimited members)")
- [ ] Verify non-group subscriptions still show the standard format (#ID for Name)
- [ ] Verify the group name links to the subscription edit page

### Group filter
- [ ] Verify a "Group subscriptions" / "Non-group subscriptions" dropdown appears in the filter bar
- [ ] Select "Group subscriptions" and verify only group-enabled subscriptions appear (including those inheriting group status from their product)
- [ ] Select "Non-group subscriptions" and verify group subscriptions are excluded
- [ ] Verify the filter works in combination with other filters (product, customer, etc.)
- [ ] Reset to "All subscriptions" and verify all subscriptions appear again
- [ ] Set a subscription's group enabled setting to "no" while its product is group-enabled - verify it's excluded from "Group subscriptions"
- [ ] Test with both HPOS enabled and disabled sites

### Search
- [ ] Search for a group subscription by its custom group name
- [ ] Verify the subscription appears in search results
- [ ] Search for a term that does not match any group name and verify it does not return false positives

### Unit tests
- [ ] Run `n test-php --filter Test_Group_Subscriptions` and verify all 49 tests pass


includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -182,9 +239,17 @@ public static function get_subscription_settings( $subscription ) {
...
+		$owner_name          = trim( $subscription->get_formatted_billing_full_name() );
 		$settings            = self::get_product_settings( $product_id );
 		$settings['enabled'] = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ? \wc_string_to_bool( $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ) : $settings['enabled'];
 		$settings['limit']   = (int) $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true ) ?: $settings['limit']; // phpcs:ignore Universal.Operators.DisallowShortTernary.Found
+		$settings['name']    = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'name', true ) ?
+								$subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'name', true ) :
+								sprintf(
+									/* translators: %s: The subscription owner's name. */
+									__( '%s Group', 'newspack-plugin' ),
+									$owner_name ? $owner_name . '’s' : __( 'Unnamed', 'newspack-plugin' )
```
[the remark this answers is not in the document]
dkoo: This would result in `Unnamed’s Group`, which isn't grammatically correct. Let's skip this for now.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -139,6 +157,45 @@ public static function add_custom_product_pricing_options( $custom_product_prici
...
+		if ( 'order_title' !== $column ) {
+			return $column_content;
+		}
+		if ( ! Group_Subscription::is_group_subscription( $subscription ) ) {
+			return $column_content;
+		}
+		$settings     = self::get_subscription_settings( $subscription );
+		$members      = Group_Subscription::get_members( $subscription );
+		$member_count = count( $members );
+		$limit        = $settings['limit'] > 0
```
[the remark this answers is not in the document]
dkoo: This will be computed only for subscriptions in the current view. Default max is 20 and I tested at up to 200, so I don't think this is a huge concern.

tests/unit-tests/content-gate/group-subscriptions.php
```diff
@@ -425,6 +429,101 @@ public function test_get_group_subscriptions_for_user_with_no_memberships() {
...
+	public function test_group_name_defaults_to_owner_name() {
+		$owner_id  = $this->create_reader_user();
+		$group_sub = $this->create_group_subscription( $owner_id );
+
+		// Set billing name on the subscription so get_formatted_billing_full_name() returns a real name.
+		$group_sub->data['billing_first_name'] = 'Jane';
+		$group_sub->data['billing_last_name']  = 'Doe';
+
+		$settings = Group_Subscription_Settings::get_subscription_settings( $group_sub );
+
```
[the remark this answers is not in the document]
dkoo: Let's address this when it becomes an issue in the future, if ever.

tests/unit-tests/content-gate/group-subscriptions.php
```diff
@@ -425,6 +429,101 @@ public function test_get_group_subscriptions_for_user_with_no_memberships() {
...
+	 * Test get_subscription_settings() returns a default name based on the
+	 * subscription owner's billing name when no custom name is set.
+	 */
+	public function test_group_name_defaults_to_owner_name() {
+		$owner_id  = $this->create_reader_user();
+		$group_sub = $this->create_group_subscription( $owner_id );
+
+		// Set billing name on the subscription so get_formatted_billing_full_name() returns a real name.
+		$group_sub->data['billing_first_name'] = 'Jane';
+		$group_sub->data['billing_last_name']  = 'Doe';
```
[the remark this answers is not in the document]
dkoo: Let's not worry about this for now since fixing this would just be adding more methods to the mock.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +467,271 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+							AND oim.meta_key = '_product_id'
+						WHERE oim.meta_value IN ($placeholders)", // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare, WordPress.DB.PreparedSQL.NotPrepared
+						...$product_ids
+					)
+				)
+			);
+		}
+
+		return array_values( array_unique( array_merge( $enabled_ids, $product_sub_ids ) ) );
+	}
```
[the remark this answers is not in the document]
dkoo: 14f848c addresses this by caching the query results in a transient for 5 minutes. I'm not sure this is necessary, but it might help with sites that have many (thousands) of Group subscriptions.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +467,271 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+		}
+		$term = $query->get( 's' );
+		if ( empty( $term ) ) {
+			return $search;
+		}
+		$like    = '%' . $wpdb->esc_like( $term ) . '%';
+		$or_clause = $wpdb->prepare( ' OR ( np_group_name.meta_value LIKE %s )', $like );
+
+		// Insert the OR clause before the closing parenthesis of the search condition.
+		$search = preg_replace( '/\)\s*$/', $or_clause . ' )', $search );
```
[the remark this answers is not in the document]
dkoo: Addressed in 14f848c.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +467,271 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+						FROM {$wpdb->prefix}woocommerce_order_items oi
+						INNER JOIN {$wpdb->prefix}woocommerce_order_itemmeta oim
+							ON oi.order_item_id = oim.order_item_id
+							AND oim.meta_key = '_product_id'
+						WHERE oim.meta_value IN ($placeholders)", // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare, WordPress.DB.PreparedSQL.NotPrepared
+						...$product_ids
+					)
+				)
+			);
+		}
```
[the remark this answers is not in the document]
dkoo: Addressed in 14f848c.

dkoo:
Note: pushed fixes to handle some edge cases:

- 5e9f22f fixes querying for group and non-group subscriptions on sites not using HPOS
- 6acdadf fixes querying for product variations—these should not inherit Group settings from their parent product or any other sibling variations
- 5772d59 fixes an edge case where a subscription has the "enabled" setting set to "no" when its product is set to "yes"—the subscription-level "no" setting should override the product-inherited setting.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +475,354 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+	 *
+	 * Collects IDs from two sources:
+	 * 1. Subscriptions with the group enabled meta set directly.
+	 * 2. Subscriptions whose product has group subscriptions enabled (inheritance).
+	 *
+	 * Results are cached in a transient for 5 minutes.
+	 *
+	 * @return int[] Array of subscription IDs.
+	 */
+	public static function get_group_subscription_ids() {
```
adekbadek, line 647: **Perf scaling.** Each cache miss runs three queries with `posts_per_page => -1` and `subscription_status => 'any'`: subs with `enabled=yes`, subs with `enabled=no`, and all `product`/`product_variation` posts with `enabled=yes`, then loops `wcs_get_subscriptions_for_product()` per product. On a publisher with thousands of subscriptions/products this stalls the admin list. Consider building the ID list in a single SQL/joined-meta query, or skipping the full ID list for `non-group` mode and using `meta_query` with `compare = NOT EXISTS` / `value = no` instead.

(Combined with the cache-invalidation note below, this is the most worth-fixing finding — every save invalidates the cache, so the cold path is hit constantly.)

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -225,6 +293,11 @@ public static function update_subscription_settings( $subscription, $settings )
 		}
 		if ( $should_save ) {
 			$subscription->save();
+
+			// Clear the cached group subscription IDs if the enabled setting changed.
+			if ( isset( $settings['enabled'] ) ) {
```
adekbadek: **Cache invalidation is too eager.** The meta box always posts `_newspack_group_subscription_enabled` (whether checked or not), so `isset( $settings['enabled'] )` is always true on subscription save. The transient is busted on every save, even when the value didn't change — defeating the 5-min TTL and exposing the perf cost in `get_group_subscription_ids()` above.

Clear only when the value actually changed (compare against `$previous_settings['enabled']`).

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -39,6 +40,26 @@ public static function init() {
...
+		// Group subscription filter dropdown on subscription list table.
+		\add_action( 'woocommerce_order_list_table_restrict_manage_orders', [ __CLASS__, 'add_group_subscription_filter' ] );
+		\add_action( 'restrict_manage_posts', [ __CLASS__, 'add_group_subscription_filter' ] );
+
+		// Filter subscription list table query by group status.
+		\add_filter( 'woocommerce_shop_subscription_list_table_prepare_items_query_args', [ __CLASS__, 'filter_subscriptions_by_group' ] );
+		\add_filter( 'request', [ __CLASS__, 'filter_subscriptions_by_group_cpt' ] );
+
+		// Clear group subscription IDs cache when product group settings change.
+		\add_action( 'woocommerce_process_product_meta', [ __CLASS__, 'maybe_clear_cache_on_product_save' ] );
```
adekbadek, line 56: **Missing cache invalidation on subscription delete/trash.** When a group subscription is trashed/deleted, the cached ID list lingers for up to 5 minutes — `Group subscriptions` will still match against IDs that no longer exist. Consider adding `wp_trash_post` / `before_delete_post` hooks (gated on `'shop_subscription' === get_post_type( $post_id )`) to clear the transient.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +475,354 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+		$term = $query->get( 's' );
+		if ( empty( $term ) ) {
+			return $search;
+		}
+		$like      = '%' . $wpdb->esc_like( $term ) . '%';
+		$or_clause = $wpdb->prepare( ' OR ( np_group_name.meta_value LIKE %s )', $like );
+
+		// Insert the OR clause inside the existing grouped search condition.
+		// WP's search clause can end with )) or ) depending on search terms.
+		if ( preg_match( '/\)\)\s*$/', $search ) ) {
```
adekbadek: **Fragile search rewrite.** `WP_Query::parse_search()` produces clauses shaped differently with `exact`, `sentence`, multi-term, or other plugins' `posts_search` filters running before this one — the exact bracketing has shifted between WP versions. When neither regex matches, `preg_replace` silently returns the original `$search` unchanged: search-by-group-name fails *quietly* with no signal.

Safer alternatives:
- Wrap the original clause from outside: `( {$search} OR ( np_group_name.meta_value LIKE %s ) )`.
- Or run a separate ID lookup and inject via `pre_get_posts` `post__in`.

Also worth covering this code path with a unit test — it's the trickiest piece in the diff and currently has zero coverage.

adekbadek (commented):
Tested in an isolated env (CPT and HPOS) — column rendering, filter dropdown (Group / Non-group / Reset), and group-name search all behave per the test plan. Precedence works end-to-end: a subscription with `_newspack_group_subscription_enabled = no` against a group-enabled product is correctly excluded from `Group subscriptions`; subscriptions inheriting from the product are correctly included. Default-name fallback (`<owner>’s Group`) renders, custom names persist and are searchable. CI green on the latest commit. **Outcome: comments only — no blockers.**

Automated reviewers used: deep-review (Opus) and standards-review (superpowers:code-reviewer). Codex was unavailable (CLI not installed locally) so no third opinion this round.

Most worth addressing before merge:

1. **Perf scaling of `get_group_subscription_ids()`** — three full-table scans with `posts_per_page => -1` per cache miss. Inline note on the function.
2. **Cache invalidation** — busts on every subscription save (because the meta box always posts `enabled`), and there is no hook for trash/delete. Inline note.
3. **Regex rewrite of WP's search WHERE** — fragile and silently no-ops on shape changes; also untested.
4. **Test coverage** — the four new tests cover the `name` setting only; the precedence rules and search injection (the most subtle code in the diff) are not exercised.
5. **`wp_kses_post( woocommerce_wp_text_input(...) )`** — `woocommerce_wp_text_input()` echoes and returns `null`, so the kses wrapper is a no-op. On PHP 8.1+ it triggers `preg_replace(): Passing null to parameter #3 ... deprecated` — I observed this live in `debug.log` while the meta box was rendering. Pre-existing pattern in this file, but the PR extends it.

Rest is polish (a11y label on the filter dropdown, missing cache invalidation on subscription delete, the column markup replacing rather than appending, brittle `$_GET['page']` guard for HPOS search, DRY of the two near-identical filter methods, and a few nits) — all in inline notes.

Verified browser screenshots and a screen recording are saved locally; happy to attach if useful.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -182,9 +242,17 @@ public static function get_subscription_settings( $subscription ) {
...
+		$owner_name          = trim( $subscription->get_formatted_billing_full_name() );
 		$settings            = self::get_product_settings( $product_id );
 		$settings['enabled'] = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ? \wc_string_to_bool( $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ) : $settings['enabled'];
 		$settings['limit']   = (int) $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true ) ?: $settings['limit']; // phpcs:ignore Universal.Operators.DisallowShortTernary.Found
+		$settings['name']    = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'name', true ) ?
+								$subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'name', true ) :
+								sprintf(
+									/* translators: %s: The subscription owner's name. */
+									__( '%s Group', 'newspack-plugin' ),
+									$owner_name ? $owner_name . '’s' : __( 'Unnamed', 'newspack-plugin' )
```
adekbadek: **i18n nit on the unnamed-owner fallback.**

With a billing name → `"Jane Doe’s Group"`. With no billing name → `"Unnamed Group"` (no `’s`). The asymmetry is harmless, but composing the phrase in PHP via `sprintf( '%s Group', ... )` means translators can't reorder `X's Group` → `Group de X` in other locales — only the trailing word can vary.

Consider making the entire phrases translatable instead, e.g.:

```php
$settings['name'] = $owner_name
    ? sprintf( __( '%s’s Group', 'newspack-plugin' ), $owner_name )
    : __( 'Unnamed group', 'newspack-plugin' );
```

Minor: the meta is fetched twice on the previous two lines — assigning to a local first would help readability.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -290,6 +363,22 @@ public static function add_group_subscription_options( $subscription ) {
 						<?php \esc_html_e( 'Group subscription enabled', 'newspack-plugin' ); ?>
 					</label>
 				</p>
+				<div class="form-row">
+					<?php
+					echo wp_kses_post(
+						\woocommerce_wp_text_input(
+							[
+								'id'            => self::GROUP_SUBSCRIPTION_META_PREFIX . 'name',
+								'name'          => self::GROUP_SUBSCRIPTION_META_PREFIX . 'name',
```
adekbadek: **`wp_kses_post()` is wrapping a `null` return.** `woocommerce_wp_text_input()` *echoes* its output and returns `void`/`null`. The kses wrapper is dead code, and on PHP 8.1+ it actively triggers:

```
PHP Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /wp-includes/kses.php on line 2018
```

I observed this in `debug.log` repeatedly while the meta box rendered.

Fix: just call `\woocommerce_wp_text_input( [...] );` directly, or wrap the call in `ob_start()` / `ob_get_clean()` if filtering is genuinely intended.

(I see this pattern is pre-existing on the line below and elsewhere in the file — but this PR adds another instance, so worth flagging here.)

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -139,6 +160,45 @@ public static function add_custom_product_pricing_options( $custom_product_prici
...
+		}
+		$settings     = self::get_subscription_settings( $subscription );
+		$members      = Group_Subscription::get_members( $subscription );
+		$member_count = count( $members );
+		$limit        = $settings['limit'] > 0
+			? $settings['limit']
+			: __( 'unlimited', 'newspack-plugin' );
+
+		return sprintf(
+			'<a href="%s"><strong>%s</strong></a> (%s)',
```
adekbadek: **Column override swallows WCS's own markup.** `filter_subscription_column_content()` returns a fully-constructed `<a>`, replacing whatever WCS placed in `order_title` (status pills, preview affordances, anything WCS adds in future). Consider appending alongside `$column_content`, or at minimum reproducing the standard `#ID for Name` prefix so future WCS additions aren't silently lost on group rows.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +475,354 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+
+		if ( 'shop_subscription' !== $order_type ) {
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$selected = isset( $_GET['_newspack_group_subscription'] ) ? \sanitize_text_field( \wp_unslash( $_GET['_newspack_group_subscription'] ) ) : '';
+
+		?>
+		<select name="_newspack_group_subscription" id="_newspack_group_subscription">
```
adekbadek, line 530: **a11y: filter dropdown is missing an accessible label.** Other WC filters in the same toolbar use `<label class="screen-reader-text" for="…">`. Worth matching here for parity (and screen-reader users):

```php
printf( '<label class="screen-reader-text" for="_newspack_group_subscription">%s</label>', esc_html__( 'Filter by group subscription status', 'newspack-plugin' ) );
```

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +475,354 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+	 *
+	 * @return array The meta keys with the group name meta key added.
+	 */
+	public static function add_group_name_hpos_search_field( $meta_keys ) {
+		if ( ! is_admin() ) {
+			return $meta_keys;
+		}
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$page = isset( $_GET['page'] ) ? \sanitize_text_field( \wp_unslash( $_GET['page'] ) ) : '';
+		if ( 'wc-orders--shop_subscription' !== $page ) {
```
adekbadek: **HPOS guard is brittle.** Coupling to the literal page slug `wc-orders--shop_subscription` means a future WC slug change silently disables group-name search — and AJAX/REST contexts that don't carry `page` in the query string skip the filter entirely. Worth at least a `// @TODO replace with a stabler signal if WC exposes one.` comment, or hooking off `current_screen` instead.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -386,13 +475,354 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+	}
+
+	/**
+	 * Filter the subscription list table query by group subscription status (HPOS).
+	 *
+	 * @param array $query_args The query args for the list table.
+	 *
+	 * @return array The filtered query args.
+	 */
+	public static function filter_subscriptions_by_group( $query_args ) {
```
adekbadek, line 777: **DRY.** `filter_subscriptions_by_group()` and `filter_subscriptions_by_group_cpt()` are ~30 lines of mirrored `post__in`/`post__not_in` logic — only the early guard differs. Extract a private helper that takes `array $args` and returns the mutated args; both filters call into it.

tests/unit-tests/content-gate/group-subscriptions.php
```diff
@@ -425,6 +429,101 @@ public function test_get_group_subscriptions_for_user_with_no_memberships() {
 		$this->assertEmpty( $result );
 	}
 
+	// -------------------------------------------------------------------------
```
adekbadek, line 432: **Test coverage gap.** The four new tests exercise the `name` setting (default fallback / custom override / update-and-read-back / empty fallback). None of the substantively new behavior is covered:

- The filter precedence rules: subscription `yes` overrides product, subscription `no` overrides product inheritance, variable parents are skipped, etc.
- `get_group_subscription_ids()` itself.
- Cache invalidation paths.
- Search injection.

Per the project's TDD/spec philosophy, the precedence logic in particular is the highest-value addition — it's the most subtle code in this PR and the most likely to regress quietly.

tests/mocks/wc-mocks.php
```diff
@@ -343,6 +343,12 @@ public function update_dates( $dates ) {
 			$this->data['dates'][ $type ] = $date;
 		}
 	}
+	public function get_formatted_billing_full_name() {
+		$first = $this->data['billing_first_name'] ?? '';
+		$last  = $this->data['billing_last_name'] ?? '';
+		$name  = trim( "$first $last" );
+		return $name ? $name : '';
+	}
```
adekbadek, line 350: Nit: the ternary is a no-op. `trim( '' )` already returns `''`, so `return $name ? $name : '';` simplifies to `return $name;` (or just `return trim( "$first $last" );` directly).

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e9814085b — removed the `wp_kses_post()` wrapper for the new text input. The pre-existing instances elsewhere in the file remain unchanged as out-of-scope for this PR.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e9814085b. Two bugs uncovered while tracing this:

1. `update_subscription_settings()` was comparing `wc_bool_to_string($value)` (`'yes'`/`'no'`) against `$previous_settings[$key]` (`bool`), so for the `enabled` key the strict comparison was always true and `$should_save` flipped on every save — even when nothing changed.
2. The cache-clear condition was `isset( $settings['enabled'] )`, which was always true since the meta box always passes the key.

Now we track which keys actually changed (after normalizing both sides of the comparison) and only invalidate the cache if `enabled` is among them.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e9814085b. Added four hooks to cover both storage modes:

- `wp_trash_post` and `before_delete_post` (CPT) — gated on `'shop_subscription' === get_post_type( $post_id )`
- `woocommerce_trash_subscription` and `woocommerce_delete_subscription` (HPOS)

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e9814085b. Switched to wrapping the existing clause from outside:

```php
$inner  = preg_replace( '/^\s*AND\s+/', '', $search );
$search = " AND ( {$inner} OR {$or_clause} ) ";
```

This is robust against `exact`/`sentence`/multi-term and other plugins' `posts_search` filters running before ours, since we don't rely on the inner shape.

Also added five unit tests covering: non-subscription queries (skip), empty search (skip), simple clause wrapping, multi-term clauses, and special-character escaping.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e9814085b. Made the whole phrases translatable per your suggestion:

```php
if ( $name_meta ) {
    $settings['name'] = $name_meta;
} elseif ( $owner_name ) {
    $settings['name'] = sprintf( __( '%s’s Group', 'newspack-plugin' ), $owner_name );
} else {
    $settings['name'] = __( 'Unnamed group', 'newspack-plugin' );
}
```

Also fetching each meta value into a local once instead of twice.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e9814085b. The column now prepends our group info (wrapped in a `<div class=\"newspack-group-subscription__column-info\">`) before the original `\$column_content`, preserving status pills and any future WCS additions. The block-level wrapper provides the line break naturally.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Deferring this for now. Combined with the cache-invalidation fix in e9814085b (which addresses the cache being busted on every subscription save), the 5-minute transient should hold for the full TTL on most installs and the cold path becomes much rarer.

Re-architecting around a single joined query would mean dropping the WCS API helpers (`wcs_get_subscriptions`, `wcs_get_subscriptions_for_product`) which are useful for correctness across CPT/HPOS modes. The `NOT EXISTS` shortcut for `non-group` mode also doesn't cover the product-inheritance + subscription-level opt-out cases cleanly.

Will track this as a follow-up if it shows up as an actual bottleneck on a publisher with many group subscriptions.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Followed up in 047cb2a10 to also fix the pre-existing instance flagged in your comment (the limit input on line 408). The two other `wp_kses_post()` calls remaining in this file are correct usage — they sanitize actual HTML strings returned from `sprintf` (containing `<a>` and `<span>` tags), not null returns from echo-style helpers.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e50cf8f08. Now uses `get_current_screen()` with `wcs_get_page_screen_id( 'shop_subscription' )` — same pattern as `admin_enqueue_scripts` in this file. Returns the meta keys unchanged in AJAX/REST contexts (where `get_current_screen()` is null), which is the correct no-op since those aren't subscription list-table searches.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in e50cf8f08. Extracted a private `apply_group_filter( $args, $filter, $group_ids )` helper. Both `filter_subscriptions_by_group()` (HPOS) and `filter_subscriptions_by_group_cpt()` (CPT) keep their distinct entry guards and call into the shared helper for the merge logic.

tests/unit-tests/content-gate/group-subscriptions.php
dkoo: Partially addressed in e50cf8f08 — added two tests covering the transient cache hit path and `clear_group_subscription_ids_cache()` invalidation.

The full precedence-rule tests (subscription `yes` overrides product, `no` overrides product inheritance, variable parents skipped, etc.) require real `shop_subscription` and product posts in the test database, since `wcs_get_subscriptions()` and `wcs_get_subscriptions_for_product()` query the actual posts/order tables. Our current `tests/mocks/wc-mocks.php` stores subscriptions in a `$subscriptions_database` global that those WCS functions don't see.

I'll open a follow-up issue to build out the integration test infrastructure needed for end-to-end coverage of `get_group_subscription_ids()`. Agreed it's the most subtle code in this PR and deserves real coverage.

tests/mocks/wc-mocks.php
dkoo: Fixed in e50cf8f08. Simplified to `return trim( "$first $last" );` — the ternary was a workaround for a phpcs rule against short ternaries (`?:`) but isn't needed at all since `trim('')` already returns `''`.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Fixed in 24bfee4fe — added the `<label class="screen-reader-text">` for the dropdown to match the pattern used by other WC filters in the toolbar.

tests/unit-tests/content-gate/group-subscriptions.php
dkoo: Issue created: NPPD-1496

dkoo:
@adekbadek thanks for the review and feedback! I've pushed some changes to address your comments, so this is ready for a re-review.

adekbadek (approved):
Re-reviewed. All 12 inline items addressed across 4 commits:

- Cache invalidation now busts only when `enabled` actually changes; trash/delete hooks added for both CPT (`wp_trash_post`, `before_delete_post`) and HPOS (`woocommerce_trash_subscription`, `woocommerce_delete_subscription`).
- Search clause now wrapped from outside (`AND ( {$inner} OR {$or} )`) — robust against `exact`/`sentence`/multi-term and other plugins' filters; +5 unit tests.
- i18n on group label, `wp_kses_post(null)` cleanup (own input + pre-existing line), a11y `screen-reader-text` label, HPOS guard via `get_current_screen()`, DRY filter helper extracted, `trim` ternary nit fixed.

Two deferrals noted and acceptable IMO:
- 3× full-table scan: rationale is sound — cache hit rate is now real and WCS helpers cover CPT/HPOS correctness; `NOT EXISTS` doesn't cleanly handle product-inheritance + sub-level opt-out.
- Full precedence test coverage tracked in NPPD-1496.

CI green. Approving.

PR #4668, opened 2026-04-16
feat(integrations): redesign configure view

chickenn00dle:
### All Submissions:

Redesigns the per-integration configure view to match the Figma designs from [DSGNEWS-156](https://linear.app/a8c/issue/DSGNEWS-156/integration-configuration-panel).

https://github.com/user-attachments/assets/7359907b-eeae-4f12-b1b8-208c7b03886c

**PHP:**
- Added `Metadata::get_grouped_default_fields()` to return outgoing metadata fields grouped by section (Identity, Registration, Engagement, etc.)
- Added `grouped_options` to the outgoing metadata field in the REST API response (backward-compatible addition alongside existing `options`)

**Frontend:**
- Redesigned `configure-view.js` from a flat single-column layout to a 2-column sectioned layout (matching the content-gates edit pattern)
- Settings split into 3 sections separated by dividers: **Settings** (ESP-specific fields), **Inbound** (incoming metadata checkboxes), **Outbound** (accordion groups by category)
- Save button moved from inline to header actions via `setHeaderData`
- Integration name and description now set dynamically in the section header
- Added `defaultOpen` prop to the shared `Accordion` component

**Behavior change:**
- `Integrations::register_integrations()` no longer force-re-enables ESP on every registration. Previously `self::enable( 'esp' )` ran on every page load regardless of the saved option, which meant a user-driven Disable would not stick across requests. ESP is now auto-enabled only on first registration (when the option is unset). For existing sites where someone disabled ESP, that disable will now persist.

Closes DSGNEWS-156.

1. Enable both feature flags in `wp-config.php` — `NEWSPACK_SYNC_METADATA_VERSION_1` (gates the v1.0 metadata schema) and `NEWSPACK_INTEGRATIONS_SETTINGS_ENABLED` (gates the Audience > Integrations UI) — and ensure Newspack Newsletters is configured with an ESP and at least one active list
2. Navigate to Newspack > Audience > Integrations
3. Click "Configure" on the Newsletter ESP card
4. Verify the 3-section layout: Settings fields at top, Inbound checkboxes in the middle, Outbound accordion groups at the bottom
5. Verify the first Outbound accordion group (Identity) is expanded by default, others are collapsed
6. Verify the Save button appears in the header and works correctly
7. Verify back navigation returns to the integrations list

### Other information:

### DIFF
--- a/includes/reader-activation/class-integrations.php
+++ b/includes/reader-activation/class-integrations.php
@@ -384,14 +384,7 @@ class Integrations {
 			if ( empty( $fields ) ) {
 				continue;
 			}
-			$can_sync = $integration->can_sync( true );
-			// Filter out the "not enabled" error — that's a user-controlled state,
-			// not a requirement. The frontend uses can_sync for unmet prerequisites only.
-			if ( is_wp_error( $can_sync ) ) {
-				$can_sync->remove( 'ras_esp_sync_not_enabled' );
-			}
-			$can_sync_value = ! ( is_wp_error( $can_sync ) && $can_sync->has_errors() );
-			$result[ $id ]  = [
+			$result[ $id ] = [
 				'id'          => $id,
 				'name'        => $integration->get_name(),
 				'description' => $integration->get_description(),
--- a/includes/reader-activation/class-integrations.php
+++ b/includes/reader-activation/class-integrations.php
@@ -391,8 +391,6 @@ class Integrations {
 				'enabled'     => self::is_enabled( $id ),
 				'is_set_up'   => $integration->is_set_up(),
 				'setup_url'   => $integration->get_setup_url(),
-				'can_sync'    => $can_sync_value,
-				'setup_url'   => $integration->get_setup_url(),
 				'settings'    => $integration->get_settings_config(),
 			];
 		}
--- a/includes/reader-activation/class-integrations.php
+++ b/includes/reader-activation/class-integrations.php
@@ -390,10 +390,8 @@ class Integrations {
 			if ( is_wp_error( $can_sync ) ) {
 				$can_sync->remove( 'ras_esp_sync_not_enabled' );
 			}
-			$can_sync_value = is_wp_error( $can_sync ) && $can_sync->has_errors()
-				? $can_sync->get_error_message()
-				: true;
-			$result[ $id ] = [
+			$can_sync_value = ! ( is_wp_error( $can_sync ) && $can_sync->has_errors() );
+			$result[ $id ]  = [
 				'id'          => $id,
 				'name'        => $integration->get_name(),
 				'description' => $integration->get_description(),
@@ -401,6 +399,7 @@ class Integrations {
 				'is_set_up'   => $integration->is_set_up(),
 				'setup_url'   => $integration->get_setup_url(),
 				'can_sync'    => $can_sync_value,
+				'setup_url'   => $integration->get_setup_url(),
 				'settings'    => $integration->get_settings_config(),
 			];
 		}
--- a/includes/reader-activation/class-integrations.php
+++ b/includes/reader-activation/class-integrations.php
@@ -385,6 +385,14 @@ class Integrations {
 				continue;
 			}
 			$can_sync = $integration->can_sync( true );
+			// Filter out the "not enabled" error — that's a user-controlled state,
+			// not a requirement. The frontend uses can_sync for unmet prerequisites only.
+			if ( is_wp_error( $can_sync ) ) {
+				$can_sync->remove( 'ras_esp_sync_not_enabled' );
+			}
+			$can_sync_value = is_wp_error( $can_sync ) && $can_sync->has_errors()
+				? $can_sync->get_error_message()
+				: true;
 			$result[ $id ] = [
 				'id'          => $id,
 				'name'        => $integration->get_name(),
@@ -392,6 +400,7 @@ class Integrations {
 				'enabled'     => self::is_enabled( $id ),
 				'is_set_up'   => $integration->is_set_up(),
 				'setup_url'   => $integration->get_setup_url(),
+				'can_sync'    => $can_sync_value,
 				'settings'    => $integration->get_settings_config(),
 			];
 		}
--- a/includes/reader-activation/class-integrations.php
+++ b/includes/reader-activation/class-integrations.php
@@ -384,6 +384,7 @@ class Integrations {
 			if ( empty( $fields ) ) {
 				continue;
 			}
+			$can_sync = $integration->can_sync( true );
 			$result[ $id ] = [
 				'id'          => $id,
 				'name'        => $integration->get_name(),
--- a/includes/reader-activation/integrations/class-esp.php
+++ b/includes/reader-activation/integrations/class-esp.php
@@ -305,6 +305,14 @@ class ESP extends Integration {
 			);
 		}
 
+		$newsletters_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'newspack-newsletters' );
+		if ( $newsletters_configuration_manager->is_esp_set_up() === false ) {
+			$errors->add(
+				'ras_esp_sync_not_set_up',
+				__( 'ESP sync is not set up.', 'newspack-plugin' )
+			);
+		}
+
 		if ( ! Integrations::is_enabled( $this->get_id() ) ) {
 			$errors->add(
 				'ras_esp_sync_not_enabled',
--- a/includes/reader-activation/integrations/class-esp.php
+++ b/includes/reader-activation/integrations/class-esp.php
@@ -13,6 +13,7 @@ use Newspack\Reader_Activation\Integrations;
 use Newspack\Reader_Activation;
 use Newspack_Newsletters_Contacts;
 use Newspack_Newsletters_Subscription;
+use Newspack\Configuration_Managers;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -49,10 +50,18 @@ class ESP extends Integration {
 	/**
 	 * Get the URL where the user can set up the ESP.
 	 *
+	 * Delegates to the Newsletters configuration manager so the page slug
+	 * lives in one place. Falls back to the same hardcoded URL when the
+	 * configuration manager isn't resolvable yet.
+	 *
 	 * @return string The Newspack Newsletters settings page URL.
 	 */
 	public function get_setup_url() {
-		return admin_url( 'edit.php?post_type=newspack_nl_cpt&page=newspack-newsletters' );
+		$newsletters_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'newspack-newsletters' );
+		if ( is_wp_error( $newsletters_configuration_manager ) ) {
+			return admin_url( 'edit.php?post_type=newspack_nl_cpt&page=newspack-newsletters' );
+		}
+		return $newsletters_configuration_manager->get_settings_url();
 	}
 
 	/**
--- a/includes/reader-activation/integrations/class-esp.php
+++ b/includes/reader-activation/integrations/class-esp.php
@@ -305,14 +305,6 @@ class ESP extends Integration {
 			);
 		}
 
-		$newsletters_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'newspack-newsletters' );
-		if ( $newsletters_configuration_manager->is_esp_set_up() === false ) {
-			$errors->add(
-				'ras_esp_sync_not_set_up',
-				__( 'ESP sync is not set up.', 'newspack-plugin' )
-			);
-		}
-
 		if ( ! Integrations::is_enabled( $this->get_id() ) ) {
 			$errors->add(
 				'ras_esp_sync_not_enabled',
--- a/includes/reader-activation/integrations/class-esp.php
+++ b/includes/reader-activation/integrations/class-esp.php
@@ -13,7 +13,6 @@ use Newspack\Reader_Activation\Integrations;
 use Newspack\Reader_Activation;
 use Newspack_Newsletters_Contacts;
 use Newspack_Newsletters_Subscription;
-use Newspack\Configuration_Managers;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -35,16 +34,16 @@ class ESP extends Integration {
 	}
 
 	/**
-	 * Whether the ESP service provider is configured.
+	 * Whether the ESP integration is ready to sync.
 	 *
-	 * @return bool True if an ESP provider is selected and configured.
+	 * Mirrors the readiness gate used by get_settings_config() so the configure
+	 * UI never advertises a card as set up while the underlying settings call
+	 * short-circuits to an empty config.
+	 *
+	 * @return bool True if an ESP provider is selected and at least one list is active.
 	 */
 	public function is_set_up() {
-		$newsletters_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'newspack-newsletters' );
-		if ( is_wp_error( $newsletters_configuration_manager ) ) {
-			return false;
-		}
-		return (bool) $newsletters_configuration_manager->is_esp_set_up();
+		return Reader_Activation::is_esp_configured();
 	}
 
 	/**
--- a/includes/reader-activation/integrations/class-integration.php
+++ b/includes/reader-activation/integrations/class-integration.php
@@ -809,6 +809,7 @@ abstract class Integration {
 				);
 			}
 			if ( 'outgoing_metadata_fields' === $field['key'] ) {
+				// TODO: Drop $field['options'] for outgoing_metadata_fields once consumers have migrated to grouped_options.
 				$field['options']         = Sync\Metadata::get_default_fields();
 				$field['grouped_options'] = Sync\Metadata::get_grouped_default_fields();
 			}
--- a/includes/reader-activation/integrations/class-integration.php
+++ b/includes/reader-activation/integrations/class-integration.php
@@ -809,7 +809,8 @@ abstract class Integration {
 				);
 			}
 			if ( 'outgoing_metadata_fields' === $field['key'] ) {
-				$field['options'] = Sync\Metadata::get_default_fields();
+				$field['options']         = Sync\Metadata::get_default_fields();
+				$field['grouped_options'] = Sync\Metadata::get_grouped_default_fields();
 			}
 			$config[] = $field;
 		}
--- a/includes/reader-activation/sync/class-metadata.php
+++ b/includes/reader-activation/sync/class-metadata.php
@@ -289,9 +289,11 @@ class Metadata {
 	 * `newspack_ras_metadata_keys` filter are respected. Fields added by the filter
 	 * that don't belong to any class are collected in an "Additional" group.
 	 *
-	 * @return array[] Array of [ 'section' => string, 'fields' => string[] ].
+	 * @return array<int, array{section: string, fields: list<string>}> List of
+	 *   groups, each with a non-empty section label and an ordered list of field
+	 *   names. May be filtered by `newspack_ras_grouped_metadata_fields`.
 	 */
-	public static function get_grouped_default_fields() {
+	public static function get_grouped_default_fields(): array {
 		$classes          = self::get_metadata_classes();
 		$available_fields = array_values( array_unique( array_values( self::get_all_fields( true ) ) ) );
 		$groups           = [];
--- a/includes/reader-activation/sync/class-metadata.php
+++ b/includes/reader-activation/sync/class-metadata.php
@@ -284,27 +284,56 @@ class Metadata {
 	 * Get the list of possible fields to be synced, grouped by section.
 	 *
 	 * Returns an array of groups, each with a 'section' label and 'fields' array.
-	 * Only includes non-legacy classes with a section name.
+	 * Only includes non-legacy classes with a section name. Fields are intersected
+	 * with the filtered available fields list so extensions using the
+	 * `newspack_ras_metadata_keys` filter are respected. Fields added by the filter
+	 * that don't belong to any class are collected in an "Additional" group.
 	 *
 	 * @return array[] Array of [ 'section' => string, 'fields' => string[] ].
 	 */
 	public static function get_grouped_default_fields() {
-		$classes = self::get_metadata_classes();
-		$groups  = [];
+		$classes          = self::get_metadata_classes();
+		$available_fields = array_values( array_unique( array_values( self::get_all_fields( true ) ) ) );
+		$groups           = [];
+		$grouped_fields   = [];
+
 		foreach ( $classes as $class ) {
 			if ( $class::is_available() ) {
 				$section = $class::get_section_name();
 				if ( empty( $section ) ) {
 					continue;
 				}
-				$fields   = array_values( array_unique( array_values( $class::get_fields() ) ) );
-				$groups[] = [
+
+				$fields = array_values( array_unique( array_values( $class::get_fields() ) ) );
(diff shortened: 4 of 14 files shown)


src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -46,32 +77,90 @@ export const ConfigureView = ( { integrations, loading, pendingChanges, saving,
...
 
-	const hasPending = pendingChanges[ integrationId ] && Object.keys( pendingChanges[ integrationId ] ).length > 0;
+	const handleCheckboxListChange = ( fieldKey, currentValue, optionName, checked ) => {
+		const selected = Array.isArray( currentValue ) ? currentValue : [];
+		const newValue = checked ? [ ...selected, optionName ] : selected.filter( f => f !== optionName );
+		onFieldChange( integrationId, fieldKey, newValue );
+	};
 
 	return (
 		<WizardsTab isFetching={ loading }>
```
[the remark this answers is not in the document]
chickenn00dle: This is now populated within the configure view via the `setHeaderData` hook so gonna ignore.

src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -46,32 +77,90 @@ export const ConfigureView = ( { integrations, loading, pendingChanges, saving,
...
+						<Divider alignment="full-width" variant="tertiary" marginTop={ 32 } marginBottom={ 32 } />
+						<Grid columns={ 2 } gutter={ 32 } noMargin>
+							<SectionHeader heading={ 2 } title={ __( 'Outbound', 'newspack-plugin' ) } noMargin />
+							<div>
+								{ ( outboundField.grouped_options || [] ).map( ( group, index ) => {
+									const currentValue = getFieldValue( outboundField );
+									const selected = Array.isArray( currentValue ) ? currentValue : [];
+									return (
+										<Accordion key={ group.section } title={ group.section } defaultOpen={ index === 0 }>
+											<Grid columns={ 1 } rowGap={ 8 } noMargin>
```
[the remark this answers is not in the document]
chickenn00dle: The api should always provide grouped_options so this shouldn't be a concern. I will address the filter problem in the other related comment so resolving this one.

packages/components/src/accordion/index.js
```diff
@@ -17,11 +17,16 @@ import classNames from 'classnames';
...
 	return (
-		<details className={ classNames( 'newspack-accordion', { 'newspack-accordion--is-open': isOpen } ) }>
-			<summary onClick={ () => setIsOpen( ! isOpen ) }>
+		<details className={ classNames( 'newspack-accordion', { 'newspack-accordion--is-open': isOpen } ) } open={ isOpen }>
+			<summary
+				onClick={ e => {
+					e.preventDefault();
+					setIsOpen( ! isOpen );
+				} }
+			>
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in https://github.com/Automattic/newspack-plugin/pull/4668/commits/e025e8821369dc6d55fa094e8eb97ee42e75e8c2

includes/reader-activation/sync/class-metadata.php
```diff
@@ -280,6 +280,33 @@ public static function get_key( $key ) {
...
+					continue;
+				}
+				$fields   = array_values( array_unique( array_values( $class::get_fields() ) ) );
+				$groups[] = [
+					'section' => $section,
+					'fields'  => $fields,
+				];
+			}
+		}
+		return $groups;
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in https://github.com/Automattic/newspack-plugin/pull/4668/commits/e025e8821369dc6d55fa094e8eb97ee42e75e8c2

chickenn00dle:
This one is ready for review, but depends on https://github.com/Automattic/newspack-plugin/pull/4665. I'm gonna add the blocked label until that PR is approved and merged.

src/wizards/audience/views/integrations/settings-section.js
```diff
@@ -1,76 +1,94 @@
...
+/**
+ * Icon configuration per integration ID.
+ * Only ESP exists today. When new integrations are added (DSGNEWS-157),
+ * this moves to the PHP API response.
+ */
+const INTEGRATION_ICONS = {
+	esp: {
+		node: <Icon icon={ envelope } />,
+		fill: '#003da5',
+		backgroundColor: '#dfe7f4',
```
adekbadek: Hardcoded hex values (`#003da5`, `#dfe7f4`, `#757575`, `#f0f0f0`) in the `INTEGRATION_ICONS` / `DEFAULT_ICON` palette. These should come from design tokens / `--wp-admin-theme-color` / WP `wp-colors` Sass vars rather than literals in JSX. The `#003da5` here looks intended as the WP admin theme color, and the greys map to `wp-colors.$gray-700` / `$gray-100`. If `CardFeature`'s icon API only accepts strings, that's worth raising with the component owner.

adekbadek (commented):
The redesign itself works as described in the PR body and matches the Figma intent: 2-column sectioned layout, Save in header, accordion groups with Identity expanded by default, back nav. 

The two issues worth flagging up front: the `is_set_up()` / `get_settings_config()` mismatch and the `enabled && isSetUp` collapse on the card both have real user-visible consequences (empty configure view; hidden Disable control). Plus a small set of suggestions: hardcoded hex colors in the icon palette, dead `onConfigure` ternary, stale `description` prop on the `:integrationId` section, missing PHP test for `Metadata::get_grouped_default_fields()`, undocumented behavior change in `Integrations::register_integrations()`, and the empty `<h1>` from `WizardsTab` on every configure page.

One docs note: the README/PR body mentions only `NEWSPACK_SYNC_METADATA_VERSION_1`, but `Audience_Integrations::is_enabled()` also requires `NEWSPACK_INTEGRATIONS_SETTINGS_ENABLED`. The "How to test" steps don't currently work without that second flag.

No CI failures, no PHP debug-log errors triggered by browser testing.

src/wizards/audience/views/integrations/settings-section.js
```diff
@@ -1,76 +1,94 @@
...
+							const goToSetup = () => {
+								window.location.href = setup_url;
+							};
+							return (
+								<CardFeature
+									key={ id }
+									title={ name }
+									description={ description }
+									icon={ INTEGRATION_ICONS[ id ] || DEFAULT_ICON }
+									enabled={ isEnabled && isSetUp }
```
adekbadek: `enabled={ isEnabled && isSetUp }` collapses the "enabled-but-no-longer-set-up" state into the visual disabled state. Because `CardFeature` only renders `moreControls` (the dropdown that contains "Disable") in the enabled state, a publisher who enabled ESP and then lost the prerequisite (e.g. removed their newsletter provider) can't disable it from this UI. Meanwhile the backend still includes it in `get_active_integrations()` and keeps trying to sync. Drive the card enabled state / Disable affordance from the backend `enabled` flag, not the AND of both.

src/wizards/audience/views/integrations/settings-section.js
```diff
@@ -1,76 +1,94 @@
...
+									enableLabel={ isSetUp ? __( 'Enable', 'newspack-plugin' ) : __( 'Connect', 'newspack-plugin' ) }
+									configureLabel={ needsSetup ? __( 'Configure', 'newspack-plugin' ) : undefined }
+									onEnable={ needsSetup ? goToSetup : () => onToggleEnabled( id, true ) }
+									onConfigure={
+										needsSetup ? goToSetup : () => ( onConfigure ? onConfigure( id ) : history?.push( `/settings/${ id }` ) )
```
adekbadek: `onConfigure` is destructured and used in `onConfigure ? onConfigure( id ) : history?.push( ... )`, but the parent never passes it (`index.js` only spreads `sharedProps`). One branch of the ternary is dead. Either drop the `onConfigure` fallback or actually wire it up; right now it reads like an incomplete escape hatch.

src/wizards/audience/views/integrations/settings-section.js
```diff
@@ -1,76 +1,94 @@
...
+									}
+									badgeText={ undefined }
+									badgeLevel={ undefined }
+									moreControls={
+										isEnabled
+											? [
+													{
```
adekbadek, line 77: `badgeText={ undefined }` and `badgeLevel={ undefined }` are no-ops; `CardFeature` already handles their absence. Drop for clarity.

src/wizards/audience/views/integrations/index.js
```diff
@@ -86,24 +87,34 @@ const AudienceIntegrations = ( props, ref ) => {
...
-					},
+					props: sharedProps,
+				},
+				{
+					path: '/settings/:integrationId',
+					render: ConfigureView,
+					props: sharedProps,
+					backNav: '#/settings',
+					isHidden: true,
+					description: __( 'Syncs reader data with your Newspack Newsletters email service provider.', 'newspack-plugin' ),
```
adekbadek: The `description` prop is hardcoded as the ESP-specific copy (`"Syncs reader data with your Newspack Newsletters email service provider."`) on a generic `:integrationId` route. Today it's redundant for ESP because `setHeaderData`'s `sectionDescription` overrides it (see `wizard/index.js:247`); for any future integration it would silently mislabel the page. Either drop this prop entirely (the store value wins) or set it to something integration-agnostic.

src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -0,0 +1,166 @@
...
+	};
+
+	const handleCheckboxListChange = ( fieldKey, currentValue, optionName, checked ) => {
+		const selected = Array.isArray( currentValue ) ? currentValue : [];
+		const newValue = checked ? [ ...selected, optionName ] : selected.filter( f => f !== optionName );
+		onFieldChange( integrationId, fieldKey, newValue );
+	};
+
+	return (
+		<WizardsTab isFetching={ loading }>
```
adekbadek, line 116: No `title` prop is passed to `WizardsTab`, but `WizardsTab` always renders `<h1>{title}</h1>` — so every valid configure route emits an empty `<h1>` (visible in the live DOM, eats vertical space below the real section header). Either pass a title, switch to a plain wrapper, or guard the `<h1>` in `WizardsTab` itself.

src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -0,0 +1,166 @@
...
+		}
+		const settings = [];
+		let inbound = null;
+		let outbound = null;
+		for ( const field of integration.settings ) {
+			if ( field.key === 'incoming_metadata_fields' ) {
+				inbound = field;
+			} else if ( field.key === 'outgoing_metadata_fields' ) {
+				outbound = field;
+			} else {
```
adekbadek, line 39: This effect's deps include `hasPending` and `saving`, both of which change on every keystroke in any text field. Each change posts a fresh `SET_HEADER_DATA` action and the wizard store fans the update out via `useSelect( getHeaderData )`. Not catastrophic, but you can avoid the per-keystroke churn by separating "set name/description once" from "update the action's `disabled` flag" — or by reading `hasPending`/`saving` inside the action handler via a ref so they aren't effect deps.

Also, the not-found early return below doesn't reset `headerData`, so navigating directly from `/settings/esp` to `/settings/<bogus>` *can* leave the previous integration's name in the breadcrumb in some races (in my test `ResetHeaderData` did fire on path change so this didn't reproduce, but worth defending against).

includes/reader-activation/class-integrations.php
```diff
@@ -215,8 +215,10 @@ public static function register_integrations() {
 		// Hook for other plugins/code to register their integrations.
 		do_action( 'newspack_reader_activation_register_integrations' );
 
-		// hardcode ESP integration as enabled for now.
-		self::enable( 'esp' );
+		// Auto-enable ESP on first registration only.
```
adekbadek, line 220: Real behavior change worth calling out in the PR description: previously `self::enable( 'esp' )` ran on every registration, so ESP was force-re-enabled on every page load regardless of the saved option. Now it only runs when the option is unset (`false`). For any existing site where someone disabled ESP, that disable will now actually stick across requests. That's almost certainly the desired fix, but it's a buried semantic change in a PR titled "redesign configure view" — flagging it explicitly in the body (or splitting it into its own commit) would help reviewers and `git blame`.

includes/reader-activation/integrations/class-esp.php
```diff
@@ -28,11 +29,33 @@ class ESP extends Integration {
...
+	 * Whether the ESP service provider is configured.
+	 *
+	 * @return bool True if an ESP provider is selected and configured.
+	 */
+	public function is_set_up() {
+		$newsletters_configuration_manager = Configuration_Managers::configuration_manager_class_for_plugin_slug( 'newspack-newsletters' );
+		if ( is_wp_error( $newsletters_configuration_manager ) ) {
+			return false;
+		}
+		return (bool) $newsletters_configuration_manager->is_esp_set_up();
```
adekbadek: `is_set_up()` returns true as soon as the ESP has credentials (delegated through `Newsletters_Configuration_Manager::is_esp_set_up()`). But `get_settings_config()` short-circuits with `[]` when `Reader_Activation::is_esp_configured()` is false, which additionally requires at least one active newsletter list. Result: a publisher with a Mailchimp API key but no active list sees an enabled "Configure" card, clicks it, and lands on a configure page with a back arrow, an integration title… and an empty body with a Save button. See screenshot below. Either align the two readiness checks (use the same gate for both), or send the user to Newsletters setup when they're "set up" but not "configured."

11-configure-empty-redesign

includes/reader-activation/integrations/class-esp.php
```diff
@@ -28,11 +29,33 @@ class ESP extends Integration {
...
 			'esp',
-			__( 'ESP', 'newspack-plugin' ),
-			__( 'Sync reader data and activity to the connected email service provider.', 'newspack-plugin' )
+			__( 'Newsletter ESP', 'newspack-plugin' ),
+			__( 'Syncs reader data with your Newspack Newsletters email service provider.', 'newspack-plugin' )
 		);
 	}
 
+	/**
+	 * Whether the ESP service provider is configured.
```
adekbadek: The setup URL is hardcoded as `edit.php?post_type=newspack_nl_cpt&page=newspack-newsletters`. There's already a `Newsletters_Configuration_Manager` for cross-plugin coupling — consider asking it for the URL (e.g. a `get_settings_url()` method) so this silently doesn't break if the newsletters plugin moves the page slug.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -712,7 +737,8 @@ function ( $incoming_field ) {
 				);
```
adekbadek, line 809: Adding `grouped_options` alongside `options` is backward compatible, but the frontend now reads only `grouped_options` for outgoing-metadata and the flat `options` is dead in this code path. A `// TODO: drop $field['options'] for outgoing_metadata_fields once consumers have migrated.` would mark this for removal so it doesn't become permanent dead weight in the REST response.

includes/reader-activation/sync/class-metadata.php
```diff
@@ -280,6 +280,62 @@ public static function get_key( $key ) {
...
+	 *
+	 * Returns an array of groups, each with a 'section' label and 'fields' array.
+	 * Only includes non-legacy classes with a section name. Fields are intersected
+	 * with the filtered available fields list so extensions using the
+	 * `newspack_ras_metadata_keys` filter are respected. Fields added by the filter
+	 * that don't belong to any class are collected in an "Additional" group.
+	 *
+	 * @return array[] Array of [ 'section' => string, 'fields' => string[] ].
+	 */
+	public static function get_grouped_default_fields() {
```
adekbadek: Two requests on this new method:

1. **Add a return type / structured PHPDoc.** It's a public static API consumed by REST and the frontend; declaring `: array` and a shape via `@return array<int, array{section: string, fields: list<string>}>` would lock the contract.
2. **Add at least one PHP test.** This method has non-trivial gating: filters out classes without `get_section_name()`, intersects per-class fields against `get_all_fields( true )` (so `newspack_ras_metadata_keys` is respected), drops empty groups, and rolls filter-only orphans into an "Additional" group. Sibling tests already exist in `tests/unit-tests/reader-activation-sync/`. A test covering (a) section-name gating, (b) the intersection-with-filter-removed-fields path, (c) the "Additional" bucket, (d) the new `newspack_ras_grouped_metadata_fields` filter would lock it down.

src/wizards/audience/views/integrations/configure-view.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.newspack-configure-view {
+	// Align section headers to the top of the grid row.
+	.newspack-grid .newspack-section-header {
+		margin-top: 0;
+		margin-bottom: 0;
+	}
+
+	// Accordion overrides: top/bottom borders only, no box border.
```
adekbadek, line 10: Overriding `.newspack-accordion` borders from a consumer view is a smell — the shared component doesn't own its own border anymore. If "top/bottom borders only, no box border" is a real second variant of the Accordion, push it into `packages/components/src/accordion/` as a `variant` prop rather than reaching in.

src/wizards/audience/views/integrations/configure-view.scss
```diff
@@ -0,0 +1,41 @@
+@use "~@wordpress/base-styles/colors" as wp-colors;
+
+// Settings section: tighten title-to-description spacing and constrain description width.
+.newspack-wizard__content:has(.newspack-audience-integrations) {
```
adekbadek: The `.newspack-wizard__content:has(.newspack-audience-integrations)` selector is feature-scoped CSS reaching out to override the wizard chrome (description color/width, sibling `SectionHeader` margin, etc.). Same pattern in `src/admin/style.scss:198-213`. If those spacings are right for this view, they're probably right elsewhere too — consider lifting them into the wizard's own SCSS or exposing wizard props rather than `:has()` gymnastics.

src/wizards/audience/views/integrations/settings-section.js
chickenn00dle: Updated in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/settings-section.js
chickenn00dle: Done in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/settings-section.js
chickenn00dle: Removed in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/settings-section.js
chickenn00dle: Dropped in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/index.js
chickenn00dle: Dropped in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/configure-view.js
chickenn00dle: Guarded in `WizardsTab` in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/configure-view.js
chickenn00dle: Resolved in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

includes/reader-activation/class-integrations.php
chickenn00dle: Updated PR description with a **Behavior Change** section

includes/reader-activation/integrations/class-esp.php
chickenn00dle: Aligned using only `Reader_Activation::is_esp_configured()` in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

includes/reader-activation/sync/class-metadata.php
chickenn00dle: Both of these have been added in https://github.com/Automattic/newspack-plugin/commit/f33818596b6a2e7dd57c1eddd85c23ca3f9d39db

src/wizards/audience/views/integrations/configure-view.scss
chickenn00dle: I'm not sure that this is a real second variant as I don't see this styling in any other wizard page. I'll skip this for now, but if you have a strong preference, I can move this into the accordion component.

includes/reader-activation/integrations/class-esp.php
chickenn00dle: Good idea. Updated in https://github.com/Automattic/newspack-plugin/pull/4668/commits/9c9e84a67ac91c26904b00b946f83e6a6677d621

includes/reader-activation/integrations/class-integration.php
chickenn00dle: Added the comment in https://github.com/Automattic/newspack-plugin/pull/4668/commits/9c9e84a67ac91c26904b00b946f83e6a6677d621

src/wizards/audience/views/integrations/configure-view.scss
chickenn00dle: Lifted in to wizard styles in https://github.com/Automattic/newspack-plugin/pull/4668/commits/9c9e84a67ac91c26904b00b946f83e6a6677d621

chickenn00dle:
Thanks for the review @adekbadek! I think I've resolved all of the feedback. Let me know if I've missed anything!

adekbadek (approved):
Re-reviewed at 9c9e84a6. All findings from the previous round are addressed:

- Hex literals in the icon palette → design tokens (`colors[ 'primary-600' ]` etc.)
- `enabled={ isEnabled && isSetUp }` → `enabled={ isEnabled }`. Browser-confirmed: Disable is now reachable from the More menu in the enabled-but-not-set-up state.
- `is_set_up()` now delegates to `Reader_Activation::is_esp_configured()`, so the readiness gate matches `get_settings_config()`. Browser-confirmed: clicking Configure on an unconfigured ESP redirects to Newsletters Settings instead of an empty form.
- `WizardsTab` `title` is now optional and the `<h1>` is guarded — no more empty heading on the configure page.
- `setHeaderData` split into separate effects for static identity vs. action state, plus an explicit reset on the not-found path.
- Hardcoded setup URL moved to `Newspack_Newsletters_Configuration_Manager::get_settings_url()` with a fallback.
- `Metadata::get_grouped_default_fields()` got `: array`, the "Additional" bucket for filter-only orphans, the `newspack_ras_grouped_metadata_fields` filter, and a new test class with five well-targeted methods (`Test_Grouped_Metadata`).
- TODO comment added next to `grouped_options` so the dead `options` array doesn't become permanent.
- `:has()` SCSS coupling lifted into wizard-level styles.
- PR description updated with a Behavior Change section calling out the `Integrations::register_integrations()` enable() semantic change.
- Bonus: `<Accordion>` refactored to use native `<details onToggle>` instead of intercepting summary clicks — drops two `eslint-disable` directives and improves keyboard a11y.

The Accordion border override in `configure-view.scss` is left as-is, which is a reasonable YAGNI call given there's only one consumer of that visual.

CI green, no PHP debug-log noise from browsing the redesigned views. LGTM.

PR #4670, opened 2026-04-16
fix(access-control): allow `0` member limit values to be shown

dkoo:
### All Submissions:

Fixes a bug in the Group Subscription class where a subscription's override value of `0` for the member limit setting is interpreted by the UI as being not set, and will instead show the inherited product's limit setting.

1. Enable Group features for a product and set the limit to a non-zero value.
2. Create or edit a subscription and set the limit setting for the subscription to `0`. Save.
3. On `trunk`, observe that the `0` value isn't shown—instead, the inherited product value is shown.
4. On this branch, confirm that the `0` value is shown. Also confirm that other Group subscriptions that don't have any override value for the limit option still show the inherited product value.

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
+++ b/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
@@ -227,8 +227,10 @@ class Group_Subscription_Settings {
 		if ( ! $product ) {
 			return $settings;
 		}
-		$settings['enabled'] = $product->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ? \wc_string_to_bool( $product->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ) : self::DEFAULT_SETTINGS['enabled'];
-		$settings['limit']   = (int) $product->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true ) ?: self::DEFAULT_SETTINGS['limit']; // phpcs:ignore Universal.Operators.DisallowShortTernary.Found
+		$enabled_meta        = $product->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true );
+		$limit_meta          = $product->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true );
+		$settings['enabled'] = '' !== $enabled_meta ? \wc_string_to_bool( $enabled_meta ) : $settings['enabled']; // Empty string means the meta is unset; any other value, including 'no' or false, is a real override.
+		$settings['limit']   = '' !== $limit_meta ? (int) $limit_meta : $settings['limit']; // Empty string means the meta is unset; any other value, including '0', is a real override.
 
 		/**
 		 * Filter the group subscription settings for a product.
@@ -257,8 +259,8 @@ class Group_Subscription_Settings {
 		$enabled_meta        = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true );
 		$limit_meta          = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true );
 		$name_meta           = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'name', true );
-		$settings['enabled'] = $enabled_meta ? \wc_string_to_bool( $enabled_meta ) : $settings['enabled'];
-		$settings['limit']   = (int) $limit_meta ?: $settings['limit']; // phpcs:ignore Universal.Operators.DisallowShortTernary.Found
+		$settings['enabled'] = '' !== $enabled_meta ? \wc_string_to_bool( $enabled_meta ) : $settings['enabled']; // Empty string means the meta is unset; any other value, including 'no' or false, is a real override.
+		$settings['limit']   = '' !== $limit_meta ? (int) $limit_meta : $settings['limit']; // Empty string means the meta is unset; any other value, including '0', is a real override.
 		if ( $name_meta ) {
 			$settings['name'] = $name_meta;
 		} elseif ( $owner_name ) {
--- a/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
+++ b/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
@@ -14,6 +14,16 @@ use Newspack\Group_Subscription_Settings;
 class Test_Group_Subscription_Settings extends WP_UnitTestCase {
 
 	/**
+	 * Set up test fixtures.
+	 */
+	public static function set_up_before_class() {
+		parent::set_up_before_class();
+
+		// Include WC mocks.
+		require_once dirname( __DIR__, 4 ) . '/mocks/wc-mocks.php';
+	}
+
+	/**
 	 * Tear down: reset subscriptions and products databases.
 	 */
 	public function tear_down() {
--- a/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
+++ b/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
@@ -44,52 +44,70 @@ class Test_Group_Subscription_Settings extends WP_UnitTestCase {
 	}
 
 	/**
-	 * Build a subscription linked to a product with a non-zero group subscription limit.
+	 * Build a subscription linked to a product, optionally setting group subscription
+	 * meta on either side and arbitrary subscription data.
 	 *
-	 * @param int        $product_limit         The product's group subscription limit.
-	 * @param mixed|null $subscription_override If not null, set as the subscription's limit meta override.
+	 * Meta keys are passed without the GROUP_SUBSCRIPTION_META_PREFIX; the helper
+	 * applies the prefix.
+	 *
+	 * @param array $product_meta      Map of meta key => value to set on the product.
+	 * @param array $subscription_meta Map of meta key => value to set on the subscription.
+	 * @param array $subscription_args Extra arguments merged into the subscription data
+	 *                                 (e.g. billing_first_name, billing_last_name).
 	 *
 	 * @return WC_Subscription
 	 */
-	private function make_subscription_with_product_limit( $product_limit, $subscription_override = null ) {
-		$product_id = 123;
+	private function make_subscription_with_product( $product_meta = [], $subscription_meta = [], $subscription_args = [] ) {
+		$product_id            = 123;
+		$prefixed_product_meta = [];
+		foreach ( $product_meta as $key => $value ) {
+			$prefixed_product_meta[ Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . $key ] = $value;
+		}
 		wc_create_mock_product(
 			[
 				'id'   => $product_id,
-				'meta' => [
-					Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled' => 'yes',
-					Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'limit'   => (string) $product_limit,
-				],
+				'meta' => $prefixed_product_meta,
 			]
 		);
 
 		$subscription = wcs_create_subscription(
-			[
-				'customer_id'    => 1,
-				'status'         => 'active',
-				'billing_period' => 'month',
-				'items'          => [
-					new WC_Order_Item_Product( [ 'product_id' => $product_id ] ),
+			array_merge(
+				[
+					'customer_id'    => 1,
+					'status'         => 'active',
+					'billing_period' => 'month',
+					'items'          => [
+						new WC_Order_Item_Product( [ 'product_id' => $product_id ] ),
+					],
 				],
-			]
+				$subscription_args
+			)
 		);
 
-		if ( null !== $subscription_override ) {
-			$subscription->update_meta_data( Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', $subscription_override );
+		foreach ( $subscription_meta as $key => $value ) {
+			$subscription->update_meta_data( Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . $key, $value );
 		}
 
 		return $subscription;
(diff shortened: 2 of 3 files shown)


includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -184,7 +184,7 @@ public static function get_subscription_settings( $subscription ) {
 		$product_id          = WooCommerce_Subscriptions::get_subscription_product_id( $subscription );
 		$settings            = self::get_product_settings( $product_id );
 		$settings['enabled'] = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ? \wc_string_to_bool( $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ) : $settings['enabled'];
-		$settings['limit']   = (int) $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true ) ?: $settings['limit']; // phpcs:ignore Universal.Operators.DisallowShortTernary.Found
+		$settings['limit']   = '' === $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true ) ? $settings['limit'] : (int) $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true );
```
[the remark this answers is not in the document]
dkoo: eeb0cb2c9 refactors so that the `subscription->get_meta(...)` is called only once and cached in a local var.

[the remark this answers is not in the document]
dkoo: eeb0cb2c9 adds a unit test for this scenario.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -0,0 +1,57 @@
...
+				'billing_period' => 'month',
+			]
+		);
+		$sub->update_meta_data( Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', 'yes' );
+		// Set the limit override to 0 (string, as stored by WooCommerce meta).
+		$sub->update_meta_data( Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', '0' );
+
+		$settings = Group_Subscription_Settings::get_subscription_settings( $sub );
+
+		remove_filter( 'newspack_group_subscription_product_settings', $filter );
```
adekbadek: I think this test doesn't actually catch the regression as currently written. The mock subscription has no items, so `WooCommerce_Subscriptions::get_subscription_product_id()` returns `false`, and `Group_Subscription_Settings::get_product_settings(false)` short-circuits at the `if ( ! $product )` guard *before* the `newspack_group_subscription_product_settings` filter is applied. The inherited product limit ends up being `DEFAULT_SETTINGS['limit']` (0), not 10 — so `assertSame( 0, $settings['limit'] )` passes whether the production fix is in place or not (I confirmed by re-running the test against the unpatched logic).

To make it a real regression guard, the inherited limit needs to actually reach `get_subscription_settings` as a non-zero value. Probably easiest to add a real product item to the subscription via `$sub->add_product( $product )` and put `_newspack_group_subscription_limit => 10` on the product so `get_product_settings` runs end-to-end. Hooking `newspack_group_subscription_settings` (the post-merge filter) instead of the product-level one would also work.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
+ * Tests for Group_Subscription_Settings.
+ *
+ * @package Newspack\Tests
+ * @group group-subscription-settings
+ */
+
+use Newspack\Group_Subscription_Settings;
```
adekbadek, line 9: Every other test file in the suite that uses the WC mocks has an explicit `require_once` for `tests/mocks/wc-mocks.php`. This file works today only because the sibling `class-woocommerce-subscriptions.php` loads them first during PHPUnit discovery, which is load-order coupling. Worth adding for robustness:

```php
require_once __DIR__ . '/../../../../mocks/wc-mocks.php';
```

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -0,0 +1,57 @@
...
+		$sub->update_meta_data( Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', 'yes' );
+		// Set the limit override to 0 (string, as stored by WooCommerce meta).
+		$sub->update_meta_data( Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', '0' );
+
+		$settings = Group_Subscription_Settings::get_subscription_settings( $sub );
+
+		remove_filter( 'newspack_group_subscription_product_settings', $filter );
+
+		$this->assertSame( 0, $settings['limit'], 'A subscription limit of 0 should override the product limit of ' . $product_limit );
+	}
```
adekbadek, line 146: Once the inherited-value problem above is sorted, consider rounding out the matrix with the other two branches the refactor touches: (a) override unset → inherits the product limit, (b) non-zero override (e.g. `'5'` over product limit `10` → `5`). Cheap to add and they'd lock in the full contract.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
+ * Tests for Group_Subscription_Settings.
+ *
+ * @package Newspack\Tests
+ * @group group-subscription-settings
```
adekbadek: Nit – `@group group-subscription-settings` is unique to this file. Sibling test uses `@group WooCommerce_Subscriptions_Integration` – consider reusing for consistency, or dropping the annotation.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * Tear down: reset subscriptions and products databases.
+	 */
+	public function tear_down() {
+		global $subscriptions_database, $products_database;
+		$subscriptions_database = [];
+		$products_database      = [];
+		parent::tear_down();
+	}
```
adekbadek, line 44: Nit – the sibling `class-woocommerce-subscriptions.php` test resets these globals in `set_up()`, which isolates the fixture against pollution from earlier tests in the same suite. Adopting the same convention here would be a touch safer than only resetting in `tear_down()`.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -0,0 +1,57 @@
...
+			$settings['enabled'] = true;
+			$settings['limit']   = $product_limit;
+			return $settings;
+		};
+		add_filter( 'newspack_group_subscription_product_settings', $filter );
+
+		$sub = wcs_create_subscription(
+			[
+				'customer_id'    => 1,
+				'status'         => 'active',
```
adekbadek: Nit – per the project convention of using descriptive variable names in tests, `$sub` and `$filter` are on the generic side. `$subscription` (mirrors the production code) and `$product_settings_filter` would read more clearly.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -184,7 +184,11 @@ public static function get_subscription_settings( $subscription ) {
 		$product_id          = WooCommerce_Subscriptions::get_subscription_product_id( $subscription );
 		$settings            = self::get_product_settings( $product_id );
 		$settings['enabled'] = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ? \wc_string_to_bool( $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled', true ) ) : $settings['enabled'];
-		$settings['limit']   = (int) $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true ) ?: $settings['limit']; // phpcs:ignore Universal.Operators.DisallowShortTernary.Found
+		$limit               = $subscription->get_meta( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit', true );
```
adekbadek: While you're in here: the same short-ternary pattern still lives one method up at line 161 in `get_product_settings()`:

```php
$settings['limit'] = (int) $product->get_meta( ... ) ?: self::DEFAULT_SETTINGS['limit'];
```

It's currently benign because `DEFAULT_SETTINGS['limit']` is `0` (so falsy-collapsing `0` to `0` is a no-op), but if that default ever changes to a non-zero value the same bug resurfaces. Worth either mirroring the explicit check here or leaving a follow-up note – up to you.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -184,7 +184,11 @@ public static function get_subscription_settings( $subscription ) {
...
+		// Allow 0 as a valid override limit.
```
adekbadek: Nit – the comment describes the symptom; documenting the contract reads a bit better imo:

```suggestion
		// Empty string means the meta is unset; any other value, including '0', is a real override.
```

adekbadek (approved):
Fix is correct and minimal. A few inline suggestions, mostly around the test (I think it currently passes whether or not the production change is in place — easy to fix).

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: a347a7a applies the same logic to the `enabled` meta (respecting any non `''` value as a valid override) and applies that same logic for both meta values to both product and subscription getter methods.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Updated in 17577303a.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Good call—updated in ef5c89fe8.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Updated in ef5c89fe8.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Updated in ef5c89fe8.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Good catch — confirmed locally that the test passed against unpatched logic. Fixed in 8d2ff22eb.

Took the product-fixture route you suggested:

- Patched `wcs_get_canonical_product_id()` in `tests/mocks/wc-mocks.php` so it returns `$item->get_product_id()` when available (the previous `null` was blocking the inheritance chain).
- Replaced the `newspack_group_subscription_product_settings` filter with a real mock product carrying `_newspack_group_subscription_enabled` / `_newspack_group_subscription_limit` meta, linked to the subscription via a `WC_Order_Item_Product` item.

Verified the regression guard works: temporarily reverted the production fix and the zero-override test fails (`Failed asserting that 10 is identical to 0.`), exactly the bug we shipped the fix for. Full PHPUnit suite (1034 tests) still passes with the mock change.

tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
dkoo: Added both in 8d2ff22eb. The shared fixture is now a small helper (`make_subscription_with_product_limit( $product_limit, $subscription_override = null )`) and the matrix reads cleanly:

- `test_inherits_product_limit_when_subscription_override_unset` — override unset → inherits product limit (10).
- `test_zero_subscription_limit_overrides_product_limit` — `'0'` override beats product 10 (the regression guard).
- `test_nonzero_subscription_limit_overrides_product_limit` — `'5'` override beats product 10.

The two new cases don't strictly catch the original bug, but together they pin down the full inheritance/override contract so a future refactor can't silently break either branch.

dkoo:
Pushed 167f14137 to round out unit coverage for the rest of `Group_Subscription_Settings::get_subscription_settings()`. The class now has a dedicated test for every branch of every setting it returns:

| Setting | Coverage |
|---|---|
| `limit` | inherits product when unset · `'0'` override beats non-zero product (regression guard) · non-zero override beats product |
| `enabled` | inherits product when unset · `'no'` override beats product `'yes'` · `'yes'` override takes effect when product unset |
| `name` | explicit name meta wins (even when an owner billing name is present) · falls back to `"{Owner}'s Group"` from billing name · falls back to `"Unnamed group"` when neither is set |

Refactored the per-test fixture into a single helper, `make_subscription_with_product( $product_meta, $subscription_meta, $subscription_args )`, that takes arbitrary meta maps (without the `_newspack_group_subscription_` prefix — the helper applies it) plus extra subscription data for the `name`-fallback cases that need billing fields. Each test reads as one line of setup + one assertion.

9 tests, 9 assertions, full PHPUnit suite (1034 tests) still green.

adekbadek (approved):
Test fixture rewritten to use a real product mock so the regression guard now actually fails against unpatched code. Full inheritance/override matrix and `enabled`/`name` coverage round out the suite. Extending the falsy-override symmetry to `enabled` meta is a small scope creep but principled and covered.

PR #4673, opened 2026-04-16
feat(co-authors): support CAP core entity store alongside legacy store

rbcorrales:
### All Submissions:

Adds dual-version support for Co-Authors Plus so the plugin works with both the current released CAP (which uses a `cap/authors` Redux store) and the upcoming CAP 3.8+ (https://github.com/Automattic/co-authors-plus/pull/1217, which replaces the store with the core entity store and resolves coauthors via taxonomy term IDs + a REST endpoint).

When the legacy `cap/authors` store is present, the existing synchronous path is used. When it is absent, the hook reads `getEditedPostAttribute('coauthors')` for term IDs and resolves them via `/coauthors/v1/authors-by-term-ids`. The Query Loop path (reading `newspack_author_info` from the post entity) is unchanged.

Also extracts `useAuthorTokens` from `src/bylines/index.js` into its own module so it can be unit tested independently, and fixes a pre-existing `useSelect` reactivity issue in the bylines sidebar where the selector returned a function rather than values.

Closes [NEWS-1732](https://linear.app/a8c/issue/NEWS-1732), .

This PR should be tested against both the current stable CAP and the new CAP ([co-authors-plus#1217](https://github.com/Automattic/co-authors-plus/pull/1217), branch `issue-1204`). To use the new CAP: check out that branch, run `npm install && npm run build`, and activate the plugin.

**Verify the problem on trunk (before this PR):**

1. With new CAP active, stay on the `trunk` branch of newspack-plugin.
2. Open a post with 2+ coauthors in the editor.
3. The byline block (block theme) renders only the post author, ignoring the other coauthors.
4. The avatar block (block theme) renders only one avatar instead of all coauthors.
5. The custom bylines sidebar only shows the post author in the byline, ignoring the other coauthors. The token picker is empty because only the post_author fallback is available.
6. The console shows a `useSelect hook returns different values... Non-equal value keys: coAuthors` warning.

**Switch to this PR branch and verify the fix:**

**Custom bylines sidebar (any theme):**

7. Open a post with 2+ WP user coauthors assigned via CAP.
8. Open the "Byline" sidebar panel and enable "Enable custom byline."
9. Verify the default byline is auto-populated with all WP user coauthors (guest authors should NOT appear as tokens).
10. Click "Edit byline" and verify the token picker shows the correct WP users.
11. Remove a token, verify it reappears in the available picker. Re-add it.
12. Save and reload. Verify the custom byline persists.

**Guest-only post:**

13. Create a post with only a guest author as the coauthor (no WP users).
14. Enable custom byline. The preview should be empty (not auto-seeded with the post author). This is intentional: guest authors cannot be tokenized with `[Author id=N]`.

**Post-author fallback (no CAP coauthors):**

15. Create a fresh post with default post_author only (do not add coauthors via CAP sidebar).
16. Custom byline toggle should auto-insert "By [post author name]."

**Byline block + Avatar block (block theme):**

17. Activate `newspack-block-theme`.
18. Create a new post, add 2-3 coauthors via CAP's "Authors" sidebar (include at least one guest author if `NEWSPACK_ENABLE_CAP_GUEST_AUTHORS` is defined).
19. Insert `newspack/byline` and `newspack/avatar` blocks.
20. Verify the byline renders all coauthors (e.g. "By Alice, Bob and Guest Author") and the avatar block renders one avatar per coauthor.
21. Verify guest author avatars use the uploaded featured image, not gravatar.
22. Save, reload, and verify data persists. You may briefly see only the post author before the full list renders, depending on network/rendering speed.
23. View the post on the frontend and confirm the server-rendered output matches.

**Author profile block (block theme):**

24. On a single post with multiple coauthors, verify the `newspack-blocks/author-profile` block (contextual mode) displays the correct authors with nested avatar and social blocks.

**Backwards compatibility (legacy CAP):**

25. Switch CAP to the current stable version (e.g., `git checkout 3.7.0 && npm install && npm run build`).
26. Repeat steps 7-24. All should work identically via the legacy `cap/authors` store path.
27. Verify in DevTools: `wp.data.select('cap/authors')?.getAuthors` should be a function on legacy CAP, undefined on new CAP.

**Console:**

28. Check for the absence of the previously-seen `useSelect hook returns different values... Non-equal value keys: coAuthors` warning.

### Other information:

### DIFF
--- a/src/bylines/hooks/use-author-tokens.js
+++ b/src/bylines/hooks/use-author-tokens.js
@@ -1,0 +1,59 @@
+/**
+ * WordPress dependencies
+ */
+import { useSelect } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
+
+/**
+ * Internal dependencies
+ */
+import { useCoAuthors } from '../../shared/hooks/use-coauthors';
+
+const BASE_QUERY = {
+	_fields: 'id,name',
+	context: 'view',
+};
+
+/**
+ * WP user coauthors (or post_author fallback) as `{ id, name }` byline tokens.
+ * Guests are excluded since the `[Author id=N]` shortcode requires a WP user ID.
+ *
+ * @param {number} postId The post ID.
+ * @return {Object[]} Array of `{ id, name }` tokens.
+ */
+export function useAuthorTokens( postId ) {
+	const { authors: coAuthors, isCapAvailable, isLoading } = useCoAuthors( postId );
+
+	const postAuthor = useSelect( select => {
+		const editorStore = select( 'core/editor' );
+		const authorId = editorStore?.getEditedPostAttribute?.( 'author' );
+		if ( ! authorId ) {
+			return null;
+		}
+		return select( coreStore ).getUser( authorId, BASE_QUERY );
+	}, [] );
+
+	// Prevent `insertDefaultByline()` from seeding a truncated byline during new-CAP async resolve.
+	if ( isLoading ) {
+		return [];
+	}
+
+	// When coauthors exist, return only WP users — never fall back to post_author,
+	// which would credit the wrong person on guest-only posts.
+	if ( isCapAvailable && coAuthors && coAuthors.length > 0 ) {
+		const wpUsers = coAuthors
+			.filter( author => author.isGuest === false )
+			.filter( ( author, index, arr ) => arr.findIndex( other => other.id === author.id ) === index );
+		return wpUsers.map( author => ( {
+			id: Number( author.id ),
+			name: author.display_name,
+		} ) );
+	}
+
+	// No coauthors (or CAP inactive) — fall back to the post author.
+	if ( postAuthor ) {
+		return [ postAuthor ];
+	}
+
+	return [];
+}
--- a/src/bylines/hooks/use-author-tokens.js
+++ b/src/bylines/hooks/use-author-tokens.js
@@ -15,14 +15,15 @@ const BASE_QUERY = {
 };
 
 /**
- * WP user coauthors (or post_author fallback) as `{ id, name }` byline tokens.
- * Guests are excluded since the `[Author id=N]` shortcode requires a WP user ID.
+ * WP user coauthors as `{ id, name }` byline tokens. Falls back to the `post_author` only when
+ * no coauthors are present and CAP is inactive (or has no term IDs assigned). Guests are excluded
+ * since the `[Author id=N]` shortcode requires a WP user ID.
  *
  * @param {number} postId The post ID.
  * @return {Object[]} Array of `{ id, name }` tokens.
  */
 export function useAuthorTokens( postId ) {
-	const { authors: coAuthors, isCapAvailable, isLoading } = useCoAuthors( postId );
+	const { authors: coAuthors, isCapAvailable, isLoading, hasCoauthorTermIds } = useCoAuthors( postId );
 
 	const postAuthor = useSelect( select => {
 		const editorStore = select( 'core/editor' );
@@ -35,6 +36,14 @@ export function useAuthorTokens( postId ) {
 
 	// Prevent `insertDefaultByline()` from seeding a truncated byline during new-CAP async resolve.
 	if ( isLoading ) {
+		return [];
+	}
+
+	// New CAP has term IDs assigned but resolution returned no authors — likely a permission-denied
+	// REST response (the `authors-by-term-ids` endpoint requires `edit_others_posts`) or a transient
+	// error. Returning `[]` instead of falling back to `post_author` avoids silently crediting the
+	// wrong person on a post where coauthors are explicitly set.
+	if ( hasCoauthorTermIds && ( ! coAuthors || coAuthors.length === 0 ) ) {
 		return [];
 	}
 
--- a/src/shared/hooks/use-coauthors.js
+++ b/src/shared/hooks/use-coauthors.js
@@ -6,11 +6,10 @@ import { useState, useEffect, useMemo } from '@wordpress/element';
 import { store as coreStore } from '@wordpress/core-data';
 import apiFetch from '@wordpress/api-fetch';
 
-/**
- * CoAuthors Plus store name.
- */
+// CAP_STORE is the legacy Redux store, removed in CAP 3.8+. Its absence triggers the new-CAP path.
 const CAP_STORE = 'cap/authors';
 const COAUTHORS_ENDPOINT = '/coauthors/v1/coauthors';
+const COAUTHORS_BY_TERM_IDS_ENDPOINT = '/coauthors/v1/authors-by-term-ids';
 
 // Module-level cache for guest author avatar URLs, keyed by user_nicename.
 // Prevents duplicate REST requests when components re-mount or when
@@ -22,6 +21,66 @@ const guestAvatarCache = {};
 // mount simultaneously (e.g. Query Loop) and all check the cache before
 // any fetch has completed.
 const inflightRequests = {};
+
+// Term-ID to author object cache (new CAP). A cached `false` means the server omitted the ID.
+const coauthorDetailsCache = {};
+
+// In-flight term-ID fetches keyed by sorted CSV, so concurrent mounts share one REST call.
+const inflightTermResolves = {};
+
+// Stable empty-state reference so setState is a no-op when already empty.
+const EMPTY_AUTHORS = Object.freeze( [] );
+
+/**
+ * Fetch coauthor details for an array of taxonomy term IDs (new CAP).
+ * Deduplicates concurrent requests and caches results by term ID.
+ *
+ * @param {number[]} termIds Array of coauthor taxonomy term IDs.
+ * @return {Promise} Resolves once results are in `coauthorDetailsCache`.
+ */
+function fetchCoauthorsByTermIds( termIds ) {
+	if ( ! termIds || termIds.length === 0 ) {
+		return Promise.resolve();
+	}
+	const uncachedIds = termIds.filter( id => ! ( id in coauthorDetailsCache ) );
+	if ( uncachedIds.length === 0 ) {
+		return Promise.resolve();
+	}
+	const key = [ ...uncachedIds ].sort( ( a, b ) => a - b ).join( ',' );
+	if ( inflightTermResolves[ key ] ) {
+		return inflightTermResolves[ key ];
+	}
+	inflightTermResolves[ key ] = apiFetch( {
+		path: `${ COAUTHORS_BY_TERM_IDS_ENDPOINT }?ids=${ uncachedIds.join( ',' ) }`,
+	} )
+		.then( results => {
+			if ( Array.isArray( results ) ) {
+				results.forEach( author => {
+					if ( author?.termId ) {
+						coauthorDetailsCache[ author.termId ] = author;
+					}
+				} );
+			}
+			// Mark IDs that the server omitted as permanently unresolved (e.g. deleted author).
+			// Only done on a successful response — transient errors leave cache untouched to allow retries.
+			uncachedIds.forEach( id => {
+				if ( ! ( id in coauthorDetailsCache ) ) {
+					coauthorDetailsCache[ id ] = false;
+				}
+			} );
+		} )
+		.catch( error => {
+			// Do not poison the cache on transient errors (network, 5xx). Concurrent calls are
+			// already deduped via `inflightTermResolves`, and leaving the cache untouched lets
+			// the next render retry once the issue clears.
+			// eslint-disable-next-line no-console
+			console.warn( '[Newspack] Failed to resolve coauthor term IDs:', error );
+		} )
+		.finally( () => {
+			delete inflightTermResolves[ key ];
+		} );
+	return inflightTermResolves[ key ];
+}
 
 /**
  * Fetch a single coauthor's avatar URLs, deduplicating concurrent requests.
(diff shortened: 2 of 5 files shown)


src/shared/hooks/use-coauthors.js
```diff
@@ -23,6 +22,66 @@ const guestAvatarCache = {};
...
+			uncachedIds.forEach( id => {
+				if ( ! ( id in coauthorDetailsCache ) ) {
+					coauthorDetailsCache[ id ] = false;
+				}
+			} );
+		} )
+		.catch( error => {
+			// Do not poison the cache on transient errors (network, 5xx). Concurrent calls are
+			// already deduped via `inflightTermResolves`, and leaving the cache untouched lets
+			// the next render retry once the issue clears.
```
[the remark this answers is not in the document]
rbcorrales: Addressed in dd7d88c.

src/shared/hooks/use-coauthors.js
```diff
@@ -83,66 +150,140 @@ function extractNicenameFromLink( link ) {
...
+			// No CAP detected (plugin inactive).
+			return { legacyCapAuthors: null, coauthorTermIds: null, restAuthors: null, isCapAvailable: false };
 		},
 		[ postId, postType, skip ]
 	);
 
+	// Resolve coauthor term IDs to author data via REST (new CAP path only).
+	// Uses a module-level cache + in-flight dedup so concurrent mounts share a single fetch.
+	const [ resolvedTermAuthors, setResolvedTermAuthors ] = useState( EMPTY_AUTHORS );
+	const [ isLoading, setIsLoading ] = useState( false );
```
adekbadek: `isLoading` starts `false` from `useState` and only flips to `true` inside the effect's first run. That leaves an unguarded window between mount and effect firing where `useCoAuthors` returns `{ coauthorTermIds: [15,16,17], isCapAvailable: true, isLoading: false }` even though resolution hasn't started yet. `useAuthorTokens`'s `if ( isLoading ) return []` guard catches the in-flight window but not this pre-flight one — in that window it falls through to the `postAuthor` branch as if resolution were complete.

Reproduced this with source-level instrumentation against CAP 4.0.1: at ~380ms post-mount `useAuthorTokens` returns `[postAuthor]`, at ~467ms `isLoading` flips to `true` and the guard kicks in, at ~669ms `resolvedTermAuthors` lands and tokens become correct. The pre-flight window is small in interactive use (a human won't click that fast) but it's deterministic for code that reads tokens during initial render — and it widens with slow networks or larger `coauthors` arrays.

Fix is to derive `isLoading` synchronously during render rather than via `useState`. Something like:

```js
const ids = termIdsKey === null || termIdsKey === ''
  ? []
  : termIdsKey.split( ',' ).map( Number ).filter( Number.isInteger );
const hasUncached = ids.some( id => ! ( id in coauthorDetailsCache ) );
// Then return isLoading: hasUncached || asyncIsLoading
```

That closes the gap because on the very first render where `termIdsKey === '15,16,17'` and the cache is empty, `isLoading` is already `true` synchronously — no commit ever exposes "have term IDs but isLoading false."

src/shared/hooks/use-coauthors.js
```diff
@@ -23,6 +22,66 @@ const guestAvatarCache = {};
...
+	const uncachedIds = termIds.filter( id => ! ( id in coauthorDetailsCache ) );
+	if ( uncachedIds.length === 0 ) {
+		return Promise.resolve();
+	}
+	const key = [ ...uncachedIds ].sort( ( a, b ) => a - b ).join( ',' );
+	if ( inflightTermResolves[ key ] ) {
+		return inflightTermResolves[ key ];
+	}
+	inflightTermResolves[ key ] = apiFetch( {
+		path: `${ COAUTHORS_BY_TERM_IDS_ENDPOINT }?ids=${ uncachedIds.join( ',' ) }`,
```
adekbadek, line 55: Confirmed against the upstream endpoint: `/coauthors/v1/authors-by-term-ids`'s `permission_callback` is `Endpoints::can_edit_coauthors()`, which delegates to `Co_Authors_Plus::current_user_can_set_authors()` and checks `$current_user->allcaps['edit_others_posts']`. Authors editing their own post (capable of `edit_post`, not `edit_others_posts`) will hit 403, the catch handler logs, `coAuthors` stays empty, and `useAuthorTokens` falls back to `postAuthor` — so the byline/avatar previews and custom-byline seeding silently credit the wrong person despite term IDs being assigned.

Two ways out: surface a read-permitted source (e.g. add a `read_coauthors` permission callback variant, or expose author display data in the `getEntityRecord` post payload), or treat 403 as "preserve current state" rather than "set empty" so consumers don't lose what they had. The existing `console.warn` log doesn't help users — they just see the wrong byline.

src/shared/hooks/use-coauthors.js
```diff
@@ -23,6 +22,66 @@ const guestAvatarCache = {};
...
+					coauthorDetailsCache[ id ] = false;
+				}
+			} );
+		} )
+		.catch( error => {
+			// Do not poison the cache on transient errors (network, 5xx). Concurrent calls are
+			// already deduped via `inflightTermResolves`, and leaving the cache untouched allows
+			// a later call to `fetchCoauthorsByTermIds` to retry after this in-flight request clears.
+			// eslint-disable-next-line no-console
+			console.warn( '[Newspack] Failed to resolve coauthor term IDs:', error );
```
adekbadek: This is the only `console.warn` in the hook, and the test mocks it precisely because it's noisy. Every transient 5xx in production logs to every editor's console, and the empty `coAuthors` already signals failure to consumers. Consider gating behind a debug flag or just dropping it.

src/shared/hooks/use-coauthors.js
```diff
@@ -83,66 +150,140 @@ function extractNicenameFromLink( link ) {
...
 				display_name: author.display || author.value || author.label,
 				user_nicename: author.value,
 				isGuest: author.userType === 'guest-author',
 			} ) );
 		}
 
-		// REST API authors from newspack_author_info.
+		// New CAP authors resolved from term IDs: { id, termId, displayName, userNicename, userType, ... }
+		if ( resolvedTermAuthors && resolvedTermAuthors.length > 0 ) {
+			return resolvedTermAuthors.map( author => ( {
```
adekbadek, line 304: Cross-version normalization: legacy CAP returns numeric `id`, new-CAP returns string `id`. `useAuthorTokens` already coerces with `Number()`, but other consumers of `useCoAuthors` won't necessarily — and the in-file dedupe filter (`other.id === author.id`) silently fails across the version boundary. Coerce at the source so every consumer sees a consistent shape.

src/shared/hooks/use-coauthors.js
```diff
@@ -83,66 +150,140 @@ function extractNicenameFromLink( link ) {
...
- * @return {Object} Authors array and availability state.
+ * @return {Object} `{ authors, isCapAvailable, isLoading }`. `isLoading` is true only while
+ *                  the new-CAP term-ID resolution is in flight. Legacy CAP is synchronous.
  */
 export function useCoAuthors( postId, postType = 'post', skip = false ) {
 	// Return raw store references from useSelect to avoid creating new objects
 	// on every render (which triggers useSelect memoization warnings).
 	// The .map() transformations happen in useMemo below.
-	const { capAuthors, restAuthors, isCapAvailable } = useSelect(
+	const { legacyCapAuthors, coauthorTermIds, restAuthors, isCapAvailable } = useSelect(
```
adekbadek: `coauthorTermIds` returns the raw array reference from `getEditedPostAttribute('coauthors')`, which is a fresh array each render. `useSelect` will see referential changes and re-trigger downstream renders even when IDs haven't actually changed. Joining/serializing inside the selector (or returning a stable empty-state sentinel) avoids the churn — and you already do similar work for `EMPTY_AUTHORS` elsewhere.

src/shared/hooks/use-coauthors.js
```diff
@@ -6,11 +6,10 @@ import { useState, useEffect, useMemo } from '@wordpress/element';
 import { store as coreStore } from '@wordpress/core-data';
 import apiFetch from '@wordpress/api-fetch';
 
-/**
- * CoAuthors Plus store name.
- */
+// CAP_STORE is the legacy Redux store, removed in CAP 3.8+. Its absence triggers the new-CAP path.
```
adekbadek: "removed in CAP 3.8+" reads as version history relative to today. CAP went 3.7 → 4.0.0 with this change (4.0.0 shipped 2026-04-23, 4.0.1 on 2026-04-24), so the version line is already inaccurate. Phrasing it as a state of the world ages better, e.g. "Detection key for the optional Redux store published by CAP <4.0. When present and exposing `getAuthors`, prefer the synchronous path; otherwise fall back to term-ID resolution."

src/bylines/hooks/use-author-tokens.js
```diff
@@ -0,0 +1,59 @@
...
+import { useCoAuthors } from '../../shared/hooks/use-coauthors';
+
+const BASE_QUERY = {
+	_fields: 'id,name',
+	context: 'view',
+};
+
+/**
+ * WP user coauthors (or post_author fallback) as `{ id, name }` byline tokens.
+ * Guests are excluded since the `[Author id=N]` shortcode requires a WP user ID.
```
adekbadek: The summary says "WP user coauthors (or post_author fallback) as `{id, name}` byline tokens" but the actual behavior is more nuanced — when CAP is available with at least one coauthor, the post_author fallback is intentionally suppressed even on a guest-only post (returns `[]`). Worth tightening to "post_author fallback only when no coauthors are present (or CAP is inactive)" so the contract matches the implementation.

adekbadek (approved):
Tested locally against the released CAP 4.0.1 (which contains PR #1217's changes — tags 4.0.0 and 4.0.1 both include the new-CAP commit). On a post with three WP-user coauthors:

- The frontend byline block server-renders correctly: `By Author 1, Author 2 and Editor 1`.
- The CAP Authors sidebar lists all three coauthors.
- The new-CAP path's REST resolution to `/coauthors/v1/authors-by-term-ids` fires as expected and resolves to the right three authors.
- With the resolution complete, "Enable custom byline" auto-populates the default byline with all three WP user tokens.

A couple of correctness things worth tightening before merge — see inline. The two I'd most like a second look at are the `isLoading` race on first commit (reproducible at ~380ms post-mount where `useAuthorTokens` returns the post-author fallback as if resolved data) and the 403 path on the upstream endpoint for users without `edit_others_posts`. Neither is catastrophic but both lead to the byline silently seeding with the wrong author.

src/shared/hooks/use-coauthors.js
chickenn00dle: Added processedKey state in https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a. isLoading now derives synchronously to true when `coauthorTermIdsKey` differs from the last-processed key, closing the gap between mount and effect firing.

src/shared/hooks/use-coauthors.js
chickenn00dle: Added `hasCoauthorTermIds` to the return value so consumers can distinguish "no coauthors set" from "coauthors set but resolution unavailable" (403 / 5xx). https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a

src/shared/hooks/use-coauthors.js
chickenn00dle: Dropped in https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a

src/shared/hooks/use-coauthors.js
chickenn00dle: Coerced id: `Number(author.id)` in the new-CAP map so legacy and new paths return consistent numeric ids. https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a

src/shared/hooks/use-coauthors.js
chickenn00dle: Selector now returns `coauthorTermIdsKey` (stable CSV string) instead of the raw array. https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a

src/shared/hooks/use-coauthors.js
chickenn00dle: Reworded in https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a

src/bylines/hooks/use-author-tokens.js
chickenn00dle: Reworded in https://github.com/Automattic/newspack-plugin/pull/4673/commits/4d95e28effe43882a432a7f3fe8f76e28381c90a

chickenn00dle:
Thanks for the review @adekbadek! This should be ready for another round of review.

adekbadek (approved):
Re-reviewed `4d95e28e`. All prior inline feedback addressed cleanly:

- `isLoading` pre-flight race fixed via synchronous `processedKey` derivation.
- 403 path closed via `hasCoauthorTermIds` + guard in `useAuthorTokens`.
- Selector returns CSV string for stable identity.
- Cross-version `id` shape coerced to `Number`.
- Doc/comment rewords as suggested.

Approving.

PR #4674, opened 2026-04-17
feat(access-control): add Specific posts content rule

dkoo:
### All Submissions:

Adds a new **Specific posts** content rule to Access Control gates so publishers can restrict individual posts or pages by ID, without having to match them through a broader post-type or taxonomy rule.

The rule appears last in the gate editor (after Categories and Tags) and uses a `FormTokenField` that searches posts by title or numeric ID across all post types supported by content gates (defaults: posts, pages, plus anything registered via the `newspack_content_gate_supported_post_types` filter).

Semantically this rule behaves differently from the others: it's an **inclusion override**. If a post's ID is listed in a gate's `specific_posts` rule, the gate applies to that post regardless of whether the gate's other rules (post types, categories, tags) would normally match it. The rule's UI description spells this out: *"Also restrict specific posts, even if not covered by other rules above."* A publisher can still combine it with the other rules — e.g. a gate that restricts all posts in the Entertainment category AND additionally gates a specific page that isn't in Entertainment.

Screenshot 2026-04-17 at 12 37 46 PM

Implementation highlights:

* New rule slug `specific_posts` registered in `Content_Rules::get_content_rules()`, `include_only` (no include/exclude toggle), with a custom REST endpoint.
* New admin REST endpoint `GET /newspack/v1/wizard/newspack-audience-access-control/posts-search` returns `{ id, name, type_label }` for published posts matching a title search or `include` lookup. Numeric searches are treated as post-ID lookups so publishers can paste IDs directly.
* `Content_Restriction_Control::get_post_gates()` now checks `specific_posts` before the existing AND evaluation — a match short-circuits to "gate applies", and the rule is skipped inside the AND loop so it never participates in AND logic. A guard prevents a gate whose only rule is `specific_posts` from being vacuously included when nothing matches.
* The existing `ContentRuleControlTokenField` already supported custom endpoints; this PR just genericizes its variable names and log messages (previously referred to "taxonomy" in places).
* PHPUnit bootstrap now defines `NEWSPACK_CONTENT_GATES` so Content Gates code paths (including the new REST endpoint) can be exercised under test.

Closes NPPD-1457.

1. On a test site with `define( 'NEWSPACK_CONTENT_GATES', true );` in `wp-config.php`, create or edit an Access Control gate (Newspack → Access control).
2. Scroll to the bottom of the rule list — verify **Specific posts** appears last (after Tags) with the description "Also restrict specific posts, even if not covered by other rules above."
3. Toggle Specific posts on. Verify there is no include/exclude toggle (the rule is include-only) and that a search token field is shown.
4. Type a post title — verify autocomplete shows suggestions as `"<id>: <title> (Post)"` for posts and `"<id>: <title> (Page)"` for pages. Select one.
5. Type a numeric post ID — verify the specific post with that ID is returned.
6. Save the gate. Reload the editor — verify the saved tokens re-display with their titles (not raw IDs).
7. Activate the gate (registration rule active, status Published).
8. With the gate configured to restrict **Post types = post** AND **Specific posts = [some page ID]**, in an incognito window:
   * Visit the page listed in Specific posts — the gate should appear (override in effect despite the page not matching the post_types rule).
   * Visit a regular post — the gate should appear (AND-chain still works).
   * Visit a different page not in Specific posts — no gate (override doesn't match, AND-chain fails the page on post_types).
9. With a gate whose ONLY content rule is Specific posts = [some-post-id]:
   * Visit the specific post — gate appears.
   * Visit any other post — no gate (guards against the AND loop vacuously succeeding on zero non-specific rules).
10. Run `composer run phpcs` and the PHPUnit suite (`phpunit --filter Test_Content_Gates`) — both should be clean. The full suite should also pass.

### Other information:

Override semantics and rule ordering are covered by new PHPUnit tests in `tests/unit-tests/content-gate/content-gates.php`. Full `Test_Content_Gates` suite: 37/37 passing. Full newspack-plugin suite: 939/939.

### DIFF
--- a/includes/content-gate/class-content-restriction-control.php
+++ b/includes/content-gate/class-content-restriction-control.php
@@ -154,8 +154,32 @@ class Content_Restriction_Control {
 				continue;
 			}
 
+			// Inclusion override: if this post ID is listed in any specific_posts rule
+			// for this gate, the gate applies regardless of other rules.
+			$specific_match = false;
 			foreach ( $content_rules as $content_rule ) {
-				$is_exclusion = isset( $content_rule['exclusion'] ) && $content_rule['exclusion'];
+				if ( 'specific_posts' === $content_rule['slug']
+					&& ! empty( $content_rule['value'] )
+					&& in_array( (int) $post_id, array_map( 'intval', (array) $content_rule['value'] ), true )
+				) {
+					$specific_match = true;
+					break;
+				}
+			}
+			if ( $specific_match ) {
+				$post_gates[] = $gate;
+				continue;
+			}
+
+			// Standard AND evaluation across remaining rules.
+			$has_non_specific_rule = false;
+			foreach ( $content_rules as $content_rule ) {
+				if ( 'specific_posts' === $content_rule['slug'] ) {
+					// Skip — already evaluated above as an override-only rule.
+					continue;
+				}
+				$has_non_specific_rule = true;
+				$is_exclusion          = isset( $content_rule['exclusion'] ) && $content_rule['exclusion'];
 				if ( $content_rule['slug'] === 'post_types' ) {
 					$post_type = get_post_type( $post_id );
 					if ( $is_exclusion ? in_array( $post_type, $content_rule['value'], true ) : ! in_array( $post_type, $content_rule['value'], true ) ) {
@@ -180,6 +204,13 @@ class Content_Restriction_Control {
 					}
 				}
 			}
+
+			// If the gate ONLY had a specific_posts rule and we got here, it means the
+			// override didn't match — don't include this gate.
+			if ( ! $has_non_specific_rule ) {
+				continue;
+			}
+
 			$post_gates[] = $gate;
 		}
 		return $post_gates;
--- a/includes/wizards/audience/class-audience-content-gates.php
+++ b/includes/wizards/audience/class-audience-content-gates.php
@@ -545,6 +545,9 @@ class Audience_Content_Gates extends Wizard {
 			if ( empty( $ids ) ) {
 				return rest_ensure_response( [] );
 			}
+			// Broader status filter when hydrating saved tokens so the editor
+			// keeps showing items whose status changed since the gate was saved.
+			$args['post_status']    = [ 'publish', 'draft', 'pending', 'private', 'future' ];
 			$args['post__in']       = $ids;
 			$args['posts_per_page'] = min( count( $ids ), 100 );
 			$args['orderby']        = 'post__in';
--- a/includes/wizards/audience/class-audience-content-gates.php
+++ b/includes/wizards/audience/class-audience-content-gates.php
@@ -329,6 +329,30 @@ class Audience_Content_Gates extends Wizard {
 				],
 			]
 		);
+
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/audience-content-gates/posts-search',
+			[
+				'methods'             => 'GET',
+				'callback'            => [ $this, 'posts_search' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+				'args'                => [
+					'search'   => [
+						'type'              => 'string',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'include'  => [
+						'type'              => 'string',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'per_page' => [
+						'type'    => 'integer',
+						'default' => 10,
+					],
+				],
+			]
+		);
 	}
 
 	/**
@@ -487,5 +511,70 @@ class Audience_Content_Gates extends Wizard {
 			$updated_gates[] = $updated_gate;
 		}
 		return rest_ensure_response( $updated_gates );
+	}
+
+	/**
+	 * REST callback: search published posts across all post types supported by content gates.
+	 *
+	 * Returns items in the shape consumed by ContentRuleControlTokenField:
+	 * `[{ id, name, type_label }]`. Supports `search` (by title/content) and `include`
+	 * (comma-separated IDs to hydrate saved tokens regardless of search match).
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 * @return \WP_REST_Response
+	 */
+	public function posts_search( $request ) {
+		$post_types = array_column( Content_Restriction_Control::get_available_post_types(), 'value' );
+
+		$args = [
+			'post_type'      => $post_types,
+			'post_status'    => 'publish',
+			'posts_per_page' => max( 1, min( 50, (int) $request->get_param( 'per_page' ) ) ),
+			'orderby'        => 'title',
+			'order'          => 'ASC',
+			'no_found_rows'  => true,
+		];
+
+		$include = $request->get_param( 'include' );
+		if ( ! empty( $include ) ) {
+			$ids = array_filter( array_map( 'absint', explode( ',', $include ) ) );
+			if ( empty( $ids ) ) {
+				return rest_ensure_response( [] );
+			}
+			$args['post__in']       = $ids;
+			$args['posts_per_page'] = count( $ids );
+			$args['orderby']        = 'post__in';
+		}
+
+		$search = $request->get_param( 'search' );
+		if ( ! empty( $search ) ) {
+			// Numeric search: treat as a post ID lookup.
+			if ( is_numeric( $search ) ) {
+				$args['p'] = absint( $search );
+			} else {
+				$args['s'] = $search;
+			}
+		}
+
+		$query = new \WP_Query( $args );
+
+		$labels = [];
+		foreach ( $post_types as $pt ) {
+			$obj = get_post_type_object( $pt );
+			$labels[ $pt ] = $obj && isset( $obj->labels->singular_name ) ? $obj->labels->singular_name : $pt;
+		}
+
+		$data = array_map(
+			function( $post ) use ( $labels ) {
+				return [
+					'id'         => (int) $post->ID,
+					'name'       => $post->post_title !== '' ? $post->post_title : sprintf( '#%d', $post->ID ),
+					'type_label' => $labels[ $post->post_type ] ?? $post->post_type,
+				];
+			},
+			$query->posts
+		);
+
+		return rest_ensure_response( $data );
 	}
 }
--- a/includes/wizards/audience/class-audience-content-gates.php
+++ b/includes/wizards/audience/class-audience-content-gates.php
@@ -347,8 +347,12 @@ class Audience_Content_Gates extends Wizard {
 						'sanitize_callback' => 'sanitize_text_field',
 					],
 					'per_page' => [
-						'type'    => 'integer',
-						'default' => 10,
+						'type'              => 'integer',
+						'default'           => 10,
+						'minimum'           => 1,
+						'maximum'           => 50,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
 					],
 				],
 			]
@@ -529,7 +533,7 @@ class Audience_Content_Gates extends Wizard {
 		$args = [
 			'post_type'      => $post_types,
 			'post_status'    => 'publish',
-			'posts_per_page' => max( 1, min( 50, (int) $request->get_param( 'per_page' ) ) ),
+			'posts_per_page' => (int) $request->get_param( 'per_page' ),
 			'orderby'        => 'title',
 			'order'          => 'ASC',
 			'no_found_rows'  => true,
@@ -542,7 +546,7 @@ class Audience_Content_Gates extends Wizard {
 				return rest_ensure_response( [] );
 			}
 			$args['post__in']       = $ids;
-			$args['posts_per_page'] = count( $ids );
+			$args['posts_per_page'] = min( count( $ids ), 100 );
 			$args['orderby']        = 'post__in';
 		}
 
--- a/includes/wizards/audience/class-audience-content-gates.php
+++ b/includes/wizards/audience/class-audience-content-gates.php
@@ -332,7 +332,7 @@ class Audience_Content_Gates extends Wizard {
 
 		register_rest_route(
 			NEWSPACK_API_NAMESPACE,
-			'/wizard/audience-content-gates/posts-search',
+			'/wizard/' . $this->slug . '/posts-search',
 			[
 				'methods'             => 'GET',
 				'callback'            => [ $this, 'posts_search' ],
--- a/tests/class-newspack-unit-tests-bootstrap.php
+++ b/tests/class-newspack-unit-tests-bootstrap.php
@@ -84,6 +84,10 @@ class Newspack_Unit_Tests_Bootstrap {
 	 * Load Newspack.
 	 */
 	public function load_newspack() {
+		// Enable the Content Gates feature so its wizard and REST endpoints register under test.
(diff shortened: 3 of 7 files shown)


tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1006,4 +1006,285 @@ public function test_newsletter_content_rule_matches_listed_post() {
...
+				'post_status' => 'draft',
+				'post_title'  => 'Searchable Draft',
+			] 
+		);
+		$published_page = $this->factory->post->create(
+			[
+				'post_status' => 'publish',
+				'post_type'   => 'page',
+				'post_title'  => 'Searchable Page',
+			]
```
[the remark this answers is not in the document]
dkoo: ceb2d2a0f

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1006,4 +1006,285 @@ public function test_newsletter_content_rule_matches_listed_post() {
...
+				'post_status' => 'publish',
+				'post_title'  => 'A',
+			] 
+		);
+		$post_b = $this->factory->post->create(
+			[
+				'post_status' => 'publish',
+				'post_title'  => 'B',
+			] 
+		);
```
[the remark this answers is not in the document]
dkoo: ceb2d2a0f

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1006,4 +1006,285 @@ public function test_newsletter_content_rule_matches_listed_post() {
...
+	 */
+	public function test_posts_search_endpoint_supports_include() {
+		wp_set_current_user( $this->factory->user->create( [ 'role' => 'administrator' ] ) );
+
+		$post_a = $this->factory->post->create(
+			[
+				'post_status' => 'publish',
+				'post_title'  => 'A',
+			] 
+		);
```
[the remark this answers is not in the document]
dkoo: ceb2d2a0f

tests/unit-tests/content-gate/content-gates.php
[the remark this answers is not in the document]
dkoo: ceb2d2a0f

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -488,4 +512,69 @@ public function update_gate_priorities( $request ) {
...
+		$include = $request->get_param( 'include' );
+		if ( ! empty( $include ) ) {
+			$ids = array_filter( array_map( 'absint', explode( ',', $include ) ) );
+			if ( empty( $ids ) ) {
+				return rest_ensure_response( [] );
+			}
+			$args['post__in']       = $ids;
+			$args['posts_per_page'] = count( $ids );
+			$args['orderby']        = 'post__in';
+		}
```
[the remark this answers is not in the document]
dkoo: Since the UI has you selecting posts one ID at a time, I doubt this will grow to the extent where it becomes a performance issue.

tests/unit-tests/content-gate/content-gates.php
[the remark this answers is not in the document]
dkoo: ceb2d2a0f

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -329,6 +329,30 @@ public function register_api_endpoints() {
...
+					'search'   => [
+						'type'              => 'string',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'include'  => [
+						'type'              => 'string',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'per_page' => [
+						'type'    => 'integer',
```
adekbadek: `per_page` lacks `minimum`/`maximum` and a `sanitize_callback`. The runtime `max(1, min(50, ...))` does the right thing but `per_page=0` silently becomes 1 instead of a 400, and the schema doesn't document the bounds. IMO `'minimum' => 1, 'maximum' => 50, 'sanitize_callback' => 'absint'` should be added so callers get a clear validation error.

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -488,4 +512,69 @@ public function update_gate_priorities( $request ) {
...
+		$search = $request->get_param( 'search' );
+		if ( ! empty( $search ) ) {
+			// Numeric search: treat as a post ID lookup.
+			if ( is_numeric( $search ) ) {
+				$args['p'] = absint( $search );
+			} else {
+				$args['s'] = $search;
+			}
+		}
+
```
adekbadek, line 565: When `include` is used to rehydrate saved tokens and one of those IDs points to a post that's since been unpublished/trashed, `post_status => 'publish'` will silently drop it. The editor will quietly lose tokens on reload with no indication that anything is stale. The status filter should be broadened when `include` is set (e.g. include `draft`/`pending`/`private`/`future`), or a notice should be surfaced when an ID can't be hydrated.

includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -488,4 +512,69 @@ public function update_gate_priorities( $request ) {
...
+		];
+
+		$include = $request->get_param( 'include' );
+		if ( ! empty( $include ) ) {
+			$ids = array_filter( array_map( 'absint', explode( ',', $include ) ) );
+			if ( empty( $ids ) ) {
+				return rest_ensure_response( [] );
+			}
+			$args['post__in']       = $ids;
+			$args['posts_per_page'] = count( $ids );
```
adekbadek: When `include` is set, `posts_per_page` is overridden to `count($ids)` with no upper bound. Should be capped (e.g. `min(count($ids), 100)`) so a stray caller passing many hundreds of IDs can't trigger an unbounded query.

tests/class-newspack-unit-tests-bootstrap.php
```diff
@@ -84,6 +84,10 @@ public function __construct() {
 	 * Load Newspack.
 	 */
 	public function load_newspack() {
+		// Enable the Content Gates feature so its wizard and REST endpoints register under test.
+		if ( ! defined( 'NEWSPACK_CONTENT_GATES' ) ) {
+			define( 'NEWSPACK_CONTENT_GATES', true );
+		}
```
adekbadek: Defining `NEWSPACK_CONTENT_GATES` for the whole suite means every other test class now runs with the feature on. If anything in the suite asserts behaviour with the feature off (or relies on the routes/CPT not existing), it'll silently flip. IMO this should be defined in a `setUpBeforeClass()` on the gate-specific test class instead, or gated on a class-level `@group`. Worth confirming there's no test elsewhere that depends on the off state.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1006,4 +1006,285 @@ public function test_newsletter_content_rule_matches_listed_post() {
...
+	 */
+	public function test_specific_posts_rule_is_registered() {
+		$rules = Content_Rules::get_content_rules();
+		$this->assertArrayHasKey( 'specific_posts', $rules, 'specific_posts rule is registered' );
+
+		$rule = $rules['specific_posts'];
+		$this->assertSame( __( 'Specific posts', 'newspack-plugin' ), $rule['name'] );
+		$this->assertSame( [], $rule['default'] );
+		$this->assertTrue( $rule['include_only'], 'specific_posts is include-only (no exclusion mode)' );
+		$this->assertSame( '/' . NEWSPACK_API_NAMESPACE . '/wizard/newspack-audience-access-control/posts-search', $rule['endpoint'], 'endpoint matches the route Task 2 must register' );
```
adekbadek: Tiny one — the assertion message references "Task 2" from the planning doc. That should be dropped in favour of describing what the assertion actually checks (e.g. `'endpoint matches the registered REST route'`); planning artifacts don't survive the work.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1006,4 +1006,285 @@ public function test_newsletter_content_rule_matches_listed_post() {
...
+		$this->assertStringContainsString( __( 'restrict specific posts', 'newspack-plugin' ), $rule['description'], 'description signals override behavior' );
```
adekbadek: `__( 'restrict specific posts', 'newspack-plugin' )` isn't a real translation key (the actual user-facing string is "Also restrict specific posts, even if not covered by other rules above."), so this just registers a phantom string with the i18n machinery. A plain `assertStringContainsString( 'restrict specific posts', $rule['description'] )` does the same job without that.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1006,4 +1006,285 @@ public function test_newsletter_content_rule_matches_listed_post() {
...
+
+		// Must be the LAST rule in the list.
+		$keys = array_keys( $rules );
+		$this->assertSame( 'specific_posts', end( $keys ), 'specific_posts appears last' );
```
adekbadek: Asserting position via `end($keys)` couples this data-layer test to UI ordering — anyone adding a new rule slug below `specific_posts` for unrelated reasons will fail this test. IMO the "appears last in the editor" assertion belongs in a UI-layer test, or this one should be relaxed to "specific_posts is registered".

adekbadek (changes requested):
Nice rule, the override semantics are clean and the tests read well as a spec. Left one blocker inline on `class-content-restriction-control.php` (the editor-side matcher in `post-settings.js` hasn't kept pace with this new override — gates with `specific_posts` are reported as not applying in the post-editor sidebar even when the backend correctly restricts the post).

A few smaller suggestions inline alongside it (REST schema bounds, behaviour around stale `include` IDs, and a couple of test-only nits).

includes/content-gate/class-content-restriction-control.php
```diff
@@ -154,8 +154,32 @@ public static function get_post_gates( $post_id = null ) {
...
+					&& in_array( (int) $post_id, array_map( 'intval', (array) $content_rule['value'] ), true )
+				) {
+					$specific_match = true;
+					break;
+				}
+			}
+			if ( $specific_match ) {
+				$post_gates[] = $gate;
+				continue;
+			}
```
adekbadek, line 172: Heads-up that the editor-side matcher hasn't kept pace with this backend override.

`gateMatchesPost()` in `src/content-gate/editor/post-settings.js` (not touched by this PR) only knows about `post_types` and entries in `taxonomyMap`. Any rule slug it doesn't recognise short-circuits the gate to "doesn't apply" — so a gate using `specific_posts` is reported in the post-editor sidebar as not applying to a post, even when this branch above correctly restricts it.

Reproduced on a test site: a gate with `specific_posts: [5]` correctly returns from `Content_Restriction_Control::get_post_gates(5)` but the editor panel for page 5 shows "No gates apply to this post." Editors using this rule will be told their content is unrestricted while readers are actually being gated — a trust/visibility problem more than a functional one.

The JS matcher should be updated to mirror the new override semantics: if the current post id is in `rule.value`, short-circuit to "applies"; if a gate's only rule is `specific_posts` and the post isn't listed, don't apply.

includes/wizards/audience/class-audience-content-gates.php
dkoo: Addressed in 0b1704a14 — added `minimum: 1`, `maximum: 50`, `sanitize_callback: 'absint'`, and `validate_callback: 'rest_validate_request_arg'` (the last one is required for WP REST to actually enforce the bounds; without it `minimum`/`maximum` are silently ignored). Runtime clamp removed. New test `test_posts_search_endpoint_per_page_below_minimum` confirms `per_page=0` returns 400.

includes/wizards/audience/class-audience-content-gates.php
dkoo: Addressed in 0b1704a14 — `posts_per_page` is now `min(count($ids), 100)`. New test `test_posts_search_endpoint_caps_include_results` passes 105 IDs and asserts the response is capped at 100.

includes/wizards/audience/class-audience-content-gates.php
dkoo: Addressed in 6231a08ca — went with the broadened-status approach. When `include` is set, `post_status` is now `['publish', 'draft', 'pending', 'private', 'future']` so the editor keeps showing tokens whose status changed after the gate was saved. Search-mode stays publish-only so users can only ADD published items. New test `test_posts_search_endpoint_include_hydrates_non_published` covers both directions. Trashed posts still won't hydrate (intentional — a trash is user-actioned removal).

includes/content-gate/class-content-restriction-control.php
dkoo: Addressed in 0c99eec35 — `gateMatchesPost()` now mirrors the PHP `get_post_gates()` semantics: an inclusion-override pre-pass (current post ID in any `specific_posts` rule short-circuits to `true`), `specific_posts` is skipped inside the AND `every()`, and a `hasNonSpecificRule` guard prevents a gate-with-only-`specific_posts` from being included when the override doesn't match. The current post ID is now threaded through `useSelect` via `getCurrentPostId()`. Bundle rebuilt, lint clean.

tests/unit-tests/content-gate/content-gates.php
dkoo: Addressed in ab390975c — message now reads `'endpoint matches the registered REST route'`. Good catch.

tests/unit-tests/content-gate/content-gates.php
dkoo: Addressed in ab390975c — dropped the `end($keys)` ordering assertion. Agree with the framing: the "appears last" property is a UX guarantee, not a data-layer one. The Linear spec still requires `specific_posts` last in the registration array; a future engineer would need to preserve that, but the unit test isn't the right enforcer.

tests/class-newspack-unit-tests-bootstrap.php
dkoo: Addressed in d768ae68d — reverted the bootstrap define and moved it into `Test_Content_Gates::setUpBeforeClass()`, plus reset `$GLOBALS['wp_rest_server']` and re-fire `rest_api_init` so the audience-content-gates routes register with the flag on. The reset is needed because in full-suite runs another test class can fire `rest_api_init` first, caching the server without our routes. Verified the full suite (942 tests) and filtered `Test_Content_Gates` (40 tests) both pass under this scoping. Confirmed grep: no existing test depends on the feature being off.

tests/unit-tests/content-gate/content-gates.php
dkoo: Addressed in ab390975c — dropped the `__()` wrapper around the substring; it's now a plain `'restrict specific posts'`. (Note: that wrapping was introduced post-push by an accepted Copilot suggestion in 175cfd951; I should have caught it then.)

adekbadek (approved):
Re-reviewed. All 8 inline threads addressed across 5 commits:

- Editor matcher (`gateMatchesPost`) correctly mirrors PHP `get_post_gates()` semantics — override pre-pass, AND-skip for `specific_posts`, `hasNonSpecificRule` guard, `postId` threaded through deps.
- `per_page` schema gets full validation incl. `validate_callback`; `include` path capped at `min(count(ids), 100)`.
- `post_status` broadens when `include` is set so stale tokens stay visible.
- Bootstrap leak isolated into `setUpBeforeClass` with `wp_rest_server` reset + `rest_api_init` re-fire.
- Test nits (assertion message, `__()` wrapper, `end($keys)` ordering) all taken.

CI green. Approving.

PR #4675, opened 2026-04-17
fix(access-control): ignore content rules with empty value

dkoo:
### All Submissions:

A small improvement to avoid a potentially confusing gate configuration. If a gate has a content rule enabled but with an empty value, it should be ignored to allow the gate's other content rules to be evaluated.

1. Define `define( 'NEWSPACK_CONTENT_GATES', true );` in wp-config.php and deactivate WooCommerce Memberships.
2. In **Audience > Access control** add or edit a content gate and select "Choose specific content". 
3. For the content rules, enable the **Post types** rule but deselect all options. Enable **Categories** and select one or more terms. Set any access rules and save.
4. Visit a post that matches one of the categories selected in the content rules. On `trunk`, observe that the post is NOT gated—this is because the post is also evaluated against the empty Post Types rule, which will never match any posts.
5. On this branch, confirm that the post IS gated because it matches a term in the **Categories** rule.
6. Confirm that other posts that don't have one of the restricted categories are not gated.
7. Confirm the new unit tests pass 👇 

### Other information:


tests/unit-tests/content-gate/content-gates.php
```diff
@@ -233,6 +233,24 @@ public function test_content_rules() {
...
+		// Update content rules to add an empty post_type value.
+		Content_Rules::update_gate_content_rules(
+			$this->gate_ids[2],
+			[
+				[
+					'slug'  => 'post_types',
+					'value' => [],
+				],
+				[
+					'slug'  => 'category',
```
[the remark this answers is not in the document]
dkoo: Noticed this and fixed in e6ed906

includes/content-gate/class-content-restriction-control.php
```diff
@@ -155,6 +155,10 @@ public static function get_post_gates( $post_id = null ) {
 			}
 
 			foreach ( $content_rules as $content_rule ) {
+				// Rule with an empty value should pass through.
+				if ( empty( $content_rule['value'] ) ) {
+					continue;
+				}
```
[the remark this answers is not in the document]
dkoo: 12d3a4bc4 refactors this to filter out rules with empty values before iterating through the rules, and 5cb5e4e0 moves the filtering earlier so the `empty( $content_rules )` check works on the filtered rules.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -155,6 +155,10 @@ public static function get_post_gates( $post_id = null ) {
 			}
 
 			foreach ( $content_rules as $content_rule ) {
+				// Rule with an empty value should pass through.
+				if ( empty( $content_rule['value'] ) ) {
+					continue;
```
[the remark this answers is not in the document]
dkoo: 12d3a4b updates the JS matcher to filter out the empty content rules.

src/content-gate/editor/post-settings.js
```diff
@@ -20,7 +20,8 @@ const { gates = [], taxonomyMap = {}, canEditGates = false } = window.newspackCo
  * @return {boolean} Whether all rules match.
  */
 function gateMatchesPost( contentRules, postType, termsByTax ) {
-	return contentRules.every( rule => {
+	const filteredRules = contentRules.filter( rule => rule.value?.length > 0 );
```
adekbadek: When all rules are filtered out, `filteredRules` is `[]` and `[].every()` returns `true`, so the gate matches every post. The PHP side now skips gates whose rules are all empty, so this creates the inverse of the bug being fixed — verified in the editor on a post the runtime does not gate.

Two reasonable options:

- Add `if ( filteredRules.length === 0 ) return false;` before the `.every()` call.
- Or do the filtering PHP-side in `Content_Gate::enqueue_post_settings_script` so empty rules never reach the editor in the first place — that also keeps the localized payload smaller and co-locates the filter with the existing `empty( $gate['content_rules'] )` skip.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -150,6 +150,16 @@ public static function get_post_gates( $post_id = null ) {
 			}
 
 			$content_rules = $gate['content_rules'];
+
+			// Filter out rules with empty values.
+			$content_rules = is_array( $content_rules ) ? array_values(
+				array_filter(
+					$content_rules,
+					function( $content_rule ) {
+						return ! empty( $content_rule['value'] );
```
adekbadek: The two empty-checks aren't quite the same shape: PHP uses `! empty( $content_rule['value'] )` (null, 0, '0', '', [] all count as empty), JS uses `rule.value?.length > 0` (only arrays/strings). All current rule values are arrays so they agree today, but if a future rule type stores a scalar value the two sides could diverge. Worth aligning on something explicit, e.g. `Array.isArray( rule.value ) && rule.value.length > 0` in JS mirrored by `is_array( ... ) && count( ... ) > 0` in PHP.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -233,6 +233,24 @@ public function test_content_rules() {
...
+					'value' => [ $cat1 ],
+				],
+			]
+		);
+		$gates = Content_Restriction_Control::get_post_gates( $post1 );
+		$this->assertCount( 1, $gates, 'One gate for the post in category 1' );
+		$this->assertEquals( $this->gate_ids[2], $gates[0]['id'], 'Rule with an empty value should pass through' );
```
adekbadek: Tiny wording thing: "Rule with an empty value should pass through" reads as if the empty rule passes through to evaluation, but it's actually filtered out (and the *other* rule is what makes the gate match). Something like "Empty rule is ignored, remaining category rule still matches" would describe what the assertion proves more directly.

adekbadek (changes requested):
The PHP fix works as described — verified the gating behavior:

- Created a gate with `content_rules = [{post_types: []}, {category: [9]}]`.
- A post in the gated category is correctly gated.
- A post outside the gated category is correctly not gated.

There is one issue introduced on the editor side: when a gate's `content_rules` are *all* filtered out as empty, `gateMatchesPost` returns `true` (because `[].every()` is `true`), and the post-settings panel reports the gate as applying to every post. The PHP path correctly skips gates with no remaining rules, so the editor and the runtime now disagree for exactly the empty-rule case this patch handles. Reproduced by setting one gate's rules to `[{slug: "post_types", value: []}]` and opening any post in the editor — the panel falsely shows "Gates that apply to this post: …" while the public page is not gated.

Worth fixing before merge so the editor doesn't mislead authors. A couple of smaller things in the inlines.

src/content-gate/editor/post-settings.js
dkoo: Since #4674, `gateMatchesPost` verifies the existence of any non-specific-post rules before considering the gate a match, which essentially solves this.

Additionally, 1f98a88 moves the filtering all the way back to the initial getter method `Content_Rules::get_gate_content_rules`, so that any content rules with empty values are filtered out before they're evaluated by any PHP or JS functions. This simplifies things by sidestepping the need to make any filtering match across methods.

includes/content-gate/class-content-restriction-control.php
dkoo: Addressed in 1f98a88. We now only filter out rules with an empty value if the value is an array and the count of that array is `0`.

tests/unit-tests/content-gate/content-gates.php
dkoo: Updated in 1f98a88: `Rule with an empty array-like value is ignored; category rule still matches`

dkoo:
@adekbadek thanks for catching the issues here. After merging in the latest changes from #4674, this became a bit easier to solve due to additional checks added by that branch that ensure that a gate with no content rules is never considered a match. 1f98a88 refactors the fix for this PR by moving the filtering of content rules to the earliest possible moment, avoiding the need to replicate this logic across other methods and across PHP vs. JS (and avoiding the need for any changes to JS methods). I also resolved a sort of merge conflict where a `per_page` maximum for the `posts-search` endpoint set by #4674 was conflicting with the new default `per_page` value of `100` set by #4669.

All tests are passing and I'm not seeing any matching posts indicating unexpected gate matches in the editor sidebar for gates with empty content rules, now.

adekbadek (approved):
Refactoring the filter to the source method (`Content_Rules::get_gate_content_rules`) is a cleaner fix than what was originally suggested — editor and runtime now consume identical filtered rules, so the false-match in the post-settings panel is structurally impossible. Test wording updated to match. CI green.

PR #4676, opened 2026-04-17
feat(reader-activation): configure ESP incoming fields from schema

leogermani:
### All Submissions:

Teach the Reader Activation ESP integration to consume the enriched schema now exposed by the newsletter service provider, so each contact field surfaces in the admin UI with its correct name, value type, options (for dropdown / radio / listbox / checkbox), and default access-rule / segmentation-criterion eligibility.

- Override `configure_incoming_field()` in the ESP integration to populate each `Incoming_Field` from the schema stored in its raw data — `name`, `value_type`, `matching_function`, `options`, `description`, `is_access_rule`, `is_segment_criteria` are all applied when present.
- Apply the same configuration in `get_available_incoming_fields()` so the admin picker reflects the defaults instead of raw keys.

### Related PR

Requires [Automattic/newspack-newsletters#2082](https://github.com/Automattic/newspack-newsletters/pull/2082), which renames the provider method and extends the returned schema this PR consumes. This PR will fail against `trunk` of newspack-newsletters until that PR is merged, because the old `get_contact_fields()` method has been removed in favor of `get_contact_fields_for_integrations()`.

1. Start the workspace and check out this branch plus the paired newspack-newsletters branch (`feat/esp-contact-fields-integrations-schema`).
2. Configure **Mailchimp** as the ESP with a list that has a mix of merge-field types including a dropdown.
3. Navigate to the Audience / Reader Activation ESP section where promoted contact fields are picked and verify:
   - Each field displays the human-readable name from the ESP, not just the raw key.
   - Dropdown / radio fields display the correct option list.
   - Fields of type `text / number / date / radio / dropdown` are flagged eligible for access-rule and segmentation promotion; other types are not.
4. Enable a Mailchimp `dropdown`-type field as a promoted field. Go to **Audience → Content Gates**, create a gate using that field as an access rule, and verify the option list is present in the picker.
5. Create a popups segment using the same field and confirm the options surface in the segmentation criteria UI.
6. Repeat steps 3–5 with ActiveCampaign configured as the ESP (include a `dropdown` field).
7. Regression check: any promoted fields previously enabled before this PR should continue to evaluate correctly for existing gates and segments.

### Other information:


includes/reader-activation/integrations/class-esp.php
```diff
@@ -425,9 +425,49 @@ public function get_available_incoming_fields() {
...
+		}
+		if ( isset( $raw['description'] ) ) {
+			$field->set_description( $raw['description'] );
+		}
+		if ( ! empty( $raw['is_access_rule'] ) ) {
+			$field->set_is_access_rule( true );
+		}
+		if ( ! empty( $raw['is_segment_criteria'] ) ) {
+			$field->set_is_segment_criteria( true );
+		}
```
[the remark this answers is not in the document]
leogermani: Addressed in b787db947. configure_incoming_field() now guards each setter: scalar fields are cast to string only when is_scalar(), options is set only when it's an array, and the promotion flags are parsed with wp_validate_boolean() so stringy "false" doesn't become truthy.

includes/reader-activation/integrations/class-esp.php
```diff
@@ -425,9 +425,49 @@ public function get_available_incoming_fields() {
...
+		return $field;
```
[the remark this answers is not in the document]
leogermani: Addressed in b787db947. Added tests/unit-tests/integrations/class-test-esp.php with five cases: full schema applied, empty schema preserves defaults, invalid types are ignored, wp_validate_boolean parsing, and get_available_incoming_fields() pipes each field through configure_incoming_field().

includes/reader-activation/integrations/class-esp.php
```diff
@@ -423,11 +423,55 @@ public function get_available_incoming_fields() {
...
+			$field->set_matching_function( (string) $raw['matching_function'] );
+		}
+		if ( ! empty( $raw['options'] ) && is_array( $raw['options'] ) ) {
+			$field->set_options( $raw['options'] );
+		}
+		if ( isset( $raw['description'] ) && is_scalar( $raw['description'] ) ) {
+			$field->set_description( (string) $raw['description'] );
+		}
+		if ( isset( $raw['is_access_rule'] ) && \wp_validate_boolean( $raw['is_access_rule'] ) ) {
+			$field->set_is_access_rule( true );
```
adekbadek: Asymmetric with the other property assignments in this method: this branch can only set the flag to `true`, never back to `false`. I verified the consequence: pre-setting the flag to `true` and then calling `configure_incoming_field` with `is_access_rule => false` leaves the flag `true`. Suggest:

```php
if ( isset( $raw['is_access_rule'] ) ) {
    $field->set_is_access_rule( wp_validate_boolean( $raw['is_access_rule'] ) );
}
```

…and a paired test that pre-sets the flag to `true` then passes `false`. Same change for `is_segment_criteria` below.

includes/reader-activation/integrations/class-esp.php
```diff
@@ -423,11 +423,55 @@ public function get_available_incoming_fields() {
...
+			$field->set_value_type( (string) $raw['value_type'] );
+		}
+		if ( ! empty( $raw['matching_function'] ) && is_scalar( $raw['matching_function'] ) ) {
+			$field->set_matching_function( (string) $raw['matching_function'] );
+		}
+		if ( ! empty( $raw['options'] ) && is_array( $raw['options'] ) ) {
+			$field->set_options( $raw['options'] );
+		}
+		if ( isset( $raw['description'] ) && is_scalar( $raw['description'] ) ) {
+			$field->set_description( (string) $raw['description'] );
```
adekbadek, line 470: `isset() && is_scalar()` here is more permissive than the `! empty()` used for `name` / `value_type` / `matching_function` above — an explicit `description => ''` will overwrite the default, but an explicit `name => ''` is silently ignored (preserves `key` fallback). Pick one pattern, even if the current effect is harmless.

tests/unit-tests/integrations/class-test-esp.php
```diff
@@ -0,0 +1,267 @@
...
+
+		$this->assertSame( 'Organization', $result[0]->get_name() );
+		$this->assertTrue( $result[0]->is_access_rule() );
+		$this->assertTrue( $result[0]->is_segment_criteria() );
+
+		$this->assertSame( 'VIP', $result[1]->get_name() );
+		$this->assertSame( 'boolean', $result[1]->get_value_type() );
+		$this->assertFalse( $result[1]->is_access_rule() );
+
+		\Newspack_Newsletters_Contacts::reset_calls();
```
adekbadek: `reset_calls()` is called manually at the end of each test method (here, line 223, line 265). If any earlier assertion throws, the fixture leaks into the next test. Move to `tear_down()` so cleanup is unconditional. Same for the `$fields_fixture` reset.

tests/unit-tests/integrations/class-test-esp.php
```diff
@@ -0,0 +1,267 @@
...
+				'is_segment_criteria' => true,
+			],
+			[
+				'key'        => 'is_vip',
+				'name'       => 'VIP',
+				'value_type' => 'boolean',
+			],
+		];
+
+		$esp = new class() extends ESP {
```
adekbadek: All three test methods declare the same `extends ESP { public function get_master_list_id() { return 'test-list'; } }` shim (line 139, 187, 249). Extract a `make_esp_with_master_list()` helper, or set the option in `set_up()` so the real method returns a value.

tests/unit-tests/integrations/class-test-esp.php
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Tests for the ESP integration's configure_incoming_field() behavior.
+ *
+ * @package Newspack\Tests\Unit\Integrations
+ */
+
+namespace Newspack\Tests\Unit\Integrations;
+
+use Newspack\Reader_Activation\Integrations\ESP;
+use Newspack\Reader_Activation\Integrations\Incoming_Field;
```
adekbadek, line 11: The mock class `Newspack_Newsletters_Contacts` is loaded by `tests/unit-tests/reader-activation-sync.php` at file scope — works in the full suite but fragile if anyone ever runs this file directly. Consider an explicit `require_once __DIR__ . '/../../mocks/newsletters-mocks.php';` near the top, like the existing pattern in `content-gate/class-premium-newsletters.php`.

src/wizards/audience/views/integrations/settings-field.js
```diff
@@ -34,18 +34,21 @@ export const SettingsField = ( { field, value, onChange } ) => {
...
 					<Grid columns={ 3 } rowGap={ 16 }>
-						{ options.map( fieldName => (
+						{ normalizedOptions.map( ( { value: optionValue, label: optionLabel } ) => (
 							<CheckboxControl
 								className="newspack-checkbox-control"
-								key={ fieldName }
-								label={ fieldName.replace( ': ', '' ) }
-								checked={ selectedFields.includes( fieldName ) }
+								key={ optionValue }
+								label={ optionLabel.replace( ': ', '' ) }
```
adekbadek: Now that this branch renders ESP-derived labels (real names, no `: ` prefix), this strip is a no-op for the new shape and breaks legitimate labels — e.g. `'Plan: Gold'` becomes `'PlanGold'`. Worth either dropping it, pushing label normalization to the PHP side, or commenting on which legacy case it preserves.

adekbadek (changes requested):
Reviewed alongside the paired newsletters PR. Schema-key alignment between the two is clean, the new test file does real work (each method covers a regression you'd want to catch), and the JS option-shape normalization correctly bridges old (string) and new (`{ value, label }`) configs.

Two things I'd push for before merge, plus a handful of smaller items inline:

1. **`is_access_rule` and `is_segment_criteria` can be turned on by the schema but never off** — the setter only fires when the value is truthy. I confirmed this with a quick smoke test: pre-setting the flag to `true` and then running `configure_incoming_field` with `is_access_rule => false` leaves the flag at `true`. Today's call sites always start from a fresh `Incoming_Field` so this isn't a live bug, but it diverges from how the other six properties are handled and will quietly break if anything ever reuses a configured field. Suggest unconditional assignment when the key is present.
2. **Existing enabled fields silently degrade after upgrade.** Sites that had incoming fields enabled before this lands have `raw_data` stored under the old shape (no `name`, no `is_access_rule`, no `options`). After both PRs land, those entries fall back to defaults until an admin re-saves the audience integrations page. Worth either a read-side rebuild, or at least a callout in the PR description so upgraders know to re-save.

Inline comments below for the smaller items.

includes/reader-activation/integrations/class-esp.php
adekbadek: Addressed in 7debf9b3. Both setters now fire unconditionally when the schema key is present, so an explicit `false` resets a previously-true flag. New test `test_configure_can_reset_boolean_flags_to_false` covers the negative path. Picking up this work while @leogermani is unavailable.

includes/reader-activation/integrations/class-esp.php
adekbadek: Addressed in 7debf9b3. Aligned `name`/`value_type`/`matching_function` to all use `isset() && is_scalar() && '' !== (string) ...`. Kept `description` permissive on purpose so a provider can clear a previously-set description, and added a comment explaining the asymmetry.

tests/unit-tests/integrations/class-test-esp.php
adekbadek: Addressed in 7debf9b3. `reset_calls()`, filter cleanup, and the option delete are all in `tear_down()` now.

tests/unit-tests/integrations/class-test-esp.php
adekbadek: Addressed in 7debf9b3. Extracted to `make_esp_with_master_list()` and called from every test that needs a stub list id.

tests/unit-tests/integrations/class-test-esp.php
adekbadek: Addressed in 7debf9b3. Added `require_once` for the newsletters mock at the top of the file.

src/wizards/audience/views/integrations/settings-field.js
adekbadek: Addressed in 7debf9b3. Replaced the unconditional `replace(': ', '')` with `replace(/:\s*$/, '')` so legitimate labels like `'Plan: Gold'` survive while legacy trailing-colon labels (`'Signup UTM: '`, `'Payment UTM: '`) are still cleaned.

adekbadek (approved):
All inline items addressed (resolved threads above). I picked up the remaining work since @leogermani is out:

- Symmetric boolean assignment so `is_access_rule` / `is_segment_criteria` can be reset to `false` from the schema, with a paired test covering the negative path.
- Guard alignment on `name` / `value_type` / `matching_function`; intentional `description` permissiveness documented inline so a provider can clear a previously-set help text.
- `Promoted_Fields::parse_list_value()` recognizes ActiveCampaign's `||val||val||` multi-select format under `list__in` / `list__not_in`, with leading-and-trailing wrapping required so non-AC strings containing `||` mid-value are left alone.
- Sites that had incoming fields enabled before this lands are no longer silently degraded — `get_enabled_incoming_fields()` rebuilds entries on read by overlaying the live provider schema when stored `raw_data` is in the legacy shape, with a graceful fallback if the live fetch fails.
- Test housekeeping: `tear_down()` cleanup, helper extraction, mock `require_once` at file scope.

Suite green locally (1004 / 1004 with 1 pre-existing skip) and on CI.

Approving so the coordinated merge with #2082 can proceed.

PR #4685, opened 2026-04-28
feat(content-gate): expose institutional IP allowlist endpoint

rbcorrales:
### All Submissions:

Adds a `GET /newspack/v1/institutional-access/ip-allowlist` REST endpoint, gated by `manage_options`, that returns the institutional IP allowlist configured for the site. Designed for external systems to mirror the list locally and decide gating without hitting the origin per request.

The endpoint exposes only IP rules. Email domain and reader data rules are not surfaced. Entries that fail IPv4 or CIDR validation are dropped before being returned. A `newspack_content_gate_ip_allowlist` filter is provided so other plugins can augment the list.

This PR also includes a security fix for a bug in `IP_Access_Rule::ip_matches_ranges()` (introduced in #4574): the `(int) $bits` cast silently coerced any value that wasn't numeric to `0`, causing typo'd entries like `10.0.0.5/foo` or `10.0.0.5/` to match every IP. Validation is now tightened with `ctype_digit()` before the cast. Caused by admin input only, not externally exploitable, but operationally risky.

Closes NPPM-2770.

1. Configure an institution with a valid IP range (e.g. `192.168.1.0/24,10.0.0.5`) under Audience > Access Control > Institutions.
2. As an administrator, `curl -u admin:app_password https://your-site.com/wp-json/newspack/v1/institutional-access/ip-allowlist`. Confirm the response is a JSON array of `{ id, name, ip_ranges }` entries, sorted by `id` ascending.
3. Hit the endpoint as an unauthenticated user. Confirm a 401 response.
4. Hit the endpoint as a user without admin permissions (e.g. subscriber). Confirm a 403 response.
5. Configure a second institution with no IP range. Confirm it does NOT appear in the response.
6. Configure an institution with a mix of valid and malformed entries (e.g. `10.0.0.1,not-an-ip,10.0.0.5/foo,192.168.1.0/24,10.0.0.5/40`). Confirm only the valid ones (`10.0.0.1`, `192.168.1.0/24`) appear.
7. Confirm the existing `/check` endpoint (POST and GET) still works unchanged.
8. Verify the security fix: configure an institution with `10.0.0.0/foo` as its IP range. An unrelated visitor IP like `203.0.113.5` must NOT pass the institutional IP gate. Previously it would have, due to the loose validation.

### Other information:

### DIFF
--- a/includes/content-gate/class-ip-access-rule.php
+++ b/includes/content-gate/class-ip-access-rule.php
@@ -575,8 +575,10 @@ class IP_Access_Rule {
 	/**
 	 * Parse a comma-separated list of IPs and CIDR blocks.
 	 *
-	 * Trims whitespace, drops empty tokens, and discards anything that isn't a
-	 * valid IPv4 address or CIDR block (`<ipv4>/<0-32>`).
+	 * Trims whitespace (around tokens and around the `/` separator), drops
+	 * empty tokens, and discards anything that isn't a valid IPv4 address or
+	 * CIDR block (`<ipv4>/<0-32>`). Returned CIDR entries are emitted in their
+	 * trimmed form.
 	 *
 	 * @param string $raw Comma-separated list (e.g. `"192.168.1.0/24,10.0.0.5"`).
 	 *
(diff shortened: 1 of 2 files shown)


includes/content-gate/class-ip-access-rule.php
```diff
@@ -483,6 +572,40 @@ function showError( message, detail ) {
...
+		foreach ( $tokens as $token ) {
+			if ( strpos( $token, '/' ) !== false ) {
+				list( $subnet, $bits ) = explode( '/', $token, 2 );
+				if ( ! ctype_digit( $bits ) ) {
+					continue;
+				}
+				$bits = (int) $bits;
+				if ( $bits > 32 || ! filter_var( $subnet, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 ) ) {
+					continue;
+				}
```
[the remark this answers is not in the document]
rbcorrales: Good catch. Trimmed both `$subnet` and `$bits` after the `/` split, and the response now emits the trimmed, normalized form (e.g. `192.168.1.0 / 24` becomes `192.168.1.0/24`). Regression coverage added in `test_cidr_tolerates_whitespace_around_slash` and `test_ip_allowlist_normalizes_whitespace_around_cidr_slash`.

includes/content-gate/class-ip-access-rule.php
```diff
@@ -189,14 +207,85 @@ public static function check_external_ip_rest( $request ) {
...
+	 * CIDR range. Malformed entries are dropped silently. Email-domain and
+	 * reader-data rules are not exposed.
+	 *
+	 * @return \WP_REST_Response
+	 */
+	public static function get_ip_allowlist_rest() {
+		if ( function_exists( 'batcache_cancel' ) ) {
+			batcache_cancel();
+		}
+		nocache_headers();
```
miguelpeixe: This is superfluous with the existing permission callback. Logged-in users are not cached.

claudiulodro (approved):
Thanks!

includes/content-gate/class-ip-access-rule.php
rbcorrales: Removed both `batcache_cancel()` and `nocache_headers()` in b5b8dcd.

PR #4695, opened 2026-04-30
fix(access-control): allow access rules with `supports_anonymous` to evaluate anonymous readers

dkoo:
### All Submissions:

Fixes an important bug in the current implementation of IP-based access via Institutions: when a content gate has both Registered and Paid Access (with institutional/IP access) enabled, the Registered Access rule will gate content for non-logged-in users even if they've already passed the IP access check at the special URL. This happens before we even evaluate the Paid Access rules, so we never even get a chance to grant access to someone who should be whitelisted via Institution.

#4574 already implemented a `supports_anonymous` flag in the access rule schema, so the proposed fix here uses this to address the above scenario. If an anonymous user encounters a gate with Registered Access rule requirements, we evaluate any active Paid Access rules. Since only rules with `supports_anonymous => true` will potentially pass for an anonymous readers, this allows institutions' IP-based access rules to pass if the anonymous user has already successfully completed the IP access check, but not any other access rules.

Closes NPPD-1494.

1. Publish a content gate with both Registered + Paid Access enabled, with Paid Access rules requiring either an active subscription or institutional access to an institution that whitelists an IP range that includes your public IP.
2. As an anonymous reader, visit the Institution's access check URL with a privileged IP to set the `wp_nocache_ip` cookie.
3. On `trunk`, visit some restricted posts and observe that the posts are gated with the Registered Access layout despite you matching the institution's IP whitelist.
4. Check out this branch and refresh. Now confirm that after passing the IP check, restricted posts are shown even while not logged in.
5. As an admin, remove the institution from the content gate's access rules.
6. In the same reader session that was granted access via IP whitelist, visit more restricted posts and confirm that they're restricted again with the Registered Access gate layout.
7. Still in the same session, register a new account and visit more restricted posts, and confirm that they're restricted with the Paid Access gate layout.
8. Confirm all unit tests pass and there are no regressions in other gating logic unrelated to IP access rules.

### Other information:

### DIFF
--- a/includes/content-gate/class-content-restriction-control.php
+++ b/includes/content-gate/class-content-restriction-control.php
@@ -239,11 +239,11 @@ class Content_Restriction_Control {
 		}
 
 		$user_id = $user_id ?? get_current_user_id();
-		// Cache the user ID for the current request. Used only for self::$post_gate_layout_id_map,
-		// so the page-render's user's layout is the one templates retrieve via get_gate_layout_id.
-		// Other instances of $user_id below are intentional so that this method can be called
-		// repeatedly with different user IDs in the same execution context.
-		// See: Newspack_Premium_Newsletters::check_access, called for each user in the queue.
+		// Each call evaluates restriction for its own $user_id; only self::$user_id — used by
+		// get_gate_layout_id() to surface the page-render viewer's layout to templates — is
+		// cached across calls. This lets the same request evaluate multiple users (e.g.
+		// Newspack_Premium_Newsletters::check_access iterating over a queue) without the
+		// first caller's user ID hijacking subsequent evaluations.
 		if ( ! self::$user_id ) {
 			self::$user_id = $user_id;
 		}
--- a/includes/content-gate/class-content-restriction-control.php
+++ b/includes/content-gate/class-content-restriction-control.php
@@ -28,13 +28,6 @@ class Content_Restriction_Control {
 	private static $post_gate_layout_id_map = [];
 
 	/**
-	 * Cached user ID to check restrictions for.
-	 *
-	 * @var int
-	 */
-	private static $user_id = 0;
-
-	/**
 	 * Post meta key for exempting a post from access control restrictions.
 	 *
 	 * @var string
@@ -239,14 +232,6 @@ class Content_Restriction_Control {
 		}
 
 		$user_id = $user_id ?? get_current_user_id();
-		// Each call evaluates restriction for its own $user_id; only self::$user_id — used by
-		// get_gate_post_id() and get_gate_layout_id() to surface the page-render viewer's gate
-		// to templates — is cached across calls. This lets the same request evaluate multiple
-		// users (e.g. Newspack_Premium_Newsletters::check_access iterating over a queue)
-		// without the first caller's user ID hijacking subsequent evaluations.
-		if ( ! self::$user_id ) {
-			self::$user_id = $user_id;
-		}
 
 		// Don't restrict this post for users who can edit it.
 		if ( ! empty( $post_id ) && user_can( $user_id, 'edit_post', $post_id ) ) {
@@ -312,7 +297,12 @@ class Content_Restriction_Control {
 	}
 
 	/**
-	 * Get the current gate post ID.
+	 * Get the current gate post ID for the current user.
+	 *
+	 * Looks up the cache entry written by the most recent is_post_restricted()
+	 * call for the current user. Calls made for a *different* user (e.g.
+	 * queue workers, REST callbacks iterating over readers) write to their
+	 * own cache slot and do not surface here.
 	 *
 	 * @param int $post_id Post ID. If not given, uses the current post ID.
 	 *
@@ -328,14 +318,20 @@ class Content_Restriction_Control {
 		if ( ! $post_id ) {
 			return false;
 		}
-		if ( ! empty( self::$post_gate_id_map[ $post_id . '_' . self::$user_id ] ) ) {
-			return self::$post_gate_id_map[ $post_id . '_' . self::$user_id ];
+		$user_id = get_current_user_id();
+		if ( ! empty( self::$post_gate_id_map[ $post_id . '_' . $user_id ] ) ) {
+			return self::$post_gate_id_map[ $post_id . '_' . $user_id ];
 		}
 		return false;
 	}
 
 	/**
-	 * Get the current gate layout ID.
+	 * Get the current gate layout ID for the current user.
+	 *
+	 * Looks up the cache entry written by the most recent is_post_restricted()
+	 * call for the current user. Calls made for a *different* user (e.g.
+	 * queue workers, REST callbacks iterating over readers) write to their
+	 * own cache slot and do not surface here.
 	 *
 	 * @param int $post_id Post ID. If not given, uses the current post ID.
 	 *
@@ -351,8 +347,9 @@ class Content_Restriction_Control {
 		if ( ! $post_id ) {
 			return false;
 		}
-		if ( ! empty( self::$post_gate_layout_id_map[ $post_id . '_' . self::$user_id ] ) ) {
-			return self::$post_gate_layout_id_map[ $post_id . '_' . self::$user_id ];
+		$user_id = get_current_user_id();
+		if ( ! empty( self::$post_gate_layout_id_map[ $post_id . '_' . $user_id ] ) ) {
+			return self::$post_gate_layout_id_map[ $post_id . '_' . $user_id ];
 		}
 		return false;
 	}
--- a/includes/content-gate/class-content-restriction-control.php
+++ b/includes/content-gate/class-content-restriction-control.php
@@ -280,7 +280,7 @@ class Content_Restriction_Control {
 					// not grant access — Access_Rules treats an empty value as "no constraint" and
 					// returns true, which would silently bypass registration here.
 					$anonymous_bypass_passed = ! empty( $gate['custom_access']['active'] )
-						&& self::custom_access_passes_for_anonymous( $gate['custom_access']['access_rules'] ?? [] );
+						&& Access_Rules::evaluate_anonymous_rules( $gate['custom_access']['access_rules'] ?? [] );
 					$is_restricted  = ! $anonymous_bypass_passed;
 					$gate_layout_id = $gate['registration']['gate_layout_id'] ?? $gate['id'];
 				} elseif ( ! empty( $gate['registration']['require_verification'] ) ) {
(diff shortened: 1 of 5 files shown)


includes/content-gate/class-content-restriction-control.php
```diff
@@ -247,7 +247,7 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null,
 		}
 
 		// Don't restrict this post for users who can edit it.
-		if ( ! empty( $post_id ) && user_can( $user_id, 'edit_post', $post_id ) ) {
+		if ( ! empty( $post_id ) && user_can( self::$user_id, 'edit_post', $post_id ) ) {
```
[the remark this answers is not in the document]
dkoo: This was reverted in 66e7cd4 with a new inline explanation as it did cause regressions in the Premium Newsletters cron job that checks access for all users in the queue.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -268,12 +268,15 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null,
 			// If registration mode is active.
 			if ( ! empty( $gate['registration']['active'] ) ) {
 				// Check if user is logged in.
-				if ( $user_id === 0 ) {
-					$is_restricted  = true;
+				if ( self::$user_id === 0 ) {
+					// Anonymous visitors can still pass via the gate's custom_access rules if they
+					// match a rule with `supports_anonymous` (currently only `institution`).
+					$is_restricted  = empty( $gate['custom_access']['active'] ) || empty( $gate['custom_access']['access_rules'] )
+						|| ! Access_Rules::evaluate_rules( $gate['custom_access']['access_rules'], 0 );
```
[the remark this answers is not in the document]
dkoo: 20ebb0e adds a new suite of unit tests to cover this new behavior.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -266,7 +271,10 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null,
 			if ( ! empty( $gate['registration']['active'] ) ) {
 				// Check if user is logged in.
 				if ( $user_id === 0 ) {
-					$is_restricted  = true;
+					// Anonymous visitors can still pass via the gate's custom_access rules if they
+					// match a rule with `supports_anonymous` (currently only `institution`).
+					$is_restricted  = empty( $gate['custom_access']['active'] ) || empty( $gate['custom_access']['access_rules'] )
+						|| ! Access_Rules::evaluate_rules( $gate['custom_access']['access_rules'], 0 );
```
adekbadek: **Blocker.** This grants anonymous access whenever `Access_Rules::evaluate_rules( ..., 0 )` returns true – including when an editor saved the institutional rule active but didn't pick any institutions. With `value => []`, `Institution::evaluate( 0, [] )` returns `true` and the registration gate is silently bypassed for all logged-out visitors.

Fix at the call site by requiring a populated anonymous-capable rule/group, or push the empty-value short-circuit down into `Access_Rules::evaluate_rules` / `evaluate_rule` so future callers can't trip the same edge.

Also a minor structural nit: this `evaluate_rules(..., 0)` runs again at line 292 (`evaluate_rules( $access_rules, $user_id )` with `$user_id === 0`) for any gate that has both modes active. Cheap (institution evaluator is transient-cached) but worth computing once and reusing.

While you're here: `Access_Rules::register_rule`'s PHPDoc (`class-access-rules.php:35-46`, not in this diff) doesn't mention the new `supports_anonymous` capability. Future rule authors won't know the contract – worth documenting in the same PR or a follow-up.

includes/content-gate/class-content-restriction-control.php
```diff
@@ -239,6 +239,11 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null,
 		}
 
 		$user_id = $user_id ?? get_current_user_id();
+		// Cache the user ID for the current request. Used only for self::$post_gate_layout_id_map,
+		// so the page-render's user's layout is the one templates retrieve via get_gate_layout_id.
+		// Other instances of $user_id below are intentional so that this method can be called
```
adekbadek: Two things on this comment block:

1. "Other instances of `$user_id` below are intentional so that this method can be called repeatedly with different user IDs in the same execution context" reads as commentary on a refactor that didn't happen. Suggest reframing positively: "Each call evaluates restriction for its own `$user_id`; only `self::$user_id` (used by `get_gate_layout_id` for the page-render viewer) is cached across calls."
2. The comment is honest about a real wart – `self::$user_id` is whoever called `is_post_restricted` first in this request, and `get_gate_layout_id` reads it back. That works for the page-render path (first call IS the viewer) but if any code path – a REST callback, an early integration call, a queue worker – calls `is_post_restricted` with a non-current-user id before the page render, the layout lookup silently returns the wrong layout. The comment documents the gotcha; consider following up by deriving the layout cache key from `get_current_user_id()` at lookup time (or threading an explicit `$user_id` arg into `get_gate_layout_id`) so the static can go away. Out of scope for this bugfix.

includes/content-gate/class-institution.php
```diff
@@ -253,7 +253,7 @@ public static function user_matches_institution( $user_id, $rules, $uncached = f
 		}
 
 		if ( ! empty( $rules['ip_range'] ) ) {
-			$is_uncached = $uncached || ! empty( $user_id ) || isset( $_COOKIE[ IP_Access_Rule::COOKIE_NAME ] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+			$is_uncached = $uncached || ! empty( $user_id ) || IP_Access_Rule::is_cookie_set();
```
adekbadek, line 262: Worth a code comment near the new bypass call site (or here): the anonymous-bypass path at `class-content-restriction-control.php:277` only succeeds when `IP_Access_Rule::is_cookie_set()` is true – so a first-time on-campus visitor landing directly on a gated post and matching IP **will not** bypass the gate; they need to hit `?institutional-access=1` first. That's the correct page-cache-safety design but it's a non-obvious coupling that publishers may misread.

Would be good to lock it down with an explicit test: anonymous + matching IP + no cookie → restricted. The existing `test_anonymous_with_matching_ip_bypasses_registration` defaults `with_cookie = true` via `set_visitor_ip`, so the no-cookie variant isn't covered.

includes/content-gate/class-ip-access-rule.php
```diff
@@ -571,5 +571,14 @@ public static function get_visitor_ip() {
 		}
 		return '';
 	}
+
+	/**
+	 * Check if the IP access cookie is set.
+	 *
+	 * @return bool Whether the cookie is set.
+	 */
+	public static function is_cookie_set() {
```
adekbadek, line 587: Nit: the doc is tautological ("Check if the IP access cookie is set" / "Whether the cookie is set"). Tighten to something like: "Whether the IP-access bypass cookie was sent on the current request." That tells the reader what the cookie *does*, which is the part that matters when deciding whether to call this. Optional one-liner explaining that this helper centralises the phpcs ignore would help future grep-archaeology.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -171,6 +184,17 @@ public function tear_down() {
...
+		// phpcs:disable WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__REMOTE_ADDR__, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+		if ( null === $this->original_remote_addr ) {
+			unset( $_SERVER['REMOTE_ADDR'] );
+		} else {
+			$_SERVER['REMOTE_ADDR'] = $this->original_remote_addr;
+		}
+		unset( $_COOKIE[ IP_Access_Rule::COOKIE_NAME ] );
+		// phpcs:enable
+		delete_transient( Institution::TRANSIENT_KEY );
+		wp_set_current_user( 0 );
```
adekbadek: `wp_set_current_user( 0 )` here is likely redundant: `WP_UnitTestCase::tear_down` already resets the global user on rollback, and `reset_visitor_state()` resets it at the end of each scenario. Not harmful, but it can surprise sibling tests that explicitly set a user and expect their assertions to run before tear_down resets it. Consider dropping it from `tear_down` and keeping only the per-scenario reset.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1385,4 +1413,479 @@ public function test_posts_search_endpoint_caps_include_results() {
...
+					],
+				],
+			]
+		);
+
+		$matching_user = $this->factory->user->create(
+			[
+				'role'       => 'subscriber',
+				'user_email' => 'a@university.edu',
+			]
```
adekbadek: Nit: trailing space after `]` before `)`. Same on line 1871. PHPCS should flag it; `npm run fix:php` would clean both up.

tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1385,4 +1413,479 @@ public function test_posts_search_endpoint_caps_include_results() {
...
+					],
+				],
+			]
+		);
+
+		wp_set_current_user( 0 );
+		$this->set_visitor_ip( '10.1.2.3' );
+		$this->reset_restriction_cache();
+
+		$this->assertFalse(
```
adekbadek, line 1546: Two coverage notes on this test:

1. The assertion is end-to-end on the registration-layout ID, which depends on `get_gate_layout_id` reading the seeded `self::$user_id`. Asserting `self::$user_id === 0` after the call (via reflection like `reset_restriction_cache()` does) would catch a regression in the seeding logic directly, rather than via downstream layout behavior – which is exactly the bug class this PR's revert addresses.
2. Adjacent gap: today's tests assert OR-of-singletons. An anonymous visitor with a matching IP, where the access rules are an AND-group of `[institution, email_domain]`, isn't directly probed. The expected outcome (restricted, because `email_domain` fails for `user_id = 0`) is worth pinning down in a sibling test.

adekbadek (commented):
Fix is well-scoped: three small production changes (`supports_anonymous` wiring, the `self::$user_id`/`$user_id` revert, and the `is_cookie_set()` extraction) plus a focused PHPUnit addition. Tests cover the matrix well (matching IP bypass, stale cookie + changed IP must not bypass, layout selection, multi-user iteration, registration-only and custom-access-only gates, etc.).

One blocker on the new anonymous-bypass branch and a handful of suggestions – everything is in inline comments.

includes/content-gate/class-content-restriction-control.php
dkoo: Addressed in ab6001aef. Tightened the registration-bypass to require at least one populated, `supports_anonymous`-capable rule via a new `Content_Restriction_Control::custom_access_passes_for_anonymous` helper — an institution rule with `value => []` is now treated as "not configured" rather than matching everyone. Also captured the result so `evaluate_rules` doesn't run twice on the same gate (your structural nit). The empty-value assumption is documented inline at `92eca4c23`.

includes/content-gate/class-content-restriction-control.php
dkoo: 1. Comment reframed in ecc623e81 and refined in be629cc5a to call out both `get_gate_post_id()` and `get_gate_layout_id()`.

2. The `self::$user_id` static refactor remains out of scope here — tracking separately as a follow-up.

includes/content-gate/class-institution.php
dkoo: Addressed in ecc623e81. Added a comment near the `$is_uncached` line in `Institution::user_matches_institution` spelling out that an anonymous on-campus visitor must complete `/institutional-access` first to set the cookie. Also added `test_anonymous_with_matching_ip_without_cookie_is_restricted` in `008c7c0fd` to lock the coupling down.

includes/content-gate/class-ip-access-rule.php
dkoo: Addressed in ecc623e81, refined in be629cc5a after a Codex review caught that the cookie can also be set via email-domain or reader-data matches in `check_ip_rest`, not only IP range. The doc now describes what the cookie does, including the phpcs-centralization note.

tests/unit-tests/content-gate/content-gates.php
dkoo: Addressed in 06867c75a. Dropped from `tear_down`; per-scenario `reset_visitor_state()` is the only reset point now. Verified the 8 sibling tests that set non-zero users via `wp_set_current_user` are self-contained (don't rely on the dropped tear_down line).

tests/unit-tests/content-gate/content-gates.php
dkoo: Addressed in 06867c75a.

tests/unit-tests/content-gate/content-gates.php
dkoo: Both addressed:

- AND-group case: `test_anonymous_with_matching_ip_and_grouped_with_email_domain_is_restricted` in 008c7c0fd.
- Direct seeding assertion: first added to the existing test in 008c7c0fd, then replaced in dd0800c41 with a stronger standalone test (`test_is_post_restricted_seeds_user_id_from_first_caller_only`) after code review caught that the original reflection assertion couldn't fail under any meaningful regression — `reset_restriction_cache` zeros the static before the call, so an anonymous evaluation also writing 0 is a tautology. The new test seeds via a logged-in caller and verifies subsequent callers (different user, anonymous) don't overwrite the seed — pinning the "first caller wins" contract.

dkoo:
@adekbadek great catch on that glaring regression allowing anonymous readers to bypass restrictions via an empty Institution rule. I've addressed this in ab6001a (with a slight refactor in a0d3a1a63 for clarity and better naming of the new `Access_Rules::evaluate_anonymous_rules` method) and addressed your other feedback in subsequent commits.

includes/content-gate/class-content-restriction-control.php
dkoo: Opened #4701 to address this.

adekbadek (approved):
LGTM

PR #4697, opened 2026-05-02
feat(access-control): add group subscription identifier to metadata

dkoo:
### All Submissions:

Surfaces the user's content-access group identity (group subscription names and matching institution names) to two destinations, as outlined in [NPPD-1451](https://linear.app/a8c/issue/NPPD-1451/add-group-subscription-identifier-to-metadata-schema):

1. **ESPs / integrations**, via a new `NP_Content_Access_Group` contact metadata field that joins the existing `NP_Content_Access` and `NP_Content_Access_Source` fields synced by `Reader_Activation\Sync\Contact_Metadata\Content_Gate`.
2. **Google Analytics 4**, via a new `group` custom event parameter on the gtag config payload. Group and Institution names are anonymized in this parameter to avoid potentially leaking PII due to the fallback names for unnamed groups including the subscription owner's full name (e.g., `gtag("config", ..., {"group":"Group 123, Institution 456", ...})`). Defaults to `none` when there are no matching groups. Only added when `Content_Gate::is_newspack_feature_enabled()` is true.

The new value reflects active group subscriptions the user owns or is a member of, plus institutions whose rules the user matches via any means (verified email domain, IP range with cache-bypass cookie or current visitor, reader_data). Names are deduplicated and sorted naturally.

This PR also includes a few smaller changes that came up while wiring up the new field:
- Adds a `$strict` parameter to `Access_Rules::has_active_subscription()` so callers can distinguish "user owns the subscription" from "user is a member of someone's group subscription" — used to choose between product names and `'group'` for `NP_Content_Access_Source`.
- Tightens `Institution::user_matches_institution()`'s "uncached" detection so the IP check only fires for the current visitor (not arbitrary `$user_id` values passed from background metadata syncs).
- Refactors `Content_Gate` metadata source/group label collection into a shared `collect_labels` helper.
- Migrates `Teams_For_Memberships::handle_esp_sync_contact()` off the deprecated `Sync_Metadata::filter_enabled_fields()` to `Integration::filter_enabled_outgoing_fields()`.

Closes NPPD-1451.

#### ESP metadata (`NP_Content_Access_Group`)

1. With an ESP connected (e.g., Mailchimp), enable the new `Content Access Group` field in **Reader Activation → ESP → outgoing fields**.
2. Create a published content gate with custom access enabled, requiring an active subscription for product P.
3. Create a group subscription for product P with a name (e.g., "ACME Corp"). Add a reader as a member.
4. Trigger a metadata sync for that reader (e.g., re-save their account or log them in/out). Confirm in your ESP that the contact has:
   - `NP_Content_Access` = `Yes`
   - `NP_Content_Access_Source` = `group`
   - `NP_Content_Access_Group` = `ACME Corp`
5. Repeat with the reader as the **owner** of the group subscription instead of a member. Confirm `NP_Content_Access_Source` = `<product name>` and `NP_Content_Access_Group` still includes `ACME Corp`.
6. Add a published Institution whose `email_domain` matches the reader's verified email. Add the institution to the gate's access rules. Re-sync and confirm both the group subscription and institution names appear (sorted, comma-separated) in `NP_Content_Access_Group`.

#### GA4 `group` parameter

1. With Google Site Kit installed and the Newspack content-gating feature enabled (`define('NEWSPACK_CONTENT_GATES', true)`), view any page logged in as the reader from above.
2. In the page source, locate the `gtag("config", ...)` call. Confirm the config object includes `"group":"<group/institution names>"`.
3. Log out and refresh. Confirm `"group":"none"`.
4. Add the reader to additional group subscriptions and matching institutions. Confirm all names appear comma-separated and sorted naturally.
5. Cancel one of the group subscriptions and refresh the page. Confirm its name is excluded.

#### Regression checks

- Confirm existing `Content Access` metadata flows still work (`NP_Content_Access`, `NP_Content_Access_Source`).
- Confirm Institution IP-based access still works for both logged-in users and anonymous users with the `wp_nocache_ip` cache-bypass cookie set.
- Confirm all unit tests pass: `n test-php`.

### Other information:

### DIFF
--- a/includes/content-gate/class-access-rules.php
+++ b/includes/content-gate/class-access-rules.php
@@ -296,9 +296,10 @@ class Access_Rules {
 	 *
 	 * @param int   $user_id     User ID.
 	 * @param array $product_ids Required product IDs.
+	 * @param bool  $strict      If true, only consider active subscriptions owned by $user_id (ignore group subscription memberships).
 	 * @return bool
 	 */
-	public static function has_active_subscription( $user_id, $product_ids ) {
+	public static function has_active_subscription( $user_id, $product_ids, $strict = false ) {
 		$has_subscription = false;
 
 		// Check user's own subscriptions.
@@ -307,7 +308,7 @@ class Access_Rules {
 		}
 
 		// Check group subscriptions the user is a member of.
-		if ( ! $has_subscription && function_exists( 'wcs_get_subscription' ) ) {
+		if ( ! $strict && ! $has_subscription && function_exists( 'wcs_get_subscription' ) ) {
 			$group_subscriptions = Group_Subscription::get_group_subscriptions_for_user( $user_id );
 			foreach ( $group_subscriptions as $subscription ) {
 				if ( ! $subscription || ! $subscription->has_status( WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES ) ) {
@@ -334,8 +335,9 @@ class Access_Rules {
 		 * @param bool  $has_subscription Whether the user has an active subscription.
 		 * @param int   $user_id          User ID.
 		 * @param array $product_ids      Required product IDs.
+		 * @param bool  $strict           If true, only consider active subscriptions owned by $user_id (ignore group subscription memberships).
 		 */
-		return apply_filters( 'newspack_access_rules_has_active_subscription', $has_subscription, $user_id, $product_ids );
+		return apply_filters( 'newspack_access_rules_has_active_subscription', $has_subscription, $user_id, $product_ids, $strict );
 	}
 
 	/**
--- a/includes/content-gate/class-access-rules.php
+++ b/includes/content-gate/class-access-rules.php
@@ -355,6 +355,16 @@ class Access_Rules {
 	 * Whether the user has an active subscription for one of the given products.
 	 * Also checks if the user is a member of a group subscription with the required products.
 	 *
+	 * Note: `$strict` only constrains the built-in ownership / group-membership checks.
+	 * The `newspack_access_rules_has_active_subscription` filter is always applied and
+	 * its return value is the final result, so a third-party filter callback can grant
+	 * access even when `$strict` is true. Filter authors should opt in to the 4th `$strict`
+	 * arg (`accepted_args` >= 4) and respect it — e.g., short-circuit and return
+	 * `$has_subscription` unchanged when `$strict` is true and the access claim isn't
+	 * strictly an owned subscription. Otherwise callers using `$strict` to distinguish
+	 * owner-vs-member access (e.g., `Content_Gate` source labels) may misclassify
+	 * filter-granted access as local ownership.
+	 *
 	 * @param int   $user_id     User ID.
 	 * @param array $product_ids Required product IDs.
 	 * @param bool  $strict      If true, only consider active subscriptions owned by $user_id (ignore group subscription memberships).
--- a/includes/content-gate/class-institution.php
+++ b/includes/content-gate/class-institution.php
@@ -236,9 +236,28 @@ class Institution {
 	}
 
 	/**
+	 * Per-request cache of [inst_id => decoded_name] maps.
+	 *
+	 * @var array<string,array<int,string>>
+	 */
+	private static $names_cache = [];
+
+	/**
+	 * Reset the per-request matching-names cache.
+	 *
+	 * Tests, CLI workers, and invalidation hooks call this to bust the memoization in
+	 * `get_matching_names_for_user()` / `get_matching_ids_for_user()`. Distinct from
+	 * `invalidate_cache()`, which clears the underlying institutions transient.
+	 */
+	public static function reset_matching_cache() {
+		self::$names_cache = [];
+	}
+
+	/**
 	 * Get the sorted, deduplicated names of institutions whose rules a user matches.
 	 *
-	 * Result is memoized per request, keyed by user ID and the optional institution filter.
+	 * Memoized per request via {@see self::get_matching_map_for_user()} — see that helper
+	 * for cache scope, the IP/cookie context dependency, and invalidation.
 	 *
 	 * @param int        $user_id            User ID.
 	 * @param array|null $institution_filter Optional list of institution post IDs. If non-empty, only
@@ -248,8 +267,50 @@ class Institution {
 	 * @return string[] Sorted, deduplicated institution names.
 	 */
 	public static function get_matching_names_for_user( $user_id, $institution_filter = null ) {
-		static $cache = [];
+		$map   = self::get_matching_map_for_user( $user_id, $institution_filter );
+		$names = array_values( array_unique( array_values( $map ) ) );
+		sort( $names, SORT_NATURAL | SORT_FLAG_CASE );
+		return $names;
+	}
 
+	/**
+	 * Get the IDs of institutions whose rules a user matches.
+	 *
+	 * Returns institution post IDs. Shares the per-request cache with
+	 * {@see self::get_matching_names_for_user()}. Suitable for callers that want a stable,
+	 * non-PII identifier (e.g., GA4 anonymized labels) and don't need the display name.
+	 *
+	 * @param int        $user_id            User ID.
+	 * @param array|null $institution_filter Optional list of institution post IDs (same semantics
+	 *                                       as {@see self::get_matching_names_for_user()}).
+	 *
+	 * @return int[] Sorted institution post IDs.
+	 */
+	public static function get_matching_ids_for_user( $user_id, $institution_filter = null ) {
+		$ids = array_keys( self::get_matching_map_for_user( $user_id, $institution_filter ) );
+		sort( $ids, SORT_NUMERIC );
+		return $ids;
+	}
+
+	/**
+	 * Build the [inst_id => decoded_name] map for the user, memoized per request.
+	 *
+	 * Cache key includes `get_current_user_id()` because `user_matches_institution()`'s IP
+	 * branch is context-dependent on the current visitor (see the comment in that method).
+	 * Without that the same `$user_id` could legitimately resolve to different sets across
+	 * requests in a long-running worker that swaps the current user.
+	 *
+	 * IDs are read with `get_post_field( 'post_title', ... )` rather than `get_the_title( ... )`
+	 * so the value going to GA4/ESP is a raw post title, not one that's been through
+	 * `the_title` filters (texturization, third-party plugins) that can introduce entities
+	 * or markup.
+	 *
+	 * @param int        $user_id            User ID.
+	 * @param array|null $institution_filter Optional list of institution post IDs.
+	 *
+	 * @return array<int,string> Map of institution post ID to decoded post title.
+	 */
+	private static function get_matching_map_for_user( $user_id, $institution_filter = null ) {
 		$user_id = (int) $user_id;
 
 		// Normalize the filter so [], null, and unsorted/duplicate inputs share a cache key.
@@ -259,26 +320,26 @@ class Institution {
 		if ( null !== $normalized_filter ) {
 			sort( $normalized_filter, SORT_NUMERIC );
 		}
-		$cache_key = $user_id . '|' . ( null === $normalized_filter ? '' : implode( ',', $normalized_filter ) );
-		if ( isset( $cache[ $cache_key ] ) ) {
-			return $cache[ $cache_key ];
+		// Include the current user in the cache key because the IP-rule branch of
+		// user_matches_institution() short-circuits when $user_id !== get_current_user_id().
+		$cache_key = $user_id . '|' . get_current_user_id() . '|' . ( null === $normalized_filter ? '' : implode( ',', $normalized_filter ) );
+		if ( isset( self::$names_cache[ $cache_key ] ) ) {
+			return self::$names_cache[ $cache_key ];
 		}
 
-		$names = [];
+		$map = [];
 		foreach ( self::get_cached_institutions() as $inst_id => $rules ) {
-			if ( null !== $normalized_filter && ! in_array( (int) $inst_id, $normalized_filter, true ) ) {
+			$inst_id = (int) $inst_id;
+			if ( null !== $normalized_filter && ! in_array( $inst_id, $normalized_filter, true ) ) {
 				continue;
 			}
 			if ( self::user_matches_institution( $user_id, $rules ) ) {
-				$names[] = html_entity_decode( (string) get_the_title( $inst_id ), ENT_QUOTES | ENT_HTML5, 'UTF-8' );
+				$map[ $inst_id ] = html_entity_decode( (string) \get_post_field( 'post_title', $inst_id ), ENT_QUOTES | ENT_HTML5, 'UTF-8' );
 			}
 		}
 
-		$names = array_values( array_unique( $names ) );
-		sort( $names, SORT_NATURAL | SORT_FLAG_CASE );
-
-		$cache[ $cache_key ] = $names;
-		return $names;
+		self::$names_cache[ $cache_key ] = $map;
+		return $map;
 	}
 
 	/**
@@ -308,6 +369,13 @@ class Institution {
 			// none of these and will see the gate — they must first complete the IP check at
 			// /institutional-access (or ?institutional-access=1) to set the cookie before
 			// subsequent gated requests can evaluate their IP.
+			//
+			// Caveat: the IP_Access_Rule::is_cookie_set() disjunct accepts any non-current
+			// $user_id when the *current visitor* carries the cache-bypass cookie, so a
+			// cookie-bearing on-campus visitor can still cause the current request's IP to
+			// be matched against another user's institution range. This is a narrow,
+			// pre-existing edge case (the cookie is only set after the visitor has already
+			// passed an institutional-access check on this site).
 			$is_uncached = $uncached || ( ! empty( $user_id ) && (int) $user_id === get_current_user_id() ) || IP_Access_Rule::is_cookie_set();
 			if ( $is_uncached && IP_Access_Rule::ip_matches_ranges( IP_Access_Rule::get_visitor_ip(), $rules['ip_range'] ) ) {
 				return true;
--- a/includes/content-gate/class-institution.php
+++ b/includes/content-gate/class-institution.php
@@ -236,6 +236,52 @@ class Institution {
 	}
 
 	/**
+	 * Get the sorted, deduplicated names of institutions whose rules a user matches.
+	 *
+	 * Result is memoized per request, keyed by user ID and the optional institution filter.
+	 *
+	 * @param int        $user_id            User ID.
+	 * @param array|null $institution_filter Optional list of institution post IDs. If non-empty, only
+	 *                                       institutions whose ID is in the list are considered.
+	 *                                       Pass null or an empty array to scan every cached institution.
+	 *
+	 * @return string[] Sorted, deduplicated institution names.
+	 */
+	public static function get_matching_names_for_user( $user_id, $institution_filter = null ) {
+		static $cache = [];
+
+		$user_id = (int) $user_id;
+
+		// Normalize the filter so [], null, and unsorted/duplicate inputs share a cache key.
+		$normalized_filter = is_array( $institution_filter ) && ! empty( $institution_filter )
+			? array_values( array_unique( array_map( 'absint', $institution_filter ) ) )
+			: null;
+		if ( null !== $normalized_filter ) {
+			sort( $normalized_filter, SORT_NUMERIC );
+		}
+		$cache_key = $user_id . '|' . ( null === $normalized_filter ? '' : implode( ',', $normalized_filter ) );
+		if ( isset( $cache[ $cache_key ] ) ) {
+			return $cache[ $cache_key ];
+		}
+
+		$names = [];
+		foreach ( self::get_cached_institutions() as $inst_id => $rules ) {
+			if ( null !== $normalized_filter && ! in_array( (int) $inst_id, $normalized_filter, true ) ) {
+				continue;
+			}
+			if ( self::user_matches_institution( $user_id, $rules ) ) {
+				$names[] = html_entity_decode( (string) get_the_title( $inst_id ), ENT_QUOTES | ENT_HTML5, 'UTF-8' );
+			}
+		}
+
+		$names = array_values( array_unique( $names ) );
+		sort( $names, SORT_NATURAL | SORT_FLAG_CASE );
+
+		$cache[ $cache_key ] = $names;
+		return $names;
+	}
+
+	/**
 	 * Check if a user matches an institution's rules (OR logic).
 	 *
 	 * @param int   $user_id  User ID.
@@ -254,12 +300,15 @@ class Institution {
 
 		if ( ! empty( $rules['ip_range'] ) ) {
 			// IP evaluation is page-cache-safe only when the response would be uncached anyway:
-			// the caller flagged it as such, the visitor is logged in, or the visitor carries
-			// the IP-access bypass cookie. A first-time anonymous on-campus visitor landing
-			// directly on a gated post matches none of these and will see the gate — they must
-			// first complete the IP check at /institutional-access (or ?institutional-access=1)
-			// to set the cookie before subsequent gated requests can evaluate their IP.
-			$is_uncached = $uncached || ! empty( $user_id ) || IP_Access_Rule::is_cookie_set();
+			// the caller flagged it as such, the *current visitor* is logged in (so the IP we
+			// would read is theirs), or the visitor carries the IP-access bypass cookie. We
+			// require $user_id === get_current_user_id() to avoid attributing the requestor's
+			// IP to a different user during background metadata sync (admin/cron/webhook). A
+			// first-time anonymous on-campus visitor landing directly on a gated post matches
+			// none of these and will see the gate — they must first complete the IP check at
+			// /institutional-access (or ?institutional-access=1) to set the cookie before
+			// subsequent gated requests can evaluate their IP.
+			$is_uncached = $uncached || ( ! empty( $user_id ) && (int) $user_id === get_current_user_id() ) || IP_Access_Rule::is_cookie_set();
 			if ( $is_uncached && IP_Access_Rule::ip_matches_ranges( IP_Access_Rule::get_visitor_ip(), $rules['ip_range'] ) ) {
 				return true;
 			}
--- a/includes/content-gate/class-institution.php
+++ b/includes/content-gate/class-institution.php
@@ -253,7 +253,7 @@ class Institution {
 		}
 
 		if ( ! empty( $rules['ip_range'] ) ) {
-			$is_uncached = $uncached || ! empty( $user_id ) || isset( $_COOKIE[ IP_Access_Rule::COOKIE_NAME ] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+			$is_uncached = $uncached || ( ! empty( $user_id ) && $user_id === get_current_user_id() ) || isset( $_COOKIE[ IP_Access_Rule::COOKIE_NAME ] ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
 			if ( $is_uncached && IP_Access_Rule::ip_matches_ranges( IP_Access_Rule::get_visitor_ip(), $rules['ip_range'] ) ) {
 				return true;
 			}
--- a/includes/plugins/google-site-kit/class-googlesitekit.php
+++ b/includes/plugins/google-site-kit/class-googlesitekit.php
@@ -324,6 +324,9 @@ class GoogleSiteKit {
 			}
 		}
 
+		// Dedupe by name — two subscriptions can share a display name, and a
+		// group name can collide with an institution title.
+		$names = array_values( array_unique( $names ) );
 		sort( $names, SORT_NATURAL | SORT_FLAG_CASE );
 		return $names;
 	}
--- a/includes/plugins/google-site-kit/class-googlesitekit.php
+++ b/includes/plugins/google-site-kit/class-googlesitekit.php
@@ -260,12 +260,72 @@ class GoogleSiteKit {
 		// If reader has any currently active non-donation subscriptions.
 		$params['is_subscriber'] = empty( $reader_data['active_subscriptions'] ) ? 'no' : 'yes';
 
+		// Content access groups: group subscriptions (owned or member) and matching institutions.
+		if ( Content_Gate::is_newspack_feature_enabled() ) {
+			$group_names     = self::get_user_group_names( $current_user );
+			$params['group'] = empty( $group_names ) ? 'none' : implode( ', ', $group_names );
+		}
+
 		/**
 		 * Filters the custom parameters passed to GA4.
 		 *
 		 * @param array $params Custom parameters sent to GA4.
 		 */
 		return apply_filters( 'newspack_ga4_custom_parameters', $params );
+	}
+
+	/**
+	 * Get the sorted list of group names a user is associated with.
+	 *
+	 * Includes active group subscriptions the user owns or is a member of, and
+	 * institutions whose rules the user matches via any means.
+	 *
+	 * @param \WP_User $user The user to inspect.
+	 * @return string[] Sorted, deduplicated group names.
+	 */
+	private static function get_user_group_names( $user ) {
+		$names = [];
+		if ( ! $user || ! $user->ID ) {
+			return $names;
+		}
+		$user_id = (int) $user->ID;
+
+		// Group subscriptions: owned + member.
+		$candidates = [];
+		if ( class_exists( __NAMESPACE__ . '\Group_Subscription' ) ) {
+			$candidates = Group_Subscription::get_group_subscriptions_for_user( $user_id );
+		}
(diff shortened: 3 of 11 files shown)


includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -268,6 +274,60 @@ function( $category ) {
...
 
+	/**
+	 * Get the sorted list of group names a user is associated with.
+	 *
+	 * Includes active group subscriptions the user owns or is a member of, and
+	 * institutions whose rules the user matches via any means.
+	 *
+	 * @param \WP_User $user The user to inspect.
+	 * @return string[] Sorted, deduplicated group names.
+	 */
```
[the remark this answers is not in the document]
dkoo: Addressed in e2d4b7d4b6263a43b2ff31fc18133fcec1c3c84f — `get_user_group_names()` now dedupes by name via `array_values( array_unique( $names ) )` before sorting. Added a unit test covering the case where two distinct group subs share the same display name.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
```diff
@@ -176,24 +185,85 @@ private static function get_source_labels( $slug, $value, $user_id ) {
...
 							}
 						}
 					}
 					if ( ! empty( $names ) ) {
 						return $names;
 					}
+					// If they don't have an active subscription, check if they're a member of a group subscription.
+					if ( Access_Rules::has_active_subscription( $user_id, $value ) ) {
+						return [ 'group' ];
+					}
```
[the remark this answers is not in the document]
dkoo: Addressed in e2d4b7d4b6263a43b2ff31fc18133fcec1c3c84f — get_source_labels() now does the strict (ownership) check first, so an owner of a sub matching an `any subscription` rule (empty `$value`) is no longer mislabeled as `group`. Added a unit test covering the empty-value owned case.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
```diff
@@ -176,24 +185,85 @@ private static function get_source_labels( $slug, $value, $user_id ) {
...
+						continue;
+					}
+					if ( ! $subscription->has_status( WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES ) ) {
+						continue;
+					}
+					foreach ( $value as $product_id ) {
+						if ( $subscription->has_product( $product_id ) ) {
+							$group_settings  = Group_Subscription_Settings::get_subscription_settings( $subscription );
+							$group_names[]   = wp_specialchars_decode( $group_settings['name'] );
+							$seen[ $sub_id ] = true;
```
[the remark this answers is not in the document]
dkoo: Addressed in e2d4b7d4b6263a43b2ff31fc18133fcec1c3c84f — get_group_labels() now treats an empty/non-array `$value` as `match any product`, so active group subs surface their names regardless. Added a unit test covering the empty-value group-member case.

dkoo:
Addressed all blockers and most suggestions from the code review in 7426ab3d0.

**Blockers**
- **B1**: Restored the `Institution::user_matches_institution()` IP-cache tightening that was lost in the trunk merge. The IP check now only fires when `$user_id === get_current_user_id()` (or the cache-bypass cookie is set), so the requestor's IP can no longer leak to non-current users during background metadata sync. **B2** is resolved by B1.
- **B3**: Added an `is_array()` guard in `get_group_labels()`'s institution branch (now folded into the new helper). A malformed institution rule (empty / null / scalar `value`) no longer fatals — confirmed by a new data-provider regression test.
- **B4**: `Group_Subscription::get_group_subscriptions_for_user()` now checks `Group_Subscription_Settings::get_subscription_settings($sub)['enabled']` directly instead of going through `is_group_subscription()`, so the My-Account/Memberships UI-visibility suppression no longer bleeds into the data layer.

**Suggestions**
- **S2 + S3 + S4 + S7 + S10**: Extracted two reusable helpers — `Group_Subscription::get_group_names_for_user( $user_id, ?array $product_filter = null )` and `Institution::get_matching_names_for_user( $user_id, ?array $institution_filter = null )`. Both memoize per request (S4), read `get_subscription_settings()` once (S3), decode names with `html_entity_decode( ENT_QUOTES | ENT_HTML5 )` so `&mdash;` / `&#8217;` / etc. survive ESP/GA4 transport (S7), and use `WooCommerce_Connection::get_active_subscriptions_for_user()` for the owned path so status + gifting filters are centralized (S10). Both `Content_Gate::get_group_labels()` and `GoogleSiteKit::get_user_group_names()` delegate to these.
- **S6**: Dropped the `class_exists()` guards in the GA4 helper (the classes ship in the same plugin).
- **S8**: GA4 helper now early-returns when `! Reader_Activation::is_user_reader( $user )`, consistent with the surrounding `is_reader` / `is_subscriber` params.
- **S9**: Replaced the brittle one-shot filter ordering in `test_subscription_filter_source` with an unconditional `__return_true` filter + unregistered product, so the strict per-product lookup naturally falls through regardless of how many times the filter fires.
- **S5**: Added regression tests for malformed institution rule values (data provider covering empty string, empty array, null, scalar int, scalar string) and a cross-user IP leak prevention test.

All 1136 tests pass.

**Deferred**
- **S1** (string label change: `'Subscription'` → `'subscription'`, `'group'` reused for group-membership subscription access vs institution): kept the new taxonomy because it's clearer, but happy to discuss release notes / segmentation guidance separately.
- **Nits** N1–N6: open to addressing in this PR if you'd like, otherwise good follow-up candidates.

includes/content-gate/class-access-rules.php
```diff
@@ -395,8 +396,9 @@ public static function has_active_subscription( $user_id, $product_ids ) {
 		 * @param bool  $has_subscription Whether the user has an active subscription.
 		 * @param int   $user_id          User ID.
 		 * @param array $product_ids      Required product IDs.
+		 * @param bool  $strict           If true, only consider active subscriptions owned by $user_id (ignore group subscription memberships).
 		 */
-		return apply_filters( 'newspack_access_rules_has_active_subscription', $has_subscription, $user_id, $product_ids );
+		return apply_filters( 'newspack_access_rules_has_active_subscription', $has_subscription, $user_id, $product_ids, $strict );
```
adekbadek, line 411: **Suggestion:** `$strict` flag doesn't constrain the `newspack_access_rules_has_active_subscription` filter — `$strict` is applied only to the built-in ownership/group checks; the `apply_filters( 'newspack_access_rules_has_active_subscription', … )` at line 401 runs in the non-strict path either way. `newspack-network`'s `class-access.php` registers this filter (`accepted_args=3`, unaware of the new 4th arg) and grants access network-wide, so a user with only a *network* subscription returns `true` for the strict-ownership probe in `get_source_labels()` (`Access_Rules::has_active_subscription( $user_id, $value, true )`) — they're then classified as a local owner and the gate's configured product names are enumerated as their `Content_Access_Source`. The names happen to be the gate's own products so the output isn't *wrong*, but the `$strict` contract (“only active subscriptions owned by `$user_id`”) isn't honored when a filter participates. Either short-circuit the filter when `$strict` is true, or add a doc note that `$strict` only constrains the built-in checks and filter authors should respect the 4th arg.

includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -260,6 +260,12 @@ function( $category ) {
 		// If reader has any currently active non-donation subscriptions.
 		$params['is_subscriber'] = empty( $reader_data['active_subscriptions'] ) ? 'no' : 'yes';
 
+		// Content access groups: group subscriptions (owned or member) and matching institutions.
+		if ( Content_Gate::is_newspack_feature_enabled() ) {
+			$group_names     = self::get_user_group_names( $current_user );
```
adekbadek: **Blocker:** `get_user_group_names()` adds uncached DB work to every logged-in pageview — Invoked from `get_custom_event_parameters()` (the `googlesitekit_gtag_opt` filter) — once per `wp_footer` render for every logged-in *reader* on a content-gate site. Per pageview it does `wcs_get_users_subscriptions()` + a `wcs_get_subscription()` per sub (via `WooCommerce_Connection::get_active_subscriptions_for_user()`) + a `get_user_meta()` + per-sub `Group_Subscription_Settings::get_subscription_settings()` reads, plus a loop over the institution list running `Institution::user_matches_institution()` per institution — none of it cached. `7426ab3d` improved this (extracted into per-request-memoized helpers; non-readers now early-return) but didn't remove the cost: the memoization only de-dupes repeated calls within one request, and the gtag builder calls it exactly once, so the per-pageview hit is unchanged for the population this affects. Two lines up, `is_subscriber` already reuses the cached `Reader_Data` payload (`active_subscriptions`); source the owned-sub IDs from there, or memoize the gtag params themselves — should land before the feature ships.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -240,4 +251,95 @@ public static function get_group_subscriptions_for_user( $user_id, $ids_only = f
...
+	 *
+	 * @param int        $user_id        User ID.
+	 * @param array|null $product_filter Optional list of product IDs. If non-empty, only subscriptions
+	 *                                   containing at least one of these products contribute a name.
+	 *                                   Pass null or an empty array to include every active group sub.
+	 *
+	 * @return string[] Sorted, deduplicated group names.
+	 */
+	public static function get_group_names_for_user( $user_id, $product_filter = null ) {
+		static $cache = [];
```
adekbadek: **Suggestion:** New static name caches: no reset hook, unbounded across users, request-stable assumption — `Group_Subscription::get_group_names_for_user()` (line 268) and `Institution::get_matching_names_for_user()` (`class-institution.php:251`) memoize on `(user_id, filter)` in a function-local `static $cache` with no eviction and no `reset_cache()` accessor (unlike `Content_Gate::reset_cache()`). In a long-running process — a WP-CLI bulk metadata sync, or an Action Scheduler worker processing many `newspack_esp_sync_contact` jobs in one PHP run — this accumulates one entry per processed contact for the worker's lifetime and serves a stale list if a subscription/membership changes between jobs in the same worker. Separately, `get_matching_names_for_user`'s value depends on `get_current_user_id()` / the visitor IP / the cache-bypass cookie for `ip_range` rules, none of which are in the cache key — latent today (no live call site evaluates the same `$user_id` under a different current-user context within a request, and the current user is stable per request) but worth folding into the key or noting in a comment. Adding a `reset_cache()` would also help test isolation.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -229,7 +229,18 @@ public static function get_group_subscriptions_for_user( $user_id, $ids_only = f
 		$subscription_ids = array_map( 'absint', \get_user_meta( $user_id, self::GROUP_SUBSCRIPTION_USER_META_KEY, false ) );
 		$subscriptions    = [];
 		foreach ( $subscription_ids as $subscription_id ) {
-			$subscriptions[] = $ids_only ? $subscription_id : \wcs_get_subscription( $subscription_id );
+			$subscription = \wcs_get_subscription( $subscription_id );
+			if ( ! $subscription ) {
+				continue;
+			}
+			// Check the group-enabled meta directly rather than calling self::is_group_subscription(),
```
adekbadek, line 279: **Suggestion:** Decoupling from `is_group_subscription()` re-exposes member group subs in My Account when Woo Memberships is active — `get_group_subscriptions_for_user()` now reads the `enabled` meta directly instead of calling `is_group_subscription()` — correct for data-layer callers, but it removes the `is_account_page()` short-circuit that suppressed group-subscription behavior in My Account on Woo Memberships sites. `Group_Subscription_MyAccount::inject_member_group_subscriptions()` (hooked on `wcs_get_users_subscriptions`, in `class-group-subscription-myaccount.php`) therefore now injects a member's group subscriptions into the My-Account subscriptions list even when Memberships is active — the exact scenario that short-circuit suppressed; they render as plain (non-group) subscriptions there since `is_group_subscription()` still returns `false` on the account page, so the group-specific UI is skipped. Visible behavior change for Memberships sites — confirm it's intended, or gate `inject_member_group_subscriptions()` on `! Memberships::is_active()` until Access Control fully replaces Memberships.

includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -260,6 +260,12 @@ function( $category ) {
...
+			$params['group'] = empty( $group_names ) ? 'none' : implode( ', ', $group_names );
```
adekbadek: **Nit:** `group` GA4 param + `NP_Content_Access_Group` are a new (PII-adjacent) data category leaving the site — The `group` gtag param and the new `NP_Content_Access_Group` ESP merge field carry publisher-defined group-subscription names and institution titles. For an *unnamed* group subscription, `Group_Subscription_Settings::get_subscription_settings()` falls back to `"{owner's billing full name}'s Group"` — so a group owner's legal name now flows to Google Analytics and to ESP records for every member of that group. It's a deliberate consequence of opting into content gates and the GA surface is feature-flag-gated, but it's a new class of data leaving the site — worth a publisher-facing docs note, and maybe reconsidering the unnamed-group default.

includes/content-gate/class-institution.php
```diff
@@ -235,6 +235,52 @@ public static function evaluate( $user_id, $institution_ids ) {
...
+			return $cache[ $cache_key ];
+		}
+
+		$names = [];
+		foreach ( self::get_cached_institutions() as $inst_id => $rules ) {
+			if ( null !== $normalized_filter && ! in_array( (int) $inst_id, $normalized_filter, true ) ) {
+				continue;
+			}
+			if ( self::user_matches_institution( $user_id, $rules ) ) {
+				$names[] = html_entity_decode( (string) get_the_title( $inst_id ), ENT_QUOTES | ENT_HTML5, 'UTF-8' );
```
adekbadek: **Nit:** `get_the_title()` runs the `the_title` filter on a string headed for GA4/ESP — `html_entity_decode( (string) get_the_title( $inst_id ), … )` — `get_the_title()` runs the `the_title` filter (texturize, plus anything third-party plugins hook), so the value can pick up entities that `html_entity_decode` then partly undoes, or even markup that won't be stripped before it goes to GA4/the ESP. `get_post_field( 'post_title', $inst_id )` (or `wp_strip_all_tags( get_the_title( … ) )`) would be a more predictable source for a transport string. Low impact.

includes/content-gate/class-institution.php
```diff
@@ -254,12 +300,15 @@ public static function user_matches_institution( $user_id, $rules, $uncached = f
...
-			$is_uncached = $uncached || ! empty( $user_id ) || IP_Access_Rule::is_cookie_set();
+			// the caller flagged it as such, the *current visitor* is logged in (so the IP we
+			// would read is theirs), or the visitor carries the IP-access bypass cookie. We
+			// require $user_id === get_current_user_id() to avoid attributing the requestor's
+			// IP to a different user during background metadata sync (admin/cron/webhook). A
+			// first-time anonymous on-campus visitor landing directly on a gated post matches
+			// none of these and will see the gate — they must first complete the IP check at
+			// /institutional-access (or ?institutional-access=1) to set the cookie before
+			// subsequent gated requests can evaluate their IP.
+			$is_uncached = $uncached || ( ! empty( $user_id ) && (int) $user_id === get_current_user_id() ) || IP_Access_Rule::is_cookie_set();
```
adekbadek, line 379: **Nit:** `IP_Access_Rule::is_cookie_set()` branch still allows cross-user IP attribution; the comment doesn't note it — The `(int) $user_id === get_current_user_id()` clause closes the common leak (a frontend pageview's IP being attributed to whichever `$user_id` a synchronous ESP sync happens to evaluate during that request). But the `|| IP_Access_Rule::is_cookie_set()` disjunct still treats the request as “uncached” for an *arbitrary* `$user_id` when the visitor carries the cache-bypass cookie — so a cookie-bearing visitor can cause the current request's IP to be matched against another user's institution range. Narrow and largely pre-existing, but the comment at line 305 says the `=== get_current_user_id()` check is there “to avoid attributing the requestor's IP to a different user” without noting the cookie branch reopens that for cookie-bearing visitors. Worth a one-line caveat in the comment, or scoping the cookie branch to the current user too.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
```diff
@@ -173,27 +180,70 @@ private static function get_access_source_labels( $evaluations, $user_id ) {
...
+
+			case 'institution':
+				// A malformed institution rule (missing/empty/scalar value) matches everyone
+				// per Institution::evaluate(), but there's no specific institution to attribute.
+				if ( ! is_array( $value ) || empty( $value ) ) {
+					return [];
+				}
+				return Institution::get_matching_names_for_user( $user_id, $value );
 
 			default:
```
adekbadek, line 251: **Nit:** A `reader_data`-only access rule yields no `Content_Access_Source` and no `Content_Access_Group` — When a gate grants access purely via a `reader_data` rule, `get_source_labels()` and `get_group_labels()` both fall through to `default => []`, so the contact gets `Content_Access = Yes` but `Content_Access_Source = ''` and `Content_Access_Group = ''`. The empty `Content_Access_Source` for `reader_data` is pre-existing trunk behavior; the new `Content_Access_Group` inherits the same gap. (Institutions matched *via* `reader_data` are handled — only the bare `reader_data` rule type isn't.) Probably fine; worth a one-line confirmation against the NPPD-1451 schema whether a “reader data” label is expected here.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
```diff
@@ -173,27 +180,70 @@ private static function get_access_source_labels( $evaluations, $user_id ) {
...
 							if ( $product ) {
-								$names[] = $product->get_name();
+								$names[] = html_entity_decode( (string) $product->get_name(), ENT_QUOTES | ENT_HTML5, 'UTF-8' );
 							}
 						}
 					}
-					if ( ! empty( $names ) ) {
-						return $names;
-					}
+					return ! empty( $names ) ? $names : [ 'subscription' ];
```
adekbadek, line 199: **Nit:** Owned-subscription fallback string lowercased: `'Subscription'` → `'subscription'` — On trunk, the owned-subscription source-label fallback (owner matched, but no product name resolved) was `'Subscription'` with a capital S; the rewrite emits `'subscription'`. Worth confirming the lowercasing is intentional rather than an accident of the rewrite. (The broader value-vocabulary changes — `institution` rules now labeled `institution` rather than `group`, group members labeled `group` rather than the product name — are acknowledged and accepted since the Access Control feature is still in alpha, so no migration note is needed; this is just the capitalization.)

tests/unit-tests/plugins/google-site-kit.php
```diff
@@ -0,0 +1,231 @@
...
+	 */
+	private static $owner_id;
+
+	/**
+	 * Enable the content gating feature flag and load WC mocks.
+	 */
+	public static function set_up_before_class() {
+		parent::set_up_before_class();
+		if ( ! defined( 'NEWSPACK_CONTENT_GATES' ) ) {
+			define( 'NEWSPACK_CONTENT_GATES', true );
```
adekbadek, line 48: **Nit:** `define( 'NEWSPACK_CONTENT_GATES', true )` in `set_up_before_class` leaks to the rest of the suite — Constants can't be undefined, so once this test class runs, `NEWSPACK_CONTENT_GATES` stays `true` for every subsequent test class in the same PHP process — any test that assumes the content-gate feature is off would see it on. The suite passes today, a quick grep confirms nothing else asserts the flag is off in tests, and this matches the codebase's existing pattern for constant-gated features — flagging only so it's a conscious choice (the GA4 `group` param only exists when the flag is on, so there isn't an obviously cleaner option).

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -240,4 +251,95 @@ public static function get_group_subscriptions_for_user( $user_id, $ids_only = f
...
+		}
+		$cache_key = $user_id . '|' . ( null === $normalized_filter ? '' : implode( ',', $normalized_filter ) );
+		if ( isset( $cache[ $cache_key ] ) ) {
+			return $cache[ $cache_key ];
+		}
+
+		$candidates = [];
+
+		// Owned active subscriptions, already filtered by status (and product, if provided) and gifting.
+		$owned_ids = WooCommerce_Connection::get_active_subscriptions_for_user(
```
adekbadek, line 382: **Nit:** Gifting filter applied to owned subs but not member subs in `get_group_names_for_user()` — Owned candidates come from `WooCommerce_Connection::get_active_subscriptions_for_user()`, which excludes gifted subscriptions where the user isn't the recipient; the member branch (`get_group_subscriptions_for_user()`) applies no gifting check. So a gifted group subscription would be excluded if encountered via ownership but included via membership. This mirrors the existing asymmetry in `Access_Rules::has_active_subscription()`'s non-strict branch, so it's consistent with current behavior — but it's a latent edge case worth a comment if gifted group subscriptions are a real scenario.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
```diff
@@ -150,14 +157,14 @@ private static function get_access_source_labels( $evaluations, $user_id ) {
 					if ( ! $rule['passes'] ) {
 						continue;
 					}
-					foreach ( self::get_source_labels( $rule['slug'], $rule['value'], $user_id ) as $label ) {
-						$sources[ $label ] = true;
+					foreach ( call_user_func( $resolver, $rule['slug'], $rule['value'], $user_id ) as $label ) {
```
adekbadek: **Nit:** `call_user_func( $resolver, … )` can be `$resolver( … )` — In `collect_labels()`, `$resolver` is always `[ self::class, 'get_source_labels' ]` or `[ self::class, 'get_group_labels' ]`, so `$resolver( $rule['slug'], $rule['value'], $user_id )` (or a first-class callable at the call site) works directly — cosmetic, slightly cleaner than `call_user_func()`.

adekbadek (changes requested):
Request changes — one blocker: `get_user_group_names()` adds uncached per-pageview DB work to the GA4 config path on content-gate sites; should be fixed before the feature ships (details + suggested fix inline). The rest of the PR is in good shape — the cross-user IP-attribution risk is closed with a regression test, the shared name helpers are clean, PHP suites are green in CI; a few suggestions and nits inline.

includes/plugins/google-site-kit/class-googlesitekit.php
dkoo: Partially addressed in 15488f71a. The per-pageview cost is now lower:
- GA4 path uses anonymized IDs (`Group {id}` / `Institution {id}`) via a new `Group_Subscription::get_group_ids_for_user()` / `Institution::get_matching_ids_for_user()` pair. No name-resolution work on the GA4 path — `get_subscription_settings()` is only called for callers that need names.
- Static `reset_cache()` is now hookable and is fired on subscription status changes and group-member user-meta events, so CLI workers don't accumulate stale entries.

What's NOT yet done in this PR: cross-request memoization (transient / persistent object cache) of the per-user result. The remaining per-pageview cost is one `wcs_get_users_subscriptions()` + per-sub `get_subscription_settings()` + the institution loop, all of which lean on WC/WP internal caches. That's a real win over the previous full path but doesn't drop to zero. I'd like to land this PR with the current improvement and open a follow-up issue for the cross-request layer — it needs proper invalidation orchestration across subscription / membership / institution change hooks, and reasonable as its own change. Happy to take it as a blocker on this PR if you'd prefer.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
dkoo: Addressed in 15488f71a. Both helpers now have:
- Class-level static cache (was function-local).
- Public `reset_cache()` (`Group_Subscription::reset_cache()`) and `reset_matching_cache()` (`Institution::reset_matching_cache()`) accessors. Tests now call these in `set_up`/`tear_down`.
- `Group_Subscription::init()` hooks `reset_cache()` to `woocommerce_subscription_status_updated` and to `added_user_meta` / `updated_user_meta` / `deleted_user_meta` for the group-member meta key, so long-running CLI workers don't accumulate stale entries across jobs.
- `Institution::get_matching_map_for_user()` cache key now includes `get_current_user_id()` so the IP-rule branch's current-user dependency doesn't produce stale hits.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
dkoo: Addressed in 15488f71a. `Group_Subscription_MyAccount::inject_member_group_subscriptions()` now early-returns when `Memberships::is_active()`, restoring the previous behavior on Memberships sites. The suppression now lives at the UI layer (where it belongs) rather than inside the shared `is_group_subscription()` predicate that data-layer callers also use.

includes/plugins/google-site-kit/class-googlesitekit.php
dkoo: Addressed in 15488f71a. The GA4 `group` param now emits anonymized identifiers (`Group {sub_id}`, `Institution {inst_id}`) instead of human-readable names, so the unnamed-group fallback's owner billing full name never leaves the site via GA4. The ESP path (`NP_Content_Access_Group` contact metadata) still emits names, since that's a deliberate publisher-driven sync surface where the value's semantic identity matters. New test `test_group_never_emits_publisher_facing_names` asserts the anonymization contract.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
dkoo: Addressed in 15488f71a — `get_source_labels()` now returns `[ 'reader_data' ]` for the `reader_data` rule slug, so a reader-data-only rule no longer yields an empty `Content_Access_Source`. (`Content_Access_Group` remains empty for this rule type, which I think is correct: `reader_data` doesn't designate a specific group / institution / product, so there's nothing concrete to attribute as a group identifier.)

includes/reader-activation/sync/contact-metadata/class-content-gate.php
dkoo: Confirmed intentional — the value vocabulary is normalized to lowercase across the board (`subscription` / `group` / `institution` / `domain` / `reader_data`) for consistency. As you noted, the broader vocabulary changes are accepted while the feature is in alpha, so no migration note needed.

includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
dkoo: Addressed in 15488f71a with an inline docblock on `Group_Subscription::get_settings_map_for_user()` explicitly calling out the gifting-filter asymmetry between the owned path (`WooCommerce_Connection::get_active_subscriptions_for_user()`, which filters gifted subs the user isn't the recipient of) and the member path (`get_group_subscriptions_for_user()`, which doesn't). Behavior unchanged — mirrors the pre-existing pattern in `Access_Rules::has_active_subscription()`.

tests/unit-tests/plugins/google-site-kit.php
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	/**
+	 * Clean up after each test.
+	 */
+	public function tear_down() {
+		delete_user_meta( self::$user_id, Group_Subscription::GROUP_SUBSCRIPTION_USER_META_KEY );
+		Institution::invalidate_cache();
+		wp_set_current_user( 0 );
+		parent::tear_down();
```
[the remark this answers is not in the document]
dkoo: Addressed in 15488f71a. Institution post IDs created via `Institution::create()` are now tracked in a `$institution_ids` array and deleted (plus `Institution::TRANSIENT_KEY` cleared) in `tear_down()`. Also reset the new `Group_Subscription` / `Institution` per-request caches in both `set_up` and `tear_down` for stronger isolation.

tests/unit-tests/content-gate/class-content-gate-metadata.php
```diff
@@ -23,19 +26,125 @@ class Newspack_Test_Content_Gate_Metadata extends WP_UnitTestCase {
...
+	}
+
+	/**
+	 * Clean up after each test.
+	 */
+	public function tear_down() {
+		delete_user_meta( self::$user_id, Group_Subscription::GROUP_SUBSCRIPTION_USER_META_KEY );
+		Institution::invalidate_cache();
+		parent::tear_down();
+	}
```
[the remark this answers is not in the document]
dkoo: Addressed in 15488f71a — same fix as the parallel comment in `google-site-kit.php`. `create_institution()` now appends to a class `$institution_ids` array and `tear_down()` deletes them + clears `Institution::TRANSIENT_KEY`.

includes/content-gate/class-access-rules.php
dkoo: Addressed in 15488f71a with a docblock note on `Access_Rules::has_active_subscription()` documenting that `$strict` only constrains the built-in ownership / group-membership checks, and that filter authors must opt into the 4th `$strict` arg and respect it. Going with the doc-only option here — short-circuiting the filter when `$strict=true` would break newspack-network's current grant intent. Will open a follow-up to update newspack-network's filter callback to accept the 4th arg and short-circuit appropriately.

includes/content-gate/class-institution.php
dkoo: Addressed in 15488f71a — `Institution::get_matching_map_for_user()` now reads titles via `get_post_field( 'post_title', $inst_id )` instead of `get_the_title( $inst_id )`, so the value heading to GA4/ESP isn't run through `the_title` filters.

includes/content-gate/class-institution.php
dkoo: Addressed in 15488f71a with a one-line caveat in the inline comment, noting that the `IP_Access_Rule::is_cookie_set()` disjunct still accepts any non-current `$user_id` when the visitor carries the cache-bypass cookie. Leaving the behavior itself as-is — the cookie is only set after the visitor has already passed an institutional-access check on this site, so the cross-attribution window is narrow. Easy to tighten later if it becomes a concern.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
dkoo: Addressed in 15488f71a — `collect_labels()` now invokes `$resolver( $rule['slug'], $rule['value'], $user_id )` directly. (The `call_user_func()` had been added by the Copilot autofix in d350ee47 in response to an early Copilot review that incorrectly claimed array callables would fatal under direct invocation; PHP 7+ handles this fine.)

includes/content-gate/class-access-rules.php
dkoo: Follow-up: No changes needed for the Newspack Network implementation of the `newspack_access_rules_has_active_subscription` filter hook, for now. Network's filter callback only filters this hook so it can check whether a user owns a network subscription that was propagated throughout the network. Subscriptions are currently only synced across the network for owned subscriptions, not group memberships. If we augment this behavior to sync group subscription memberships across the network in the future, we will want to add handling for the `$strict` param in Network's filter callback to differentiate between owned vs. group membership subscriptions.

includes/plugins/google-site-kit/class-googlesitekit.php
adekbadek: `15488f71a` covers the bulk: anonymized IDs remove the name work, the per-request memo returns 0 queries on a repeat call (verified locally), and the `reset_cache` hooks on `woocommerce_subscription_status_updated` + the group-member `*_user_meta` bust it.

Worth asking whether the cross-request transient is needed at all. Residual per logged-in pageview: ~7 queries for the owned-sub walk + ~2 for institutions on a cold transient (which amortizes; the sub walk doesn't). Bounded and warm-cache-friendly within the request.

A per-user transient also can't safely cover the institution side: `user_matches_institution()`'s IP branch matches the *current visitor's* IP and short-circuits unless `$user_id === get_current_user_id()`. A `prefix_u{uid}` key with no visitor dimension would cache an on-campus match and serve it off-campus — the class of bug the cross-user IP-leak regression test added in this PR guards against.

Proposed: land as-is plus a one-line docblock on `get_user_group_labels()` noting it's deliberately request-scoped because the institution branch is per-visitor. Follow-up *issue* contingent on profiling, not a committed PR.

adekbadek (approved):
Round-2 changes (`7426ab3d` + `15488f71a`) cover the earlier feedback — see resolved threads. The GA4 path is fine to land as-is for the reasons in the blocker thread.

CI is currently red on `lint-php` and `build-distributable`, but both fail with the same composer auth error cloning `github.com/joshtronic/php-loremipsum.git` — infrastructure, not the diff. `test-php`, `test-js`, `lint-js-scss` all pass; a re-run should clear it.

Optional: the one-line docblock on `get_user_group_labels()` from the blocker thread.

tests/unit-tests/plugins/google-site-kit.php
dkoo: I think it's okay to move ahead with this since the constant requirement is temporary and will be removed in the near future, WDYT?

includes/plugins/google-site-kit/class-googlesitekit.php
dkoo: Agreed, I think the cross-request transient is unnecessary at this point, and the per-request memoization is the more important performance consideration for now. 7a3cdbcd0 adds a line to the docblock for `get_user_group_labels()` as suggested.

PR #4700, opened 2026-05-04
refactor(access-control): scope gate layout cache to current user

dkoo:
## Summary

Removes the `Content_Restriction_Control::$user_id` "first-caller-wins" static and keys the gate-cache lookups (`get_gate_post_id`, `get_gate_layout_id`) on `get_current_user_id()` directly.

Follow-up to #4695 — see [the inline thread on the original PR](https://github.com/Automattic/newspack-plugin/pull/4695#discussion_r3180597018) for context. @adekbadek flagged that the static was "whoever called `is_post_restricted` first in this request, and `get_gate_layout_id` reads it back. That works for the page-render path (first call IS the viewer) but if any code path – a REST callback, an early integration call, a queue worker – calls `is_post_restricted` with a non-current-user id before the page render, the layout lookup silently returns the wrong layout."

## What changed

- Drops `private static $user_id = 0;` from `Content_Restriction_Control`.
- Drops the seed-on-first-call block in `is_post_restricted` (the static no longer exists).
- `get_gate_post_id()` and `get_gate_layout_id()` resolve `$user_id = get_current_user_id()` at lookup time. They only return a cache entry written for the actual current user.
- `is_post_restricted()` already wrote cache entries keyed by its call-local `$user_id` (unchanged from PR #4695). Calls made for other users (`Premium_Newsletters::process_queue`, REST callbacks, integration hooks) populate their own cache slots and are inert to the page-render lookup.

No public API change — the dropped static was `private`, and the three method signatures (`is_post_restricted`, `get_gate_post_id`, `get_gate_layout_id`) are unchanged.

## Test changes

- Replaces `test_is_post_restricted_seeds_user_id_from_first_caller_only` (which pinned the now-removed seeding contract) with `test_get_gate_layout_id_does_not_return_other_users_cached_layout`. The new test sets up the queue-worker scenario directly: calls `is_post_restricted(false, $post_id, $queue_user)`, switches `wp_set_current_user($page_user)`, and asserts `get_gate_layout_id` and `get_gate_post_id` return `false` (no cross-user leak). RED on the prior code (returned a stale layout ID); GREEN here.
- Removes the obsolete `get_cached_user_id()` reflection helper and the `self::$user_id` reset from `reset_restriction_cache()`.

## Base / merge order

Targets `fix/ip-access-plus-registered-access-rules` (PR #4695) so the diff shows only the refactor. Once #4695 merges, this PR can be rebased onto `trunk`.

## Test plan

- [ ] PHPUnit suite is green locally (`n test-php` from `repos/newspack-plugin`): 1054 tests, 1 unrelated pre-existing skip.
- [ ] Manual smoke: gated post renders the right gate for the current viewer when `Premium_Newsletters` queue worker has previously run for other users in the same request.

### DIFF
--- a/includes/content-gate/class-access-rules.php
+++ b/includes/content-gate/class-access-rules.php
@@ -296,9 +296,10 @@ class Access_Rules {
 	 *
 	 * @param int   $user_id     User ID.
 	 * @param array $product_ids Required product IDs.
+	 * @param bool  $strict      If true, only consider active subscriptions owned by $user_id (ignore group subscription memberships).
 	 * @return bool
 	 */
-	public static function has_active_subscription( $user_id, $product_ids ) {
+	public static function has_active_subscription( $user_id, $product_ids, $strict = false ) {
 		$has_subscription = false;
 
 		// Check user's own subscriptions.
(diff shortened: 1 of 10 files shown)


dkoo:
Closing —  picked up the wrong head branch. Reopening with the correct refactor branch.

PR #4702, opened 2026-05-05
fix(content-gate): pass redirectToLayout explicitly on Save

miguelpeixe:
### All Submissions:

p1778008459631959-slack-C0ANXE4NDT3

When saving a new gate from Audience → Access Control, the user is being redirected to the gate layout editor with a malformed query string:

```
/wp-admin/admin.php?action=newspack_edit_gate_layout[...]gate_mode=%5Bobject%20Object%5D
```

The root cause is the `handleCreate` being passed directly as the action for the Save button. When the button fires, the click handler is invoked with the synthetic event as the first argument. The if ( redirectToLayout !== '' ) guard passes, and getEditGateLayoutUrl( data.id, event ) is called. addQueryArgs stringifies the event to [object Object].

TypeScript didn't catch this because the = '' default makes handleCreate's signature assignable to () => void.

The fix is to wrap the action so the event is discarded.

1. Go to Audience → Access Control → Add new gate.
2. Configure rules so the gate can be saved (title + at least one content rule + an active mode), then click Save.
3. Confirm you land on the #/content-gates list view
4. Open an existing gate and click Configure registration / Configure custom access — should still navigate to the layout editor with the correct gate_mode.

### Other information:

### DIFF
--- a/src/wizards/audience/views/content-gates/edit/index.tsx
+++ b/src/wizards/audience/views/content-gates/edit/index.tsx
@@ -336,7 +336,7 @@ const Edit = ( { match, updateGatesData, slug = AUDIENCE_CONTENT_GATES_WIZARD_SL
 			{
 				type: 'primary',
 				label: __( 'Save', 'newspack-plugin' ),
-				action: isNew ? handleCreate : handleSave,
+				action: isNew ? () => handleCreate() : handleSave,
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works!

PR #4703, opened 2026-05-05
fix: remove colour label changes to avoid bleeding into other blocks

laurelfulford:
### All Submissions:

The Author Profile block's Social Links block includes some code that relabels the Color and Background colour pickers to Icon Color and Icon Background. This is having an unintended side effect and bleeding into other block settings.

This PR changes the approach to match how this is handled in Core's Social Links block, to (hopefully!) make it a little less fragile (the alternative seemed to be checking if the Author Profile social links block was active again and again, and adding/removing the label, but that seemed subpar). 

Another option would be just to show the default 'Text' prefix. It's not as nice UX-wise, but much simpler under the hood. It doesn't match Core's Social Links block, but it matches the labelling in Core's new Icon block coming in WP 7.0.

Closes NPPD-1458

1. Using the Block Theme, edit one of the Single Post templates in the Site Editor (after resetting it if it has saved changes). These templates will already have the Author Profile block in place.
2. In the site editor, click on any of the other blocks -- like the Post Title block -- and note the colour settings:

CleanShot 2026-05-05 at 14 37 06

4. Refresh the editor, and ensure standard blocks are no longer inheriting the 'Icon' prefix on the colour picker.
5. Edit the Author Profile block and select the Social Icons block.
6. Confirm the Colors panel shows up on the Style tab, under the Styles panel 
7. Ensure each colour has the prefix 'Icon'
8. Try togglng the block between Default and Brand styles, and confirm that the colour pickers go away when "Brand" is picked, and appear again when "Default" is picked.
9. Try changing the icon colours and saving - confirm colours can be changed and saved, and switching back to the Brand style switches back to the branded colours.

### Other information:

### DIFF
--- a/src/blocks/author-profile-social/block.json
+++ b/src/blocks/author-profile-social/block.json
@@ -19,6 +19,24 @@
 		"iconSize": {
 			"type": "number",
 			"default": 24
+		},
+		"iconColor": {
+			"type": "string"
+		},
+		"customIconColor": {
+			"type": "string"
+		},
+		"iconColorValue": {
+			"type": "string"
+		},
+		"iconBackgroundColor": {
+			"type": "string"
+		},
+		"customIconBackgroundColor": {
+			"type": "string"
+		},
+		"iconBackgroundColorValue": {
+			"type": "string"
 		}
 	},
 	"styles": [
@@ -26,13 +44,6 @@
 		{ "name": "brand", "label": "Brand" }
 	],
 	"supports": {
-		"color": {
-			"enableContrastChecker": false,
-			"background": true,
-			"text": true,
-			"link": false,
-			"__experimentalSkipSerialization": [ "text", "background" ]
-		},
 		"html": false,
 		"layout": {
 			"allowSwitching": false,
--- a/src/blocks/author-profile-social/class-author-profile-social-block.php
+++ b/src/blocks/author-profile-social/class-author-profile-social-block.php
@@ -221,8 +221,6 @@ final class Author_Profile_Social_Block {
 
 	/**
 	 * Get wrapper attributes (class, style, etc.) for the block.
-	 * Sets block context so core includes default class, custom className, and other supports.
-	 * Color serialization is skipped via block.json so colors are applied only as CSS vars.
 	 *
 	 * @param WP_Block $block      Block instance.
 	 * @param array    $attributes Block attributes.
@@ -230,49 +228,30 @@ final class Author_Profile_Social_Block {
 	 * @return string HTML attributes for the wrapper element.
 	 */
 	private static function get_block_wrapper_attributes( WP_Block $block, array $attributes, int $icon_size ): string {
-		$previous = \WP_Block_Supports::$block_to_render ?? null;
-		\WP_Block_Supports::$block_to_render = $block->parsed_block;
-
-		$wrapper_attributes = get_block_wrapper_attributes(
+		return get_block_wrapper_attributes(
 			[
 				'class' => 'author-profile-social__list',
 				'style' => self::get_wrapper_style( $attributes, $icon_size ),
 			]
 		);
-
-		\WP_Block_Supports::$block_to_render = $previous;
-
-		return $wrapper_attributes;
 	}
 
 	/**
-	 * Convert a preset token (var:preset|type|slug) to a CSS variable reference.
-	 *
-	 * @param string $value Raw value, e.g. "var:preset|color|primary" or "#fff".
-	 * @return string CSS value, e.g. "var(--wp--preset--color--primary)" or "#fff".
-	 */
-	private static function preset_to_css( string $value ): string {
-		if ( preg_match( '/^var:preset\|([^|]+)\|(.+)$/', $value, $matches ) ) {
-			return sprintf( 'var(--wp--preset--%s--%s)', $matches[1], $matches[2] );
-		}
-		return $value;
-	}
-
-	/**
-	 * Resolve a color value from attributes (preset slug or custom style token).
+	 * Resolve a color value from the block's icon color attributes.
+	 * Prefers the preset slug (so theme switches reflect new palette values)
+	 * and falls back to the saved hex/CSS value when no preset is set.
 	 *
 	 * @param array  $attributes Block attributes.
-	 * @param string $preset_key Top-level preset attribute key (e.g. "textColor").
-	 * @param string $style_key  Key under style.color (e.g. "text").
+	 * @param string $preset_key Preset slug attribute key (e.g. "iconColor").
+	 * @param string $value_key  Resolved CSS value attribute key (e.g. "iconColorValue").
 	 * @return string|null CSS color value or null.
 	 */
-	private static function resolve_color( array $attributes, string $preset_key, string $style_key ): ?string {
+	private static function resolve_color( array $attributes, string $preset_key, string $value_key ): ?string {
 		if ( ! empty( $attributes[ $preset_key ] ) && is_string( $attributes[ $preset_key ] ) ) {
 			return sprintf( 'var(--wp--preset--color--%s)', $attributes[ $preset_key ] );
 		}
-		$custom = $attributes['style']['color'][ $style_key ] ?? null;
-		if ( ! empty( $custom ) && is_string( $custom ) ) {
-			return self::preset_to_css( $custom );
+		if ( ! empty( $attributes[ $value_key ] ) && is_string( $attributes[ $value_key ] ) ) {
+			return $attributes[ $value_key ];
 		}
 		return null;
 	}
@@ -281,7 +260,6 @@ final class Author_Profile_Social_Block {
 	 * Build wrapper inline style with CSS variables for icon sizing and color.
 	 * Margin is handled natively by get_block_wrapper_attributes().
 	 * Gap is handled by WP layout support (outputs scoped <style> tag per block).
-	 * Color classes/inline styles are skipped via __experimentalSkipSerialization in block.json.
 	 *
 	 * @param array $attributes Block attributes.
 	 * @param int   $icon_size  Icon size in pixels.
@@ -292,8 +270,8 @@ final class Author_Profile_Social_Block {
 		$is_brand = ! empty( $attributes['className'] ) && str_contains( $attributes['className'], 'is-style-brand' );
 
 		if ( ! $is_brand ) {
-			$icon_color      = self::resolve_color( $attributes, 'textColor', 'text' );
-			$icon_background = self::resolve_color( $attributes, 'backgroundColor', 'background' );
+			$icon_color      = self::resolve_color( $attributes, 'iconColor', 'iconColorValue' );
+			$icon_background = self::resolve_color( $attributes, 'iconBackgroundColor', 'iconBackgroundColorValue' );
 
 			if ( null !== $icon_color ) {
 				$parts[] = sprintf( '--icon-color: %s;', $icon_color );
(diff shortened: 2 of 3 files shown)


src/blocks/author-profile-social/edit.jsx
```diff
@@ -67,7 +68,12 @@ export default function Edit( { attributes, setAttributes, clientId } ) {
 	const iconBackground = ! isBrand ? resolveColor( backgroundColor, styleAttr?.color?.background ) : undefined;
 
 	// Hide color panel when "Brand" is active; rename labels when "Default".
+	// Only run while this block is selected — otherwise the global sidebar
+	// query leaks the relabel/hide into every other block's color panel.
 	useEffect( () => {
+		if ( ! isSelected ) {
+			return;
+		}
```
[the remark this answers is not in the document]
laurelfulford: I went a different way -- this first approach is way too fragile.

src/blocks/author-profile-social/edit.jsx
```diff
@@ -28,91 +38,63 @@ const fetchAllServiceKeys = () => {
...
 
-		updateColorPanel( sidebar );
-
-		const observer = new MutationObserver( () => updateColorPanel( sidebar ) );
-		observer.observe( sidebar, { childList: true, subtree: true } );
-
-		return () => observer.disconnect();
-	}, [ isBrand ] );
+	const resolvedIconColor = ! isBrand ? iconColor?.color || iconColorValue : undefined;
+	const resolvedIconBackground = ! isBrand ? iconBackgroundColor?.color || iconBackgroundColorValue : undefined;
```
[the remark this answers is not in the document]
laurelfulford: This is okay - it's a new block, not on live sites, so we don't need to worry (yet) about preserving existing settings.

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -221,58 +221,37 @@ private static function render_social_flat( array $attributes, WP_Block $block,
...
 	 */
 	private static function get_block_wrapper_attributes( WP_Block $block, array $attributes, int $icon_size ): string {
-		$previous = \WP_Block_Supports::$block_to_render ?? null;
-		\WP_Block_Supports::$block_to_render = $block->parsed_block;
-
-		$wrapper_attributes = get_block_wrapper_attributes(
+		return get_block_wrapper_attributes(
 			[
 				'class' => 'author-profile-social__list',
 				'style' => self::get_wrapper_style( $attributes, $icon_size ),
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [dc07ebf](https://github.com/Automattic/newspack-plugin/pull/4703/commits/dc07ebf9d27306c602150432d6543a02810ab7e8).

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -292,8 +270,8 @@ private static function get_wrapper_style( array $attributes, int $icon_size ):
 		$is_brand = ! empty( $attributes['className'] ) && str_contains( $attributes['className'], 'is-style-brand' );
 
 		if ( ! $is_brand ) {
-			$icon_color      = self::resolve_color( $attributes, 'textColor', 'text' );
-			$icon_background = self::resolve_color( $attributes, 'backgroundColor', 'background' );
+			$icon_color      = self::resolve_color( $attributes, 'iconColor', 'iconColorValue' );
+			$icon_background = self::resolve_color( $attributes, 'iconBackgroundColor', 'iconBackgroundColorValue' );
```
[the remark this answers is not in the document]
laurelfulford: This is fine - again, it's an in-development block, so we don't need to preserve existing block settings.

src/blocks/author-profile-social/block.json
```diff
@@ -19,20 +19,31 @@
 		"iconSize": {
```
adekbadek, line 19: **Blocker:** Removing `color.text` / `color.background` supports without a `deprecated` registration is a silent-data-loss path for the ~30 days of posts (production since v6.37.0 / 2026-04-13) that have been saving `textColor` / `backgroundColor` / `style.color.text` / `style.color.background` under the previous schema. Because `save: () => <InnerBlocks.Content />` returns no attribute markup, block validation won't fail loudly — the old keys are dropped from parsed attributes on first load, and on the frontend the new PHP `resolve_color()` never reads them, so the configured colours disappear silently and the data is gone on next resave. Verified in-env by saving a post with the old shape and loading the frontend: no `--icon-color` / `--icon-background` set despite the saved attrs.

Needs a `deprecated` entry registering the previous attribute shape with a `migrate(attrs)` mapping `textColor → iconColor`, `style.color.text → iconColorValue`, `backgroundColor → iconBackgroundColor`, `style.color.background → iconBackgroundColorValue` (and stripping the empty `style.color`).

src/blocks/author-profile-social/edit.jsx
```diff
@@ -28,91 +38,63 @@ const fetchAllServiceKeys = () => {
...
+		},
+		{
+			colorValue: iconBackgroundColor?.color || iconBackgroundColorValue,
+			onColorChange: value => {
+				setIconBackgroundColor( value );
+				setAttributes( { iconBackgroundColorValue: value } );
+			},
+			label: __( 'Icon background', 'newspack-plugin' ),
+		},
+	];
```
adekbadek, line 90: **Suggestion:** Worth manually confirming that picking a theme **preset** colour writes to `iconColor` (the slug attribute), not just `iconColorValue`. `ColorGradientSettingsDropdown`'s `onColorChange` passes a single value, and `withColors`' setter is responsible for resolving it back to a slug by iterating the palette — Core's `core/social-links` uses the same pattern, but if for any reason `iconColor` never gets populated, the docblock claim in `class-author-profile-social-block.php` ("Prefers the preset slug so theme switches reflect new palette values") doesn't hold and the frontend always emits a raw hex.

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -246,33 +248,21 @@ private static function get_block_wrapper_attributes( WP_Block $block, array $at
...
 	 * @param array  $attributes Block attributes.
-	 * @param string $preset_key Top-level preset attribute key (e.g. "textColor").
-	 * @param string $style_key  Key under style.color (e.g. "text").
+	 * @param string $preset_key Preset slug attribute key (e.g. "iconColor").
+	 * @param string $value_key  Resolved CSS value attribute key (e.g. "iconColorValue").
 	 * @return string|null CSS color value or null.
 	 */
-	private static function resolve_color( array $attributes, string $preset_key, string $style_key ): ?string {
+	private static function resolve_color( array $attributes, string $preset_key, string $value_key ): ?string {
 		if ( ! empty( $attributes[ $preset_key ] ) && is_string( $attributes[ $preset_key ] ) ) {
```
adekbadek: **Suggestion:** `resolve_color()` returns `var(--wp--preset--color--<slug>)` whenever the slug is set, even if the active theme doesn't define that slug — the resulting CSS variable is unresolved and the icon renders uncoloured on the frontend. The editor side (via `withColors`'s `getColorObjectByAttributeValues`) falls back to `iconColorValue` in that case, so editor preview and frontend can diverge on theme switch.

Either make the fallback symmetric (prefer `iconColorValue` when present, regardless of slug), or drop the "theme switches reflect new palette values" claim from the docblock.

src/blocks/author-profile-social/class-author-profile-social-block.php
```diff
@@ -221,16 +221,18 @@ private static function render_social_flat( array $attributes, WP_Block $block,
...
 	 *
 	 * @param WP_Block $block      Block instance.
 	 * @param array    $attributes Block attributes.
 	 * @param int      $icon_size  Icon size in pixels.
 	 * @return string HTML attributes for the wrapper element.
 	 */
 	private static function get_block_wrapper_attributes( WP_Block $block, array $attributes, int $icon_size ): string {
-		$previous = \WP_Block_Supports::$block_to_render ?? null;
+		// Inner blocks have already rendered by this point in the InnerBlocks
+		// path, leaving $block_to_render pointing at the last child — so the
```
adekbadek: **Nit:** Good that the rationale is documented now. The actual reason for the snapshot is more defensive than "the inner blocks have already rendered": each inner `WP_Block::render()` snapshots/restores `$block_to_render` around its own callback, so this should already be back to the parent's `parsed_block` by the time `get_block_wrapper_attributes()` is called. The snapshot here guards against a third-party filter or future Core change breaking that chain. Worth rewording the comment to reflect "defensive against the inner-render chain leaking the static" rather than "already rendered".

adekbadek (changes requested):
The bug fix itself is well-targeted and works end-to-end in the Site Editor — Article Subtitle and other unrelated blocks now show plain *Text*/*Background* labels, and the new in-block colour panel works as described, including the Brand-style hide/show.

One blocker before merge: removing the `color` supports from `block.json` without a `deprecated` registration is a silent-data-loss path for posts saved under the current shipped version. 

A handful of polish items below — see the inline comments.

src/blocks/author-profile-social/block.json
laurelfulford: @adekbadek I previously poo-poo'd this when Copilot suggested it because this version of the Author Profile block (and this issue) won't show up unless you're using a block theme. It is _possible_ it's used with either our block theme or another theme, but the odds seem very low.

I think we should be safe without a fallback, but I'm willing to talk it out further! I might be being too laissez-faire about it 😅

src/blocks/author-profile-social/edit.jsx
laurelfulford: Good call! Looks like it does -- I got both `base` and `accent` when using random colours from the theme's baked in palette: 

```
 <!-- wp:newspack/author-profile-social {"iconColor":"base","iconColorValue":"#ffffff","iconBackgroundColor":"accent","iconBackgroundColorValue":"#003DA5","className":"is-style-default","style":{"spacing":{"padding":{"top":"var:preset|spacing|20"}}}}    
  -->       
```

src/blocks/author-profile-social/class-author-profile-social-block.php
laurelfulford: Good call! I ended up switching to `var(--wp--preset--color--<slug>, <hex>)` when both are present in [d2cd993](https://github.com/Automattic/newspack-plugin/pull/4703/commits/d2cd993493e8360f056db40e59ac416c28b350ae), so if the variable isn't defined in a theme the hex will be used instead. Also updated the docblock to reflect that!

src/blocks/author-profile-social/class-author-profile-social-block.php
laurelfulford: Code comment beefed up in [35e8c87](https://github.com/Automattic/newspack-plugin/pull/4703/commits/35e8c87a1030cb382902b8f9a5e860b7706b93cf) - thanks!

laurelfulford:
I think this is ready for your eyes again, @adekbadek, though it's just a comment addressing the blocker, not a code change.

If you have any hesitations about my "it'll be fiiiiiiiine" response there, let me know - there's no harm adding the deprecated code, I just wanted to avoid adding complexity if we can, but that shouldn't be at the expense of UX.

src/blocks/author-profile-social/block.json
adekbadek: Ok, that's fine then 👌

adekbadek (approved):

laurelfulford:
Thanks @adekbadek!

PR #4706, opened 2026-05-06
fix(subscriptions): allow migrated subs to resubscribe

miguelpeixe:
### All Submissions:

Migrated subscriptions do not have previous payments on record, which prevents resubscription. That restriction is applied on the Woo Subscriptions side to prevent circumventing sign-up fees.

We need to filter into `wcs_can_user_resubscribe_to` in order to allow resubscription of migrated subs.

1. In a fresh session, purchase a subscription with a new reader account
2. In the admin, edit the subscription and change its status to expired
3. Back as the reader, visit the subscription page in "My Account" and confirm there's no regression, and you can resubscribe without issues
4. Now, as an admin, create a subscription manually, assigning it to the reader and to a subscription product
5. Change the manually created subscription status to "expired."
6. As the reader, confirm you are not able to resubscribe (there are no previous payments)
7. Now, via shell, add a stripe subscription ID meta to simulate a migration:

```php
$sub = wcs_get_subscription( {sub_id} );
$sub->add_meta_data( '_stripe_subscription_id', 1234 );
$sub->save();
```

8. Refresh the subscription page in My Account and confirm you are now able to resubscribe

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
+++ b/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
@@ -22,6 +22,7 @@ class WooCommerce_Subscriptions {
 		add_filter( 'woocommerce_subscriptions_product_trial_length', [ __CLASS__, 'limit_free_trials_to_one_per_user' ], 10, 2 );
 		add_filter( 'wcs_get_users_subscriptions', [ __CLASS__, 'filter_subscriptions_for_account_page' ], 10, 1 );
 		add_filter( 'woocommerce_subscriptions_can_item_be_switched', [ __CLASS__, 'allow_migrated_subscription_switch' ], 10, 3 );
+		add_filter( 'wcs_can_user_resubscribe_to', [ __CLASS__, 'allow_migrated_subscription_to_resubscribe' ], 10, 3 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -344,5 +345,47 @@ public static function get_subscription_product_id( $subscription ) {
...
+			empty( $subscription ) ||
+			! user_can( $user_id, 'subscribe_again', $subscription->get_id() ) || // phpcs:ignore WordPress.WP.Capabilities.Unknown
+			! $subscription->has_status( [ 'pending-cancel', 'cancelled', 'expired', 'trash' ] ) ||
+			$subscription->get_total() <= 0 ||
+			$subscription->contains_unavailable_product()
+		) {
+			return false;
+		}
+
+		$migrated_meta = [ '_piano_subscription_id', '_stripe_subscription_id' ];
```
chickenn00dle, line 381: My concern here is as we get migrations from more and more sources this will quickly become outdated.

Maybe we can use the `created_via` meta since this will be added to every migrated subscription regardless of source: https://github.com/woocommerce/woocommerce-subscriptions-importer-exporter/blob/e3dc275038cb02e6382b3377ac1719e556c0c8f5/includes/class-wcs-importer.php#L389

WDYT?

chickenn00dle (approved):
Works wonderfully @miguelpeixe!

I'm okay to merge as is, but I left a comment regarding the meta we are checking.

includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
miguelpeixe: I've never seen that. We don't use `WCS_Importer` in https://github.com/automattic/newspack-subscription-migrations, so I don't think we'll have that flag.

But looking closer, we do [have `newspack-stripe` for the Stripe importer](https://github.com/Automattic/newspack-subscription-migrations/blob/8107b913874652309d620a21a25a4d3d390e929b/includes/class-woocommerce-connection.php#L17) but nothing for Piano.

I'd rather keep it as is for now.

chickenn00dle: > I've never seen that. We don't use WCS_Importer in https://github.com/automattic/newspack-subscription-migrations

Right, but we use it as part of our migration process. We use `newspack-subscription-migrations` to generate the import file that is then run by `woocommerce-subscriptions-importer-exporter` to actually create the subscriptions.

In any case, I'm fine with merging as is!

miguelpeixe: Ah, I see, thank you for clarifying. So would that `newspack-stripe` replace the original `created_via`?

chickenn00dle: Honestly, its my first time seeing this class 😄. But it looks like its only used as part of the `newspack stripe sync-stripe-subscriptions-to-wc` command. Not sure if we even use this in any of our migrations anymore, but if we do then it looks like it would replace it.

PR #4714, opened 2026-05-11
fix(stripe): prevent re-stamping `_stripe_customer_id` on non-Stripe subscriptions

dkoo:
### All Submissions:

On dual-gateway sites (Stripe plugin + WooPayments both active), the upstream Stripe plugin unconditionally re-stamps `_stripe_customer_id` onto subscriptions during renewal, even when the subscription's payment method is `woocommerce_payments`. When the Stripe customer no longer exists in Stripe, this causes "No such customer" errors and failed renewals.

This adds three Newspack-side guards to `WooCommerce_Gateway_Stripe`:

- **`update_post_metadata` filter** (`maybe_block_stripe_customer_id_post_meta_update`) — blocks raw `update_post_meta( $id, '_stripe_customer_id', ... )` calls from the Stripe plugin's legacy path in `abstract-wc-stripe-payment-gateway.php`. Uses `wcs_is_subscription()` (not `get_post_type()`) to correctly identify subscription IDs on HPOS sites where data sync is disabled and subscriptions are not stored in `wp_posts`.

- **`woocommerce_before_subscription_object_save` action** (`maybe_strip_stripe_customer_id_before_save`) — strips `_stripe_customer_id` from in-memory meta before WC CRUD commits, covering three `$subscription->update_meta_data() + save()` paths in the Stripe plugin's subscription trait. Hook name is derived from `WC_Subscription::$object_type = 'subscription'`, fired by `WC_Abstract_Order::save()`.

- **`wcs_renewal_order_created` action** (`clear_stripe_customer_id_on_renewal`) — fires after the renewal order is created but **before payment processing begins**. Clears `_stripe_customer_id` from both the renewal order and its parent subscription. On HPOS sites with sync disabled, meta is purged from both the HPOS store (WC CRUD `save()`) and `wp_postmeta` (direct `delete_post_meta()`) to ensure all read paths see a clean slate before WooPayments processes the payment.

All guards are no-ops for subscriptions using any Stripe gateway variant (`stripe`, `stripe_sepa`, `stripe_klarna`, etc.) and for non-subscription objects. An empty payment method is intentionally treated as non-Stripe.

**Known scope boundary:** The `update_post_metadata` filter and pre-save action target subscription objects only. If a third-party plugin were to write `_stripe_customer_id` directly to a renewal order ID (not a subscription ID), those two guards would not intercept it. The `wcs_renewal_order_created` hook covers initial order creation; all four documented Stripe plugin re-stamp paths operate on subscription objects, so this boundary is acceptable.

Closes NPPM-2761.

Prerequisites: a site with both the WooCommerce Stripe plugin and WooPayments active (dual-gateway), and WooCommerce Subscriptions.

1. As a reader, purchase a subscription using the Stripe payment gateway (not WooPayments).
2. As an admin, reconnect that gateway to a different Stripe account so that the customer ID won't exist in Stripe anymore.
3. On `release`: as the reader, go to My Account > Payment Information and attempt to add a new payment method via the WooPayments gateway. Observe a failure with error response: `No such customer`.
4. Check out this branch, repeat step 3, and confirm that adding the new payment method succeeds.
5. As an admin, trigger a renewal for the subscription ("Process renewal order") and confirm it succeeds.
6. Run the PHP test suite: `n test-php` — expect 14 tests passing in the `WooCommerce_Gateway_Stripe` group with no failures.

### Other information:

**Upstream bug:** [woocommerce/woocommerce-gateway-stripe#2661](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2661) (filed Jul 2023, unresolved)

**Note:** On HPOS-enabled sites with data sync off, manual cleanup of stale `_stripe_customer_id` meta also requires deleting the legacy `wp_postmeta` row directly. The `wcs_renewal_order_created` guard handles this automatically for new renewals; existing stale subscriptions are cleaned at their next renewal attempt.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,6 @@
 	},
 	"require": {
 		"composer/installers": "~2.0",
-		"joshtronic/php-loremipsum": "^2.1",
 		"google/auth": "^1.15",
 		"woocommerce/action-scheduler": "^3.9"
 	},
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -245,20 +245,24 @@ class WooCommerce_Gateway_Stripe {
 	 * Clear stale _stripe_customer_id from a renewal order and its parent
 	 * subscription when the subscription uses a non-Stripe payment method.
 	 *
+	 * Hooked on wcs_renewal_order_created, which is an apply_filters hook in WCS.
 	 * Fires after the renewal order is created but before payment processing
 	 * begins, intercepting the window where WooPayments reads _stripe_customer_id
 	 * from the renewal order and would encounter an invalid Stripe customer.
 	 *
-	 * On HPOS sites with data sync disabled, meta is purged from both the HPOS
-	 * store (via WC CRUD save()) and wp_postmeta (via delete_post_meta()) to
-	 * ensure all read paths see a clean state.
+	 * On HPOS sites with data sync disabled, the renewal order is purged from both
+	 * the HPOS store (via WC CRUD save()) and wp_postmeta (via delete_post_meta()).
+	 * The subscription's HPOS store is cleaned at its next natural save via the
+	 * woocommerce_before_subscription_object_save guard; we do not save it here to
+	 * avoid triggering WCS side-effects mid-filter.
 	 *
 	 * @param \WC_Order        $renewal_order The renewal order just created.
 	 * @param \WC_Subscription $subscription  The parent subscription.
+	 * @return \WC_Order The renewal order, passed through for the filter chain.
 	 */
 	public static function clear_stripe_customer_id_on_renewal( $renewal_order, $subscription ) {
 		if ( str_starts_with( $subscription->get_payment_method(), 'stripe' ) ) {
-			return;
+			return $renewal_order;
 		}
 
 		// Clear from the renewal order — this is what WooPayments reads during payment processing.
@@ -268,15 +272,17 @@ class WooCommerce_Gateway_Stripe {
 			delete_post_meta( $renewal_order->get_id(), '_stripe_customer_id' );
 		}
 
-		// Also clean the subscription (root source of the stale value).
-		// Note: $subscription->save() re-fires woocommerce_before_subscription_object_save, which
-		// calls maybe_strip_stripe_customer_id_before_save() again — this is intentional and
-		// idempotent (delete_meta_data on an already-absent key is a no-op).
+		// Also clean the subscription's in-memory meta and wp_postmeta (root source of the stale value).
+		// We intentionally do NOT call $subscription->save() here: wcs_renewal_order_created is an
+		// apply_filters hook, and triggering a full WC CRUD save on the subscription mid-filter can
+		// cause unexpected side-effects. The woocommerce_before_subscription_object_save guard will
+		// strip _stripe_customer_id from HPOS when the subscription is next saved naturally by WCS.
 		if ( $subscription->get_meta( '_stripe_customer_id' ) ) {
 			$subscription->delete_meta_data( '_stripe_customer_id' );
-			$subscription->save();
 			delete_post_meta( $subscription->get_id(), '_stripe_customer_id' );
 		}
+
+		return $renewal_order;
 	}
 }
 WooCommerce_Gateway_Stripe::init();
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -33,7 +33,10 @@ class WooCommerce_Gateway_Stripe {
 		 *   1. update_post_metadata filter — intercepts raw update_post_meta() calls to
 		 *      wp_postmeta (the legacy and data-sync storage paths). Only registered when
 		 *      both WCS and the Stripe plugin are active to avoid hook overhead on sites
-		 *      where neither dependency is present.
+		 *      where neither dependency is present. The class_exists/function_exists check
+		 *      is deferred to plugins_loaded because WordPress loads plugins alphabetically:
+		 *      newspack-plugin loads before woocommerce-gateway-stripe and woocommerce-
+		 *      subscriptions, so the symbols are not yet defined at Newspack include time.
 		 *      NOTE: this filter fires only for wp_postmeta writes; it does NOT intercept
 		 *      HPOS-only writes to wc_orders_meta, which is why guard #2 is also required.
 		 *
@@ -45,11 +48,22 @@ class WooCommerce_Gateway_Stripe {
 		 *      renewal order (and the parent subscription in-memory) before payment
 		 *      processing begins, self-healing stale values that pre-date the fix.
 		 */
+		// Priority 20 ensures this runs after WC_Stripe and woocommerce-subscriptions
+		// have completed their own plugins_loaded callbacks (both register at default priority 10).
+		add_action( 'plugins_loaded', [ __CLASS__, 'maybe_register_post_meta_guard' ], 20 );
+		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
+		add_filter( 'wcs_renewal_order_created', [ __CLASS__, 'clear_stripe_customer_id_on_renewal' ], 10, 2 );
+	}
+
+	/**
+	 * Conditionally register the update_post_metadata filter once all plugins
+	 * have loaded their main files. See NPPM-2761 init() comment block for the
+	 * load-order rationale.
+	 */
+	public static function maybe_register_post_meta_guard() {
 		if ( class_exists( 'WC_Stripe' ) && function_exists( 'wcs_is_subscription' ) ) {
 			add_filter( 'update_post_metadata', [ __CLASS__, 'maybe_block_stripe_customer_id_post_meta_update' ], 10, 3 );
 		}
-		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
-		add_filter( 'wcs_renewal_order_created', [ __CLASS__, 'clear_stripe_customer_id_on_renewal' ], 10, 2 );
 	}
 
 	/**
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -196,6 +196,9 @@ class WooCommerce_Gateway_Stripe {
 	 * @return mixed Null to allow, true to block.
 	 */
 	public static function maybe_block_stripe_customer_id_post_meta_update( $check, $object_id, $meta_key ) {
+		if ( null !== $check ) {
+			return $check; // Respect earlier filter short-circuits.
+		}
 		if ( '_stripe_customer_id' !== $meta_key ) {
 			return $check;
 		}
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -28,6 +28,7 @@ class WooCommerce_Gateway_Stripe {
 		// subscriptions whose payment method is not a Stripe gateway.
 		add_filter( 'update_post_metadata', [ __CLASS__, 'maybe_block_stripe_customer_id_post_meta_update' ], 10, 3 );
 		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
+		add_action( 'wcs_renewal_order_created', [ __CLASS__, 'clear_stripe_customer_id_on_renewal' ], 10, 2 );
 	}
 
 	/**
@@ -233,6 +234,41 @@ class WooCommerce_Gateway_Stripe {
 			return;
 		}
 		$subscription->delete_meta_data( '_stripe_customer_id' );
+	}
+
+	/**
+	 * Clear stale _stripe_customer_id from a renewal order and its parent
+	 * subscription when the subscription uses a non-Stripe payment method.
+	 *
+	 * Fires after the renewal order is created but before payment processing
+	 * begins, intercepting the window where WooPayments reads _stripe_customer_id
+	 * from the renewal order and would encounter an invalid Stripe customer.
+	 *
+	 * On HPOS sites with data sync disabled, meta is purged from both the HPOS
+	 * store (via WC CRUD save()) and wp_postmeta (via delete_post_meta()) to
+	 * ensure all read paths see a clean state.
+	 *
+	 * @param \WC_Order        $renewal_order The renewal order just created.
+	 * @param \WC_Subscription $subscription  The parent subscription.
+	 */
+	public static function clear_stripe_customer_id_on_renewal( $renewal_order, $subscription ) {
+		if ( str_starts_with( $subscription->get_payment_method(), 'stripe' ) ) {
+			return;
+		}
+
+		// Clear from the renewal order — this is what WooPayments reads during payment processing.
+		if ( $renewal_order->get_meta( '_stripe_customer_id' ) ) {
+			$renewal_order->delete_meta_data( '_stripe_customer_id' );
+			$renewal_order->save();
+			delete_post_meta( $renewal_order->get_id(), '_stripe_customer_id' );
+		}
+
+		// Also clean the subscription (root source of the stale value).
+		if ( $subscription->get_meta( '_stripe_customer_id' ) ) {
+			$subscription->delete_meta_data( '_stripe_customer_id' );
+			$subscription->save();
+			delete_post_meta( $subscription->get_id(), '_stripe_customer_id' );
+		}
 	}
 }
 WooCommerce_Gateway_Stripe::init();
--- a/includes/plugins/class-woocommerce-gateway-stripe.php
+++ b/includes/plugins/class-woocommerce-gateway-stripe.php
@@ -23,6 +23,11 @@ class WooCommerce_Gateway_Stripe {
 
 		add_filter( 'wc_stripe_generate_payment_request', [ __CLASS__, 'add_payment_request_metadata' ], 10, 2 );
 		add_filter( 'wc_stripe_intent_metadata', [ __CLASS__, 'add_intent_metadata' ], 10, 2 );
+
+		// NPPM-2761: Prevent Stripe plugin from re-stamping _stripe_customer_id on
+		// subscriptions whose payment method is not a Stripe gateway.
+		add_filter( 'update_post_metadata', [ __CLASS__, 'maybe_block_stripe_customer_id_post_meta_update' ], 10, 3 );
+		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +178,58 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+	 * @return mixed Null to allow, true to block.
+	 */
+	public static function maybe_block_stripe_customer_id_post_meta_update( $check, $object_id, $meta_key ) {
+		if ( '_stripe_customer_id' !== $meta_key ) {
+			return $check;
+		}
+		if ( ! function_exists( 'wcs_is_subscription' ) || ! function_exists( 'wcs_get_subscription' ) || ! \wcs_is_subscription( $object_id ) ) {
+			return $check;
+		}
+		$subscription = \wcs_get_subscription( $object_id );
```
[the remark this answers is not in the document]
dkoo: Addressed in e2f9345.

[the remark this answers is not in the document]
dkoo: Addressed in e2f9345.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+	 *
+	 * Returning a non-null value from the update_{meta_type}_metadata filter
+	 * short-circuits the write.
+	 *
+	 * @param mixed  $check     The value to return. Null allows the write; true blocks it.
+	 * @param int    $object_id Post ID being written to.
+	 * @param string $meta_key  Meta key being written.
+	 * @return mixed Null to allow, true to block.
+	 */
+	public static function maybe_block_stripe_customer_id_post_meta_update( $check, $object_id, $meta_key ) {
```
adekbadek, line 256: **Blocker:** `update_post_metadata` filter is registered unconditionally — fires on every postmeta write site-wide — The filter is registered on every page load regardless of whether WCS or the Stripe plugin is even active, so every `update_post_meta()` call across the entire site pays the cost of the hook invocation + key comparison. Two improvements: (a) gate the `add_filter()` registration behind `function_exists('wcs_is_subscription') && class_exists('WC_Stripe')` so it isn't registered when both aren't present; (b) make the boundary with HPOS-only meta explicit in the comment block — `update_post_metadata` does NOT fire on HPOS-only writes to `wc_orders_meta`, which is the entire reason guard #2 (`woocommerce_before_subscription_object_save`) exists. The current docstring at L184 doesn't call this out clearly.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+			return $check;
+		}
+		// Use wcs_is_subscription() rather than get_post_type() here: on HPOS sites with
+		// data sync disabled, subscriptions are not stored in wp_posts and get_post_type()
+		// would return false, silently bypassing this guard.
+		if ( ! function_exists( 'wcs_is_subscription' ) || ! function_exists( 'wcs_get_subscription' ) || ! \wcs_is_subscription( $object_id ) ) {
+			return $check;
+		}
+		$subscription = \wcs_get_subscription( $object_id );
+		if ( $subscription && ! str_starts_with( $subscription->get_payment_method(), 'stripe' ) ) {
```
adekbadek: **Blocker:** `str_starts_with( $payment_method, 'stripe' )` is too loose a predicate for the load-bearing 'is Stripe?' check — The 'is Stripe?' predicate is used in all three guards (L213, L238, L264) and is a prefix match on the gateway id. The chosen polarity ('preserve when stripe-prefixed, strip otherwise') generalises beyond the documented failure case in NPPM-2761, which is specifically `_payment_method = woocommerce_payments`. The Linear ticket's stated outcome reads *'When a subscription's `_payment_method` is not `stripe`'* — singular, not a prefix family — and the broad-vs-narrow choice is not engaged anywhere in the thread; the broader form appears to have been carried forward from dkoo's initial sketch without explicit comparison.

Broad polarity opens two avenues of false-positive harm:

- A future or third-party gateway whose id starts with `stripe` but isn't actually Stripe-billed would be silently treated as Stripe and retain stale `_stripe_customer_id`.
- A Stripe-billed integration with a non-`stripe`-prefix gateway id (custom shim, rebrand) would have its valid `_stripe_customer_id` aggressively stripped on every renewal.

The empty-`payment_method` edge (free-trial / state-transition subscriptions, which WCS interprets as the manual/no-gateway signal) is currently treated as 'strip' too. Worth weighing whether 'preserve' is the right stance for those.

**Risk if not addressed:** On a site where the user-meta fallback is also empty (guest checkout, deleted user, account-merge fallout), the stripped `_stripe_customer_id` cannot be recovered server-side and the renewal constructs a Payment_Information with an empty customer; WooPayments creates a NEW Stripe customer for the renewal, decoupling the recurring payment from the original card-on-file and potentially failing the charge until the publisher manually re-links the customer.

**Risk if not addressed:** A bulk script or admin operation that touches subscriptions mid-edit (e.g. the payment-method-change UI saving the subscription before the new method is committed) strips every touched subscription's customer id; recovery requires Stripe-side customer lookup per record.

**Consider:** Flip the polarity to target the bug case directly: `$subscription->get_payment_method() === 'woocommerce_payments'`. This narrows the action to exactly the case NPPM-2761 documents and eliminates the false-positive class entirely (custom shims, rebrands, empty `payment_method`). One open question worth verifying first: whether WooPayments uses a single canonical id (`woocommerce_payments`) or splits into `woocommerce_payments_*` sibling ids for variants (Card, SEPA Debit, Klarna, Afterpay, BNPL). If split, use `str_starts_with('woocommerce_payments')` on the symmetric polarity — same prefix structure as today, but the prefix-collision risk on the WooPayments side is essentially zero.

composer.json
```diff
@@ -9,7 +9,6 @@
 	},
 	"require": {
 		"composer/installers": "~2.0",
```
adekbadek, line 11: **Suggestion:** Unrelated dependency removal bundled with the Stripe fix — The `joshtronic/php-loremipsum` removal plus inlined `LIPSUM_WORDS` constant in `class-starter-content-generated.php` is functionally unrelated to the Stripe re-stamping fix. The PR notes it was rolled in from a previously merged PR, but it still inflates the diff and complicates a clean revert if the Stripe guards need to be backed out. Future similar work: split unrelated dependency hygiene into its own PR so each change can be reverted independently.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -23,6 +23,12 @@ public static function init() {
 
 		add_filter( 'wc_stripe_generate_payment_request', [ __CLASS__, 'add_payment_request_metadata' ], 10, 2 );
 		add_filter( 'wc_stripe_intent_metadata', [ __CLASS__, 'add_intent_metadata' ], 10, 2 );
+
+		// NPPM-2761: Prevent Stripe plugin from re-stamping _stripe_customer_id on
+		// subscriptions whose payment method is not a Stripe gateway.
+		add_filter( 'update_post_metadata', [ __CLASS__, 'maybe_block_stripe_customer_id_post_meta_update' ], 10, 3 );
+		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
+		add_action( 'wcs_renewal_order_created', [ __CLASS__, 'clear_stripe_customer_id_on_renewal' ], 10, 2 );
```
adekbadek: **Suggestion:** Register `wcs_renewal_order_created` with `add_filter`, not `add_action` — The hook is dispatched via `apply_filters` in WCS and the callback intentionally returns `$renewal_order` (per commit `d35aad16b` which fixed a fatal from losing the return value). `add_action` is a thin alias around `add_filter` so this works today, but a future maintainer reading `add_action` may reasonably drop the `return` statement, re-introducing the fatal. Register with `add_filter` so the registration site reflects the documented contract.

includes/plugins/class-woocommerce-gateway-stripe.php
adekbadek: **Suggestion:** Cast `get_payment_method()` to string before `str_starts_with` (PHP 8.1+ null deprecation) — `str_starts_with()` with a `null` haystack triggers a deprecation notice on PHP 8.1+ and errors on PHP 9. `WC_Order::get_payment_method()` is documented to return a string, but in-memory state on a freshly-built subscription or one with corrupted data can be null. Cast for safety: `str_starts_with( (string) $subscription->get_payment_method(), 'stripe' )`. Same pattern at L238 and L264.

adekbadek: **Suggestion:** Extract the 'is Stripe?' predicate to a private helper — it's duplicated across all three guards — `str_starts_with( $subscription->get_payment_method(), 'stripe' )` appears at L213, L238, and L264. Even small drift across the three sites is a footgun: a future change to the polarity (see the blocker above), the gateway-id family, or even just an added variant id has to touch all three or one guard silently behaves differently from the others. Extract to a private helper like `is_woopayments_subscription( \WC_Subscription $subscription ): bool` so the decision lives in one place. Combined with the polarity flip suggested above, this becomes especially load-bearing — the right polarity in one helper is hard to get wrong; the same polarity transcribed three times is not.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+	public static function clear_stripe_customer_id_on_renewal( $renewal_order, $subscription ) {
+		if ( str_starts_with( $subscription->get_payment_method(), 'stripe' ) ) {
+			return $renewal_order;
+		}
+
+		// Clear from the renewal order — this is what WooPayments reads during payment processing.
+		if ( $renewal_order->get_meta( '_stripe_customer_id' ) ) {
+			$renewal_order->delete_meta_data( '_stripe_customer_id' );
+			$renewal_order->save();
+			delete_post_meta( $renewal_order->get_id(), '_stripe_customer_id' );
```
adekbadek, line 329: **Suggestion:** HPOS data-sync semantics for `delete_post_meta()` deserve a clarifying comment — On HPOS-no-sync sites, `delete_post_meta()` is a no-op against the canonical row (which lives in `wc_orders_meta`). `$renewal_order->save()` at L271 writes via CRUD to whichever store is authoritative. The `delete_post_meta()` call exists to scrub legacy CPT and sync-table rows. The current docblock at L253-254 says the renewal order is purged from both stores, which is slightly misleading — clarify that `delete_post_meta()` is a belt-and-braces scrub for CPT/sync rows and is intentionally a no-op on HPOS-only sites.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+		}
+
+		// Also clean the subscription's in-memory meta and wp_postmeta (root source of the stale value).
+		// We intentionally do NOT call $subscription->save() here: wcs_renewal_order_created is an
+		// apply_filters hook, and triggering a full WC CRUD save on the subscription mid-filter can
+		// cause unexpected side-effects. The woocommerce_before_subscription_object_save guard will
+		// strip _stripe_customer_id from HPOS when the subscription is next saved naturally by WCS.
+		if ( $subscription->get_meta( '_stripe_customer_id' ) ) {
```
adekbadek: **Suggestion:** Subscription cleanup on renewal leaves the HPOS row stale on no-sync sites — The renewal handler deliberately skips `$subscription->save()` after stripping `_stripe_customer_id` (L280-283), citing apply_filters re-entry concerns, and relies on `woocommerce_before_subscription_object_save` to clean HPOS at the next natural save. On HPOS-no-sync sites this leaves a window where `$subscription->get_meta('_stripe_customer_id')` from a freshly-loaded subscription still returns the stale value, because `delete_post_meta()` only touched `wp_postmeta`, which is not authoritative. If any code path reads `_stripe_customer_id` off the parent subscription mid-renewal it would still see the stale id. Either call `save()` and accept the side-effects, or add a comment explicitly naming this stale window.

tests/unit-tests/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -0,0 +1,375 @@
...
+		);
+
+		WooCommerce_Gateway_Stripe::maybe_strip_stripe_customer_id_before_save( $subscription );
+
+		$this->assertSame(
+			'',
+			$subscription->get_meta( '_stripe_customer_id' ),
+			'_stripe_customer_id should be stripped when payment_method is empty (treated as non-Stripe).'
+		);
+	}
```
adekbadek, line 284: **Suggestion:** Missing coverage: empty `payment_method` on the post-meta filter and renewal handler — `test_stripe_customer_id_stripped_for_empty_payment_method` covers empty-payment-method behavior on the `maybe_strip_stripe_customer_id_before_save` path, but the same edge case is not exercised against `maybe_block_stripe_customer_id_post_meta_update` or `clear_stripe_customer_id_on_renewal`. Given that empty-payment-method-as-non-Stripe is explicitly intentional, all three guards should be pinned by tests so a future refactor doesn't quietly diverge.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+		if ( null !== $check ) {
+			return $check; // Respect earlier filter short-circuits.
+		}
+		if ( '_stripe_customer_id' !== $meta_key ) {
+			return $check;
+		}
+		// Use wcs_is_subscription() rather than get_post_type() here: on HPOS sites with
+		// data sync disabled, subscriptions are not stored in wp_posts and get_post_type()
+		// would return false, silently bypassing this guard.
+		if ( ! function_exists( 'wcs_is_subscription' ) || ! function_exists( 'wcs_get_subscription' ) || ! \wcs_is_subscription( $object_id ) ) {
```
adekbadek: **Nit:** Redundant `function_exists( 'wcs_get_subscription' )` check — `wcs_is_subscription` and `wcs_get_subscription` ship together in the WCS package. Either drop the second `function_exists` or move it inline immediately before the call on L212 with a defensive `null` fallback.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+	 * the HPOS store (via WC CRUD save()) and wp_postmeta (via delete_post_meta()).
+	 * The subscription's HPOS store is cleaned at its next natural save via the
+	 * woocommerce_before_subscription_object_save guard; we do not save it here to
+	 * avoid triggering WCS side-effects mid-filter.
+	 *
+	 * @param \WC_Order        $renewal_order The renewal order just created.
+	 * @param \WC_Subscription $subscription  The parent subscription.
+	 * @return \WC_Order The renewal order, passed through for the filter chain.
+	 */
+	public static function clear_stripe_customer_id_on_renewal( $renewal_order, $subscription ) {
```
adekbadek, line 320: **Nit:** `clear_stripe_customer_id_on_renewal` name understates the subscription cleanup — The callback name leads with the renewal-order cleanup, but the method also operates on the parent subscription at L280-283. Either rename to `clear_stale_stripe_customer_id_on_renewal` or expand the docblock summary to lead with both targets.

includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -173,5 +179,110 @@ public static function disable_link_by_default( $settings, $old_settings ) {
...
+		if ( str_starts_with( $subscription->get_payment_method(), 'stripe' ) ) {
+			return $renewal_order;
+		}
+
+		// Clear from the renewal order — this is what WooPayments reads during payment processing.
+		if ( $renewal_order->get_meta( '_stripe_customer_id' ) ) {
```
adekbadek: **Nit:** Use `meta_exists()` instead of truthy check on `get_meta()` return — `$renewal_order->get_meta( '_stripe_customer_id' )` returns `''` when the meta is absent (per WC_Data default). The truthy check works because `''` is falsy, but a valid Stripe customer id starting with `'0'` would also short-circuit. Safer pattern: `$renewal_order->meta_exists( '_stripe_customer_id' )`. Same on L280 for the subscription. Low risk in practice (real `cus_xxx` ids never evaluate falsy) but intent is clearer.

tests/mocks/wc-mocks.php
```diff
@@ -366,7 +375,22 @@ function wc_get_checkout_url() {
 	return 'https://example.com/checkout';
 }
 function wcs_is_subscription( $order ) {
```
adekbadek, line 383: **Nit:** Mocks cannot exercise the HPOS-no-sync divergence the production code is defending against — The mock `wcs_is_subscription` only recognises IDs registered via `wcs_create_subscription`. On a real HPOS-no-sync site, the production concern is the opposite: a subscription id exists in the HPOS orders table but NOT in `wp_posts`, and `get_post_type()` would return false. The mock has no notion of separate stores. The tests pass on logic, but cannot verify the inline comment at L206-208 that motivates `wcs_is_subscription`. Worth either an integration-style test against a real WC+WCS install or a 'verified on HPOS-no-sync env' note in the PR description.

adekbadek (changes requested):
Two things to settle on the predicate before merge — polarity (broad `str_starts_with('stripe')` vs narrow `'woocommerce_payments'`) and duplication across the three guards. Details inline.

Before merge: verify against ≥2 real publisher accounts including one HPOS-no-sync site. The failure mode is publisher-config-specific (gateway transitions, HPOS+sync state, stale meta lineage) and the mock-based tests can't exercise the HPOS divergence the production code is defending.

composer.json
dkoo: Unfortunately I needed to cherry-pick this commit from alpha because without it, CI will fail unconditionally and prevent the hotfix release jobs from being able to run.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: gated the `add_filter` registration behind `class_exists( 'WC_Stripe' ) && function_exists( 'wcs_is_subscription' )`; updated the docblock and the `init()` comment to explicitly call out that the filter covers only wp_postmeta writes and that guard #2 handles HPOS.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: flipped the polarity to `str_starts_with( (string) $subscription->get_payment_method(), 'woocommerce_payments' )`, covering the base gateway ID and all split UPE LPM variants (`woocommerce_payments_sepa` etc.). Extracted the predicate into a private `is_woopayments_subscription()` helper so all three guards share a single decision point. Empty payment_method now falls through to 'not WooPayments → leave alone', which is the correct behaviour under the flipped polarity.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: changed to `add_filter`.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: the `(string)` cast is inside the new `is_woopayments_subscription()` helper, applied in one place for all three guards.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: updated the docblock to clarify that `$renewal_order->save()` writes to the authoritative store, and that `delete_post_meta()` is a belt-and-braces scrub for legacy CPT / data-sync rows that is intentionally a no-op on HPOS-only sites.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: added an explicit 'stale-window caveat' comment naming the HPOS-only case and why it is acceptable (the only NPPM-2761 consumer reads from the renewal ORDER, which is fully cleaned above).

tests/unit-tests/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: added `test_post_meta_filter_allows_empty_payment_method_subscription` and `test_renewal_preserves_customer_id_for_empty_payment_method`; updated the existing empty-payment-method test to reflect the new polarity (preserved, not stripped).

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: removed both `function_exists` guards from inside the callback — both functions are guaranteed to exist because the filter is only registered when `function_exists( 'wcs_is_subscription' )` is true.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Addressed in ad6baa9: switched to `meta_exists()` on both the renewal order and subscription checks.

tests/mocks/wc-mocks.php
dkoo: Agreed — the unit test mocks cannot simulate the HPOS-only divergence. The inline comment at the registration site in `init()` now explicitly documents this boundary (filter covers wp_postmeta only; guard #2 covers HPOS). Will add a note to the PR description confirming the HPOS-no-sync path was verified manually on newbedfordlight.org.

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Along with flipping the polarity to check for WooPayments specifically instead of any "not Stripe" gateway, ad6baa9 adds an `is_woopayments_subscription()` method to consolidate all checks.

dkoo:
@adekbadek thanks as always for the thorough and clear review feedback. ad6baa9 should address all blockers and suggestions and most nits above. I left the blocker threads unresolved so you can confirm and resolve if you want to.

dkoo:
797b1ba addresses a load-order issue caught in a follow-up review: the original `class_exists('WC_Stripe') && function_exists('wcs_is_subscription')` gate ran at Newspack include time, before WordPress had loaded the other plugins (alphabetical: `newspack-plugin` < `woocommerce-gateway-stripe` < `woocommerce-subscriptions`), so the guard was silently never registered in production.

**Fix:** defer the gate to `plugins_loaded` at priority 20 (both Stripe and WCS register their own symbols on `plugins_loaded` at the default priority 10, so we need to run after them).

**Verified live** in an isolated env with all three plugins active:
- `update_post_metadata` filter is now registered with `Newspack\WooCommerce_Gateway_Stripe::maybe_block_stripe_customer_id_post_meta_update` at prio=10
- Functional test: `update_post_meta($woopay_sub_id, '_stripe_customer_id', 'cus_x')` → write blocked, stored value remains empty
- Functional test: `update_post_meta($stripe_sub_id, '_stripe_customer_id', 'cus_y')` → write passes through, value stored
- All 17 PHPUnit tests still pass

includes/plugins/class-woocommerce-gateway-stripe.php
dkoo: Updated in 797b1ba to address a load-order issue caught in a follow-up review. [Details in this follow-up comment](https://github.com/Automattic/newspack-plugin/pull/4714#issuecomment-4454009381).

adekbadek (approved):
LGTM. All blockers and suggestions addressed; the `is_woopayments_subscription()` helper is a clean consolidation and the polarity flip is the right call. Bonus catch in `797b1ba` — the load-order fix is a sharper guard than the version this review asked for.

Confirming and resolving the two blocker threads.

includes/plugins/class-woocommerce-gateway-stripe.php
adekbadek: Confirmed — gated registration + the `plugins_loaded` priority 20 deferral in 797b1ba is the right shape. Resolving.

includes/plugins/class-woocommerce-gateway-stripe.php
adekbadek: Confirmed — narrow `'woocommerce_payments'` predicate (covering UPE variants) via the shared helper is the right call. Resolving.

PR #4716, opened 2026-05-11
fix: prevent content-gate editor.scss styles from getting chunked into the common.css file

laurelfulford:
### All Submissions:

This one was weird! 

In https://github.com/Automattic/newspack-plugin/pull/4646, a new file `block-visibility.tsx` started importing `content-gate/editor.scss`. Newspack Plugin's Webpack has `minChunks: 2`; since this was the second file editor.css was imported into, it started getting included in the common.css, which caused the `.components-base-control` styles in that CSS file to bleed into more post types.

The `.components-base-control` style wasn't very specific, but this only started being an issue when the styles were imported into more editors.

This PR does two things:
1. It moves the styles block-visiblity.tsx needed into its own CSS file, so editor.scss isn't being imported twice, and isn't getting added to the common.scss.
2. It makes the `.components-base-control` styles more specific (adding content gate-specific CSS classes) since this issue was still happening when you edited a Content Gate. 

Closes NPPM-2812

1. On trunk, edit a post and click the publish date in the sidebar; note the appearance of the date picker. 
2. Apply this PR and run `npm run build`; refresh the post editor and confirm the date picker looks okay. 
3. Using Woo Memberships, set up a Content Gate and confirm the same date picker looks okay.
4. Set up access control (add `define( 'NEWSPACK_CONTENT_GATES', true );` to your wp-config.php, and set up a content gate).
5. Edit a post, and add a group block. Confirm the options in the 'Access Control' panel look okay.
6. Smoke test some of the other steps from https://github.com/Automattic/newspack-plugin/pull/4646 and confirm things look okay.

### Other information:

### DIFF
--- a/src/content-gate/editor/block-visibility.tsx
+++ b/src/content-gate/editor/block-visibility.tsx
@@ -22,7 +22,7 @@ import { __, sprintf } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import './editor.scss';
+import './block-visibility.scss';
 
 /**
  * Target block types that receive access control attributes.
--- a/src/content-gate/editor/editor.scss
+++ b/src/content-gate/editor/editor.scss
@@ -5,9 +5,8 @@
 	display: none;
 }
 
-.components-base-control {
+.newspack-content-gate-panel .components-base-control {
 	margin-bottom: 16px;
-	.components-datetime__time-wrapper &,
 	&:last-child {
 		margin-bottom: 0;
 	}
--- a/src/content-gate/editor/editor.scss
+++ b/src/content-gate/editor/editor.scss
@@ -8,6 +8,7 @@
 
 .components-base-control {
 	margin-bottom: 16px;
+	.components-datetime__time-wrapper &,
 	&:last-child {
 		margin-bottom: 0;
 	}
--- a/src/content-gate/editor/editor.scss
+++ b/src/content-gate/editor/editor.scss
@@ -1,4 +1,5 @@
 @use "~@wordpress/base-styles/colors" as wp-colors;
+@use "~@wordpress/base-styles/variables" as wp-vars;
 @use "../vars" as gate;
 
 .edit-post-post-visibility {
@@ -6,7 +7,7 @@
 }
 
 .newspack-content-gate-panel .components-base-control {
-	margin-bottom: 16px;
+	margin-bottom: wp-vars.$grid-unit-20;
 	&:last-child {
 		margin-bottom: 0;
 	}
--- a/src/content-gate/editor/editor.scss
+++ b/src/content-gate/editor/editor.scss
@@ -1,5 +1,4 @@
 @use "~@wordpress/base-styles/colors" as wp-colors;
-@use "~@wordpress/base-styles/variables" as wp-vars;
 @use "../vars" as gate;
 
 .edit-post-post-visibility {
@@ -59,28 +58,3 @@
 	display: none;
 }
 
-/**
- * Access control block visibility panel.
- */
-.newspack-access-control-block-visibility-panel {
-	.components-base-control,
-	.components-form-token-field {
-		width: 100%;
-	}
-	// Gutenberg's `.block-editor-block-inspector p:not(.components-base-control__help)` rule
-	// strips the emotion margin-top on FormTokenField's help text. Nested here to bump
-	// specificity above Gutenberg's and restore the gap.
-	.components-form-token-field .components-form-token-field__help {
-		margin-top: wp-vars.$grid-unit-10;
-	}
-	.components-toggle-control {
-		margin-top: wp-vars.$grid-unit-10;
-		width: 100%;
-		.components-base-control__field {
-			margin-bottom: wp-vars.$grid-unit-05;
-		}
-		.components-base-control__help {
-			margin-top: 0;
-		}
-	}
-}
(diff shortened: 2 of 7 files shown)


src/content-gate/editor/editor.scss
```diff
@@ -8,6 +7,7 @@
 
 .components-base-control {
 	margin-bottom: 16px;
+	.components-datetime__time-wrapper &,
 	&:last-child {
 		margin-bottom: 0;
 	}
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [1c9859b](https://github.com/Automattic/newspack-plugin/pull/4716/commits/1c9859b9f4d5bd7616ef80d55d0558410746d25f)

src/content-gate/editor/block-visibility.tsx
```diff
@@ -22,7 +22,7 @@ import { __, sprintf } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import './editor.scss';
+import './block-visibility.scss';
```
adekbadek, line 25: **Blocker:** Moving these styles into a dedicated `block-visibility.scss` makes this the only importer, so Webpack now emits `dist/content-gate-block-visibility.css` instead of folding the rules into `commons.css`. But `Block_Visibility` (`includes/content-gate/class-block-visibility.php`) only `wp_enqueue_script`s `content-gate-block-visibility.js` — there's no matching `wp_enqueue_style`. 

Verified in a built env: on the post editor the script loads but the stylesheet does not, so the Access Control panel loses its `width: 100%`, FormTokenField help spacing, and ToggleControl spacing. Add `wp_enqueue_style( 'newspack-content-gate-block-visibility', Newspack::plugin_url() . '/dist/content-gate-block-visibility.css', [], $asset['version'] );` next to the script enqueue.

image

src/content-gate/editor/editor.scss
```diff
@@ -1,12 +1,11 @@
 @use "~@wordpress/base-styles/colors" as wp-colors;
-@use "~@wordpress/base-styles/variables" as wp-vars;
 @use "../vars" as gate;
 
 .edit-post-post-visibility {
 	display: none;
 }
 
-.components-base-control {
+.newspack-content-gate-panel .components-base-control {
```
adekbadek, line 9: **Suggestion:** Scoping this rule under `.newspack-content-gate-panel` only covers the two panels in `editor.js` that received the class. The `content-gate-metering-panel` (`metering-settings.js`: Checkbox/Text/Select controls) and `content-gate-post-exemptions-panel` (`post-settings.js`: a ToggleControl) also render `.components-base-control`s, don't carry the class, and don't use `__nextHasNoMarginBottom` — so they lose the 16px spacing they previously got from the global rule. Either add the class to those panels too or pull it into a shared constant so the next panel doesn't rediscover this.

src/content-gate/editor/editor.scss
```diff
@@ -1,12 +1,11 @@
...
 	margin-bottom: 16px;
```
adekbadek: **Nit:** `16px` is hardcoded here while the sibling `block-visibility.scss` uses `wp-vars.$grid-unit-*` tokens. Value carried over verbatim so it's not introduced by this PR, but the two files now sit side by side with inconsistent spacing conventions — `$grid-unit-20` (16px) would match.

adekbadek (changes requested):
The chunking diagnosis is correct and the NPPM-2812 schedule-popup fix is verified working (the global `.components-base-control` margin no longer reaches `commons.css`). One blocker and some comments.

src/content-gate/editor/block-visibility.tsx
laurelfulford: 🤦‍♀️ Helps to actually enqueue the CSS to have it used. Thanks @adekbadek! This should be fixed in [4d466c6](https://github.com/Automattic/newspack-plugin/pull/4716/commits/4d466c65bcab0539338d89384d3e541e0f4a8728).

src/content-gate/editor/editor.scss
laurelfulford: Confirmed that this got cozy on the Metering Settings - before: 

CleanShot 2026-05-26 at 13 49 51

Should look like this now: 

CleanShot 2026-05-26 at 13 50 30

Best I can tell Access control settings looks the same in either case, likely because they just have the one toggle right now -- adding the class doesn't hurt, either:
 
CleanShot 2026-05-26 at 14 02 42

Classes added to both in [e5a95b5](https://github.com/Automattic/newspack-plugin/pull/4716/commits/e5a95b5e7bf5626862a4ce3fe2ee4d6af50d98f8).

src/content-gate/editor/editor.scss
laurelfulford: Updated to `$grid-unit-20` in [15c6a2b](https://github.com/Automattic/newspack-plugin/pull/4716/commits/15c6a2be6a9d51ecf1fd3010a81f222da57a5972)

laurelfulford:
Thanks @adekbadek -- everything should be addressed! Can you take another peek and let me know if anything else looks amiss!

adekbadek (approved):
All three items addressed in code: missing CSS enqueue, `newspack-content-gate-panel` scoping class on Metering + PostSettings, and `$grid-unit-20` token. Verified via the before/after screenshots.

laurelfulford:
Thanks for the review, @adekbadek! 🙌

PR #4723, opened 2026-05-13
feat(alert-manager): route alerts to Slack via newspack_log

miguelpeixe:
### All Submissions:

Adds a `forward_alert_to_log` listener on `newspack_alert` in `Alert_Manager`. The listener emits `do_action( 'newspack_log', $code, $message, [ 'log_level' => 3, ... ] )`, which Newspack Manager's `Logger` already routes through Jetpack to the Newspack Slack alerts channel.

Existing precedent for `log_level => 3` → Slack alert is in `newspack-manager`'s `class-data-report.php:352`.

Closes NPPD-1443.

1. Check out this branch on a site with `newspack-manager` active and connected to a hub (so `newspack_log` actually round-trips through Jetpack).
2. From `wp shell` (or a quick snippet), fire a synthetic alert:

   ```php
   do_action( 'newspack_alert', [
       'type'      => 'sync_retry_exhausted',
       'severity'  => 'error',
       'message'   => 'Synthetic test from alert-log-forward PR',
       'context'   => [ 'integration_id' => 'mailchimp' ],
       'timestamp' => time(),
   ] );
   ```

3. Confirm a level 3 log is fired with the alert message

### Other information:

### DIFF
--- a/includes/class-alert-manager.php
+++ b/includes/class-alert-manager.php
@@ -103,7 +103,14 @@ class Alert_Manager {
 
 	/**
 	 * Forward a `newspack_alert` to the `newspack_log` action so Newspack
-	 * Manager's Logger routes it to Slack (log_level 3 = Alert channel).
+	 * Manager's Logger routes it. Severity drives the destination:
+	 *
+	 *   - severity = 'warning' → type 'debug', log_level 2 (Watch — logstash only)
+	 *   - anything else        → type 'error', log_level 3 (Alert — Slack)
+	 *
+	 * Only the human-readable `message` is forwarded; the full alert
+	 * `context` is intentionally dropped to avoid leaking source payloads
+	 * (which may include reader identifiers) into downstream logs.
 	 *
 	 * When Newspack Manager isn't active, `newspack_log` is a no-op.
 	 *
@@ -118,14 +125,16 @@ class Alert_Manager {
 			? (string) $alert['type']
 			: 'newspack_alert';
 
+		$severity   = is_scalar( $alert['severity'] ?? null ) ? (string) $alert['severity'] : '';
+		$is_warning = 'warning' === $severity;
+
 		do_action(
 			'newspack_log',
 			$code,
 			(string) $alert['message'],
 			[
-				'type'      => 'error',
-				'log_level' => 3,
-				'data'      => $alert['context'] ?? [],
+				'type'      => $is_warning ? 'debug' : 'error',
+				'log_level' => $is_warning ? 2 : 3,
 			]
 		);
 	}
--- a/includes/class-alert-manager.php
+++ b/includes/class-alert-manager.php
@@ -96,8 +96,38 @@ class Alert_Manager {
 		add_action( 'newspack_sync_retry_exhausted', [ __CLASS__, 'handle_sync_retry_exhausted' ] );
 		add_action( 'newspack_data_event_retry_exhausted', [ __CLASS__, 'handle_data_event_retry_exhausted' ] );
 		add_action( 'newspack_integration_health_check_failed', [ __CLASS__, 'handle_health_check_failed' ] );
+		add_action( 'newspack_alert', [ __CLASS__, 'forward_alert_to_log' ] );
 		add_action( self::PATTERN_SCAN_HOOK, [ __CLASS__, 'scan_failure_patterns' ] );
 		add_action( 'init', [ __CLASS__, 'schedule_pattern_scan' ] );
+	}
+
+	/**
+	 * Forward a `newspack_alert` to the `newspack_log` action so Newspack
+	 * Manager's Logger routes it to Slack (log_level 3 = Alert channel).
+	 *
+	 * When Newspack Manager isn't active, `newspack_log` is a no-op.
+	 *
+	 * @param mixed $alert The alert payload fired by this class.
+	 */
+	public static function forward_alert_to_log( $alert ) {
+		if ( ! is_array( $alert ) || ! isset( $alert['message'] ) || ! is_scalar( $alert['message'] ) || '' === (string) $alert['message'] ) {
+			return;
+		}
+
+		$code = is_scalar( $alert['type'] ?? null ) && '' !== (string) $alert['type']
+			? (string) $alert['type']
+			: 'newspack_alert';
+
+		do_action(
+			'newspack_log',
+			$code,
+			(string) $alert['message'],
+			[
+				'type'      => 'error',
+				'log_level' => 3,
+				'data'      => $alert['context'] ?? [],
+			]
+		);
 	}
 
 	/**
--- a/includes/class-alert-manager.php
+++ b/includes/class-alert-manager.php
@@ -105,12 +105,21 @@ class Alert_Manager {
 	 * Forward a `newspack_alert` to the `newspack_log` action so Newspack
 	 * Manager's Logger routes it. Severity drives the destination:
 	 *
-	 *   - severity = 'warning' → type 'debug', log_level 2 (Watch — logstash only)
-	 *   - anything else        → type 'error', log_level 3 (Alert — Slack)
+	 *   - severity = 'error' or 'critical' → type 'error', log_level 3
+	 *     (Alert — Slack)
+	 *   - anything else (incl. 'warning', unknown, or missing severity) →
+	 *     type 'debug', log_level 2 (Watch — logstash only)
 	 *
-	 * Only the human-readable `message` is forwarded; the full alert
-	 * `context` is intentionally dropped to avoid leaking source payloads
-	 * (which may include reader identifiers) into downstream logs.
+	 * Only known error severities escalate to Slack so an unanticipated
+	 * alert shape (e.g. a third-party `newspack_alert` with no severity)
+	 * lands in Watch rather than paging on-call.
+	 *
+	 * Only the human-readable `message` is forwarded as free text. Any
+	 * contact email carried in the alert `context` is passed through
+	 * Logger's first-class `user_email` param — a structured field that is
+	 * not part of the Slack message body — instead of being interpolated
+	 * into `message`. The rest of the `context` is intentionally dropped to
+	 * avoid leaking source payloads into downstream logs.
 	 *
 	 * When Newspack Manager isn't active, `newspack_log` is a no-op.
 	 *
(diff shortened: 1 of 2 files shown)


tests/unit-tests/alert-manager.php
```diff
@@ -357,6 +358,73 @@ function ( $data ) use ( &$fire_count ) {
...
+		add_action(
+			'newspack_log',
+			function ( $code, $message, $params ) use ( &$captured ) {
+				$captured = compact( 'code', 'message', 'params' );
+			},
+			10,
+			3
+		);
+
+		Alert_Manager::forward_alert_to_log(
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 82bc3d330. The test now dispatches via `do_action( 'newspack_alert', $alert )` after re-registering the listener (tear_down strips it), so the action wiring is exercised end-to-end. Same for the malformed-payload cases.

includes/class-alert-manager.php
```diff
@@ -96,10 +96,40 @@ public static function init() {
...
+			: 'newspack_alert';
+
+		do_action(
+			'newspack_log',
+			$code,
+			(string) $alert['message'],
+			[
+				'type'      => 'error',
+				'log_level' => 3,
+				'data'      => $alert['context'] ?? [],
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 82bc3d330. Dropped the `data` field entirely — only the human-readable `message` is forwarded now. The alert messages already include the relevant identifiers in a curated form, so passing the full source context downstream was redundant.

includes/class-alert-manager.php
```diff
@@ -96,10 +96,40 @@ public static function init() {
...
+		$code = is_scalar( $alert['type'] ?? null ) && '' !== (string) $alert['type']
+			? (string) $alert['type']
+			: 'newspack_alert';
+
+		do_action(
+			'newspack_log',
+			$code,
+			(string) $alert['message'],
+			[
+				'type'      => 'error',
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 82bc3d330. Severity now drives both `type` and `log_level`: `warning` maps to `type=debug`+`log_level=2` (Watch — logstash only), and everything else falls back to `type=error`+`log_level=3` (Alert — Slack). New test `test_warning_severity_alert_uses_watch_level` pins the warning path.

includes/class-alert-manager.php
```diff
@@ -96,10 +96,49 @@ public static function init() {
...
+	/**
+	 * Forward a `newspack_alert` to the `newspack_log` action so Newspack
+	 * Manager's Logger routes it. Severity drives the destination:
+	 *
+	 *   - severity = 'warning' → type 'debug', log_level 2 (Watch — logstash only)
+	 *   - anything else        → type 'error', log_level 3 (Alert — Slack)
+	 *
+	 * Only the human-readable `message` is forwarded; the full alert
+	 * `context` is intentionally dropped to avoid leaking source payloads
+	 * (which may include reader identifiers) into downstream logs.
```
chickenn00dle: The docblock here says `context` is dropped to avoid leaking reader identifiers, but the email is already interpolated into `message` itself by the upstream handlers — see `handle_sync_retry_exhausted` (`$payload['contact']['email']` via `sprintf`) and the `same_user` branch of `scan_failure_patterns` (where `$group_value` *is* the email). Since `message` is the field that gets forwarded to Slack, the dropping-context defense doesn't actually cover the highest-volume PII case.

Two ways to resolve:

1. Stop interpolating the email into the upstream messages and pass it through Logger's first-class `user_email` param from this listener (the alert's `context.contact.email` is right there).
2. Drop the misleading docblock claim about reader identifiers.

includes/class-alert-manager.php
```diff
@@ -96,10 +96,49 @@ public static function init() {
...
+		if ( ! is_array( $alert ) || ! isset( $alert['message'] ) || ! is_scalar( $alert['message'] ) || '' === (string) $alert['message'] ) {
+			return;
+		}
+
+		$code = is_scalar( $alert['type'] ?? null ) && '' !== (string) $alert['type']
+			? (string) $alert['type']
+			: 'newspack_alert';
+
+		$severity   = is_scalar( $alert['severity'] ?? null ) ? (string) $alert['severity'] : '';
+		$is_warning = 'warning' === $severity;
```
chickenn00dle: The current direction treats anything other than `'warning'` as Slack-worthy. All four current dispatch sites hard-code `severity => 'error'` so this works today, but a future alert with missing or unknown severity will page Slack — exactly the kind of "noisy unknown shape" we'd usually want in Watch, not Slack.

Consider inverting so only known error severities escalate, defaulting everything else to Watch:

```php
$is_error = in_array( $severity, [ 'error', 'critical' ], true );
do_action(
    'newspack_log',
    $code,
    (string) $alert['message'],
    [
        'type'      => $is_error ? 'error' : 'debug',
        'log_level' => $is_error ? 3 : 2,
    ]
);
```

Matches the docblock above (which only documents `'error'` and `'warning'`) and is more conservative if a third party ever fires `newspack_alert` with a shape we didn't anticipate.

includes/class-alert-manager.php
```diff
@@ -96,10 +96,49 @@ public static function init() {
 		add_action( 'newspack_sync_retry_exhausted', [ __CLASS__, 'handle_sync_retry_exhausted' ] );
 		add_action( 'newspack_data_event_retry_exhausted', [ __CLASS__, 'handle_data_event_retry_exhausted' ] );
 		add_action( 'newspack_integration_health_check_failed', [ __CLASS__, 'handle_health_check_failed' ] );
+		add_action( 'newspack_alert', [ __CLASS__, 'forward_alert_to_log' ] );
```
chickenn00dle, line 99: Out of scope for this PR (the PR doesn't change *whether* alerts fire), but worth flagging as a likely follow-up: with this listener registered, every retry-exhaustion alert now lands directly in Slack. The three direct `handle_*_exhausted` paths fire `newspack_alert` every time without dedup — only `scan_failure_patterns` has transient-based suppression. If an ESP is down, this could turn into hundreds of identical `sync_retry_exhausted` Slack messages per hour from a single site.

Could be addressed with a small transient keyed by `integration_id + reason` for a few minutes. Probably worth a follow-up NPPD issue.

tests/unit-tests/alert-manager.php
```diff
@@ -357,6 +358,109 @@ function ( $data ) use ( &$fire_count ) {
...
+
+	/**
+	 * Malformed alert payloads.
+	 */
+	public function data_malformed_alerts() {
+		return [
+			'non-array'            => [ 'string' ],
+			'missing message'      => [ [ 'type' => 'x' ] ],
+			'non-scalar message'   => [ [ 'message' => [ 'not', 'a', 'string' ] ] ],
+			'empty string message' => [ [ 'message' => '' ] ],
```
chickenn00dle, line 641: A few small additions worth considering — each is a one-liner / data-provider entry:

- **Missing `type` key** — locks in the `'newspack_alert'` default code at the fallback (currently exercised only implicitly by the warning-severity test, which happens to provide `type`).
- **Unknown severity** (`'critical'`, `'info'`, `''`, missing key) — confirms the chosen default routing. Especially relevant if you flip the default per the comment on line 129.
- **Numeric / boolean `message`** (`0`, `'0'`, `false`, `42`) — the `is_scalar` + `'' === (string) ...` check has subtle edges: numeric `0` casts to `'0'` (non-empty, would forward); `false` casts to `''` (correctly skipped). Pinning either behavior locks down the intent.

The `data_malformed_alerts` provider pattern is great — keep it.

chickenn00dle (approved):
LGTM — local checks (lint, PHPUnit for `Alert_Manager`: 18 tests / 72 assertions) and CI are all green. The `newspack_log` contract — `( $code, $message, $params )` with `type ∈ error|debug` and `log_level ∈ 1..4` — matches `Logger::log` and the precedent at `newspack-manager`'s `class-data-report.php:352`. Severity-to-level mapping aligns; dropping `context` from the forwarded params is correct (passing as `data` would JSON-encode reader identifiers into the log payload).

Approving — none of the inline comments are blockers, just refinements worth considering before merge.

includes/class-alert-manager.php
miguelpeixe: Addressed in 5686646cd, option (a). Stopped interpolating the email into the upstream messages (`handle_sync_retry_exhausted` and the `same_user` branch of `scan_failure_patterns`) and now forward it through Logger's structured `user_email` param via a new `get_alert_user_email()` helper — it pulls `context.contact.email` for sync/handler payloads and `context.group_value` for `contact_email`-grouped pattern alerts. `user_email` is a structured field (not part of the Slack message body), and the docblock now reflects this accurately. New tests assert the email reaches `user_email` and is absent from `message`.

includes/class-alert-manager.php
miguelpeixe: Addressed in 5686646cd. Inverted as suggested: `$is_error = in_array( $severity, [ 'error', 'critical' ], true )` — only known error severities escalate to Slack (type `error`, log_level 3); everything else (incl. `warning`, unknown, missing) goes to Watch (type `debug`, log_level 2). Docblock updated. New `data_severity_routing` provider pins error/critical/warning/info/empty/missing.

includes/class-alert-manager.php
miguelpeixe: Agreed it's out of scope here. Tracked as a follow-up: NPPD-1534 — "Dedup direct retry-exhaustion alerts before forwarding to Slack" (sub-issue of NPPD-1443).

tests/unit-tests/alert-manager.php
miguelpeixe: Addressed in 5686646cd. Added `false message` to the `data_malformed_alerts` provider (casts to `''`, correctly skipped); added `test_alert_without_type_uses_default_code` (pins the `newspack_alert` fallback code), `test_numeric_zero_message_is_forwarded` (`0` → `'0'`, forwarded), and a `data_severity_routing` provider covering unknown/missing severity. Suite: 28 tests / 98 assertions, green.

miguelpeixe:
Thanks, @chickenn00dle! I've addressed the suggestions.

chickenn00dle (approved):

PR #4718, opened 2026-05-12
chore(deps): drop joshtronic/php-loremipsum, inline starter content lipsum

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

The `joshtronic/php-loremipsum` package is used in the starter content generator (`Starter_Content_Generated::get_lipsum()`) to produce dummy paragraphs and words for the demo site bootstrapped by the setup wizard. It's no longer available, so this PR removes the dependency and inlines a tiny equivalent generator (an embedded word list plus two small helpers, `lipsum_words()` and `lipsum_paragraphs()`).

The public signature of `get_lipsum( $type, $amount )` is unchanged, so the three existing callers (post titles, excerpts, body paragraphs) keep working exactly as before. Output remains lowercase space-separated words for `'words'` and sentence-cased paragraphs joined by `PHP_EOL` for `'paras'`.

Closes NPPM-2827

### How to test the changes in this Pull Request:

1. Pull this branch and run `composer install` in `newspack-plugin`. Confirm `vendor/joshtronic/` no longer exists and `composer.lock` no longer mentions `joshtronic/php-loremipsum`.
2. On a fresh site, run the Newspack setup wizard (or `wp newspack starter-content generate` if available) and let it generate starter posts.
3. Verify the generated posts have:
   - Titles that look like lorem-ipsum word strings (7–14 words, sentence-cased, no periods).
   - Excerpts of ~20–30 lowercase words ending in a period.
   - Bodies composed of several paragraphs of varied sentence length, separated by paragraph breaks.
4. Smoke test directly via WP-CLI:
   ```bash
   wp eval 'echo Newspack\Starter_Content_Generated::get_lipsum("words", 10), "\n---\n", Newspack\Starter_Content_Generated::get_lipsum("paras", 2);'
   ```
   Confirm `words` returns a single space-separated string with no punctuation and no leading uppercase, and `paras` returns paragraphs separated by `\n`, each containing multiple sentences.

### Other information:

[...]

PR #4724, opened 2026-05-13
fix(parse.ly): update meta_type default

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPM-2833/possible-conflict-between-yoast-and-our-required-parsely-code-that-may

This PR updates the default meta_type parse.ly config to `repeated_metas` which is [recommended by parse.ly ](https://docs.parse.ly/installation-resources/installation-tools/wordpress-plugin/#basic-optional-settings). This also implements a migrator so existing sites with Parse.ly get the updated setting.

We need to do this to avoid the conflict pointed out in the linear task above when Yoast is active.

**Removing parsely**
These tests will require removing parsely and parsely settings several times. You can do so with the following:

```
wp plugin delete wp-parsely
wp option delete parsely
```

1. On `release`, add the following snippet to a site where parsely is removed:
```
	$request = new WP_REST_Request( 'POST', '/newspack/v1/plugins/wp-parsely/configure' );
	$response = rest_do_request( $request );
} );
```
2. Reload the page and go to Parsely settings.
3. Confirm you see json_ld selected for Metadata Format:
Screenshot 2026-05-14 at 09 41 25

4. Checkout this branch and reload the page
5. Confirm Metadata Format is now set to `repeated_metas`
6. Now remove parsely once again
7. Ensure the snippet from step 1 is still present, then reload the page. (You'll likely need to navigate away and back again to the parsely settings page if you removed following the steps above)
8. Confirm Metadata Format is set to `repeated_metas`

### Other information:


chickenn00dle:
Just noting the lint failure was fixed here in trunk: https://github.com/Automattic/newspack-plugin/pull/4718

includes/plugins/class-parsely.php
```diff
@@ -0,0 +1,53 @@
...
+		if ( get_option( self::META_TYPE_MIGRATION_OPTION ) ) {
+			return;
+		}
+
+		if ( ! is_plugin_active( 'wp-parsely/wp-parsely.php' ) ) {
+			return;
+		}
+
+		$parsely_settings = get_option( 'parsely', [] );
+		if ( is_array( $parsely_settings ) && isset( $parsely_settings['meta_type'] ) && 'json_ld' === $parsely_settings['meta_type'] ) {
```
adekbadek: **Blocker:** Migration misses (and permanently marks done) sites with no stored meta_type key — The condition requires `meta_type` to be explicitly stored as `json_ld`, but wp-parsely's own default for an absent key is also `json_ld` (it merges `OPTION_DEFAULTS` at read time in `get_options()`). So a parsely-active site whose stored option never persisted `meta_type` renders the conflicting JSON-LD this PR targets, yet is skipped here — and line 50 still marks it migrated, so it's stuck on the broken output with no re-run path. Reproduced in an isolated env: a stored option of `{"apikey":"example.com"}` (no `meta_type`) renders `json_ld`, is skipped, and gets flagged `v1`.

**Consider:** treat an absent key as the legacy default, e.g. `$meta_type = $parsely_settings['meta_type'] ?? 'json_ld';` then rewrite when `'json_ld' === $meta_type` (this also covers an empty stored option array).

includes/plugins/class-parsely.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
adekbadek, line 1: **Suggestion:** Classmap not regenerated for the new class — `Newspack\Parsely` is absent from `vendor/composer/autoload_classmap.php`. It loads fine at runtime because `class-newspack.php` `include_once`s it explicitly, so there's no functional bug, but per the project convention (classmap autoload — run `composer dump-autoload` after adding a PHP file) the classmap should be regenerated so the class is discoverable via autoload like its siblings. Note the classmap is build-time generated, so CI may cover this; worth confirming.

includes/plugins/class-parsely.php
```diff
@@ -0,0 +1,53 @@
...
+		add_action( 'admin_init', [ __CLASS__, 'migrate_meta_type' ] );
+	}
+
+	/**
+	 * Migrate existing Parse.ly installs from the legacy `json_ld` meta_type
+	 * default to `repeated_metas`, which avoids conflicts with Yoast SEO's
+	 * own JSON-LD output. Runs once per site.
+	 */
+	public static function migrate_meta_type() {
+		if ( get_option( self::META_TYPE_MIGRATION_OPTION ) ) {
```
adekbadek, line 48: **Suggestion:** No test for the migrator's branches — The migrator has several non-obvious branches (already-migrated short-circuit, plugin-inactive short-circuit, non-array option, missing `meta_type` key, value !== `json_ld`, and the happy path). Per the repo's testing conventions this is exactly the kind of logic that benefits from a small `WP_UnitTestCase` as a self-proving spec — in particular the idempotency guarantee and the matching rule (which the blocker above shows is easy to get subtly wrong).

includes/plugins/class-parsely.php
adekbadek: **Suggestion:** Silently overrides a deliberate publisher json_ld choice — The migrator cannot distinguish "this site still has Newspack's old `json_ld` default" from "a publisher deliberately chose `json_ld`" — both look identical in the stored option. Switching the latter is a silent behavior change. Given the conflict is Yoast-driven the blanket switch is defensible, but it's worth calling out in the PR (or a comment) that the override is intentional and unconditional, since on a site without Yoast `json_ld` may be a legitimate working choice.

includes/plugins/class-parsely.php
```diff
@@ -0,0 +1,53 @@
...
+			return;
+		}
+
+		$parsely_settings = get_option( 'parsely', [] );
+		if ( is_array( $parsely_settings ) && isset( $parsely_settings['meta_type'] ) && 'json_ld' === $parsely_settings['meta_type'] ) {
+			$parsely_settings['meta_type'] = 'repeated_metas';
+			update_option( 'parsely', $parsely_settings );
+		}
+
+		update_option( self::META_TYPE_MIGRATION_OPTION, 'v1' );
```
adekbadek: **Suggestion:** Migrator re-runs on every admin_init for non-parsely sites — The sentinel `update_option( META_TYPE_MIGRATION_OPTION, 'v1' )` is written only after the `is_plugin_active()` early-return at line 40. On sites without wp-parsely — the majority of Newspack installs — completion is never recorded, so `migrate_meta_type()` re-runs its `get_option()` lookups on every single `admin_init`, forever. The per-request cost is two cached option reads with no compounding side effect, so it's cheap, but it's an unusual shape for a "runs once" migrator and the docblock's "Runs once per site" is misleading for that population.

includes/plugins/class-parsely.php
```diff
@@ -0,0 +1,53 @@
...
+class Parsely {
+	/**
+	 * Option to track whether the meta_type migration has run.
+	 *
+	 * @var string
+	 */
+	const META_TYPE_MIGRATION_OPTION = 'newspack_parsely_meta_type_migrated';
+
+	/**
+	 * Initialize hooks and filters.
```
adekbadek: **Nit:** Docblock inaccuracies — `init()`'s docblock reads "Initialize hooks and filters" but only one action is registered; the class docblock is the generic "Main class." carried over from sibling templates. Minor, against the comment-describes-current-state convention — "Register hooks." / a one-line description of the Parse.ly migrator would be precise.

includes/plugins/class-parsely.php
adekbadek, line 48: **Nit:** 'v1' sentinel value is never read — The gate at line 36 is a plain truthy check (`if ( get_option( ... ) )`), so storing `'v1'` versus `true`/`1` makes no behavioral difference today — and a future `'v2'` re-migration would be short-circuited by the same truthy guard before it could compare versions. Either store `true` and drop the implied versioning, or if versioned re-runs are genuinely anticipated, compare the stored version explicitly.

adekbadek (changes requested):
Verified end-to-end in an isolated env: the new default serves `repeated_metas`, and a stored `meta_type=json_ld` flips correctly on the real `admin_init` path.

One blocker and a few suggestions and nits.

includes/plugins/class-parsely.php
chickenn00dle: Good catch — fixed in 117704c1d. An absent `meta_type` is now treated as the legacy default (`$meta_type = $parsely_settings['meta_type'] ?? 'json_ld'`), so a stored option without the key (and an empty option array) is migrated too instead of being skipped-and-flagged. Covered by `test_migrates_absent_meta_type` and `test_migrates_empty_option`.

includes/plugins/class-parsely.php
chickenn00dle: No change needed here: `vendor/` is gitignored and the classmap isn't tracked — it's regenerated at build/CI time, so there's nothing to commit. The class loads fine via the explicit `include_once` in `class-newspack.php` and will be in the generated classmap like its siblings after the build runs.

includes/plugins/class-parsely.php
chickenn00dle: Added in 117704c1d — `tests/unit-tests/parsely.php` covers the branches: explicit `json_ld`, absent key, empty option, non-`json_ld` left untouched, inactive-plugin stays-pending, and the idempotency guarantee (7 tests pass).

includes/plugins/class-parsely.php
chickenn00dle: Agreed it's worth calling out — documented in 117704c1d. The migrator docblock now states the switch is intentional and unconditional for any site still rendering `json_ld`, and that a deliberate publisher `json_ld` choice is indistinguishable from the old default and gets switched too, which is desired since the Yoast conflict applies regardless of intent.

includes/plugins/class-parsely.php
chickenn00dle: Fixed the misleading docblock in 117704c1d. The re-run is intentional: keeping it pending until wp-parsely is active means a site that installs Parse.ly later still gets migrated (writing the sentinel before the active-check would permanently skip those). The docblock now says it runs on every `admin_init` until wp-parsely is active, then migrates once — and notes the cheap two-cached-reads cost for sites that never activate it.

includes/plugins/class-parsely.php
chickenn00dle: Fixed in 117704c1d — `init()` is now "Register hooks." and the class docblock describes the Parse.ly migrator instead of the generic "Main class."

includes/plugins/class-parsely.php
chickenn00dle: Fixed in 117704c1d — stores `true` now and dropped the implied versioning. This is a one-time migration; if a future re-migration is ever needed it'd warrant its own option anyway.

chickenn00dle:
Thanks for the review @adekbadek! This is ready for another look.

adekbadek (approved):
Migration blocker resolved: an absent `meta_type` key and an empty option both migrate now via the `?? 'json_ld'` default, with unit tests locking each branch (including the inactive-plugin stays-pending and idempotency cases). The `admin_init` re-run for late wp-parsely activations and the unconditional override of a deliberate `json_ld` choice are both documented and sound.

adekbadek (approved):
Migration blocker resolved: an absent `meta_type` key and an empty option both migrate now via the `?? 'json_ld'` default, with unit tests locking each branch (including the inactive-plugin stays-pending and idempotency cases). The `admin_init` re-run for late wp-parsely activations and the unconditional override of a deliberate `json_ld` choice are both documented and sound.

PR #4729, opened 2026-05-14
feat(block-theme): add search overlay block

laurelfulford:
### All Submissions:

This PR adds a Search Overlay Block to be used in the Newspack Block Theme. It's inserted as a button into the block theme, and when clicked, will open an overlay with the search block in it. The button appearance (colours, fonts, border radius), label (changing text or hiding text/icon) and overlay (colour) are customizeable, but the search -- for now -- is not). 

When Jetpack's instant search is enabled, it will be opened instead.

A version of this functionality is currently baked into the theme using template parts. The downside is that that's pretty fragile - changing the structure or a class name could break how it works. My moving the functionality to block, we make sure it's easier to use, add, and customize.

Closes NPPD-1449.

1. On a test site using the Block Theme, open the Site Editor. 
2. Add the Search Overlay block (under the Newspack category, or by searching for 'Search'):

CleanShot 2026-05-18 at 11 45 26

3. Save and view on the frontend in a separate tab so you can flip between editor and frontend.
4. Try clicking the search button and confirm it opens a search overlay:

CleanShot 2026-05-14 at 13 08 32

5. Try running a search from the overlay and confirm it redirects you to the search results page.
6. Open the search a few times and confirm you can close it:
    * With the close button (top-right corner)
    * By clicking the overlay background
    * By hitting the Esc key
7. While you're testing the above, also confirm that the input is focused on when you open the search overlay.

#### Test the visual stuff
1. In the editor sidebar > Styles, confirm you have three options: Default, Icon only, Text only.
2. Try testing each and toggling between the editor and front-end to review:
    * Default should show the search icon and text
    * Icon only should show only the icon, with the text wrapped in a screen-reader-text CSS class.
    * Text only should show only the text label.
3. With the text visible, try editing the text on the button to change the label to something other than "Search".
4. Try testing the customization options -- colour settings, typography settings, border radius... confirm everything's applied and looks right on the front-end. 
5. Try changing the overlay colour specifically and testing on the front-end -- the close button in the-top right corner should remain contrast-y against the overlay, so it should be black against a really light colour, and white against a darker colour.

#### Confirm the block is accessible

1. With overlay closed, inspect the button on the front-end -- it should have the attributes `aria-expanded="false", aria-controls="newspack-overlay-search-panel-…"`
2. Inspect the panel itself and confirm it has the `aria-hidden="true"`, `inert` attributes
3. Open the and confirm the button switches to `aria-expanded="true"` and the panel loses `aria-hidden` and `inert`
4. Close the overlay → attributes flip back.

#### Jetpack Instant Search

1. Enable Jetpack's instant search
2. Try opening an overlay and running a search - confirm Jetpack's instant search is opened instead of redirecting you to the search results page.

#### Get weird & throw a search party! 

1. Set up a few instances of the search button on the same page. Give the overlays different settings so you can tell them apart.
2. Test triggering them from the front-end. Each search button should open is own overlay.
3. Confirm that the input still gets focus when the search overlay is opened.
4. Ensure that the `esc` key still closes the open overlay.

### Other information:

---

### Update (scope)

In addition to introducing the new Search Overlay block, this PR includes two related changes to existing Newspack blocks:

1. **Block style registration moved to `block.json`** — for `overlay-search`, `overlay-menu-trigger`, **and `my-account-button`**, the `default` / `icon-only` / `text-only` style variations are now declared in `block.json` `styles` arrays instead of JS-only `registerBlockStyle()` calls. This makes the variations visible to `theme.json` so themes can target them via `styles.blocks.<name>.variations.<style>`.

2. **Overlay Menu inner blocks are now permanently locked** — `overlay-menu-trigger` and `overlay-menu-panel` now bake `lock: { move: true, remove: true }` into their `block.json` attribute defaults. Combined with the existing `supports.lock: false`, the lock UI is hidden and the inner blocks of every `newspack/overlay-menu` instance — including ones already saved in shipped patterns — cannot be moved or deleted, even via the parent block's "Modify" template-bypass UI.

#### Extra test steps for the scope updates

- **theme.json variations**: with the companion theme PR applied, set an Overlay Search / Overlay Menu Trigger / My Account Button to the **Icon only** style. The button should render as a square (0.75rem padding on all four sides) without any pattern-level overrides.
- **Inner block locking**: in the Site Editor, expand a header pattern that uses the Overlay Menu. Click the trigger button or panel child. Confirm no Lock button in the toolbar, no drag handle, no Delete affordance — even after attempting the parent's "Modify" template option.

Companion theme PR: [`Automattic/newspack-block-theme#452`](https://github.com/Automattic/newspack-block-theme/pull/452)


src/blocks/overlay-search/view.js
```diff
@@ -0,0 +1,207 @@
...
+	const match = bg.match( /rgba?\(([^)]+)\)/ );
+	if ( ! match ) {
+		return 'white';
+	}
+	const [ r, g, b ] = match[ 1 ].split( ',' ).map( s => parseFloat( s.trim() ) );
+	if ( ! Number.isFinite( r ) || ! Number.isFinite( g ) || ! Number.isFinite( b ) ) {
+		return 'white';
+	}
+	const yiq = ( r * 299 + g * 587 + b * 114 ) / 1000;
+	return yiq >= 128 ? 'black' : 'white';
```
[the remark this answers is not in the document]
laurelfulford: > For semi-transparent colors like rgba(0,0,0,0.1) over a light page, the RGB values make this choose white even though the effective background is very light, leaving the close icon with poor contrast.

I'm not sure how well we can perfect the contrast calculations - this assumes the site's background is white, which yes, would mean a very low contrast X. But if the site's background is black, a white close button would be preferred here.

Going to leave for now but this might be a good argument for adding a close button colour control down the road if it becomes an issue.

src/blocks/search-overlay/class-search-overlay-block.php
```diff
@@ -0,0 +1,235 @@
...
+		<div
+			id="<?php echo \esc_attr( $panel_id ); ?>"
+			class="newspack-search-overlay__panel"
+			role="dialog"
+			aria-modal="true"
+			aria-hidden="true"
+			inert="true"
+			aria-label="<?php \esc_attr_e( 'Search', 'newspack-plugin' ); ?>"
+			<?php if ( '' !== $overlay_color ) : ?>
+				style="background:<?php echo \esc_attr( $overlay_color ); ?>"
```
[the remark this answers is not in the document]
laurelfulford: Should be fixed in [2c35533](https://github.com/Automattic/newspack-plugin/pull/4729/commits/2c35533947e134bf7fc0163f9365357b043423f4).

src/blocks/search-overlay/class-search-overlay-block.php
```diff
@@ -0,0 +1,235 @@
...
+		);
+	}
+
+	/**
+	 * Render the block.
+	 *
+	 * @param array $attributes Block attributes.
+	 * @return string Block HTML.
+	 */
+	public static function render_block( $attributes ) {
```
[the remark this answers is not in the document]
laurelfulford: Tests added in [3fd1c8a](https://github.com/Automattic/newspack-plugin/pull/4729/commits/3fd1c8a6ccb33024b45c9e36f2cce2f96a1122e2).

src/blocks/search-overlay/view.js
```diff
@@ -0,0 +1,207 @@
...
+		document.addEventListener( 'keydown', onEsc );
+		escCleanup = () => document.removeEventListener( 'keydown', onEsc );
+
+		// Focus the search input. Delayed so the panel is fully laid out
+		// (visibility/display transitions complete) before focusing.
+		setTimeout( () => {
+			const searchInput = panel.querySelector( 'input[type="search"]' );
+			const focusTarget = searchInput || getVisibleFocusable( panel )[ 0 ] || closeBtn;
+			if ( focusTarget && document.contains( focusTarget ) ) {
+				focusTarget.focus();
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [704c243](https://github.com/Automattic/newspack-plugin/pull/4729/commits/704c24317da19691c237afab1486e39fa390eea5).

laurelfulford:
Hey @thomasguillot! for the styles here I tried to match the placement and animation of the template parts. Can you poke around with that plus the button style settings, and let me know if anything needs changing? Thanks!

thomasguillot (commented):
Hey @laurelfulford — 22 commits on top of yours. Grouped summary:

### Block rename: `newspack/search-overlay` → `newspack/overlay-search`
Matches the `overlay-menu` family — slug, paths, PHP class, CSS prefix `newspack-overlay-search__*`, test group, webpack entry. Tests pass.

### Markup parity: overlay-menu trigger now wraps in `.wp-block-button`
Both blocks render identical markup, so both pick up the theme's `.wp-block-button > .wp-block-button__link:has(svg)` rule.

### Icons → `Newspack_UI_Icons` registry
Removed inline `ICON_*` constants from all three PHP files; added a `search` entry to the registry (using the canonical WP path).

### Token alignment (raw rems and stale refs → theme tokens)
- Scrim: `var(--wp--custom--color--neutral-70, var(--newspack-ui-color-neutral-70, #{colors.\$neutral-700}))` — same chain on both blocks.
- Close buttons: button via `--wp--custom--spacing--65`, icon via `--wp--preset--spacing--40`, corners via `--wp--custom--border--radius-small`.
- Admin-bar offset: `top: var(--wp-admin--admin-bar--height, 0)` on both panels.
- Fixed token bugs in overlay-menu: `--wp--preset--spacing--10` → `--wp--custom--spacing--10`; `4rem` fallback on `--wp--custom--spacing--85`.
- overlay-menu panel `panelBackgroundColor` / `panelTextColor` defaults set via `block.json` (`base` / `contrast`).

### Stacking
Both panels: `z-index: 99998` (just below the WP admin toolbar). overlay-menu scrim: `99997`.

### Close button polish (parity across both blocks)
`:focus-visible { outline: 2px solid; outline-offset: 1px; }`, rounded corners, `display: block` on SVG. overlay-search uses grid centering; overlay-menu keeps its inline-flex (icon + "Close" text).

### Cleanup
Dropped redundant trigger flex rules from both blocks — the theme rule now provides `inline-flex` + `gap`.

---

### Still open
1. **Slash inserter** — Newspack blocks don't register in the Site Editor iframe (`window.newspack_blocks` undefined inside). Pre-existing, not this PR.
2. **From earlier code review** (not addressed):
   - `get_block_wrapper_attributes()` lands on inner `<button>` instead of outer wrapper.
   - `style: dist/blocks.css` may not enqueue in template-part usage.
   - Jetpack branch passes `href` through `get_block_wrapper_attributes()`.

More to come.

thomasguillot:
Follow-up commits addressing findings from a Codex review pass — pushed in `c031f2f0c`:

### P1: View script's webpack split-chunk dependency wasn't enqueued
The compiled `overlay-search-block.js` and `overlay-menu-block.js` bundles wait for chunk `4223` (the `dist/commons.js` split-chunk) before running their entry callback. The asset.php files only declare `@wordpress/*` deps, so `newspack_commons` was never enqueued on frontends without Reader Activation active. The trigger click handler would silently never initialize on those sites.

**Fix:**
- Both view scripts now import `domReady` from `@wordpress/dom-ready` instead of the internal `src/utils/index.js` helper. Asset.php now correctly lists `wp-dom-ready` as a dep.
- Both block render callbacks (`Overlay_Search_Block::render_block` and `Overlay_Menu_Block::render_block`) now call `\Newspack\Newspack::load_common_assets()` to ensure `newspack_commons` is enqueued whenever the block renders.

### P2: Duplicate `core/search` input IDs across instances
`Overlay_Search_Block::render_panel` cached the rendered `core/search` HTML in a `static $search_html` variable, which meant every panel on a page reused the same auto-generated input ID, producing invalid DOM and breaking label-for associations.

**Fix:** dropped the static cache; `core/search` now renders fresh per panel so `wp_unique_id()` generates a distinct input ID for each instance.

### Codex P3 (trigger SVG fill) — not a real bug
Codex flagged that `Newspack_UI_Icons` SVGs have `fill="none"` on the root, suggesting icon-only triggers would render blank. Verified on the live frontend (`https://pr-4729-search-overlay.local/`): the theme rule `.wp-block-button > .wp-block-button__link:has(svg) svg { fill: currentcolor; … }` overrides the presentation attribute via CSS specificity, and the `<path>` inherits the resolved color. SVGs render correctly. No fix applied.

All 15 PHPUnit tests still pass after these changes.

thomasguillot:
Two more iterations pushed on top of the Codex fixes:

### Slide-duration scales with panel width (`6fd9eafb`)
The 250ms slide felt right for the smallest drawer but way too fast on `large` / `x-large`. Pulled the duration into a CSS custom property and let each width modifier override it. The opacity transition stays at 125ms (the fade should feel quick regardless of size); only the directional slide scales.

| Width modifier | Slide duration |
|---|---|
| `x-small` (300px) | 250ms |
| `small` (410px) | 275ms |
| `medium` (632px) | 325ms |
| `large` (964px) | 400ms |
| `x-large` (1296px) | 475ms |
| `full-screen` (fade) | 200ms (was 125ms) |

Future tweaks are now single-value edits on the modifiers, not edits to the property list.

### Close-icon contrast switched to APCA (`f6b0e867`)
The close button's `color` is computed in `view.js` by sampling the panel's resolved background and picking black or white. Two changes:

1. **YIQ → APCA W3.** The previous formula was the 1953 NTSC weighting (`r×299 + g×587 + b×114`); now using APCA's perceptual luminance (`sRGBtoY`) with polarity-aware `Lc` computation. Compares `|Lc(black, bg)|` vs `|Lc(white, bg)|` and returns whichever wins. More accurate against mid-tone overlay colors.
2. **Black/white returns swapped to layered token chains.** Returns `var(--wp--custom--color--neutral-100, var(--newspack-ui-color-neutral-100, #000000))` for black and the matching `0`-suffixed chain for white. Same fallback pattern we use for the scrim background.
3. **Call moved from `init()` into `openOverlay()`.** The contrast now recomputes on every open, so runtime changes to the panel background (theme swap, `overlayColor` edit) are picked up instead of being captured once at page load.

Helpers (`sRGBtoY`, `apcaLc`, `pickContrastColor`) each have a JSDoc block explaining what they return and why.

src/blocks/overlay-menu/panel/block.json
```diff
@@ -33,11 +33,11 @@
...
 		"panelBackgroundColor": {
 			"type": "string",
-			"default": ""
+			"default": "var(--wp--preset--color--base)"
 		},
 		"panelTextColor": {
 			"type": "string",
-			"default": ""
+			"default": "var(--wp--preset--color--contrast)"
 		}
```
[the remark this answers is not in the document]
thomasguillot: Addressed in ab7ab42c7 — `panelBackgroundColor` and `panelTextColor` now go through `safecss_filter_attr` before being concatenated into the inline `style`. Verified `var(--wp--preset--color--*)` passes through and `background: red; background-image: url(javascript:...)` strips the injection.

tests/mocks/jetpack-mock.php
```diff
@@ -0,0 +1,35 @@
+<?php // phpcs:disable WordPress.Files.FileName.InvalidClassFileName
+/**
+ * Jetpack class stub for tests.
+ *
+ * Tests that exercise code paths gated on Jetpack's presence (e.g. the search
+ * overlay block's instant-search handoff) include this file. The class is
+ * defined globally once and behavior is toggled via `Jetpack::$test_modules`.
```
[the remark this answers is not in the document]
thomasguillot: Addressed in ab7ab42c7 — docblock now says `Jetpack::$test_active_modules`.

src/blocks/overlay-search/style.scss
```diff
@@ -0,0 +1,101 @@
...
+			outline: 2px solid;
+			outline-offset: 1px;
+		}
+
+		svg {
+			display: block;
+			width: var(--wp--preset--spacing--40, 1.5rem);
+			height: var(--wp--preset--spacing--40, 1.5rem);
+			fill: currentcolor;
+		}
```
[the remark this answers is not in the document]
thomasguillot: Verified false on the live frontend — the block theme's `.wp-block-button > .wp-block-button__link:has(svg) svg { fill: currentcolor; ... }` rule matches our trigger markup (the button is wrapped in `.wp-block-button`) and overrides the SVG's `fill="none"` presentation attribute via CSS specificity. Icons render with the inherited contrast color on the rendered page. Same explanation given in [the Codex follow-up comment](https://github.com/Automattic/newspack-plugin/pull/4729#issuecomment-4520710308).

src/blocks/overlay-menu/style.scss
```diff
@@ -55,13 +33,26 @@
...
+		outline: 2px solid;
+		outline-offset: 1px;
+	}
+
+	svg {
+		display: block;
+		width: var( --wp--preset--spacing--40, 1.5rem );
+		height: var( --wp--preset--spacing--40, 1.5rem );
+		fill: currentcolor;
+	}
```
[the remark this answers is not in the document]
thomasguillot: Same as the overlay-search SVG fill thread — the theme rule `.wp-block-button > .wp-block-button__link:has(svg) svg { fill: currentcolor; ... }` applies because we wrapped the overlay-menu trigger in `.wp-block-button` (commit `2c2bae3`). Verified on the live frontend that triggers render with the correct color.

src/blocks/overlay-menu/trigger/edit.js
```diff
@@ -55,28 +55,31 @@ export default function OverlayMenuTriggerEdit( { attributes, setAttributes, cli
...
-					withoutInteractiveFormatting
-				/>
-			</button>
+			<div className="wp-block-buttons is-layout-flex">
+				<div className="wp-block-button">
+					<button
+						{ ...blockProps }
+						type="button"
+						aria-controls={ instanceId ? `newspack-overlay-panel-${ instanceId }` : undefined }
+						onClick={ e => e.preventDefault() }
```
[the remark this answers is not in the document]
thomasguillot: Addressed in add22363c — the trigger button now conditionally sets `aria-label="Menu"` only when `triggerText` is empty or whitespace-only, so non-empty values aren't double-announced.

src/blocks/overlay-menu/trigger/class-overlay-menu-trigger-block.php
```diff
@@ -62,22 +59,25 @@ public static function render_block( array $attributes, string $content, \WP_Blo
...
+					<?php if ( $show_icon ) : ?>
+						<span class="overlay-menu__icon" aria-hidden="true">
+							<?php echo \Newspack\Newspack_UI_Icons::get_svg( 'menu' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+						</span>
+					<?php endif; ?>
+					<span class="<?php echo esc_attr( $text_class ); ?>">
+						<?php echo esc_html( $trigger_text ); ?>
+					</span>
+				</button>
+			</div>
```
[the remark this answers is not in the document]
thomasguillot: Addressed in add22363c — `$trigger_text` is now normalized: whitespace-only values fall back to `__( 'Menu', 'newspack-plugin' )`, mirroring the pattern in `Overlay_Search_Block::render_block`.

src/blocks/overlay-search/edit.js
```diff
@@ -0,0 +1,101 @@
...
+	const isIconVisible = ! isTextOnly;
+
+	const blockProps = useBlockProps( {
+		className: classnames(
+			blockClassName,
+			'wp-block-button__link',
+			'newspack-overlay-search__trigger',
+			colorProps.className,
+			borderProps.className
+		),
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 48c500031 — `wp-element-button` is now included in the editor trigger's class list, matching the frontend PHP render.

src/blocks/overlay-menu/panel/class-overlay-menu-panel-block.php
```diff
@@ -74,12 +71,20 @@ public static function render_block( array $attributes, string $content, \WP_Blo
...
+			if ( '' !== $safe_bg ) {
+				$panel_styles[] = $safe_bg;
+			}
 		}
 		if ( $panel_text_color ) {
-			$panel_styles[] = 'color:' . esc_attr( $panel_text_color );
+			$safe_color = safecss_filter_attr( 'color: ' . $panel_text_color );
+			if ( '' !== $safe_color ) {
+				$panel_styles[] = $safe_color;
+			}
```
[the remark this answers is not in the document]
thomasguillot: The sibling overlay-search test suite already covers the equivalent `safecss_filter_attr` behavior on the same code path: [`test_overlay_color_strips_css_injection`](https://github.com/Automattic/newspack-plugin/blob/feat/search-overlay-block/tests/unit-tests/class-test-overlay-search-block.php) renders the block with `overlayColor = 'red; background-image: url(javascript:alert(1))'` and asserts `javascript:` and `background-image` are stripped. Since both blocks invoke `safecss_filter_attr` against the user-supplied color string before output, that test gives us confidence the function strips injections as expected — it's WordPress core behavior, not specific to either block. Adding a parallel test on overlay-menu-panel would require bootstrapping a new test file (no suite exists there yet) just to re-verify core WP behavior; skipping for scope.

tests/unit-tests/class-test-overlay-search-block.php
```diff
@@ -0,0 +1,275 @@
...
+		$this->assertDoesNotMatchRegularExpression( $pattern, $this->render( [ 'className' => 'is-style-icon-only' ] ) );
+	}
+
+	/**
+	 * A valid `overlayColor` reaches the panel as an inline `background` style.
+	 */
+	public function test_overlay_color_renders_inline_style() {
+		$output = $this->render( [ 'overlayColor' => '#123456' ] );
+
+		$this->assertMatchesRegularExpression( '/<div[^>]*class="newspack-overlay-search__panel"[^>]*style="background:\s*#123456"/', $output );
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 2bdd615df — regex now allows an optional trailing semicolon (`#123456;?\s*"`). Verified empirically that current WP `safecss_filter_attr` doesn't emit a trailing semicolon for single declarations, but the more permissive pattern is future-proof.

src/blocks/overlay-menu/panel/block.json
```diff
@@ -33,11 +33,11 @@
 		},
 		"panelBackgroundColor": {
 			"type": "string",
-			"default": ""
+			"default": "var(--wp--preset--color--base)"
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 2bdd615df — `panelBackgroundColor` default is now `var(--wp--preset--color--base, #ffffff)`.

src/blocks/overlay-menu/panel/block.json
```diff
@@ -33,11 +33,11 @@
...
 		},
 		"panelTextColor": {
 			"type": "string",
-			"default": ""
+			"default": "var(--wp--preset--color--contrast)"
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 2bdd615df — `panelTextColor` default is now `var(--wp--preset--color--contrast, #000000)`.

tests/mocks/jetpack-mock.php
```diff
@@ -0,0 +1,35 @@
...
+
+		/**
+		 * Whether `$module` is in the active list for this test.
+		 *
+		 * @param string $module Module slug.
+		 * @return bool
+		 */
+		public static function is_module_active( $module ) {
+			return in_array( $module, self::$test_active_modules, true );
+		}
```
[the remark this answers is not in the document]
laurelfulford: Updated in [59fb57d](https://github.com/Automattic/newspack-plugin/pull/4729/commits/59fb57da70b8fb50cf6d7f8c7190c552d9504fa7)!

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+	/**
+	 * Render the block.
+	 *
+	 * @param array $attributes Block attributes.
+	 * @return string Block HTML.
+	 */
+	public static function render_block( array $attributes ): string {
+		// The compiled view script depends on dist/commons.js (a webpack split chunk).
+		// Ensure commons is enqueued so the entry callback can execute on the frontend.
+		\Newspack\Newspack::load_common_assets();
```
chickenn00dle, line 48: `Newspack::load_common_assets()` enqueues `newspack-commons` JS *and* the `newspack-commons` CSS handle — the CSS depends on `wp-components`, pulling the editor-oriented stylesheet into every frontend page that contains an overlay-search block. The view script does need the webpack `commons` chunk (split-chunk dependency), but the CSS side is almost certainly dead weight on the frontend.

Inherited from overlay-menu, but this PR doubles the surface where it ships. Worth a follow-up to split frontend JS-only commons from full commons.

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+			: 'newspack-overlay-search__label';
+
+		ob_start();
+		?>
+		<div class="wp-block-buttons is-layout-flex">
+			<div class="wp-block-button">
+				<a <?php echo $wrapper_attributes; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>>
+					<?php if ( ! $is_text_only ) : ?>
+						<span class="newspack-overlay-search__icon" aria-hidden="true">
+							<?php echo \Newspack\Newspack_UI_Icons::get_svg( 'search' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
```
chickenn00dle: `Newspack_UI_Icons::print_svg()` is the prevailing pattern in this codebase (~75 call sites) and runs SVGs through `wp_kses` with an SVG allowlist, which would let you drop the `phpcs:ignore` annotation. Same applies on lines 167 and 226, and on `overlay-menu/panel/class-overlay-menu-panel-block.php:202` / `overlay-menu/trigger/class-overlay-menu-trigger-block.php:466`.

The icons are static class constants so it's a security wash — just a cleanliness / consistency-with-rest-of-codebase suggestion.

src/blocks/overlay-search/edit.js
```diff
@@ -0,0 +1,102 @@
...
+	__experimentalUseColorProps as useColorProps,
+	__experimentalGetSpacingClassesAndStyles as useSpacingProps,
+	/* eslint-enable @wordpress/no-unsafe-wp-apis */
+} from '@wordpress/block-editor';
+import { Icon } from '@wordpress/components';
+
+export default function OverlaySearchEdit( { attributes, setAttributes, clientId } ) {
+	const { triggerText, className: blockClassName, overlayColor } = attributes;
+
+	const borderProps = useBorderProps( attributes );
```
chickenn00dle: With `supports.color` and `supports.__experimentalBorder.radius` declared in `block.json` and `useBlockProps()` already in use (apiVersion 3), the editor merges color and border classes/styles into `blockProps` automatically.

Manually pulling them via `useColorProps` / `useBorderProps` and spreading into the same `blockProps` risks double-applied inline styles via internal merge behavior of unstable APIs. Worth verifying that removing both calls (lines 20–21 imports, 30–31 hooks, and the spreads on 41–55) doesn't change the editor render, then dropping.

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+		$trigger_text = '' === trim( (string) $attributes['triggerText'] )
+			? $defaults['triggerText']
+			: $attributes['triggerText'];
+
+		$classes      = explode( ' ', (string) $attributes['className'] );
+		$is_icon_only = in_array( 'is-style-icon-only', $classes, true );
+		$is_text_only = in_array( 'is-style-text-only', $classes, true );
+
+		// `wp_unique_id()` returns a per-request counter ("1", "2", ...); prefix once for a clean DOM id.
+		$panel_id = 'newspack-overlay-search-panel-' . \wp_unique_id();
```
chickenn00dle: Two small cleanups:

1. `wp_unique_id()` accepts the prefix as an argument — `wp_unique_id( 'newspack-overlay-search-panel-' )` — which is the established Newspack pattern (see e.g. `reader-registration/index.php:99`).
2. The counter is burned even when the Jetpack branch (line 68) makes the panel unnecessary. Defer the call until after the Jetpack early return.

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+	 * @param bool   $is_text_only Whether the active style hides the icon.
+	 * @return string Trigger anchor HTML.
+	 */
+	private static function render_jetpack_trigger( string $trigger_text, bool $is_icon_only, bool $is_text_only ): string {
+		// No `aria-label` — the visible label (or the screen-reader-text span when
+		// icon-only) already supplies the accessible name.
+		$wrapper_attributes = \get_block_wrapper_attributes(
+			[
+				'class' => 'wp-element-button wp-block-button__link newspack-overlay-search__trigger jetpack-search-filter__link',
+				'href'  => \esc_url( \add_query_arg( 's', '', \home_url( '/' ) ) ),
```
chickenn00dle: `esc_url()` here is redundant — the href ends up as an attribute on the anchor returned by `get_block_wrapper_attributes`, which escapes attribute values on output. Either drop the inner `esc_url` or, if defensive belt-and-suspenders is preferred, leave a brief comment explaining the double-escape for future maintainers.

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+	private static function render_trigger_button( string $trigger_text, string $panel_id, bool $is_icon_only, bool $is_text_only ): string {
+		// No `aria-label` — the visible label (or the screen-reader-text span when
+		// icon-only) already supplies the accessible name.
+		$wrapper_attributes = \get_block_wrapper_attributes(
+			[
+				'class'           => 'wp-element-button wp-block-button__link newspack-overlay-search__trigger',
+				'type'            => 'button',
+				'aria-expanded'   => 'false',
+				'aria-controls'   => $panel_id,
+				'data-overlay-id' => $panel_id,
```
chickenn00dle: `data-overlay-id` is set on the trigger button but `view.js` only ever reads `aria-controls` to find the panel — `data-overlay-id` is never queried. Either remove it or add a comment explaining intent (e.g. parity with overlay-menu, future hook).

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+	 *
+	 * @param string $panel_id      DOM id of the panel.
+	 * @param string $overlay_color Background color of the panel (any valid CSS color, supports RGBA).
+	 * @return string Panel HTML.
+	 */
+	private static function render_panel( string $panel_id, string $overlay_color ): string {
+		// Render core/search per instance so each panel gets its own unique input id.
+		$search_html = \render_block(
+			[
+				'blockName'    => 'core/search',
```
chickenn00dle, line 190: `render_block( [ 'blockName' => 'core/search', ... ] )` re-renders a fresh core/search block per overlay-search instance. For a header with multiple instances (desktop + mobile + footer), this runs the full core block render pipeline N times per request.

Acceptable for v1 — but if profiling shows it as a hot path, consider caching the rendered form HTML once and substituting `wp_unique_id`-suffixed input IDs per instance.

src/blocks/overlay-search/class-overlay-search-block.php
```diff
@@ -0,0 +1,244 @@
...
+		ob_start();
+		?>
+		<div
+			id="<?php echo \esc_attr( $panel_id ); ?>"
+			class="newspack-overlay-search__panel"
+			role="dialog"
+			aria-modal="true"
+			aria-hidden="true"
+			inert
+			aria-label="<?php \esc_attr_e( 'Search', 'newspack-plugin' ); ?>"
```
chickenn00dle: The panel's `aria-label` is hardcoded to "Search" while the trigger uses the customizable `triggerText`. A screen-reader user with a customized trigger ("Find articles", "Browse", etc.) would hear the button labelled differently from the dialog it opens.

Consider passing the customized `$trigger_text` into the panel's `aria-label`, falling back to "Search" when empty.

src/blocks/overlay-search/view.js
```diff
@@ -0,0 +1,255 @@
...
+ * panel. When Jetpack Instant Search rendered the trigger as
+ * `.jetpack-search-filter__link`, no panel exists and this controller bails.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+
+const FOCUSABLE_SELECTOR =
```
chickenn00dle, line 15: `FOCUSABLE_SELECTOR`, `getVisibleFocusable()`, the focus-trap loop, the Esc handler, and the "return focus to opener" pattern here are duplicated almost verbatim from `overlay-menu/view.js`. With a third overlay block landing in the codebase, it might be time to extract these into something like `src/blocks/shared/overlay-controller.js`.

The behaviors are also subtly diverging — overlay-menu uses `setTimeout( ..., 50 )` for focus, overlay-search uses double `requestAnimationFrame` — both correct, but consistency matters going forward. Not a blocker for this PR; worth a follow-up issue.

chickenn00dle (commented):
Looks good @laurelfulford! Tested and everything looks good. Below are some non-blocking comments from the review skill that I thought were worth looking at. Feel free to ignore any that you feel aren't relevant (even if its all of them).

chickenn00dle (approved):

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: Thanks @chickenn00dle! I made a new task for this here to untangle after this gets merged: NPPD-1563

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: 🤦‍♀️ Thanks @chickenn00dle! Updated all instances in this block to `print_svg()` in [305953f](https://github.com/Automattic/newspack-plugin/pull/4729/commits/305953f5657c264b923555b646bd23aef52cd6b4) - I snuck in a fix for the Overlay Menu block, too, it falls outside of this PR but it's so small, it seemed safe!

src/blocks/overlay-search/edit.js
laurelfulford: Removing them doesn't affect how things look in the editor - removed in [4fcac57](https://github.com/Automattic/newspack-plugin/pull/4729/commits/4fcac5787db37090398c9a99bffd435e19892598)!

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: Tidied up in [4f8b255](https://github.com/Automattic/newspack-plugin/pull/4729/commits/4f8b255f2735e7ce6079fd63ea2d9eff6eff0b13).

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: This one seems _relatively_ low risk (a site's unlikely to have more than 2 instances of the block to cover the desktop/mobile headers but better safe than sorry! I made NPPD-1564 to investigate further!

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: Removed in [c4a2602](https://github.com/Automattic/newspack-plugin/pull/4729/commits/c4a2602658005d60c3930bd01158b93ffa6865a2)!

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: Removed in [27bd3af](https://github.com/Automattic/newspack-plugin/pull/4729/commits/27bd3af79ded6364c7c059c1a8d095b799638f77)!

src/blocks/overlay-search/class-overlay-search-block.php
laurelfulford: Updated in [db4b4e8](https://github.com/Automattic/newspack-plugin/pull/4729/commits/db4b4e81ff3c36cf9a441375488d05c85583e4b4)!

src/blocks/overlay-search/view.js
laurelfulford: Good call! Issue created here for further follow-up NPPD-1565

PR #4730, opened 2026-05-14
feat(emails): inline EmailPreview component

kmwilkerson:
## What this PR does

Adds an inline email preview component ([NPPD-1525](https://linear.app/a8c/issue/NPPD-1525)) that renders a live thumbnail of each email's HTML inside the DataViews grid built in #4727. Replaces the envelope-icon placeholder with real previews — readers see the actual rendered template at a glance.

image

## ⚠️ Stacked on #4727

**Base branch is `nppd-945-unified-emails-newspack-slice`, not `trunk`.** 

The diff in this PR shows only the NPPD-1525 work. The slice 1 changes (`#4727`) are excluded from the diff by virtue of the base branch.

## Code changes

**Backend** (`includes/wizards/newspack/class-email-preview.php`)

- New `Email_Preview` class registers `GET /wizard/newspack-settings/emails/{post_id}/preview`
- Returns the cached `EMAIL_HTML_META` post meta with sample-value token substitution applied
- Falls back to rendering the template file if no cached HTML is stored
- Substitutes 29 template tokens: reader/transaction values use stable fakes (e.g. "Sample Reader", "$25.00", "May 14, 2026"); site/branding values use real publisher values from `get_option`; action URLs use `#` anchors

**Frontend** (`src/wizards/newspack/views/settings/emails/`)

- New `EmailPreview` React component
- IntersectionObserver lazy-load via `isVisible` state bridge — cards only fetch when scrolled into view, avoiding 24 simultaneous requests on initial render
- Iframe `srcDoc` rendering with empty `sandbox=""` for security (scripts in email HTML cannot execute)
- Spinner loading state, envelope icon fallback on render error
- ResizeObserver-based proportional scaling: source iframe rendered at 1200×900, scaled via inline `transform` to match container width while preserving 4:3 aspect ratio
- Wired into `emails.tsx` to replace the envelope-icon placeholder in the Preview column

**Loader fix** (`includes/class-newspack.php`)

- Fixed missing tab indentation on the new `class-email-preview.php` include line

**Tests**

- 6 PHPUnit tests (`tests/unit-tests/email-preview.php`): substitution map, preview HTML with stored meta, template fallback, nonexistent post, wrong post type returns 404, successful API response
- 5 Jest tests (`src/wizards/newspack/views/settings/emails/email-preview.test.js`): loading state, successful iframe render, error fallback, lazy-load gating, correct endpoint path

## Manual testing

Tested locally on katie.local (Homebrew-served Apache + PHP, symlinked plugin) with the slice 1 grid view. Verified:

- ✅ Each card renders a live iframe thumbnail of the rendered email HTML
- ✅ Token substitution working (publisher name, dates, sample amounts visible in rendered previews)
- ✅ IntersectionObserver gates fetching — initial page load doesn't trigger all 24 requests
- ✅ Iframe scales proportionally as container width changes
- ✅ Fallback envelope icon renders when REST returns an error

### All Submissions:

### DIFF
--- a/includes/wizards/newspack/class-email-preview.php
+++ b/includes/wizards/newspack/class-email-preview.php
@@ -1,0 +1,297 @@
+<?php
+/**
+ * Email preview rendering for the Settings → Emails screen.
+ *
+ * Renders a publisher-facing preview of a transactional email by substituting
+ * known template tokens with realistic sample values. Used by the unified
+ * email management UI to display a per-card thumbnail.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Wizards\Newspack;
+
+use Newspack\Emails;
+use Newspack_Newsletters;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Email Preview Class.
+ */
+class Email_Preview {
+
+	/**
+	 * Get the rendered HTML for an email post, with sample token values substituted.
+	 *
+	 * Falls back to the registered template file's HTML when the post's saved
+	 * EMAIL_HTML_META is empty (i.e. the email has never been customized).
+	 *
+	 * @param int $post_id ID of the email post.
+	 *
+	 * @return string|false Rendered HTML, or false if the email can't be resolved.
+	 */
+	public static function get_preview_html( int $post_id ) {
+		if ( ! self::is_supported() ) {
+			return false;
+		}
+
+		$html = self::get_source_html( $post_id );
+		if ( empty( $html ) ) {
+			return false;
+		}
+
+		return self::apply_sample_substitutions( $html );
+	}
+
+	/**
+	 * Get the source HTML for an email post.
+	 *
+	 * Reads the saved EMAIL_HTML_META first; falls back to the registered
+	 * template's default email_html when that meta is empty.
+	 *
+	 * @param int $post_id ID of the email post.
+	 *
+	 * @return string Source HTML, or empty string if unavailable.
+	 */
+	private static function get_source_html( int $post_id ): string {
+		$html = get_post_meta( $post_id, Newspack_Newsletters::EMAIL_HTML_META, true );
+		if ( ! empty( $html ) ) {
+			return $html;
+		}
+
+		// Fallback: look up the registered template for this post type and use its default HTML.
+		$type = get_post_meta( $post_id, Emails::EMAIL_CONFIG_NAME_META, true );
+		if ( empty( $type ) ) {
+			return '';
+		}
+
+		// Trigger the template load by requesting the email config. This is the
+		// same path Emails::get_email_config_by_type() uses; we just want the
+		// raw template HTML, which is available via reflection on the loaded config.
+		$configs = apply_filters( 'newspack_email_configs', [] );
+		if ( ! isset( $configs[ $type ], $configs[ $type ]['template'] ) ) {
+			return '';
+		}
+
+		$template_data = include $configs[ $type ]['template'];
+		if ( ! is_array( $template_data ) || empty( $template_data['email_html'] ) ) {
+			return '';
+		}
+
+		return $template_data['email_html'];
+	}
+
+	/**
+	 * Apply sample-value substitutions to email HTML.
+	 *
+	 * Site/branding tokens use the publisher's real site config (so the preview
+	 * reflects their actual branding). Reader/transaction tokens use stable
+	 * fake values. Action URLs are replaced with anchor placeholders so
+	 * preview iframes don't trigger live navigation.
+	 *
+	 * @param string $html Source HTML containing *TOKEN* placeholders.
+	 *
+	 * @return string HTML with tokens substituted.
+	 */
+	private static function apply_sample_substitutions( string $html ): string {
+		foreach ( self::get_sample_substitutions() as $token => $value ) {
+			$html = str_replace( $token, $value, $html );
+		}
+		return $html;
+	}
+
+	/**
+	 * Get the substitution map of email-template tokens to sample values.
+	 *
+	 * @return array Map of `*TOKEN*` => sample value.
+	 */
+	public static function get_sample_substitutions(): array {
+		$site_logo_url   = wp_get_attachment_url( get_theme_mod( 'custom_logo' ) );
+		$site_title      = get_bloginfo( 'name' );
+		$site_url        = get_bloginfo( 'wpurl' );
+		$reply_to_email  = function_exists( 'Newspack\Emails::get_reply_to_email' ) ? Emails::get_reply_to_email() : get_bloginfo( 'admin_email' );
+		$site_address    = self::get_site_address();
+		$site_contact    = $site_address
+			? sprintf( '<strong>%s</strong> — %s', $site_title, $site_address )
+			: $site_title;
+
+		return [
+			// Site / branding — real values from the publisher's config.
+			'*SITE_TITLE*'             => $site_title,
+			'*SITE_URL*'               => $site_url,
+			'*SITE_LOGO*'              => $site_logo_url ? esc_url( $site_logo_url ) : '',
+			'*SITE_ADDRESS*'           => $site_address,
+			'*SITE_CONTACT*'           => $site_contact,
+			'*CONTACT_EMAIL*'          => sprintf( '<a href="mailto:%s">%s</a>', $reply_to_email, $reply_to_email ),
+
+			// Reader identity — stable sample values.
+			'*BILLING_FIRST_NAME*'     => 'Sample',
+			'*BILLING_LAST_NAME*'      => 'Reader',
+			'*BILLING_NAME*'           => 'Sample Reader',
+			'*PENDING_EMAIL_ADDRESS*'  => 'sample.reader@example.com',
+
+			// Transaction / subscription details — stable sample values.
+			'*AMOUNT*'                 => '$25.00',
+			'*PAYMENT_METHOD*'         => 'Visa ending in 4242',
+			'*PRODUCT_NAME*'           => 'Monthly Membership',
+			'*BILLING_FREQUENCY*'      => 'monthly',
+			'*DATE*'                  => gmdate( get_option( 'date_format', 'F j, Y' ) ),
+			'*CANCELLATION_TITLE*'     => __( 'Subscription Cancelled', 'newspack-plugin' ),
+			'*CANCELLATION_TYPE*'      => __( 'subscription', 'newspack-plugin' ),
+
+			// Action URLs — anchors so preview clicks don't navigate.
+			'*ACCOUNT_URL*'            => '#',
+			'*CANCELLATION_URL*'       => '#',
+			'*EMAIL_CANCELLATION_URL*' => '#',
+			'*EMAIL_VERIFICATION_URL*' => '#',
+			'*VERIFICATION_URL*'       => '#',
+			'*RECEIPT_URL*'            => '#',
+			'*MAGIC_LINK_URL*'         => '#',
+			'*PASSWORD_RESET_LINK*'    => '#',
+			'*SET_PASSWORD_LINK*'      => '#',
+			'*DELETION_LINK*'          => '#',
+			'*WP_LOGIN_URL*'           => '#',
+
+			// OTP code — stable sample value.
+			'*MAGIC_LINK_OTP*'         => '123456',
+		];
+	}
+
+	/**
+	 * Get the site's store address as a formatted string.
+	 *
+	 * Mirrors the logic in Emails::get_email_payload() so the preview
+	 * shows the same address format the real email would use.
+	 *
+	 * @return string Formatted site address, or empty string.
+	 */
+	private static function get_site_address(): string {
+		if ( class_exists( 'WC' ) ) {
+			$base_address  = WC()->countries->get_base_address();
+			$base_city     = WC()->countries->get_base_city();
+			$base_postcode = WC()->countries->get_base_postcode();
+		} else {
+			$base_address  = get_option( 'woocommerce_store_address', '' );
+			$base_city     = get_option( 'woocommerce_store_city', '' );
+			$base_postcode = get_option( 'woocommerce_store_postcode', '' );
+		}
+
+		if ( ! $base_address ) {
+			return '';
+		}
+
+		if ( ! $base_city && ! $base_postcode ) {
+			return $base_address;
+		}
+
+		return sprintf(
+			/* translators: 1: street address, 2: city, 3: postcode. */
+			__( '%1$s, %2$s %3$s', 'newspack-plugin' ),
+			$base_address,
+			$base_city,
+			$base_postcode
+		);
+	}
+
+	/**
+	 * Is email preview supported on this install?
+	 *
+	 * Requires Newspack Newsletters (the same dependency that gates
+	 * email management in general).
+	 *
+	 * @return bool
+	 */
+	private static function is_supported(): bool {
+		return class_exists( 'Newspack_Newsletters' );
+	}
+
+	/**
+	 * Initialize the class. Hooked from class-newspack.php inclusion.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function init(): void {
+		add_action( 'rest_api_init', [ __CLASS__, 'register_rest_routes' ] );
+	}
+
+	/**
+	 * Register the email-preview REST endpoint.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function register_rest_routes(): void {
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'wizard/newspack-settings/emails/(?P<post_id>\d+)/preview',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => [ __CLASS__, 'api_get_preview' ],
+				'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
+				'args'                => [
+					'post_id' => [
+						'required'          => true,
+						'type'              => 'integer',
+						'sanitize_callback' => 'absint',
+					],
+				],
+			]
+		);
+	}
+
+	/**
+	 * REST handler: return preview HTML for an email post.
+	 *
+	 * @param \WP_REST_Request $request Request object.
+	 *
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public static function api_get_preview( $request ) {
+		$post_id = (int) $request->get_param( 'post_id' );
+
+		$post = get_post( $post_id );
+		if ( ! $post || Emails::POST_TYPE !== $post->post_type ) {
+			return new \WP_Error(
+				'newspack_email_preview_not_found',
+				__( 'Email not found.', 'newspack-plugin' ),
+				[ 'status' => 404 ]
+			);
+		}
+
+		$html = self::get_preview_html( $post_id );
+		if ( false === $html ) {
+			return new \WP_Error(
+				'newspack_email_preview_unavailable',
+				__( 'Email preview is unavailable.', 'newspack-plugin' ),
+				[ 'status' => 500 ]
+			);
+		}
+
+		return rest_ensure_response(
+			[
+				'html'    => $html,
+				'post_id' => $post_id,
+			]
+		);
+	}
+
+	/**
+	 * Permissions check for the preview endpoint. Mirrors other Newspack email endpoints.
+	 *
+	 * @codeCoverageIgnore
+	 * @return bool|\WP_Error
+	 */
+	public static function api_permissions_check() {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return new \WP_Error(
+				'newspack_rest_forbidden',
+				esc_html__( 'You cannot use this resource.', 'newspack-plugin' ),
+				[ 'status' => 403 ]
+			);
+		}
+		return true;
+	}
+}
+
+
+Email_Preview::init();
--- a/includes/wizards/newspack/class-email-preview.php
+++ b/includes/wizards/newspack/class-email-preview.php
@@ -41,7 +41,7 @@ class Email_Preview {
 			return false;
 		}
 
-		return self::apply_sample_substitutions( $html );
+		return self::apply_sample_substitutions( $html, $post_id );
 	}
 
 	/**
@@ -95,15 +95,26 @@ class Email_Preview {
 	 * fake values. Action URLs are replaced with anchor placeholders so
 	 * preview iframes don't trigger live navigation.
 	 *
-	 * @param string $html Source HTML containing *TOKEN* placeholders.
+	 * @param string $html    Source HTML containing *TOKEN* placeholders.
+	 * @param int    $post_id The email post being previewed (passed to the filter).
 	 *
 	 * @return string HTML with tokens substituted.
 	 */
-	private static function apply_sample_substitutions( string $html ): string {
-		foreach ( self::get_sample_substitutions() as $token => $value ) {
-			$html = str_replace( $token, $value, $html );
-		}
-		return $html;
+	private static function apply_sample_substitutions( string $html, int $post_id = 0 ): string {
+		$substitutions = self::get_sample_substitutions();
+
+		/**
+		 * Filters the sample substitution map used for email previews.
+		 *
+		 * Allows plugins to inject sample values for custom tokens
+		 * (e.g. group-subscription invite tokens, future token additions).
+		 *
+		 * @param array $substitutions Map of `*TOKEN*` => sample value.
+		 * @param int   $post_id       The email post being previewed (0 if unknown).
+		 */
+		$substitutions = apply_filters( 'newspack_email_preview_substitutions', $substitutions, $post_id );
+
+		return strtr( $html, $substitutions );
 	}
 
 	/**
@@ -115,7 +126,7 @@ class Email_Preview {
 		$site_logo_url   = wp_get_attachment_url( get_theme_mod( 'custom_logo' ) );
 		$site_title      = get_bloginfo( 'name' );
 		$site_url        = get_bloginfo( 'wpurl' );
-		$reply_to_email  = function_exists( 'Newspack\Emails::get_reply_to_email' ) ? Emails::get_reply_to_email() : get_bloginfo( 'admin_email' );
+		$reply_to_email  = Emails::get_reply_to_email();
 		$site_address    = self::get_site_address();
 		$site_contact    = $site_address
 			? sprintf( '<strong>%s</strong> — %s', $site_title, $site_address )
@@ -128,7 +139,7 @@ class Email_Preview {
 			'*SITE_LOGO*'              => $site_logo_url ? esc_url( $site_logo_url ) : '',
 			'*SITE_ADDRESS*'           => $site_address,
 			'*SITE_CONTACT*'           => $site_contact,
-			'*CONTACT_EMAIL*'          => sprintf( '<a href="mailto:%s">%s</a>', $reply_to_email, $reply_to_email ),
+			'*CONTACT_EMAIL*'          => sprintf( '<a href="%s">%s</a>', esc_url( 'mailto:' . $reply_to_email ), esc_html( $reply_to_email ) ),
 
 			// Reader identity — stable sample values.
 			'*BILLING_FIRST_NAME*'     => 'Sample',
@@ -141,7 +152,7 @@ class Email_Preview {
 			'*PAYMENT_METHOD*'         => 'Visa ending in 4242',
 			'*PRODUCT_NAME*'           => 'Monthly Membership',
 			'*BILLING_FREQUENCY*'      => 'monthly',
-			'*DATE*'                   => gmdate( get_option( 'date_format', 'F j, Y' ) ),
+			'*DATE*'                   => wp_date( get_option( 'date_format', 'F j, Y' ) ),
 			'*CANCELLATION_TITLE*'     => __( 'Subscription Cancelled', 'newspack-plugin' ),
 			'*CANCELLATION_TYPE*'      => __( 'subscription', 'newspack-plugin' ),
 
--- a/includes/wizards/newspack/class-email-preview.php
+++ b/includes/wizards/newspack/class-email-preview.php
@@ -74,7 +74,12 @@ class Email_Preview {
 			return '';
 		}
 
-		$template_data = include $configs[ $type ]['template'];
+		$template_path = $configs[ $type ]['template'];
+		if ( ! is_readable( $template_path ) ) {
+			return '';
+		}
+
+		$template_data = include $template_path;
 		if ( ! is_array( $template_data ) || empty( $template_data['email_html'] ) ) {
 			return '';
 		}
--- a/includes/wizards/newspack/class-email-preview.php
+++ b/includes/wizards/newspack/class-email-preview.php
@@ -136,7 +136,7 @@ class Email_Preview {
 			'*PAYMENT_METHOD*'         => 'Visa ending in 4242',
 			'*PRODUCT_NAME*'           => 'Monthly Membership',
 			'*BILLING_FREQUENCY*'      => 'monthly',
-			'*DATE*'                  => gmdate( get_option( 'date_format', 'F j, Y' ) ),
+			'*DATE*'                   => gmdate( get_option( 'date_format', 'F j, Y' ) ),
 			'*CANCELLATION_TITLE*'     => __( 'Subscription Cancelled', 'newspack-plugin' ),
 			'*CANCELLATION_TYPE*'      => __( 'subscription', 'newspack-plugin' ),
 
--- a/src/wizards/newspack/views/settings/emails/email-preview.tsx
+++ b/src/wizards/newspack/views/settings/emails/email-preview.tsx
@@ -4,13 +4,17 @@
  * Lazy-loads via IntersectionObserver: the REST fetch only fires once the
  * component scrolls into view. On success an iframe with srcDoc displays the
  * rendered HTML; on error an envelope icon placeholder is shown instead.
+ *
+ * Rendering contract mirrors NewsletterPreview in newspack-newsletters:
+ * 848 px source viewport, 1 : 1 aspect ratio, fade-in via `is-ready` class,
+ * and iframe height measured from the loaded document.
  */
 
 /**
  * WordPress dependencies.
  */
 import apiFetch from '@wordpress/api-fetch';
-import { useState, useEffect, useRef } from '@wordpress/element';
+import { useState, useEffect, useRef, useCallback } from '@wordpress/element';
 import { Spinner } from '@wordpress/components';
 import { Icon, envelope } from '@wordpress/icons';
 
@@ -23,15 +27,18 @@ interface EmailPreviewProps {
 	postId: number;
 }
 
-const IFRAME_WIDTH = 600;
+const IFRAME_WIDTH = 848;
 
 const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 	const containerRef = useRef< HTMLDivElement >( null );
+	const iframeRef = useRef< HTMLIFrameElement >( null );
 	const [ isVisible, setIsVisible ] = useState( false );
 	const [ html, setHtml ] = useState< string | null >( null );
 	const [ isLoading, setIsLoading ] = useState( false );
 	const [ hasError, setHasError ] = useState( false );
 	const [ scale, setScale ] = useState( 0 );
+	const [ iframeHeight, setIframeHeight ] = useState< number | null >( null );
+	const [ isReady, setIsReady ] = useState( false );
 
 	// Observe visibility — fetch only when the thumbnail enters the viewport.
 	useEffect( () => {
@@ -79,13 +86,17 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 		return () => ro.disconnect();
 	}, [] );
 
-	// Fetch preview HTML once visible.
+	// Fetch preview HTML once visible. Reset state on postId change.
 	useEffect( () => {
 		if ( ! isVisible ) {
 			return;
 		}
 
 		setIsLoading( true );
+		setIsReady( false );
+		setIframeHeight( null );
+		setHasError( false );
+		setHtml( null );
 		apiFetch< { html: string; post_id: number } >( {
 			path: `/newspack/v1/wizard/newspack-settings/emails/${ postId }/preview`,
 		} )
@@ -100,9 +111,42 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 			} );
 	}, [ isVisible, postId ] );
 
+	// Handle iframe load: wait for stylesheets and images, then measure height and reveal.
+	const handleIframeLoad = useCallback( () => {
+		const doc = iframeRef.current?.contentDocument;
+		if ( ! doc ) {
+			return;
+		}
+
+		const awaitLoad = ( el: HTMLLinkElement | HTMLImageElement ) =>
+			new Promise< void >( resolve => {
+				el.addEventListener( 'load', () => resolve(), { once: true } );
+				el.addEventListener( 'error', () => resolve(), { once: true } );
+			} );
+
+		const linkPromises = Array.from( doc.querySelectorAll< HTMLLinkElement >( 'link[rel="stylesheet"]' ) )
+			.filter( link => ! link.sheet )
+			.map( awaitLoad );
+		const imgPromises = Array.from( doc.querySelectorAll< HTMLImageElement >( 'img' ) )
+			.filter( img => ! img.complete )
+			.map( awaitLoad );
+
+		// 8 s safety so a slow asset never strands the spinner.
+		const safety = setTimeout( () => {
+			setIframeHeight( doc.body.scrollHeight );
+			setIsReady( true );
+		}, 8000 );
+
+		Promise.all( [ ...linkPromises, ...imgPromises ] ).then( () => {
+			clearTimeout( safety );
+			setIframeHeight( doc.body.scrollHeight );
+			setIsReady( true );
+		} );
+	}, [] );
+
 	return (
-		<div ref={ containerRef } className="newspack-email-preview">
-			{ isLoading && (
+		<div ref={ containerRef } className={ `newspack-email-preview${ isReady ? ' is-ready' : '' }` }>
+			{ ( isLoading || ( html && ! isReady ) ) && ! hasError && (
 				<div className="newspack-email-preview__placeholder">
 					<Spinner />
 				</div>
@@ -112,14 +156,19 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 					<Icon icon={ envelope } size={ 48 } />
 				</div>
 			) }
-			{ html && ! hasError && ! isLoading && scale > 0 && (
+			{ html && ! hasError && scale > 0 && (
 				<iframe
+					ref={ iframeRef }
 					className="newspack-email-preview__iframe"
 					srcDoc={ html }
 					sandbox=""
 					tabIndex={ -1 }
 					title="Email preview"
-					style={ { transform: `scale(${ scale })` } }
+					onLoad={ handleIframeLoad }
+					style={ {
+						transform: `scale(${ scale })`,
+						height: iframeHeight ? `${ iframeHeight }px` : undefined,
+					} }
 				/>
 			) }
 		</div>
--- a/src/wizards/newspack/views/settings/emails/email-preview.tsx
+++ b/src/wizards/newspack/views/settings/emails/email-preview.tsx
@@ -44,7 +44,7 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 		}
 
 		const observer = new IntersectionObserver(
-			( entries ) => {
+			entries => {
 				if ( entries[ 0 ]?.isIntersecting ) {
 					setIsVisible( true );
 				}
--- a/src/wizards/newspack/views/settings/emails/email-preview.tsx
+++ b/src/wizards/newspack/views/settings/emails/email-preview.tsx
@@ -35,16 +35,18 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 
 	// Observe visibility — fetch only when the thumbnail enters the viewport.
 	useEffect( () => {
+		if ( isVisible ) {
+			return;
+		}
 		const node = containerRef.current;
 		if ( ! node ) {
 			return;
 		}
 
 		const observer = new IntersectionObserver(
-			( [ entry ] ) => {
-				if ( entry.isIntersecting ) {
+			( entries ) => {
+				if ( entries[ 0 ]?.isIntersecting ) {
 					setIsVisible( true );
-					observer.disconnect();
 				}
 			},
 			{ rootMargin: '200px' }
@@ -52,7 +54,7 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 
 		observer.observe( node );
 		return () => observer.disconnect();
-	}, [] );
+	}, [ isVisible ] );
 
 	// Measure container width and compute iframe scale.
 	useEffect( () => {
--- a/src/wizards/newspack/views/settings/emails/email-preview.tsx
+++ b/src/wizards/newspack/views/settings/emails/email-preview.tsx
@@ -23,12 +23,15 @@ interface EmailPreviewProps {
 	postId: number;
 }
 
+const IFRAME_WIDTH = 600;
+
 const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 	const containerRef = useRef< HTMLDivElement >( null );
 	const [ isVisible, setIsVisible ] = useState( false );
 	const [ html, setHtml ] = useState< string | null >( null );
 	const [ isLoading, setIsLoading ] = useState( false );
 	const [ hasError, setHasError ] = useState( false );
+	const [ scale, setScale ] = useState( 0 );
 
 	// Observe visibility — fetch only when the thumbnail enters the viewport.
 	useEffect( () => {
@@ -49,6 +52,21 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 
 		observer.observe( node );
 		return () => observer.disconnect();
+	}, [] );
+
+	// Measure container width and compute iframe scale.
+	useEffect( () => {
+		const node = containerRef.current;
+		if ( ! node ) {
+			return;
+		}
+
+		const ro = new ResizeObserver( ( [ entry ] ) => {
+			setScale( entry.contentRect.width / IFRAME_WIDTH );
+		} );
+
+		ro.observe( node );
+		return () => ro.disconnect();
 	}, [] );
 
 	// Fetch preview HTML once visible.
@@ -84,8 +102,15 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 					<Icon icon={ envelope } size={ 48 } />
 				</div>
 			) }
-			{ html && ! hasError && ! isLoading && (
-				<iframe className="newspack-email-preview__iframe" srcDoc={ html } sandbox="" tabIndex={ -1 } title="Email preview" />
+			{ html && ! hasError && ! isLoading && scale > 0 && (
+				<iframe
+					className="newspack-email-preview__iframe"
+					srcDoc={ html }
+					sandbox=""
+					tabIndex={ -1 }
+					title="Email preview"
+					style={ { transform: `scale(${ scale })` } }
+				/>
 			) }
 		</div>
 	);
--- a/src/wizards/newspack/views/settings/emails/email-preview.tsx
+++ b/src/wizards/newspack/views/settings/emails/email-preview.tsx
@@ -161,7 +161,7 @@ const EmailPreview: React.FC< EmailPreviewProps > = ( { postId } ) => {
 					ref={ iframeRef }
 					className="newspack-email-preview__iframe"
 					srcDoc={ html }
-					sandbox=""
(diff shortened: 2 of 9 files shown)


thomasguillot (commented):
Nice work on the lazy-load + scaling plumbing. Before this lands, please align the rendering contract with our existing `NewsletterPreview` in newspack-newsletters so the two grids feel like the same surface side-by-side. **These are separate plugins — copy the relevant logic into newspack-plugin rather than extracting a shared package.** Use newspack-newsletters as a read-only reference for the shape.

### Reference (read-only, in `newspack-newsletters`)

- Component: `src/components/newsletter-preview/index.js` — see lines 43–215, particularly the `useInlineStyles` hook's iframe-readiness detection (lines 70–193) and the `is-ready`/fade-in pattern (lines 195–214).
- Styles: `src/components/newsletter-preview/style.scss` — note the `.block-editor-block-preview__container { opacity: 0; transition: opacity 200ms ease-out; }` → `.is-ready &` fade.
- Grid wrapper: `src/admin-shell/screens/layouts-list/fields.js` `PreviewCard` (lines 197–242) — uses `aspectRatio: '1'`, `viewportWidth={ 848 }`, and `rootMargin: "200px"`.
- Lazy helper: `src/admin-shell/screens/layouts-list/lazy-preview.js`.

### Concrete changes to make in this PR

All paths below are in `newspack-plugin`.

**1. Aspect ratio: 1 : 1**

`src/wizards/newspack/views/settings/emails/email-preview.scss`

```diff
-.newspack-email-preview {
-    width: 100%;
-    aspect-ratio: 4 / 3;
-    overflow: hidden;
-    ...
-    background: #f6f7f7;
+.newspack-email-preview {
+    width: 100%;
+    aspect-ratio: 1;
+    overflow: hidden;
+    ...
+    background: transparent;
```

Rationale: newsletters layouts grid uses 1 : 1 (`<LazyPreview placeholderStyle={{ aspectRatio: '1' }} />`). Cards sit side-by-side in the same DataView shell, so matching makes them visually coherent. Transparent background lets the card chrome show through consistently.

**2. Source viewport: 848 px wide, intrinsic height**

`src/wizards/newspack/views/settings/emails/email-preview.tsx`

Replace the `IFRAME_WIDTH = 600` constant and the fixed-size iframe with **width 848, height measured from the loaded document**. Concretely:

```tsx
const IFRAME_WIDTH = 848; // matches NewsletterPreview's `viewportWidth={ 848 }`
```

Update the SCSS so `__iframe` no longer hardcodes height:

```scss
&__iframe {
    width: 848px;
    height: auto; // set dynamically after load
    border: 0;
    position: absolute;
    top: 0;
    left: 0;
    transform-origin: top left;
    pointer-events: none;
}
```

Then measure the natural document height after the iframe loads and set it inline. Add an `onLoad` handler:

```tsx
const iframeRef = useRef<HTMLIFrameElement>( null );
const [ iframeHeight, setIframeHeight ] = useState< number | null >( null );
const [ isReady, setIsReady ] = useState( false );

const handleIframeLoad = () => {
    const doc = iframeRef.current?.contentDocument;
    if ( ! doc ) {
        return;
    }
    // Mirror NewsletterPreview's readiness detection: wait for stylesheets
    // and images before measuring + revealing, otherwise the height we read
    // is mid-layout and the iframe pops in unstyled.
    const awaitLoad = ( el: HTMLLinkElement | HTMLImageElement ) =>
        new Promise< void >( resolve => {
            el.addEventListener( 'load', () => resolve(), { once: true } );
            el.addEventListener( 'error', () => resolve(), { once: true } );
        } );
    const linkPromises = Array.from(
        doc.querySelectorAll< HTMLLinkElement >( 'link[rel="stylesheet"]' )
    )
        .filter( link => ! link.sheet )
        .map( awaitLoad );
    const imgPromises = Array.from( doc.querySelectorAll< HTMLImageElement >( 'img' ) )
        .filter( img => ! img.complete )
        .map( awaitLoad );
    // 8s safety so a slow asset never strands the spinner — matches NewsletterPreview line 81.
    const safety = setTimeout( () => {
        setIframeHeight( doc.body.scrollHeight );
        setIsReady( true );
    }, 8000 );
    Promise.all( [ ...linkPromises, ...imgPromises ] ).then( () => {
        clearTimeout( safety );
        setIframeHeight( doc.body.scrollHeight );
        setIsReady( true );
    } );
};
```

And on the `<iframe>` element:

```tsx
<iframe
    ref={ iframeRef }
    className="newspack-email-preview__iframe"
    srcDoc={ html }
    sandbox=""
    tabIndex={ -1 }
    title="Email preview"
    onLoad={ handleIframeLoad }
    style={ {
        transform: `scale(${ scale })`,
        height: iframeHeight ? `${ iframeHeight }px` : undefined,
    } }
/>
```

Scale stays `containerWidth / IFRAME_WIDTH` — but since `IFRAME_WIDTH` is now 848, the math changes accordingly.

Also: the PR description says **1200 × 900** but the SCSS as written is **600 × 1800**. Please remove that line from the PR description once 848 × auto is the actual value.

**3. Ready-state fade-in**

Add the `is-ready` class pattern from NewsletterPreview.

Component:

```tsx
<div
    ref={ containerRef }
    className={ `newspack-email-preview${ isReady ? ' is-ready' : '' }` }
>
```

SCSS:

```scss
.newspack-email-preview {
    ...
    &__iframe {
        ...
        opacity: 0;
        transition: opacity 200ms ease-out;
    }

    &.is-ready &__iframe {
        opacity: 1;
    }
}
```

Keep the existing spinner placeholder mounted while `! isReady` so the transition is spinner → fade-in iframe (no white flash).

**4. Render gating**

The current condition is `html && ! hasError && ! isLoading && scale > 0`. After the changes above, also render the iframe as soon as `html` arrives (so `onLoad` can fire and measure) — the **iframe itself** is hidden by `opacity: 0` until `isReady` flips, and the spinner placeholder stays mounted in the meantime. Concretely:

```tsx
{ html && ! hasError && scale > 0 && (
    <iframe ... />
) }
{ ( isLoading || ( html && ! isReady ) ) && ! hasError && (
    <div className="newspack-email-preview__placeholder">
        <Spinner />
    </div>
) }
```

**5. Iframe-state reset on `postId` change**

When `postId` changes, reset `html`, `isReady`, `iframeHeight`, `hasError` so the next preview doesn't show stale content while loading. Add to the fetch effect:

```tsx
useEffect( () => {
    if ( ! isVisible ) {
        return;
    }
    setIsLoading( true );
    setIsReady( false );
    setIframeHeight( null );
    setHasError( false );
    setHtml( null );
    apiFetch...
}, [ isVisible, postId ] );
```

**6. Tests**

Update `src/wizards/newspack/views/settings/emails/email-preview.test.js`:

- The "renders iframe on successful fetch" test will need to simulate `onLoad`. After `srcdoc` is set, fire `iframe.dispatchEvent(new Event('load'))` and assert the container gains the `is-ready` class.
- Stub `iframe.contentDocument` to return an object with `querySelectorAll` returning `[]` and `body.scrollHeight` returning a number, so `handleIframeLoad` resolves immediately.
- Add a test for the reset-on-postId-change behaviour.

### Out of scope (don't do)

- Don't extract a shared component into a package — these are separate plugins, duplication is fine.
- Don't honour `meta.background_color` / fonts here (system emails don't carry that meta; the newsletters version reads it from layout meta).
- Don't reuse `LazyPreview` from newsletters — your inline `IntersectionObserver` already matches its `rootMargin: '200px'` contract, which is what matters.

### Acceptance check

When done, an Email card and a Newsletter Layout card placed side-by-side should:

- Be the same shape (1 : 1).
- Show a comparable amount of vertical content (header + first content block roughly fills the square).
- Fade in from spinner with no FOUC.
- Show transparent background behind the rendered HTML.

includes/wizards/newspack/class-email-preview.php
```diff
@@ -0,0 +1,297 @@
...
+
+		// Trigger the template load by requesting the email config. This is the
+		// same path Emails::get_email_config_by_type() uses; we just want the
+		// raw template HTML, which is available via reflection on the loaded config.
+		$configs = apply_filters( 'newspack_email_configs', [] );
+		if ( ! isset( $configs[ $type ], $configs[ $type ]['template'] ) ) {
+			return '';
+		}
+
+		$template_data = include $configs[ $type ]['template'];
```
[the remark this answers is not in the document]
kmwilkerson: Addressed in 92d6a42. Added `is_readable()` guard before the template include.

src/wizards/newspack/views/settings/emails/email-preview.tsx
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		const observer = new IntersectionObserver(
+			entries => {
+				if ( entries[ 0 ]?.isIntersecting ) {
+					setIsVisible( true );
+				}
+			},
+			{ rootMargin: '200px' }
+		);
```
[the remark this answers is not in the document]
kmwilkerson: Addressed in 92d6a42. Added `typeof` guards for both `IntersectionObserver` and `ResizeObserver` — falls back to eager loading and scale=1 respectively.

kmwilkerson:
Addressed in f889cd6. All six items from the review:

1. **Aspect ratio 1:1** — `aspect-ratio: 1` in SCSS, background set to `transparent`
2. **Source viewport 848px** — `IFRAME_WIDTH = 848`, iframe width 848px in SCSS with `height: auto`
3. **Ready-state fade-in** — `is-ready` class on container, iframe starts at `opacity: 0` with `transition: opacity 200ms ease-out`, revealed when `is-ready` is set
4. **Render gating** — iframe mounts as soon as HTML arrives (hidden via opacity), spinner stays until `isReady`
5. **Iframe height measurement** — `onLoad` handler waits for stylesheets + images via Promise.all, measures `contentDocument.body.scrollHeight`, 8s safety timeout
6. **State reset on postId change** — resets `html`, `isReady`, `iframeHeight`, `hasError` at the top of the fetch effect

Tests updated: `is-ready` class assertion after simulated `onLoad`, `contentDocument` stub, and reset-on-postId-change test. Build and 217 tests pass.

thomasguillot:
## Code review — EmailPreview component (slice 2 of unified emails)

Reviewed at `2d6fe465b` against base branch `nppd-945-unified-emails-newspack-slice` (PR #4727 head `ebb58fbed`). Structured handoff for the agent that will pick this up — each finding includes file, line, severity, why, and remediation. There is **one real bug** (finding 1), one security trade-off worth a comment in code (finding 2), and a handful of resilience / cleanup items.

---

### Bugs

#### 1. `function_exists()` cannot test a `Class::method` — `Emails::get_reply_to_email()` is never called
`includes/wizards/newspack/class-email-preview.php:118`

```php
$reply_to_email = function_exists( 'Newspack\Emails::get_reply_to_email' )
    ? Emails::get_reply_to_email()
    : get_bloginfo( 'admin_email' );
```

`function_exists()` only tests global functions. Verified locally:

```
php -r 'var_dump(function_exists("DateTime::format"));'   // bool(false)
php -r 'var_dump(method_exists("DateTime", "format"));'    // bool(true)
```

So the conditional is **always false** — `Emails::get_reply_to_email()` is unreachable, and `*CONTACT_EMAIL*` always falls back to `admin_email`. On sites where Reader Activation sets a `contact_email_address` option (which is exactly the *publisher-facing* reply-to behaviour previews should show), the preview will display the wrong address relative to the real outbound email.

Replace with `method_exists`:
```php
$reply_to_email = method_exists( Emails::class, 'get_reply_to_email' )
    ? Emails::get_reply_to_email()
    : get_bloginfo( 'admin_email' );
```
Even simpler: `Emails` is `use`-imported at the top of the file and lives in the same plugin, so the guard is unnecessary. Just call `Emails::get_reply_to_email()` directly.

Pin it with a PHPUnit case: set the `newspack_ras_contact_email_address` option (or filter `newspack_reply_to_email`), generate a preview from a template containing `*CONTACT_EMAIL*`, and assert the rendered HTML contains the configured address rather than `admin_email`. The current substitution-map test would not have caught this because it doesn't render against a real template.

---

### Security

#### 2. `sandbox="allow-same-origin"` weakens the iframe sandbox — document the trade-off
`src/wizards/newspack/views/settings/emails/email-preview.tsx:164`

The commit message (`2d6fe465b`) explains why the change was made: without `allow-same-origin`, `iframe.contentDocument` is `null`, so the `handleIframeLoad` measurement logic can't read `body.scrollHeight` or query stylesheets. That's correct — same-origin is required for the measurement contract borrowed from `NewsletterPreview`.

But the security analysis isn't in the code. The relevant facts:
- `allow-scripts` is **not** in the sandbox attribute, so scripts in the email HTML cannot execute. Good.
- `allow-same-origin` alone (without `allow-scripts`) does not let the iframe steal cookies or `localStorage` because it can't run JS.
- But **`<form action="…">` submissions still work without scripts**, and with `allow-same-origin` they submit from the same origin (i.e. with the admin's cookies). Email HTML rarely contains forms, but the preview reads HTML from `EMAIL_HTML_META`, which is publisher-editable. A malicious editor (or a compromised plugin writing to that meta) could embed a form pointing at an admin endpoint and trigger it on preview render.

Realistic risk: low — this is admin-only code and the attacker would already need write access to email post meta. But the safest hardening that preserves the measurement contract is to inject a `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src *; style-src 'unsafe-inline' *; font-src *;">` into the srcDoc, which blocks form action / link navigation from the iframe entirely. This is also where you'd want `<base target="_blank">` in case any anchor escapes the sandbox.

Minimum bar: add a comment above the `sandbox` attribute explaining (a) why `allow-same-origin` is required, (b) why scripts are still blocked, and (c) the residual form-action risk. The next reader should not need this review thread to reach the same conclusion.

---

### Resilience / lifecycle

#### 3. In-flight fetch isn't cancelled on `postId` change or unmount
`src/wizards/newspack/views/settings/emails/email-preview.tsx:90-112`

```tsx
useEffect( () => {
    if ( ! isVisible ) return;
    setIsLoading( true );
    // ...resets...
    apiFetch( { /* ... */ } )
        .then( response => setHtml( response.html ) )
        .catch( () => setHasError( true ) )
        .finally( () => setIsLoading( false ) );
}, [ isVisible, postId ] );
```

If `postId` changes (DataViews re-uses rows when scrolling) or the component unmounts, the old promise still resolves and writes to state. The state-reset at the top of the new run happens *before* the next fetch, but the OLD `then` callback can still fire **after** the new fetch's `then`, leaving the iframe showing stale HTML.

Fix:
```tsx
useEffect( () => {
    if ( ! isVisible ) return;
    let cancelled = false;
    setIsLoading( true );
    setIsReady( false );
    setIframeHeight( null );
    setHasError( false );
    setHtml( null );
    apiFetch< { html: string; post_id: number } >( {
        path: `/newspack/v1/wizard/newspack-settings/emails/${ postId }/preview`,
    } )
        .then( response => { if ( ! cancelled ) setHtml( response.html ); } )
        .catch( () => { if ( ! cancelled ) setHasError( true ); } )
        .finally( () => { if ( ! cancelled ) setIsLoading( false ); } );
    return () => { cancelled = true; };
}, [ isVisible, postId ] );
```

Same `cancelled` pattern already exists in `NewsletterPreview` (`repos/newspack-newsletters/src/components/newsletter-preview/index.js:80`) — worth aligning.

#### 4. Safety timeout in `handleIframeLoad` is not cleared on unmount or postId change
`src/wizards/newspack/views/settings/emails/email-preview.tsx:135-144`

```tsx
const safety = setTimeout( () => { setIframeHeight( ... ); setIsReady( true ); }, 8000 );
Promise.all( [...] ).then( () => { clearTimeout( safety ); /* ... */ } );
```

`clearTimeout( safety )` only fires when Promise.all resolves. If the user scrolls the row out of view, switches tabs, or the component re-mounts before 8 seconds, the timer still ticks and calls `setIframeHeight`/`setIsReady` on a stale render. React will warn about state updates on an unmounted component (or silently overwrite state on the same postId, depending on timing).

Hoist the timeout into a ref or attach it via an effect with a cleanup:
```tsx
const handleIframeLoad = useCallback( () => {
    // ... build promises ...
    const safetyId = setTimeout( finalize, 8000 );
    Promise.all( [...] ).then( () => {
        clearTimeout( safetyId );
        finalize();
    } );
    // No way to clean from here — either ref it, or track `cancelled` via the parent effect.
}, [] );
```

Cleaner: drive the load tracking from the same `useEffect` that fetches HTML, so a single `cancelled` flag covers both fetch and post-load measurement.

#### 5. `Promise.all().then()` runs even after the safety timeout fires — redundant state writes
Same block as finding (4). If Promise.all resolves after 8 s (slow asset), both code paths fire `setIframeHeight` and `setIsReady`. React batches the second call, but it's still wasted work and obscures intent. The fix from (4) (single cancellation source) resolves this too.

#### 6. No `onError` handler on the iframe element
`src/wizards/newspack/views/settings/emails/email-preview.tsx:160-172`

Network-level iframe failures (rare with `srcDoc`, but not impossible) leave the spinner spinning until the 8 s safety. Add `onError={ () => setHasError( true ) }`.

#### 7. Initial paint shows nothing for a frame
`src/wizards/newspack/views/settings/emails/email-preview.tsx:39, 159`

```tsx
const [ scale, setScale ] = useState( 0 );
// ...
{ html && ! hasError && scale > 0 && <iframe ... /> }
```

`scale === 0` until ResizeObserver fires (one tick after mount). For that tick, neither the spinner nor the iframe is rendered (the spinner is gated by `isLoading || (html && !isReady)`, and right after fetch resolves both flags can be false simultaneously for a frame). Cleanest fix: initialise `scale` to `null` and use `scale !== null && scale > 0` in the guard, OR always render the spinner placeholder when `! isReady`.

---

### PHP — quality & correctness

#### 8. `*CONTACT_EMAIL*` substitution embeds raw `$reply_to_email` into HTML
`includes/wizards/newspack/class-email-preview.php:131`

```php
'*CONTACT_EMAIL*' => sprintf( '<a href="mailto:%s">%s</a>', $reply_to_email, $reply_to_email ),
```

This mirrors `Emails::get_email_payload()` (`includes/emails/class-emails.php:215-218`), which has the same issue. If the reply-to address ever contains characters that need escaping (it shouldn't for a sanitised email, but defence in depth), this becomes an HTML/URL injection vector. Recommend:
```php
'*CONTACT_EMAIL*' => sprintf(
    '<a href="%s">%s</a>',
    esc_url( 'mailto:' . $reply_to_email ),
    esc_html( $reply_to_email )
),
```
Calling out the parallel bug in `class-emails.php` so this slice can either fix both or file a follow-up.

#### 9. `*DATE*` uses `gmdate()` rather than site time
`includes/wizards/newspack/class-email-preview.php:144`

```php
'*DATE*' => gmdate( get_option( 'date_format', 'F j, Y' ) ),
```

Real transactional emails use the publisher's timezone (`wp_date()` or `date_i18n()`). For a preview, a UTC date may show "May 14" while the publisher's locale shows "May 15". Use `wp_date( get_option( 'date_format', 'F j, Y' ) )` — same call signature, but applies the site timezone and locale.

#### 10. Template fallback uses raw `include`
`includes/wizards/newspack/class-email-preview.php:82`

```php
$template_data = include $template_path;
```

Email templates today return arrays, but `include` runs arbitrary PHP. If a template file ever runs side-effecting code (action hooks, option writes), the preview request will trigger it. The `is_readable()` guard added in commit `92d6a42` protects against missing files but not against malicious or stateful templates.

Mitigations:
- Tighten the template path to a known safe directory: `if ( 0 !== strpos( realpath( $template_path ), realpath( NEWSPACK_ABSPATH . 'includes/templates' ) ) ) return '';`.
- Or: load the template via `Emails::get_email_config_by_type()` instead of `include`-ing the path yourself. The class already has the loading semantics worked out (`class-emails.php:340 load_email_template`).

#### 11. No caching — every grid render does 12-24 substitution passes per preview
`includes/wizards/newspack/class-email-preview.php:103-107`

```php
foreach ( self::get_sample_substitutions() as $token => $value ) {
    $html = str_replace( $token, $value, $html );
}
```

29 `str_replace` passes over the HTML, called per `post_id`, with no transient cache. For the visible grid (12-24 cards depending on viewport), that's 12-24 × 29 passes per page view. On a publisher's slow shared host this adds up.

Two cheap wins:
1. Use `strtr( $html, $map )` — single pass over the string instead of N passes:
   ```php
   return strtr( $html, self::get_sample_substitutions() );
   ```
2. Cache per `post_id` keyed on `post_modified_gmt` of the email post (transient or `wp_cache_*`). Invalidates automatically when the publisher edits the email.

#### 12. Substitution map has no filter hook
`includes/wizards/newspack/class-email-preview.php:114-164`

`get_sample_substitutions()` is `public static`, which suggests it's extension-ready, but there's no `apply_filters( 'newspack_email_preview_substitutions', $substitutions, $post_id )` to let other plugins inject sample values for their own tokens (e.g. group-subscription invite tokens, future tokens from PR follow-ups). Without this, every new token will require editing this file.

Add the filter, and have `apply_sample_substitutions()` accept `$post_id` so the filter can vary by email type.

#### 13. `*SITE_ADDRESS*` / `*SITE_CONTACT*` logic duplicates `Emails::get_email_payload`
`includes/wizards/newspack/class-email-preview.php:174-200` vs `includes/emails/class-emails.php:186-211`

The address-formatting logic is copy-pasted from `Emails::get_email_payload`. The docblock acknowledges this ("Mirrors the logic in Emails::get_email_payload()") but doesn't reference the line range, so the next refactor of either will silently diverge. Two options:
- Extract `Emails::get_site_address(): string` (or similar) as a public static helper and call it from both.
- At minimum, expand the docblock to "Mirror of `Emails::get_email_payload()` lines 186-211; keep in sync."

#### 14. `function_exists`/`class_exists` checks for `Newspack_Newsletters`
`includes/wizards/newspack/class-email-preview.php:210-212, 14-15`

```php
use Newspack_Newsletters;
// ...
return class_exists( 'Newspack_Newsletters' );
```

Importing a class via `use` doesn't trigger autoloading, so the `class_exists` check works. But the file unconditionally reads `Newspack_Newsletters::EMAIL_HTML_META` at line 58 — if `Newspack_Newsletters` is missing, `get_source_html()` runs before `is_supported()` is called from `get_preview_html()`, and the constant access would fail.

Trace: `get_preview_html()` (line 34) → `is_supported()` (line 35) returns false → returns false. **Before** reaching `get_source_html()`. So the order is safe.

But `get_source_html()` is `private static` and only called from `get_preview_html()` which guards, so it's defensive enough. Worth a one-line comment saying so.

#### 15. `*MAGIC_LINK_OTP*` is a single fixed string
`includes/wizards/newspack/class-email-preview.php:162`

`'*MAGIC_LINK_OTP*' => '123456'` — a 6-digit fixed sample. If `*MAGIC_LINK_OTP*` ever lives in HTML that uses character-count constraints (very tight inline styles), `123456` is a reasonable typical width. No issue, but worth mentioning that if the real OTP format changes (e.g. 8 chars), this sample needs to track it.

---

### Tests

#### 16. The PHP tests don't cover the `function_exists` bug surface
`tests/unit-tests/email-preview.php:69-90`

The substitution-map test asserts keys exist; it doesn't verify that `*CONTACT_EMAIL*`'s value resolves through `Emails::get_reply_to_email()` when the option is set. Add a case that:
1. Sets `update_option( 'newspack_ras_contact_email_address', 'reply@example.org' )` (or uses the filter).
2. Generates a preview of a template containing `*CONTACT_EMAIL*`.
3. Asserts the rendered HTML contains `reply@example.org`, not `admin_email`.

This pins finding (1).

#### 17. Permission-check path isn't asserted
`tests/unit-tests/email-preview.php`

`api_permissions_check()` is annotated `@codeCoverageIgnore`, but a negative test (non-admin user → 403) round-trips the security contract. Two lines:
```php
public function test_api_permissions_check_rejects_non_admin() {
    wp_set_current_user( self::factory()->user->create( [ 'role' => 'subscriber' ] ) );
    $result = Email_Preview::api_permissions_check();
    self::assertInstanceOf( 'WP_Error', $result );
}
```

#### 18. 404 test should assert the HTTP status, not just the error code
`tests/unit-tests/email-preview.php:133-143`

```php
self::assertEquals( 'newspack_email_preview_not_found', $response->get_error_code() );
```

Add:
```php
self::assertEquals( 404, $response->get_error_data()['status'] );
```

#### 19. Jest tests don't cover the race condition (finding 3) or the safety-timeout (finding 4)
`src/wizards/newspack/views/settings/emails/email-preview.test.js`

Specifically missing:
- `postId` changes mid-fetch: old fetch resolves *after* new fetch. Assert the displayed `srcDoc` reflects the *new* postId. Easiest setup: two `mockImplementationOnce` returning controlled promises; resolve them out of order.
- 8 s safety timeout: mock `setTimeout`, advance Jest's fake timers, assert `is-ready` flips even when no load event fires.
- iframe `onError` (after finding 6 is added).

#### 20. `simulateIframeLoad` stubs `contentDocument.querySelectorAll` to always return `[]`
`src/wizards/newspack/views/settings/emails/email-preview.test.js:60`

That means the Promise.all path always resolves with zero promises (synchronously). The real iframe will have an asynchronous-load story. The test never exercises the link/img load-wait path. Add a test that stubs `querySelectorAll` to return a fake link with no `.sheet` and an unresolved load event, then asserts the spinner stays visible until the load event fires.

#### 21. `renders loading state while fetching` relies on Spinner's role=presentation
`src/wizards/newspack/views/settings/emails/email-preview.test.js:74-81`

```js
expect( screen.getByRole( 'presentation' ) ).toBeTruthy();
```

This works because `<svg role="presentation">` lives inside `<Spinner>`. If the Spinner implementation in `@wordpress/components` ever drops that role, the test breaks for the wrong reason. Prefer `expect( document.querySelector( '.newspack-email-preview__placeholder' ) ).toBeInTheDocument()`.

---

### CSS

#### 22. Placeholder colour is a hardcoded hex
`src/wizards/newspack/views/settings/emails/email-preview.scss:36`

```scss
color: #949494;
```

Per Newspack's design conventions (see `packages/colors/DEVELOPMENT.md`), admin surfaces should use WordPress core colour tokens or the Newspack colour palette. Replace with:
```scss
@use "~@wordpress/base-styles/colors" as wp-colors;
// ...
color: wp-colors.$gray-400;
```
(Or the equivalent Newspack neutral. The existing `emails.scss` already imports `wp-colors` — same pattern.)

#### 23. `pointer-events: none` on the iframe blocks user-initiated focus
`src/wizards/newspack/views/settings/emails/email-preview.scss:20`

Combined with `tabIndex={ -1 }` on the iframe element, this correctly makes the preview a non-interactive thumbnail. Good. No change.

#### 24. `transition: opacity 200ms` on the iframe means the spinner can briefly overlap
`src/wizards/newspack/views/settings/emails/email-preview.scss:22`

The container shows the spinner while `! isReady`, and the iframe is `opacity: 0` until `is-ready` is added. During the 200ms transition both can be visible. If the spinner has a solid background-coloured placeholder, no flash; if it's transparent, the iframe shows through. Worth a manual smoke test on slow connections.

---

### Repo / process

#### 25. PR diff includes a 354-line README that's unrelated to this PR
`includes/reader-activation/integrations/README.md`

The README addition comes from commit `dcb057df5 docs(integrations): add framework README (#4722)`, which is an unrelated merged PR. It shows in this diff because the base branch (`nppd-945-unified-emails-newspack-slice`) was cut before #4722 merged to trunk. When PR #4727 rebases on / merges into trunk, this file will fall out of the diff naturally.

Nothing to fix here, but flagging so the reviewer agent doesn't waste cycles auditing it. The base branch could be rebased onto current trunk to clean up, but rebasing a stacked-PR base mid-review is usually more disruptive than the noise it removes.

#### 26. PR is stacked on PR #4727 — review order matters
`baseRefName: nppd-945-unified-emails-newspack-slice`

This PR's correctness depends on:
- `EmailItem.post_id` being a valid email post ID (PR #4727's `api_get_email_settings()` contract).
- `Emails::POST_TYPE` posts existing for every row (also #4727).

If PR #4727 merges with any changes to those contracts (e.g. the duplicate-fetch consolidation tracked as NPPD-1531), the preview component may need a touchup. Keep the stack tight and merge in order.

#### 27. Preview component is not exposed via `useWizardData` store
`src/wizards/newspack/views/settings/emails/email-preview.tsx:100`

Same pattern as the slice 1 review (PR #4727 finding 5): `apiFetch` directly bypasses the `@wordpress/data` store. For 24 cards this is fine — the cardinality is bounded. But if the preview is ever reused in a wizard with paginated emails, the store would dedupe requests across cards with the same `post_id`. Note for future consolidation.

---

### Verified / non-issues
- `IntersectionObserver` and `ResizeObserver` guards correctly fall back to "eager load" and "scale=1" (`email-preview.tsx:48-51, 72-75`). Addresses Copilot finding 3247899534.
- `is_readable()` guard before `include` (`class-email-preview.php:78`) addresses Copilot finding 3247899474.
- `tabIndex={ -1 }` plus `pointer-events: none` correctly takes the iframe out of focus/click reach for screen readers and pointer users — the row is the actionable unit.
- The `is-ready` opacity-fade contract is borrowed cleanly from `NewsletterPreview` (`repos/newspack-newsletters/src/components/newsletter-preview/index.js:80-106`). Pattern is sound.
- `permission_callback` uses `manage_options` — matches every other Newspack settings endpoint.
- POST_TYPE 404 check in `api_get_preview()` (`class-email-preview.php:258`) means non-email posts can't be probed for HTML content.

---

### Suggested split for the agent picking this up

1. **Fix-up commit (must-have):** finding (1) `function_exists` bug + finding (16) test that pins it. Two-line code change plus a short test.
2. **Security-comment commit (must-have):** finding (2) — comment block above the `sandbox` attribute explaining the trade-off, optionally plus a CSP `<meta>` injected into the srcDoc.
3. **Resilience commit:** findings (3), (4), (5), (6), (7) — unify under a single `cancelled` flag in the fetch effect, clear the safety timeout, add `onError`, defend initial paint. About 30 lines.
4. **PHP polish commit:** findings (8), (9), (11), (12) — escape `mailto`, use `wp_date`, `strtr` instead of foreach `str_replace`, add the filter hook. About 15 lines.
5. **Test fill-in commit:** findings (17), (18), (19), (20), (21). About 50 lines of test code.
6. **Follow-up tickets:** (10) template-path hardening, (13) extract `Emails::get_site_address`, (27) data-store consolidation alongside NPPD-1531.

Findings (1), (2), and (3) are the ones most likely to bite real users. Everything else is polish or defensive.

kmwilkerson:
Addressed in 6b8552b37. All actionable findings from the review:

**Bugs**
- Finding 1 — `function_exists()` → call `Emails::get_reply_to_email()` directly (it's `use`-imported, guard unnecessary). Pinned with PHPUnit test that filters `newspack_reply_to_email` and asserts the preview resolves to the custom address.

**Security**
- Finding 2 — Added inline comment above `sandbox="allow-same-origin"` documenting: why it's required (contentDocument access), why scripts are blocked, and the residual form-action risk.

**Resilience**
- Finding 3 — Added `cancelled` flag to the fetch effect with cleanup function, matching the `NewsletterPreview` pattern.
- Finding 4 — Safety timeout tracked via `safetyTimerRef`, cleared in the fetch effect's cleanup and when `finalize()` runs.
- Finding 5 — Single `finalize()` function with a `finalized` guard prevents double state writes from both the Promise.all path and the safety timeout.
- Finding 6 — Added `onError={ () => setHasError( true ) }` on the iframe element.
- Finding 7 — Init `scale` to `null` (not `0`). Spinner now shows via a unified `showSpinner` variable that covers loading, fetched-but-not-ready, and the initial-paint frame.

**PHP polish**
- Finding 8 — `*CONTACT_EMAIL*` now uses `esc_url( 'mailto:' . $email )` and `esc_html( $email )`.
- Finding 9 — Replaced `gmdate()` with `wp_date()` so preview dates match the site timezone.
- Finding 11 — Replaced the foreach `str_replace` loop with a single `strtr()` call.
- Finding 12 — Added `apply_filters( 'newspack_email_preview_substitutions', $substitutions, $post_id )` so plugins can inject sample values for custom tokens.

**Tests**
- Finding 16 — PHPUnit: `test_contact_email_uses_reply_to_email()` pins the `function_exists` bug fix.
- Finding 17 — PHPUnit: `test_api_permissions_check_rejects_non_admin()` asserts subscriber gets WP_Error.
- Finding 18 — PHPUnit: 404 test now also asserts `$response->get_error_data()['status'] === 404`.
- Finding 19 — Jest: `cancelled fetch does not update state when postId changes mid-flight` — two controlled promises resolved out of order, asserts stale HTML is discarded. Safety-timeout and iframe-onError tests are noted as not testable in jsdom (auto-load behavior prevents pending-asset simulation).
- Finding 21 — Jest: Replaced `screen.getByRole('presentation')` with `document.querySelector('.newspack-email-preview__placeholder')`.

**CSS**
- Finding 22 — Replaced hardcoded `#949494` with `wp-colors.$gray-400`.

**Deferred (follow-up tickets)**
- Finding 10 — Template-path hardening (restrict `include` to known safe directory)
- Finding 13 — Extract `Emails::get_site_address()` shared helper
- Finding 27 — Data-store consolidation alongside NPPD-1531

thomasguillot:
## Re-review at `6b8552b37`

Verified the fixup against the earlier review. Reading the actual code — every actionable finding is addressed correctly, with appropriate test coverage. Deferred items are reasonable. Two small notes worth a glance.

### Verified resolved

| # | Finding | Where it landed | Notes |
|---|---|---|---|
| 1 | `function_exists()` → call `Emails::get_reply_to_email()` directly | `class-email-preview.php:129` | `$reply_to_email = Emails::get_reply_to_email();` — guard removed. Pinned by `test_contact_email_uses_reply_to_email` (`tests/unit-tests/email-preview.php:99-112`), which filters `newspack_reply_to_email` and asserts the custom address resolves into the preview HTML. |
| 2 | Document the `allow-same-origin` trade-off | `email-preview.tsx:196-201` | Multi-line comment block above the sandbox attribute explains: (a) why same-origin is needed (contentDocument access), (b) why scripts are still blocked, (c) the residual form-action risk. Reads exactly as I'd want a future reader to encounter it. |
| 3 | Cancel in-flight fetch on `postId` change / unmount | `email-preview.tsx:96-134` | `let cancelled = false;` checked in `.then` / `.catch` / `.finally`. Effect returns a cleanup that flips it. Mirrors the pattern in `NewsletterPreview`. |
| 4 | Clear safety timeout on cleanup | `email-preview.tsx:35, 99-102, 128-134, 163-166, 172` | `safetyTimerRef = useRef<…>(null)` — cleared in the fetch-effect cleanup AND inside `finalize()`. Belt and suspenders. |
| 5 | Single source of truth for finalize | `email-preview.tsx:157-169` | `let finalized = false; const finalize = () => { if ( finalized ) return; finalized = true; … }`. No double state writes. |
| 6 | `onError` on iframe | `email-preview.tsx:206` | `onError={ () => setHasError( true ) }`. |
| 7 | Initial-paint blank-frame | `email-preview.tsx:40, 177, 191` | `scale` defaults to `null`; iframe is gated by `scale !== null && scale > 0`; spinner gated by a unified `showSpinner = ! hasError && ! isReady && ( isLoading \|\| Boolean( html ) )`. Covers loading + fetched-but-not-ready + the post-ResizeObserver pre-scale frame. |
| 8 | Escape `*CONTACT_EMAIL*` | `class-email-preview.php:142` | `sprintf( '<a href="%s">%s</a>', esc_url( 'mailto:' . $reply_to_email ), esc_html( $reply_to_email ) )`. |
| 9 | `wp_date()` instead of `gmdate()` | `class-email-preview.php:155` | `wp_date( get_option( 'date_format', 'F j, Y' ) )` — site timezone + locale. |
| 11 | `strtr` instead of 29 `str_replace` passes | `class-email-preview.php:117` | Single-pass substitution. |
| 12 | Filter hook on substitutions | `class-email-preview.php:106-115` | `apply_filters( 'newspack_email_preview_substitutions', $substitutions, $post_id )` with a real docblock explaining when to use it. `apply_sample_substitutions()` signature widened to accept `$post_id` so the filter has context. |
| 16 | PHPUnit test for finding (1) | `tests/unit-tests/email-preview.php:99-112` | See finding (1) above. |
| 17 | Permission-check negative test | `tests/unit-tests/email-preview.php:155-159` | Subscriber → `WP_Error`. |
| 18 | 404 test asserts the status | `tests/unit-tests/email-preview.php:174` | `assertEquals( 404, $response->get_error_data()['status'] )` added next to the error-code assertion. |
| 19 | Race-condition Jest test | `email-preview.test.js:178-215` | Excellent — controlled `firstPromise`, second fetch resolves first, then the stale first resolves and is *not* allowed to overwrite the iframe. The exact race I described. Clean assertion: srcDoc must contain "Second email" and NOT "First email". |
| 21 | `getByRole('presentation')` → `.newspack-email-preview__placeholder` querySelector | `email-preview.test.js:80, 114` | Replaced in both the loading and error tests. No longer depends on the implementation detail of WordPress Spinner's role. |
| 22 | Hardcoded `#949494` → design token | `email-preview.scss:1, 38` | `@use "~@wordpress/base-styles/colors" as wp-colors;` import added; `color: wp-colors.$gray-400`. |

### Deferred (acknowledged in author response)
- **(10) Template-path hardening** — `include $template_path` is gated by `is_readable()` only; the attacker bar is "compromise the `newspack_email_configs` filter," which already implies PHP execution capability, so a follow-up `realpath()`-within-known-dir check is defense-in-depth rather than load-bearing. Reasonable to defer.
- **(13) Extract `Emails::get_site_address()` shared helper** — cosmetic dedupe with `Emails::get_email_payload()` lines 186-211. Drift risk exists but is small.
- **(27) Data-store consolidation alongside NPPD-1531** — same scope as the slice 1 follow-up.

### Two small new observations

#### A. `apply_sample_substitutions` `$post_id = 0` default is unreachable
`includes/wizards/newspack/class-email-preview.php:103`

```php
private static function apply_sample_substitutions( string $html, int $post_id = 0 ): string {
```

The method is `private static` and only called from `get_preview_html()` (line 44), which always passes the actual `$post_id`. The `= 0` default never fires. Tiny nit — either drop the default, or make the method `public` so the filter can be invoked standalone for ad-hoc previews. Whichever, it'd be more honest about intent.

#### B. `test_contact_email_uses_reply_to_email` cleans up via `remove_all_filters` at the end
`tests/unit-tests/email-preview.php:111`

```php
remove_all_filters( 'newspack_reply_to_email' );
```

If the assertions above it fail, the filter leaks to subsequent tests in the class — every later `Emails::get_reply_to_email()` call would still return `reply@example.org`. Belt-and-suspenders: move the removal into a `tear_down()` override, or wrap the test body in `try { … } finally { remove_all_filters( … ); }`. Minor — the test passes today.

#### C. Jest skip note for safety-timeout / iframe `onError` is the right call
`src/wizards/newspack/views/settings/emails/email-preview.test.js:217-221`

The comment at the bottom explains exactly why these two defensive paths aren't unit-tested:
> jsdom automatically fires the iframe load event when srcDoc is set, which prevents us from simulating pending-asset scenarios.

That's correct — and it's the kind of transparency a future maintainer needs. If the project ever spins up Playwright/Cypress for the wizards screen, this is the place to add the two missing scenarios. No action for this PR.

#### D. README cherry-pick still present in the diff
`includes/reader-activation/integrations/README.md`

Same as last round — falls out when PR #4727 merges. No action.

---

### Verdict

Ship-ready. The fixup is comprehensive: every behavioural fix lands cleanly, the race-condition test is genuinely good (catches the exact bug it's named for, not a proxy), and the security comment is the kind of inline rationale a future reader needs. The three deferred items are all reasonable for follow-up tickets.

Original review for the audit trail: https://github.com/Automattic/newspack-plugin/pull/4730#issuecomment-4461092203

tests/unit-tests/email-preview.php
```diff
@@ -0,0 +1,197 @@
...
+					'description' => __( 'Email for testing preview.', 'newspack' ),
+					'template'    => dirname( NEWSPACK_PLUGIN_FILE ) . '/includes/templates/reader-revenue-emails/receipt.php',
+					'category'    => 'test',
+				];
+				return $types;
+			}
+		);
+	}
+
+	/**
```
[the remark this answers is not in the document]
kmwilkerson: Addressed in 0dfbb1a — callback stored in `$this->config_filter_callback` and removed in `tear_down()`.

src/wizards/newspack/views/settings/emails/email-preview.tsx
```diff
@@ -0,0 +1,217 @@
...
+					 * cookies, but this is admin-only code and the email HTML is
+					 * publisher-controlled post meta. */
+					sandbox="allow-same-origin"
+					tabIndex={ -1 }
+					title="Email preview"
+					onLoad={ handleIframeLoad }
+					onError={ () => setHasError( true ) }
+					style={ {
+						transform: `scale(${ scale })`,
+						height: iframeHeight ? `${ iframeHeight }px` : undefined,
```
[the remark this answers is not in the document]
kmwilkerson: Already addressed in the prior review round — there's an inline comment above the sandbox attribute explaining the trade-off: `allow-same-origin` is required for `contentDocument` access (height measurement), but `allow-scripts` is NOT present so JS can't execute.

src/wizards/newspack/views/settings/emails/emails.test.js
```diff
@@ -10,6 +10,17 @@ import { render, screen, waitFor } from '@testing-library/react';
...
+global.ResizeObserver = class {
+	constructor( callback ) {
+		this.callback = callback;
+	}
+	observe() {
+		this.callback( [ { contentRect: { width: 300 } } ] );
+	}
+	disconnect() {}
+};
+
```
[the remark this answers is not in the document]
kmwilkerson: Addressed in 0dfbb1a — removed the unnecessary global ResizeObserver mock.

includes/reader-activation/integrations/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+Each integration:
+
+- Pushes contact data (Newspack → external) on reader activity events.
+- Pulls contact data (external → Newspack) on a recurring schedule.
+- Declares its own settings, metadata prefix, and outgoing fields.
+- Exposes external fields back to Newspack for use as segmentation criteria and content gate access rules.
+
+The framework is built on top of [Data Events](../../data-events/README.md) and uses WooCommerce [ActionScheduler](https://actionscheduler.org/) for retries and per-integration job grouping.
+
```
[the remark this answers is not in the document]
kmwilkerson: This README was added in the base branch (slice 1), not in this PR. It's picked up in the diff because the base branch diverged from trunk. No action needed here.

includes/wizards/newspack/class-email-preview.php
```diff
@@ -0,0 +1,313 @@
...
+
+	/**
+	 * Get the site's store address as a formatted string.
+	 *
+	 * Mirrors the logic in Emails::get_email_payload() so the preview
+	 * shows the same address format the real email would use.
+	 *
+	 * @return string Formatted site address, or empty string.
+	 */
+	private static function get_site_address(): string {
```
dkoo, line 185: ### 🟡 Non-blocking suggestion: extract formatted site address string to a shared helper method

This method replicates some logic in the `Emails::get_email_payload()` method almost exactly (the docblock even says so!). This creates some redundancy and could result in mismatching output if the `Emails::get_email_payload()` logic ever changes. If we move the logic into a new helper method (e.g. `Emails::get_site_address()`) that both `Email_Preview` and `Emails` classes can use, this simplifies things and prevents future drift.

includes/wizards/newspack/class-email-preview.php
```diff
@@ -0,0 +1,313 @@
...
+		$site_logo_url   = wp_get_attachment_url( get_theme_mod( 'custom_logo' ) );
+		$site_title      = get_bloginfo( 'name' );
+		$site_url        = get_bloginfo( 'wpurl' );
+		$reply_to_email  = Emails::get_reply_to_email();
+		$site_address    = self::get_site_address();
+		$site_contact    = $site_address
+			? sprintf( '<strong>%s</strong> — %s', $site_title, $site_address )
+			: $site_title;
+
+		return [
```
dkoo, line 135: ### 🟡 Non-blocking: static token map could get out-of-sync

This is probably okay for the initial implementation due to how we've added placeholder tokens sort of ad hoc for each email template as needed, but the static map here has potential for easily getting out-of-sync if we add/remove placeholders or make any changes to existing ones in the future.

To properly address this, we'd likely need to make some refactors to how we're adding placeholder tokens in each email template's settings object. We'd want some kind of API to register placeholders in a single shared map that would include config such as the sample value and maybe a data replacement type + sanitization callback. Then, email template configs would reference placeholders from this shared map so we could be sure all email templates are using the same placeholders in the same way.

I'd say this is non-blocking because it would balloon the scope of this PR, but I'd strongly consider a separate follow-up PR to revisit and standardize how we handle all the different placeholders.

includes/wizards/newspack/class-email-preview.php
```diff
@@ -0,0 +1,313 @@
...
+			'*PENDING_EMAIL_ADDRESS*'  => 'sample.reader@example.com',
+
+			// Transaction / subscription details — stable sample values.
+			'*AMOUNT*'                 => '$25.00',
+			'*PAYMENT_METHOD*'         => 'Visa ending in 4242',
+			'*PRODUCT_NAME*'           => 'Monthly Membership',
+			'*BILLING_FREQUENCY*'      => 'monthly',
+			'*DATE*'                   => wp_date( get_option( 'date_format', 'F j, Y' ) ),
+			'*CANCELLATION_TITLE*'     => __( 'Subscription Cancelled', 'newspack-plugin' ),
+			'*CANCELLATION_TYPE*'      => __( 'subscription', 'newspack-plugin' ),
```
dkoo, line 157: ### 🔴 Make all sample strings translatable

Some sample strings here are translatable while others aren't. Let's put them all through `__()` for consistency.


```suggestion
			'*BILLING_FIRST_NAME*'     => __( 'Sample', 'newspack-plugin' ),
			'*BILLING_LAST_NAME*'      => __( 'Reader', 'newspack-plugin' ),
			'*BILLING_NAME*'           => __( 'Sample Reader', 'newspack-plugin' ),
			'*PENDING_EMAIL_ADDRESS*'  => __( 'sample.reader@example.com', 'newspack-plugin' ),

			// Transaction / subscription details — stable sample values.
			'*AMOUNT*'                 => __( '$25.00', 'newspack-plugin' ),
			'*PAYMENT_METHOD*'         => __( 'Visa ending in 4242', 'newspack-plugin' ),
			'*PRODUCT_NAME*'           => __( 'Monthly Membership', 'newspack-plugin' ),
			'*BILLING_FREQUENCY*'      => __( 'monthly', 'newspack-plugin' ),
			'*DATE*'                   => wp_date( get_option( 'date_format', 'F j, Y' ) ),
			'*CANCELLATION_TITLE*'     => __( 'Subscription Cancelled', 'newspack-plugin' ),
			'*CANCELLATION_TYPE*'      => __( 'subscription', 'newspack-plugin' ),
```

src/wizards/newspack/views/settings/emails/email-preview.tsx
```diff
@@ -0,0 +1,217 @@
...
+			cancelled = true;
+			if ( safetyTimerRef.current ) {
+				clearTimeout( safetyTimerRef.current );
+				safetyTimerRef.current = null;
+			}
+		};
+	}, [ isVisible, postId ] );
+
+	// Handle iframe load: wait for stylesheets and images, then measure height and reveal.
+	const handleIframeLoad = useCallback( () => {
```
dkoo, line 138: ### 🟡 Stale `iframe` load handler (flagged by 2/6)

The fetch effect cancels stale network responses, but `handleIframeLoad()` has no generation guard. If `postId` changes mid-load, the previous iframe's `Promise.all([...linkPromises, ...imgPromises])` resolves later and `finalize()` sets `isReady`/`iframeHeight` against the _new_ preview, resulting in visual flash and wrong dimensions.

**Suggested fix:** Add a per-load ref and bail in `finalize` if it doesn't match the current load.

includes/wizards/newspack/class-email-preview.php
```diff
@@ -0,0 +1,313 @@
...
+			? sprintf( '<strong>%s</strong> — %s', $site_title, $site_address )
+			: $site_title;
+
+		return [
+			// Site / branding — real values from the publisher's config.
+			'*SITE_TITLE*'             => $site_title,
+			'*SITE_URL*'               => $site_url,
+			'*SITE_LOGO*'              => $site_logo_url ? esc_url( $site_logo_url ) : '',
+			'*SITE_ADDRESS*'           => $site_address,
+			'*SITE_CONTACT*'           => $site_contact,
```
dkoo, line 141: ### 🔴 Escape raw text tokens

These sample strings come from options or theme mods and are interpolated raw into HTML before reaching `srcDoc`. Escape all output for tighter security:

```suggestion
			'*SITE_TITLE*'             => esc_html( $site_title ),
			'*SITE_URL*'               => esc_url( $site_url ),
			'*SITE_LOGO*'              => $site_logo_url ? esc_url( $site_logo_url ) : '',
			'*SITE_ADDRESS*'           => esc_html( $site_address ),
			'*SITE_CONTACT*'           => esc_html( $site_contact ),
```

tests/unit-tests/email-preview.php
```diff
@@ -0,0 +1,210 @@
...
+	 * Setup.
+	 */
+	public function set_up() {
+		parent::set_up();
+
+		$this->config_filter_callback = function ( $types ) {
+			$types[ self::$test_config_name ] = [
+				'name'        => self::$test_config_name,
+				'label'       => __( 'Test preview config', 'newspack' ),
+				'description' => __( 'Email for testing preview.', 'newspack' ),
```
dkoo, line 42: ### 🔴 Wrong text domain

```suggestion
				'label'       => __( 'Test preview config', 'newspack-plugin' ),
				'description' => __( 'Email for testing preview.', 'newspack-plugin' ),
```

src/wizards/newspack/views/settings/emails/email-preview.tsx
```diff
@@ -0,0 +1,217 @@
...
+			clearTimeout( safetyTimerRef.current );
+			safetyTimerRef.current = null;
+		}
+
+		setIsLoading( true );
+		setIsReady( false );
+		setIframeHeight( null );
+		setHasError( false );
+		setHtml( null );
+		apiFetch< { html: string; post_id: number } >( {
```
dkoo, line 109: ### 🟡 Performance: redundant fetches on `DataView` re-renders

Currently, most `DataViews` remounts (pagination/sort/search) will unnecessarily re-fetch the same `postId`s. Also, changing the view style to "Table" instead of "Grid" will trigger preview re-fetches even though this view doesn't show any thumbnails.

**Suggested fix:** Cache fetched preview HTML in a map by `postId` (maybe in the parent component so it persists without a memo/ref) and load from the cache if available to avoid redundant REST calls.

dkoo: ### 🟡 Non-blocking: consider `useWizardApiFetch` instead

Newer wizard views use this shared hook instead of calling `apiFetch` directly; consider updating to match.

src/wizards/newspack/views/settings/emails/emails.tsx
dkoo, line 70: ### 🟡 Non-blocking: consider `useWizardApiFetch` instead

Same as above: newer wizard views use this shared hook instead of calling `apiFetch` directly; consider updating to match.

src/wizards/newspack/views/settings/emails/email-preview.tsx
```diff
@@ -0,0 +1,217 @@
...
+					srcDoc={ html }
+					/* sandbox: allow-same-origin is required so handleIframeLoad can
+					 * read contentDocument (body.scrollHeight, stylesheet load state).
+					 * allow-scripts is NOT present, so JS cannot execute.
+					 * Residual risk: a <form> in the HTML could submit with admin
+					 * cookies, but this is admin-only code and the email HTML is
+					 * publisher-controlled post meta. */
+					sandbox="allow-same-origin"
+					tabIndex={ -1 }
+					title="Email preview"
```
dkoo, line 204: ### 🟡 Untranslated string

Let's make sure raw strings are translatable. Note: will need to import `__` from `@wordpress/i18n`

```suggestion
					title={ __( 'Email preview', 'newspack-plugin' ) }
```

dkoo (commented):
@kmwilkerson Great work—tests well and looks great in-browser! Commented inline with some minor blockers and a handful of non-blocking suggestions. The ones with a red 🔴 are the most important comments; others are less important and could be deferred if necessary.

Will review #4727 separately.

kmwilkerson:
## Closing as superseded by #4732

After verification against both branches, this PR's contents have been fully absorbed into [#4732](https://github.com/Automattic/newspack-plugin/pull/4732) (slice 2 / NPPD-1527). Specifically:

- **`class-email-preview.php`** — every line in this PR's version exists verbatim in #4732, plus ~140 additional lines for WC email handling (`get_wc_preview_html()`, `get_wc_preview_cache_key()`, `try_block_render()`, logger integration, and routing logic that distinguishes `woo_email` posts from Newspack posts).
- **`email-preview.tsx`** — #4732 is again a strict superset, with two improvements this PR lacked:
  - i18n: `title={ __( 'Email preview', 'newspack-plugin' ) }` instead of a hardcoded string
  - Race-condition fix: `awaitLink`/`awaitImg` attach listeners before checking loaded state, closing a TOCTOU window where an asset could finish between the filter check and listener attachment
- **Tests** — `email-preview.test.js` and the PHPUnit `email-preview.php` are byte-for-byte identical across both branches. #4732 additionally extends `emails.test.js` and `emails-section.php` with WC-specific coverage.

All of @thomasguillot's 27-finding review on this PR — 848px viewport, `is-ready` fade, cancelled flag, `safetyTimerRef`, finalized guard, `onError` handler, `scale` initialized to null, etc. — was addressed in commit `6b8552b3` here and verified "ship-ready" by Thomas. Those patterns all carried into #4732.

No code is lost in this close.

### Tracked follow-ups

@dkoo's review here flagged two items that remain unaddressed in #4732's version too:

1. `esc_html()` on token substitution values
2. Translatable sample strings

These are now tracked separately in [NPPD-1555](https://linear.app/a8c/issue/NPPD-1555) and will be picked up after #4732 merges.

Thanks to everyone who reviewed here — the work absolutely shipped, just under a different PR number. NPPD-1525 will close as Done when #4732 lands.

PR #4731, opened 2026-05-15
fix(integrations): sync reader-account deletion to ESP

chickenn00dle:
### All Submissions:

* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and VIP Go coding standards?

When a reader deletes their account, Newspack is supposed to propagate that deletion to the connected ESP (Mailchimp / ActiveCampaign / Constant Contact). At some point that stopped working — deletions were silently swallowed and the contact lingered in the ESP forever. This PR fixes that, and adds per-integration controls so publishers can choose what "deletion" actually means for each ESP.

[dsgnews-126-deletion-sync-demo.webm](https://github.com/user-attachments/assets/84e174d1-62da-42fe-9931-b6b488ae3472)

#### What changed for publishers

A new section appears on every integration's configure page under *Audience → Integrations → Configure*:

- **Sync user account deletion** — a checkbox (on by default) that controls whether reader deletions reach this integration at all.
- **How to sync deletion** — appears only when the checkbox is on, with two options:
  - **Delete contact immediately** — removes the contact from the ESP entirely. This is the default for integrations that can hard-delete (like the Newsletter ESP). Integrations that can't hard-delete don't show this option at all, so publishers can't pick a mode that would just error on every deletion.
  - **Sync deletion metadata** — leaves the contact in the ESP but stamps it with deletion metadata: an `NP_Account_Deleted` field carrying the deletion timestamp, plus `NP_Membership_Status = user-deleted` (kept for backward compatibility with automations that keyed on the old deletion signal). Publishers can wire their own automation — a Mailchimp journey, an ActiveCampaign automation, etc. — to act on it however they like: archive, tag, unsubscribe, anything.

Existing sites migrate automatically from the legacy `sync_esp_delete` boolean — no publisher action required. Because both legacy states propagated *something* on deletion, deletion sync stays **on** either way; the old boolean only picks the handling mode: `true` → **delete immediately**, `false` → **sync deletion metadata** (preserving the old "don't hard-delete, but still signal the deletion" posture for sites that had opted out of hard delete). Sites that never configured it fall through to the defaults.

> **Behavior change worth a release note:** on existing sites running the v1 integrations path, account deletions now **hard-delete** the ESP contact by default. Publishers who prefer the non-destructive signal can switch any integration to *Sync deletion metadata*, and sites that had explicitly opted out of the legacy hard delete are migrated to that mode automatically.

Closes DSGNEWS-126.

Spin up an isolated env and connect a real ESP so outbound calls actually leave the box:

```bash
n env create dsgnews-126 --worktree newspack-plugin:fix/integrations-sync-deletion --up
n setup --env dsgnews-126 --yes
```

Then connect Newspack Newsletters to a test list on Mailchimp / ActiveCampaign / Constant Contact.

For each test below, create a reader (e.g. via My Account registration), confirm the contact landed in your ESP list, then delete the reader via the WP admin Users screen or My Account → Delete Account.

1. **Default — Delete immediately.** Don't change any settings. After deleting the reader, the contact should be **removed** from the ESP.
2. **Sync deletion metadata (flag mode).** Go to *Audience → Integrations → Newsletter ESP → Configure*. Keep the "Sync user account deletion" checkbox on; switch "How to sync deletion" to *Sync deletion metadata*. Delete a fresh reader. The contact should **stay** in the ESP, now carrying `NP_Account_Deleted` (a `Y-m-d H:i:s` timestamp) **and** `NP_Membership_Status = user-deleted`.
3. **Sync off.** Uncheck "Sync user account deletion" — the "How to sync deletion" select should disappear. Delete a fresh reader. The contact should remain **untouched** in the ESP.
4. **UI reactivity.** On the configure page, toggle the checkbox a few times — the "How to sync deletion" select should show/hide instantly without needing to save.
5. **Legacy migration.** On a site that previously set `newspack_reader_activation_sync_esp_delete` to `false`, load the ESP configure page: "Sync user account deletion" should be **on** with handling defaulted to *Sync deletion metadata* (not skipped). With it set to `true`, handling should default to *Delete contact immediately*.

**Legacy metadata mode.** If a site is still on the legacy metadata mode, the old `reader_deleted` code path runs instead of the new dispatcher. Verify deletion still propagates and that the `sync_esp_delete` boolean is honored.

**Automated tests:**

```bash
n test-php --filter Test_Account_Deletion
n test-php --filter Newspack_Test_Contact_Sync_Connector
n test-php --filter Newspack_Test_Data_Events_Listeners_Deletion
n test-php --filter Test_Esp
```

<details>
<summary>Technical details for reviewers</summary>

**Root cause of the bug.** Both deletion event handlers read `$data['user']['data']['user_email']`. The listeners passed a `WP_User` object whose `ArrayAccess::offsetGet('data')` returns `null`, so the `isset()` guards short-circuited and both handlers returned early without syncing.

**Listener fix.** `reader_deleted` and `reader_delete_sync` now emit `email` as a top-level primitive, consistent with peer listeners (`reader_logged_in`, `reader_verified`).

**Handler split by metadata version.** Previously both the legacy `reader_deleted` and the v1 `reader_delete_sync` handlers registered on every site and both fired on `delete_user`. Now `Contact_Sync_Connector::register_handlers()` registers only one of the two, gated on `Sync\Metadata::get_version()`. Stops double-processing.

**Framework additions:**
- `Integration::delete_contact( string $email )` — new base method, default `WP_Error('not_implemented')`. ESP overrides it to call `Newspack_Newsletters_Contacts::delete()`.
- `Integration::supports_hard_delete(): bool` — capability gate (default `false`; ESP returns `true`). When an integration can't hard-delete, `get_account_deletion_fields()` hides the `delete` option and defaults handling to `flag`, so third-party integrations that only implement `push_contact_data()` aren't exposed as a delete mode that would just return `not_implemented`.
- `Integration::get_account_deletion_fields()` — auto-appended to every integration's settings via `get_settings_fields()`. Provides the two new fields described above.
- `Contact_Sync::handle_account_deletion()` — new central dispatcher called from `reader_delete_sync`. Iterates active integrations and routes each one according to its own settings. Fires `newspack_sync_contact_failed` on errors (with a `mode` discriminator so Alert_Manager can tell delete-mode failures from flag-mode ones) and writes per-integration outcomes to the current ActionScheduler audit log, matching the patterns in `push_to_integrations`. Transient ESP failures are retried via a dedicated `RETRY_DELETION_HOOK` keyed on email + mode (the WP user is gone by retry time).
- Flag-mode metadata is re-injected after `prepare_contact()` with the integration's prefix (so it survives the v1 outgoing-fields filter): `Account_Deleted` (a `Sync\Metadata::DATE_FORMAT` timestamp) and `Membership_Status = user-deleted`, both Title_Case_With_Underscores to match peer prefixed metadata.
- Frontend: small `fieldIsVisible` predicate in `configure-view.js` honors a new `condition` shape on field declarations — generic enough that other integrations can use it too.

**Follow-ups from code review (jason10lee / Copilot):**
- **Deletion can no longer be silently resurrected at shutdown.** `handle_account_deletion()` records the deleted email in a request-scoped set, and `run_queued_syncs()` refuses to re-push any email in it. Without this, an upsert queued by a *later* event in the same Action Scheduler batch (e.g. the `subscription_updated` fired when WooCommerce Subscriptions cancels during the `delete_user` cascade) could recreate the contact at shutdown — a right-to-be-forgotten failure that depended on `delete_user` hook ordering.
- **Smarter legacy migration.** `sync_account_deletion` / `account_deletion_handling` no longer use the naive value-copy option map; `Integration::migrate_account_deletion_setting()` derives both from the single legacy `sync_esp_delete` boolean (`true → delete`, `false → flag`, sync always on). This keeps opted-out sites signalling deletions instead of turning them into no-ops.
- **`membership_status = user-deleted`** is preserved in flag mode for backward compatibility with publisher automations that keyed on the pre-integrations delete-sync signal.
- Minor: PHP 8.3 return/param type hints on the new extension-surface methods; deletion timestamp references `Sync\Metadata::DATE_FORMAT` rather than a hardcoded string; the `fieldIsVisible` boolean coercion handles the literal `'0'` / `'false'` strings explicitly; and `update_settings_field_value()` uses `add_option()` when the option is absent so an unchecked default-`true` checkbox actually persists.

</details>

### Other information:


src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -117,6 +117,17 @@ export const ConfigureView = ( { integrations, loading, pendingChanges, saving,
...
 
+	const fieldIsVisible = field => {
+		if ( ! field.condition || typeof field.condition !== 'object' ) {
+			return true;
+		}
+		const ref = settingsFields.find( f => f.key === field.condition.field );
+		if ( ! ref ) {
+			return true;
+		}
+		return getFieldValue( ref ) === field.condition.equals;
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in 609aa0b. The fieldIsVisible predicate now coerces both sides for boolean conditions so '1'/'0'/'true'/'false' from WP options all match equals: true / equals: false.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -655,6 +670,48 @@ public function get_enabled_outgoing_fields_keys( $prefixed = false ) {
...
+				'label'       => __( 'Sync user account deletion', 'newspack-plugin' ),
+				'description' => __( 'When a reader account is deleted, propagate the deletion to this integration.', 'newspack-plugin' ),
+				'default'     => true,
+			],
+			[
+				'key'         => 'account_deletion_handling',
+				'type'        => 'select',
+				'label'       => __( 'How to sync deletion', 'newspack-plugin' ),
+				'description' => __( 'Choose whether to remove the contact from the integration immediately, or to upsert it with an account-deleted flag so the integration can decide what to do.', 'newspack-plugin' ),
+				'default'     => 'delete',
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in 83dffd4. Added Integration::supports_hard_delete() (default false). For integrations that don't override it, the 'delete' option is hidden from the select and the field defaults to 'flag', so third-party integrations no longer fail by default on every deletion. ESP overrides to true.

tests/unit-tests/data-events/class-contact-sync-connector.php
```diff
@@ -0,0 +1,232 @@
...
+	 */
+	private function reset_data_events_handlers() {
+		$reflection = new \ReflectionClass( Data_Events::class );
+		$property   = $reflection->getProperty( 'actions' );
+		$property->setAccessible( true );
+		$actions = $property->getValue();
+		foreach ( $actions as $action_name => $handlers ) {
+			$actions[ $action_name ] = [];
+		}
+		$property->setValue( null, $actions );
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in f5dac49. The test now snapshots Data_Events::$actions in set_up and restores it in tear_down, so reset_data_events_handlers no longer leaks handler wipes across the suite.

tests/unit-tests/reader-activation-sync.php
```diff
@@ -47,6 +47,11 @@ public function set_option( $value ) {
 	 * Test whether reader data can be synced.
 	 */
 	public function test_can_esp_sync() {
+		if ( defined( 'NEWSPACK_ALLOW_READER_SYNC' ) && NEWSPACK_ALLOW_READER_SYNC ) {
+			$this->markTestSkipped(
+				'NEWSPACK_ALLOW_READER_SYNC was defined by a prior test; default-disabled assertion is moot.'
+			);
+		}
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in f5dac49. The markTestSkipped guard is reverted. The new deletion tests use the newspack_reader_activation_is_syncing_allowed filter (added in 83dffd4) instead of defining the constant globally, so this default-disabled assertion no longer needs to be skipped.

includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -233,6 +233,151 @@ private static function push_to_integrations( $contact, $context, $existing_cont
...
+				// Push through the integration's normal pipeline so prepare_contact applies metadata
+				// prefixing and outgoing-field filtering to publisher-configured metadata. Then
+				// re-inject the account_deleted signal: it's a system-level signal that must
+				// always reach the ESP on deletion, independent of the integration's
+				// outgoing-fields config. Apply the integration's prefix so the field is named
+				// consistently with other metadata on the ESP side.
+				$integration_contact = $integration->prepare_contact( $flag_contact );
+				$prefixed_key        = $integration->get_metadata_prefix() . 'account_deleted';
+				$integration_contact['metadata'] = $integration_contact['metadata'] ?? [];
+				$integration_contact['metadata'][ $prefixed_key ] = $flag_contact['metadata']['account_deleted'];
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in 609aa0b. handle_account_deletion now passes the flag-mode contact through the newspack_esp_sync_contact filter before prepare_contact, restoring parity with Contact_Sync::sync's filter chain.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -655,6 +670,48 @@ public function get_enabled_outgoing_fields_keys( $prefixed = false ) {
...
+	 * @return array Array of settings field declarations.
+	 */
+	public function get_account_deletion_fields() {
+		return [
+			[
+				'key'         => 'sync_account_deletion',
+				'type'        => 'checkbox',
+				'label'       => __( 'Sync user account deletion', 'newspack-plugin' ),
+				'description' => __( 'When a reader account is deleted, propagate the deletion to this integration.', 'newspack-plugin' ),
+				'default'     => true,
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in 83dffd4. update_settings_field_value now detects a missing option via a null sentinel and creates it with add_option in that case, so an unchecked sync_account_deletion persists on a fresh site.

tests/unit-tests/data-events/class-listeners-deletion.php
```diff
@@ -0,0 +1,70 @@
...
+			'newspack_data_event_dispatch',
+			function( $action_name, $timestamp, $data, $client_id ) {
+				$this->captured[ $action_name ] = $data;
+			},
+			10,
+			4
+		);
+	}
+
+	/**
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in f5dac49. The dispatch listener is now stored in a class property and removed in tear_down() so it doesn't pile up across the suite.

tests/unit-tests/integrations/class-test-account-deletion.php
```diff
@@ -0,0 +1,408 @@
...
+		Integrations::register( $this->integration );
+	}
+
+	/**
+	 * Tear down the test environment after each test.
+	 */
+	public function tear_down() {
+		$this->reset_integrations();
+		Integrations::register_integrations();
+		parent::tear_down();
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in f5dac49. tear_down() now deletes Integrations::OPTION_NAME alongside resetting the registry, so test spy IDs don't leak into the enabled-integrations option.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -26,6 +26,7 @@ abstract class Integration {
 		'active_campaign_master_list'     => 'newspack_reader_activation_active_campaign_master_list',
 		'constant_contact_list_id'        => 'newspack_reader_activation_constant_contact_list_id',
 		'sync_esp_delete'                 => 'newspack_reader_activation_sync_esp_delete',
+		'sync_account_deletion'           => 'newspack_reader_activation_sync_esp_delete',
```
jason10lee: **Blocker:** Legacy migration turns 'remove from lists' into a no-op for opted-out sites — `sync_account_deletion` lazily migrates from the legacy `newspack_reader_activation_sync_esp_delete` boolean (line 29). But the legacy semantics were three-way: `sync_esp_delete = false` meant *don't hard-delete, but still remove the contact from all lists* (`update_lists( $email, [], … )`, still present in the unchanged legacy branch at `class-contact-sync-connector.php:221`) and it always flagged `membership_status='user-deleted'`. After migration on a v1 site, `false` maps to `sync_account_deletion = false`, which **skips the integration entirely** (`continue` at `class-contact-sync.php:307`). So a site that deliberately chose 'remove from lists but don't delete' now does **nothing** at the ESP on deletion — no delete, no list removal, no flag. There's no migration target that reproduces the middle state.

**Risk if not addressed:** If publishers deliberately set 'remove from lists but do not hard-delete' as a privacy posture, after migration their deleted readers remain fully active in every ESP list indefinitely with no deletion signal whatsoever — silently reintroducing the lingering-contact / right-to-be-forgotten exposure this change exists to fix, for exactly that configuration.

**Consider:** Provide a migration target that preserves the old 'do not delete but remove from lists' behavior, or at minimum keep performing the list removal when deletion sync is disabled, so deletions on previously opted-out sites do not silently become no-ops.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -655,6 +686,61 @@ public function get_enabled_outgoing_fields_keys( $prefixed = false ) {
...
+	public function get_account_deletion_fields() {
+		$supports_hard_delete = $this->supports_hard_delete();
+
+		// When the integration supports hard delete, expose both options and default
+		// to `delete` (matches the historical sync_esp_delete=true default for ESP).
+		// Otherwise expose only `flag` and default to it — no point letting publishers
+		// pick a mode that will just return `not_implemented` on every deletion.
+		$handling_options = [
+			[
+				'value' => 'flag',
```
jason10lee, line 734: **Blocker:** Existing v1 sites flip to hard-delete by default and lose the deletion signal — Pre-PR, non-legacy sites flagged deletions by writing `membership_status = 'user-deleted'` via `Contact_Sync::sync()` and never hard-deleted. Post-PR the ESP declares `supports_hard_delete() = true` and the deletion-handling field defaults to `'delete'` (line 707), so the same sites now **hard-delete** contacts by default, and the `membership_status='user-deleted'` signal disappears (replaced by an `NP_Account_Deleted` timestamp that is only written in *flag* mode). Publishers with ESP automations keyed on `membership_status='user-deleted'` will silently stop receiving that trigger, and contacts that were previously only flagged are now destroyed. This may be the intended product direction, but the flip from flag to destructive-delete-by-default plus the dropped automation signal is a cross-publisher change.

**Risk if not addressed:** If publishers across multiple sites depend on the prior deletion signal for downstream automations (suppression lists, win-back flows, compliance reporting), then on every deletion that signal silently disappears and contacts previously only flagged are now permanently destroyed by the new default — an irreversible, cross-publisher behavior change shipped without a release note or impact review.

**Consider:** Add an explicit release note and run a cross-publisher impact review for the deletion-signal change, since deletions stop writing membership_status='user-deleted' and may now hard-delete by default where they previously only flagged.

includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -233,6 +243,179 @@ private static function push_to_integrations( $contact, $context, $existing_cont
...
+		if ( empty( $context ) ) {
+			$context = static::$context;
+		}
+
+		// Drop any upsert queued earlier in this same request for the deleted email.
+		// Without this, the shutdown queue (see sync()/run_queued_syncs()) would run
+		// after the ESP delete below and recreate the contact — e.g. when WCS cancels
+		// subscriptions during the delete_user cascade and fires subscription_updated,
+		// which queues a Contact_Sync via the Data Event dispatcher.
+		unset( self::$queued_syncs[ $email ] );
```
jason10lee, line 302: **Blocker:** Shutdown re-sync queue can silently resurrect a deleted contact — `handle_account_deletion()` clears the pending upsert for the deleted email with `unset( self::$queued_syncs[ $email ] )` (line 283), but that only removes syncs already queued at that instant. `run_queued_syncs()` (~line 997) keeps no record of which emails were deleted this request — for an email with no live WP user it re-pushes `$queued_sync['contact']` and calls `Contact_Sync::sync()`, recreating the contact at the ESP. Deletion data events are replayed sequentially in a single Action Scheduler batch, so if any other reader event for the same email (e.g. the `subscription_updated` from the WooCommerce Subscriptions cancel cascade) is dispatched *after* `reader_delete_sync`, the deletion runs first, the `unset()` clears nothing useful, and the later sync re-creates the contact at shutdown — while the deletion path still logs success. Correctness depends on `delete_user` hook ordering, which isn't guaranteed.

**Risk if not addressed:** If a subscription-cancellation event (or any other contact sync) for the deleted reader is processed after the deletion in the same batch, a right-to-be-forgotten deletion is silently undone: the contact reappears at the ESP, the logs claim the delete succeeded, and nothing flags that it came back — recreating the exact lingering-contact exposure this change set out to close, with no signal to detect or fix it.

**Consider:** Track deleted emails for the lifetime of the request (e.g. a static set populated in the deletion handler) and have the shutdown queue refuse to re-push any email in that set, so a later event in the same batch cannot resurrect a deleted contact regardless of dispatch order.

includes/reader-activation/integrations/class-integration.php
```diff
@@ -243,6 +244,36 @@ abstract public function can_sync( $return_errors = false );
...
+	 * When false, the account-deletion settings UI hides the "delete immediately"
+	 * option and falls back to `flag` mode by default — so third-party integrations
+	 * that only implement `push_contact_data()` aren't exposed as a delete-mode
+	 * option that would just return `not_implemented` on every deletion.
+	 *
+	 * Override and return true alongside a `delete_contact()` implementation.
+	 *
+	 * @return bool True if the integration implements delete_contact().
+	 */
+	public function supports_hard_delete() {
```
jason10lee: **Suggestion:** New public integration API methods lack PHP 8.3 type hints — The new extension-surface methods don't carry the param/return types their peers in the same file already use (`supports_frontend_registration(): bool`, `get_registration_key(): string`). `supports_hard_delete()` (`class-integration.php:259`, `class-esp.php:360`) should be `: bool`; `delete_contact( $email )` (`class-integration.php:273`, `class-esp.php:370`) should be `delete_contact( string $email )`. Integration authors will subclass these, so typing them now avoids churn later.

includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -233,6 +243,179 @@ private static function push_to_integrations( $contact, $context, $existing_cont
...
+		$integrations = Integrations::get_active_integrations();
+		$errors       = [];
+
+		// Build the flag-mode contact once. The timestamp uses the same format as peer
+		// datetime metadata fields (Sync\Contact_Metadata::DATE_FORMAT — 'Y-m-d H:i:s')
+		// so publishers can apply consistent automation rules across reader events.
+		$flag_contact          = $contact;
+		$flag_contact['email'] = $email;
+		$flag_contact['metadata'] = isset( $flag_contact['metadata'] ) ? $flag_contact['metadata'] : [];
+		$flag_contact['metadata']['account_deleted'] = gmdate( 'Y-m-d H:i:s' );
```
jason10lee: **Nit:** Deletion timestamp hardcodes the format the comment says is a constant — The comment says the timestamp matches `Sync\Contact_Metadata::DATE_FORMAT`, but the code hardcodes `gmdate( 'Y-m-d H:i:s' )`. If that constant ever changes, the deletion timestamp silently diverges from peer datetime fields. Reference the constant directly to keep them in lockstep.

src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -117,6 +117,24 @@ export const ConfigureView = ( { integrations, loading, pendingChanges, saving,
...
+			return true;
+		}
+		const refValue = getFieldValue( ref );
+		// For boolean conditions, coerce both sides — values can arrive from WP options
+		// as scalar strings (`'1'`/`'0'`/`''`) after migration or from the REST layer,
+		// so strict equality would hide dependent fields until the parent is re-saved.
+		if ( typeof field.condition.equals === 'boolean' ) {
+			return Boolean( refValue ) === field.condition.equals;
+		}
+		return refValue === field.condition.equals;
```
jason10lee, line 138: **Nit:** Boolean-coercion fix still mishandles the literal string '0' — The visibility fix coerces with `Boolean( refValue )`, but `Boolean('0') === true` in JS — the exact `'0'` case the inline comment calls out would still show a dependent field when the parent is the string `'0'`. In practice `getFieldValue` only ever yields a JS boolean or `''`/`true`, so `'0'` isn't reachable today and the effect is cosmetic (the select would show, but deletion is still skipped server-side). Either treat `'0'`/`'false'` as false explicitly, or drop the `'0'` claim from the comment. (Refines the already-resolved Copilot point — the fix is correct for the reachable cases.)

jason10lee (changes requested):
The core fix is sound and genuinely repairs a silent, GDPR-relevant deletion failure, landing with real retry/logging and strong test coverage. 

Automated review has flagged several blockers, all on the deletion path itself: (1) the shutdown re-sync queue can resurrect a just-deleted contact if another event for the same email is processed after the deletion in the same batch — the `unset()` guard is order-dependent; (2) existing v1 sites flip to hard-delete by default and lose the `membership_status='user-deleted'` signal publishers may automate on; and (3) the legacy migration turns 'remove from lists but don't delete' into a no-op, so opted-out sites stop propagating deletions entirely. 

The remaining items are minor (type hints, a timestamp constant, a cosmetic JS coercion, duplicate event dispatch). 

Impact review flagged the above blockers, but otherwise had nothing else to say to us.

Finally, GitHub has flagged a conflict that'll need to be resolved before merge.  This doesn't appear to directly impact the review.

**Nits on impacted code outside these commits**:
- `includes/data-events/listeners.php:44` — **Nit:** Both deletion events dispatch even though only one handler is registered — Each `delete_user` now fires *both* the `reader_deleted` and `reader_delete_sync` listeners, so two data events are dispatched (each its own async ActionScheduler action / webhook POST) even though `Contact_Sync_Connector` only registers a handler for one of them per site. The unhandled event still travels the dispatch pipeline (webhook subscribers, etc.). Harmless for correctness; could gate the listener emission on `Metadata::get_version()` the same way handler registration is, if event volume matters.

includes/reader-activation/integrations/class-integration.php
chickenn00dle: Good catch — addressed in e0ca4a6. `sync_account_deletion` / `account_deletion_handling` no longer copy the legacy boolean straight across. `Integration::migrate_account_deletion_setting()` now derives both from `sync_esp_delete`: since both legacy states propagated *something* on deletion, sync stays on, and the boolean only picks the mode — `true → delete`, `false → flag`. So an opted-out (`false`) site now flags the deletion instead of going silent.

Reproducing the exact "remove from lists but don't delete" middle state would need a third mode; mapping to flag felt like the right balance between preserving intent and not adding surface. WDYT?

includes/reader-activation/integrations/class-integration.php
chickenn00dle: Thanks — the hard-delete-by-default is a deliberate product call and stays: the whole point of the change is to actually propagate right-to-be-forgotten deletions. I've flagged it as a release-note-worthy behavior change in the PR description, and the impact review already ran.

On the dropped signal though, you're right — e0ca4a6 re-injects `membership_status = user-deleted` in flag mode (alongside the new `Account_Deleted` timestamp), so publisher automations keyed on the old value keep working.

includes/reader-activation/sync/class-contact-sync.php
chickenn00dle: Great find — fixed in e0ca4a6. `handle_account_deletion()` now records the deleted email in a request-scoped set, and `run_queued_syncs()` skips any email in it, so a sync queued by a later event in the same Action Scheduler batch can't resurrect the contact regardless of `delete_user` hook order. Added a regression test (`test_run_queued_syncs_does_not_resurrect_deleted_email`).

includes/reader-activation/integrations/class-integration.php
chickenn00dle: Done in e0ca4a6 — `supports_hard_delete(): bool` and `delete_contact( string $email )` on both the base class and ESP (plus the two test fixtures, since the return-type covariance would otherwise fatal).

includes/reader-activation/sync/class-contact-sync.php
chickenn00dle: Fixed in e0ca4a6 — now `gmdate( Sync\Metadata::DATE_FORMAT )` (already imported), and the comment matches. Same constant the peer datetime fields use, so they stay in lockstep.

src/wizards/audience/views/integrations/configure-view.js
chickenn00dle: Addressed in e0ca4a6 — `fieldIsVisible` now treats `''` / `'0'` / `'false'` as false explicitly instead of leaning on `Boolean( '0' )`. Agree it's cosmetic today given `getFieldValue` only ever yields a boolean or `''`, but cheap to make correct.

chickenn00dle:
Thanks for the thorough review @jason10lee — all addressed in e0ca4a6, with each inline thread answered and resolved. Quick rundown against your summary:

**Blockers**
- **Shutdown re-sync resurrection** — `handle_account_deletion()` now records the deleted email for the request and `run_queued_syncs()` refuses to re-push it, so a later event in the same batch can't recreate the contact. Regression test added.
- **v1 hard-delete-by-default + dropped signal** — hard-delete-by-default is intentional (the point is to actually propagate deletions) and now called out as a behavior change in the PR description. The `membership_status = user-deleted` signal is preserved in flag mode alongside the new `Account_Deleted` timestamp.
- **Legacy `false` → no-op** — migration now maps legacy `false` to flag mode instead of skipping, so opted-out sites keep signalling deletions.

**Minor** — type hints, the `DATE_FORMAT` constant, and the JS `'0'`/`'false'` coercion are all in.

**Deferred (one):** gating the `reader_deleted` / `reader_delete_sync` *listener* emission on metadata version. It's harmless for correctness as you noted, and gating it would stop v1 sites emitting the `reader_deleted` **webhook** event, which external subscribers may depend on — so I left it rather than risk a silent webhook regression. Happy to revisit if you feel strongly.

**Conflict** — resolving the merge conflict with `trunk` now.

includes/reader-activation/integrations/class-integration.php
jason10lee: Yeah, doesn't look like we'll want to add more surface to maintain here.  Figure we've got enough here for future developers (or their agents) to establish intent if the question comes up again.  Thanks!

includes/reader-activation/integrations/class-integration.php
```diff
@@ -861,6 +966,38 @@ function( $field ) {
...
+	 * @return mixed|null The migrated value, or null if there is no legacy option to migrate.
+	 */
+	private function migrate_account_deletion_setting( $key, $option_name ) {
+		$legacy_value = \get_option( self::LEGACY_SYNC_DELETE_OPTION, null );
+		if ( null === $legacy_value ) {
+			return null;
+		}
+		$migrated = 'sync_account_deletion' === $key
+			? true
+			: ( \wp_validate_boolean( $legacy_value ) ? 'delete' : 'flag' );
```
jason10lee: **Nit (non-blocking):** This migration derives `'delete'` purely from the legacy boolean, without consulting `supports_hard_delete()` — so it can persist `account_deletion_handling = 'delete'` for an integration that can't hard-delete. The deletion loop calls `delete_contact()` on `'delete'` mode with no capability guard (`class-contact-sync.php:344`), so such an integration would return `not_implemented` on every deletion (error log + retry + alert) — a narrow re-opening of the earlier field-default fix, but via the migration path instead of the UI.

Unreachable for first-party code (the legacy option only governed ESPs, which all support hard delete); it'd only bite a third-party integration that's active on a legacy-`true` site. A one-line clamp closes it:
```php
$migrated = 'sync_account_deletion' === $key
    ? true
    : ( \wp_validate_boolean( $legacy_value ) && $this->supports_hard_delete() ? 'delete' : 'flag' );
```
Fine to defer or wontfix — flagging for completeness.

jason10lee (approved):
Re-reviewed the responses in `e0ca4a6` — all six threads are addressed, and I verified each against the diff rather than the reply text. Approving with one nit that was introduced with the migration fix and that *we* will probably never trigger.

**Blockers, all resolved:**
- **Legacy-migration no-op** — the straight value-copy is gone; `migrate_account_deletion_setting()` now keeps deletion sync *on* for opted-out sites and maps the legacy boolean to the handling mode (`false → flag`), so a previously opted-out site emits a deletion signal instead of silently doing nothing. Agree with the `flag` mapping over reintroducing a third mode.
- **Dropped `user-deleted` signal** — flag mode re-injects `membership_status = user-deleted` both raw and prefixed (`NP_Membership_Status`), so it survives `prepare_contact` and keeps existing publisher automations working.
- **Shutdown-queue resurrection** — the request-scoped `$deleted_emails` set + the `run_queued_syncs()` skip is exactly the fix, and the regression test proves order-independence (deletes an email, then queues a later upsert for it *and* a control email, asserts only the deleted one is withheld).

**Nits/suggestion** — type hints, `Metadata::DATE_FORMAT`, and the explicit `'0'`/`'false'` coercion in `fieldIsVisible` are all done correctly.

On the hard-delete-by-default product decision: 👍 on keeping it, and thanks for the release note in the description. Left one non-blocking follow-up nit inline about the migration path. Not a merge blocker.

includes/reader-activation/integrations/class-integration.php
chickenn00dle: Done in d4086aee6 — `migrate_account_deletion_setting()` now clamps the `delete` target on `supports_hard_delete()`, so a legacy `true` value only migrates to `delete` when the integration can actually hard-delete; otherwise it falls to `flag`. Mirrors the field default in `get_account_deletion_fields()`, so the migration and UI paths default identically now. Added a regression test (`test_migration_legacy_true_maps_to_flag_when_hard_delete_unsupported`) covering a non-hard-delete integration on a legacy-`true` site.

Issue #75, opened 2019-05-28
Text input behavior doesn't match latest designs, bug with 'active' class


claudiulodro:
**Describe the bug**
Originally reported by @justinshreve at https://github.com/Automattic/newspack-plugin/pull/72#issuecomment-496292548

> I've noticed a few behaviors with the components that differ from what I see in the Figma Component file. Let me know if you want me to help tackle these.
> 
> For inputs, the label should still be displayed when active:
> 
> <img width="830" alt="Screen Shot 2019-05-27 at 3 34 32 PM" src="https://user-images.githubusercontent.com/689165/58436765-0e4e3680-8095-11e9-81ca-e3fa65f041a8.png">
> 
> I also noticed that the active class does not get applied if the input is empty. That means the input looks unfocused even though it is. There should be some kind of active but empty state.
> 
> I also noticed that if you tab/keyboard navigate into the inputs, the active class is not appended, even if you begin typing.

claudiulodro:
This is not high priority, as it looks like we'll eventually migrate to `calypso-ui` components.

thomasguillot:
Out of date designs

PR #4733, opened 2026-05-18
feat(cli): add fix-memberships command (NPPM-386)

adekbadek:
### All Submissions:

Transplants `wp newspack-subscription-migrations fix-memberships` from [newspack-subscription-migrations#36](https://github.com/Automattic/newspack-subscription-migrations/pull/36) into newspack-plugin as `wp newspack fix-memberships`.

Addresses NPPM-386.

**Why move it here:**
- It lives next to the closely-related `wp newspack migrate-expired-subscriptions` and `wp newspack teams-for-memberships diagnostics` commands. Same shape (single-site WC Memberships/Subscriptions cleanup CLI), same gating pattern (`function_exists( 'wc_memberships' )`).
- Ships with every Newspack install instead of requiring publishers to side-load `newspack-subscription-migrations`.
- Closes #36 as part of this PR (see comment there).

**What the command does** (dry-run by default, `--live` to apply):

1. **Primary pass** – walks users who have an active subscription for a plan product but no active membership, and:
   - Reclaims a network-managed membership when a local subscription exists (strips `_managed_by_newspack_network` / `_remote_id` / `_remote_site_url` and relinks to the local subscription).
   - Relinks subscription-linked memberships that lost their `_product_id` meta.
   - Reactivates inactive memberships whose subscription is active (clears `_end_date` and unschedules expiration events to avoid re-expiry via cron).
   - Creates a new membership when none exists for the plan/user.
2. **Second pass** – clears stale `_end_date` on memberships whose linked subscription is still active (independent of the primary pass).

**Handles:**
- Team products (`_wc_memberships_for_teams_plan` meta on products) and team plans.
- Variable / variable-subscription products (expands to children).
- Transferred subscriptions (billing email belongs to a different WP user who already has a membership linked to the subscription – skipped to avoid duplicates).
- Team subscription payers (skipped when they have no membership for the plan).
- HPOS-enabled sites (branches the subscription join on `woocommerce_custom_orders_table_enabled`).
- Network-managed memberships (per-membership classification rather than a blanket skip).

**Differences vs the source #36:**
- Conformed to newspack-plugin namespacing (`Newspack\CLI\Fix_Memberships`) and the existing CLI registration pattern in `class-initializer.php` (gated on `wc_memberships`, like the teams-diagnostics command is gated on `WC_Memberships_For_Teams_Loader`).
- Split the original 700-line `fix_memberships` method into focused private helpers (`process_user`, `handle_user_without_local_membership`, `handle_user_with_inactive_membership`, etc.).
- Extracted the repeated 4-call reactivation sequence into a `reactivate_if_inactive( $membership )` helper used everywhere a reactivation was open-coded in #36 (5 sites).
- Extracted `clear_network_managed_meta( $membership_id )` for the three `_managed_by_newspack_network` / `_remote_id` / `_remote_site_url` deletions.
- Method renamed from `fix_memberships` to `run` to avoid `Generic.NamingConventions.ConstructorName.OldStyle` (the class is `Fix_Memberships`, so a `fix_memberships` method tripped the PHP4 constructor detector).
- Removed the always-`true` default for `--verbose`. Opt-in like every other Newspack CLI.

**Worth considering as a follow-up (not done here):** `Membership_Expiry::prevent_membership_expiration()` does a structurally similar reactivation when an alternate active subscription is found. It currently omits `unschedule_expiration_events()` + `set_end_date( '' )` because it runs as a pre-expiration filter; this PR's helper does both. Pulling the helper out into something `Membership_Expiry` can also call would let both code paths share one implementation. Left out of this PR to keep behavior of existing code unchanged.

1. Local site with WooCommerce, WooCommerce Subscriptions, and WooCommerce Memberships active.
2. Create a membership plan with at least one linked product, plus a customer with an active subscription for that product.
3. Manually break the membership state (e.g. set the membership to `expired` while the subscription is still `active`, or delete the membership entirely).
4. Run `wp newspack fix-memberships --verbose` and confirm the dry-run output describes the right corrective action without changing data.
5. Run `wp newspack fix-memberships --verbose --live` and confirm the membership is reactivated / relinked / created.
6. Optional: with `newspack-network` installed, mark a membership as `_managed_by_newspack_network` while a local subscription exists, and verify the command reclaims it under `--live`.
7. Optional: enable HPOS (`woocommerce_custom_orders_table_enabled=yes`) and re-run – the command should detect HPOS and use the orders-table join.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? – no automated tests; behavior is exercised on data the local env doesn't have. See the manual test plan above.
* [x] Have you successfully ran tests with your changes locally? – PHPCS clean; PHP syntax clean.

### DIFF
--- a/includes/cli/class-fix-memberships.php
+++ b/includes/cli/class-fix-memberships.php
@@ -504,7 +504,16 @@ class Fix_Memberships {
 			return;
 		}
 
-		$plan                 = wc_memberships_get_membership_plan( $plan_id );
+		// The team-plan fallback reads the plan ID from product meta, which outlives a deleted
+		// plan post, so the plan may not load even though we have an ID for it.
+		$plan = wc_memberships_get_membership_plan( $plan_id );
+		if ( ! $plan ) {
+			$log_line = sprintf( 'Membership plan (#%d) for subscription (#%d) could not be loaded, skipping.', $plan_id, $latest_active_subscription_id );
+			WP_CLI::warning( $log_line );
+			self::$command_results['skipped'][] = $log_line;
+			return;
+		}
+
 		$plan_product_id_list = self::get_full_plan_product_id_list( $plan, $plan_id );
 		$product_ids          = array_values( array_intersect( $subscription_product_ids, $plan_product_id_list ) );
 		if ( empty( $product_ids ) ) {
--- a/includes/cli/class-fix-memberships.php
+++ b/includes/cli/class-fix-memberships.php
@@ -59,7 +59,8 @@ class Fix_Memberships {
 	 * : Produce more output.
 	 *
 	 * [--limit=<n>]
-	 * : Limit the number of customers processed.
+	 * : Limit the number of records processed. Applies to both passes: at most <n>
+	 * customers in the first pass, and at most <n> memberships in the stale end date pass.
 	 *
 	 * ## EXAMPLES
 	 *
@@ -72,8 +73,6 @@ class Fix_Memberships {
 	 * @return void
 	 */
 	public function run( $args, $assoc_args ) {
-		global $wpdb;
-
 		if ( ! function_exists( 'wc_memberships_get_membership_plan' ) ) {
 			WP_CLI::error( 'WooCommerce Memberships plugin is not active.' );
 		}
@@ -101,25 +100,24 @@ class Fix_Memberships {
 		}
 		WP_CLI::line( '' );
 
-		$all_product_ids = self::collect_plan_product_ids();
-		if ( empty( $all_product_ids ) ) {
+		$plan_product_pairs = self::collect_plan_product_pairs();
+		if ( empty( $plan_product_pairs ) ) {
 			WP_CLI::warning( 'No products found linked to any membership plan.' );
 			return;
 		}
 
-		$plan_product_ids = implode( ',', array_map( 'intval', $all_product_ids ) );
-		$is_using_hpos    = 'yes' === get_option( 'woocommerce_custom_orders_table_enabled' );
+		$is_using_hpos = 'yes' === get_option( 'woocommerce_custom_orders_table_enabled' );
 
 		WP_CLI::line( $is_using_hpos ? 'Site is using HPOS.' : 'Site is not using HPOS.' );
 
-		$affected_users = self::query_affected_users( $plan_product_ids, $is_using_hpos );
+		$affected_users = self::query_affected_users( $plan_product_pairs, $is_using_hpos );
 
 		if ( false !== $limit ) {
-			WP_CLI::warning( sprintf( 'Results limited to %d.', $limit ) );
+			WP_CLI::warning( sprintf( 'Results limited to %d per pass.', $limit ) );
 			$affected_users = array_slice( $affected_users, 0, $limit );
 		}
 
-		WP_CLI::line( sprintf( 'Will process %d customers.', count( $affected_users ) ) );
+		WP_CLI::line( sprintf( 'Will process %d customer/plan pairs.', count( $affected_users ) ) );
 		WP_CLI::line( '' );
 
 		$site_url = get_option( 'siteurl' );
@@ -128,16 +126,20 @@ class Fix_Memberships {
 			self::process_user( $result, $site_url );
 		}
 
-		self::fix_stale_end_dates( $is_using_hpos );
+		self::fix_stale_end_dates( $is_using_hpos, $limit );
 		self::print_summary();
 	}
 
 	/**
-	 * Collect all product IDs (including team products and variations) linked to any membership plan.
+	 * Collect every (product, plan) pair that grants a membership, including team
+	 * products and variations.
 	 *
-	 * @return int[]
+	 * Pairs rather than a product => plan map: the same product may grant more than
+	 * one plan, and collapsing that would silently drop a plan from the reconciliation.
+	 *
+	 * @return array<int,array{product_id:int,plan_id:int}>
 	 */
-	private static function collect_plan_product_ids() {
+	private static function collect_plan_product_pairs() {
 		$plans = get_posts(
 			[
 				'post_type'      => 'wc_membership_plan',
@@ -145,20 +147,19 @@ class Fix_Memberships {
 			]
 		);
 
-		// Regular plan products.
-		$parent_product_ids = array_reduce(
-			$plans,
-			function( $acc, $post ) {
-				$ids = get_post_meta( $post->ID, '_product_ids', true );
-				return array_merge( is_array( $ids ) ? $ids : [], $acc );
-			},
-			[]
-		);
-
-		// Team products (woocommerce-memberships-for-teams stores these via _wc_memberships_for_teams_plan on products).
-		$team_product_ids = [];
+		$parent_pairs = [];
 		foreach ( $plans as $plan ) {
-			$team_products    = get_posts(
+			// Regular plan products.
+			$product_ids = get_post_meta( $plan->ID, '_product_ids', true );
+			foreach ( is_array( $product_ids ) ? $product_ids : [] as $product_id ) {
+				$parent_pairs[] = [
+					'product_id' => (int) $product_id,
+					'plan_id'    => (int) $plan->ID,
+				];
+			}
+
+			// Team products (woocommerce-memberships-for-teams stores these via _wc_memberships_for_teams_plan on products).
+			$team_products = get_posts(
 				[
 					'post_type'      => [ 'product', 'product_variation' ],
 					'post_status'    => 'any',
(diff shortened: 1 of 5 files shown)


includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,996 @@
...
+	 * @param bool   $is_using_hpos Whether HPOS is enabled.
+	 * @return array<int,array<string,string|null>>
+	 */
+	private static function query_affected_users( $plan_product_ids, $is_using_hpos ) {
+		global $wpdb;
+
+		if ( $is_using_hpos ) {
+			$active_subscriptions_query = "
+			WITH ActiveSubscriptions AS (
+				SELECT subscriptions.customer_id AS customer_user_id,
```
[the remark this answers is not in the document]
adekbadek: Good catch - fixed in aec2005 by rewriting query_affected_users() to use derived tables instead of CTEs, so it no longer requires MySQL 8.0 / MariaDB 10.2. The three WITH blocks became inline subqueries in the FROM / LEFT JOIN clauses; the projection, join keys and the COALESCE(m.active_memberships_count, 0) = 0 filter are unchanged, and both the HPOS and non-HPOS branches were converted.

Verified equivalence rather than assuming it: I ran the old CTE query and the new derived-table query side by side against a local database with real data (8 users with an active subscription on a plan product) and diffed the two result sets - identical, same columns and same rows. The non-HPOS variant was executed too, to confirm it parses.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,996 @@
...
+		}
+
+		$subscription_ids              = array_map( 'intval', $subscription_ids );
+		$latest_active_subscription_id = max( $subscription_ids );
+		$latest_active_subscription    = wcs_get_subscription( $latest_active_subscription_id );
+
+		// Detect transferred subscription: billing email belongs to a different WP user
+		// who already has a membership linked to this subscription. Skip to avoid duplicates.
+		if ( self::subscription_was_transferred( $latest_active_subscription, $latest_active_subscription_id, $user, $user_id ) ) {
+			return;
```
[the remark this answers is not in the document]
adekbadek: Already addressed in aec2005c (pushed a few minutes after this comment landed, so the review was against the prior diff). `wcs_get_subscription()` is now followed by an `instanceof \WC_Subscription` guard that logs a skip line and returns before `subscription_was_transferred()` / `is_skippable_team_payer()` are reached.

Following the same reasoning I found one more unguarded load one step down and fixed it in 759e21b: `wc_memberships_get_membership_plan()` can also return false, because the team-plan fallback in `find_plan_id_for_subscription_products()` reads the plan ID from `_wc_memberships_for_teams_plan` product meta, which outlives a deleted plan post. `get_full_plan_product_id_list()` would then fatal on `$plan->get_product_ids()`. Same skip-and-log treatment.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+		WP_CLI::line( sprintf( 'Will process %d customers.', count( $affected_users ) ) );
+		WP_CLI::line( '' );
+
+		$site_url = get_option( 'siteurl' );
+
+		foreach ( $affected_users as $result ) {
+			self::process_user( $result, $site_url );
+		}
+
+		self::fix_stale_end_dates( $is_using_hpos );
```
dkoo: **Severity:** suggestion

`fix_stale_end_dates()` runs unconditionally and does not honor `--limit`. `run()` slices `$affected_users` to `--limit` for the first pass, but this second pass then queries and mutates *every* subscription-linked membership site-wide with a stale `_end_date`. An operator running `--live --limit=10` as a cautious canary would reasonably expect at most ~10 mutations but would trigger unbounded writes across the whole site.

**Risk if not addressed:** On a large site where thousands of memberships carry a stale `_end_date` pointing at a still-active subscription, one unattended run the operator believed was scoped to a handful of customers mutates thousands of memberships, with no way to have previewed the real scope from the limited run.

**Consider:** thread the limit (or the just-processed affected-user set) into the second pass, or at minimum document in the `## OPTIONS` block that `--limit` only bounds the first pass.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+
+		$sql_query = "
+		SELECT s.customer_user_id,
+			s.subscription_ids,
+			im.membership_ids,
+			im.membership_statuses
+		FROM ( $active_subscriptions_query ) s
+		LEFT JOIN ( $active_memberships_query ) m ON s.customer_user_id = m.customer_user_id
+		LEFT JOIN ( $inactive_memberships_query ) im ON s.customer_user_id = im.customer_user_id
+		WHERE COALESCE(m.active_memberships_count, 0) = 0;
```
dkoo, line 302: **Severity:** suggestion

The affected-users query collapses all plan products to a single active-membership count per customer and excludes the customer whenever `COALESCE(active_memberships_count, 0) = 0` is false — i.e. when they have *any* active membership for *any* plan product. A reader with an active Plan A membership plus an active Plan B subscription whose membership is missing is therefore never returned, and a customer with multiple broken subscriptions collapses to one row. If multi-plan customers are in scope, the active-membership match needs to be per plan/product rather than per customer.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+			if ( ! $reclaim_id ) {
+				$reclaim_id = $managed_membership_ids[0];
+			}
+
+			$log_line = sprintf( 'Reclaiming network-managed membership (#%d) for local subscription (#%d) for user %s.', $reclaim_id, $latest_active_subscription_id, $user->user_email );
+			if ( self::$live ) {
+				self::clear_network_managed_meta( (int) $reclaim_id );
+				$membership = new \WC_Memberships_Integration_Subscriptions_User_Membership( $reclaim_id );
+				$membership->set_subscription_id( $latest_active_subscription_id );
+				$membership->set_order_id( $latest_active_subscription->get_parent_id() );
```
dkoo, line 517: **Severity:** suggestion

The direct-reclaim branch clears the network meta and sets `set_subscription_id()` + `set_order_id()`, but unlike the other two relink paths (`relink_existing_membership`, `reclaim_or_relink_plan_membership`) it never calls `set_product_id()`. A reclaimed network-managed membership that had no local `_product_id` stays product-less after reclaim. It self-heals on a second run (once `_subscription_id` is set the relink path picks it up and sets the product), but a single `--live` pass leaves it incompletely linked. Setting the product ID here would give single-pass convergence.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+			$membership         = new \WC_Memberships_Integration_Subscriptions_User_Membership( $membership_id );
+			$log_line           = sprintf(
+				'Membership #%d (%s, %s): clearing end_date %s (subscription #%s).',
+				$membership_id,
+				$row['user_email'],
+				$row['membership_status'],
+				$row['end_date'],
+				$row['subscription_id']
+			);
+			$needs_reactivation = ! in_array( $row['membership_status'], [ 'wcm-active', 'wcm-free_trial' ], true );
```
dkoo, line 948: **Severity:** suggestion

This pass reactivates any membership whose status is not active/free_trial as long as its `_subscription_id` points to a `wc-active` subscription, with no check for whether the inactive state was deliberate — and activation emails are globally suppressed (line 85), so the change is silent to reader and staff.

**Risk if not addressed:** If actively-subscribed customers had memberships revoked for a business reason (chargeback, refund-in-progress, abuse hold) while the subscription had not yet been cancelled, a single `--live` run silently reinstates paid access for all of them with no audit trail beyond stdout.

**Consider:** excluding (or requiring explicit opt-in for) statuses that represent a deliberate revocation rather than reactivating every non-active status.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+		LEFT JOIN {$wpdb->prefix}postmeta mp ON memberships.ID = mp.post_id AND mp.meta_key = '_product_id'
+		WHERE memberships.post_type = 'wc_user_membership'
+		AND memberships.post_status IN ('wcm-active', 'wcm-free_trial')
+		AND mp.meta_value IN ({$plan_product_ids})
+		GROUP BY memberships.post_author
+		";
+
+		$inactive_memberships_query = "
+		SELECT memberships.post_author AS customer_user_id,
+			GROUP_CONCAT(memberships.ID) AS membership_ids,
```
dkoo: **Severity:** nit

`subscription_ids` / `membership_ids` are assembled with `GROUP_CONCAT` and parsed back via `explode`/`max`. The item-meta joins emit one row per matching order item, so IDs are duplicated in the concatenation and long lists can hit `group_concat_max_len` (default 1024) and truncate silently — a cut trailing entry could feed `max()` a bogus value. Low probability, but a `SET SESSION group_concat_max_len` guard or a comment would document the assumption. (`membership_statuses` is also concatenated but never consumed.)

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+				]
+			);
+			$membership = new \WC_Memberships_Integration_Subscriptions_User_Membership( $membership->get_id() );
+			$membership->set_start_date( $subscription->get_date( 'start' ) );
+			$membership->set_subscription_id( $subscription_id );
+			WP_CLI::success( sprintf( 'Created a membership (#%d) for user %s.', $membership->get_id(), $user->user_email ) );
+			self::$command_results['processed'][] = sprintf( 'Created a membership for user %s.', $user->user_email );
+		} catch ( \Throwable $th ) {
+			WP_CLI::warning( sprintf( 'Could not create a membership for user %s.', $user->user_email ) );
+		}
```
dkoo, line 874: **Severity:** nit

The `create_membership()` catch emits `WP_CLI::warning` but pushes nothing to `$command_results`, so a failed creation vanishes from the final summary while the dry-run "would create" path is counted. Push a `skipped` line in the catch for accounting parity.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1014 @@
...
+	 *     wp newspack fix-memberships
+	 *     wp newspack fix-memberships --live --verbose
+	 *     wp newspack fix-memberships --limit=50
+	 *
+	 * @param array $args       Positional arguments.
+	 * @param array $assoc_args Assoc arguments.
+	 * @return void
+	 */
+	public function run( $args, $assoc_args ) {
+		global $wpdb;
```
dkoo: **Severity:** nit

`global $wpdb;` is declared in `run()`, `process_user()`, and `handle_user_without_local_membership()` but never used in any of them (all DB access is delegated to helpers). Dead declarations — safe to drop.

dkoo (changes requested):
Release risk: HIGH — one blocker that must be resolved before merge (machine-local dependency symlinks committed to the branch), plus a `--limit` footgun in the destructive second pass worth addressing.

Thanks for the careful transplant — the command logic is in good shape. The dry-run/`--live` gating is disciplined (every mutation sits behind `--live` with a matching dry-run "would" branch), the raw SQL interpolates only integer-sanitized values or uses `$wpdb->prepare()`, the HPOS and non-HPOS branches are at parity, and the create path is well-guarded so it relinks rather than duplicates an existing membership.

**Blocker — remove the committed `node_modules` and `vendor` symlinks.** This branch commits two absolute symlinks (added in aec2005c):

- `node_modules` → `/Users/adbo/lol/a8c/newspack-docker/repos/newspack-plugin/node_modules`
- `vendor` → `/Users/adbo/lol/a8c/newspack-docker/repos/newspack-plugin/vendor`

They point at an absolute path on one machine, so on every other checkout and in CI they dangle, `npm ci` / `composer install` find tracked paths where they expect to create local dependency directories, and they would ship inside the released plugin zip as broken symlinks. `.gitignore` only excludes the directories (`/node_modules/`, `/vendor/`), so the symlink files slipped past it. Please `git rm --cached node_modules vendor` and drop them from the branch (and consider a gitignore entry that also matches the symlink form).

The inline comments below are non-blocking; the `--limit` one has the most operational teeth. Separately, worth noting there are no automated tests for a 1000-line data-mutating command — given the `--live` blast radius, a few PHPUnit tests around the branch-selection logic and the HPOS/non-HPOS query builders would de-risk future edits. Not a merge blocker given the dry-run default, but a good follow-up.

includes/cli/class-fix-memberships.php
adekbadek: Fixed in 003e86f. `--limit` now bounds both passes: `run()` threads it into `fix_stale_end_dates()`, which slices the result set, and the header line reports found vs. processed ("Found 3 ... processing 1"). The `## OPTIONS` text says so explicitly now.

Verified on a seeded env (3 memberships with a stale `_end_date` on active subscriptions). Before: `--limit=1` printed "Will process 1 customers" and then mutated all 3. After: `Results limited to 1 per pass.` / `Found 3 ... processing 1.` and only membership #29 is touched.

includes/cli/class-fix-memberships.php
adekbadek: Fixed in 003e86f. The affected-users query is now grouped by (customer, plan) instead of by customer: subscriptions join a product => plan derived table and group by `pp.plan_id`, and both membership subqueries group by `post_author, post_parent`, with the outer joins matching on customer **and** plan. Keying memberships off `post_parent` rather than `_product_id` also means a membership that lost its product link still counts against the right plan.

Reproduced the miss before fixing it. Seeded a reader with an active Plan A membership plus an active Plan B subscription and no Plan B membership: the old query returned `Will process 2 customers` and never surfaced them. After the change: `Will process 3 customer/plan pairs`, and a `--live` pass leaves that reader holding both memberships (`#20 plan "Plan A" active`, `#36 plan "Plan B" active`).

`collect_plan_product_ids()` became `collect_plan_product_pairs()` and returns (product, plan) pairs rather than a map, so a product that grants more than one plan does not collapse.

includes/cli/class-fix-memberships.php
adekbadek: Fixed in 003e86f. The direct-reclaim branch now resolves the product from the subscription item that grants the reclaimed membership's plan (`find_subscription_product_id_for_plan()`, matching on the membership's `post_parent`) and calls `set_product_id()` alongside `set_subscription_id()` / `set_order_id()`, so a single `--live` pass converges.

Verified against a seeded network-managed membership with no `_product_id`: after one live run, `membership #26: _product_id='12' _subscription_id='25' managed='' status=wcm-active`. The pre-fix code left `_product_id` empty until a second run.

includes/cli/class-fix-memberships.php
adekbadek: Addressed in 003e86f. `membership_statuses` is gone (it was never consumed), the remaining `GROUP_CONCAT`s use `DISTINCT` so the item-meta join fan-out no longer duplicates IDs, and the query sets `SET SESSION group_concat_max_len = 1000000` up front with a comment explaining why the lists must not truncate.

includes/cli/class-fix-memberships.php
adekbadek: Fixed in 003e86f. The catch now pushes to `$command_results['skipped']` and includes the exception message, so a failed creation shows in the summary.

Verified by forcing `wc_memberships_create_user_membership()` to throw on a seeded user: the run ends with `Skipped 1: Could not create a membership for user … : Cannot create User Membership: Content, title, and excerpt are empty.` Previously that user vanished from the summary entirely.

includes/cli/class-fix-memberships.php
adekbadek: Fixed in 003e86f - dropped from `run()`, `process_user()` and `handle_user_without_local_membership()`. The six remaining `global $wpdb;` declarations are all in methods that query directly.

includes/cli/class-fix-memberships.php
adekbadek: Partly addressed in 003e86f, with a deliberate deviation on the status filter - flagging it so you can push back.

Added the audit trail: every reactivation this command performs now writes a membership note (`add_audit_note()`), so the change is visible on the membership itself in wp-admin rather than only in the operator's stdout. Wired into all three reactivation paths - `reactivate_if_inactive()`, the relink path in `handle_user_with_inactive_membership()`, and the stale-end-date pass. Verified on a seeded env; after a `--live` run the notes read:

```
membership #23: Reactivated and relinked to subscription #22 by `wp newspack fix-memberships` (was expired).
membership #26: Reactivated by `wp newspack fix-memberships` (was cancelled): a linked subscription is active.
```

What I did **not** do is exclude statuses that could represent a deliberate revocation. WooCommerce Memberships itself sets `cancelled` (and `paused` for on-hold) when the subscription integration reacts to a subscription state change, so `cancelled` is the dominant status of exactly the drift this command exists to repair - excluding it by default would make the command a no-op on its main case, and putting it behind an opt-in flag means the flag is always passed, which buys nothing. A status filter cannot distinguish "cancelled by the integration" from "cancelled by staff"; the note does at least make the reinstatement reviewable after the fact.

Happy to add an `--exclude-status=<list>` escape hatch if you would rather operators be able to carve out a status their site uses as a revocation marker - say the word and I will.

adekbadek:
@dkoo thanks - all points addressed, replies inline on each thread. Summary of the two raised in the review body:

**Blocker - committed `node_modules` / `vendor` symlinks.** Removed in d101141. `git rm --cached node_modules vendor` and `.gitignore` now uses `/node_modules` and `/vendor` without the trailing slash so the symlink form is matched too, not just the directory. Confirmed neither path is tracked on the branch any more:

```
$ git ls-tree origin/feat/fix-memberships-cli --name-only | grep -E "^(vendor|node_modules)$"
(no output)
```

Worth noting they were not merely cosmetic: the dangling `vendor` symlink is what made `wp plugin activate newspack-plugin` die silently when I mounted this branch into a fresh container - the autoloader was simply not there.

**No automated tests for a 1000-line data-mutating command.** Agreed, and I have not added them in this PR. In the meantime I verified every change against a disposable env with seeded fixtures (multi-plan reader, expired membership, network-managed membership with no `_product_id`, three memberships carrying a stale `_end_date`, and a forced creation failure), running the pre-fix code first to confirm each bug reproduced and then the fix. Both the HPOS and non-HPOS variants of the affected-users and stale-end-date queries were executed to confirm they parse. Details in the thread replies.

Happy to follow up with PHPUnit coverage of the branch-selection logic and the two query builders as a separate PR - the query builders in particular are now pure functions of the (product, plan) pair list, which makes them straightforward to assert against.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1115 @@
...
+	 * @param array<int,array{product_id:int,plan_id:int}> $plan_product_pairs Product/plan pairs from collect_plan_product_pairs().
+	 * @param bool                                         $is_using_hpos Whether HPOS is enabled.
+	 * @return array<int,array<string,string|null>>
+	 */
+	private static function query_affected_users( $plan_product_pairs, $is_using_hpos ) {
+		global $wpdb;
+
+		// The ID lists below are GROUP_CONCAT-ed and parsed back in PHP, so a truncated
+		// list would silently feed max() a bogus ID. Raise the cap for this connection.
+		$wpdb->query( 'SET SESSION group_concat_max_len = 1000000' ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
dkoo, line 233: **Severity:** suggestion

The `SET SESSION group_concat_max_len` bump is the right instinct (so the `GROUP_CONCAT` membership-id lists parse back without silent truncation). One caveat worth a comment: on a multi-host DB layer (HyperDB/LudicrousDB, which the VIP-Go standards target), the `SET SESSION` query and the subsequent large `get_results()` SELECT are both reads and may be routed to *different* physical connections, so the raised cap can be a no-op on the connection that actually runs the GROUP_CONCAT. On Atomic/single-DB it's fine, and now that `membership_statuses` is gone and the concats are `DISTINCT` the per-(customer,plan) lists are tiny — so this is defensive rather than a live risk, just noting the assumption.

includes/cli/class-fix-memberships.php
```diff
@@ -0,0 +1,1115 @@
...
+		$stale_end_dates = $wpdb->get_results( $stale_end_dates_query, ARRAY_A );
+		// phpcs:enable WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+
+		if ( empty( $stale_end_dates ) ) {
+			return;
+		}
+
+		$found_count = count( $stale_end_dates );
+		if ( false !== $limit && $found_count > $limit ) {
+			$stale_end_dates = array_slice( $stale_end_dates, 0, $limit );
```
dkoo: **Severity:** suggestion

`--limit` is now honored here, but it's applied with `array_slice()` *after* `get_results()` has already loaded the full unbounded result set (the `posts` + two `postmeta` self-joins + users/orders scan) into memory. So `--limit` bounds how many memberships get mutated — the important blast-radius control, and correctly fixed — but not the query cost or peak memory. Pushing the limit into SQL (`LIMIT %d`) would let the flag also protect very large sites. Non-blocking, and it matches the existing shape of the first pass (line 117 slices after fetching all affected users).

dkoo (approved):
Re-reviewed the two commits since my last pass (`d101141` + `003e86f`). Both blocking concerns from my previous review are resolved.

**Symlink blocker — resolved.** The committed `node_modules` and `vendor` symlinks are gone: `git ls-files` no longer tracks either, and `.gitignore` now ignores the symlink *files* (`/node_modules`, `/vendor`, no trailing slash), so they can't slip back into the plugin zip on another checkout or in CI.

**`--limit` bounding — resolved.** `run()` now threads the limit into `fix_stale_end_dates()`, which slices its result set, and the header reports found-vs-processed ("Found N ... processing M"). Both passes are now bounded, with no off-by-one.

The per-(customer, plan) scoping rewrite is correct: both membership subqueries group by `post_author, post_parent` and both outer joins match on customer **and** plan, so a reader with a valid Plan A membership and a broken Plan B subscription now surfaces for Plan B instead of being silently suppressed. The UNION ALL product=>plan derived table is integer-cast throughout (injection-safe), the LEFT->INNER join changes are behavior-preserving, and the `set_product_id()` on the reclaim path, the `GROUP_CONCAT(DISTINCT ...)` + `group_concat_max_len` bump, the `trash` exclusion, the skipped-accounting on create failure, and the audit notes all check out.

The deliberate choice to reactivate `cancelled`/`expired` memberships rather than exclude them is reasonable — that's the dominant drift this command exists to repair, and the new audit note makes each reinstatement reviewable after the fact. The offered `--exclude-status` escape hatch would be a nice-to-have but isn't needed to merge.

Two optional follow-ups inline; neither blocks. Thanks for the thorough turnaround on the whole list.


**Manual QA (isolated env).** Verified the command end-to-end against a WooCommerce Subscriptions + Memberships site seeded with a plan/product and customers in each broken state: the dry-run correctly identified the reactivate, create, and stale-end-date cases, `--live` applied them, and the healthy control membership was left untouched.

includes/cli/class-fix-memberships.php
adekbadek: Added a comment in caa6885 noting the multi-host caveat: on HyperDB/LudicrousDB the `SET SESSION` and the SELECT may land on different physical read connections, so the bump can be a no-op there; it holds on Atomic's single DB, and with `DISTINCT` + per-(customer,plan) grouping the concatenated lists are tiny, so it stays defensive rather than load-bearing. No behaviour change - just documenting the assumption as you suggested.

includes/cli/class-fix-memberships.php
adekbadek: Fixed in caa6885. The stale-end-date pass now appends `LIMIT <int>` to the query (integer-only, no user input), so `--limit` bounds what the query materialises, not just the mutation count. To keep the found-vs-processed reporting you liked, when a limit is set it runs a cheap `SELECT COUNT(DISTINCT p.ID)` over the same FROM/WHERE for the true total - so a canary still sees how much it left untouched without loading every row.

Verified on a seeded env (3 active memberships with stale `_end_date` on active subs):
- `--limit=1` dry run: `Found 3 ... processing 1.`
- `--live --limit=2`: `Found 3 ... processing 2.`; afterwards #69/#72 have their `_end_date` cleared and #75 is untouched (still `2026-06-23 …`).

Left the first pass slicing in PHP as-is, since that query is already grouped/derived and you noted it as accepted precedent - happy to align it too if you would prefer both in SQL.

PR #4742, opened 2026-05-18
fix(reader-activation): Content Gate metadata for legacy ESP sync

miguelpeixe:
### All Submissions:

The Content Access contact-metadata fields (`NP_Content Access`, `NP_Content Access Source`, `NP_Content Access Group`) were previously only available to sites running the v1 Reader Activation sync metadata schema. Adopting v1 requires a disruptive cutover that renames/splits most legacy fields and breaks ESP segments built on the old field names.

This makes the Content Access fields available on the **legacy** sync schema as well, so sites can adopt them without the v1 cutover. The fields flow through the existing legacy sync and normalization path with no other behavior changes; field names are unchanged.

Additionally, the Content Access fields are now offered (in both the legacy and v1 schemas) **only when the Content Gate feature is enabled** (`NEWSPACK_CONTENT_GATES`). On sites without the feature, the fields no longer appear in the per-integration field selector and are not synced as permanently-empty values.

No backfill of existing ESP contacts is included — existing contacts populate the fields on their next per-contact sync (unchanged limitation).

1. On a site **without** `NEWSPACK_CONTENT_GATES` defined, on the legacy sync schema (default), open the Reader Activation ESP sync field selector and confirm there is **no** "Content Access" group/fields.
2. Define `define( 'NEWSPACK_CONTENT_GATES', true );` in `wp-config.php`. Reload the field selector and confirm a "Content Access" group now appears with the three fields, on the **legacy** schema (no `NEWSPACK_SYNC_METADATA_VERSION*` constant needed).
3. Create a published Content Gate with custom access active (e.g. an email-domain rule the test reader matches). Enable the "Content Access" field for the ESP integration.
4. Trigger a contact sync for a reader that passes the gate; confirm the ESP receives `NP_Content Access` = `Yes` (and Source/Group as applicable).
5. Confirm a reader that does not pass the gate syncs `NP_Content Access` = `No`.
6. Disable the "Content Access" field in the field selector; confirm it is no longer included in the sync payload (legacy basic/payment fields continue to sync unchanged).
7. Repeat steps 2-6 with the v1 schema (`define( 'NEWSPACK_SYNC_METADATA_VERSION_1', true );`) to confirm v1 behavior is unchanged aside from the shared feature-flag gating.

### Other information:

### DIFF
--- a/includes/reader-activation/sync/class-metadata.php
+++ b/includes/reader-activation/sync/class-metadata.php
@@ -47,6 +47,7 @@ class Metadata {
 			$classes = [
 				'Legacy_Basic',
 				'Legacy_Payment',
+				'Content_Gate',
 			];
 		} else {
 			$classes = [
--- a/includes/reader-activation/sync/class-metadata.php
+++ b/includes/reader-activation/sync/class-metadata.php
@@ -38,7 +38,7 @@ class Metadata {
 	/**
 	 * Get the metadata classes to be used for syncing contact metadata to the ESP.
 	 *
-	 * These are the metadata classes that will be used in case get_version is not legacy.
+	 * These are the metadata classes that will be used to build the full set of contact metadata fields.
 	 *
 	 * @return array List of metadata classes.
 	 */
--- a/includes/reader-activation/sync/contact-metadata/class-content-gate.php
+++ b/includes/reader-activation/sync/contact-metadata/class-content-gate.php
@@ -8,6 +8,8 @@
 namespace Newspack\Reader_Activation\Sync\Contact_Metadata;
 
 use Newspack\Reader_Activation\Sync\Contact_Metadata;
+use Newspack\Reader_Activation\Sync\Legacy_Metadata;
+use Newspack\Reader_Activation\Sync\Metadata;
 use Newspack\Access_Rules;
 use Newspack\Content_Gate as Content_Gate_CPT;
 use Newspack\Group_Subscription;
@@ -78,26 +80,36 @@ class Content_Gate extends Contact_Metadata {
 
 		$custom_access_gates = self::get_custom_access_gates();
 
-		// No custom access gates configured — nothing to evaluate.
 		if ( empty( $custom_access_gates ) ) {
-			return [
+			$metadata = [
 				'Content_Access'        => '',
 				'Content_Access_Source' => '',
 				'Content_Access_Group'  => '',
 			];
+		} else {
+			$evaluations = [];
+			foreach ( $custom_access_gates as $gate ) {
+				$evaluations[] = User_Gate_Access::evaluate_gate_for_user( $gate, $this->user->ID );
+			}
+
+			$user_id  = $this->user->ID;
+			$metadata = [
+				'Content_Access'        => self::has_content_access( $evaluations ) ? 'Yes' : 'No',
+				'Content_Access_Source' => implode( ', ', self::collect_labels( $evaluations, $user_id, [ self::class, 'get_source_labels' ] ) ),
+				'Content_Access_Group'  => implode( ', ', self::collect_labels( $evaluations, $user_id, [ self::class, 'get_group_labels' ] ) ),
+			];
 		}
 
-		$evaluations = [];
-		foreach ( $custom_access_gates as $gate ) {
-			$evaluations[] = User_Gate_Access::evaluate_gate_for_user( $gate, $this->user->ID );
+		// In legacy mode the main sync path does not run a normalize step on
+		// the merged contact, so each metadata class must return keys in the
+		// prefixed shape (matching Legacy_Basic / Legacy_Payment). Without this,
+		// raw Content_Access keys are silently dropped at the ESP push.
+		if ( 'legacy' === Metadata::get_version() ) {
+			$normalized = Legacy_Metadata::normalize_contact_data( [ 'metadata' => $metadata ] );
+			return $normalized['metadata'] ?? [];
 		}
 
-		$user_id = $this->user->ID;
-		return [
-			'Content_Access'        => self::has_content_access( $evaluations ) ? 'Yes' : 'No',
-			'Content_Access_Source' => implode( ', ', self::collect_labels( $evaluations, $user_id, [ self::class, 'get_source_labels' ] ) ),
-			'Content_Access_Group'  => implode( ', ', self::collect_labels( $evaluations, $user_id, [ self::class, 'get_group_labels' ] ) ),
-		];
+		return $metadata;
 	}
 
 	/**
--- a/includes/reader-activation/sync/contact-metadata/class-content-gate.php
+++ b/includes/reader-activation/sync/contact-metadata/class-content-gate.php
@@ -41,7 +41,7 @@ class Content_Gate extends Contact_Metadata {
 	 * @return boolean
 	 */
 	public static function is_available() {
-		return true;
+		return Content_Gate_CPT::is_newspack_feature_enabled();
 	}
 
 	/**
--- a/includes/reader-activation/sync/contact-metadata/class-content-gate.php
+++ b/includes/reader-activation/sync/contact-metadata/class-content-gate.php
@@ -52,7 +52,7 @@ class Content_Gate extends Contact_Metadata {
 	 * @return string
 	 */
 	public static function get_section_name() {
-		return __( 'Content Access', 'newspack' );
+		return __( 'Content Access', 'newspack-plugin' );
 	}
 
 	/**
--- a/tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
+++ b/tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
@@ -141,6 +141,7 @@ class Test_Content_Gate_Legacy extends WP_UnitTestCase {
 		$this->assertArrayHasKey( 'Content_Access', $fields, 'Legacy schema should expose the Content_Access field.' );
 		$this->assertSame( 'Content Access', $fields['Content_Access'] );
 		$this->assertArrayHasKey( 'Content_Access_Source', $fields );
+		$this->assertArrayHasKey( 'Content_Access_Group', $fields );
 	}
 
 	public function test_is_available_follows_content_gate_feature_flag() {
@@ -273,7 +274,7 @@ class Test_Content_Gate_Legacy extends WP_UnitTestCase {
 		);
 	}
 
-	public function test_institution_rule_path_yields_group_source() {
+	public function test_institution_rule_path_yields_institution_source() {
 		$institution_id = Institution::create( 'Test University', '', [ 'email_domain' => 'example.com' ] );
 		$this->assertNotInstanceOf( WP_Error::class, $institution_id );
 
@@ -294,9 +295,9 @@ class Test_Content_Gate_Legacy extends WP_UnitTestCase {
 
 		$this->assertSame( 'Yes', $normalized['metadata']['NP_Content Access'] );
 		$this->assertSame(
-			'group',
+			'institution',
 			$normalized['metadata']['NP_Content Access Source'],
-			'Institution-rule passes must yield the "group" source label.'
+			'Institution-rule passes must yield the "institution" source label.'
 		);
 	}
 }
--- a/tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
+++ b/tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
@@ -6,6 +6,7 @@
  */
 
 use Newspack\Content_Gate;
+use Newspack\Institution;
 use Newspack\Reader_Activation;
 use Newspack\Reader_Activation\Sync\Metadata;
 use Newspack\Reader_Activation\Sync\Contact_Metadata\Content_Gate as Content_Gate_Metadata;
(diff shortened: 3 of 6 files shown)


dkoo (changes requested):
Changeset is tiny and clean. However, manual and unit testing fail for a couple of reasons:

### 🔴 Missing `Content_Access_Group` field

The failing unit test is because the `Content_Access_Group` field isn't in the parent `release` branch yet. This field was added in #4697, which hasn't yet made it to `release`.

```
Test_Content_Gate_Legacy::test_legacy_schema_exposes_content_access_fields
Failed asserting that an array has the key 'Content_Access_Group'.
```

**Suggested fix:** I opened #4746 to cherry-pick that PR to `release` as a hotfix. Once that PR lands we can rebase this branch on `release` to resolve things. Alternatively, you can just cherry-pick that same commit hash `8d71495` to this branch since it's also a hotfix, and then I'll close #4746.

### 🔴 Unprefixed meta fields get silently dropped before the sync

Unlike `Legacy_Basic` and `Legacy_Payment` classes, `Content_Gate` doesn't normalize its own fields using `Legacy_Metadata::normalize_contact_data()` before appending its fields, so the `Content Access` fields get appended without `NP_` (or custom) prefix. This causes the fields to be silently dropped from the contact before getting synced to the ESP.

**Suggested fix:** Normalize all fields, either by adding a handler to prefix `Content_Gate` fields before returning for a legacy sync, or by moving `Legacy_Metadata::normalize_contact_data()` outside of the individual `Legacy_*` classes and to a later point where it can run on all fields collected for a legacy sync but before it gets passed to the sync handler.

includes/reader-activation/sync/class-metadata.php
```diff
@@ -47,6 +47,7 @@ protected static function get_metadata_classes() {
 			$classes = [
 				'Legacy_Basic',
 				'Legacy_Payment',
+				'Content_Gate',
```
dkoo, line 50: ### 🟡 v1 `prepare_contact()` accepts prefixed keys without checking current availability

`class-integration.php:733` admits already-prefixed keys based on enabled-field names only, not the live availability map. If an external filter/contact includes `NP_Content Access` while the flag is off (and the field remains in saved settings from a prior flag-on period), it can pass through. **Suggested fix:** also require membership in the current `Sync\Metadata::get_keys()` values.

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
```diff
@@ -0,0 +1,157 @@
...
+			]
+		);
+		Reader_Activation::set_reader_verified( self::$user_id );
+	}
+
+	public function tear_down() {
+		Metadata::$version = self::$original_version;
+		Content_Gate_Metadata::reset_cache();
+		parent::tear_down();
+	}
```
dkoo, line 79: ### 🟡 Test mutates enabled outgoing fields without restoring in `tear_down()`

Legacy test suite calls `Metadata::update_fields()` but never restores prior state. Risk of order-dependent failures in the broader suite. **Suggested fix:** snapshot in `set_up`, restore in `tear_down`.

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
```diff
@@ -0,0 +1,157 @@
...
+	 *
+	 * @var int
+	 */
+	private static $user_id;
+
+	public static function set_up_before_class() {
+		parent::set_up_before_class();
+		require_once dirname( __DIR__, 2 ) . '/mocks/wc-mocks.php';
+		if ( ! defined( 'NEWSPACK_CONTENT_GATES' ) ) {
+			define( 'NEWSPACK_CONTENT_GATES', true );
```
dkoo, line 55: ### 🟡 `NEWSPACK_CONTENT_GATES` define leaks for the entire PHPUnit process

PHP defines cannot be unset, so any future "flag-off" test in the same process would be silently neutralized. **Suggested fix:** The existing test at https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-legacy-metadata/tests/unit-tests/content-gate/content-gates.php#L46-L59 explains the leak risk — mirror that, or use process isolation (@runInSeparateProcess).

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
```diff
@@ -0,0 +1,157 @@
...
+		$contact    = Metadata::get_contact_with_metadata( self::$user_id );
+		$normalized = Metadata::normalize_contact_data( $contact );
+
+		$this->assertArrayNotHasKey(
+			'NP_Content Access',
+			$normalized['metadata'],
+			'Content Access must be dropped when not enabled for the integration.'
+		);
+	}
+}
```
dkoo, line 303: ### 🟡 Test coverage gaps

Not covered: v1-schema parity (all tests force `Metadata::$version = 'legacy'`), flag-off behavior, no-gate-matches case (`Content_Access = 'No'`), prefixed field normalization, multiple gates, institution and group subscription rules path (which would populate the missing `Group` field), subscription-rule path (the WC product-name lookup).

dkoo (commented):
Additional feedback in inline comments. Also, this doesn't originate in this PR, but the `includes/reader-activation/sync/contact-metadata/class-content-gate.php` file has incorrect `newspack` text domains throughout—let's find/replace with `newspack-plugin` to keep it in line with the rest of the plugin.

includes/reader-activation/sync/class-metadata.php
miguelpeixe: Addressed in db0df29 — `prepare_contact()` now requires the unprefixed field name to also be in the live `Sync\Metadata::get_keys()` values, so stale prefixed keys from a prior flag-on period are dropped. Added `Test_Prepare_Contact::test_already_prefixed_stale_enabled_field_filtered` to cover the case.

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
miguelpeixe: Addressed in db0df29 — `set_up` now snapshots `Metadata::get_fields()` and `tear_down` restores it via `Metadata::update_fields()`.

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
miguelpeixe: Addressed in db0df29 — `set_up_before_class` now carries a docblock mirroring the one in `Test_Content_Gates`, explaining that the `NEWSPACK_CONTENT_GATES` define leaks for the rest of the process and that any future flag-off test must use `@runInSeparateProcess`.

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
miguelpeixe: Partially addressed in db0df29. Added: v1-schema parity (`test_v1_schema_also_exposes_content_access_fields`), no-gate-matches case (`test_no_gate_matches_yields_content_access_no`), prefixed normalization (`test_legacy_normalize_keeps_already_prefixed_content_access_key`), multiple gates (`test_multiple_gates_aggregate_source_labels`), and institution path (`test_institution_rule_path_yields_group_source`). Also dropped the broken `Content_Access_Group` assertion that was failing CI — the field doesn't exist on this branch. Skipped: flag-off behavior (define leak prevents in-process testing — would need `@runInSeparateProcess`); subscription-rule path (would require WC subscription mocks beyond the existing `wc_get_product` stub — happy to add if you'd like).

miguelpeixe:
@dkoo Thanks for the review!

> **🔴 Unprefixed Content_Gate fields silently dropped**

Fixed in f3077f1. `Content_Gate::get_metadata()` now runs its output through `Legacy_Metadata::normalize_contact_data()` when `Metadata::get_version() === 'legacy'`, returning prefixed keys to match `Legacy_Basic` / `Legacy_Payment`. Added `Test_Content_Gate_Legacy::test_content_gate_fields_arrive_prefixed_from_get_contact_with_metadata` to lock in the wire-format expectation (the existing test had the wrong path because it called `Metadata::normalize_contact_data()` explicitly — production doesn't in legacy mode). Updated `Test_Content_Gate_Metadata` to force `Metadata::$version = '1.0'` so its raw-shape assertions still hold; the legacy wire format is now tested in `Test_Content_Gate_Legacy` instead.

> **🔴 Missing `Content_Access_Group` field**

I worked around the CI failure by dropping the assertion (so other fixes could land green). Once #4746 lands on `release` I'll rebase this branch and restore the `Content_Access_Group` assertion.

tests/unit-tests/reader-activation-sync/class-test-content-gate-legacy.php
dkoo: Thanks! You're clear to add back the `Content_Access_Group` test now that #4746 has landed.

dkoo (commented):
@miguelpeixe latest changes look good, thanks for the updates! I'll hold off approving until merge conflicts are resolved.

miguelpeixe:
@dkoo Rebased onto `release` now that #4746 landed — `Content_Access_Group` is on `release`, so:

- Restored the `Content_Access_Group` assertion in `Test_Content_Gate_Legacy::test_legacy_schema_exposes_content_access_fields` (was the workaround drop from before).
- Resolved conflicts in `includes/reader-activation/sync/contact-metadata/class-content-gate.php` (kept release's 3-key `collect_labels` shape for v1 and stacked the legacy normalize on top) and `tests/.../class-content-gate-metadata.php` (merged the version-snapshot setup with the new group/institution test scaffolding).
- Updated the institution-path test to expect the new `'institution'` source label (release renamed it from `'group'` in #4697).

54/54 tests in the `Content_Gate_Metadata`, `Content_Gate_Legacy`, and `prepare_contact` groups pass locally. New tip: cf33f4b2f.

includes/reader-activation/sync/contact-metadata/class-content-gate.php
```diff
@@ -78,26 +80,36 @@ public function get_metadata() {
...
 
-		$evaluations = [];
-		foreach ( $custom_access_gates as $gate ) {
-			$evaluations[] = User_Gate_Access::evaluate_gate_for_user( $gate, $this->user->ID );
+		// In legacy mode the main sync path does not run a normalize step on
+		// the merged contact, so each metadata class must return keys in the
+		// prefixed shape (matching Legacy_Basic / Legacy_Payment). Without this,
+		// raw Content_Access keys are silently dropped at the ESP push.
+		if ( 'legacy' === Metadata::get_version() ) {
+			$normalized = Legacy_Metadata::normalize_contact_data( [ 'metadata' => $metadata ] );
```
dkoo, line 108: ### 🟢 Nit: `newspack_esp_sync_normalize_contact` filter now fires multiple times per sync with partial contact payloads

Because `Content_Gate::get_metadata()` now also internally calls `Legacy_Metadata::normalize_contact_data()` ([class-legacy-metadata.php:267](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-legacy-metadata/includes/reader-activation/sync/class-legacy-metadata.php#L267) — `apply_filters( 'newspack_esp_sync_normalize_contact', $contact )`), the filter fires once per metadata class that internally normalizes (Legacy_Basic + Content_Gate), each time receiving only that class's partial metadata (no email/name). On the retry path it fires a third time on the merged contact.

**Suggestion:** since [`Metadata::normalize_contact_data()`](https://github.com/Automattic/newspack-plugin/blob/feat/content-gate-legacy-metadata/includes/reader-activation/sync/class-metadata.php#L436) already defers to `Legacy_Metadata::normalize_contact_data()` if `Metadata::get_version()` returns `legacy`, we could pass merged contact data through this method once instead of per class. This would also result in cleaner logs as currently each class that calls `Legacy_Metadata::normalize_contact_data()` will log its output separately. Non-blocking because the end result is the same, though.

includes/reader-activation/sync/class-metadata.php
dkoo: ### 🟢 Nit: docblock for `get_metadata_classes()` is now stale

Docblock says *"used in case get_version is not legacy"* but the array now branches both ways. Update to reflect version-dependent list:

```suggestion
	 * These are the metadata classes that will be used to build the full set of contact metadata fields.
```

dkoo (approved):
Passes manual testing, and I see all three `NP_Content Access` fields being synced now. A couple of non-blocking nitpicks inline, but nothing urgent.

includes/reader-activation/sync/class-metadata.php
miguelpeixe: Addressed in 3f29a4f — applied the suggested docblock.

miguelpeixe:
@dkoo Thanks for the approval and the follow-up nits!

**Text domain fix** — Done in 3f29a4f. Find/replaced `'newspack'` → `'newspack-plugin'` in `class-content-gate.php` (only one occurrence on this branch, at the `get_section_name()` translation).

**`newspack_esp_sync_normalize_contact` fires multiple times** — Agreed this is the architecturally cleaner shape, but I'd like to defer it to a follow-up PR since it touches `Legacy_Basic` / `Legacy_Payment` (currently the only legacy normalizers) and changes where the merged contact gets normalized in the main sync path. The fix is mechanical but it's outside the scope of "unblock Content Gate fields for legacy sync" — keeping it separate makes the bisect/revert story cleaner if anything regresses. I'll open the follow-up after this lands and tag you.

dkoo (approved):
Sounds good! Thanks, @miguelpeixe

PR #4743, opened 2026-05-19
fix(wc-memberships): skip expiry filter for non-subscription memberships

adekbadek:
### All Submissions:

`Membership_Expiry::prevent_membership_expiration()` is hooked into `wc_memberships_expire_user_membership`, which fires for **every** user membership. The handler assumed every membership was a subscription-tied one (`\WC_Memberships_Integration_Subscriptions_User_Membership`) and called `get_subscription_id()` on it. For plain `WC_Memberships_User_Membership` objects (no linked subscription), that method does not exist, so the call throws `Uncaught Error: Call to undefined method WC_Memberships_User_Membership::get_subscription_id()`.

This adds an early `method_exists()` guard that returns the incoming `$cancel_membership` value unchanged when the membership is not subscription-tied, so the filter is effectively a no-op for memberships it cannot reason about. The docblock type for `$user_membership` is also widened from the integration class to the base class to match reality.

1. With WC Memberships active **but without WC Subscriptions** (or with a manually-granted membership that has no linked subscription), expire that membership in the admin or via `wp wc memberships ...`. Confirm no PHP fatal is logged.
2. With WC Subscriptions active and a subscription-tied membership, cancel the underlying subscription. Confirm the existing behavior still works: if another active subscription for the same product exists, the membership is reassigned and not expired; otherwise it expires normally.
3. Run the included unit test: `composer test -- --filter test_passes_through_non_subscription_membership`. It must pass; without the fix it errors with `Call to undefined method ... ::get_subscription_id()`.
4. Run the full plugin test suite to confirm no regressions: `composer test`.

### Other information:

### DIFF
--- a/includes/plugins/wc-memberships/class-membership-expiry.php
+++ b/includes/plugins/wc-memberships/class-membership-expiry.php
@@ -28,11 +28,16 @@ class Membership_Expiry {
 	/**
 	 * Prevent membership expiration if there are other active subscriptions.
 	 *
-	 * @param bool                                                      $cancel_membership Whether to cancel the membership when the subscription is cancelled (default true).
-	 * @param \WC_Memberships_Integration_Subscriptions_User_Membership $user_membership The subscription-tied membership.
+	 * @param bool                            $cancel_membership Whether to cancel the membership when the subscription is cancelled (default true).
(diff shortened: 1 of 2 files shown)


jason10lee (approved):
Approved!  Targeted fix, tests clean, even adds to the test suite.  (Thank you!)

PR #4747, opened 2026-05-20
fix: prevent recursion in lite-site primary color on block themes

chickenn00dle:
### All Submissions:

* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and VIP Go coding standards?

`Lite_Site::get_primary_color()` calls `wp_get_global_settings()` when running on a block theme, which re-fires the `wp_theme_json_data_theme` filter. If the lite-site code is itself executing inside that filter, this caused unbounded recursion and a fatal error.

This adds a re-entry guard so the method returns `currentcolor` when invoked while `wp_theme_json_data_theme` is already on the filter stack, breaking the loop while still producing a valid CSS value.

1. Activate a block theme (e.g. `newspack-block-theme`) on a Newspack site with the Lite Site feature enabled.
2. Load any front-end page and confirm there are no PHP fatal errors or "maximum function nesting level" warnings.
3. Verify the lite-site styling still renders with the expected primary color in the resulting CSS.
4. Switch back to a classic theme (e.g. `newspack-theme`) and confirm the primary color resolution still works unchanged.

### Other information:

### DIFF
--- a/includes/lite-site/class-lite-site.php
+++ b/includes/lite-site/class-lite-site.php
@@ -375,6 +375,10 @@ class Lite_Site {
 	 */
 	public static function get_primary_color() {
 		if ( wp_is_block_theme() ) {
+			// Guard against re-entry via wp_theme_json_data_theme — wp_get_global_settings() re-fires that filter and recurses.
+			if ( doing_filter( 'wp_theme_json_data_theme' ) ) {
+				return 'currentcolor';
+			}
 			$settings = wp_get_global_settings();
 			$palettes = $settings['color']['palette'] ?? [];
 			$palette  = [];


laurelfulford (approved):
Thanks @chickenn00dle! 

This fix looks good, and stops the recursion issue in its tracks! 

Claude had two non-blocking potentially enhancement-like suggestions for the future:

> Fallback is `'currentcolor'` — a CSS keyword, not a hex. It's valid as a CSS value but unusual for a "primary color" return. It does mean the newsletter button on the very first new-newsletter page load gets `currentcolor` instead of #003DA5 (the theme's `palette[0]`). For subsequent requests the cached `wp_get_global_settings()` result probably means it resolves correctly. So this is mostly a cosmetic loss on the initial filter pass.

I think this goes beyond the scope of the fix and is something we can fix in the newsletter directly if we need to, with something like [this](https://github.com/Automattic/newspack-newsletters/compare/fix/oom-block-theme-issue?expand=1). 

> Guard only covers `wp_theme_json_data_theme` — `wp_get_global_settings()` also fires  `wp_theme_json_data_user` and `wp_theme_json_data_blocks`. If anyone hooks `get_primary_color()` from one of those in the future, recursion returns. Could broaden to all three or use a static `$resolving` flag. But "the actual filter that bit us" is fine for now.

This feels like looking for trouble right now - maybe something to harden in the future if we think we'll need it, but not the problem at hand right now 🙂 

Anyway, all that to say 🚢

PR #4750, opened 2026-05-21
fix(editor): exclude Customizer Additional CSS from iframed editor

miguelpeixe:
### All Submissions:

WordPress 7.0 iframes the post editor whenever every block in the post uses Block API v3+. The iframed editor pulls in the front-end style pipeline, which includes the Customizer's **Additional CSS**. Any front-end rule that targets \`body { ... }\` (or other broad selectors) therefore lands on the editor canvas itself, distorting the editing UX on sites that have customized the body element.

This PR filters `block_editor_settings_all` to drop the specific entry WP core injects for Customizer Additional CSS — identified by `__unstableType === 'user'` AND `isGlobalStyles === false`. The front-end is unaffected: the CSS is still printed by `wp_custom_css_cb()` on `wp_head`.

A new `newspack_strip_customizer_css_in_editor` filter is exposed as an escape hatch for sites that intentionally want Customizer CSS in the editor.

Scope notes:
- Only the Customizer entry is removed. The block-theme **Global Styles → Additional CSS** entry (`isGlobalStyles === true`) is left untouched.
- Per-block Custom CSS introduced in WP 7.0 is not affected — that's scoped to individual blocks.

1. On a site running WP 7.0, add a rule like `.wp-block-quote { background: red; }` in **Appearance → Customize → Additional CSS**.
2. Draft a post  and add a quote block
3. Before this fix on `release`: the block turns red. After this fix: the block stays normal; the front-end is still red.
4. Confirm the escape hatch: add `add_filter( 'newspack_strip_customizer_css_in_editor', '__return_false' );` in mu-plugins → editor canvas turns red again.

### Other information:

* [x] Have you successfully ran tests with your changes locally? (PHPCS clean)

### DIFF
--- a/includes/class-patches.php
+++ b/includes/class-patches.php
@@ -44,6 +44,9 @@ class Patches {
 
 		// Fix an issue when running The Events Calendar where all posts block items have same date.
 		add_action( 'tribe_events_views_v2_after_make_view', [ __CLASS__, 'remove_tec_extra_excerpt_filtering' ], 1 );
+
+		// Prevent Customizer "Additional CSS" from leaking into the iframed block editor in WP 7.0+.
+		add_filter( 'block_editor_settings_all', [ __CLASS__, 'strip_customizer_css_from_editor' ], 999 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-patches.php
```diff
@@ -519,6 +524,49 @@ public static function remove_tec_extra_excerpt_filtering() {
...
+			return $settings;
+		}
+
+		$settings['styles'] = array_values(
+			array_filter(
+				$settings['styles'],
+				function ( $entry ) {
+					return ! (
+						isset( $entry['__unstableType'] )
+						&& 'user' === $entry['__unstableType']
```
laurelfulford, line 561: Nit: I think WordPress tends to leave a lot of 'unstable' and experimental' stuff in the code, but it might be good to flag in a comment just in case it changes down the road! (like ` // Note: __unstableType is a private WP core key. Re-verify on WP upgrades.`) 

A test could also help, but I think that's beyond the scope of this PR!

laurelfulford (approved):
Thanks @miguelpeixe! 

On trunk, I applied this CSS snippet, which is representative of some CSS that has caused issues (but WAY uglier):

```
body:after {
   position: fixed;
   content: "";
   background: pink;
   left: 0;
   right: 0;
   bottom: 0;
   z-index: 10000;
}

body {
    background: linear-gradient(90deg,rgba(42, 123, 155, 1) 0%, rgba(87, 199, 133, 1) 50%, rgba(237, 221, 83, 1) 100%);
}
```

It created this masterpiece:

Image

But on this branch, it looks back to normal! 

Image

I added one small nit that's more future-facing -- it's definitely not a problem that needs to be address right now! 🚢

includes/class-patches.php
miguelpeixe: Thanks, @laurelfulford! Added a note in 1d318af.

PR #4755, opened 2026-05-26
fix(advertising): stabilize placement button width, add a11y labels

thomasguillot:
### All Submissions:

Two small UX/accessibility refinements to the Ads placement cards (Advertising › Display Ads › Placements):

- **No more width jump on the Edit/Cancel toggle.** Previously the button visibly resized when its label switched from "Edit" to "Cancel", causing a layout shift in the card header. The button now reserves the width of the longer label via a CSS grid stack — both labels live in the DOM, and only the inactive one is `visibility: hidden`. Width is derived from the actual rendered text (no magic pixel value), so the fix holds across translations.
- **Screen reader context for placement actions.** "Edit", "Cancel", and "Enable" buttons repeat once per placement card, with no audible difference between them. Each button now gets an `aria-label` that includes the placement name — e.g. "Edit Sticky Footer", "Cancel editing Sticky Footer", "Enable Sticky Footer". The visible compact text is unchanged.

1. Go to **Advertising › Display Ads › Placements** (`/wp-admin/admin.php?page=newspack-ads-display-ads#/placements`).
2. On any enabled placement (e.g. Global Above Header), click **Edit**. Verify the button text swaps to **Cancel** without the button width changing or surrounding header elements shifting.
3. Click **Cancel**. Same check in reverse — no width jump.
4. Open the same page with a screen reader (VoiceOver, NVDA, etc.) and tab through the action buttons. Confirm each button is announced with its placement name appended, e.g. "Edit Global Above Header, button" and "Enable Sticky Footer, button".
5. For an enabled placement, expand it and confirm the announced label switches to "Cancel editing &lt;placement name&gt;".
6. Confirm no regression on the disabled placement state — **Enable** button still toggles the placement on as before.

### Other information:

### DIFF
--- a/src/wizards/advertising/views/placements/index.js
+++ b/src/wizards/advertising/views/placements/index.js
@@ -189,6 +189,13 @@ const Placements = () => {
 							hasAdUnit = Object.keys( placement.hooks ).every( hookKey => !! placement.data?.hooks?.[ hookKey ]?.ad_unit );
 						}
 
+						/* translators: %s: placement name (e.g. "Sticky Footer"). */
+						const editButtonLabel = sprintf( __( 'Edit %s', 'newspack-plugin' ), placement.name );
+						/* translators: %s: placement name (e.g. "Sticky Footer"). */
+						const cancelButtonLabel = sprintf( __( 'Cancel editing %s', 'newspack-plugin' ), placement.name );
+						/* translators: %s: placement name (e.g. "Sticky Footer"). */
+						const enableButtonLabel = sprintf( __( 'Enable %s', 'newspack-plugin' ), placement.name );
+
 						return (
 							<CardForm
 								key={ key }
@@ -204,6 +211,7 @@ const Placements = () => {
 										<Button
 											variant="tertiary"
 											size="compact"
+											aria-label={ isEditing ? cancelButtonLabel : editButtonLabel }
 											disabled={ inFlight || ( !! editingPlacement && ! isEditing ) }
 											onClick={ () => {
 												if ( isEditing ) {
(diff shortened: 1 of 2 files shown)


src/wizards/advertising/views/placements/index.js
```diff
@@ -214,12 +222,20 @@ const Placements = () => {
 												}
 											} }
 										>
-											{ isEditing ? __( 'Cancel', 'newspack-plugin' ) : __( 'Edit', 'newspack-plugin' ) }
+											<span style={ { display: 'inline-grid', justifyItems: 'center' } }>
```
adekbadek: **Nit:** The width-stabilization uses three inline `style` objects, where the rest of this view expresses presentational styling via `classnames` + SCSS. A small class (e.g. `.newspack-wizard-ads-placement__toggle-label` with the grid stack) would match house style, be reusable, and go through autoprefixer/lint. Non-blocking for a 17-line fix.

src/wizards/advertising/views/placements/index.js
```diff
@@ -204,6 +211,7 @@ const Placements = () => {
 										<Button
 											variant="tertiary"
 											size="compact"
+											aria-label={ isEditing ? cancelButtonLabel : editButtonLabel }
```
adekbadek, line 215: **Suggestion:** Worth a one-line comment noting this `aria-label` is load-bearing for the button's accessible name. Because the grid stack keeps both "Cancel" and "Edit" in the DOM (only `visibility: hidden`, which prunes the hidden span from the accessibility tree), the `aria-label` is the only thing producing a clean name today — if a future edit drops it, the accessible name silently becomes the concatenated `CancelEdit`. (WCAG 2.5.3 Label-in-Name is satisfied now: each label contains its visible word.)

adekbadek (approved):
OK

src/wizards/advertising/views/placements/index.js
thomasguillot: Done in e150a5e79 — moved the grid stack into `.newspack-wizard-ads-placements__toggle-label` in style.scss and switched the spans to `classnames`. Used the file's existing plural `placements__` prefix to match `__snackbar`.

src/wizards/advertising/views/placements/index.js
thomasguillot: Added in e150a5e79 — a one-line comment above the `aria-label` noting it's load-bearing for the button's accessible name.

PR #4757, opened 2026-05-26
feat(integrations): details modal and Run now for integration logs

chickenn00dle:
### All Submissions:

* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and VIP Go coding standards?

Adds two affordances to the per-integration activity logs view (Audience › Integrations › Logs) to make troubleshooting Action Scheduler retries practical:

- **View details** — opens a modal showing the full action payload (decoded `args`, with `extended_args` preferred when the indexed column was truncated), metadata (action ID, group, scheduled date, attempts, last attempt), and the per-action Action Scheduler log trail.
- **Run now** — for `pending` actions only, runs the Action Scheduler action immediately and refreshes the row with the post-run status. Mirrors WooCommerce's default AS admin behavior.

Both REST endpoints sit behind the existing wizard permission gate. The detail endpoint validates that the action belongs to the integration before responding, returning a generic 404 on cross-group access so other-group actions can't be probed through it. Errors thrown by the action's callback during a Run are surfaced as a 200 response with `status: 'failed'` and the last AS log message — AS itself records the failure and writes the log, so the UI just refetches and shows the result in a snackbar.

Follow-up to #4671 per Miguel's non-blocking review feedback.

Closes DSGNEWS-160.

Screenshot 2026-05-26 at 17 40 36

1. Spin up an isolated env: `n env create dsgnews-160 --worktree newspack-plugin:feat/improve-integrations-activity-logs --up && n setup --env dsgnews-160 --yes --woocommerce`.
2. Visit Newspack › Audience › Integrations and open the **Logs** view for an integration that has scheduled actions (Mailchimp, ActiveCampaign, etc.). If you need to generate fixtures, register a reader or trigger a data event so the integration enqueues actions.
3. On a `pending` row, open the row "…" menu → **View details**. Verify the modal shows: event label + status badge, action ID, scheduled (GMT), group, attempts, last attempt (or `—`), pretty-printed args JSON in a `<pre>` block, and the AS log list.
4. On a `pending` row, click **Run now**. Verify a success snackbar appears ("Action completed.") and the row's status flips to **Success** in place.
5. On a row whose hook callback fails (or temporarily break a callback to force one), click **Run now**. Verify a failure snackbar shows the AS log message and the row flips to **Failed**.
6. On any `complete`, `failed`, or `canceled` row, confirm **Run now** is hidden from the row menu but **View details** still opens correctly.
7. Force a not-found state: open **View details**, then rewrite the `action_id` segment of the request in the devtools network tab and re-issue. The modal should show "This action no longer exists."
8. If you can catch an `in-progress` action in the list, confirm its status badge renders as "In progress" (no raw-string fallback).

### Other information:

Automated tests were intentionally not added — Action Scheduler is heavy to mock for low value here. Verification is via the manual QA steps above in an isolated env. PHP/JS/SCSS linting and `n build` all pass.

### DIFF
--- a/src/wizards/audience/views/integrations/logs-view.js
+++ b/src/wizards/audience/views/integrations/logs-view.js
@@ -24,13 +24,14 @@ const DEFAULT_VIEW = {
 	perPage: 25,
 	sort: { field: 'timestamp', direction: 'desc' },
 	search: '',
-	fields: [ 'timestamp', 'event', 'status' ],
+	fields: [ 'timestamp', 'email', 'event', 'status' ],
 	filters: [],
 	layout: {
 		styles: {
-			timestamp: { width: '75%' },
-			event: { width: '15%' },
-			status: { width: '10%' },
+			timestamp: { width: '35%' },
+			email: { width: '30%' },
+			event: { width: '20%' },
+			status: { width: '15%' },
 		},
 	},
 };
@@ -38,7 +39,7 @@ const DEFAULT_VIEW = {
 export const LogsView = ( { integrations, match } ) => {
 	const integrationId = match?.params?.integrationId;
 	const integration = integrationId ? integrations[ integrationId ] : null;
-	const { addNotice, setHeaderData } = useDispatch( WIZARD_STORE_NAMESPACE );
+	const { addNotice, removeNotice, setHeaderData } = useDispatch( WIZARD_STORE_NAMESPACE );
 
 	const [ data, setData ] = useState( [] );
 	const [ total, setTotal ] = useState( 0 );
@@ -116,6 +117,12 @@ export const LogsView = ( { integrations, match } ) => {
 				enableSorting: true,
 			},
 			{
+				id: 'email',
+				label: __( 'Email', 'newspack-plugin' ),
+				render: ( { item } ) => item.email || '—',
+				enableSorting: false,
+			},
+			{
 				id: 'event',
 				label: __( 'Event', 'newspack-plugin' ),
 				getValue: ( { item } ) => item.event,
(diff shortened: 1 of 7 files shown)


src/wizards/audience/views/integrations/logs-view.js
```diff
@@ -158,6 +144,77 @@ export const LogsView = ( { integrations, match } ) => {
...
+						message = response.message || __( 'Action failed.', 'newspack-plugin' );
+					} else {
+						message = response.message || __( 'Action processed.', 'newspack-plugin' );
+					}
+					addNotice( {
+						message,
+						type: response.status === 'failed' ? 'error' : 'success',
+						id: `integration-action-run-${ actionId }`,
+					} );
+				} )
```
miguelpeixe, line 189: Non-blocking: The "Running action…" notice isn't replaced by the success/failure result because `addNotice` appends without deduplication by id. We should dispatch removeNotice(id) before the result notice.

miguelpeixe (changes requested):
It's looking great!

The "Success" status can be misleading here. The retry action succeeded because it didn't throw an error to AS, but still failed to sync and created a new sync retry action. To avoid confusion, we should change the label to "Completed".

Image

miguelpeixe:
As discussed internally, it'd be good to have the user email (if available in the action metadata) surfaced as a table column and item in the details.

src/wizards/audience/views/integrations/logs-view.js
chickenn00dle: Good catch — fixed in 616158082. `ADD_NOTICE` appends without deduping by id, so the running notice now dispatches `removeNotice( noticeId )` immediately before the success/failure/error notice (in both the `.then` and `.catch`), replacing it in place. Updated the comment to reflect the real behavior.

chickenn00dle:
@miguelpeixe re: surfacing the user email — done in 2714a32bb (with a follow-up fix in ca60d161a).

The email is now shown both as a table column in the logs view and as an item in the action details modal. The integration retry actions reference the contact by `user_id` rather than carrying an email directly, so the backend resolves the current account email from that ID (preferring an explicit `email`/`user_email` key when present, and falling back to `previous_email` for email-change retries when the user can no longer be resolved). The list endpoint primes the user cache with a single `cache_users()` query so populating the column doesn't fire a query per row. Rows with no resolvable email render as `—`.

miguelpeixe (approved):
Thank you for revising!

